Issue #186, opened 2019-08-30
Fix syntax classes to use new versions of textX


yguclu:
Since version 2.1 of **textX**, user-defined syntax classes must provide access to their keyword arguments through properties with the same name. This requires a few changes on our side.

yguclu:
Commit 12b12e6 should have solved the issue.

PR #191, opened 2019-09-10
[issue #186] Update attribute names in headers syntax classes.

yguclu:


yguclu:
I suggest using the **Rebase and merge** option.

PR #209, opened 2019-11-20
Add tests for pyccel

EmilyBourne:
including failing test demonstrating import problem. (Missing use statement in fortran code)

### DIFF
--- a/bandit.yml
+++ b/bandit.yml
@@ -3,5 +3,4 @@
 #
 # NOTE: ideally, we would like to only skip this check in our unit tests, but
 #       we do not know if this is possible.
-skips: ['B101']
-skips: ['B603']
+skips: ['B101','B603']
--- a/bandit.yml
+++ b/bandit.yml
@@ -4,3 +4,4 @@
 # NOTE: ideally, we would like to only skip this check in our unit tests, but
 #       we do not know if this is possible.
 skips: ['B101']
+skips: ['B603']
--- a/bandit.yml
+++ b/bandit.yml
@@ -3,4 +3,7 @@
 #
 # NOTE: ideally, we would like to only skip this check in our unit tests, but
 #       we do not know if this is possible.
-skips: ['B101','B603']
+skips:
+  - B101  # Ignore assert statements
+  - B404  # Ignore warnings about importing subprocess
+  - B603  # Ignore warnings about calling subprocess.Popen without shell=True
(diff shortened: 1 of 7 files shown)


tests/run_tests_py3.sh
```diff
@@ -17,4 +17,4 @@ python3 -m pytest tests/epyccel/test_epyccel_functions.py
 python3 -m pytest tests/epyccel/test_epyccel_modules.py
 python3 -m pytest tests/epyccel/test_arrays.py
 python3 -m pytest tests/epyccel/test_loops.py
-
```
yguclu: @EmilyBourne Can you add this line also to file `.travis.yml` in the root directory?

tests/pyccel/test_pyccel.py
```diff
@@ -0,0 +1,39 @@
+import subprocess
```
yguclu: @EmilyBourne Codacy complains about a few things in this file. Can you please have a look at it?

EmilyBourne: Codacy doesn't like the use of subprocess which I use to run pyccel via pytest. I can't think of any other way to do this in python. The alternative would be a bash script but ideally that would require a bash equivalent of pytest. Do you have any suggestions?

yguclu: Yes, there is probably no way around it. But before we give up on this, have you tried using `shell=False`?

EmilyBourne: I've managed to remove the shell=True, but it doesn't seem to be enough to satisfy codacy

yguclu: Thanks for trying! The warnings raised by Codacy are less serious now (notice the orange highlight instead of the red we had previously), and your code looks good to me.

@ratnania: Can we reduce the strictness level of our Codacy analysis?

yguclu: @EmilyBourne: While we wait for Ahmed's input we can skip rule [B603](https://docs.openstack.org/bandit/latest/plugins/b603_subprocess_without_shell_equals_true.html) in file [bandit.yml](https://github.com/pyccel/pyccel/blob/master/bandit.yml).

bandit.yml
```diff
@@ -4,3 +4,4 @@
 # NOTE: ideally, we would like to only skip this check in our unit tests, but
 #       we do not know if this is possible.
 skips: ['B101']
+skips: ['B603']
```
yguclu: No, `skips` should be a list that contains both strings...

yguclu:
I will take care of the last two errors raised by Codacy

PR #210, opened 2019-11-25
Devel f2py compatible

saidctb:


yguclu:
@saidctb Please have a look at Codacy's warnings, we must fix them all in order to merge. In most cases you just have to remove some unused imports. If you have questions on how to fix the other issues, please just ask me.

EmilyBourne:
@saidctb f2py flag does not work as expected. See newly added test

tests/pyccel/test_pyccel.py
```diff
@@ -20,19 +29,40 @@ def get_fortran_output(path_dir,test_file):
...
 
 @pytest.mark.xfail
 def test_imports():
-    pyccel_test("test_imports.py")
+    pyccel_test("test_imports.py","funcs.py")
 
 def test_funcs():
     pyccel_test("test_funcs.py")
+
+def test_f2py_compat():
```
yguclu: @EmilyBourne We should remove any `.so` module from the `path_dir` directory before running this test, otherwise Python will try to import `return_one` from the shared library rather then the pure Python one (possibly raising an `ImportError`). You can achieve this by adding a `teardown_module` function to this file, where you remove the shared library just created.

yguclu: Instead of calling a `teardown` function explicitly, I would prefer to create a `teardown_module` function that gets called automatically by `pytest` (see https://docs.pytest.org/en/latest/xunit_setup.html). A `setup_module` function also makes sense in this case.

Both functions (`setup_module` and `teardown_module`) could call your own `teardown` function. The advantage of using this structure is that `pytest` takes care of calling these functions even in the case of an assertion error.

EmilyBourne: I noticed that the function teardown was being called anyway after every function. I think this is better in case multiple tests want to use the same file. E.g once to test functions then once to test imports by importing the file containing the functions. I haven't seen this functionality documented though, only teardown_module, teardown_function(function), teardown_class

PR #214, opened 2019-12-11
fix issue 198

saidctb:


pyccel/ast/core.py
```diff
@@ -4669,29 +4669,27 @@ def get_assigned_symbols(expr):
...
             elif isinstance(var, IndexedElement):
                 var = var.base
-            free_symbols = var.free_symbols
-            symbols = list(free_symbols)
+                symbols.append(var)
+            return symbols
         except:
 
             # TODO must raise an Exception here
             #      this occurs only when parsing lapack.pyh
```
yguclu: @saidctb Here the TODO suggests to raise an `Exception`, but you added a `pass` statement instead. Why?

saidctb: I don't know what @ratnania  has in mind here so I didn't do anything

yguclu: Have you tried raising an Exception and see if we run into any problems?

PR #217, opened 2020-01-07
Issue 216

EmilyBourne:


setup.py
```diff
@@ -39,7 +39,7 @@
 install_requires = [
     'numpy',
     'scipy',
-    'sympy>=1.2',
+    'sympy==1.4',
```
yguclu: It's better to be less restrictive here: see [setup.py](https://github.com/pyccel/sympde/blob/master/setup.py) in `sympde` library.

EmilyBourne: Sorry I didn't realise you could put two conditions. Fixed now

PR #226, opened 2020-01-21
Import alias

EmilyBourne:
Allow all 4 import syntaxes:
"from module import func1, func2"
"from module import func1 as f1, func2 as f2"
"import module"
"import module as mod"

### DIFF
--- a/pyccel/parser/syntactic.py
+++ b/pyccel/parser/syntactic.py
@@ -465,9 +465,7 @@ class SyntaxParser(BasicParser):
                           bounding_box=stmt.absolute_bounding_box,
                           severity='error')
 
-        scope = stmt
-        while scope.parent != None and not isinstance(scope, (RedBaron, DefNode)):
-            scope = scope.parent
+        scope = stmt.root
 
         # in an import statement, we can have seperate target by commas
         ls = self._visit(stmt.value)
--- a/pyccel/parser/syntactic.py
+++ b/pyccel/parser/syntactic.py
@@ -332,10 +332,24 @@ class SyntaxParser(BasicParser):
             names.append(strip_ansi_escape.sub('', a.value))
 
         if len(names) == 1:
-            return names[0]
+            value = names[0]
         else:
+            value = DottedName(*names)
 
-            return DottedName(*names)
+        if not stmt.target:
+            return value
+
+        old = value
+        new = self._visit(stmt.target)
+
+        # TODO improve
+
+        if isinstance(old, str):
+            old = old.replace("'", '')
+        if isinstance(new, str):
+            new = new.replace("'", '')
+
+        return AsName(new, old)
 
     def _visit_NameAsNameNode(self, stmt):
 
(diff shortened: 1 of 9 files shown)


EmilyBourne:
When using the syntax "import module", all uses of the module in the file are located and module.func is replaced with module_func using redbaron. Would it be better to use a random string instead of module_ for the new name of the function?

pyccel/parser/syntactic.py
```diff
@@ -451,13 +465,56 @@ def _visit_ImportNode(self, stmt):
                           bounding_box=stmt.absolute_bounding_box,
                           severity='error')
 
+        scope = stmt
+        while scope.parent != None and not isinstance(scope, (RedBaron, DefNode)):
```
yguclu: Line 458 checks that `stmt.parent` is an object of type `RedBaron` or `DefNode`, otherwise it creates an error report. Then line 463 creates an error report if the same object is of type `DefNode`. Hence `stmt.parent` must be of type `RedBaron` when we enter this cycle, or am I missing something here?

yguclu:
> When using the syntax "import module", all uses of the module in the file are located and module.func is replaced with module_func using redbaron. Would it be better to use a random string instead of module_ for the new name of the function?

I suggest a combination of the two approaches, by replacing `module.func` with `module_n1abcx_func` where `n1abcx` is an alphanumeric random string.

pyccel/parser/syntactic.py
EmilyBourne: You are correct. I missed this restriction

saidctb:
the problem with doing string manipulations is that we loose the notion of the scoop, for example if we have this two statements  `import mod1 as a`   and  `import mod2 as a` in two different scoops, we will not be able to detect the behaviour of one of the imports.
here is a complete example  where we can see the problem.
```
import math as np
a = np.cos(1)
def f():
    import numpy as np
    b = np.sin([1, 2, 3])

print(a)
print(b)
```
that's why the changes must be done in the semantics part where we have the notion of the scoop and not in the syntactic.
for the other cases where we have a statement of this kind `from numpy import cos as func`
this should work fine

EmilyBourne:
> the problem with doing string manipulations is that we loose the notion of the scoop, for example if we have this two statements `import mod1 as a` and `import mod2 as a` in two different scoops, we will not be able to detect the behaviour of one of the imports.
> here is a complete example where we can see the problem.
> 
> ```
> import math as np
> a = np.cos(1)
> def f():
>     import numpy as np
>     b = np.sin([1, 2, 3])
> 
> print(a)
> print(b)
> ```
> 
> that's why the changes must be done in the semantics part where we have the notion of the scoop and not in the syntactic.
> for the other cases where we have a statement of this kind `from numpy import cos as func`
> this should work fine

This is why I had the lines :
```
    scope = stmt
    while scope.parent != None and not isinstance(scope, (RedBaron, DefNode)):
        scope = scope.parent
```

in order to restrict the changes to the def or the file depending on what the parent was (https://github.com/pyccel/pyccel/pull/226/commits/979274ca1701eeaa334d4ee8e403e7ca9ca2b77c). However I agree that my method wouldn't work in the following case:
```
a = np.cos(1)
def f():
    import numpy as np
    b = np.sin([1, 2, 3])

import math as np
print(a)
print(b)
```

I don't think we need to worry about this though. The statement `import numpy as np` will be treated in the function _visit_ImportNode which raises an error if the parent of the import statement is not RedBaron. This is equivalent to saying that it must have global scope

saidctb:
the second case is okay it should never work , if you manage to fix the first one than it should be okay

EmilyBourne:
Changing to different branch to revert changes and move step to preprocessing

PR #231, opened 2020-01-24
Import preprocess

EmilyBourne:
Allow all 4 import syntaxes:
"from module import func1, func2"
"from module import func1 as f1, func2 as f2"
"import module"
"import module as mod"

### DIFF
--- a/pyccel/parser/syntactic.py
+++ b/pyccel/parser/syntactic.py
@@ -446,6 +446,13 @@ class SyntaxParser(BasicParser):
                       bounding_box=stmt.absolute_bounding_box,
                       severity='error')
 
+        ls = self._visit(stmt.value)
+        ls = get_default_path(ls)
+        expr = Import(ls)
+        expr.set_fst(stmt)
+        self.insert_import(expr)
+        return expr
+
     def _visit_FromImportNode(self, stmt):
 
         if not isinstance(stmt.parent, (RedBaron, DefNode)):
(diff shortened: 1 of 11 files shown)


pyccel/parser/syntactic.py
```diff
@@ -440,25 +441,6 @@ def _visit_NameNode(self, stmt):
             val = strip_ansi_escape.sub('', stmt.value)
             return Symbol(val)
 
-    def _visit_ImportNode(self, stmt):
```
saidctb: can you put back this method and raise an error if we use it instead of deleting it

pyccel/parser/syntactic.py
```diff
@@ -441,23 +442,9 @@ def _visit_NameNode(self, stmt):
...
-        # in an import statement, we can have seperate target by commas
-        ls = self._visit(stmt.value)
-        ls = get_default_path(ls)
-        expr = Import(ls)
-        expr.set_fst(stmt)
-        self.insert_import(expr)
-        return expr
+        errors.report(PYCCEL_UNEXPECTED_IMPORT,
+                      bounding_box=stmt.absolute_bounding_box,
+                      severity='error')
```
saidctb: can you also add this part  : 
ls = self._visit(stmt.value)
ls = get_default_path(ls)
expr = Import(ls)
expr.set_fst(stmt)
self.insert_import(expr)
return expr
because we continue converting even if we have an error

saidctb: can you also add this part :
ls = self._visit(stmt.value)
ls = get_default_path(ls)
expr = Import(ls)
expr.set_fst(stmt)
self.insert_import(expr)
return expr
because we continue converting even if we have an error

PR #234, opened 2020-02-03
Reorganize Pyccel build process into unique pipeline

yguclu:


EmilyBourne:
epyccel does not work for modules if called from the folder containing the module

Issue #236, opened 2020-02-18
Problem with conversion and printing of non-square arrays


krystophny:
The following code seems to produce wrong matrices and printout in the Fortran version. Some things with row and columns get mixed up in the initialization apparently.

Python:
```python
from numpy import array
A = array([[1., 2., 3.], [4., 5., 6.], [7., 8., 9.]])
B = array([[1., 2.], [4., 5.], [7., 7.]])
C = array([[1., 2., 3.], [4., 5., 6.]])
print("=== A ===")
print(A[0, :])
print(A[1, :])
print(A[2, :])
print("=== B ===")
print(B[0, :])
print(B[1, :])
print(B[2, :])
print("=== C ===")
print(C[0, :])
print(C[1, :])
```
Output:
```
=== A ===
[1. 2. 3.]
[4. 5. 6.]
[7. 8. 9.]
=== B ===
[1. 2.]
[4. 5.]
[7. 7.]
=== C ===
[1. 2. 3.]
[4. 5. 6.]
```

Pyccelized Fortran:
```Fortran
allocate(A(0:2, 0:2))
A = reshape((/ 1.0d0, 2.0d0, 3.0d0, 4.0d0, 5.0d0, 6.0d0, 7.0d0, 8.0d0, &
      9.0d0 /),(/ 3, 3 /))
allocate(B(0:2, 0:1))
B = reshape((/ 1.0d0, 2.0d0, 3.0d0, 4.0d0, 5.0d0, 6.0d0 /),(/ 3, 2 /))
allocate(C(0:1, 0:2))
C = reshape((/ 1.0d0, 2.0d0, 3.0d0, 4.0d0, 5.0d0, 6.0d0 /),(/ 2, 3 /))
print *, "=== A ==="
print *, A(:, 0)
print *, A(:, 1)
print *, A(:, 2)
print *, "=== B ==="
print *, B(:, 0)
print *, B(:, 1)
print *, B(:, 2)
print *, "=== C ==="
print *, C(:, 0)
print *, C(:, 1)
```
Output:
```
 === A ===
   1.0000000000000000        2.0000000000000000        3.0000000000000000     
   4.0000000000000000        5.0000000000000000        6.0000000000000000     
   7.0000000000000000        8.0000000000000000        9.0000000000000000     
 === B ===
   1.0000000000000000        2.0000000000000000        3.0000000000000000     
   4.0000000000000000        5.0000000000000000        6.0000000000000000     
[...]

PR #237, opened 2020-02-18
Devel numpyext

yguclu:
This fixes issue #236 and (temporarily) #243


EmilyBourne:
At least some of the errors are coming from the change in the default integer precision. It is changed for the arguments. But not for the dummy variables or any local variables. Their definition seems to be done at line 2070 of ast/core.py. But if I modify that to use the default_precision dictionary then the function call is a problem (see tests/pyccel/scripts/test_funcs.py for example code):
 ```
print *, sum_to_n(4)
        1
Error: Type mismatch in argument 'n' at (1); passed INTEGER(4) to INTEGER(8)
```
My impression is that fixing the problem in ast/core.py is enough. But there seem to be quite a lot of other places where the precision is set manually without checking the dictionary. The ones I found (for ints, but I think real/complex are mostly in the same place) are:
parser/syntax/headers.py: lines 62, 106,
parser/semantic.py: lines 646, 901, 963

I will push a commit with this fix, but it won't fix the problem for literal types. Something like this will need to be written instead:
`print *, sum_to_n(4_8)`

yguclu:
> At least some of the errors are coming from the change in the default integer precision. It is changed for the arguments. But not for the dummy variables or any local variables. Their definition seems to be done at line 2070 of ast/core.py. But if I modify that to use the default_precision dictionary then the function call is a problem (see tests/pyccel/scripts/test_funcs.py for example code):
> 
> ```
> print *, sum_to_n(4)
>        1
> Error: Type mismatch in argument 'n' at (1); passed INTEGER(4) to INTEGER(8)
> ```
> 
> My impression is that fixing the problem in ast/core.py is enough. But there seem to be quite a lot of other places where the precision is set manually without checking the dictionary. The ones I found (for ints, but I think real/complex are mostly in the same place) are:
> parser/syntax/headers.py: lines 62, 106,
> parser/semantic.py: lines 646, 901, 963
> 
> I will push a commit with this fix, but it won't fix the problem for literal types. Something like this will need to be written instead:
> `print *, sum_to_n(4_8)`

@EmilyBourne Thanks for the bugfix. I am going to look into literal constants now.

yguclu:
All unit tests are passing now.

yguclu:
I am waiting for some feedback from @ratnania before merging this into master.

yguclu:
@EmilyBourne You can use [`shutil`](https://docs.python.org/3/library/shutil.html#module-shutil) to remove the directories

Issue #241, opened 2020-02-20
Booleans fail


krystophny:
Trying the following with epyccel/pyccel fails with a gfortran compile error. Reason is that it tries to set logical(kind=0) which doesn’t work in Fortran. Opened branch `devel-base` with a test case in `tests/epyccel/test_base.py`.

```python
@types('bool', 'bool')
def cmp_bool(a, b):
    c = False
    if a == b:
        c = True
    return c
```

produces

```fortran
function cmp_bool(a, b) result(c)

  implicit none
  logical(kind=0) :: c !!!!!!!!!! HERE IT FAILS !!!!!!!!!! 
  logical(kind=1), intent(in)  :: a 
  logical(kind=1), intent(in)  :: b 

  c = .False.
  if (a == b ) then
    c = .True.
  end if
  return
end function
```

krystophny:
Another problem: printing `a == b` done currently doesn't work, Fortran requires `a .eq. b` instead.

yguclu:
Actually, since Fortran 90/95 `==` is accepted and much clearer than `.eq.`. Maybe somewhere we pass the flag `-std=f77` to the compiler?

krystophny:
Logicals are still different. They need `.eqv.` and `.neqv.` operators, and don't even work with `.eq.` which is now `==`.

yguclu:
Yes, I see your point now! Sorry for the confusion.

krystophny:
...also a is b and a is True won't work as a workaround. I'm getting the error
```
NotImplementedError: Only None rhs is allowed in Is statement
```

krystophny:
In `_visit_ComparisonNode(self, stmt)` we have a lot of property flags in first and second arguments. These are just set as atomic symbols with all other flags on false. Could one somehow set `is_boolean flag` to `True` or will that break something?

... or do these flags mean something different and should all be false if first and second are set as symbols with `is_atom=True` ?

krystophny:
Update: equality and inequality `==`, `!=` and `is` are working in `devel-base` with unit tests. What is failing now still is the `is not` clause, probably because I did something wrong / too general in https://github.com/pyccel/pyccel/blob/50eb37989adcfde5cc900168f3512380ecda9cba/pyccel/parser/semantic.py#L2770-L2792

yguclu:
@krystophny @EmilyBourne Can we close this issue?

EmilyBourne:
Ok with me

krystophny:
Closed

PR #242, opened 2020-02-21
Boolean issues

krystophny:
#241 - not finished until `is not` is working


pyccel/codegen/printing/fcode.py
```diff
@@ -1991,11 +2005,17 @@ def _print_Assert(self, expr):
...
         lhs = self._print(expr.lhs)
-        return 'present({})'.format(lhs)
+        rhs = self._print(expr.rhs)
+        a = expr.args[0]
+        b = expr.args[1]
+        if ((a.is_Boolean or isinstance(a.dtype, NativeBool)) and
+            (b.is_Boolean or isinstance(b.dtype, NativeBool))):
+            return '{} .eqv. {}'.format(lhs, rhs)
+        if isinstance(expr.rhs, Nil):
+            return 'present({})'.format(lhs)
```
EmilyBourne: This should be tested

pyccel/ast/datatypes.py
```diff
@@ -8,7 +8,7 @@
 from sympy import Eq, Ne, Lt, Gt, Le, Ge
 
 
-default_precision = {'real': 8, 'int': 8, 'complex': 8, 'bool':1, 'float':8}
+default_precision = {'real': 8, 'int': 8, 'complex': 8, 'bool':4, 'float':8}
```
yguclu: Interesting that we should use 4 Bytes to store a boolean variable. Can you explain why we need this?

EmilyBourne: If you have a function with an argument of type logical(kind=1) and you pass .True. then gfortran will complain that the types don't match

yguclu: What worries me is that we have no way to determine the kind parameter value in Fortran. For integers there is `selected_int_kind`, for floats `selected_real_kind` and for strings `selected_char_kind`. But for booleans there is no equivalent function.

yguclu: If we want to ensure compatibility between integer and logical variables, maybe these should have the same kind. But I say "maybe" here! I suspect that we should always print logicals without specifying their kind, hence using the default kind provided by the machine. I wonder how f2py handles this issue.

yguclu: @chris @EmilyBourne Thanks for all this work!
You need to fix some small Codacy issue before we can proceed with this merge.
And I am also waiting for @saidctb approval for this.

krystophny: I think C handles it the same way. There is no smaller unit than 4 Bytes on 32bit x86 and they kept it for x64 also probably for compatibility. So internally you always reserve an integer when you want a bool.

PR #246, opened 2020-02-25
Add coverage test requiring 80% coverage

EmilyBourne:


EmilyBourne:
Fails. Coverage is currently at 60%.
Should stdlib/external be removed from coverage?

EmilyBourne:
Calls to pyccel are not managed by coverage as they occur in subprocesses

PR #247, opened 2020-02-25
Coverage

EmilyBourne:


EmilyBourne:
Fails. Coverage is currently at 62%.
Should stdlib/external be removed from coverage?

.travis.yml
```diff
@@ -58,7 +64,12 @@ script:
...
 #  - python ../tests/test_pyccel_lapack.py --libs='blas lapack' --execute
 #  - python ../tests/test_pyccel_mpi.py --compiler=mpif90
+  - coverage combine
+  - coverage xml
+  - rm ${SITE_DIR}/pyccel_cov.pth
   - cd -
 
 after_script:
   - rm -rf __pytest__
+  - export CODACY_PROJECT_TOKEN=d5e147769dad48158549d7236cb5d53a
```
krystophny: This one should not be written directly, but set as a protected environment variable in Travis CI. See https://travis-ci.org/pyccel/fffi/settings for an example.

.travis.yml
```diff
@@ -25,6 +25,7 @@ before_install:
 #  - sudo apt-get install libcr-dev mpich2
 #  - sudo apt-get install mpich libmpich-dev # libhdf5-mpich-dev
   - sudo apt-get install libopenmpi-dev openmpi-bin # libhdf5-openmpi-dev
+  - sudo apt-get install default-jre
```
krystophny: One could avoid installing a full JRE by just loading `bash <(curl -Ls https://coverage.codacy.com/get.sh)` directly from codacy, see https://github.com/pyccel/fffi/blob/master/.travis.yml

EmilyBourne: Interesting
I added that line because when I ran `bash <(curl -Ls https://coverage.codacy.com/get.sh)` locally on my computer it complained that I had not installed java.
That line was therefore necessary to satisfy the requirement : Java JRE 8 and higher
stated on https://github.com/codacy/codacy-coverage-reporter
Presumably travis already has a version of java installed somewhere which allows you to remove this line.

Do you think it is better to remove this line to avoid installing unnecessary packages or to leave it in case travis changes its internal packages at some point?

.travis.yml
EmilyBourne: @yguclu 
I suspect I don't have the rights to do this. When I try to set the variable I get a message:
`Oh no! There was an error saving this environment variable.`
If I try again it tells me "Variable with this name is already defined."
But, given the output from travis I am sure that the variable isn't defined

krystophny: Now that it's in the Git history anyway, doesn't matter ;-) but for the future this would be the clean way. In my opinion we can leave it in.

.travis.yml
krystophny: Oh, in that case could be that really travis has it. Then I would leave  line inside, as it doesn't cost anything. For some reason I added also `apt install jq` in fffi, as this was mentioned somewhere - but it's not needed. So everything OK also here!

.travis.yml
EmilyBourne: In that case this change either needs to be rolled back or Yaman needs to add the environment variable before carrying out the merge

yguclu: @krystophny We should probably have asked you before doing this ;-) The Codacy documentation that we read did not mention Travis unfortunately. I suppose I can revoke that token and create a new one?

yguclu: Now this is what we should have read: https://support.codacy.com/hc/en-us/articles/207279819-Coverage

setup.py
```diff
@@ -49,6 +49,7 @@
     'mpi4py',
     'termcolor',
     'fastcache',
+    'coverage',
```
yguclu: Why do we need to add `coverage` to our dependencies? Pyccel can be installed without coverage, and all tests can also be run without it, correct?

.travis.yml
yguclu: I have revoked the previous Codacy API token, created a new one, and uploaded it to Travis.

krystophny: That’s good, this way we don’t have to rewrite history (else this should be done before merging to master)

setup.py
krystophny: Agreed

EmilyBourne: You're right. We should move the installation to the travis script. Should it be in the install section or the before_script section (or the before_install section) ?

yguclu: I would put it in `before_script` as it's only needed by the tests.

PR #249, opened 2020-02-27
add recursive functions for subroutines and pass intent(in) by value

saidctb:


EmilyBourne:
> Please add proper unit tests using `pytest`, both in `tests/pyccel` and `tests/epyccel`. Note that we may already have some tests with recursive functions, currently marked with `pytest.mark.xfail`.

The failing hop benchmark tests are all recursive functions

tests/pyccel/scripts/hope_benchmarks_decorators/fib.py
```diff
@@ -1,6 +1,6 @@
 from pyccel.decorators import types
 
-@types('int')
+@types('int', results='int')
```
yguclu: @saidctb I understand that recursive functions need the `results` type to be explicitly provided by the user. I then have two questions:
1. What happens if the user does **not** provide `results` for a recursive function?
2. What happens if the user provides `results` but the function is **not** recursive?

Do you detect the errors? If so, what exceptions do you raise? I think you need to write two unit tests to check the cases above, and verify that the expected behavior is obtained (see https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions).

pyccel/parser/semantic.py
```diff
@@ -2786,20 +2779,18 @@ def _visit_FunctionDef(self, expr, **settings):
...
-               funcs = list(funcs.funcs)+[vec_func]
-           else:
-               self.namespace.sons_scopes['sc_'+ name] = self.namespace.sons_scopes[name]
-               funcs = funcs.rename('sc_'+ name)
-               funcs = [funcs, vec_func]
-
-           funcs = Interface(name, funcs)
-           self.insert_function(funcs)
-
+#        TODO move this to codegen
```
yguclu: Can we actually move this to codegen within this pull request?

pyccel/parser/semantic.py
```diff
@@ -2532,26 +2531,26 @@ def _visit_FunctionDef(self, expr, **settings):
...
-            index     = create_variable(expr.body)
-            range_    = Function('range')(Function('len')(arg))
-            args      = symbols(args)
-            args[index_arg] = vec_arg[index]
-            body_vec        = Assign(args[index_arg], Function(name)(*args))
-            body_vec.set_fst(expr.fst)
-            body_vec   = [For(index, range_, [body_vec], strict=False)]
-            header_vec = header.vectorize(index_arg)
-            vec_func   = expr.vectorize(body_vec, header_vec)
+#        TODO move this to codegen
```
yguclu: Can we actually move this to codegen within this pull request?

tests/pyccel/scripts/hope_benchmarks_decorators/fib.py
saidctb: 1. If the user doesn't provide us the return type, then we can't deduce the return type, we should improve this in the future so that we can deduce the return type automatically.
2. In the case of a non recursive function ,the return type provided by the user doesn't matter, so it will be ignored for the moment.
for the first case we don't raise a proper exception error , so yes we should add it , as for the second case it will work fine and i don't think we should raise an error

pyccel/parser/semantic.py
saidctb: I don't know if we should provide that for the moment as it was never working from the beginning, it was generating the right code but the behaviour in python and fortran was different

tests/pyccel/scripts/hope_benchmarks_decorators/fib.py
EmilyBourne: If the results type is completely ignored, it might be a good idea to raise a warning even if we don't raise an error. It could lead to unexpected behaviour for the user (e.g. function returns an int but the user specified a real). It's not a big problem, but a warning could be useful so the user is aware

pyccel/parser/semantic.py
yguclu: > I don't know if we should provide that for the moment as it was never working from the beginning, it was generating the right code but the behaviour in python and fortran was different

I see. Then please create a dedicated "enhancement" issue and mention it here.

tests/pyccel/scripts/hope_benchmarks_decorators/fib.py
yguclu: In case 2 (user provides results for non-recursive function) I suggest that we simply check if the provided type is the same as the inferred type. If not we should raise an exception.

PR #250, opened 2020-03-02
Devel ast

saidctb:


pyccel/codegen/printing/pycode.py
```diff
@@ -190,6 +190,10 @@ def _print_Zeros(self, expr):
     def _print_ZerosLike(self, expr):
         return 'zeros_like('+ self._print(expr.rhs)+')'
 
+    def _print_Max(self, expr):
+        args = ','.join(self._print(e) for e in expr.args)
+        return 'max({})'.format(args)
+
```
EmilyBourne: If the _print_Max function is added then it would make sense to also add a _print_Min function at the same time

yguclu: I agree.

yguclu: @saidctb And maybe you could add an empty space after the comma, in order to avoid the `pip8` module to complain after the violation of rule
```
E231 missing whitespace after ‘,’
```

saidctb: the requested changes have been added

PR #257, opened 2020-03-11
Remove mpi4py dependency

### LINKED ISSUE #255: Make mpi4py requirement optional?
Since not everyone intends to use MPI, it would be nice to have mpi4py requirement optional. This would also allow for easier use on Windows (see #194) without requiring a compiler and and installing MSMPI. Related to #251 .

yguclu:
closes #255

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ install:
   - pip install .
 
 before_script:
+  - pip install mpi4py
   - pip install coverage
   - mkdir __pytest__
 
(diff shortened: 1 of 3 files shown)


EmilyBourne:
If we remove dependencies that are only relevant for the tests should we not also remove pytest?

krystophny:
pytest should also be removed as a dependency in my opinion, as it’s not needed to run pyccel

yguclu:
Yes, we could remove `pytest` from our dependencies. But in general I would like the user to always run the tests on his/her machine after installing the library, so I was a little reluctant to remove pytest.

yguclu:
So I have also removed `pytest` from our dependencies and updated the README file. I will proceed with merging.

PR #262, opened 2020-03-12
Multiple function outputs

### LINKED ISSUE #203: When functions have multiple outputs the order is not conserved
The order of the out parameters is not deterministic for functions returning tuples.

For example calling pyccel on the a file containing the following code:
```
from pyccel.decorators import external, pure

@external
@pure
def get_2_3():
    return 2,3
```

can generate both of the following versions of fortran code :

```
module tmp

implicit none




contains

pure subroutine get_2_3(Dummy_117, Dummy_76)

implicit none
integer(kind=4), intent(out)  :: Dummy_117
integer(kind=4), intent(out)  :: Dummy_76

Dummy_117 = 2
Dummy_76 = 3
return
end subroutine

end module
```
(where the out parameters are in the original order and the code provides the same result when run through f2py)

and :
```
module tmp

implicit none




contains

pure subroutine get_2_3(Dummy_357, Dummy_255)

implicit none
integer(kind=4), intent(out)  :: Dummy_357
integer(kind=4), intent(out)  :: Dummy_255

Dummy_255 = 2
Dummy_357 = 3
return
end subroutine

end module
```
(where the values are returned in the wrong order)

yguclu:
closes #203, but some unit tests are needed.

### DIFF
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -730,13 +730,10 @@ class CodeBlock(Basic):
         ls = []
         for i in body:
             if isinstance(i, CodeBlock):
-                ls += i.body
-            elif isinstance(i, (Assign, For, AugAssign, FunctionalFor,
-                            Application, Expr, IfTernaryOperator, EmptyLine)):
-                ls.append(i)
+                ls += list(i.body)
             else:
+                ls.append(i)
 
-                raise TypeError('statement of type {} not supported yet'.format(type(i)))
         obj = Basic.__new__(cls, ls)
         if isinstance(ls[-1], (Assign, AugAssign)):
             obj.set_fst(ls[-1].fst)
@@ -1566,6 +1563,7 @@ class Program(Basic):
 
         if not iterable(body):
             raise TypeError('body must be an iterable')
+        body = CodeBlock(body)
 
         if not iterable(classes):
             raise TypeError('classes must be an iterable')
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -1685,11 +1685,12 @@ class For(Basic):
             if not cond_iter:
                 raise TypeError('iter must be an iterable')
 
-            if not iterable(body):
-                raise TypeError('body must be an iterable')
+            if iterable(body):
+                body = CodeBlock((sympify(i, locals=local_sympify) for i in
+                             body))
+            elif not isinstance(body,CodeBlock):
+                raise TypeError('body must be an iterable or a Codeblock')
 
-            body = Tuple(*(sympify(i, locals=local_sympify) for i in
-                         body), sympify=False)
         return Basic.__new__(cls, target, iter, body, local_vars)
 
     @property
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -4462,9 +4462,13 @@ class If(Basic):
             if not isinstance(cond, (bool, Relational, Boolean, Is, IsNot)):
                 raise TypeError('Cond %s is of type %s, but must be a Relational, Boolean, Is, IsNot, or a built-in bool.'
                                  % (cond, type(cond)))
-            if not isinstance(ce[1], (list, Tuple, tuple)):
-                raise TypeError('body is not iterable')
-            newargs.append(ce)
+            if isinstance(ce[1], (list, Tuple, tuple)):
+                body = CodeBlock(ce[1])
+            elif isinstance(ce[1], CodeBlock):
+                body = ce[1]
+            else:
+                raise TypeError('body is not iterable or CodeBlock')
+            newargs.append((cond,body))
 
         return Basic.__new__(cls, *newargs)
 
@@ -4472,7 +4476,7 @@ class If(Basic):
     def bodies(self):
         b = []
         for i in self._args:
-            b += i[1]
+            b.append( i[1])
         return b
 
 
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -1030,9 +1030,10 @@ class While(Basic):
     def __new__(cls, test, body, local_vars=[]):
         test = sympify(test, locals=local_sympify)
 
-        if not iterable(body):
-            raise TypeError('body must be an iterable')
-        body = Tuple(*(sympify(i, locals=local_sympify) for i in body),sympify=False)
+        if iterable(body):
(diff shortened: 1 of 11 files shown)


tests/pyccel/test_pyccel.py
```diff
@@ -82,26 +83,56 @@ def teardown(path_dir = None):
         elif not f.endswith(".py"):
             os.remove(file_name)
 
+def compare_pyth_fort_output_by_type( p_output, f_output, dtype=float ):
+
+    if dtype is str:
+        assert(p_output.strip()==f_output.strip())
+    elif dtype is complex:
+        p_output, f_output = compare_pyth_fort_output_by_type( p_output, f_output, float)
+        p_output, f_output = compare_pyth_fort_output_by_type( p_output, f_output, float)
```
yguclu: Can you explain why you call the function twice with the same arguments? Do the two calls process the real and imaginary parts, respectively?

tests/pyccel/test_pyccel.py
```diff
@@ -82,26 +83,56 @@ def teardown(path_dir = None):
...
+        f = f_match.group().lower() in ['true', 't', '1']
+        assert(p==f)
+        p_output=p_output[p_match.span()[1]:]
+        f_output=f_output[f_match.span()[1]:]
+
+    elif dtype is float or dtype is int:
+        if dtype is float:
+            rx = re.compile('[-0-9.eE]+')
+        elif dtype is int:
+            rx = re.compile('[-0-9eE]+')
```
yguclu: Why do you include the characters `eE` in the regular expression?

tests/pyccel/scripts/test_multiple_results.py
```diff
@@ -0,0 +1,72 @@
...
+print(max(f3()))
+
+
+@types('real','real')
+def f4(x, y = 2.5):
+    x = x + y
+    return x+y, x-y
+
+for k in range(2):
+    print(f4(i,j))
```
yguclu: Did you intend to have the loop change the function arguments? If this is the case, I cannot see how that works here.

tests/pyccel/test_pyccel.py
EmilyBourne: Yes exactly. I process 2 floats so the real then imaginary parts of a complex

tests/pyccel/test_pyccel.py
EmilyBourne: I didn't at first but something was printing in scientific notation (ie 2.0E-02)

tests/pyccel/scripts/test_multiple_results.py
EmilyBourne: No it just needs to change something so that the difference can be seen if the call is before the for loop rather than in it

tests/pyccel/test_pyccel.py
yguclu: Even the integers?

tests/pyccel/scripts/test_multiple_results.py
yguclu: I see.

pyccel/ast/core.py
```diff
@@ -730,15 +730,12 @@ def __new__(cls, body):
         ls = []
         for i in body:
             if isinstance(i, CodeBlock):
-                ls += i.body
-            elif isinstance(i, (Assign, For, AugAssign, FunctionalFor,
-                            Application, Expr, IfTernaryOperator, EmptyLine)):
-                ls.append(i)
+                ls += list(i.body)
```
yguclu: I cannot see why one needs to convert `i.body` to a list. In order to extend a list (by either using `+=` or calling the `extend` method) it is sufficient that the argument is an iterable object.

tests/pyccel/test_pyccel.py
EmilyBourne: I don't know. I thought it was better to put it in for things like 2e2 just in case rather than leave it out and have it fail later. But we can remove it if you don't think it's necessary

pyccel/ast/core.py
EmilyBourne: @saidctb advised doing this as he wants `i.body` to become a tuple. But if `+=` works with a tuple then I agree that this isn't necessary

tests/pyccel/test_pyccel.py
yguclu: You might be right, it depends on the string formatting that one uses after all. We will improve this in the future if need be.

pyccel/ast/core.py
saidctb: I didn't know that it works with a tuple , then yes we don't need to convert it

pyccel/parser/semantic.py
```diff
@@ -1095,7 +1097,7 @@ def _infere_type(self, expr, **settings):
             return self._infere_type(expr.value)
 
         elif isinstance(expr, IfTernaryOperator):
```
yguclu: Is an `IfTernaryOperator` the right-hand side expression of
```
a = 0 if b is None else 1
```
?

And if so, do we have unit tests for it?

pyccel/ast/core.py
yguclu:
```
In [1]: a = list()

In [2]: a.extend?                                                                                                  
Docstring: L.extend(iterable) -> None -- extend list by appending elements from the iterable
Type:      builtin_function_or_method
```

pyccel/parser/semantic.py
EmilyBourne: Yes exactly. There are some tests in tests/codegen/scripts/Functional_Stmts.py

yguclu: I know that this is not relevant to this pull request, but we should maybe check if this really works with pyccel/epyccel. If not we should create an issue.

pyccel/parser/semantic.py
```diff
@@ -1703,6 +1707,102 @@ def _visit_Max(self, expr, **settings):
         args = self._visit(expr.args, **settings)
         return Max(*args)
 
+    def _visit_lhs_Assign(self, lhs, d_var, rhs, **settings):
```
yguclu: I like the logic implemented in this function, but I think we should change its name.

Every other method in class `SemanticParser` follows the naming convention `_visit_TypeName` where `TypeName` is indeed the name of the type of the argument. But we do not have a type `lhs_Assign`, so I think this should be called differently.

yguclu:
@EmilyBourne: Excellent job! We can merge this now

PR #266, opened 2020-03-16
Devel ast

saidctb:


EmilyBourne:
We should add the tests for all the numpy and scipy functions too. So far I have only added the math functions

EmilyBourne:
Tests with shape are broken. Shape returns a tuple. We should therefore fix the tuples as per issue #271 first

EmilyBourne:
The following functions have been marked as xfail but should be fixed before completing the pull request:
- [ ] test_tan_call
- [ ] test_tan_phrase

EmilyBourne:
Several return_type tests are also failing. We do not always return the expected type. These tests are:
- test_fabs_return_type
- test_sqrt_return_type
- test_floor_return_type

pyccel/parser/semantic.py
```diff
@@ -43,56 +33,38 @@
...
+from pyccel.ast import Subroutine
+from pyccel.ast import Zeros, Where, Linspace, Diag, EmptyLike
 from pyccel.ast import StarredArguments
 from pyccel.ast import inline, subs, create_variable, extract_subexpressions
 from pyccel.ast.core import get_assigned_symbols
 
-from pyccel.ast.core      import local_sympify, int2float, Pow, _atomic
+from pyccel.ast.core      import Pow, Add, Mul, And, Or, _atomic
+from pyccel.ast.core      import Eq, Ne, Lt, Le, Gt, Ge
+from pyccel.ast.core      import BooleanTrue, BooleanFalse
```
EmilyBourne, line 59: This line is new but it is masked by
```
from sympy.logic.boolalg       import BooleanTrue, BooleanFalse
```
Before it is used

yguclu: So are `BooleanTrue` and `BooleanFalse` defined in pyccel or sympy or both?

EmilyBourne: Both (see line 219 of ast/core.py)

yguclu: As the objective of this PR is to replace native Sympy objects with Pyccel AST nodes, I suppose that we should try to reduce the number of modules that import Sympy. Here I would like `pyccel.parser.semantic` not to use Sympy at all.

@saidctb @EmilyBourne Do you agree?

EmilyBourne:
TODO (@saidctb ) :
- [ ] Fix tan(x)
- [ ] Remove as many sympy imports as possible
     - [ ] min
     - [ ] max
     - [ ] ceiling

pyccel/parser/semantic.py
saidctb: done

EmilyBourne:
The windows tests are failing because although the function argument is correctly marked as `integer(4)` the value passed is an `integer(8)`
See error:
```
C:\Users\travis\build\pyccel\pyccel\tests\pyccel\project_rel_imports\project\folder2\__pyccel__\mod3.f90:20:13:
 Dummy_3736 = add2(100_8, sum_to_n_squared(n))
             1
Error: Type mismatch in argument 'x' at (1); passed INTEGER(8) to INTEGER(4)
```

EmilyBourne:
The failing test was due to incorrect code in codegen/printing/fcode.py:
```
 def _print_Integer(self, expr):
        printed = CodePrinter._print_Integer(self, expr)
        return "%s_8" % printed

def _print_Zero(self, expr):
        return "0_8"
```
We should print the precision of the Integer however when I make this change there are lots of errors that appear. This is due to the fact that carrying out any operations on a pyccel Integer gives a sympy Integer. This occurs for example in _print_Range:
```
def _print_Range(self, expr):
        stop  = self._print(expr.stop-1)
        step  = self._print(expr.step)
        return '{0}, {1}, {2}'.format(start, stop, step)
```
`expr.stop` is a pyccel Integer, but `expr.stop-1` is a sympy Integer

yguclu:
@EmilyBourne Does this mean that we have a name clash in the Fortran printer, which uses the name of an object's class to choose which print function to call? It seems to me that `_print_Integer` will be used to print both `sympy.Integer` and `pyccel.Integer`...

EmilyBourne:
@yguclu Yes exactly, but I think we shouldn't be trying to print sympy.Integer. If we are it is because something has not been treated correctly

pyccel/ast/core.py
```diff
@@ -177,7 +181,113 @@ def __init__(self, var):
...
+        expr = Pow(args[0], args[1], evaluate=False)
+        return expr
+
+    def _eval_evalf(self,prec):
+        return sp_Pow(self.base,self.exp).evalf(prec)
+
+    @property
+    def rank(self):
+        # TODO: Use broadcasting rules to decide shape (https://docs.scipy.org/doc/numpy/user/basics.broadcasting.html)
+        return max(getattr(a,'rank',0) for a in self._args)
```
EmilyBourne, line 201: There is already a Pow class defined in this file. Which Pow should we be using?

EmilyBourne:
The following diff would fix all bugs. However it would be better to fix the bugs before applying these changes as they make the internal problems invisible:
```
diff --git a/pyccel/codegen/printing/fcode.py b/pyccel/codegen/printing/fcode.py
index cb02385c..17af7bbf 100644
--- a/pyccel/codegen/printing/fcode.py
+++ b/pyccel/codegen/printing/fcode.py
@@ -50,6 +50,7 @@ from pyccel.ast.datatypes import NativeSymbol, NativeString
 from pyccel.ast.datatypes import NativeInteger
 from pyccel.ast.datatypes import NativeRange, NativeTensor, NativeTuple
 from pyccel.ast.datatypes import CustomDataType
+from pyccel.ast.datatypes import default_precision
 from pyccel.ast.numbers import Integer, Float
 
 from pyccel.ast import builtin_import_registery as pyccel_builtin_import_registery
@@ -2209,16 +2210,16 @@ class FCodePrinter(CodePrinter):
         return "%sd0" % printed
 
     def _print_Integer(self, expr):
-        return "{0}_{1}".format(str(expr.p), expr.precision)
+        return "{0}_{1}".format(str(expr.p), default_precision['int'])
 
     def _print_Zero(self, expr):
-        return "0_{0}".format(expr.precision)
+        return "0_{0}".format(default_precision['int'])
 
     def _print_One(self, expr):
-        return "1_{0}".format(expr.precision)
+        return "1_{0}".format(default_precision['int'])
 
     def _print_NegativeOne(self, expr):
-        return "-1_{0}".format(expr.precision)
+        return "-1_{0}".format(default_precision['int'])
 
     def _print_IndexedBase(self, expr):
         return self._print(expr.label)
```

EmilyBourne:
@EmilyBourne 
The remaining problem is due to IndexedBase.__getitem__() evaluating its argument. The fix for this requires an additional class to be created

yguclu:
In **pyccel.ast.core** we have classes
- `IndexedVariable(IndexedBase, PyccelAstNode)`, and
- `IndexedElement(Indexed, PyccelAstNode)`.

I suppose we should use those classes instead of the Sympy ones?

EmilyBourne:
@yguclu 
Exactly. I think we just need to remove these lines:
https://github.com/pyccel/pyccel/blob/master/pyccel/ast/core.py#L4356-L4361
But I will look into it tomorrow

yguclu:
On Windows many x-failing tests are now passing!
We should remove the `pytest.mark.xfail` annotation from them.

pyccel/ast/basic.py
```diff
@@ -1,6 +1,6 @@
 from sympy.core.basic import Basic as sp_Basic
 
-__all__ = ('Basic',)
+__all__ = ('Basic', 'PyccelAstNode')
```
yguclu, line 3: @saidctb: Now that we have a `PyccelAstNode` class, do we still need a separate `Basic` class? It is not clear to me why we need some Pyccel AST nodes to inherit from one class, and some from the other. Maybe it is just a matter of giving `Basic` a more meaningful name.

saidctb: it's not used yet, that's why we should keep it separate for the moment until we validate it
and set its attribute, we  then can merge it with Basic

saidctb: the attributes are not used that's what I mean
if you prefer we can wait until we validate the type inference mechanism with it

pyccel/ast/core.py
```diff
@@ -177,7 +177,75 @@ def __init__(self, var):
...
+        args_ = [self.base._subs(old, new),self.exp._subs(old, new)]
+        args  = [args_[i] if args_[i] else args[i] for i in range(len(args))]
+        expr = Pow(args[0], args[1], evaluate=False)
+        return expr
+
+    def _eval_evalf(self,prec):
+        return sp_Pow(self.base,self.exp).evalf(prec)
+
+    @property
+    def rank(self):
```
yguclu, line 197: Given that `Pow` inherits the `rank` property from `PyccelAstNode`, I cannot see a good reason to override it here. It seems more natural to me to compute the correct value of `self._rank` when we initialize a `Pow` object in the `__init__` function, which is called after `__new__`:
```
 def __init__(cls, *args, **kwargs):
    self._rank = max(getattr(a,'rank',0) for a in self._args)
    # TODO: Use broadcasting rules to compute self._shape
```
@EmilyBourne, @saidctb: What do you think?

EmilyBourne: I agree

EmilyBourne: I agree

pyccel/ast/datatypes.py
```diff
@@ -77,23 +78,24 @@
 )
 
 #==============================================================================
-default_precision = {'real': 8, 'int': 8, 'complex': 8, 'bool':4, 'float':8}
-dtype_and_precision_registry = { 'real':('real',8),
-                                 'double':('real',8),
-                                 'float':('real',8),       # sympy.Float
-                                 'pythonfloat':('real',8), # built-in float
+
+default_precision = {'real': 8, 'int': numpy.dtype(int).alignment, 'complex': 8, 'bool':4, 'float':8}
```
yguclu, line 82: When is `real` used? I wonder if we could remove this key from all dictionaries in this module.

pyccel/ast/datatypes.py
```diff
@@ -77,23 +78,24 @@
...
+                                 'complex':('complex',default_precision['complex']),
                                  'complex64':('complex',4),
                                  'complex128':('complex',8),
                                  'int8' :('int',1),
                                  'int16':('int',2),
                                  'int32':('int',4),
                                  'int64':('int',8),
-                                 'int'  :('int',8),
+                                 'int'  :('int', default_precision['int']),
                                  'integer':('int',4),
```
yguclu, line 97: When is `integer` used?

pyccel/ast/datatypes.py
saidctb: it is used everywhere a Variable can be declared only by `real` not `float` or `double`

pyccel/ast/datatypes.py
saidctb: it is also used everywhere we probably need to set its precision to default_precision

EmilyBourne: I can't find it used anywhere

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -0,0 +1,1359 @@
...
+        from numpy import floor
+        a = floor(x)*floor(y)
+        return a
+
+    f2 = epyccel(floor_phrase)
+    x = rand()
+    y = rand()
+    assert(isclose(f2(x,y), floor_phrase(x,y), rtol=1e-15, atol=1e-15))
+
+@pytest.mark.xfail
```
yguclu, line 446: I don't recall why this test fails. Could you add an explanation?

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -0,0 +1,1359 @@
...
+        from numpy import fabs
+        a = fabs(x)*fabs(y)
+        return a
+
+    f2 = epyccel(fabs_phrase)
+    x = rand()
+    y = rand()
+    assert(isclose(f2(x,y), fabs_phrase(x,y), rtol=1e-15, atol=1e-15))
+
+@pytest.mark.xfail
```
yguclu, line 59: Could you write the reason why this test fails?

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -0,0 +1,1359 @@
...
+        from numpy import absolute
+        a = absolute(x)*absolute(y)
+        return a
+
+    f2 = epyccel(absolute_phrase)
+    x = rand()
+    y = rand()
+    assert(isclose(f2(x,y), absolute_phrase(x,y), rtol=1e-15, atol=1e-15))
+
+@pytest.mark.xfail
```
yguclu, line 94: Could you write the reason why this test fails?

pyccel/parser/semantic.py
```diff
@@ -102,33 +77,21 @@
 
 #==============================================================================
 
-from sympy.core.function       import Function, FunctionClass, Application
-from sympy.core.numbers        import ImaginaryUnit, IntegerConstant
-from sympy.logic.boolalg       import Boolean, BooleanTrue, BooleanFalse
+from sympy.core.function       import Application, UndefinedFunction
```
yguclu, line 80: Could you explain why moving from `Function` to `UndefinedFunction`? Thanks

pyccel/ast/datatypes.py
saidctb: when you use the function `str_dtype` it's converted to it

pyccel/codegen/printing/fcode.py
```diff
@@ -481,7 +465,8 @@ def _print_Print(self, expr):
 
     def _print_SymbolicPrint(self, expr):
         # for every expression we will generate a print
-        _iprint = lambda e: "print *, 'sympy> {}'".format(e)
+        def _iprint(e):
+            return "print *, 'sympy> {}'".format(e)
```
yguclu, line 469: Is there a reason for removing the lambda function?

pyccel/ast/datatypes.py
saidctb: yes we can

pyccel/codegen/printing/fcode.py
```diff
@@ -481,7 +465,8 @@ def _print_Print(self, expr):
...
         code = ''
         for a in expr.expr:
             code = '{code}{p}'.format(code=code, p=_iprint(a))
```
yguclu, line 472: Concatenating a string within a for loops seems a little wasteful. Maybe write
```
code = "".join(map(_iprint, expr.expr))
```
But aren't we missing a carriage return `'\n'` between statements?

pyccel/ast/datatypes.py
yguclu: So `real` is used for the precision of Pyccel variables? And `float` represents a Python float? What about `double` then, where is it used?

pyccel/ast/datatypes.py
yguclu: I would also prefer to remove `integer` and use `int` everywhere, unless there is a good reason to keep both.

pyccel/ast/datatypes.py
saidctb: everything is converted to `real`

pyccel/parser/semantic.py
saidctb: to be able to subclass it like this 
```
class UndefinedFunction(sp_UndefinedFunction, PyccelAstNode):
    pass
```
we can't do the same with Function

pyccel/codegen/printing/fcode.py
EmilyBourne: I think I removed this as it gave a codacy warning:
`W0108: Lambda may not be necessary (unnecessary-lambda)`
Although it may have meant that we can use this expression without defining a function (it is only used once) rather than that it would be better to use a function rather than a lambda function
```
Name: Unnecessary lambda

Description: A lambda is used that calls through to a function without modifying any parameters
```

pyccel/ast/headers.py
```diff
@@ -417,13 +416,6 @@ def __new__(cls, name, args, master, master_args, results=None):
         if not isinstance(master, (str, FunctionDef, Interface)):
             raise ValueError('Expecting a master name of FunctionDef')
 
-        # we sympify everything since a macro is operating on symbols
-        if not(args is None):
-            args = [sympify(a, locals=local_sympify) for a in args]
-
-        if not(master_args is None):
-            master_args = [sympify(a, locals=local_sympify) for a in master_args]
-
```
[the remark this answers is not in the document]
EmilyBourne: These lines were used. But I don't think they were doing anything. As these objects are function arguments they are either symbols or ValuedArguments (default values), in the case of default values the value needs to be treated by pyccel not by sympy so I moved the treatment to parser/semantic.py
https://github.com/pyccel/pyccel/pull/266/files#diff-47dd0da6ceb30ff82ba7d591e55cccb7R3277-R3278

tests/epyccel/recognised_functions/test_math_funcs.py
```diff
@@ -0,0 +1,402 @@
...
+
+
+@pytest.mark.xfail
+def test_sqrt_return_type():
+    @types('real')
+    def sqrt_return_type_real(x):
+        from math import sqrt
+        a = sqrt(x)
+        return a
+    #TODO we should use cmath instead of math
```
EmilyBourne, line 72: Does this comment refer to both functions or just the complex function? cmath and math are 2 different packages. We need to decide if we want to support the additional package

EmilyBourne: I see that this test fails because math does not handle complex numbers. The test therefore shouldn't exist in this file

tests/epyccel/recognised_functions/test_numpy_funcs.py
EmilyBourne: Sorry this test shouldn't fail. It was failing because numpy returns numpy types rather than python types. It is therefore necessary to use `.item()` in order to get the underlying class

pyccel/ast/headers.py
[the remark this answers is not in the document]
yguclu: I see. Perfect, thanks

pyccel/parser/semantic.py
yguclu: Interesting.. thanks!

pyccel/ast/numpyext.py
```diff
@@ -333,7 +333,7 @@ def __new__(cls, arg, index=None):
                                 IndexedBase)):
             raise TypeError('Uknown type of  %s.' % type(arg))
 
-        # TODO add check on index: must be Integer or Variable with dtype=int
+        # TODO add check on index: must be sp_Integer or Variable with dtype=int
         # TODO [YG, 09.10.2018]: Verify why index should be passed at all (not in Numpy!)
```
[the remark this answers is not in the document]
yguclu: I actually meant the comment below, which dates back to October 2018! :wink:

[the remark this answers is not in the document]
EmilyBourne: I got rid of both as I don't think we need to ensure that the index is an Integer or Variable with dtype=int as this will pose problems fo rthings such as x[i*j] where the type of index would be Mul. If every object has a dtype then it may be possible to test this

pyccel/codegen/printing/fcode.py
```diff
@@ -465,11 +465,7 @@ def _print_Print(self, expr):
 
     def _print_SymbolicPrint(self, expr):
         # for every expression we will generate a print
-        def _iprint(e):
-            return "print *, 'sympy> {}'".format(e)
-        code = ''
-        for a in expr.expr:
-            code = '{code}{p}'.format(code=code, p=_iprint(a))
```
yguclu, line 487: So is this loop not needed at all? Why is that?

EmilyBourne: Because I mis-typed my correction :/

pyccel/codegen/printing/fcode.py
EmilyBourne: I think we probably are. But I don't know exactly what a symbolic print is. @saidctb ?

pyccel/codegen/printing/fcode.py
```diff
@@ -465,7 +465,7 @@ def _print_Print(self, expr):
 
     def _print_SymbolicPrint(self, expr):
         # for every expression we will generate a print
-        code = ''.join(["print *, 'sympy> {}'".format(a)])
```
yguclu: I think that passing a generator expression to `str.join()` should be more efficient than passing a list.
You may simply remove the square brackets!

pyccel/ast/basic.py
yguclu: For now we keep things as they are. We refer to issue #305 for improvements on our type inference mechanism.

pyccel/ast/datatypes.py
yguclu: Clean up of the dictionaries in `pyccel.ast.datatypes` should be done when we work on issue #305.

PR #268, opened 2020-03-17
Testing sympy>=1.5

### LINKED ISSUE #267: Errors with sympy>=1.5.0
Several unit tests fail with sympy>=1.5.0 . This is why pyccel now locks sympy at <1.5 in setup.py . According to @yguclu this is due to an upstream bug in sympy. There is now a branch `devel-sympy15` to investigate and/or find workarounds.

krystophny:
Testing sympy>=1.5 . Not ready for merging until it fixes #267 .


krystophny:
`string_types` and `range` have been dropped from `sympy.core.compatibility` in upstream git and were only needed for Python2 compatibility. Replaced them by usual `str` and `range`.

PR #271, opened 2020-03-17
Devel tuples

### LINKED ISSUE #265: Clearly define tuples
- [x] Create tuple class with NotImplementedErrors where appropriate
       - dtype is tuple
- [x] Allow allocation with the following syntax: `a, b, c = 1, False, 3.4`
- [x] Allow allocation of inhomogeneous tuple `a = (1, False, 3.4)`
- [x] Pass tuple to function using *args syntax
- [x] Access tuple element `a[0]`
- [x] Raise error when accessing tuples using indices `a[i]`

EmilyBourne:
Close when fixes #265


EmilyBourne:
3 classes have been created:
- NativeTuple which represents the tuple type
- PythonTuple which represents a tuple in python
- TupleVariable which represents a variable which is a tuple. This is necessary as a separate entity to Variable as it does not have one distinct type

EmilyBourne:
PythonTuple and TupleVariable interfaces should be harmonised

EmilyBourne:
Remaining failing test in Codegen is due to the following line :
```
s = [(x1, y1, z1)  for x1 in range(1,30) for y1 in range(x1,30) for z1 in range(y1,1000)]
```

We should discuss whether we want this line to be permitted. It is probably useful for examples such as:
```
s = np.array([(x1, y1, z1)  for x1 in range(1,30) for y1 in range(x1,30) for z1 in range(y1,1000)])
```
but it may not make much sense on it's own

yguclu:
Are there any remaining failing tests in codegen as of now?

EmilyBourne:
There are no more failing tests

yguclu:
> There are no more failing tests

So how do you deal with this line now?
```
s = [(x1, y1, z1)  for x1 in range(1,30) for y1 in range(x1,30) for z1 in range(y1,1000)]
```

EmilyBourne:
> > There are no more failing tests
> 
> So how do you deal with this line now?
> 
> ```
> s = [(x1, y1, z1)  for x1 in range(1,30) for y1 in range(x1,30) for z1 in range(y1,1000)]
> ```

The tuple is recognised as being homogeneous and is dealt with in the same way as a list. So in this case a 2d array is created

yguclu:
Codacy complaint

> arguments-differ (W0221)
> Parameters differ from %s %r method Used when a method has a different number of arguments than in the implemented interface or in an overridden method.

should not apply to class constructors. I will ignore it.

pyccel/ast/core.py
```diff
@@ -2113,7 +2116,7 @@ def __new__(
         class_type = cls_base \
             or dtype.__class__.__name__.startswith('Pyccel')
         alloweddtypes = (NativeBool, NativeRange, NativeString,
```
yguclu: I have just realized that `NativeBool` is checked for twice in the `if/elif` block below, so maybe we should remove it from the tuple above?

As a side note, I think the `if/elif` block below should be improved: it is more natural to raise a TypeError after an `else` statement, making it clear that we are covering all possible cases (by looking at the block as it is, one is left with some doubts about it).

pyccel/ast/core.py
```diff
@@ -2439,6 +2442,49 @@ def _sympystr(self, printer):
         value = sstr(self.value)
         return '{0}={1}'.format(name, value)
 
+class TupleVariable(Variable):
+
+    """Represents a valued variable in the code.
```
yguclu: Since you are giving a proper docstring to this new class, I think it is worth following [Numpy's dosctring guidelines](https://numpydoc.readthedocs.io/en/latest/format.html#sections), so that one day we will be able to create automatic documentation for Pyccel. In particular you should add the header
```
Parameters
----------
```
and add a dashed line below "Examples".

pyccel/ast/core.py
```diff
@@ -3710,7 +3756,7 @@ def __new__(
         ):
         if isinstance(dtype, str):
             dtype = datatype(dtype)
-        elif not isinstance(dtype, DataType):
+        elif not isinstance(dtype, (DataType, NativeTuple)):
```
yguclu: See earlier comment

pyccel/ast/core.py
```diff
@@ -4091,7 +4137,7 @@ def __new__(
             raise TypeError('datatype must be provided')
         if isinstance(dtype, str):
             dtype = datatype(dtype)
-        elif not isinstance(dtype, DataType):
+        elif not isinstance(dtype, (DataType, NativeTuple)):
```
yguclu: See earlier comment

pyccel/ast/core.py
```diff
@@ -4491,6 +4537,14 @@ class IfTernaryOperator(If):
 
     pass
 
+class StarredArguments(Basic):
+    def __new__(cls, args):
+        return Basic.__new__(cls, args)
+
+    @property
+    def args(self):
```
yguclu: See earlier comment about overriding `Basic.args`

pyccel/ast/datatypes.py
```diff
@@ -138,6 +139,45 @@ class NativeList(DataType):
     _name = 'List'
     pass
 
+class NativeTuple(Basic):
```
yguclu: See earlier comment. It seems strange that `NativeList` subclasses `DataType` while `NativeTuple` does not.

tests/codegen/fcode/scripts/ListComprehension.py
```diff
@@ -5,3 +5,5 @@
 z = [i*j*k1 for i in range(200) for j in range(0,i,2) for k1 in y]
 
 s = [(x1, y1, z1)  for x1 in range(1,30) for y1 in range(x1,30) for z1 in range(y1,1000)]
+
+t = [(a1, False, c1)  for a1 in range(1,30) for b1 in range(a1,30) for c1 in range(b1,1000)]
```
yguclu: Will this become an homogeneous list of tuples, all with same `dtype=(int, bool, int)`? Or should it fail?

Could you add a comment line above each of these expressions, to explain what behavior is expected?

tests/pyccel/scripts/test_tuples.py
```diff
@@ -0,0 +1,42 @@
...
+print(e)
+print(f)
+print(g)
+print(h)
+print(i)
+print(ai_0)
+print(x[idx])
+
+idx_2 = (0,1,2)
+print(x[idx,idx_2,1])
```
yguclu: Here you are indexing a multi-dimensional array using an iterable of integers (or a scalar integer) along each direction (https://numpy.org/devdocs/user/basics.indexing.html#indexing-multi-dimensional-arrays). That's very impressive! If this is fully implemented and it works, then maybe you could create a dedicated issue and link this pull request to it?

tests/semantic/scripts/expressions.py
```diff
@@ -28,3 +28,5 @@
 
 # this statement will be ignored at the codegen
 x = ad is None
```
yguclu: Is this comment still correct? Shouldn't `x` just be a boolean variable, and more specifically `x := False`?

pyccel/ast/core.py
```diff
@@ -2022,7 +2025,7 @@ def __new__(
         if isinstance(dtype, str) or str(dtype) == '*':
 
             dtype = datatype(str(dtype))
-        elif not isinstance(dtype, DataType):
+        elif not isinstance(dtype, (DataType, NativeTuple)):
             raise TypeError('datatype must be an instance of DataType.')
```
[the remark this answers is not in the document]
EmilyBourne: This change is something I pushed a few hours ago. I was originally subclassing DataType, but the problem is that DataType subclasses Singleton. Singleton ensures that only one instance of this class exists. This makes sense for objects such as NativeInt which have a fixed type however it is useful for the tuple to include information about it's contents. In order for this to work each different tuple needs its own NativeTuple which then contains NativeInts etc.

It may be possible to modify the usage so that the subclass types are determined from the variable. I will try to modify this

tests/semantic/scripts/expressions.py
EmilyBourne: I don't think this statement is entirely correct anymore. It might be good to implement this behaviour.
Currently the presence of `ad is None` is used to mark the variable as optional. If the user writes:
`if (ad is None):`
Then this is interpreted as `If (.not. present(ad))`
However the assignment case hasn't been handled and so a warning still appears and the line still isn't printed.
This is a quick fix, however a program is not the correct place to test whether it works.
I don't know whether it should have different behaviour in a program or if we should just leave it as is as there is no reason to write this in a program unless you are using None as a variable, which cannot be translated anyway

pyccel/ast/core.py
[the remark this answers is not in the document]
yguclu: I see the problem now. Theoretically speaking, each call to `tuple()` defines a new DataType, which depends on the type of the arguments. Sometimes we only care if an object is a NativeTuple, and sometimes we really want to know its specific type.

pyccel/ast/core.py
EmilyBourne: How do you suggest clarifying that? With an `elif:pass` ?

pyccel/ast/core.py
EmilyBourne: This also seems to be true in the class `Concatenate`.
Does anyone have a suggestion for a better property name there? @saidctb ?

tests/semantic/scripts/expressions.py
yguclu: I see. This should work inside of a function with optional arguments, but it makes no sense in a program. Thanks for looking into that.

pyccel/ast/core.py
yguclu: Yes, I would write a `pass` and maybe a line of comment above it.

pyccel/ast/core.py
yguclu: Assuming that `Concatenate` is used for String concatenation as the docstring suggests, I am not sure if it is ever used in Pyccel. I can find a single place in Pyccel where an object of this class is created, it's in `SemanticParser._visit_Add()`: https://github.com/pyccel/pyccel/blob/54125929d992a54b2d7739b7c4a3827381977de0/pyccel/parser/semantic.py#L1519

(@saidctb: Can you tell us when and how `Concatenate` is used?)

yguclu:
It appears that we have a problem with Travis CI, again...

yguclu:
@EmilyBourne Could you please push some dummy commit to see if we can trigger Travis CI once again? Maybe update the version...

EmilyBourne:
Travis is now passing:
https://travis-ci.org/github/pyccel/pyccel/builds/666494342?utm_medium=notification&utm_source=github_status
It seems that it is not reporting its results due to a problem with the codacy curl command

yguclu:
I am afraid I will need to setup our Travis CI integration from scratch. This may take a while.

PR #274, opened 2020-03-19
Devel pycode

saidctb:


EmilyBourne:
Can we test the results of the codegen step rather than just testing if the codegen step doesn't fail?

EmilyBourne:
We should decide what the best method is for the tests. There are several files which resemble tests/codegen/pycode/test_codegen.py . I don't find them particularly helpful for debugging as all the tests are in the same pytest. A comment at the start of these files states that they are this way as we need to change directory for imports:
```
# Note that we need to change the directory for tests involving the import
# statement
```
however this is no longer the case. I suggest that we reorganise these files as I have done with test_semantic so that we have more granular information about pass/fails in travis.
What do you think @yguclu  @saidctb ?

saidctb:
@EmilyBourne here we only test the code generation process and not the correctness of the generated code

EmilyBourne:
@saidctb I agree but surely that should be part of a unit test?

saidctb:
@EmilyBourne  yes i agree

yguclu:
I agree, we should reorganize these tests following Emily's example with test_semantic.

I would like to stress that we have many features in pyccel but still too few tests, so from my point of view adding a well designed unit tests is more important than adding features at this stage. All this work now (which should have been done when the code was written) will be very useful in the future.

yguclu:
@saidctb Thanks for the effort you put into the unit tests!

PR #278, opened 2020-03-20
Allow absolute imports in complex directory structure

### LINKED ISSUE #277: Allow absolute import in complex directory structure
If we have a project with a directory structure like the following,
```
├── Makefile
└── proj
       ├── __init__.py
       ├── dir_a
       │   ├── __init__.py
       │   ├── mod_a1.py
       │   └── mod_a2.py
       └── dir_b
            ├── __init__.py
            ├── mod_b1.py
            └── mod_b2.py
```
and `proj.dir_b.mod_b1` depends on some function `f` in `proj.dir_a.mod_a2`, the former module can import from the latter by using one of two alternatives:

1. `from proj.dir_a.mod_a2 import f` (**absolute** import)
2. `from ..dir_a.mod_a2 import f` (**relative** import)

Currently Pyccel only supports relative imports (2), but we would like to also support absolute ones (1). To do so I can suggest two options:

1. Assume that Pyccel's current directory contains the project root directory (as in the case where pyccel is called from the Makefile above);
2. Search for the module in `sys.path`, hence following Python's own import logic.

yguclu:
Fixes #277


yguclu:
@EmilyBourne, @saidctb:
1. Since we are not using `.pyh`header files extensively yet, I am going to create a specific issue about them and leave it to the future;
2. For large projects that use a Makefile in each folder, absolute imports will work if one of the following conditions apply:
   1. The project was installed in a directory in `sys.path`;
   2. The root directory of the project was added to the `PYTHONPATH` environmental variable.

Therefore I think this branch can be considered ready for merge. We might want to further clean up the `tests/pyccel` folder, though. Any help or suggestion is appreciated.

PR #286, opened 2020-03-26
Devel tuples

EmilyBourne:


EmilyBourne:
See tests/pyccel/scripts/runtest_tuples.py for working syntax

pyccel/ast/builtins.py
```diff
@@ -262,13 +262,25 @@ def __new__(cls, args):
...
+            shape = [len(self._args)]
+            if self.is_homogeneous and isinstance(self._arg_dtypes[0]['datatype'], NativeTuple):
+                shape = shape + list(self._args[0].shape)
+
+            return tuple(shape)
 
     @property
     def rank(self):
-        return 1 if self.is_homogeneous else 0
+        return max(a.rank if hasattr(a, 'rank') else 0 for a in self._args)+1
```
yguclu: Here `a.rank if hasattr(a, 'rank') else 0` may be written as `getattr(a, 'rank', 0)`, which is more concise.

pyccel/ast/builtins.py
```diff
@@ -288,7 +300,26 @@ def is_homogeneous(self):
     def set_arg_types(self,d_vars):
```
yguclu: Can you explain what `d_vars` is here? Maybe with a docstring, or at least with a comment. This will make it much easier to maintain the code later.

pyccel/ast/core.py
```diff
@@ -2519,11 +2519,38 @@ def __new__(cls, arg_vars, *args, **kwargs):
```
yguclu: A comment about initialization.

Why do we need to pass `NativeTuple()` as the second argument to the constructor of `TupleVariable`? Are we allowed to pass datatypes other than `NativeTuple()`? Or maybe in the future we might want to?

If not, it would seem more natural to me not to pass it at all.

pyccel/ast/f2py.py
```diff
@@ -127,7 +127,7 @@ def as_static_function(func, name=None):
             transpose_stmts = [Comment(f2py_template.format(kind    = int_kind,
                                                             index   = i,
                                                             name    = a.name,
-                                                            index_t = str(a.rank - i - 1)))
+                                                            index_t = str(i)))
```
yguclu: Why are we not reversing the shape anymore?

pyccel/codegen/printing/fcode.py
```diff
@@ -535,7 +538,7 @@ def _print_AnnotatedComment(self, expr):
 
     def _print_Tuple(self, expr):
         import numpy
-        shape = numpy.shape(expr)
+        shape = list(reversed(numpy.shape(expr)))
```
yguclu: So you are reversing the shape here?

EmilyBourne: I am reversing the shape so it matches the fortran indexing

pyccel/ast/core.py
yguclu: Thanks. You may need to update the example in the docstring now.

pyccel/codegen/printing/fcode.py
```diff
@@ -938,8 +952,12 @@ def _print_Declare(self, expr):
...
             #TODO fix bug when we inclue shape of type list
 
-            rankstr =  ','.join('{0}:{1}-1'.format(self._print(s),
-                                                 self._print(i)) for i in shape)
+            if var.order=='C':
+                rankstr =  ','.join('{0}:{1}-1'.format(self._print(s),
+                                                    self._print(i)) for i in shape[::-1])
+            else:
+                rankstr =  ','.join('{0}:{1}-1'.format(self._print(s),
+                                                     self._print(i)) for i in shape)
```
EmilyBourne: > Why are we not reversing the shape anymore?

@yguclu  This is due to these lines. The shape was printed in the same way for C-ordered and F-ordered arrays. I'm not sure how we got away with this before. I think it was due to the f2py line and the fact that all the initialisation was reversed anyway for arrays. Once this has been corrected in order to correctly print the tuples, the f2py sizes no longer match . However this is corrected by the comment.

I think it would be worth looking into this a bit more to check that I haven't broken anything. For the moment I am relying on the fact that the tests pass to ensure that the logic is correct, but I don't think I have fully understood why it was working before. @saidctb do you understand this?

pyccel/ast/f2py.py
EmilyBourne: See:
https://github.com/pyccel/pyccel/pull/286/files#r399179413

pyccel/codegen/printing/fcode.py
yguclu: Actually I played around with various solutions before finding the one that was implemented in the f2py printer before. I also discussed with Said a lot about it.

@EmilyBourne You might have found a different way to achieve the same goal. Interesting..

yguclu: I wonder if we even need to have an f2py comment for C ordering now..

pyccel/ast/f2py.py
yguclu: At this point it looks like we do not need to transpose the indices for f2py (`transpose_stmts` variable), and maybe not even to declare the array as having C ordering (`c_intent_stmt` variable). Can you please try and remove each of them?

EmilyBourne: If I remove both then the first test to fail is: `test_array_int32_2d_C_scalar_add`
If I remove just `transpose_stmts` then this passes but a different test fails. In this case the first to fail is : `test_array_int_2d_C_initialization`
If I remove just `c_intent_stmt` then the failing test is once again: `test_array_int32_2d_C_scalar_add`

EmilyBourne: The error for `test_array_int32_2d_C_scalar_add `is : `ValueError: failed to initialize intent(inout) array -- input not fortran contiguous`
For `test_array_int_2d_C_initialization` the initialised values are wrong:
```
E       assert False
E        +  where False = <function array_equal at 0x7f78370b72f0>(array([[1, 2, 3],\n       [4, 5, 6]]), array([[  1,   2,   4],\n       [  5,  64, 721]]))
E        +    where <function array_equal at 0x7f78370b72f0> = np.array_equal
```

pyccel/codegen/printing/fcode.py
saidctb: yes maybe we don't need the comments now

pyccel/ast/f2py.py
yguclu: So we really need those statements, and the tests are strict enough to identify any problem. Thanks for trying!

pyccel/codegen/printing/fcode.py
yguclu: Apparently we really need them, otherwise some tests fail.

Issue #194, opened 2019-10-30
Test Windows operating system support


krystophny:
After discussing with @yguclu we are planning to test Windows support with Anaconda Python and automated testing with Travis CI. All tests that don't use MPI seem to run by this strategy.

In the following steps, if using your own PC you have to restart git bash instead of running `refreshenv` as in Travis CI.

Steps (Travis uses choco package manager from https://chocolatey.org/ ):

* Install anaconda, git and mingw compilers with
```
choco install anaconda3
choco install git
choco install mingw
refreshenv
source /c/tools/Anaconda3/etc/profile.d/conda.sh
conda activate
```
(Command `conda install -c conda-forge fortran-compiler` would install flang instead of gfortran on windows.)

* Change default C compiler from M$ to mingw in `distutils.cfg` by running
```
echo -e "[build]\ncompiler = mingw32" >> /c/tools/Anaconda3/Lib/distutils/distutils.cfg
```

* Download x64 libblas.dll and liblapack.dll from https://icl.cs.utk.edu/lapack-for-windows/lapack/ to library directory
```
curl https://icl.cs.utk.edu/lapack-for-windows/libraries/VisualStudio/3.7.0/Dynamic-MINGW/Win64/libblas.dll 
-o /c/ProgramData/chocolatey/lib/mingw/tools/install/mingw64/lib/libblas.dll
curl https://icl.cs.utk.edu/lapack-for-windows/libraries/VisualStudio/3.7.0/Dynamic-MINGW/Win64/liblapack.dll 
-o /c/ProgramData/chocolatey/lib/mingw/tools/install/mingw64/lib/liblapack.dll
```

* Clone and install pyccel
```
git clone git@github.com:pyccel/pyccel.git
cd pyccel
pip install .
```

* Test with `tests\run_tests.bat`. MPI and INT64 stuff is currently deactivated.

For MPI support

* Download MS MPI runtime and SDK
```
curl -L https://github.com/microsoft/Microsoft-MPI/releases/download/v10.1.1/msmpisetup.exe -o  msmpisetup.exe
curl -L https://github.com/microsoft/Microsoft-MPI/releases/download/v10.1.1/msmpisdk.msi -o msmpisdk.msi
```

* Install MS MPI (run git bash as Administrator)
```
./msmpisetup.exe -unattend
sleep 5
msiexec //quiet //i msmpisdk.msi
refreshenv
```

[...]

PR #292, opened 2020-04-01
MPI working on Windows

krystophny:
Following instructions in #194 and https://github.com/dev-cafe/autocmake/issues/85 now the MPI tests run on Windows with these modifications. If this doesn't break other stuff, please merge.


pyccel/codegen/utilities.py
```diff
@@ -110,7 +115,14 @@ def compile_fortran(filename, compiler, flags,
     m_code = ' '.join('{}.o'.format(m) for m in modules)
     libs_flags = ' '.join('-l{}'.format(i) for i in libs)
 
-    cmd = '{0} {1} {2} "{3}" {4} "{5}" {6} {7} {8}'.format( \
+    filename = '"{}"'.format(filename)  # in case of spaces in path
+    binary = '"{}"'.format(binary)
+
+    if sys.platform == "win32" and compiler == "mpif90":
+        compiler = "gfortran"
```
EmilyBourne, line 122: Why do we change the selected compiler? If this is a default then shouldn't the switch be done when the compiler is selected? If not then shouldn't we use the compiler that the user asked for?

pyccel/codegen/utilities.py
```diff
@@ -59,6 +60,10 @@ def construct_flags(compiler,
     if compiler == "mpif90":
         if debug:
             flags += " -fbounds-check"
+        if sys.platform == "win32":
+            mpiinc = os.environ["MSMPI_INC"].rstrip('\\')
+            mpilib = os.environ["MSMPI_LIB64"].rstrip('\\')
```
EmilyBourne, line 65: Are these environment variables specific to a certain compiler?

pyccel/codegen/utilities.py
krystophny: The problem here is that `mpif90` doesn't exist on Windows. One rather writes `gfortran` with some command line argument telling it to link MPI libraries. This solution here is not very clean, as it doesn't distinguish whether the user explicitly wants `mpif90`, and whether MPI default compiler was selected. This should indeed be changed.

pyccel/codegen/utilities.py
krystophny: This is for MPI on Windows for any compiler, since Microsoft MPI is needed there.

EmilyBourne: From some googling it seems that there is also an Intel MPI for windows. However if the user wishes to use this I think we can suppose that they will specify their own flags. In this case would we end up with multiple MPI routines specified?

If this isn't compiler specific then why is it in the `if compiler == "mpif90":`? Is this once again related to the default compiler? If the user specified their own compiler then won't the MPI flags be missed?

yguclu: Depending on which MPI library and compiler the user wants to use, I think that `mpif90` should point to the correct compiler. Now the question is: how do we know if the user wants to use the Intel MPI and compiler?

yguclu: Sorry, I just read in the comment below that the `mpif90` wrapper does not exist on Windows. I find that rather strange and inconvenient.,,

EmilyBourne: For pyccel these are specified as options in the help. For epyccel they are in the kwargs

PR #303, opened 2020-04-21
detect inout arguments when lhs is a Variable

saidctb:


tests/epyccel/test_epyccel_functions.py
```diff
@@ -154,6 +154,35 @@ def f8(x,b):
...
+        x += 1
+
+    f = epyccel(f9)
+
+    x = np.zeros(10, dtype='int64')
+    x_expected = x.copy()
+
+    f9(x)
+    f(x_expected)
+    assert np.allclose(x, x_expected, rtol=1e-15, atol=1e-15)
```
yguclu, line 170: Shouldn't we check for exact equality here? This is an integer array...

tests/epyccel/test_epyccel_functions.py
```diff
@@ -154,6 +154,35 @@ def f8(x,b):
...
+        x[:] += 1
+
+    f = epyccel(f10)
+
+    x = np.zeros(10, dtype='int64')
+    x_expected = x.copy()
+
+    f10(x)
+    f(x_expected)
+    assert np.allclose(x, x_expected, rtol=1e-15, atol=1e-15)
```
yguclu, line 184: Same comment as above.

tests/epyccel/test_epyccel_functions.py
saidctb: done

tests/epyccel/test_epyccel_functions.py
saidctb: done

PR #306, opened 2020-05-05
Add arguments to rand

### LINKED ISSUE #223: Numpy rand not working
rand from numpy.random cannot be pyccelized. This is marked as to do in tests/codegen/scripts/numpyext.py

EmilyBourne:
Fixes #223

### DIFF
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -165,11 +165,7 @@ class Array(Application):
         # row-major ordering, while Fortran initial values are column-major
         shape = self.shape[::-1]
 
-        if isinstance(shape, (list, PythonTuple, Tuple, tuple)):
-            # this is a correction. problem on LRZ
-            shape_code = ', '.join('0:' + printer(Add(i, Integer(-1))) for i in shape)
-        else:
-            shape_code = '0:' + printer(Add(shape, Integer(-1)))
+        shape_code = ', '.join('0:' + printer(Add(i, Integer(-1))) for i in shape)
 
         lhs_code = printer(lhs)
         code_alloc = 'allocate({0}({1}))'.format(lhs_code, shape_code)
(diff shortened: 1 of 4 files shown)


pyccel/ast/numpyext.py
```diff
@@ -881,30 +882,59 @@ def fprint(self, printer, lhs):
...
-    def __new__(cls, arg = Nil()):
-        return Real.__new__(cls, arg)
+    _dtype = 'real'
+    _precision = default_precision['real']
 
-    def __init__(self):
-        Real.__init__(self, None)
+    def __init__(self, *args):
+        if len(args) == 0:
+            self._shape = ()
```
yguclu, line 896: Isn't `args` an empty tuple when its length is zero?

EmilyBourne: You are right!

pyccel/ast/numpyext.py
```diff
@@ -881,30 +882,59 @@ def fprint(self, printer, lhs):
...
+        stmts = []
+
+        if self.rank>0:
+            # Create statement for allocation
+            if not stack_array:
+                # Transpose indices because of Fortran column-major ordering
+                shape = self.shape[::-1]
+
+                if isinstance(self.shape, (PythonTuple,Tuple,tuple)):
+                    # this is a correction. problem on LRZ
```
yguclu, line 920: What problem do we have on LRZ?

EmilyBourne: I don't know. The allocation code was copied and pasted from the Full class (this should probably be done once. Maybe in the allocation super class that we could create for issue #307)
If I work my way back up the tree it seems that @saidctb wrote this comment originally

yguclu: @saidctb Do you remember what this problem was about? Can we just remove those lines?

saidctb: @ratnania added this part so I'm not sure

yguclu: After taking another look at it, I think that we never enter the code at line 927 just below. In fact `shape` should never be a scalar, and the comment `# this is a correction. problem on LRZ` is probably a leftover. So I suggest we replace the whole `if/else` block with just
```
shape_code = ', '.join('0:' + printer(Add(i, Integer(-1))) for i in shape)
```
The same change should be made to the `Full` class, and maybe to other places too.

EmilyBourne: I think you are right. I tried to ensure that shape is always a tuple, so it shouldn't be used. And codacy's coverage seems to indicate that it isn't used

PR #309, opened 2020-05-06
Ccode

saidctb:
Set up code generation process for generating C code


tests/codegen/ccode/scripts/loops.py
```diff
@@ -0,0 +1,119 @@
...
+#@types( 'real[:], real[:]' )
+#def product_loop_on_real_array( z, out ):
+#
+#    from numpy     import shape
+#
+#    s = shape( z )
+#    n = s[0]
+#
+#    for i in range(n):
+#        out[i] = z[i]**2
```
EmilyBourne, line 119: Should we put these tests in a separate files and mark them as xfail rather than commenting them?

saidctb: arrays are not implemented at all in C that's why i prefered to comment it rather than mark them xfails

> We should add end to end tests that check the results not just tests to see if something is generated
yes this tests just tests if the codegen mechanism works
we should definitely add tests to check the results of the functions
but this is for later, the goal of this branch is to add the ccode mechanism and be sure that it works not that it's correct.
in order for us to tests the results we need to be able to compile a c program and also be able to generate a  shared library that can be imported in Python

EmilyBourne: In order to test the results with epyccel we need to be able to compile a c program and also be able to generate a shared library that can be imported in Python.
If we just use pyccel we can add the compile command manually for basic testing. 

The test "test_imports_compile" in tests/pyccel/test_pyccel.py does this for fortran

EmilyBourne: An array in C does exist. It is constructed as a pointer:
```
int a[4];
```
would be equivalent to
```
a = np.empty(4,dtype=int)
```

saidctb: yes we need to add compilation of a c program.

saidctb: for the arrays it's not clear to me how we can implement it, for example in Fortran we can add tow arrays but we can't in C
`c = a + b ` where `a` and `b` are arrays

EmilyBourne: Obviously these kinds of operations would have to be done element-wise. We can create a separate issue for this as it seems to be beyond the scope of this pull request

saidctb: yes it should be solved in another pull request

tests/codegen/ccode/scripts/arrays.py
```diff
@@ -0,0 +1,40 @@
...
+#==============================================================================
+
+@types( 'int[:,:](order=C)' )
+def double_loop_on_2d_array_C( z ):
+
+    from numpy import shape
+
+    s = shape( z )
+    m = s[0]
+    n = s[1]
```
yguclu, line 12: We can now write `m, n = shape(z)`, so I suggest we use that syntax in our tests

tests/codegen/ccode/scripts/arrays.py
```diff
@@ -0,0 +1,40 @@
...
+            z[i,j] = i-j
+# ...
+@types( 'int[:,:](order=F)' )
+def double_loop_on_2d_array_F( z ):
+
+    from numpy import shape
+
+    s = shape( z )
+    m = s[0]
+    n = s[1]
```
yguclu, line 25: Same comment as above

tests/codegen/ccode/scripts/arrays.py
```diff
@@ -0,0 +1,40 @@
...
+        for j in range( n ):
+            z[i,j] = i-j
+# ...
+@types( 'real[:], real[:]' )
+def product_loop_on_real_array( z, out ):
+
+    from numpy     import shape
+
+    s = shape( z )
+    n = s[0]
```
yguclu, line 37: Here we can simply write `n, = shape(z)`

PR #310, opened 2020-05-06
Clean up the folders

EmilyBourne:
- src_old removed


EmilyBourne:
Leave author names and copyright for another pull request. It is important to remove the `src_old` folder as it is the source of the majority of our security issues

PR #313, opened 2020-05-13
Add compile step for C code to allow end-to-end testing

EmilyBourne:


EmilyBourne:
Print is still required for end to end printing but as this is complicated in c this should probably be a separate pull request.

A wrapper file is still required for epyccel to work. This is also complicated enough to warrant its own pull request

pyccel/codegen/f2py.py
```diff
@@ -42,7 +42,7 @@ def compile_f2py( filename, *,
     if compiler == 'gfortran' or compiler == 'gcc':
         _vendor = 'gnu95'
 
-    if compiler == 'gcc':
```
yguclu: Isn't this condition covered by line 42 above?

EmilyBourne: Oops, I must have forgotten to delete the second condition. I haven't managed to get f2py to work with gcc anyway though :/

EmilyBourne:
> I have just a general question: do we really want to use f2py when we generate C code?

No I don't think so. I have an open pull request which creates a wrapper for the c code. This works in linux although I still have a few problems with the compiler commands on windows and mac. This seemed easier than f2py as f2py has limited support for C compilers

PR #314, opened 2020-05-13
Devel cwrapper

EmilyBourne:
Add basic wrapper for C code to allow end-to-end tests with epyccel


EmilyBourne:
Windows builds are currently not working as the library -lmsvcr140 cannot be found. The command is the following:
```
C:\ProgramData\chocolatey\bin\gcc.exe -shared -s build\temp.win-amd64-3.7\Release\mod_flqa7193_wrapper.o c:\users\travis\build\pyccel\pyccel\__epyccel__\__pyccel__\mod_flqa7193.o build\temp.win-amd64-3.7\Release\mod_flqa7193.cp37-win_amd64.def -LC:\tools\Anaconda3\libs -LC:\tools\Anaconda3\PCbuild\amd64 -lpython37 -lmsvcr140 -o build\lib.win-amd64-3.7\mod_flqa7193.cp37-win_amd64.pyd
```

Presumably there is a missing include directory or the "-lmsvcr140" needs to be removed

EmilyBourne:
I cannot find why mac fails. It creates the .so file but then python's importlib cannot find it. This usually happens for one of the tests. Which test is not consistent. I cannot debug this just with travis output. Can someone with a mac have a look? @yguclu

EmilyBourne:
The windows problem seems to be related to this error:
https://github.com/Roche/pyreadstat/blob/master/windows_compilation.md#the-problems-and-their-solutions

yguclu:
> I cannot find why mac fails. It creates the .so file but then python's importlib cannot find it. This usually happens for one of the tests. Which test is not consistent. I cannot debug this just with travis output. Can someone with a mac have a look? @yguclu

Yes, I will look into this and post any news here.

pyccel/codegen/f2py.py
```diff
@@ -154,42 +157,68 @@ def create_shared_library(codegen,
...
 
-    if language == 'fortran':
+    # Name of shared library
+    if sharedlib_modname is None:
+        sharedlib_modname = module_name
+
+    sharedlib_folder = ''
+
+    if language == 'c':
+        wrapper_code = create_c_wrapper(sharedlib_modname, codegen)
```
yguclu, line 167: If we are not calling `f2py` when we generate C code, maybe we should change the name of this module?

Issue #305, opened 2020-05-04
Improve the type inference mechanism


saidctb:
The Sympy type inference mechanism has a lot of issues and we need to move out from it.
we now use Sympy to infer the type of any expression  by using the attributes `is_integer`, `is_real`, `is_complex` ... that are provided by Sympy,.
as a first step the `_infere_type` method of `pyccel.parser.semantics` should handle this by adding all the possible arithmetic expression type inference rules.
after that we may need to use the `PyccelAstNode` attributes where we infer the type automatically at the instance creation, which means that the type inference mechanism will be local to each class 
rather than it being handled by one method that we call `_infere_type`

EmilyBourne:
The `PyccelAstNode` attributes should be used and initialised in the `__init__` function wherever possible to avoid recalculating values

PR #316, opened 2020-05-13
Fix type inference in Pow

### LINKED ISSUE #315: Broken type inference in Pow
Pyccel fails to translate this code,
```
from numpy import pi
def f():
    value = (2*pi)**(3/2)
    return value
```
and raises the error
```
RecursionError: maximum recursion depth exceeded
```
The same code is correctly processed if we replace `(2*pi)` with `pi`.

saidctb:
Fixes #315
The Pow object is unable to infer its type so we added the proprieties `is_integer`, `is_real`, `is_complex`  that calculates the right type


pyccel/ast/core.py
```diff
@@ -197,6 +197,18 @@ def _eval_subs(self, old, new):
     def _eval_evalf(self,prec):
         return sp_Pow(self.base,self.exp).evalf(prec)
 
+    @property
+    def is_real(self):
+        return self._args[0].is_real and self._args[1].is_real
```
EmilyBourne, line 202: Should this not be or?
2^0.5 is real
1.5^2 is also real

Or does fortran prevent mixing types like that?

pyccel/ast/core.py
```diff
@@ -197,6 +197,18 @@ def _eval_subs(self, old, new):
...
+
+    @property
+    def is_integer(self):
+        return self._args[0].is_integer and self._args[1].is_integer
+
+    @property
+    def is_complex(self):
+        return self._args[0].is_complex and self._args[1].is_complex
```
EmilyBourne, line 210: See above comment

tests/epyccel/test_epyccel_functions.py
```diff
@@ -12,6 +12,15 @@ def clean_test():
     shutil.rmtree('__pycache__', ignore_errors=True)
     shutil.rmtree('__epyccel__', ignore_errors=True)
 
+
+def test_func_no_args_f1():
+    def f1():
+        from numpy import pi
+        value = (2*pi)**(3/2)
```
EmilyBourne, line 19: Do we know why this failed? It seems like the sympy inference mechanism ought to have been sufficient for a constant expression

pyccel/ast/core.py
saidctb: yes you are right

saidctb: if it's an integer => it's a real

pyccel/ast/core.py
saidctb: the same thing if it's an integer or real => it's a complex

tests/epyccel/test_epyccel_functions.py
saidctb: we don't evaluate the constant expression , so sympy treats it as a normal expression

pyccel/ast/core.py
EmilyBourne: Ah ok sorry

tests/epyccel/test_epyccel_functions.py
EmilyBourne: Yes I realise that, but what I mean is that sympy ought to have been capable of treating it. There shouldn't be any pyccel objects that behave differently than the equivalent sympy object would have done should there?

The last time we had a problem like this it was due to mishandling of Variables, and sympy being unable to tell if something was different from zero. Here we can't have anything like that as everything is constant

yguclu: I understand that Pow was not able to determine whether its result was `integer`, `real` or `complex`, and that is now fixed. But can anybody explain why we were getting a `RecursionError`?

saidctb: yes I'm not sure what was the problem i just added the types by myself in those methods so that sympy doesn't do it by himself

EmilyBourne: The `RecursionError` is due to sympy bouncing between 2 functions when analysing the type. Sympy's type inference mechanism is much more sophisticated than we need and checks things such as whether the exponent is negative or 0. This kind of information is not provided by pyccel.

Sympy's type inference mechanism seems quite sensitive to any missing information and seems to fall into infinite loops between 2 functions in this case (this is the second time we have seen this). This is why it is important for us to move away from sympy's type inference (see issue #305 )

PR #318, opened 2020-05-15
Devel op

saidctb:


pyccel/parser/syntactic.py
```diff
@@ -439,51 +434,60 @@ def _visit_UnitaryOperatorNode(self, stmt):
...
+            return PyccelAdd(first, second)
 
+        elif stmt.value == '*':
             if isinstance(first, (PythonTuple, Tuple, List)):
                 return Dlist(first, second)
-            return Mul(first, second)
+            if cond:
+                args  = second.args
+                first = PyccelMul(first, args[0])
+                return second.func(first, *args[1:])
```
EmilyBourne, line 452: I'm not clear on what you are doing here to ensure that the brackets are created. Have you tested this thouroughly?

pyccel/parser/syntactic.py
```diff
@@ -494,10 +498,23 @@ def _visit_BooleanOperatorNode(self, stmt):
...
         second = self._visit(stmt.second)
+
+        assoc  = isinstance(stmt.second, AssociativeParenthesisNode)
+
         if stmt.value == 'and':
-            return And(first, second, evaluate=False)
+            if not assoc and isinstance(second, PyccelOr):
+                args  = second.args
+                first = PyccelAnd(first, args[0]  )
+                return  PyccelOr (first, *args[1:])
```
EmilyBourne, line 508: How do you know that a `PyccelOr` is required?

pyccel/ast/core.py
```diff
@@ -179,85 +189,105 @@ def __init__(self, var):
...
-    def __new__(cls, *args, evaluate = False, **kwargs):
-        return sp_Add.__new__(cls, *args, evaluate = evaluate, **kwargs)
+class PyccelPow(Expr, PyccelAstNode):
+    pass
 
+class PyccelAdd(Expr, PyccelAstNode):
     @property
     def rank(self):
         # TODO: Use broadcasting rules to decide shape (https://docs.scipy.org/doc/numpy/user/basics.broadcasting.html)
         return max(getattr(a,'rank',0) for a in self._args)
```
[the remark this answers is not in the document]
saidctb: yes it's done

pyccel/ast/type_inference.py
```diff
@@ -0,0 +1,93 @@
...
+        elif expr.is_real:
+            return 'real'
+        elif expr.is_complex:
+            return 'complex'
+        elif expr.is_Boolean:
+            return 'bool'
+    raise TypeError('Unknown datatype {0}'.format(type(expr)))
+
+
+def str_dtype(dtype):
```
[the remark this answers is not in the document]
saidctb: str_dtype converts the datatypes we have which are objects to a string

pyccel/ast/type_inference.py
```diff
@@ -0,0 +1,93 @@
...
+
+        is_real    = all(a=='integer' or a=='real' for a in args)
+        is_complex = all(a=='integer' or a=='real' or a=='complex' for a in args)
+        if is_real:
+            return 'real'
+        elif is_complex:
+            return 'complex'
+
+    elif isinstance(expr, (Variable, IndexedElement, DottedVariable)):
+        return str_dtype(expr.dtype)
```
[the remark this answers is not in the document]
saidctb: yes they are

pyccel/parser/semantic.py
```diff
@@ -1524,91 +1528,118 @@ def _visit_Add(self, expr, **settings):
         elif atoms_str:
             return Concatenate(args, False)
 
-        expr_new = Add(*args, evaluate=False)
+        expr_new = expr.func(*args)
```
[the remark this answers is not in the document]
saidctb: yes sympy gives us that possibility we can use  `expr.func`

pyccel/parser/semantic.py
```diff
@@ -1498,9 +1502,9 @@ def _visit_DottedVariable(self, expr, **settings):
         raise ValueError('attribute {} not found'.format(rhs_name))
 
 
-    def _visit_Add(self, expr, **settings):
+    def _visit_PyccelAdd(self, expr, **settings):
 
-        stmts, expr = extract_subexpressions(expr)
+        #stmts, expr = extract_subexpressions(expr)
```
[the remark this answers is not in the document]
saidctb: those lines where used in the case of functional expr, they should be handled elsewhere

pyccel/parser/semantic.py
```diff
@@ -1524,91 +1528,118 @@ def _visit_Add(self, expr, **settings):
...
+        #    expr_new = CodeBlock(stmts + [expr_new])
         return expr_new
 
+    def _visit_PyccelDiv(self, expr, **settings):
+        #stmts, expr = extract_subexpressions(expr)
+        #if stmts:
+        #    stmts = [self._visit(i, **settings) for i in stmts]
+        args = [self._visit(a, **settings) for a in expr.args]
+        if isinstance(args[0], (TupleVariable, PythonTuple, Tuple, List)):
+            expr_new = self._visit(Dlist(expr.args[0], expr.args[1]))
```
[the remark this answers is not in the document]
saidctb: yes you are right, I have deleted it

pyccel/parser/semantic.py
```diff
@@ -1524,91 +1528,118 @@ def _visit_Add(self, expr, **settings):
...
+        return expr_new
 
-    def _visit_Pow(self, expr, **settings):
+    def _visit_PyccelMod(self, expr, **settings):
+        #stmts, expr = extract_subexpressions(expr)
+        #if stmts:
+        #    stmts = [self._visit(i, **settings) for i in stmts]
+        args = [self._visit(a, **settings) for a in expr.args]
+        if isinstance(args[0], (TupleVariable, PythonTuple, Tuple, List)):
+            expr_new = self._visit(Dlist(expr.args[0], expr.args[1]))
```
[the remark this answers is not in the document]
saidctb: yes you are right, I have deleted it

pyccel/parser/semantic.py
```diff
@@ -1524,91 +1528,118 @@ def _visit_Add(self, expr, **settings):
...
+    def _visit_PyccelFloorDiv(self, expr, **settings):
+        #stmts, expr = extract_subexpressions(expr)
+        #if stmts:
+        #    stmts = [self._visit(i, **settings) for i in stmts]
         args = [self._visit(a, **settings) for a in expr.args]
-        expr_new = Pow(*args, evaluate=False)
-        if stmts:
-            expr_new = CodeBlock(stmts + [expr_new])
+        if isinstance(args[0], (TupleVariable, PythonTuple, Tuple, List)):
+            expr_new = self._visit(Dlist(expr.args[0], expr.args[1]))
```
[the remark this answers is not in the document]
saidctb: yes you are right, I have deleted it

pyccel/parser/syntactic.py
saidctb: it's been changed , and yes I have added the tests in `pyccel/scrtipy/expressions.py`

pyccel/parser/syntactic.py
saidctb: because we have two cases in `BooleanOperatorNode` either `and` or `or`

EmilyBourne: Yes, but the user might put brackets in even if it's not necessary

saidctb: okay i will make it general

saidctb: the brackets case is handled in the else condition, this condition enables us to evaluate the expression from left to right instead of doing it from right to left

pyccel/ast/core.py
```diff
@@ -179,85 +191,106 @@ def __init__(self, var):
...
-    def __new__(cls, *args, evaluate = False, **kwargs):
-        return sp_Pow.__new__(cls, *args, evaluate = evaluate, **kwargs)
-
-    def __init__(self, *args, evaluate = False, **kwargs):
+class PyccelOperator(Expr, PyccelAstNode):
+    @property
+    def rank(self):
         # TODO: Use broadcasting rules to decide shape (https://docs.scipy.org/doc/numpy/user/basics.broadcasting.html)
-        self._rank = max(getattr(a,'rank',0) for a in self._args)
+        return max(getattr(a,'rank',0) for a in self._args)
```
EmilyBourne, line 198: As discussed previously. Did we not want to move this to the constructor and the use the PyccelAstNode definition of def rank? The rank will not change and this prevents it being recalculated multiple times

pyccel/codegen/printing/fcode.py
```diff
@@ -1202,7 +1208,7 @@ def _print_Assign(self, expr):
 #                    code_args = ', '.join(self._print(i) for i in func.arguments)
 #                    code_args = '{0}, {1}'.format(code_args, lhs_code)
 #                else:
-#            print('code_args > {0}'.format(code_args))
+#           F print('code_args > {0}'.format(code_args))
```
EmilyBourne, line 1211: Typo?

tests/pyccel/scripts/expressions.py
```diff
@@ -0,0 +1,68 @@
...
+a2 = 2/2j
+
+a3 = 2//2
+a4 = 2.//2
+a5 = 2//4.
+
+a6 = 2%4
+a7 = 2%4.
+a8 = 2.%4
+
```
EmilyBourne, line 11: We should add more tests using the same operator multiple times to ensure the operator treatment order is correct

pyccel/parser/syntactic.py
```diff
@@ -439,65 +452,65 @@ def _visit_UnitaryOperatorNode(self, stmt):
...
 
         first = self._visit(stmt.first)
         second = self._visit(stmt.second)
+
         if stmt.value == 'and':
-            return And(first, second, evaluate=False)
+            if isinstance(second, PyccelOr):
+                args  = second.args
+                if isinstance(second, PyccelAnd):
+                    return PyccelAnd(first, *args)
```
EmilyBourne, line 499: second can't be both a PyccelOr and a PyccelAnd

pyccel/codegen/printing/fcode.py
saidctb: yes

pyccel/parser/syntactic.py
```diff
@@ -439,65 +452,65 @@ def _visit_UnitaryOperatorNode(self, stmt):
...
+                first = PyccelAnd(first, args[0]  )
+                return  PyccelOr (first, *args[1:])
```
EmilyBourne, line 501: This seems to carry out this conversion
And(a,Or(b,c)) -> Or(And(a,b), c)
These are not equivalent

pyccel/parser/syntactic.py
```diff
@@ -439,65 +452,65 @@ def _visit_UnitaryOperatorNode(self, stmt):
...
+            else:
+                return PyccelAnd(first, second)
+
         if stmt.value == 'or':
-            return Or(first, second, evaluate=False)
+            if isinstance(second, PyccelAnd):
+                args  = second.args
+                if isinstance(second, PyccelOr):
+                    return PyccelOr(first, *args)
```
EmilyBourne, line 509: See above. The same variable can't be a PyccelAnd and a PyccelOr

pyccel/parser/syntactic.py
```diff
@@ -439,65 +452,65 @@ def _visit_UnitaryOperatorNode(self, stmt):
...
+                first = PyccelOr(first, args[0])
+                return  PyccelAnd(first, *args[1:])
```
EmilyBourne, line 511: This seems to carry out this conversion
Or(a,And(b,c))->And(Or(a,b),c)
These are not equivalent

pyccel/ast/core.py
saidctb: yes we should but  it was not  previously on the constructor it was just a property,
so i left it as it is, I didn't want to implement it and preferred to leave it for another pull request

pyccel/parser/syntactic.py
saidctb: yes i will correct that

pyccel/parser/syntactic.py
saidctb: this carry out this conversion `a or b and c`

pyccel/parser/syntactic.py
saidctb: yes you are right

pyccel/parser/syntactic.py
saidctb: this carry out this conversion `a and b or c`

tests/pyccel/scripts/expressions.py
saidctb: done

EmilyBourne: I meant tests like the one you have added as e2, but for each operator

pyccel/parser/syntactic.py
EmilyBourne: So it changes the reading order to the correct left to right?
Can PyccelAnd and PyccelOr classes have more than 2 arguments?
If so then does the logic still work for longer expressions?

pyccel/parser/syntactic.py
EmilyBourne: This seems to be the same logic as above. I am resolving this conversation so we don't have the same conversation twice

pyccel/parser/syntactic.py
saidctb: not for the moment but maybe in the futur

pyccel/parser/semantic.py
```diff
@@ -1524,91 +1528,112 @@ def _visit_Add(self, expr, **settings):
...
         return expr_new
 
-    def _visit_Mul(self, expr, **settings):
-        stmts, expr = extract_subexpressions(expr)
-        if stmts:
-            stmts = [self._visit(i, **settings) for i in stmts]
+    def _visit_PyccelMul(self, expr, **settings):
+        #stmts, expr = extract_subexpressions(expr)
+        #if stmts:
+        #    stmts = [self._visit(i, **settings) for i in stmts]
```
yguclu, line 1540: What is the purpose of leaving these lines commented out instead of removing them?
(The same question applies to several other classes below.)

pyccel/parser/syntactic.py
```diff
@@ -103,6 +101,23 @@
...
+            b    = second.args[0]
+            c    = second.args[1]
+            a    = expr.func(a,b)
+            a    = change_priority(a)
+            expr = second.func(a,c)
+        else:
+            second = change_priority(second)
+            expr   = expr.func(first, second)
+
+    return expr
```
yguclu, line 119: This function seems to be of the paramount importance. Can you add some explanation in the form of a Python comment, or even better as a docstring?

pyccel/parser/syntactic.py
yguclu: I am also puzzled by this function. Maybe @saidctb wanted to implement this identity instead?
`a and (b or c) == (a and b) or (a and c)`

tests/pyccel/scripts/expressions.py
saidctb: done

yguclu: To make it easier to add more tests, we should only print the results of the various expressions, without making any comparison to a known value. In other words, I would change `print(a1==1.0)` to `print(a1)` and so on. We may need to collect the expressions based on the output type, but this should not be a problem.

saidctb: I have tried that but they were some bugs in how we parse the results in test_pyccel
so i preferred to make the result as a boolean this works as an assert

pyccel/parser/syntactic.py
saidctb: here there is no brackets, it only changes the evaluation from left to right instead of right to left
so when we have something like `a or b and c` redbaron gives us `Or(a,And(b,c))` and i change that to `And(Or(a,b),c)` the printing of these two expression is the same because there is no brackets here

yguclu: Cool. Thanks

pyccel/ast/core.py
yguclu: I don't see what you mean. The original code for `Pow` was setting `self._rank` in the `__init__`, but now you removed that. If you do not need an `__init__` in the subclasses of `PyccelOperator`, this one will be called automatically.

pyccel/ast/numpyext.py
```diff
@@ -196,12 +201,13 @@ class NumpySum(Function):
...
 
     def __new__(cls, arg):
-        if not isinstance(arg, (list, tuple, PythonTuple, Tuple, List, Variable, Mul, Add, Pow, sp_Rational)):
+        if not isinstance(arg, (list, tuple, PythonTuple, Tuple, List, Variable, Expr)):
             raise TypeError('Uknown type of  %s.' % type(arg))
 
         return Basic.__new__(cls, arg)
 
     def __init__(self, arg):
+        from pyccel.ast.type_inference import sp_dtype, str_dtype
```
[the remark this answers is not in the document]
saidctb: yes

pyccel/parser/semantic.py
saidctb: we need to keep them to uncomment them later or move them to another place when we introduce the functional programming

yguclu: Fine with me, but we should make this clear. Maybe add a TODO statement somewhere that we can quickly search?

pyccel/ast/type_inference.py
```diff
@@ -75,8 +81,14 @@ def sp_dtype(expr):
 def str_dtype(dtype):
 
     """
-    return a sympy datatype as string
-    dtype: str, Native Type
+    This function takes a datatype and returns a sympy datatype as a string
```
yguclu, line 84: @saidctb Thanks for adding the docstring!
Do we really return the string form of a sympy datatype? Or is it our datatype?

tests/pyccel/scripts/expressions.py
yguclu: @EmilyBourne I thought we fixed the issue with the output comparison in `test_pyccel` a while ago. Do you remember?

pyccel/ast/core.py
saidctb: the diff here is wrong  you can verify the `Add`, `Mul` methods this what was changes the `Pow`
was completely removed and not just changed

tests/pyccel/scripts/expressions.py
saidctb: yes it still gives me errors in this script

pyccel/ast/type_inference.py
saidctb: a string

pyccel/parser/semantic.py
saidctb: done

tests/pyccel/scripts/expressions.py
EmilyBourne: I thought so too. I have just had a look. It is pretty basic so I would imagine that the problem is coming from complexes.
A complex in fortran is printed as 2 doubles but in python a completely imaginary complex will be printed as 1 double. This will then give problems for all subsequent results

EmilyBourne: It is therefore probably this line:
`print(a10==10j)`
that is causing problems.

The temporary workaround while waiting for a pull request with a fix would be to define the output type of this object as "float" but it's not a great solution

pyccel/ast/core.py
yguclu: Is there a reason for not writing the following?
```

    def __init__(self, *args, **kwargs):
        self._rank = max(getattr(a,'rank',0) for a in self._args)
```
This should work and it would be called automatically by any subclass that does not reimplement an `__init__` method. (I don't think `Expr` has an `__init__` method, so there should be no ambiguity on which constructor gets called.)

saidctb: yes we can , but I prefer to keep the attribute the None for the moment otherwise we will have to add them to all the nodes and modify the type inference mechanism to take that into account,
it's best to do it separately in another pull request

tests/pyccel/scripts/expressions.py
saidctb: yes i remember the problem is there

pyccel/parser/syntactic.py

tests/pyccel/scripts/expressions.py
yguclu: OK then, let's keep things as they are. We are going to fix it later.

pyccel/ast/core.py
yguclu: OK. Do we have a related issue where we can write this down?

EmilyBourne: I don't understand. Why does already having a rank for these objects mess up the type inference mechanism?

saidctb: it doesn't mess up anything , just wanted to do it separately

saidctb: yes @yguclu  we have this
https://github.com/pyccel/pyccel/issues/305

yguclu:
I think this PR is ready to be merged. @EmilyBourne: do you see something missing?

tests/pyccel/scripts/expressions.py
EmilyBourne: As things are meaning without the additional tests? Or just with the boolean results?

pyccel/ast/core.py
EmilyBourne: This issue does not mention initialising values in the constructor rather than at each function call. Do you want to add that stipulation to the issue?
Personally I don't see how the 2 are related. Calling X.rank already gives a result at the moment even if _rank is still None. I'm worried that if we don't move this code while we remember it then we will forget as an error cannot appear due to this

saidctb: it does mention that , `we may need to use the PyccelAstNode attributes where we infer the type automatically at the instance creation`.
the thing is those attributes where never used before so adding them shouldn't give any error because they were not used only in the basic cases like the `Variable` class

tests/pyccel/scripts/expressions.py
saidctb: yes we use the boolean value for the moment and fix the the problem of complex values later

EmilyBourne: So then my initial comment that we need additional tests which have the same operator twice in the expression to ensure that the left to right or right to left precedence is respected still holds

pyccel/ast/core.py
EmilyBourne: May and should are not the same thing ^^
I know they won't give an error. I meant that when we start using the attribute consistently we might use errors to track down the places where the rank has still not been initialised. In that case this wouldn't show up.
Given that it is a 1 line commit I would have thought it would have been easier to do it here the first time it's used rather than add the bad practice code to master and risk missing it when you come back to it. But if you prefer to do it later then ok. I have added a comment to the issue to clarify that we need to ensure initialisation is carried out in `__init__` where possible

tests/pyccel/scripts/expressions.py
yguclu: @EmilyBourne You're right, we should add more tests before closing this PR. Something like
```
117. / 39. / 3. == 1.   # not 9.
117 // 39 // 3  == 1    # not 9
23  %  12 %  7  == 4    # not 3
2 ** 3 ** 2     == 512  # not 64
```

saidctb: https://github.com/pyccel/pyccel/blob/374d6c5e5b2559c7cab30f88b1a2f8f446004417/tests/pyccel/scripts/expressions.py#L33
I have already added them

EmilyBourne: Ah sorry I missed that. ** seems to be missing though

yguclu: @saidctb Can you add a test with multiple successive power operators `**`?

saidctb: yes I have just added them

yguclu: Perfect, thanks!

PR #319, opened 2020-05-15
Fix epyccel for C on Windows

EmilyBourne:


EmilyBourne:
Related to https://bugs.python.org/issue32535

PR #322, opened 2020-05-19
Devel ast nodes

saidctb:


pyccel/ast/builtins.py
```diff
@@ -297,83 +257,69 @@ def __len__(self):
...
-    def dtype(self):
-        return 'int'
-
-    @property
-    def rank(self):
-        return 0
-
 #==============================================================================
-class List(Tuple):
+class List(Tuple, PyccelAstNode):
```
EmilyBourne, line 282: Should this be called PythonList (like PythonTuple) to differentiate from the sympy object?

pyccel/parser/syntactic.py
```diff
@@ -172,10 +176,11 @@ def __init__(self, inputs, **kwargs):
                           severity='fatal')
 
         preprocess_imports(red)
-        preprocess_default_args(red)
+        #preprocess_default_args(red)
```
EmilyBourne, line 179: Is this broken or handled elsewhere now?

pyccel/parser/semantic.py
```diff
@@ -670,47 +675,31 @@ def _infere_type(self, expr, **settings):
...
-            d_var['datatype'   ] = 'str'
-            d_var['allocatable'] = False
-            d_var['rank'       ] = 0
-            return d_var
-
-        elif isinstance(expr, ImaginaryUnit):
-
-            d_var['datatype'   ] = 'complex'
-            d_var['allocatable'] = False
+            d_var['datatype'   ] = '*'
```
EmilyBourne, line 698: What is a * datatype?

pyccel/ast/core.py
```diff
@@ -174,13 +175,77 @@
...
+            new_shape.append(e1)
+        elif e1 == 1:
+            new_shape.append(e2)
+        elif e2 == 2:
+            new_shape.append(e1)
+        else:
+            msg = 'operands could not be broadcast together with shapes {} {}'
+            msg = msg.format(shape_1, shape_2)
+            raise ValueError(msg)
+    return tuple(new_shape)
```
EmilyBourne, line 204: Is this the same function as in pyccel/ast/builtins.py? If so why is it copied and pasted?

pyccel/codegen/printing/fcode.py
```diff
@@ -2213,6 +2279,36 @@ def _print_ConstructorCall(self, expr):
...
+        code_args = ', '.join(self._print(i) for i in expr.args)
+        code = '{0}({1})'.format(func_name, code_args)
+        return self._get_statement(code)
+
+    def _print_NumpySqrt(self, expr):
+        arg = expr.args[0]
+        code_args = self._print(arg)
+        if arg.dtype is NativeInteger() or arg.dtype is NativeBool():
+            code_args = 'Real({})'.format(code_args)
+        code = 'sqrt({})'.format(code_args)
```
EmilyBourne, line 2301: Why do we have the dictionary numpy_ufunc_to_fortran and then not use it? Is sqrt a special case?

pyccel/codegen/printing/fcode.py
```diff
@@ -2213,6 +2279,36 @@ def _print_ConstructorCall(self, expr):
...
+        return self._get_statement(code)
+
+    def _print_MathSqrt(self, expr):
+        arg = expr.args[0]
+        code_args = self._print(arg)
+        if arg.dtype is NativeInteger() or arg.dtype is NativeBool():
+            code_args = 'Real({})'.format(code_args)
+        code = 'sqrt({})'.format(code_args)
```
EmilyBourne, line 2309: See above. Why do we have the dictionary math_ufunc_to_fortran and then not use it? Is sqrt a special case?

pyccel/ast/numpyext.py
```diff
@@ -376,21 +375,21 @@ def fprint(self, printer, lhs = None):
 
 def Shape(arg):
     if arg.shape is None:
-        return PythonTuple([PyccelArraySize(arg,i) for i in range(arg.rank)])
+        return PythonTuple(*[PyccelArraySize(arg,i) for i in range(arg.rank)])
```
EmilyBourne, line 378: Can we create a PythonTuple from a generator or do we need this *[] ?

pyccel/ast/numpyext.py
```diff
@@ -1244,155 +1150,89 @@ def __init__(self,*args):
...
+    """Base class for Numpy's universal functions."""
 
+#------------------------------------------------------------------------------
+class NumpyUfuncUnary(NumpyUfuncBase):
+    """Numpy's universal function with one argument.
+    """
+    def __init__(self, x):
+        self._shape     = x.shape
+        self._rank      = x.rank
+        self._dtype     = x.dtype if x.dtype is NativeComplex() else NativeReal()
```
EmilyBourne, line 1170: At least some of numpy's functions return the same dtype int the output as the argument. E.g. floor called on an integer returns an integer. I don't know whether floor is the exception or the rule though?

pyccel/ast/builtins.py
```diff
@@ -245,49 +181,73 @@ def fprint(self, printer):
...
-
-    @property
-    def shape(self):
-        if(self._arg_dtypes is None):
-            return [Integer(len(self._args))]
+    def __init__(self, *args, **kwargs):
+        if self.stage == 'syntactic' or len(args) == 0:
+            return
+        is_homogeneous = all(args[0].dtype==a.dtype for a in args[1:])
+        is_homogeneous = is_homogeneous and all(args[0].rank==a.rank   for a in args[1:])
```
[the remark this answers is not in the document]
saidctb: yes all the arguments must have the same type, shape and rank
shape cannot be always checked so I leave that for the fortran compiler to raise an error

pyccel/ast/builtins.py
```diff
@@ -245,49 +181,73 @@ def fprint(self, printer):
...
+                else:
+                    raise TypeError('cannot determine the type of {}'.format(self))
+                
+                shapes = [a.shape for a in args]
+                
+                if all(sh is not None for sh in shapes):
+                    if len(args) == 1:
+                        shape = args[0].shape
+                    else:
+                        shape = broadcast(args[0].shape, args[1].shape)
```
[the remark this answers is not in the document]
saidctb: your example is not gonna work for homogeneous tuples

pyccel/ast/builtins.py
```diff
@@ -297,83 +257,69 @@ def __len__(self):
...
+class List(Tuple, PyccelAstNode):
     """ Represent lists in the code with dynamic memory management."""
-
+    def __init__(self, *args, **kwargs):
+        if self.stage == 'syntactic':
+            return
+        integers  = [a for a in args if a.dtype is NativeInteger() or a.dtype is NativeBool()]
+        reals     = [a for a in args if a.dtype is NativeReal()]
+        complexes = [a for a in args if a.dtype is NativeComplex()]
+        strs      = [a for a in args if a.dtype is NativeString()]
```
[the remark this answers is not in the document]
saidctb: yes we should

pyccel/ast/builtins.py
```diff
@@ -297,83 +257,69 @@ def __len__(self):
...
+            else:
+                raise TypeError('cannot determine the type of {}'.format(self))
+            
+            shapes = [a.shape for a in args]
+            
+            if all(sh is not None for sh in shapes):
+                if len(args) == 1:
+                    shape = args[0].shape
+                else:
+                    shape = broadcast(args[0].shape, args[1].shape)
```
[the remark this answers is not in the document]
saidctb: fixed

pyccel/ast/builtins.py
```diff
@@ -511,3 +457,33 @@ def python_builtin_datatype(name):
...
+        new_shape_1 = shape_1
+    
+    new_shape = []
+    for e1,e2 in zip(new_shape_1, new_shape_2):
+        if e1 == e2:
+            new_shape.append(e1)
+        elif e1 == 1:
+            new_shape.append(e2)
+        elif e2 == 2:
+            new_shape.append(e1)
```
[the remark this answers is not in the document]
saidctb: yes we should not use broadcasting rules , so it had been deleted

pyccel/ast/core.py
saidctb: yes it's the same ,now there is only one copy

pyccel/ast/numpyext.py
saidctb: yes we can create it from a generator

pyccel/ast/numpyext.py
saidctb: floor is the exception, most of the functions returns a real or a complex

pyccel/codegen/printing/fcode.py
saidctb: those are special cases ,sqrt need an argument of type real for example

pyccel/codegen/printing/fcode.py
saidctb: yes

pyccel/parser/semantic.py
saidctb: it's a generic datatype, it's just a way of saying undetermined

pyccel/parser/syntactic.py
saidctb: it's handled elsewhere in Functioncall and print_Functioncall

tests/pyccel/scripts/runtest_tuples.py
```diff
@@ -7,8 +7,11 @@
 i = e[2]
 ai_0 = 44
 
-tup1 = (0, False, 3+1j) * 3
-tup2 = tup1 * 2
+#tup1 = (0, False, 3+1j) * 3
+#tup2 = tup1 * 2
```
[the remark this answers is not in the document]
saidctb: yes i will add issues for that

tests/symbolic/test_symbolic.py
```diff
@@ -13,6 +13,7 @@
 files = [f for f in files if (f.endswith(".py"))]
 
 @pytest.mark.parametrize( "f", files )
+@pytest.mark.xfail
```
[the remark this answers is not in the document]
saidctb: yes

pyccel/parser/syntactic.py
EmilyBourne: Then can we get rid of this line and the associated function?

pyccel/ast/builtins.py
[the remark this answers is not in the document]
EmilyBourne: Ah, sorry I meant to write shape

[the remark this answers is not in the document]
EmilyBourne: The shape must always be a known value or a PyccelArraySize right?
I think it would be better to make the tuple unhomogeneous if the sizes aren't the same or aren't known

pyccel/codegen/printing/fcode.py
EmilyBourne: Ok. But we could still use the value stored in the dictionary. It just means that if we want to change the function some day (unlikely for sqrt but it has already happened for rand for example) then it is easy to know where to look to change that

pyccel/ast/numpyext.py
EmilyBourne: Then we should create it from a generator as this is more efficient

pyccel/parser/semantic.py
EmilyBourne: Why is the datatype undetermined here? Surely the datatype of a PythonInt is 'int' for example?

tests/symbolic/test_symbolic.py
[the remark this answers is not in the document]
EmilyBourne: I didn't realise we had so few tests in the symbolic stage. That explains why they all fail. We should add reasons to the xfail mark though. See tests/codegen/ccode/test_ccode_codegen.py for an example of providing different reasons for a parameterised function

pyccel/codegen/printing/fcode.py
EmilyBourne: See above

pyccel/ast/builtins.py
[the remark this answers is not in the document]
saidctb: if the user writes a function like
```
def f(x):
    y = (x[0], x[1])
```
it will be inhomogeneous because we don't know the shape of `x`, I think it's better to let it be homogeneous by default and the exception should be inhomogeneous

pyccel/codegen/printing/fcode.py
saidctb: we are using the dictionary in `_print_NumpyUfuncBase`

pyccel/ast/builtins.py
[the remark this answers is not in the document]
EmilyBourne: Maybe we can discuss this this afternoon.
Personally I disagree. If the user wanted something contiguous then they would have probably written:
```
def f(x):
    y = x[0:2]
```
Tuples are usually used for inhomogeneous data. We have a specialised case for homogeneous data to improve performance but if we make homogeneous the default then the behaviour will be broken in cases with different shapes without there being an alternative way to write the same thing. Whereas there is an alternative way to write the homogeneous case

pyccel/codegen/printing/fcode.py
EmilyBourne: Yes. But the element 'sqrt' of the dictionary is never used. Either it should be used here so that changes to functions are always carried out in the same place even if it is a special case. Or it shouldn't be in the dictionary so that it is clear that this is not used

pyccel/ast/numpyext.py
saidctb: yes it's more memory efficient

pyccel/parser/semantic.py
saidctb: it's not an 'int' it's a python data type so the real type will be type(int) and not int we can add that later but as long as it doesn't get generated we are fine.

pyccel/codegen/printing/fcode.py
saidctb: yes it shouldn't be in the dictionary

pyccel/parser/syntactic.py
EmilyBourne: This line has been removed however the function itself is still in parser/utilities.py. Is it used elsewhere? I didn't think it was. If not then we should remove the definition too

tests/internal/scripts/mpi/nonblocking.py
```diff
@@ -40,15 +40,12 @@
...
-after  = rank + 1
+before = np.int32(rank - 1)
+after  = np.int32(rank + 1)
 if rank == 0:
     before = size - 1
 if rank == size - 1:
     after  = 0
-before = np.int32(before)
-after  = np.int32(after )
-# ...
```
yguclu, line 51: The original code was ensuring that the variables `before` and `after` passed to `mpi_irecv` are of type `int32`. The new code passes a 32-bit or a 64-bit integer depending on the MPI rank, so it does not look correct to me. What was the reason for this change?

tests/pyccel/scripts/runtest_tuples.py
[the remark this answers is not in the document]
yguclu: Do we have an issue now?

pyccel/ast/builtins.py
yguclu: Yes, we need a consistent naming scheme for our AST nodes. I plan on doing that in the near future.

pyccel/ast/numpyext.py
```diff
@@ -21,35 +20,43 @@
...
-    'Acot',
-    'Acsc',
-    'Array',
-    'Asec',
-    'Asin',
-    'Atan',
-    'Bounds',
-    'Complex',
-    'Complex64',
-    'Complex128',
```
yguclu, line 46: Did we remove some names by accident? `Complex64` and `Complex128` should probably be added back to the list (next to `Float32` and `Float64`)

pyccel/ast/numpyext.py
```diff
@@ -1160,7 +1067,7 @@ def __new__(cls, a, dtype=None, order='K', subok=True):
 
 #=======================================================================================
 
-class Bounds(Basic):
+class Bounds(Basic, PyccelAstNode):
```
yguclu, line 1070: I don't think Numpy has a `bounds` function. Is this class ever used?

pyccel/parser/syntactic.py
```diff
@@ -172,10 +176,10 @@ def __init__(self, inputs, **kwargs):
                           severity='fatal')
 
         preprocess_imports(red)
-        preprocess_default_args(red)
```
yguclu, line 175: @saidctb Can you explain why we don't need to preprocess the default arguments anymore?

pyccel/parser/syntactic.py
```diff
@@ -649,14 +669,16 @@ def _visit_DefNode(self, stmt):
...
             if results:
                 txt += ' results(' + ','.join(results) + ')'
+
             header = hdr_parse(stmts=txt)
             if name in self.namespace.static_functions:
                 header = header.to_static()
 
         body = stmt.value
-
+
```
yguclu, line 681: Trailing white spaces?

pyccel/parser/semantic.py
```diff
@@ -670,47 +671,31 @@ def _infere_type(self, expr, **settings):
...
             d_var['rank'       ] = 0
-            d_var['precision'  ] = default_precision['complex']
+            d_var['precision'  ] = 0
             return d_var
-
+
```
yguclu, line 698: Trailing white spaces?

pyccel/parser/syntactic.py
```diff
@@ -569,7 +573,24 @@ def _visit_DefArgumentNode(self, stmt):
...
+                new_stmt  = Assign(new_vars[-1], result)
+                new_stmt.set_fst(stmt)
+                assigns.append(new_stmt)
+            else:
+                new_vars.append(result)
+
+        if assigns:
+            expr = Return(new_vars, CodeBlock(assigns))
+        else:
+            expr = Return(new_vars)
```
yguclu, line 593: @saidctb I see you moved the treatment of multiple return values from the semantic to the syntactic stage. What is the advantage of this approach?

pyccel/codegen/printing/fcode.py
yguclu: There are different ways to handle the special case of `numpy.sqrt`. If we stick to this approach based on strings, I also agree that `sqrt` should not be a key of the dictionary `numpy_ufunc_to_fortran`.

pyccel/codegen/printing/fcode.py
```diff
@@ -79,22 +78,78 @@
...
+    "conjugate": "conjg"  # TODO: move to numpyext
+}
+
+numpy_ufunc_to_fortran = {
+    'NumpyAbs'  : 'abs',
+    'NumpyFloor': 'floor',  # TODO: might require special treatment with casting
+    # ---
+    'NumpyExp' : 'exp',
+    'NumpyLog' : 'Log',
+    'NumpySqrt': 'Sqrt',
```
yguclu, line 91: Since this dictionary entry is not used (`sqrt` is a special case), for the sake of clarity I suggest to comment this line out and add a comment.

pyccel/codegen/printing/fcode.py
```diff
@@ -79,22 +78,78 @@
...
+    'MathLgamma' : 'log_gamma',
+    'MathLog'    : 'log',
+    'MathLog10'  : 'log10',
+#    'MathLog1p'  : '???', # TODO
+#    'MathLog2'   : '???', # TODO
+#    'MathPow'    : '???', # TODO
+#    'MathRadians': '???', # TODO
+    'MathSin'    : 'sin',
+    'MathSinh'   : 'sinh',
+    'MathSqrt'   : 'sqrt',
```
yguclu, line 139: As above: since this dictionary entry is not used (`sqrt` is a special case), for the sake of clarity I suggest to comment this line out and add a comment.

tests/internal/scripts/mpi/nonblocking.py
saidctb: the type and precision is defined in the first assignment not the last one

pyccel/ast/numpyext.py
saidctb: I'm not sure maybe it's used

pyccel/ast/numpyext.py
saidctb: yes we should add them

pyccel/ast/core.py
```diff
@@ -2160,7 +2300,7 @@ def __new__(
                 precision = default_precision['complex']
             elif isinstance(dtype, NativeBool):
                 precision = default_precision['bool']
-
+
```
yguclu, line 2303: Trailing white spaces?

pyccel/parser/syntactic.py
saidctb: because it's handled in FunctionCall where we add the missing argument if they where optional

pyccel/parser/semantic.py
```diff
@@ -1170,12 +1019,12 @@ def _visit_list(self, expr, **settings):
         return Tuple(*ls, sympify=False)
 
     def _visit_tuple(self, expr, **settings):
-        ls = [self._visit(i, **settings) for i in expr]
-        return PythonTuple(ls)
+        ls = tuple(self._visit(i, **settings) for i in expr)
+        return ls
```
yguclu, line 1023: When do we visit a `tuple` instead of a `PythonTuple` or a Sympy `Tuple`?

pyccel/parser/semantic.py
```diff
@@ -2143,13 +1880,11 @@ def _visit_Assign(self, expr, **settings):
             stmts[-1] = stmt
             return CodeBlock(stmts)
 
-        elif isinstance(rhs, Application):
+        elif isinstance(rhs, FunctionCall):
 
             # ARA: needed for functions defined only with a header
```
yguclu, line 1885: Is this comment still accurate? If not we should modify it or remove it.

pyccel/parser/semantic.py
```diff
@@ -2178,7 +1913,7 @@ def _visit_Assign(self, expr, **settings):
 
                 # TODO imporve this will not work for
                 # the case of different results types
```
yguclu, line 1915: Is this comment still accurate?

pyccel/parser/syntactic.py
saidctb: because we need to have access to the results names in FunctionDef  when we visit it in semantics 
otherwise we can't  only after annotation

pyccel/parser/semantic.py
saidctb: I'm not sure

pyccel/parser/semantic.py
saidctb: yes

pyccel/parser/semantic.py
saidctb: no

tests/internal/scripts/mpi/nonblocking.py
EmilyBourne: Yes but when you do:
```
if rank == 0:
    before = size -1
```
you aren't using a numpy.int32 anymore.  This should raise an error I think (it may be related to issue 325)

tests/pyccel/scripts/runtest_tuples.py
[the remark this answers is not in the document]
EmilyBourne: I have created an issue now

pyccel/parser/semantic.py
EmilyBourne: According to codacy-coverage this function is used, but _visit_Tuple isn't

pyccel/ast/numpyext.py
EmilyBourne: A grep search shows that this class has an associated `_print_Bounds` function in `pyccel/codegen/printing/fcode.py` however it is not mentioned anywhere else in the codebase. The codacy-coverage results also indicate that it is not used.

The printing function looks like this:
```
def _print_Bounds(self, expr):
    var = expr.var
    rank = var.rank
    bounds = []
    for i in range(0, rank):
        l = 'lbound({var},{i})'.format(var=self._print(var),
                                       i=self._print(i+1))
        u = 'ubound({var},{i})'.format(var=self._print(var),
                                       i=self._print(i+1))

        bounds.append('{lbound}:{ubound}'.format(lbound=l, ubound=u))
    bounds = ','.join(bounds)
    return bounds
```
which means that Bounds does not represent any python concept

tests/pyccel/scripts/runtest_tuples.py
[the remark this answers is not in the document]
saidctb: ah perfect, thanks

pyccel/ast/numpyext.py
saidctb: I think we used that before to allocate memory for unknown shapes like this example
```
def f(x):
     y = zeros_like(x)
```
because here the shape of x is unknown

tests/symbolic/test_symbolic.py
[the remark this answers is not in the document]
saidctb: I have added an issue for that

tests/internal/scripts/mpi/nonblocking.py
saidctb: for the moment we are not strict on the precision and we  assign it from the first assign
so if we do 
```
a = int32(1)
a =  1
```
we still keep the first precision so changing that behaviour is another thing

tests/symbolic/test_symbolic.py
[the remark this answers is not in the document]
EmilyBourne: A reason in the xfail is still useful. Even if that reason just points back to the issue (see tests/external/test_external.py).
Although in this case I doubt that the reason why `decorator.py` fails is at all related to why `lambdas.py` fails. If you create an issue grouping all 3 problems then that issue will be much harder to close

tests/internal/scripts/mpi/nonblocking.py
EmilyBourne: This is why it currently works. However if we added a test on the return type or fix issue 325 then this will no longer be true. It is not good practice to introduce bugs in the test code as we will have errors when we have fixed pyccel's behaviour rather than having errors to show where it is broken

pyccel/parser/syntactic.py
yguclu: That makes sense, thanks!

tests/internal/scripts/mpi/nonblocking.py
saidctb: okay it's fixed

pyccel/parser/syntactic.py
```diff
@@ -59,9 +59,13 @@
 from pyccel.ast.core import PyccelPow, PyccelAdd, PyccelMul, PyccelDiv, PyccelMod, PyccelFloorDiv
 from pyccel.ast.core import PyccelEq,  PyccelNe,  PyccelLt,  PyccelLe,  PyccelGt,  PyccelGe
 from pyccel.ast.core import PyccelAnd, PyccelOr,  PyccelNot, PyccelMinus, PyccelAssociativeParenthesis
-from pyccel.ast.core  import PyccelOperator, PyccelUnary
+from pyccel.ast.core import PyccelOperator, PyccelUnary
 
-from pyccel.parser.utilities import fst_move_directives, preprocess_imports, preprocess_default_args
+from pyccel.ast.numbers import Complex
+
+from pyccel.ast.builtins import PythonComplex
```
EmilyBourne, line 66: Unused import

pyccel/parser/semantic.py
yguclu: If we don't visit a Sympy `Tuple` anymore, I recommend that we remove the `visit_Tuple` function from our code!

As for my original question, I should rephrase it: where is a Python `tuple` created in our AST? During the syntactic stage or the semantic stage? And is a Python `tuple` the correct node to have there?

pyccel/parser/syntactic.py
EmilyBourne: In the same file the following are also now unused:
```
parser/syntactic.py:5:0: W0611: Unused import traceback (unused-import)
parser/syntactic.py:13:0: W0611: Unused BinaryOperatorNode imported from redbaron (unused-import)
parser/syntactic.py:22:0: W0611: Unused AssociativeParenthesisNode imported from redbaron (unused-import)
parser/syntactic.py:54:0: W0611: Unused Dlist imported from pyccel.ast (unused-import)
parser/syntactic.py:66:0: W0611: Unused PythonComplex imported from pyccel.ast.builtins (unused-import)
```

pyccel/parser/semantic.py
```diff
@@ -960,104 +884,35 @@ def _infere_type(self, expr, **settings):
...
-                d_var['datatype'] = sp_dtype(expr)
+                d_var['datatype'   ] = expr.dtype
+                d_var['is_pointer' ] = False
+                d_var['precision'  ] = expr.precision
+
+            elif isinstance(expr, MathFunctionBase):
+                d_var['datatype' ] = expr.dtype
+                d_var['precision'] = expr.precision
+                d_var['rank'     ] = expr.rank
+                d_var['shape'    ] = expr.shape
```
EmilyBourne, line 897: I am surprised that these 2 cases are not handled in the same way. Does NumpyUfuncBase not have a shape and rank? I would have expected this to be necessary to handle broadcasting rules?

pyccel/codegen/printing/fcode.py
```diff
@@ -523,7 +578,7 @@ def _print_AnnotatedComment(self, expr):
         return '!${0} {1}'.format(accel, txt)
 
     def _print_Tuple(self, expr):
-        shape = list(reversed(numpy_shape(expr)))
+        shape = list(reversed(asarray(expr).shape))
```
EmilyBourne, line 581: Does a tuple not have a shape?

pyccel/codegen/printing/fcode.py
```diff
@@ -48,23 +48,22 @@
 from pyccel.ast.core import PyccelEq,  PyccelNe,  PyccelLt,  PyccelLe,  PyccelGt,  PyccelGe
 from pyccel.ast.core import PyccelAnd, PyccelOr,  PyccelNot, PyccelMinus, PyccelAssociativeParenthesis
```
EmilyBourne, line 49: Everything except `PyccelMinus` is an unused import

EmilyBourne:
```
codegen/printing/fcode.py:18:0: W0611: Unused S imported from sympy.core (unused-import)
codegen/printing/fcode.py:21:0: W0611: Unused precedence imported from sympy.printing.precedence (unused-import)
codegen/printing/fcode.py:47:0: W0611: Unused PyccelPow imported from pyccel.ast.core (unused-import)
codegen/printing/fcode.py:47:0: W0611: Unused PyccelMod imported from pyccel.ast.core (unused-import)
codegen/printing/fcode.py:47:0: W0611: Unused PyccelFloorDiv imported from pyccel.ast.core (unused-import)
codegen/printing/fcode.py:48:0: W0611: Unused PyccelEq imported from pyccel.ast.core (unused-import)
codegen/printing/fcode.py:48:0: W0611: Unused PyccelNe imported from pyccel.ast.core (unused-import)
codegen/printing/fcode.py:48:0: W0611: Unused PyccelLt imported from pyccel.ast.core (unused-import)
codegen/printing/fcode.py:48:0: W0611: Unused PyccelLe imported from pyccel.ast.core (unused-import)
codegen/printing/fcode.py:48:0: W0611: Unused PyccelGt imported from pyccel.ast.core (unused-import)
codegen/printing/fcode.py:48:0: W0611: Unused PyccelGe imported from pyccel.ast.core (unused-import)
codegen/printing/fcode.py:49:0: W0611: Unused PyccelAnd imported from pyccel.ast.core (unused-import)
codegen/printing/fcode.py:49:0: W0611: Unused PyccelOr imported from pyccel.ast.core (unused-import)
codegen/printing/fcode.py:49:0: W0611: Unused PyccelNot imported from pyccel.ast.core (unused-import)
codegen/printing/fcode.py:49:0: W0611: Unused PyccelAssociativeParenthesis imported from pyccel.ast.core (unused-import)
```

pyccel/ast/numpyext.py
```diff
@@ -21,35 +20,45 @@
 from .core import PyccelEq,  PyccelNe,  PyccelLt,  PyccelLe,  PyccelGt,  PyccelGe
 from .core import PyccelAnd, PyccelOr,  PyccelNot, PyccelMinus, PyccelAssociativeParenthesis
```
EmilyBourne, line 21: Unused imports:
```
ast/numpyext.py:20:0: W0611: Unused PyccelAdd imported from core (unused-import)
ast/numpyext.py:20:0: W0611: Unused PyccelMul imported from core (unused-import)
ast/numpyext.py:20:0: W0611: Unused PyccelDiv imported from core (unused-import)
ast/numpyext.py:20:0: W0611: Unused PyccelMod imported from core (unused-import)
ast/numpyext.py:20:0: W0611: Unused PyccelFloorDiv imported from core (unused-import)
ast/numpyext.py:21:0: W0611: Unused PyccelEq imported from core (unused-import)
ast/numpyext.py:21:0: W0611: Unused PyccelNe imported from core (unused-import)
ast/numpyext.py:21:0: W0611: Unused PyccelLt imported from core (unused-import)
ast/numpyext.py:21:0: W0611: Unused PyccelLe imported from core (unused-import)
ast/numpyext.py:21:0: W0611: Unused PyccelGt imported from core (unused-import)
ast/numpyext.py:21:0: W0611: Unused PyccelGe imported from core (unused-import)
ast/numpyext.py:22:0: W0611: Unused PyccelAnd imported from core (unused-import)
ast/numpyext.py:22:0: W0611: Unused PyccelOr imported from core (unused-import)
ast/numpyext.py:22:0: W0611: Unused PyccelNot imported from core (unused-import)
```

pyccel/ast/numpyext.py
yguclu: So I suppose we now handle this with classes `Shape` and `PyccelArraySize` in `numpyext`? The array bounds are never needed, because in the generated Fortran code we use 0-based indexing. The same goes for the generated C code.

pyccel/parser/semantic.py
yguclu: If we have multiple results with different types, shouldn't `rhs.dtype` be a PythonTuple?

EmilyBourne:
There seems to be a surprisingly large number of new (unused import) warnings.
To see where use the following command:
`pylint --disable=all --enable=W0611 pyccel/pyccel/*`

pyccel/codegen/printing/fcode.py
yguclu: Yes, but I think it does not return the "correct" multi-dimensional shape. This was the problem with using `numpy.shape(expr)`,which was searching for `expr.shape` and returning it.

pyccel/ast/numpyext.py
saidctb: yes you are right

pyccel/parser/semantic.py
saidctb: maybe it's not the correct node and we shouldn't have it, so it must be created somewhere

pyccel/parser/syntactic.py
saidctb: they were never properly working, but now they are broken completely

pyccel/parser/semantic.py
saidctb: it should be handled the same way

pyccel/parser/semantic.py
EmilyBourne: I think this may be coming from the shape. If we translate:
```
from numpy import shape, ones
x = ones((3,4))
s = shape(x)
```
Then we will collect the shape of x which is stored in a `tuple` not a `PythonTuple`

pyccel/parser/syntactic.py
EmilyBourne: Is this comment on the wrong thread?

pyccel/codegen/printing/fcode.py
EmilyBourne: This should be fixed when issue #328 is fixed as a tuple will always be a 1d object.
What is the "wrong" shape?

pyccel/ast/numpyext.py
yguclu: @saidctb If you agree, please remove class `Bounds` from module `numpyext` and `_print_Bounds` from `fcode`. Also check if we have a `_visit_Bounds` somewhere else.

pyccel/parser/semantic.py
saidctb: yes it should be a  PythonTuple but I'm not sure yet on how to make it working

pyccel/parser/semantic.py
saidctb: I think it's coming from multiple places

saidctb:
> I think that these renamed files are no longer tested. Is that correct?
> 
> ```
> tests/codegen/fcode/scripts/ListComprehension.py → tests/codegen/fcode/scripts/bugs/ListComprehension.py
> tests/codegen/fcode/scripts/concatenation.py → tests/codegen/fcode/scripts/bugs/concatenation.py
> tests/codegen/fcode/scripts/cross.py → tests/codegen/fcode/scripts/bugs/cross.py
> tests/codegen/fcode/scripts/generic_methods.py → tests/codegen/fcode/scripts/bugs/generic_methods.py
> tests/codegen/fcode/scripts/header_interface.py → tests/codegen/fcode/scripts/bugs/header_interface.py
> tests/codegen/fcode/scripts/interfaces.py → tests/codegen/fcode/scripts/bugs/interfaces.py
> ```
> 
> Were they passing before? If so are they in bugs because something has been broken or because the fact that they were working a fluke rather than intentional?
> If we have broken something that was working well before then we should create an issue

they were never properly working

pyccel/ast/numpyext.py
saidctb: okay

tests/symbolic/test_symbolic.py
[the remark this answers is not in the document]
saidctb: they all have the same error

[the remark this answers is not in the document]
EmilyBourne: Very good. In that case they can all have the same xfail reason. e.g:
```
pytest.mark.xfail(reason="SemanticError undefined variable, see issue #330")
```

pyccel/parser/semantic.py
yguclu: @EmilyBourne Any ideas on this point?

EmilyBourne: Not without more details on why it doesn't work. I would have expected the only change necessary would be to add Interface to the types here: https://github.com/pyccel/pyccel/blob/master/pyccel/codegen/printing/fcode.py#L2229

pyccel/codegen/printing/fcode.py
yguclu: I think this is used in
```
from numpy import array

x = array(((1, 2, 3), (4, 5, 6)))
```
The argument of `array` is a 1D `PythonTuple`, but we want to convert it to a 2D array and get its shape.

EmilyBourne: Ah sorry I didn't realise I was looking at a sympy tuple here

yguclu: By the way, here we are looking at `_print_Tuple` but the same change was made to `_print_PythonTuple`.

EmilyBourne: No it wasn't
https://github.com/pyccel/pyccel/blob/devel-ast-nodes/pyccel/codegen/printing/fcode.py#L586

pyccel/parser/semantic.py
yguclu: @saidctb Do you mean that you could change the lines
```
d_var = self._infere_type(expr.args[0], **settings)
d_var['datatype'] = sp_dtype(expr)
```
to
```
d_var['datatype' ] = expr.dtype
```
and the code would still work correctly?

pyccel/codegen/printing/fcode.py
yguclu: @EmilyBourne You're right, sorry. We handle that in the `__init__` of `numpyext.Array` directly.

pyccel/parser/semantic.py
saidctb: yes I suggest to leave that for another pull request, when we deal with type inference method

yguclu: OK.

pyccel/parser/semantic.py
EmilyBourne: One of the main places where this is used is when evaluating an If statement. An If is made up of a test and a body these are not stored in test and body structures but in a tuple `(test,body)`

yguclu: @saidctb @EmilyBourne
Thanks to both of you for clarifying this point. I think we can close this conversation now.

pyccel/parser/semantic.py
yguclu: We'll improve this in the future.

tests/epyccel/test.py
```diff
@@ -0,0 +1,4 @@
+    @types('float32', 'float32')
+    def add_real(a, b):
+        c = a+b
+        return c
```
EmilyBourne, line 4: Is this file supposed to be here?

saidctb: no I thought I have deleted it

Issue #240, opened 2020-02-20
pyccel silently fails on errors


krystophny:
Currently pyccel fails without any error messages if an exception is not handled. This is due to `try` blocks without `except`, such as in `console.py:205`. For the following example pyccel doesn’t complain but also doesn’t work:

```python

import numpy as np
from pyccel.decorators import types

@types('real[:]', 'real[:]', 'real[:]')
def kern_sqexp(x0, x1, h):
    """Generic squared exponential kernel"""
    return h[0]*np.exp(-np.sum((x1 - x0))**2/(2.0*h[1]**2))

@types('real[:]', 'real[:]', 'real[:]')
def kern_wendland4(x0, x1, h):
    """Wendland kernel for dimension <= 3"""
    r = np.abs(np.sqrt(np.sum((x1 - x0))**2)/h[1])
    if r < 1.0:
        ret = h[0]*(1.0 - r**4)*(1.0 + 4.0*r)
    else:
        ret = 0.0
    return ret

@types('real[:]', 'real[:]', 'real[:]')
def kern_wendland4_product(x0, x1, h):
    """Wendland product kernel"""
    dx = np.abs(np.sqrt((x1 - x0)**2)/h[1])
    if np.all(dx < 1.0):
        ret = h[0]*np.prod((1.0 - dx**4)*(1.0 + 4.0*dx))
    else:
        ret = 0.0
    return h[0]*np.prod((1.0 - dx**4)*(1.0 + 4.0*dx))
```

krystophny:
Update: In that case `show_traceback=False` was set in `parser/base.py` per default. Switching it on shows an error message. In any case one should maybe exit with code nonzero and write the error somewhere / inform the user that something went wrong, even if one doesn’t want to print the full stack trace. 

A similar problem concerns `x.T` or `x.transpose()` which is just pyccelized as `x` without error or warning that this operation is not supported. This can lead to wrong results without the user knowing.

EmilyBourne:
> Update: In that case `show_traceback=False` was set in `parser/base.py` per default. Switching it on shows an error message. In any case one should maybe exit with code nonzero and write the error somewhere / inform the user that something went wrong, even if one doesn’t want to print the full stack trace.

`show_traceback` can now be set to true by using the option `--verbose`

EmilyBourne:
[...]

Issue #157, opened 2018-09-11
Unreadable Errors


EmilyBourne:
Many problems throw errors which are unreadable and do not help find the problem.

E.g. enumerate does not seem to be implemented it.
Using it throws the following error:

```
Traceback (most recent call last):
  File "/usr/local/bin/pyccel", line 11, in <module>
    load_entry_point('pyccel==0.1', 'console_scripts', 'pyccel')()
  File "/usr/local/lib/python3.5/dist-packages/pyccel/commands/console.py", line 244, in pyccel
    binary=binary)
  File "/usr/local/lib/python3.5/dist-packages/pyccel/codegen/utilities.py", line 179, in execute_pyccel
    code = codegen.doprint()
  File "/usr/local/lib/python3.5/dist-packages/pyccel/codegen/codegen.py", line 299, in doprint
    code = printer(self.expr)
  File "/usr/local/lib/python3.5/dist-packages/pyccel/codegen/printing/fcode.py", line 2121, in fcode
    return FCodePrinter(settings).doprint(expr, assign_to)
  File "/usr/local/lib/python3.5/dist-packages/pyccel/codegen/printing/codeprinter.py", line 60, in doprint
    lines = self._print(expr).splitlines()
  File "/home/emily/.local/lib/python3.5/site-packages/sympy/printing/printer.py", line 257, in _print
    return getattr(self, printmethod)(expr, *args, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/pyccel/codegen/printing/fcode.py", line 274, in _print_Program
    '{sep}\n').format(funcs=funcs, sep=sep, f=self._print(i))
  File "/home/emily/.local/lib/python3.5/site-packages/sympy/printing/printer.py", line 257, in _print
    return getattr(self, printmethod)(expr, *args, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/pyccel/codegen/printing/fcode.py", line 1216, in _print_FunctionDef
    body_code = '\n'.join(self._print(i) for i in body)
  File "/usr/local/lib/python3.5/dist-packages/pyccel/codegen/printing/fcode.py", line 1216, in <genexpr>
    body_code = '\n'.join(self._print(i) for i in body)
  File "/home/emily/.local/lib/python3.5/site-packages/sympy/printing/printer.py", line 257, in _print
    return getattr(self, printmethod)(expr, *args, **kwargs)
[...]

PR #324, opened 2020-05-20
Devel error testing

EmilyBourne:
Related to issue #157 : Unreadable Errors
- Remove some of the cases that lead to pyccel crashing out with large tracebacks and move towards pyccel exiting gracefully with clean error messages

Related to issue #240 : pyccel silently fails on errors
- As mentioned in the discussion here Pyccel should always stop at the end of the current stage if an error was encountered.

Ensure that tests/errors actually tests if there are errors


pyccel/parser/semantic.py
```diff
@@ -1526,6 +1534,9 @@ def _visit_PyccelAdd(self, expr, **settings):
         atoms = [a['is_pointer'] or a['is_target'] for a in atoms if a['rank']>0]
         args  = [self._visit(a, **settings) for a in expr.args]
 
+        if None in args:
+            args = [a if a is not None else Integer(0) for a in args]
+
```
saidctb, line 1539: in which case we get `None` ?

EmilyBourne: In the case where the code is wrong and variables have not been declared. See tests/errors/semantic for examples

EmilyBourne: I think this would also be the case if the variable was declared with an expression that pyccel does not understand

saidctb: then we shouldn't modify the code to ignore it ,it will crash at printing time maybe

EmilyBourne: It crashes at various places in the code when it tries to interact with it. It is one of the reasons why our error messages are so unintelligible. In order to improve our errors it is important that it exits gracefully rather than crashing out.

We discussed previously that errors during each stage shouldn't stop that stage completing in order to find as many of the errors associated with that stage as possible. If the code crashes when we try to check something such as the dtype then this is no longer true.

If a `None` is returned then an error is also added to Errors(). If there are errors then the execution stops. If the execution will be stopped then the result won't be generated so it doesn't matter if the output would have been wrong.

We are therefore asking the rest of Semantic to ignore it but as we have an associated Error we are not ignoring it completely

saidctb: there are some blocking errors that we have no choice but to stop the process and one of them is the type inference if we can't because of None, then we should stop, modifying the code is more dangerous because we may not detect it at all

EmilyBourne: Ok. So `get_variable` (line 310 of parser/semantic.py) should never return None.
It should populate Errors with a Fatal error and raise an error if it finds nothing?

saidctb: we only continue if it's not a blocking error and we report it  so there is no risk because we are reporting it, we also need to always think about the performance, adding conditions will make the parsing slower only if it's necessary then we have no choice

saidctb: `get_variable` is different because we can cases like this one  https://github.com/pyccel/pyccel/blob/master/pyccel/parser/semantic.py#L1396.
if we don't have that kind of cases then sure we can raise an error there

EmilyBourne: Currently none of the errors in pyccel are blocking due to this line (parser/base.py:257):
```
        # flag for blocking errors. if True, an error with this flag will cause
        # Pyccel to stop
        # TODO ERROR must be passed to the Parser __init__ as argument

        self._blocking = False
```

It seems strange to me that blocking or not depends on the parser rather than the error.

I would have expected fatal errors to be blocking and normal errors to be non-blocking

saidctb: yes maybe then we should make them blocking

EmilyBourne:
Errors in parser/syntax.py and parser/semantic.py should be handled with the Errors system rather than raises. However I think this pull request is big enough already so I will create an issue for that and do it in a separate pull request

EmilyBourne:
These corrections have unearthed a test that was failing without an error due to the use of `SystemExit(0)` ensuring that the exit code of pyccel was 0.

EmilyBourne:
Fixes issue #240

Issue #331, opened 2020-06-03
Codacy-coverage not covering subprocesses


EmilyBourne:
Codacy coverage indicates that `pyccel/commands/console.py` has 0% coverage. This means that all tests in the folder `tests/pyccel/` are not currently included in the coverage calculation. The file 
`pyccel_cov.pth` must be being created in the wrong location.

EmilyBourne:
The `pyccel_cov.pth` file is correctly called, however the `.coverage.X.Y` file does not seem to be created for the subroutines. I am unable to reproduce this behaviour locally in a virtualenv

PR #332, opened 2020-06-03
WIP: Investigate issue #331

EmilyBourne:
Fixes issue #331

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ before_script:  # Coverage instructions - setup
   - sudo apt-get install default-jre
   - INSTALL_DIR=$(${PYTHON} -c "import pyccel; print(pyccel.__path__[0])")
   - SITE_DIR=$(${PYTHON} -c 'import sysconfig; print(sysconfig.get_paths()["purelib"])')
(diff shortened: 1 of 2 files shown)


EmilyBourne:
The problem has now been located!
The use of chdir confuses codacy and leads to coverage results being outputted in the wrong folder. Either the output folder must be specified in .coveragerc or the `coverage combine` command must somehow be carried out recursively

EmilyBourne:
Coverage files are now all outputted to the same folder however commands/console.py still has 0% coverage despite there being a large number of coverage output files

EmilyBourne:
The tests are being run from the root directory. This means that they are importing pyccel from the same folder. However when `pyccel` is run it uses the installed version of pyccel. Thus the files that are reported are in a different directory. The installation directory should be the directory that is being examined (this directory seems to be correctly defined in ${SITE_DIR} and incorrectly in ${INSTALL_DIR}). The fact that all epyccel tests are using the local version of pyccel (defined in ${INSTALL_DIR} ) is masking this error.

To fix the problem we must change directory before running tests
- [ ] Check if this is sufficient to ensure that the installed version of pyccel is used

We must also specify the installation directory correctly when creating the .coveragerc file
- [ ] Test

PR #341, opened 2020-06-09
Fix issue #331

EmilyBourne:
Run tests from tests/ directory. This ensures that it is the installed version of pyccel that is used rather than the local version
Ensure that all .coverage files are written to the same directory


EmilyBourne:
I have activated the analysis of this branch in codacy. I can confirm that `pyccel/commands/console.py` has more than 0% coverage.
This therefore definitely fixes issue #331.

.travis.yml
```diff
@@ -67,8 +69,10 @@ jobs:
 #      - brew install gcc
       - brew install openblas lapack
       - brew install open-mpi
-    before_script: null
-    after_script : null
+    before_script:
+      - cd tests
+    after_script :
+      - cd ..
```
yguclu, line 75: In order to keep things tidy and avoid repetitions, could we move these instructions to the `script` section?

Issue #335, opened 2020-06-08
Order should be a member of PyccelAstNode


EmilyBourne:
The `order` property is added to PyccelAstNode in PR #342, but it still need to be properly calculated in the following classes:

- [x] PythonTuple
- [ ] Len
- [x] Array (order is initialised but in Basic not the PyccelAstNode property)
- [ ] Rand
- [x] Full (order is initialised but in Basic not the PyccelAstNode property)
- [x] PyccelOperator ( and PyccelDiv )
- [x] PyccelBooleanOperator
- [x] PyccelAssociativeParenthesis
- [x] PyccelUnary
- [x] DottedVariable
- [x] FunctionCall
- [x] IndexedVariable
- [x] IndexedElement

This will require changes in different parts of the code, and we will also need to add more tests to make sure that everything works.

yguclu:
The first step should be to raise an error when an expression contains objects with different `order`. Later on we will perform transpose operations where needed.

PR #342, opened 2020-06-17
Clean type inference

saidctb:
In this PR we cleaned up the `infere_type` method in `Semantics`  where we use the `PyccelAstNode` attributes to get the necessary type informations,  now all the type calculation are done locally in each object  and the `infere_type` method only extracts  the type informations without any calculations


tests/codegen/fcode/scripts/numpyext.py
```diff
@@ -47,8 +47,8 @@
 print(np_sum(x17)==n*m*2)
 
 x18 = zeros((3, 3))
-x19 = diag(x18)
-x20 = diag(x18)
+#x19 = diag(x18)
+#x20 = diag(x18)
```
yguclu, line 51: Was this code working before? Did we support `diag` at all?

By the way, we should comment out the import statement for `diag` as well.

saidctb: it was never tested so I'm not sure

pyccel/ast/numpyext.py
```diff
@@ -1208,15 +1200,29 @@ def __init__(self, x):
         self._dtype     = NativeReal()
         self._precision = default_precision[str_dtype(self._dtype)]
 
+class NumpyMod(NumpyUfuncBinary):
+    def __init__(self, x1, x2):
+        self._shape     = x1.shape
+        self._rank      = x1.rank
+        self._dtype     = x1.dtype
+        self._precision = x1.precision
+        self._order     = x1.order
```
EmilyBourne, line 1209: Could we create a function in BasicAstNode to copy these basic properties in order to avoid code duplication?

pyccel/parser/semantic.py
```diff
@@ -682,154 +678,54 @@ def _infere_type(self, expr, **settings):
...
+                               NumpyMin, NumpyMax, PythonInt, NumpyMod,
+                               PythonFloat, PythonComplex,
+                               NumpyInt, Int32, Int64,
+                               NumpyFloat, Float32, Float64,
+                               NumpyComplex, Complex64, Complex128,
+                               Real, Imag, PythonBool, Norm,
+                               NumpyUfuncBase, MathFunctionBase, Expr, FunctionCall,
+                               Integer, Float, Complex, String,
+                               BooleanTrue, BooleanFalse,
+                               PyccelArraySize, Is, IndexedElement)):
```
EmilyBourne, line 718: Can these be grouped? E.g. `isinstance(expr, PyccelAstNode)` ?

pyccel/ast/basic.py
```diff
@@ -21,6 +21,7 @@ class PyccelAstNode:
     _rank      = None
     _dtype     = None
     _precision = None
+    _order     = None
```
EmilyBourne, line 24: I don't think the order has been initialised in all the classes which inherit from PyccelAstNode

yguclu: @EmilyBourne For some classes `_rank` and `_order` have no meaning, so there is no reason to give them a value other than the default `None`. Are there classes where we should have initialized them, but we did not?

EmilyBourne: The ones I have found where I think it does need to be set are:
- [ ] PythonTuple
- [ ] Len
- [ ] Array (order is initialised but in Basic not the PyccelAstNode property)
- [ ] Rand
- [ ] Full (order is initialised but in Basic not the PyccelAstNode property)
- [ ] PyccelOperator ( and PyccelDiv )
- [ ] PyccelBooleanOperator
- [ ] PyccelAssociativeParenthesis
- [ ] PyccelUnary
- [ ] DottedVariable
- [ ] FunctionCall
- [ ] IndexedVariable
- [ ] IndexedElement

pyccel/parser/semantic.py
saidctb: In theory yes, but we need more verification to be sure

pyccel/ast/basic.py
saidctb: in this PR `_order` is not calculated it's just created for the moment and all the instance that needs it have methods that overrides it.

pyccel/parser/semantic.py
EmilyBourne: What do we need for verification? Can we not just try and see if any of the tests fail?

pyccel/ast/numpyext.py
saidctb: done

pyccel/parser/semantic.py
yguclu: I agree, it's worth to give it a try

saidctb: the  main problem with that is almost all our nodes inherit from it which means that we can never go to the else condition where we raise an error

pyccel/ast/basic.py
yguclu: Although I like small pull requests, in this case I think that the calculation of `_order` belongs to this PR, as it would be confusing to keep it separate.

pyccel/ast/numpyext.py
```diff
@@ -1202,27 +1202,15 @@ def __init__(self, x):
 
 class NumpyMod(NumpyUfuncBinary):
     def __init__(self, x1, x2):
-        self._shape     = x1.shape
-        self._rank      = x1.rank
-        self._dtype     = x1.dtype
-        self._precision = x1.precision
-        self._order     = x1.order
+        self.copy(x1)
```
yguclu: We need to be careful here:
```
5 % np.arange(1, 5)
```
should return `array([0, 1, 2, 1])`, so copying the attributes of `5` will give the wrong result.

pyccel/ast/basic.py
```diff
@@ -42,3 +42,11 @@ def precision(self):
     @property
     def order(self):
         return self._order
+
+    def copy(self, x):
```
yguclu, line 46: Since the name `copy` is used in many Python libraries (including Numpy) to identify functions and methods that create a new instance, I recommend to change the name of this method, e.g. to `copy_attributes`.

pyccel/parser/semantic.py
yguclu: @saidctb What do you mean? I can see no `else` condition here, and no error raised.

saidctb: there is an else condition in the end of the method where we raise an error

yguclu: Sorry, you meant this of course:
```
        else:
            msg = '{expr} not yet available'.format(expr=type(expr))
            raise NotImplementedError(msg)
```

EmilyBourne: So you are worried about keys other than:
- dtype
- precision
- rank
- shape
- order
not being correctly set for classes that could be added later?
(I assume no classes are reaching the `else` condition at the moment as the tests are passing)

pyccel/ast/basic.py
saidctb: adding it will require more changes in different part of the code , we will also need to add more tests to make sure that everything works which is not the intent of this PR

pyccel/ast/basic.py
saidctb: okay

pyccel/parser/semantic.py
saidctb: yes

pyccel/ast/basic.py
yguclu: OK, I have updated the description of issue #335.

pyccel/parser/semantic.py
saidctb: yes @EmilyBourne  we shouldn't go to else unless there is an error somewhere which means it's a bug and we don't test bugs , so if we add the `PyccelAstNode`  it is the same as deleting the else condition because we are always gonna pass by the condition `isinstance(expr, PyccelAstNode)`

pyccel/ast/numpyext.py
saidctb: ah yes you are right, what I did was wrong

pyccel/parser/semantic.py
yguclu: OK let's keep this as it is for now.

EmilyBourne: We would go to `else` if we are accidentally looking at a sympy object or if the class has accidentally been created without inheriting from `PyccelAstNode`.
Effectively this list is there just so the error can act as a reminder to check if new classes are allocatable, etc.
It feels wrong to have 11 lines instead of 1, effectively throwing away the advantage of `PyccelAstNode`, but I can't think of a better way of doing this and it is important to ensure that these values are set when necessary

saidctb: okay I will try to make it work

pyccel/ast/numpyext.py
saidctb: done

pyccel/parser/semantic.py
saidctb: done

pyccel/ast/numpyext.py
```diff
@@ -1208,15 +1223,47 @@ def __init__(self, x):
...
+                    shape = broadcast(shape, a.shape)
 
+            self._shape = shape
+            self._rank  = len(shape)
+        else:
+            self._rank = max(a.rank for a in args)
+
+class NumpyMin(NumpyUfuncUnary):
+    def __init__(self, x):
+        self.copy_attributes(x)
```
EmilyBourne, line 1262: Doesn't this mean that `np.min(x)` has the same shape and rank as x?

pyccel/ast/numpyext.py
```diff
@@ -1201,8 +1224,38 @@ def __init__(self, x):
         self._precision = default_precision[str_dtype(self._dtype)]
 
 class NumpyMod(NumpyUfuncBinary):
-    def __init__(self, x1, x2):
-        self.copy_attributes(x1)
+    def __init__(self, *args):
```
yguclu, line 1227: I think we should immediately raise an error if we pass the wrong number of arguments, so I prefer the original signature. The rest of the `__init__` function could remain the same, just write
```
def __init__(self, x1, x2):
    args = (x1, x2)
    ...
```
as you did for class `Matmul`

pyccel/ast/numpyext.py
```diff
@@ -1201,8 +1224,38 @@ def __init__(self, x):
...
+            else:
+                shape = broadcast(args[0].shape, args[1].shape)
+
+                for a in args[2:]:
+                    shape = broadcast(shape, a.shape)
+
+            self._shape = shape
+            self._rank  = len(shape)
+        else:
+            self._rank = max(a.rank for a in args)
```
yguclu: This code looks sufficiently general to me, that we should be able to move it to the `__init__` of class `NumpyUfuncBinary`.

pyccel/parser/semantic.py
```diff
@@ -631,9 +623,31 @@ def _infere_type(self, expr, **settings):
 
             return d_var
 
-        elif isinstance(expr, IndexedVariable):
+        elif isinstance(expr, Dlist):
```
yguclu, line 626: What is a `Dlist`?

EmilyBourne: A Dlist is a list or a tuple multiplied by an integer. E.g:
`(1,False,4)*3`
`[1,2,3]*3`

pyccel/ast/numpyext.py
saidctb: it depends on the function no ?, for example in NumpyMod we don't take complex arguments

pyccel/ast/numpyext.py
saidctb: yes you are right

pyccel/ast/numpyext.py
saidctb: that's correct

pyccel/parser/semantic.py
yguclu: What about the concatenation of lists or tuples, do we have a special class for that too?

pyccel/ast/numpyext.py
EmilyBourne: Which numpy function does this represent?
`http://pageperso.lif.univ-mrs.fr/~francois.denis/IAAM1/numpy-html-1.14.0/reference/generated/numpy.amin.html#numpy.amin` ? (wrong name)
`https://numpy.org/doc/stable/reference/generated/numpy.minimum.html` (wrong number of arguments)
`http://pageperso.lif.univ-mrs.fr/~francois.denis/IAAM1/numpy-html-1.14.0/reference/generated/numpy.ndarray.min.html` (pyccel doesn't handle `a.min()` syntax

pyccel/ast/numpyext.py
yguclu: I think we can consider the complex case and let the Python interpreter raise an error if that is not allowed...

pyccel/ast/numpyext.py
saidctb: you can do `from numpy import min`

EmilyBourne: Ok, that's what I assumed. You are right it seems to be equivalent to amin. It's odd that I can't find documentation for it though

pyccel/ast/numpyext.py
saidctb: you have arctan2 for example can take two integers and return a real number, contrary to mod where it returns an integer if he gets integers and a real if he gets reals

EmilyBourne: We could make this a function where `complexes`, `reals`, etc are initialized unless an argument is provided in which case `complexes = []`. Then a second argument for the minimum output precision.
e.g. int+int would normally give int but with minimum precision real it gives real

It might make for a slightly complicated function but at least we won't have to write out all 30 lines for each special case

yguclu: @EmilyBourne This is a good suggestion. In `NumpyUfuncBinary` we could add a static method:
```
@staticmethod
def _initialize(obj, arg1, arg2, allow_complex=False, basic_dtype='int'):
    ...
```
which gets called by the various `__init__` functions with the proper arguments. But how many binary functions do we have?

pyccel/parser/semantic.py
EmilyBourne: The concatenation of lists or tuples appears to be broken

yguclu: @EmilyBourne Thanks for creating a new issue about it

pyccel/ast/numpyext.py
yguclu: As @saidctb pointed out in a private conversation, most "magic" binary functions need broadcasting in Numpy, e.g. `__add__`, `__mul__`, `__div__`, etc. This will be better addressed in a separate pull request.

PR #345, opened 2020-06-18
Split module program

EmilyBourne:
A file containing functions and statements will be broken into a module containing the functions and a program containing the statements. This allows the pyccelised functions to be imported from another file.

This pull request supersedes the `devel-create-extra-module` branch which has similar functionality but was incomplete and was branched from a much earlier version of master making it hard to merge.

Commits:
- Add type to hold module and program syntactic information
- Split syntactic results into module and program information
- Create a second parser if a module has been spawned from a program
- Loop over all created parsers (module and program)
- Ensure headers are saved to module
- Don't quote class name
- Variables cannot be imported
- Save module parser for imports
- Correctly identify exe file
- Comments are not program specific
- Provide verbose parameter
- Avoid double parsing. Save d_parsers to additional module after their treatment
- Ensure correct shared module name
- Rename files to avoid issue 344. Leave one copy to illustrate issue 344
- Correct n+ -1 to n-1
- Correct types decorator. Remove xfail flag
- Parse sons. This ensures they are added to sons. It must be done after namespace initialisation
- Pass lib arguments to f2py


pyccel/parser/syntactic.py
```diff
@@ -238,7 +240,60 @@ def _visit(self, stmt):
...
+            if mod_code is None:
+                mod_code = CodeBlock(mod+start)
+        assert( mod_code is not None or prog_code is not None)
+        from pyccel.ast import ParserResult
+        code = ParserResult(program   = prog_code,
+                            module    = mod_code,
+                            prog_name = prog_name,
+                            mod_name  = current_mod_name)
         code.set_fst(stmt)
         return code
```
yguclu, line 298: Can you explain the use of these additional 50 lines of code?

EmilyBourne: They break the syntax ast into 2 CodeBlocks. One for the module and one for the Program. FunctionDefs and ClassDefs always go into modules, Headers and Comments go in the same CodeBlock as the following object, EmptyLines go in the same CodeBlock as the previous line. Imports go in both CodeBlocks (this means that there can be unnecessary imports), everything else goes in the program

tests/codegen/fcode/scripts/imports.py
```diff
@@ -1,8 +1,9 @@
...
-xi = ai + 1
-
-from expressions import ad, bd
-xd = ad * bd
+#from expressions import ai
+#xi = ai + 1
+#
+#from expressions import ad, bd
+#xd = ad * bd
+xi = 2
```
yguclu, line 6: Why commenting those lines out?

pyccel/parser/syntactic.py
yguclu: Thanks for the explanation. Could you copy and paste it as a comment in the source code?

tests/codegen/fcode/scripts/imports.py
EmilyBourne: I did not put the declarations of the variables into the module. Handling variables requires an extra step where the main is called as an `init` function. The variables would also need to be saved as members of the module so that the module takes care of the declaration. This is possible but would require more work. If we want to be able to import variables then I suggest we leave this for a second PR.

yguclu:
@EmilyBourne This branch is ready to be squashed and merged. If possible, I would like to avoid the automatic commit message generated by Git. So I suggest you explicitly collect and reorder this branch's commit messages in a list to be placed in the PR's description. (Unnecessary details can be omitted.) I will then use that as the commit message.

tests/codegen/fcode/scripts/imports.py
yguclu: This is not an urgent matter, we can discuss about it later.

EmilyBourne:
@yguclu I have created the list of commit messages

yguclu:
Thanks, that looks like a good commit message!

By the way, I will drop these lines:

> This pull request supersedes the devel-create-extra-module branch which has similar functionality but was incomplete and was branched from a much earlier version of master making it hard to merge.

PR #350, opened 2020-06-24
Raise neat errors and stop pyccel failing without errors

### LINKED ISSUE #157: Unreadable Errors

### LINKED ISSUE #213: If decorators are commented then pyccel fails without an error
If there are comments amongst the decorators then pyccel fails, however no error is shown and pyccel gives no indication that it has not succeeded. E.g:

```
from pyccel.decorators import pure, external, types

@pure
@external
@types('double[:,:]','double[:]','double[:]')
#@types('int[:,:]','int[:]','int[:]')
def mat_mult(A,x,b):
    for i, l in enumerate(b):
        b[i] = 0
        for j, xi in enumerate(x):
            b[i] += A[i,j] * xi
```

### LINKED ISSUE #240: pyccel silently fails on errors

### LINKED ISSUE #307: Raise an error if an allocation function is used in an expression
An allocation expression is for example:
- full
- zeros
- ones
- rand
- empty

It is important for this to be only used in an assign to avoid unexpected temporary variables which cannot be marked as stack_array

E.g:
```
a = np.rand(2,2)*0.5 + 3.0
```
should raise an error but it can be re-written as:
```
a = np.rand(2,2)
a = a*0.5 + 3.0
```
or
```
a = np.empty(2,2)
for i in range(2):
    for j in range(2):
         a[i,j] = np.rand()*0.5 + 3.0
```

### LINKED ISSUE #327: parser/syntactic.py and parser/semantic.py should use errors.raise instead of raise
This should fix issue #157 . Once this is done all unreadable errors should be due to errors in pyccel and all errors due to input code recognised and handled by pyccel should return neat errors

### LINKED ISSUE #351: Pyccel does not raise an error if an unknown/obsolete function is imported from pyccel.decorators
This comes from the handling of pyccel.decorators in the syntactic step.
These lines: https://github.com/pyccel/pyccel/blob/master/pyccel/parser/syntactic.py#L498-L499

        if is_ignored_module(source):
            return EmptyLine()
handle the case of pyccel.decorators. This is a different method to the one used for most imports, e.g. numpy, scipy.

For these imports, an Import object is created however they are not added to the namespace. They are then handled in the function builtin_import : https://github.com/pyccel/pyccel/blob/master/pyccel/ast/utilities.py#L176

If we use the second method to handle the line from pyccel.decorators import external then we can use the inspect module to raise an error (or a warning) if the target is not defined in decorators.py

EmilyBourne:
- Remove raise statements indicating incorrect input. The only raise statements should be due to incorrect code (e.g. wrong argument type passed to a function). Use the error handling system instead (fixes #327 )
- Remove all try / except anything statements. Only catch expected errors
- Never use raise SystemExit(0). Ensure exit code shows error (fixes #240)
- catch PyccelErrors in codegen/pipeline.py to handle fatal errors
- check whether there are any errors in codegen/pipeline.py to halt translation if an error is encountered
- catch NotImplementedErrors in codegen/pipeline.py and link this to an error with a neat message (fixes #157 )
- ensure all error handling is completed in codegen/pipeline.py
- Create new class NumpyNewArray to allow error to be raised when an allocation function is used in an expression (fixes #307)
- Add error tests
- Ignore comments in decorators (fixes #213)
- Fix bug in broadcast function
- Provide access to imports of the semantic parser
- Collect PyccelOperators to catch raise thrown by broadcast without code duplication
- Move errors and associated messages to new folder to allow importing from anywhere in pyccel
- Treat imports from pyccel.decorators like all other imports (fixes #351)
- Rename StopIteration -> PyccelStopIteration in stdlib/parallel/openmp.py and stdlib/parallel/openacc.py


EmilyBourne:
Remaining try/except clauses without exception type specified:
ast/core.py:L716
```
    try:
        name = 'Dummy_' + str(abs(hash(expr)
                                  + np.random.randint(500)))[-4:]
    except:
        name = 'Dymmy_' + str(abs(np.random.randint(500)))[-4:]
```

ast/core.py:L4974
```
        try:
            var = expr.lhs
            symbols = []
            if isinstance(var, DottedVariable):
                var = expr.lhs.lhs
                while isinstance(var, DottedVariable):
                    var = var.lhs
                symbols.append(var)
            elif isinstance(var, IndexedElement):
                var = var.base
                symbols.append(var)
            elif isinstance(var, Variable):
                if var.rank:
                    symbols.append(var)
            return symbols
        except:
            #TODO should we keep the try/except clause ?

            # TODO must raise an Exception here
            #      this occurs only when parsing lapack.pyh
            raise ValueError('Unable to extract assigned variable')
```
Are these try/except clauses necessary? If so what exceptions are we expecting?

EmilyBourne:
Remaining raise statements to be handled:
stdlib/parallel/openmp.py:            raise 'StopIteration'
stdlib/parallel/openacc.py:            raise 'StopIteration'
```
class Range(object):
    def __next__(self):
        if (self.i < self.stop):
            i = self.i
            self.i = self.i + 1
        else:
            raise 'StopIteration'
```

This seems like a strange construction. Should it be changed to `raise OutOfRangeError`? Or should it be handled differently? I don't know where this function is called

saidctb:
> Remaining try/except clauses without exception type specified:
> ast/core.py:L716
> 
> ```
>     try:
>         name = 'Dummy_' + str(abs(hash(expr)
>                                   + np.random.randint(500)))[-4:]
>     except:
>         name = 'Dymmy_' + str(abs(np.random.randint(500)))[-4:]
> ```
> 
> ast/core.py:L4974
> 
> ```
>         try:
>             var = expr.lhs
>             symbols = []
>             if isinstance(var, DottedVariable):
>                 var = expr.lhs.lhs
>                 while isinstance(var, DottedVariable):
>                     var = var.lhs
>                 symbols.append(var)
>             elif isinstance(var, IndexedElement):
>                 var = var.base
>                 symbols.append(var)
>             elif isinstance(var, Variable):
>                 if var.rank:
>                     symbols.append(var)
>             return symbols
>         except:
>             #TODO should we keep the try/except clause ?
> 
>             # TODO must raise an Exception here
>             #      this occurs only when parsing lapack.pyh
>             raise ValueError('Unable to extract assigned variable')
> ```
> 
> Are these try/except clauses necessary? If so what exceptions are we expecting?

they are not necessary in my opinion

EmilyBourne:
> Remaining raise statements to be handled:
> stdlib/parallel/openmp.py: raise 'StopIteration'
> stdlib/parallel/openacc.py: raise 'StopIteration'
> 
> ```
> class Range(object):
>     def __next__(self):
>         if (self.i < self.stop):
>             i = self.i
>             self.i = self.i + 1
>         else:
>             raise 'StopIteration'
> ```
> 
> This seems like a strange construction. Should it be changed to `raise OutOfRangeError`? Or should it be handled differently? I don't know where this function is called

After discussion with @saidctb . These classes seem to exist in order to implement a `ParallelRange` however this object already exists in `ast/core.py`

The hope is that the user can do something like: `from pyccel import prange`

They do not seem to be used currently

pyccel/errors/messages.py
```diff
@@ -56,6 +56,13 @@
 INDEXED_TUPLE = 'Tuples must be indexed with constant integers for the type inference to work'
 LIST_OF_TUPLES = 'Cannot create list of non-homogeneous tuples'
 
+UNDEFINED_LAMBDA_VARIABLE = 'Unknown variables in lambda function'
```
yguclu, line 59: Did you mean "variable" (singular)?

pyccel/errors/messages.py
```diff
@@ -56,6 +56,13 @@
...
+UNDEFINED_LAMBDA_FUNCTION = 'Unknown function in lambda function'
+
+UNDEFINED_INTERFACE_FUNCTION = 'Interface functions not found'
```
yguclu, line 62: Did you mean "function" (singular)?

pyccel/errors/messages.py
```diff
@@ -56,6 +56,13 @@
...
+
+UNDEFINED_WITH_ACCES = 'The __enter__ or __exit__ method for the with construct cannot be found'
```
yguclu, line 64: Typo? ACCES >> ACCESS

pyccel/errors/messages.py
```diff
@@ -69,6 +76,11 @@
 PYCCEL_RESTRICTION_IMPORT_IN_DEF = 'Only From Import is allowed inside a def statement'
 PYCCEL_RESTRICTION_IMPORT_STAR = 'import * not allowed'
 PYCCEL_RESTRICTION_OPTIONAL_NONE = 'Variables cannot be equal to None unless they are optional arguments and None is the default value'
+PYCCEL_RESTRICTION_UNSUPPORTED_SYNTAX = 'Pyccel has encountered syntax that it does not recognise'
+PYCCEL_RESTRICTION_TODO = "Pyccel has encountered syntax that has not been implemented yet. Please create an issue at https://github.com/pyccel/pyccel/issues and provide a small example of your problem. Do not forget to specify your target language"
```
yguclu, line 80: Nice!

pyccel/errors/errors.py
```diff
@@ -304,7 +304,7 @@ def __str__(self):
         return text
 
 if __name__ == '__main__':
-    from pyccel.parser.messages import NO_RETURN_VALUE_EXPECTED
+    from pyccel.errors.messages import NO_RETURN_VALUE_EXPECTED
     from pyccel.parser.messages import INCOMPATIBLE_RETURN_VALUE_TYPE
     from pyccel.parser.messages import UNDEFINED_VARIABLE
```
yguclu, line 309: Should we get all error messaged from `pyccel.errors` now?

pyccel/codegen/printing/fcode.py
```diff
@@ -2536,6 +2546,14 @@ def _print_FunctionCall(self, expr):
 
 #=======================================================================================
 
+    def _print_Application(self, expr):
+        if isinstance(expr, PyccelAllocatableObject):
+            errors.report(FORTRAN_ALLOCATABLE_IN_EXPRESSION,
+                          symbol=expr, severity='fatal')
```
yguclu, line 2552: Can you please explain when this error is raised?

pyccel/ast/numpyext.py
```diff
@@ -90,12 +90,15 @@
     'pi': Constant('real', 'pi', value=numpy.pi),
 }
 
+class PyccelAllocatableObject(PyccelAstNode):
+    pass
```
yguclu, line 94: What does this class represent? Could you add a docstring to it?

By the way, it is a little strange to use a name containing "Object" for a class...

pyccel/ast/datatypes.py
```diff
@@ -275,12 +267,9 @@ def get_default_value(dtype):
 # TODO improve and remove try/except
```
yguclu, line 267: Can you remove this comment now?

pyccel/ast/core.py
```diff
@@ -182,7 +184,7 @@ def broadcast(shape_1, shape_2):
     b = len(shape_2)
     if a>b:
         new_shape_2 = (1,)*(a-b) + tuple(shape_2)
-        new_shape_1 = shape_2
+        new_shape_1 = shape_1
```
yguclu, line 187: Was this a bug? How did you catch it? Do we have a test for this now?

pyccel/ast/core.py
```diff
@@ -196,12 +198,12 @@ def broadcast(shape_1, shape_2):
             new_shape.append(e1)
         elif e1 == 1:
             new_shape.append(e2)
-        elif e2 == 2:
+        elif e2 == 1:
```
yguclu, line 201: Same comment as above

pyccel/ast/core.py
saidctb: yes it's a bug , the broadcasting is never tested 
because we don't allow for the moment  something like  `y = x + 1` where `x` is an ndarray
maybe we can think of a case where we need to check the shape

pyccel/errors/errors.py
EmilyBourne: Good spot. Not sure how I managed to only correct 1 of 3 imports

pyccel/ast/datatypes.py
EmilyBourne: The `TODO remove try/except` can be removed. I don't know about the  `improve` part. The comment was written by @ratnania  in 2018 and the function doesn't seem to have changed since

pyccel/codegen/printing/fcode.py
EmilyBourne: This is related to issue #307. It is raised when trying to translate code such as:
```
a = np.ones(4)*5
```

pyccel/ast/numpyext.py
EmilyBourne: This is related to issue #307. The class represents objects which are allocatable. These objects then inherit from this. This includes `Full`, `Linspace`, `Diag`, `Where`, etc
I'm not sure what exactly to put in a docstring given that my explanation of `AllocatableObject` is that it is an object that is allocatable ^^'.

I can change the name if you prefer. Should it just be called `PyccelAllocatable` ?

pyccel/errors/messages.py
EmilyBourne: It can be one or more

pyccel/errors/messages.py
EmilyBourne: It can be one or more

pyccel/ast/core.py
EmilyBourne: I think we discussed the broadcast function in a different merge. My understanding was that there would be a pull request at some point to make it cover more cases and to test it properly. But I don't see an issue for this so maybe I misremembered.
I caught this bug with a very small test I wrote while checking that a different error was raised correctly. Fixing this bug made the one below apparent thanks to our current tests

pyccel/ast/numpyext.py
yguclu: I see. Maybe more explicitly, I would say in the docstring that this class represents Python constructs corresponding to a memory allocation on the heap (unless the `@stack_array` decorator is used), and which we only accept as the right-hand side of an assignment.

Since this is only relevant to arrays at the moment, maybe this class should be called `PyccelArray` and be moved to a different module. And maybe we need to distinguish between a `PyccelStackArray` and a `PyccelHeapArray`, and return the correct one in the `__new__` constructor.

Sorry for all the "maybe"! @saidctb What do you think?

pyccel/ast/datatypes.py
yguclu: OK let's keep the "improve" then :smile:

pyccel/ast/core.py
yguclu: Do we have a unit test that would fail if we change this code back to what it was?
I think that's important.

pyccel/ast/numpyext.py
yguclu: Actually, for this PR it should be enough to change the name of this class to `NumpyNewArray` and leave it where it is. We can discuss later about further generalizations.

pyccel/ast/core.py
EmilyBourne: I have just added one

EmilyBourne:
```
# TODO raise StopIteration() instead of 'StopIteration'

#$ header class StopIteration(public, hide)
#$ header method __init__(StopIteration)
#$ header method __del__(StopIteration)
class StopIteration(object):

    def __init__(self):
        pass

    def __del__(self):
        pass
```

Codacy's error concerns this class which seems to override the standard `StopIteration`
Is this class useful?

EmilyBourne:
No errors have been raised by the `create_variable` function

yguclu:
I don't know why we should have a class that overrides the standard `StopIteration`. I assume that was not used so far?! If we want to keep it for later reference, I suggest we change its name to `PyccelStopIteration` and add a comment.

By the way, Codacy has found that the same function `make_symbol` is defined twice in file `pyccel/errors/errors.py`, at [lines 33-34](https://github.com/pyccel/pyccel/blob/e0fab7d389808a09edc07fe4e6daf7aadf6db329/pyccel/errors/errors.py#L33) and [lines 43-44](https://github.com/pyccel/pyccel/blob/e0fab7d389808a09edc07fe4e6daf7aadf6db329/pyccel/errors/errors.py#L43). In particular the latter definition overrides the one at [lines 23-24](https://github.com/pyccel/pyccel/blob/e0fab7d389808a09edc07fe4e6daf7aadf6db329/pyccel/errors/errors.py#L23) for colored output.

EmilyBourne:
I have renamed the class. It wasn't used so far. The only mention of it was in the comment above the class.

I have fixed the redefinition of `make_symbol`

PR #357, opened 2020-07-06
Replace red baron with python ast

EmilyBourne:
In this PR redbaron the python parser is removed and  replaced with the  built-in Python ast parser.
The Python ast does not parse comments , so we extend the parsed ast by inserting the comments in it .
this list of changes have also been made:
- remove the imports preprocessing and move that to the semantic's stage.
- change the `change_priority`  function  to only put parenthesis in the right place because the python ast already has the right ordering.


pyccel/parser/utilities.py
```diff
@@ -91,7 +93,7 @@ def accelerator_statement(stmt, accel):
     """
     assert(accel in ['omp', 'acc'])
 
-    if not isinstance(stmt, CommentNode): None
+    if not isinstance(stmt, CommentLine): None
```
EmilyBourne, line 96: These statements do nothing. Are they supposed to be:
`if not isinstance(stmt, CommentLine): `return None?

saidctb: they should return something

EmilyBourne:
When working on this branch the method used is:
- run the tests
- pick a failing test
- fix the test
- repeat
This method will allow us to identify the functions that haven't been touched and see if they need removing or if we need to add a test for that syntax

Currently methods that still need to be looked at include:
- with (tests/codegen/fcode/scripts/context.py, tests/preprocess/scripts/omp.py)
- ternary operator  (is this tested?)
- lambdas              (is this still tested?)
- tuple unpacking  (tests/epyccel/modules/tuples.py)
- generators           (is this tested?)
- continue              (not tested?)
- comprehension   (?)
- DotProxyList <-> ast.Attribute   (To discuss)

EmilyBourne:
Python ast does not preserve parentheses. The importance of the operators is stored in the ordering. We must therefore reinsert the parentheses. This can be done in the `change_priority` function

pyccel/codegen/printing/pycode.py
```diff
@@ -78,7 +78,8 @@ def _print_FunctionDef(self, expr):
 
         decorators = expr.decorators
         if decorators:
-            for n,args in decorators.items():
```
yguclu, line 81: It looks like the original code was buggy. Was it working before?

EmilyBourne: Yes, it was. Redbaron has a decorator object which meant we could treat it separately, but python ast stores decorators as function calls, hence the need to change the way we manage this loop

pyccel/ast/utilities.py
```diff
@@ -172,69 +172,57 @@ def builtin_function(expr, args=None):
 
     return None
 
+
 # TODO add documentation
-builtin_import_registery = ('numpy', 'numpy.linalg', 'numpy.random', 'scipy.constants', 'itertools', 'math', 'pyccel.decorators')
+builtin_import_registery = {'numpy': {**numpy_functions, **numpy_constants, 'linalg':numpy_linalg_functions, 'random':numpy_random_functions},
+        'numpy.linalg': numpy_linalg_functions,
+        'numpy.random': numpy_random_functions,
```
yguclu, line 179: Aren't these, in practice, a repetition of the keys below?
```
builtin_import_registery['numpy']['linalg']
builtin_import_registery['numpy']['random']
```

pyccel/parser/syntactic.py
```diff
@@ -229,24 +194,41 @@ def _visit(self, stmt):
         cls = type(stmt)
         syntax_method = '_visit_' + cls.__name__
         if hasattr(self, syntax_method):
-            return getattr(self, syntax_method)(stmt)
+            self._scope.append(stmt)
+            result = getattr(self, syntax_method)(stmt)
+            self._scope.pop()
+            return result
```
yguclu, line 200: Why is it necessary to append `stmt` to `self._scope` in order to visit `stmt`?

pyccel/parser/syntactic.py
```diff
@@ -385,191 +340,256 @@ def _visit_NoneType(self, stmt):
...
+        return stmt
 
-    def _visit_StringNode(self, stmt):
-        val =  stmt.value
-        if isinstance(stmt.parent,(RedBaron, DefNode)):
+    def _visit_Str(self, stmt):
+        val =  stmt.s
+        if isinstance(self._scope[-2], ast.Expr):
             return CommentBlock(val)
         return String(val)
```
yguclu, line 349: Quoting the python ast documentation (https://docs.python.org/dev/library/ast.html#node-classes)

> Changed in version 3.8: Class ast.Constant is now used for all constants.

> Deprecated since version 3.8: Old classes ast.Num, ast.Str, ast.Bytes, ast.NameConstant and ast.Ellipsis are still available, but they will be removed in future Python releases. In the meantime, instantiating them will return an instance of a different class.

So it looks like we should use `ast.Constant` and then check its value. What is our strategy for supporting multiple Python 3 versions in the future?

yguclu: Sorry, I have just realized that indeed we have a `_visit_Constant` function starting at line 419. Well done!

pyccel/parser/syntactic.py
```diff
@@ -203,21 +182,7 @@ def parse(self, verbose=False):
...
-
-            return List(*ls, sympify=False)
-        elif isinstance(stmt, (tuple, TupleNode)):
-            return PythonTuple(*ls)
-        else:
-            return Tuple(*ls, sympify=False)
+        return [self._visit(i) for i in stmt]
 
     def _visit(self, stmt):
         """Creates AST from FST."""
```
yguclu, line 188: Should this docstring read "Create Sympy AST from Python FST?"

pyccel/parser/syntactic.py
```diff
@@ -385,191 +340,256 @@ def _visit_NoneType(self, stmt):
...
-        return repr(stmt)
+        return stmt
 
-    def _visit_StringNode(self, stmt):
-        val =  stmt.value
-        if isinstance(stmt.parent,(RedBaron, DefNode)):
+    def _visit_Str(self, stmt):
+        val =  stmt.s
+        if isinstance(self._scope[-2], ast.Expr):
             return CommentBlock(val)
```
saidctb, line 348: why don't we use `get_docstring` function from the `ast` module ?
see here https://docs.python.org/3.5/library/ast.html#ast.get_docstring

pyccel/ast/utilities.py
EmilyBourne: Not quite. This handles the syntax
```
import numpy

numpy.linalg.func
```

Whereas the standard method handles 
```
import numpy.linalg as mod 
```

Maybe we could handle this differently to avoid repetition

pyccel/parser/syntactic.py
EmilyBourne: Sometimes it is useful to know the parent or grandparent of the object being visited. This information is not available from the python ast. This was the easiest way to collect it

pyccel/parser/extend_tree.py
```diff
@@ -0,0 +1,245 @@
...
+        super().__init__()
+        self.s = s
+        self.lineno     = lineno
+        self.col_offset = col_offset
+
+class CommentMultiLine(CommentLine):
+    """"New AST node representing a multi-line comment"""
+
+
+def get_comments(code):
```
yguclu, line 26: Can we have a proper docstring with `Parameters` and `Returns`?

Same comment for the other functions in this module...

yguclu:
I wonder if it could be beneficial to perform the syntactic stage exclusively at the Python AST level, by extending [ast.AST](https://docs.python.org/dev/library/ast.html#ast.AST) nodes and maybe using a subclass of [ast.NodeTransformer](https://docs.python.org/dev/library/ast.html#ast.NodeTransformer) to perform the necessary changes.

A tree of `PyccelASTNodes` would then be our intermediate representation, and it could be introduced at the semantic stage.

pyccel/ast/utilities.py
yguclu: I see. This is the fine distinction between (1) importing a sub-module and (2) getting its attributes later. The repetition is very little for now, but this may need to be improved in the future.

pyccel/parser/syntactic.py
yguclu: I am a little surprised that we need the parent or the grandparent during the syntactic stage. Could you give me an example?

EmilyBourne: Eg  a string in an expression is a comment block but a string in an assignment is a string. I believe there's one other example but I won't be able to check until Thursday

pyccel/parser/syntactic.py
EmilyBourne: I didn't realise this function existed. Does it still work if there is more than one comment block? (Ie if a docstring is used as a comment block rather than as an actual doc string?

pyccel/parser/syntactic.py
yguclu: I see, thanks!

pyccel/parser/syntactic.py
saidctb: it won't work , I think this way is better

pyccel/parser/extend_tree.py
yguclu: Anyway, we should merge this important PR now and leave minor improvements for later.

saidctb:
> I wonder if it could be beneficial to perform the syntactic stage exclusively at the Python AST level, by extending [ast.AST](https://docs.python.org/dev/library/ast.html#ast.AST) nodes and maybe using a subclass of [ast.NodeTransformer](https://docs.python.org/dev/library/ast.html#ast.NodeTransformer) to perform the necessary changes.
> 
> A tree of `PyccelASTNodes` would then be our intermediate representation, and it could be introduced at the semantic stage.

so you want us to create new Nodes that extends `ast.AST`  and use the `ast.NodeTransformer `  to create our new Ast ?

yguclu:
> > I wonder if it could be beneficial to perform the syntactic stage exclusively at the Python AST level, by extending [ast.AST](https://docs.python.org/dev/library/ast.html#ast.AST) nodes and maybe using a subclass of [ast.NodeTransformer](https://docs.python.org/dev/library/ast.html#ast.NodeTransformer) to perform the necessary changes.
> > A tree of `PyccelASTNodes` would then be our intermediate representation, and it could be introduced at the semantic stage.
> 
> so you want us to create new Nodes that extends `ast.AST` and use the `ast.NodeTransformer ` to create our new Ast ?

Not quite, I was thinking of using `ast.AST` nodes at the syntactic stage (we only need to create a few classes for annotations, pragmas, etc) and then creating a new tree of `PyccelAstNodes` in a single pass during the semantic stage (which would stay the same as now). But we'll discuss about it later.

yguclu:
This PR is able to merge, as soon as a suitable commit message is provided in its description.

pyccel/parser/syntactic.py
EmilyBourne: I don't know. Maybe 'create pyccel ast from python ast'?

I'm not clear on the difference between ast and fst.
I'm not sure how many of the results are still stored in sympy objects. I think it might be a mixture of pyccel and sympy ast still

Issue #360, opened 2020-07-13
Handle functions with multiple return statements


yguclu:
When a function contains multiple return statements, Pyccel only works if all of these are in the form `return x` and `x` is the same variable. Hence this example works
```
@types('int')
def non_negative(i):
    if i < 0:
        x = False
        return x
    else:
        x = True
        return x
```
but we would like to rewrite it as
```
@types('int')
def non_negative(i):
    if i < 0:
        return False
    else:
        return True
```
which raises an error at the annotation (semantic) stage.

This issue is due to the fact that Pyccel generates a result variable with a random name, which is different for each return statement. Instead it should generate this variable only once in the function, so that its name is unique.

PR #361, opened 2020-07-14
fix issue 360

saidctb:
Fix issue #360 (handle function with multiple return statements).
The following changes were made:
- create new return variable if returned expressions are not variables or have different names
- handle the recursivity of a function in `FunctionCall` instead of `get_function`


tests/epyccel/test_epyccel_functions.py
```diff
@@ -191,6 +191,31 @@ def f10(x):
...
+    def ackermann(m, n):
+        if m == 0:
+            return n + 1
+        elif n == 0:
+            return ackermann(m - 1, 1)
+        else:
+            return ackermann(m - 1, ackermann(m, n - 1))
+
+    f = epyccel(ackermann)
+    assert abs(f(2,3)-ackermann(2,3))<1e-16
```
yguclu, line 205: The two results should be exactly identical, as they are integer numbers...

EmilyBourne:
This looks like it works and simplifies what we have.
You might want to add a test where 2 different variables are returned. Eg 
```
@types('int,int')
def get_min(a, b)
    if a<b:
        return a
   else:
       return b
```
The tests so far only manage expressions.

Is this not a fortran problem though rather than a pyccel problem?
I thought we wanted to keep fortran modifications in thé fortran codegen? Or is this OK as it will still work in C?

pyccel/parser/syntactic.py
```diff
@@ -770,6 +757,16 @@ def _visit_FunctionDef(self, stmt):
         if 'private' in decorators.keys():
             is_private = True
 
+        returns = [i.expr for i in _atomic(body, cls=Return)]
+        assert all(len(i) == len(returns[0]) for i in returns)
+        returns = list(zip(*returns))
+        results = []
+        for i in returns:
+            if not all(i[0]==j for j in i) or not isinstance(i[0], Symbol):
```
yguclu, line 765: I am not sure I understand what this loop is doing, after you redefined `returns` as a list of tuples. Is `i` a tuple that contains all the expressions that have the same position in the return statements?

By they way, there is no need to redefine `returns` on line 762 (which is confusing), instead you can simply write
```
for i in zip(*returns):
    ...
```

saidctb:
> This looks like it works and simplifies what we have.
> You might want to add a test where 2 different variables are returned. Eg
> 
> ```
> @types('int,int')
> def get_min(a, b)
>     if a<b:
>         return a
>    else:
>        return b
> ```
> 
> The tests so far only manage expressions.
> 
> Is this not a fortran problem though rather than a pyccel problem?
> I thought we wanted to keep fortran modifications in thé fortran codegen? Or is this OK as it will still work in C?

 yes it's a Fortran problem but we can also recover the original statement if we want to print a high level code like Python or Julia

pyccel/parser/syntactic.py
saidctb: the answer is yes

yguclu:
I also agree with @EmilyBourne on the fact that we should add a few more unit tests.

tests/epyccel/test_epyccel_functions.py
saidctb: fixed

saidctb:
> Very well done! I have experimented a bit with this branch and it seems to work very nicely. I have found a couple of issues, though.
> 
> I don't know it this is related to this PR, but the following function cannot be pyccelized because of an error at the shared library generation stage:
> 
> ```
> @types('int', 'int', 'bool')
> def choose_first(x, y, verbose=False):
>     if verbose:
>         print('Choose x')
>         return x
>     else:
>         return x
> ```
> 
> Moreover, it appears that the generated Fortran subroutine is declared as `recursive`, but this could be a more general issue.

ah no , it's actually related to this issue , this must be fixed

yguclu:
To simplify things further, let's consider the Python function
```
@types('int', 'int')
def first(x, y):
    return x
``` 
Pyccel translates it to the Fortran procedure
```
recursive  subroutine first(x, y) 

implicit none
integer(kind=8), intent(inout)  :: x 
integer(kind=8), value  :: y

return
end subroutine
```
which is wrong on three accounts:
1. It should not be `recursive`;
2. It should be a function, not a subroutine;
3. There should be no `intent(inout)` arguments.

Instead, the correct Fortran code is
```
function first(x, y) result(r)

implicit none
integer(kind=8), value :: x 
integer(kind=8), value :: y
integer(kind=8) :: r

r = x
return
end function
```
Is it possible to fix at least some of these problems within this pull request?

saidctb:
> To simplify things further, let's consider the Python function
> 
> ```
> @types('int', 'int')
> def first(x, y):
>     return x
> ```
> 
> Pyccel translates it to the Fortran procedure
> 
> ```
> recursive  subroutine first(x, y) 
> 
> implicit none
> integer(kind=8), intent(inout)  :: x 
> integer(kind=8), value  :: y
> 
> return
> end subroutine
> ```
> 
> which is wrong on three accounts:
> 
>     1. It should not be `recursive`;
> 
>     2. It should be a function, not a subroutine;
> 
>     3. There should be no `intent(inout)` arguments.
> 
> 
> Instead, the correct Fortran code is
> 
> ```
> function first(x, y) result(r)
> 
> implicit none
> integer(kind=8), value :: x 
> integer(kind=8), value :: y
> integer(kind=8) :: r
> 
> r = x
> return
> end function
> ```
> 
> Is it possible to fix at least some of these problems within this pull request?

yes it should be possible to fix them all

yguclu:
We just need a commit message, as usual...

PR #362, opened 2020-07-15
Speed up Travis CI on macOS

yguclu:


yguclu:
I had to remove all commits, and commit again all changes one by one in order to see the timings on Travis.

Issue #363, opened 2020-07-16
Allocation with a random shape


saidctb:
this example illustrate the bug.
```
def f(): 
    import numpy as np 
    a = np.zeros(np.random.randint(23)) 
    print(a) 
```

EmilyBourne:
When testing this code I receive the following error:
```
ERROR at annotation (semantic) stage
pyccel: [semantic] 
 |fatal: [3,4]| Could not find randint in imported module np.random (np.random.randint)
```

Indeed, although rand and random are available `randint` does not seem to be implemented yet. The problem is therefore unrelated to the allocation

saidctb:
> When testing this code I receive the following error:
> 
> ```
> ERROR at annotation (semantic) stage
> pyccel: [semantic] 
>  |fatal: [3,4]| Could not find randint in imported module np.random (np.random.randint)
> ```
> 
> Indeed, although rand and random are available `randint` does not seem to be implemented yet. The problem is therefore unrelated to the allocation

I didn't check if `randint` was there , but I know that the shape stored will be wrong because it will be a call to the `randint` function which will give a different number each time

EmilyBourne:
I have added `randint`. The example that you provided works however if you try to use the shape of the object then you are right we do run into trouble.

From the following code:
```
def f():
    import numpy as np
    a = np.zeros(np.random.randint(23))
    b = np.shape(a)
    c = np.zeros_like(a)
```

The following fortran code is generated:
```
module test

implicit none




contains

!........................................
subroutine f()

  implicit none
  real(kind=8), allocatable  :: a (:)
  integer(kind=8)  :: b (0:1_8-1)
  real(kind=8), allocatable  :: c (:)
  real(kind=8)  :: Dummy_4258
  real(kind=8)  :: Dummy_4338
  real(kind=8)  :: Dummy_4273


  call random_number(Dummy_4258)

  allocate(a(0:floor(23_8 * Dummy_4258, kind=8) - 1_8))
  a = 0.0
  call random_number(Dummy_4338)

  b = [floor(23_8 * Dummy_4338, kind=8)]
  call random_number(Dummy_4273)

  allocate(c(0:floor(23_8 * Dummy_4273, kind=8) - 1_8))
  c = 0.0
end subroutine
!........................................

end module
```

The random expression is therefore reused. When the shape of an array is not an integer we should use NumpyArrayShape to avoid calling functions multiple times

EmilyBourne:
[...]

PR #364, opened 2020-07-16
Add numpy randint

EmilyBourne:
Add numpy's randint function and ensure that the shape of arrays initialised with one of numpy's X_like functions is correct

- Add NumpyRandInt class
- Add tests for NumpyRandInt
- Add error test to ensure neat error is correctly raised when generating an array with randint in fortran
- Use the same process_dtype method for all numpy functions where dtype can be specified
- Move PyccelArraySize to ast/core.py by removing Array special case. Initialise Variable shape with PyccelArraySize when it is unknown or non-constant. Fixes issue #363
- Add tests for issue #363
- Create 'alloc_shape' property of Variable to declare stack arrays correctly (shape is known by the initialisation function (e.g. Zeros) otherwise)


pyccel/ast/numpyext.py
```diff
@@ -76,8 +80,8 @@
     'Ones',
     'OnesLike',
     'Product',
-    'PyccelArraySize',
     'Rand',
+    'RandInt',
```
yguclu, line 84: I think at some point we will want that every class in the `numpyext` module has a name starting with "Numpy". So it may be best to call this new class `NumpyRandint`.

yguclu: Sorry for being pedantic.. can we have it with no capital "I"?

pyccel/codegen/printing/fcode.py
```diff
@@ -730,7 +730,25 @@ def _print_MathFloor(self, expr):
 
     def _print_NumpyFloor(self, expr):
         result_code = self._print_MathFloor(expr)
-        return 'real({})'.format(result_code)
+        if isinstance(expr.dtype, NativeInteger):
```
saidctb, line 733: numpy floor is always a real no ?, and also it doesn't take a complex

yguclu: Like every universal Numpy function, it takes a `dtype` optional argument: https://numpy.org/doc/stable/reference/ufuncs.html#optional-keyword-arguments

EmilyBourne: https://numpy.org/doc/stable/reference/generated/numpy.floor.html

No. Most of numpy's functions have an optional `dtype` kwarg. If this is not specified then numpy floor will return a real. However if it is specified then it will return whatever dtype is given. This dtype is unrelated to the arguments. Hence:
```
> import numpy as np
> np.floor(3.5, dtype=complex)
```
should return a complex(3.0, 0). It looks like there's a problem with this at the moment though https://stackoverflow.com/questions/47838306/getting-no-loop-matching-the-specified-signature-and-casting-error
I can't replicate the documented behaviour on my computer.

My reason for implementing this was simply so that I can get a randint using numpy floor to get the int, rather than using math floor which does not work for arrays. As it was it casts the result to a real and I would then need to cast back to an int

saidctb: you mean that we can do something like `np.floor(2.3, dtype='int')` ?

saidctb: In numpy1.18 it gives me an error

EmilyBourne: Sorry, you're right my link wasn't much use
http://numpy-discussion.10968.n7.nabble.com/floor-with-dtype-td44734.html
It seems that the dtype refers to the input and the output argument for some reason.

Can you think of another way to represent this for an array without double casting? Or do we need to have 2 implementations of floor, one for numpy's floor and one for randint?

saidctb: yes maybe we can use a floor for randint  alone

yguclu: So here is the situation:

> dtype
> >    New in version 1.6.
> >    Overrides the dtype of the calculation and output arrays. Similar to signature.

> signature
> >   Either a data-type, a tuple of data-types, or a special signature string indicating the input and output types of a ufunc. This argument allows you to provide a specific signature for the 1-d loop to use in the underlying calculation. If the loop specified does not exist for the ufunc, then a TypeError is raised. Normally, a suitable loop is found automatically by comparing the input types with what is available and searching for a loop with data-types to which all inputs can be cast safely. This keyword argument lets you bypass that search and choose a particular loop. A list of available signatures is provided by the types attribute of the ufunc object. For backwards compatibility this argument can also be provided as sig, although the long form is preferred. Note that this should not be confused with the generalized ufunc signature that is stored in the signature attribute of the of the ufunc object.

So `dtype` and `signature` allow to change the type of the input and output arrays, if the required type is actually allowed.

pyccel/ast/core.py
```diff
@@ -2367,6 +2368,10 @@ def process_shape(self, shape):
     def name(self):
         return self._name
 
+    @property
+    def alloc_shape(self):
+        return self._alloc_shape
+
```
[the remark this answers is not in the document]
EmilyBourne: It is the shape used for the allocation/declaration. If we take the following example:
```
import numpy as np
from pyccel.decorators import stack_array, types

@stack_array('a')
@types('int')
def f(n):
    a = np.ones(n)
    n = 10
    b = np.ones_like(a)
```
Then the shape of `a` needs to be saved as `(shape(a,0),)` as the expression describing it's shape is not constant. However as `a` is a stack array it's declaration needs to contain information about `n`

Similarly for the following example:
```
import numpy as np

def g():
    import numpy as np
    c = np.zeros(np.random.randint(23))
    d = np.zeros_like(c)
```
The allocation shape of `c` is the randint expression however the shape of `c` which is used whenever we manipulate it is `(shape(c,0),)`

In the second case the allocation shape of `c` is not actually necessary as the `Zeros` object has shape `(np.random.randint(23)` which is used for the allocation.

Maybe I should call `alloc_shape` -> `declare_shape` instead? It is only necessary for stack array objects whose shape is non-constant. This is important notably for all the f2py files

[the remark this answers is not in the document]
yguclu: So in your first example we have the following?
```
a.alloc_shape = (n, )
a.shape = (shape(a, 0), )

b.alloc_shape = (shape(a, 0), )
b.shape = (shape(b, 0), )
```
And you need `alloc_shape` for declaring the automatic Fortran arrays?

I am a little worried because such a declaration will be moved in the Fortran code to the declarations' section, so how can we ensure that the code is correct?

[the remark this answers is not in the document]
EmilyBourne: Yes exactly.
`alloc_shape` is needed for declaring stack arrays. In the case of a stack array (all f2py arrays) the declaration has the following form:
```
subroutine f(n)

  implicit none
  integer(kind=8), value  :: n
  real(kind=8)  :: a (0:n-1)
```
Thus the declaration (which has no knowledge of the expression used to initialise `a`) needs to have the information about `n`.

I agree that this is dangerous. I am not sure that we could declare `b` as a stack array by writing it this way. However the previous way led to b having the correct size only if it was a stack array.
I tried to avoid needing to create the `alloc_shape` variable but `stack_array`s were the one case where I couldn't work out how to.
The only other thing that I can think of is to create a `ConstantVariable` if a Variable is used to declare the size of a stack_array. This way we could be sure that it does not change afterwards and then we can feel free to use this in the shape.

Maybe we should discuss this this afternoon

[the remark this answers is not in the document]
yguclu: Sounds good. Further improvements do not belong to this PR anyway.

yguclu:
@EmilyBourne Could you add a line about `alloc_shape` in the PR description?

Issue #352, opened 2020-06-26
Assignment of an array from an expression does not raise an error and is not allocated


EmilyBourne:
The following code
```
from numpy import ones

a = ones((2,2))
b = ones((2,2))
c = ones((2,2))
d = ones((2,2))

x = a + b *2 - c / d
```

compiles without an error and generates the following:
```

program prog_test

implicit none

real(kind=8), allocatable  :: a (:,:)
real(kind=8), allocatable  :: b (:,:)
real(kind=8), allocatable  :: d (:,:)
real(kind=8), allocatable  :: c (:,:)
real(kind=8), allocatable  :: x (:,:)



allocate(a(0:2_8 - 1_8, 0:2_8 - 1_8))
a = 1.0
allocate(b(0:2_8 - 1_8, 0:2_8 - 1_8))
b = 1.0
allocate(c(0:2_8 - 1_8, 0:2_8 - 1_8))
c = 1.0
allocate(d(0:2_8 - 1_8, 0:2_8 - 1_8))
d = 1.0


x = a + b * 2_8 - c / d 



end program prog_test
```
however x is not allocated here

yguclu:
The main problem here is that, because `x` is automatically allocated on assignment, it does not use zero-based indexing.

EmilyBourne:
The provided example doesn't raise an error because the shape of x is known (2,2)

EmilyBourne:
Is there any way to unit test if x has been allocated?

PR #366, opened 2020-07-17
Fixes to PR 364 with failing tests + assorted issues

EmilyBourne:
- Provide the allocated size to function arguments. Keep the same size if the size is a PyccelArraySize
- An unknown shape is a PyccelArraySize or None. Add this logic to PyccelBooleanOperator too
- Add tests that would have been failing before corrections
- Add error test to ensure that error is raised when an array is created from an expression with unknown size (see issue #352)
- Add tests for issue #369
- If an array is created from an expression and the size is known then allocate the memory with an empty expression. Fixes issue #352
- Use the process_dtype function to find Array's dtype and precision. Fixes issue #367 
- Check for Boolean type for List. This mimmicks what is done in the Tuple class. Fixes issue #368
- Fix issue #369. When casting a boolean to a boolean change the precision. When casting something else to a boolean use a merge statement

### DIFF
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -2266,6 +2266,7 @@ class SemanticParser(BasicParser):
             if arguments:
                 for (a, ah) in zip(arguments, m.arguments):
                     d_var = self._infere_type(ah, **settings)
+                    d_var['shape'] = ah.alloc_shape
                     dtype = d_var.pop('datatype')
 
                     # this is needed for the static case
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -1392,7 +1392,7 @@ class SemanticParser(BasicParser):
                         severity='fatal', blocker=self.blocking)
                 elif lhs.rank>0 and isinstance(rhs, PyccelOperator):
                     #TODO: Provide order once issue #335 is fixed
-                    #stmts.append(Assign(lhs, Empty(lhs.alloc_shape, dtype, lhs.order)))
+                    #new_expressions.append(Assign(lhs, Empty(lhs.alloc_shape, dtype, lhs.order)))
                     new_expressions.append(Assign(lhs, Empty(lhs.alloc_shape, dtype, 'C')))
             else:
 
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -61,6 +61,7 @@ from pyccel.ast.core import PyccelAnd, PyccelOr,  PyccelNot, PyccelAssociativePa
 from pyccel.ast.core import PyccelUnary
 from pyccel.ast.core import Product, FunctionCall
 from pyccel.ast.core import PyccelArraySize
+from pyccel.ast.core import PyccelOperator
 
 from pyccel.ast.functionalexpr import FunctionalSum, FunctionalMax, FunctionalMin
 from pyccel.ast.functionalexpr import GeneratorComprehension as GC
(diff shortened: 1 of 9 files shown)


EmilyBourne:
Should we fix issue #335 before merging this?

pyccel/parser/semantic.py
```diff
@@ -1381,14 +1382,18 @@ def _assign_lhs_variable(self, lhs, d_var, rhs, **settings):
...
                 if not know_lhs_shape:
                     msg = "Cannot infer shape of right-hand side for expression {} = {}".format(lhs, rhs)
                     errors.report(PYCCEL_RESTRICTION_TODO+'\n'+msg,
                         bounding_box=(self._current_fst_node.lineno, self._current_fst_node.col_offset),
                         severity='fatal', blocker=self.blocking)
-
+                elif lhs.rank>0 and isinstance(rhs, PyccelOperator):
+                    #TODO: Provide order once issue #335 is fixed
+                    #stmts.append(Assign(lhs, Empty(lhs.alloc_shape, dtype, lhs.order)))
+                    new_expressions.append(Assign(lhs, Empty(lhs.alloc_shape, dtype, 'C')))
```
EmilyBourne, line 1396: Should we fix issue #335 before merging this PR?

yguclu: I don't know how long it would take to fix that issue. This PR looks good as it stands

EmilyBourne: I would rather leave it for another PR. I was just worried about potentially introducing a bug here

pyccel/parser/semantic.py
```diff
@@ -1381,14 +1382,18 @@ def _assign_lhs_variable(self, lhs, d_var, rhs, **settings):
...
                         or isinstance(rhs, (Variable, EmptyLike, DottedVariable))
                 if not know_lhs_shape:
                     msg = "Cannot infer shape of right-hand side for expression {} = {}".format(lhs, rhs)
                     errors.report(PYCCEL_RESTRICTION_TODO+'\n'+msg,
                         bounding_box=(self._current_fst_node.lineno, self._current_fst_node.col_offset),
                         severity='fatal', blocker=self.blocking)
-
+                elif lhs.rank>0 and isinstance(rhs, PyccelOperator):
+                    #TODO: Provide order once issue #335 is fixed
+                    #stmts.append(Assign(lhs, Empty(lhs.alloc_shape, dtype, lhs.order)))
```
yguclu, line 1395: I know this is just a comment, but nevertheless... shouldn't `stmts` be changed to `new_expressions`?

EmilyBourne: You're right. I had stmts as a local variable at one point. I must have missed this when I changed it to new_expressions

tests/epyccel/modules/arrays.py
```diff
@@ -364,6 +364,71 @@ def array_real_2d_F_initialization(a):
...
+    from numpy import full
+    z = full((2,3),5,order='F')
+    x[:] = (x // y) * x + z
+
+@types( 'int32[:]', 'int32[:]', 'int32[:]' )
+def array_int32_in_bool_out_1d_complex_3d_expr( x, y, ri ):
+    from numpy import full, int32, empty
+    z = full(3,5, dtype=int32)
+    r = empty(3, dtype=bool)
+    r = (x // y) * x > z
```
yguclu, line 413: Is this the expression you intended, or maybe you meant to write `r[:] = ...`?

EmilyBourne: You are right. It should be the same as in the functions below it

Issue #349, opened 2020-06-24
Tuple/List concatenation does not work


EmilyBourne:
The following code:
```
a = (1,2)
b = (3,4)
c = a + b
```
is translated to:
```
program prog_test

implicit none

integer(kind=8)  :: a (0:2_8-1)
integer(kind=8), allocatable  :: c (:)
integer(kind=8)  :: b (0:2_8-1)
a = [1_8, 2_8]
b = [3_8, 4_8]
c = a + b

end program prog_test
```

PR #371, opened 2020-07-24
Remove sympy objects in FunctionalFor. Create known_bugs folder. Fix string concatenation

### LINKED ISSUE #358: FunctionalFor will not accept a constant range
The following compiles without a problem:
```
m = 10
x = [i for i in range(m)]
```

But the following crashes:
```
x = [i for i in range(10)]
```
The error is:
```
  File "/home/emily/.local/lib/python3.6/site-packages/pyccel/parser/semantic.py", line 2036, in _visit_FunctionalFor
    d_var['shape'] = PythonTuple(*shape)
  File "/home/emily/.local/lib/python3.6/site-packages/pyccel/ast/builtins.py", line 199, in __init__
    integers  = [a for a in args if a.dtype is NativeInteger()]
  File "/home/emily/.local/lib/python3.6/site-packages/pyccel/ast/builtins.py", line 199, in <listcomp>
    integers  = [a for a in args if a.dtype is NativeInteger()]
AttributeError: 'Integer' object has no attribute 'dtype'
```

EmilyBourne:
Remove sympy objects in FunctionalFor to fix issue #358 . Create known_bugs folder in tests/errors to ensure that neat errors are raised for tests that we know are failing. In addition this allows working parts of these tests to still be tested which helps prevent further regressions

- Ensure that integers are PyccelIntegers and that sympy integers are not created in _visit_FunctionalFor (fixes #358)
- Add DottedVariable to isinstance(var, Variable). Raise an error for multi-level functional for
- Do not remove bugs from tests but instead check that they generate a neat error. This allows the working parts of these tests to still be tested
- Handle completely unknown variables
- Add tests for strings
- Move string concatenation to a subset of PyccelAdd (this ensures that rank/shape/etc is correctly managed)
- Create a TupleVariable rather than a Variable for a List. Handle Concatenation in PyccelAdd. Raise error for now as the Concatenate class does not handle shape/rank/dtype properties (This si the first step towards fixing #349)


pyccel/ast/core.py
```diff
@@ -237,9 +237,13 @@ def __init__(self, *args):
...
                 raise TypeError('cannot determine the type of {}'.format(self))
-            
+
+            ranks  = [a.rank for a in args]
             shapes = [a.shape for a in args]
 
+            if None in ranks:
+                self._rank  = None
+                self._shape = None
             if all(not (sh is None or isinstance(sh, PyccelArraySize)) for tup in shapes for sh in tup):
```
yguclu, line 247: It appears that `tup` could be `None` here, in which case we get the error
```
TypeError: 'NoneType' object is not iterable
```

EmilyBourne: Yes. When `tup` is `None` then so is `ranks`. I missed the `if`->`elif` on the line below when rebasing

Issue #227, opened 2020-01-22
Create module to handle naming


EmilyBourne:
We should have a module to handle names in pyccel. Currently there are multiple places where things are (or aren’t but should be) done to avoid name problems in fortran. It would be clearer if this was managed in one place. Possible problems include

- Case sensitivity:
    python is case sensitive but fortran isn't.
    x and X should be translated differently in fortran. What name should be used?
- Underscores:
    python names can start with underscores or even be underscores but fortran names can't.
    How should these variables be renamed?
- Name collisions:
    Sometimes names are generated with a random string to avoid name collisions.
    Name collisions could occur in different circumstances:
      - multiple functions epyccelized which have the same name
      - epyccelizing modules with the same name (ie. import a and import b.a)
      - renaming variables to manage "import module" syntax
- Fortran keywords
    Check that the names used in python are not fortran or c keywords
- Python keywords
    Builtin functions can be overridden in python but this is inadvisable. Therefore pyccel does not handle this information. An error should be raised if the user does this
- Lengthy names
    Names are limited to 96 characters in fortran. An error should be raised if the user has longer names

EmilyBourne:
If functions are renamed to avoid case sensitivity name collisions then the correct name can be used in a c wrapper if this is used instead of f2py

yguclu:
> If functions are renamed to avoid case sensitivity name collisions then the correct name can be used in a c wrapper if this is used instead of f2py (see issue #370)

Do you mean that after moving away from f2py we could print to Fortran the original Python names, which may contain capital letters, whenever we don't have name clashes?

EmilyBourne:
> Do you mean that after moving away from f2py we could print to Fortran the original Python names, which may contain capital letters, whenever we don't have name clashes?
No. We can already do that. Let me provide an example. The following python code:
```
def littleTest():
    return 2
```
is translated to:
```
module test

implicit none

contains

!........................................
function littleTest() result(Out_0001)

  implicit none

  integer(kind=8) :: Out_0001

  Out_0001 = 2_8
  return

end function littleTest
!........................................

end module test
```
However as fortran is not case sensitive to call this function in C, you have to call `littletest()` in c. If we write our own wrapper this would be written:
```
static PyObject * littleTest_wrapper(PyObject *self, PyObject *args)
{

    long Out_0001;
    if (!PyArg_ParseTuple(args, "" ))
        return NULL;
    Out_0001 = littletest(); // call to fortran function
[...]

PR #375, opened 2020-07-30
Fix issue #374 : Using the syntax 'import mod' can lead to name collisions

### LINKED ISSUE #374: Using the syntax 'import mod' can lead to name collisions
The following examples do not compile due to name collisions:
```
import module_a as a
value = a.value()
```
`a.value` is imported as `value`. There are therefore 2 objects called `value`

```
import module_a as a
import module_b as b
c = a.fun() + b.fun()
print(c)
```
both `a.fun` and `b.fun` are imported as `fun`. There are therefore 2 objects called `fun`

EmilyBourne:
Fix #374 by checking if the name is already used and if so creating an alias. Add missing math constants to available imports

- Add a function to the `Import` class to check if the module object has already been imported via mod.func() syntax
- If the function has not been imported then check for name collisions when saving the import via mod.func()
- Add tests for issue #374
- Add missing math constants to dict
- Correct importing constants from a module
- Reduce code duplication in name handling functions
- Create create_random_string function to handle all randomstring requirements without creating a symbol


tests/pyccel/test_pyccel.py
```diff
@@ -399,6 +399,7 @@ def test_hope_benchmarks( test_file ):
                                         "scripts/import_syntax/import_mod_func.py",
                                         "scripts/import_syntax/import_mod_as_func.py",
                                         "scripts/import_syntax/collisions.py"
```
yguclu, line 401: Missing comma

tests/pyccel/test_pyccel.py
```diff
@@ -398,7 +398,8 @@ def test_hope_benchmarks( test_file ):
                                         "scripts/import_syntax/from_mod_import_as_func.py",
                                         "scripts/import_syntax/import_mod_func.py",
                                         "scripts/import_syntax/import_mod_as_func.py",
-                                        "scripts/import_syntax/collisions.py"
+                                        "scripts/import_syntax/collisions.py",
+                                        "scripts/import_syntax/collisions3.py"
```
yguclu, line 402: Correct name of file is **collision3.py**, singular, without the `s`.

pyccel/parser/semantic.py
```diff
@@ -253,6 +254,14 @@ def annotate(self, **settings):
 
         return ast
 
+    def _get_new_name(self, current_name):
+        while current_name in self._possible_names:
+            # Generate random name. Remove dummy
+            new_suffix = create_variable(current_name).name[5:]
```
yguclu, line 260: If all we are doing here is to create a random suffix, maybe we should create a new function just for that, which would be used here and by `create_variable`

EmilyBourne: I agree. There are also several occasions in the code where `create_variable` is used to get a name but the Symbol that it returns is rejected. I think we should create a module to handle all naming problems as discussed in issue #227 . This would be the obvious place to put this function. I thought it was better to do a pull request just for name problems which would clear up these issues. Would you rather we take care of it here?

yguclu: Solving issue #227 requires more thinking, but here we could maybe take a first step and separate the creation of a new dummy variable from the generation of a random suffix, just for clarity. @saidctb What do you think?

EmilyBourne: Actually it's not as bad as I thought. Grep pulls up just 3 cases that need correcting

- codegen/printing/fcode.py:2338:                var_name = create_variable(r).name
- codegen/printing/fcode.py:2524:                var_name = create_variable(r).name
- parser/semantic.py:260:            new_suffix = create_variable(current_name).name[5:]

pyccel/parser/semantic.py
```diff
@@ -257,7 +257,7 @@ def annotate(self, **settings):
     def _get_new_name(self, current_name):
         while current_name in self._possible_names:
             # Generate random name. Remove dummy
-            new_suffix = create_variable(current_name).name[5:]
+            new_suffix = '_' + create_variable(r).name
             current_name = current_name + new_suffix
```
yguclu: Shouldn't this be the following instead?
```
current_name = current_name + '_' + create_random_string(r)
```

pyccel/parser/semantic.py
```diff
@@ -255,10 +255,12 @@ def annotate(self, **settings):
         return ast
 
     def _get_new_name(self, current_name):
+        if current_name not in self._possible_names:
+            return current_name
+        current_name = current_name + '_'
```
yguclu, line 260: It looks like in most cases this function will return the name above, which ends with an underscore.
Is this what we really want?

EmilyBourne: In most cases it won't reach the part where the underscore is added. If it reaches it then it definitely enters the while loop

yguclu: If `current_name` is not among the `_possible_names` then you add an underscore to it. After that change it will not be in `_possible_names` anymore, therefore the interpreter will not enter the loop...

EmilyBourne: Ah! You're right sorry. I wanted a do while but this is python ^^

PR #376, opened 2020-07-31
Add missing line to complete fixing issue 374

EmilyBourne:
Save new name in possible names. Split collisions2.py test into 2 files to avoid 1 test masking the other


pyccel/parser/semantic.py
```diff
@@ -256,6 +256,7 @@ def annotate(self, **settings):
 
     def _get_new_name(self, current_name):
         if current_name not in self._possible_names:
+            self._possible_names.add(current_name)
```
yguclu, line 259: Is `self._possible_names` a list? Should't we use `append` instead of `add`?

EmilyBourne: It's a set

yguclu: I see, that makes sense

PR #377, opened 2020-08-04
Fix tuple/list multiplication and addition

### LINKED ISSUE #329: Pull request #322 breaks tuple multiplication support
Statements such as the following will be broken after this pull request
```
tup1 = (0, False, 3+1j) * 3
tup2 = tup1 * 2
```

### LINKED ISSUE #349: Tuple/List concatenation does not work

### LINKED ISSUE #378: Pyccel is not compatible with textx v2.2.0
See issue https://github.com/textX/textX/issues/266

A simple fix is to change the name of the attribute `name` in `FunctionMacroStmt`

EmilyBourne:
Fix tuple/list concatenation (Fixes #349). Fix tuple/list multiplication (fixes #329). Fix textx 2.2.0 compatibility issues (fixes #378).

- Provide variables in the same way that they are accessed in TupleVariable function get_vars.
- Check order when defining shape of IndexedElement
- Correctly recognise tuples of type NativeGeneric() as inhomogenous
- Handle tuple/list concatenation (fixes #349)
- Move concatenation out of bugs folder
- Add tuple tests for multiplication and addition
- Rename name variable of FunctionMacroStmt class to work within textx 2.2.0 rules (fixes #378)


pyccel/ast/builtins.py
```diff
@@ -194,7 +194,7 @@ def __new__(cls, *args):
     def __init__(self, *args):
         if self.stage == 'syntactic' or len(args) == 0:
             return
-        is_homogeneous = all(args[0].dtype==a.dtype for a in args[1:])
+        is_homogeneous = all(False if a.dtype is NativeGeneric() else args[0].dtype==a.dtype for a in args[1:])
         is_homogeneous = is_homogeneous and all(args[0].rank==a.rank   for a in args[1:])
```
yguclu, line 198: It seems to me that these two lines could be made more readable by combining them into a single instruction:
```
is_homogeneous = all(a.dtype is not NativeGeneric() and args[0].dtype==a.dtype and args[0].rank==a.rank for a in args[1:])
```
Does that make sense?

pyccel/parser/semantic.py
```diff
@@ -1114,10 +1114,10 @@ def _handle_PyccelOperator(self, expr, **settings):
     def _visit_PyccelAdd(self, expr, **settings):
         args = [self._visit(a, **settings) for a in expr.args]
         if isinstance(args[0], (TupleVariable, PythonTuple, Tuple, List)):
-            expr_new = Concatenate(args, True)
-            errors.report(PYCCEL_RESTRICTION_TODO, symbol=expr,
-                bounding_box=(self._current_fst_node.lineno, self._current_fst_node.col_offset),
-                severity='fatal', blocker=self.blocking)
+            tuple_args = []
+            for a in args:
+                tuple_args += a.get_vars() if isinstance(a, TupleVariable) else a.args
```
yguclu, line 1119: Shouldn't a list comprehension be faster (and more readable) here?

EmilyBourne: Yes, but it would require unpacking in a list comprehension which python doesn't allow

yguclu: Right! :sweat_smile: And what about this?
```
f = lambda a: a.get_vars() if isinstance(a, TupleVariable) else a.args
tuple_args = [ai for a in args for ai in f(a)]
```
Or a little more compactly (but maybe harder to read):
```
tuple_args = [ai for a in args for ai in (a.get_vars() if isinstance(a, TupleVariable) else a.args)]
```

EmilyBourne: Won't that produce the same output?
[f(a) for a in args] will give [[a,b,c], [d,e]]
so ai in f(a) will be [a,b,c], then [d,e]

yguclu: Yes, but there is a double loop...

EmilyBourne: Yes but the second loop doesn't unpack anything either. This effectively does:
```
a = [[1,2], [3,4]]
b = [ai for ai in a]
```
as f(a) returns a list

EmilyBourne: Ah! Ok, I have understood what is going on now. You are right this works

yguclu:
```
In [1]: args = [[1, 2, 3], [4, 5, 6]]                                                                       

In [2]: tuple_args = [ai for a in args for ai in a]                                                         

In [3]: tuple_args                                                                                          
Out[3]: [1, 2, 3, 4, 5, 6]
```

pyccel/parser/semantic.py
```diff
@@ -1114,9 +1114,8 @@ def _handle_PyccelOperator(self, expr, **settings):
     def _visit_PyccelAdd(self, expr, **settings):
         args = [self._visit(a, **settings) for a in expr.args]
         if isinstance(args[0], (TupleVariable, PythonTuple, Tuple, List)):
-            tuple_args = []
-            for a in args:
-                tuple_args += a.get_vars() if isinstance(a, TupleVariable) else a.args
+            get_vars = lambda a: a.get_vars() if isinstance(a, TupleVariable) else a.args
+            tuple_args = [ai for a in args for ai in f(a)]
```
yguclu, line 1118: I'm afraid there is a typo: `f` >> `get_vars`.

EmilyBourne: Arrr, stupid mistake ^^

PR #379, opened 2020-08-05
Remove unnecessary dependencies

yguclu:
* Remove completely unnecessary dependencies: 'pylint', 'parse';
* Remove 'tabulate' dependency after deleting obsolete function that used it;
* Move 'scipy' dependency from library to unit tests (note: scipy.pi is numpy.pi);
* Update README file.


pyccel/parser/base.py
```diff
@@ -410,40 +410,40 @@ def print_namespace(self):
...
-
-            txt = tabulate(table, tablefmt='rst')
-            print (txt)
-        except NotImplementedError:
-
-            print ('------- namespace.{} -------'.format(entry))
-            for (k, v) in self.namespace[entry].items():
-                print ('{var} \t :: \t {value}'.format(var=k, value=v))
-            print ('-------------------------')
+#    def view_namespace(self, entry):
```
EmilyBourne, line 413: Should this function be moved to the obsolete folder rather than commenting it?
That is where we have the other functions which are not tested but are still useful. If it is not useful then maybe it should be removed rather than commented

README.rst
```diff
@@ -188,6 +188,7 @@ Additional packages
 
 In order to run the unit tests and to get a coverage report, three additional Python packages should be installed:::
```
EmilyBourne, line 189: four additional packages

pyccel/parser/base.py
yguclu: This is a class method, so moving it to the "obsolete" folder does not seem useful unless we also copy the class there. As for removing it, we should ask the original author (@ratnania I believe) if he thinks this function could still be useful.

EmilyBourne: Although it is a class method it doesn't seem to use any private properties of the class. If we passed a namespace (self.namespace) instead of passing the class instance then it could easily be separated from the class

README.rst
yguclu: yes, thanks!

pyccel/parser/base.py
yguclu: OK I will move this function to a separate file in the `obsolete` folder.

yguclu: Actually I don't see how this function was supposed to work. `self` is a `BasicParser` object, hence`self.namespace` is an instance of `Scope`, which has no `__getitem__` method. So the expression `self.namespace[entry]` has no meaning.

@saidctb Maybe `self.namespace` was a dictionary at some point?

EmilyBourne: I suspect it was an OrderedDict at some point. I would guess that the expected options for entry are:
- variables
- classes
- functions
- macros
- headers

And possibly
- imports

PR #380, opened 2020-08-15
Fix error message in codegen

rakati:
fix variable name

### DIFF
--- a/pyccel/codegen/codegen.py
+++ b/pyccel/codegen/codegen.py
@@ -233,7 +233,7 @@ class Codegen(object):
         language = settings.pop('language', 'fortran')
 
         if not language in ['fortran', 'c', 'python']:
-            raise ValueError('the language {} not available'.format(lanugage))
+            raise ValueError('the language {} not available'.format(language))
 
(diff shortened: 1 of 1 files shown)


pyccel/codegen/codegen.py
```diff
@@ -233,7 +233,7 @@ def doprint(self, **settings):
         language = settings.pop('language', 'fortran')
 
         if not language in ['fortran', 'c', 'python']:
-            raise ValueError('the language {} not available'.format(lanugage))
+            raise ValueError('the language {} not available'.format(language))
```
yguclu, line 236: Hi @rakati thanks for this PR!

Since we are fixing a typo here, why not also improving the error message to something that is correct English? I suggest the following:
```
raise ValueError('{} language is not available'.format(language))
```

yguclu (changes requested):

pyccel/codegen/codegen.py
saidctb: done

yguclu (approved):

PR #383, opened 2020-08-17
Fix list comprehension for multiple for loops

EmilyBourne:
The list comprehension was not working if more than 1 for was used as the method used for handling this relies on sympy. This PR fixes this by adding functions to convert simple expressions from pyccel to sympy and back again

- Ensure tuple shape is a valid pyccel object
- Add module to convert between basic sympy and pyccel expressions
- Correct Len printing for multi-dimensional objects with 'C' ordering
- Correct ceiling printer
- Fix list comprehension by converting sizes into sympy objects which can be summed over
- Add checks to see whether the upper bound is larger than the lower bound. If this is not the case then sympy will not carry out the sum correctly as it uses Karr's convention (https://docs.sympy.org/latest/modules/concrete.html#sympy.concrete.summations.Sum). The solution should be to use max(size,0) however sympy cannot sum over this object.
- Add list comprehension tests
- Add test to ensure correct error is raised for list of inhomogenous objects
- Add tests for boundary checking error


pyccel/codegen/printing/fcode.py
```diff
@@ -682,7 +682,11 @@ def _print_SumFunction(self, expr):
         return str(expr)
 
     def _print_Len(self, expr):
-        return 'size(%s,1)'%(self._print(expr.arg))
+        if (expr.arg.order == 'F'):
+            return 'size(%s,1)'%(self._print(expr.arg))
+        else:
+            return 'size({var},{idx})'.format(var = self._print(expr.arg),
+                                              idx = self._print(expr.arg.rank))
```
yguclu, line 689: Nice fix!

But the `if-else` block obscures the fact that only `idx` changes depending on `expr.arg.order`, so we could make this a bit more readable:
```
var = expr.arg
idx = 1 if var.order == 'F' else var.rank
return 'size({},{})'.format(self._print(var), self._print(idx))
```

pyccel/errors/messages.py
```diff
@@ -82,6 +82,8 @@
 PYCCEL_RESTRICTION_TODO = "Pyccel has encountered syntax that has not been implemented yet. Please create an issue at https://github.com/pyccel/pyccel/issues and provide a small example of your problem. Do not forget to specify your target language"
 PYCCEL_RESTRICTION_MULTIPLE_COMPARISONS = 'Uncovered multi operator comparison statement'
 PYCCEL_RESTRICTION_LIST_COMPREHENSION_ASSIGN = "The result of a list comprehension expression must be saved in a variable"
+PYCCEL_RESTRICTION_LIST_COMPREHENSION_SIZE = 'Could not deduce the size of this list comprehension. If you believe this expression is simple then please create an issue at https://github.com/pyccel/pyccel/issues and provide a small example of your problem.'
+PYCCEL_RESTRICTION_LIST_COMPREHENSION_LIMITS = 'Pyccel cannot handle this List comprehension. This is because there are occasions where the upper bound is smaller than the lower bound for variable {}'
```
yguclu, line 86: Did you really want to capitalize the first letter of `list` here?

yguclu (approved):

PR #384, opened 2020-08-17
Create a ClassDef instance for Numpy objects and use it to access numpy properties

### LINKED ISSUE #239: Support numpy properties of arrays
Often one uses `array.shape` or similar constructs. This could be supported by pyccel at some point alternatively to `shape(array)`.

EmilyBourne:
Create a ClassDef instance for Numpy objects. Use it to stock numpy properties. These are detected thanks to an internal `numpy_wrapper` decorator which provides the function being wrapped. Fixes #239

- Create `NumpyArrayClass ` which is a ClassDef object containing all functions accessible via array properties
- Add `numpy_wrapper` decorator to numpy methods to mark these methods as being accessed by passing the class instance as an argument rather than calling the class instance directly
- Handle `numpy_wrapper` decorator in `_visit_DottedVariable` class
- Add tests for Numpy functions: np.sum(x), np.min(x), np.max(x)
- Add tests for Numpy array methods: x.sum(), x.min(), x.max()
- Add tests for Numpy array properties: x.shape

### DIFF
--- a/tests/epyccel/recognised_functions/test_numpy_funcs.py
+++ b/tests/epyccel/recognised_functions/test_numpy_funcs.py
@@ -27,7 +27,6 @@ from conftest import *
 #    complex128
 #    complex64
 #    matmul
-#    sum
 #    prod
 #    product
 #    linspace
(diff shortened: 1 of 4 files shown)


tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -486,6 +485,27 @@ def test_shape_2d(f):
...
+    from numpy import empty
+    f1 = epyccel(test_shape_1d)
+    f2 = epyccel(test_shape_2d)
+    n1 = randint(20)
+    n2 = randint(20)
+    n3 = randint(20)
+    x1 = empty(n1,dtype = int)
+    x2 = empty((n2,n3), dtype = int)
+    assert(f1(x1) == test_shape_1d(x1))
+    assert(all(isclose(f2(x2), test_shape_2d(x2))))
```
yguclu, line 507: Why are you using `isclose` to compare tuples of integers?

EmilyBourne: Oops sorry, that's a mistake

yguclu (approved):

yguclu:
The last point in the PR description reads "TODO: Add tests", but I see that several tests were added with commit cb2d667. What did that comment refer to?

PR #385, opened 2020-08-18
Add issue template

EmilyBourne:


yguclu:
According to Codacy we should start using hyphens to separate words in a file name, and avoid underscores... I think we can agree to that for all files that do not contain source code.

yguclu (approved):

yguclu (approved):

Issue #390, opened 2020-08-19
Some data types don't work as arguments in C ex (int8 int16 bool)


EmilyBourne:
**Some data types don't work as argument ex {int8 int16 bool complex complex64 complex128}**

The following code:
```
@types('bool')
def func(x):
    return x
```

generates this error when translating to C:
```
KeyError: ('bool', 4)
```

cc: @MenoIy

EmilyBourne:
Continuing the discussion begun in issue #388 here as it pertains to fixing this issue
> When inspecting the 'bool' variable passed as argument :
> 
> ```
> from pyccel.decorators import types
> 
> @types(bool)
> def func(x):
>     return x
> ```
> 
> ```
> >>> code.routines[0].arguments[0].inspect()
> >>> Variable
>   name           = x
>   dtype          = bool
>   rank           = 0
>   allocatable    = False
>   shape          = ()
>   cls_base       = None
>   cls_parameters = None
>   is_pointer     = False
>   is_target      = False
>   is_polymorphic = False
>   is_optional    = False
> ```
> 
> and
> 
> ```
> >>>code.routines[0].arguments[0].precision
> 4
> ```
> 
> The variable type is bool and the precision is **4**. but in dictionary **dtype_registry** **in pyccel/codegen/printing/ccode.py**
> **{('bool', 1) : int}** the precision set for a boolean type is **1** . When the precision is set to 4 the code work fine but
> when working with Py_BuildValue we can't product a boolean type python object exemple :
> 
> ```
> x = False
> y = func(x)
> print(y , type(y))
> print(type(x))
> 
> >>> 0 <class 'int'>
> >>> <class 'bool'>
> ```

The precision of booleans was changed in PR #242. Probably we forgot to change the boolean type to match at that point which is why we have problems here.

> when working with Py_BuildValue we can't product a boolean type python object
You are right! I had not realised this. This explains why we have problems with booleans with f2py (see issue #370). However it seems that we can use Py_BuildValue to create `True` or `False` as explained here: https://stackoverflow.com/questions/21221242/py-buildvalue-make-tuple-with-bool

MenoIy:
Yes, as you mentioned both methods will work
```
  Py_BuildValue("O",  test ? Py_True: Py_False);
```
and
```
  Py_BuildValue("N",  PyBool_FromLong(test) );
```

[...]

PR #393, opened 2020-08-22
If and While statements working in C as in Fortran

### LINKED ISSUE #382: Add If and While statements (target language : C)
```

@types(int,int)
def f(x, y):
    r = 0
    r = x + y
    if r == 0:
      r = 1
    return r
```

```

@types(int,int)
def f(x, y):
    r = 0
    z = x + y
    while (r < z):
      r = r + 1
    return r

```

When converting those simple codes containing the **IF** and **WHILE** statements using the command >> **pyceel code.py --language c** there is a pyceel fatal error that say:

```
**ERROR at code generation stage
pyccel: [codegen]
 |fatal| Pyccel has encountered syntax that has not been implemented yet. Please create an issue at https://github.com/pyccel/pyccel/issues and provide a small example of your problem. Do not forget to specify your target language (If((PyccelEq(r, 0), CodeBlock([r := 1])), (True, CodeBlock([]))))**
```

nhamidn:
This PR adds the If and While statements to the C code printing (closes #382):

- Make the While statement working with all relational and logical operators
- Make the If statement working with all relational and logical operators
- Add more test for the While statement in epyccel

### DIFF
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -406,10 +406,6 @@ class CCodePrinter(CodePrinter):
 
 
     def _print_Program(self, expr):
-
-        name = 'prog_{0}'.format(self._print(expr.name))
-        name = name.replace('.', '_')
-
         imports  = list(expr.imports)
         imports += [Import('stdlib.h')]
         imports  = '\n'.join(self._print(i) for i in imports)
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -110,7 +110,7 @@ class CCodePrinter(CodePrinter):
         return '\n\n'.join(self._print(i) for i in expr.body)
 
     def _print_While(self,expr):
-        #print(expr.test)
+        print(expr.test)
         code = "while (%s)\n{" % self._print(expr.test)
         code = code + "\n %s" % self._print(expr.body) + "\n}"
         return (code)
@@ -145,11 +145,11 @@ class CCodePrinter(CodePrinter):
 
     def _print_PyccelAnd(self, expr):
         args = [self._print(a) for a in expr.args]
-        return '(' + ' && '.join(a for a in args) + ')'
+        return ' && '.join(a for a in args)
 
     def _print_PyccelOr(self, expr):
         args = [self._print(a) for a in expr.args]
-        return '(' + ' || '.join(a for a in args) + ')'
+        return ' || '.join(a for a in args)
 
     def _print_PyccelEq(self, expr):
         lhs = self._print(expr.args[0])
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -407,7 +407,8 @@ class CCodePrinter(CodePrinter):
 
     def _print_Program(self, expr):
 
-        name = 'prog_{0}'.format(self._print(expr.name)).replace('.', '_')
+        name = 'prog_{0}'.format(self._print(expr.name))
+        name = name.replace('.', '_')
 
         imports  = list(expr.imports)
         imports += [Import('stdlib.h')]
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -118,7 +118,6 @@ class CCodePrinter(CodePrinter):
         lines = []
         for i, (c, e) in enumerate(expr.args):
             var = self._print(e)
-            #change this line (removes an extra empty else{})
             if (var == ''):
                 break
             if i == 0:
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -1,9 +1,6 @@
 # coding: utf-8
 # pylint: disable=R0201
 
-from sympy.core import Tuple
-from pyccel.ast.numbers   import BooleanTrue
-from pyccel.ast.builtins import PythonTuple
 from pyccel.ast.core import If
 
 from sympy.core import S
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -109,6 +109,12 @@ class CCodePrinter(CodePrinter):
     def _print_Module(self, expr):
         return '\n\n'.join(self._print(i) for i in expr.body)
 
+    def _print_While(self,expr):
+        #print(expr.test)
+        code = "while (%s)\n{" % self._print(expr.test)
+        code = code + "\n %s" % self._print(expr.body) + "\n}"
+        return (code)
+
     def _print_If(self, expr):
         lines = []
         for i, (c, e) in enumerate(expr.args):
@@ -139,25 +145,21 @@ class CCodePrinter(CodePrinter):
 
     def _print_PyccelAnd(self, expr):
         args = [self._print(a) for a in expr.args]
-        return ' && '.join(a for a in args)
+        return '(' + ' && '.join(a for a in args) + ')'
 
     def _print_PyccelOr(self, expr):
         args = [self._print(a) for a in expr.args]
-        return ' || '.join(a for a in args)
+        return '(' + ' || '.join(a for a in args) + ')'
 
     def _print_PyccelEq(self, expr):
         lhs = self._print(expr.args[0])
         rhs = self._print(expr.args[1])
-        a = expr.args[0].dtype
-        b = expr.args[1].dtype
         return '{0} == {1}'.format(lhs, rhs)
 
     def _print_PyccelNe(self, expr):
         lhs = self._print(expr.args[0])
         rhs = self._print(expr.args[1])
-        a = expr.args[0].dtype
-        b = expr.args[1].dtype
-        return '{0} != {1}'.format(lhs, rhs)  
+        return '{0} != {1}'.format(lhs, rhs)
 
     def _print_PyccelLt(self, expr):
         lhs = self._print(expr.args[0])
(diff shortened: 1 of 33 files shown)


pyccel/codegen/printing/ccode.py
```diff
@@ -104,6 +109,82 @@ def _traverse_matrix_indices(self, mat):
     def _print_Module(self, expr):
         return '\n\n'.join(self._print(i) for i in expr.body)
 
+    def _print_While(self,expr):
+        print(expr.test)
```
EmilyBourne, line 113: You seem to have forgotten to remove this print

pyccel/codegen/printing/ccode.py
```diff
@@ -104,6 +109,82 @@ def _traverse_matrix_indices(self, mat):
...
+        code = "while (%s)\n{" % self._print(expr.test)
```
EmilyBourne, line 114: We should try to use the same indentation style for the braces throughout.
The following code:
```
def f():
    x = 3.0

x = 0.0

for i in range(2):
    x += 1.0
    
j = 3 
    
if (j == 2): 
    x += 1.0
elif j == 4:
    x += 1.0 
else:
    x += 1.0
    
while (j<5): 
    j+=1
```
translates to:
```
void f() {
    double x;
    x = 3.0;
}

int main(){
    double x;
    long j;
    long i;
    x = 0.0;
    for (i = 0; i < 2; i += 1) {
        x += 1.0;
    } 
    j = 3;
    if (j == 2){
        x += 1.0;
    } 
    else if (j == 4){
        x += 1.0;
    }
    else{
        x += 1.0;
    }
    while (j < 5)
    {
        j += 1;
    }
    return 0;
}
```
The for loop and the function definition use the convention `CONDITION SPACE BRACE`, the main program and the if loop use the convention `CONDITION BRACE` and the while loop uses the convention
```
CONDITION
BRACE
```

The most common conventions (https://en.wikipedia.org/wiki/Indentation_style#Brace_placement_in_compound_statements) in C/C++ are Allman (see while loop) or K&R variant 1TBS (see for and function def). Personally I prefer Allman style as it makes it easy to spot matching braces but K&R works too. The most important thing is that the same style is used throughout.

Do you have an opinion about the style @yguclu , @saidctb ?

pyccel/codegen/printing/ccode.py
```diff
@@ -104,6 +109,82 @@ def _traverse_matrix_indices(self, mat):
...
+        code = code + "\n %s" % self._print(expr.body) + "\n}"
+        return (code)
+
+    def _print_If(self, expr):
+        lines = []
+        for i, (c, e) in enumerate(expr.args):
+            #change this line (removes an extra empty else{})
+            if (self._print(e) == ''):
```
EmilyBourne, line 122: `self._print(e)` should be stored in a variable at the start of the for loop. That way it can be reused on line 136

EmilyBourne (changes requested):
Please add tests for this code.
The first step of tests can be created easily:
1. Copy the file `tests/codegen/fcode/scripts/whiles.py` into the folder `tests/codegen/ccode/scripts`
2. Copy the file `tests/codegen/fcode/scripts/ifs.py` into the folder `tests/codegen/ccode/scripts` and remove the references to the array `b`

In the epyccel folder there are already tests for if statements which we will be able to use once issue #390 is fixed however there are no tests for while loops. It would be good if some where added to the `tests/epyccel/test_loops.py` and `tests/epyccel/modules/loops.py` files

pyccel/codegen/printing/ccode.py
saidctb: i don't have any preference regarding the style but we should choose one  , we probably need to make it automatic and easy to change.

nhamidn: at this moment im gonna use the allman style for all the statements and the function definition

yguclu: I also don't have any preference about style, as long as it is consistently used by our C printer. As for being able to switch between styles automatically as suggested by @saidctb, that may be a good idea but I don't know how much that's going to complicate our code.

EmilyBourne: I think we can leave switching between styles for another pull request if we want to add that.
If @nhamidn has already used Allman style everywhere then let's stick with that for now.

@nhamidn would you mind also formatting the main in the same way (see `_print_Program`)

nhamidn: @EmilyBourne im gonna take care of it

tests/codegen/ccode/scripts/ifs.py
```diff
@@ -0,0 +1,99 @@
+from numpy import zeros
+
+x = 0
+y = 0
+z = 3
+b = zeros(64,'int')
```
EmilyBourne, line 6: See above. This line and all lines using it must be removed as arrays are not yet handled in C

EmilyBourne (changes requested):

EmilyBourne:
Remember, the tests in the `codegen` folder only test if the codegen step executes successfully. It does **not** test if the code gives the same results or even if it compiles. This is why it is much more important to have lots of tests in the `epyccel` folder

pyccel/codegen/printing/ccode.py
```diff
@@ -104,6 +109,82 @@ def _traverse_matrix_indices(self, mat):
...
+    def _print_While(self,expr):
+        code = "while (%s)\n{" % self._print(expr.test)
+        code = code + "\n %s" % self._print(expr.body) + "\n}"
+        return (code)
+
+    def _print_If(self, expr):
+        lines = []
+        for i, (c, e) in enumerate(expr.args):
+            var = self._print(e)
+            #change this line (removes an extra empty else{})
```
EmilyBourne, line 121: Is this a `TODO`? If not, then what is being changed in the line?

pyccel/codegen/printing/ccode.py
```diff
@@ -104,6 +109,82 @@ def _traverse_matrix_indices(self, mat):
...
+                break
+            if i == 0:
+                lines.append("if (%s)\n{" % self._print(c))
+            elif i == len(expr.args) - 1 and c is BooleanTrue():
+                lines.append("else\n{")
+            else:
+                lines.append("else if (%s)\n{" % self._print(c))
+            #e can be a tuple when one python statement
+            #results in multiple c statements
+            if isinstance(e, (list, tuple, Tuple, PythonTuple)):
```
EmilyBourne, line 132: Are you sure that `e` can be a `PythonTuple`? A `PythonTuple` is used to represent a tuple which was written in python, whereas the other 3 are used to collect statements

EmilyBourne (changes requested):

pyccel/codegen/printing/ccode.py
rmahjoubi: Yes this was a TODO, I just forgot it there, I will remove it

pyccel/codegen/printing/ccode.py
rmahjoubi: I think e can't be neither of those types, there is no need for that condition to exist.
I was thinking e could be a tuple when a python statement results in multiple c statements, but this isn't the case. I tested it and e is always of type `CodeBlock`.
Same goes for fortran

EmilyBourne: Ah, very good. I though that ought to be the case. In that case this if loop can be removed

pyccel/codegen/printing/ccode.py
```diff
@@ -1,6 +1,11 @@
 # coding: utf-8
 # pylint: disable=R0201
 
+from sympy.core import Tuple
```
EmilyBourne, line 4: This is a new import but I can't see where it is used

pyccel/codegen/printing/ccode.py
```diff
@@ -1,6 +1,11 @@
...
+from pyccel.ast.numbers   import BooleanTrue
+from pyccel.ast.builtins import PythonTuple
```
EmilyBourne, line 6: See above

EmilyBourne (changes requested):

pyccel/codegen/printing/ccode.py
nhamidn: yes it need to be removed, it was used in the if statement.

pyccel/codegen/printing/ccode.py
```diff
@@ -104,6 +106,74 @@ def _traverse_matrix_indices(self, mat):
...
+
+    def _print_If(self, expr):
+        lines = []
+        for i, (c, e) in enumerate(expr.args):
+            var = self._print(e)
+            if (var == ''):
+                break
+            if i == 0:
+                lines.append("if (%s)\n{" % self._print(c))
+            elif i == len(expr.args) - 1 and c is BooleanTrue():
```
EmilyBourne, line 122: `BooleanTrue` was not an unused import

pyccel/codegen/printing/ccode.py
rmahjoubi: `Tuple` and `PythonTuple` were used before I removed the condition above, they are no longer being used

EmilyBourne (approved):

tests/codegen/ccode/scripts/ifs.py
```diff
@@ -0,0 +1,93 @@
...
+if 1<0:
+    x=4
+
+if (x==2 or y==1) and z>3 :
+    x = 3
+    y = 4
+
+while x>2 and y<1:
+    x = x-1
+    y = 4
```
yguclu, line 15: Shouldn't this belong to file `whiles.py`?

tests/codegen/ccode/scripts/ifs.py
```diff
@@ -0,0 +1,93 @@
...
+
+if x>2 or y<=1:
+    x = x-1
+    y = y-1
+
+while x>2 or y<1:
+    x = 1
+    y = 3
```
yguclu, line 23: Same comment as above

tests/codegen/ccode/scripts/ifs.py
EmilyBourne: I agree but I suspect that these crept in as they are present in `tests/codegen/fcode/scripts/ifs.py`

yguclu (changes requested):
Please add a detailed description to this PR, to be used as a squash commit message when we merge branch **add_c_if_while** into **master**. See for example commit 96d9bee.

yguclu (approved):

PR #396, opened 2020-08-24
Add PyccelAssociativeParenthesis where necessary when creating an Operator

### LINKED ISSUE #392: Parentheses are not being taken into account when there are mixed 'AND' and 'OR'
Parentheses are not being taken into account when there is 'And' and 'Or' in the while and if statements, which lead to a wrong generated code in both C and Fortran 

For example this piece of python code will be generated in a wrong way

```

@types(int, int)
def f(x, y):
    while ((x==10 or y==10) and (x==20 or y== 20)):
         x = x + 1
    return 0
```

The generated code in Fortran

```
module test

implicit none

contains
function f(x, y) result(Dummy_432)

  implicit none
  integer(kind=8)  :: Dummy_432  
  integer(kind=8), value  :: x
  integer(kind=8), value  :: y

  do while (x == 10_8  .or. y == 10_8  .and. x == 20_8  .or. y == 20_8 &
      )
    x = x + 1_8
  end do
  Dummy_432 = 0_8
  return
end function

end module
```

The code generated in C which fail in the compilation stage.

```
long f(long x, long y) {
    long Dummy_125;
    while (x == 10 || y == 10 && x == 20 || y == 20)
    {
        x = x + 1;
    }
    Dummy_125 = 0;
    return Dummy_125;
}
```
Parentheses must be taken into account when there are mixed 'and' and 'or'.

EmilyBourne:
Check the precedence of the operators in the syntactic step and add in brackets where necessary. Fixes #392 

- Add precedence to Operators. Add handle_precedence method to add parentheses where necessary
   - This value corresponds to the line number in the pyccel precedence table (https://docs.python.org/3/reference/expressions.html#operator-precedence)
   - Most objects have a precedence of 17
- Remove change_priority function to rely on handle_precedence instead
- Add commutativity property to operators to reduce the number of printed brackets
- Only check precedence in syntactic stage
- Add tests for boolean operators
- Add minus brackets test to simplify debugging
- Collect PyccelAdd and PyccelOr in the obvious way using handle_precedence to manage parentheses
- Add tests for boolean operators with multiple arguments
- Handle any number of arguments for BooleanOperators


saidctb (approved):

pyccel/ast/core.py
```diff
@@ -210,11 +210,29 @@ def broadcast(shape_1, shape_2):
             errors.report(msg,severity='fatal')
     return tuple(new_shape)
 
+def handle_precedence(args, my_precedence, commutative = False):
+    precedence = [getattr(a, 'precedence', 17) for a in args]
```
yguclu, line 214: Is it safe to use 17 as a default? Why not letting Python raise an attribute error instead?

pyccel/ast/core.py
```diff
@@ -210,11 +210,29 @@ def broadcast(shape_1, shape_2):
...
+
+    return args
+
 class PyccelOperator(Expr, PyccelAstNode):
 
     def __init__(self, *args):
 
         if self.stage == 'syntactic':
+            args = handle_precedence(args, self.precedence, self._commutative)
+            self._args = tuple(args)
```
yguclu, line 235: Doesn't `handle_precedence` return a `tuple` object?

pyccel/ast/core.py
```diff
@@ -260,18 +278,30 @@ def __init__(self, *args):
...
     pass
 class PyccelDiv(PyccelOperator):
-    p = 2
+    _precedence = 13
+    _commutative = False
     def __init__(self, *args):
+
         if self.stage == 'syntactic':
+            args = handle_precedence(args, self.precedence, self._commutative)
+            self._args = tuple(args)
```
yguclu, line 304: Same comment as above. `args` is already a `tuple`

pyccel/ast/core.py
```diff
@@ -307,15 +337,22 @@ def __init__(self, *args):
...
+    _commutative = False
 
 class PyccelBooleanOperator(Expr, PyccelAstNode):
+    _precedence = 7
 
     def __init__(self, *args):
+
         if self.stage == 'syntactic':
+            args = handle_precedence(args, self.precedence, self._commutative)
+            self._args = tuple(args)
```
yguclu, line 353: As above

pyccel/ast/core.py
```diff
@@ -359,32 +407,77 @@ def __init__(self, a):
...
     _dtype = NativeBool()
     _rank  = 0
     _shape = ()
     _precision = default_precision['bool']
+    _precedence = 5
+
+    def __init__(self, *args):
+        if self.stage == 'syntactic':
+            args = handle_precedence(args, self.precedence, True)
+            self._args = tuple(args)
```
yguclu, line 444: Also here

pyccel/ast/core.py
```diff
@@ -359,32 +407,77 @@ def __init__(self, a):
...
     _dtype = NativeBool()
     _rank  = 0
     _shape = ()
     _precision = default_precision['bool']
+    _precedence = 4
+
+    def __init__(self, *args):
+        if self.stage == 'syntactic':
+            args = handle_precedence(args, self.precedence, True)
+            self._args = tuple(args)
```
yguclu, line 460: Also here

pyccel/ast/core.py
```diff
@@ -359,32 +407,77 @@ def __init__(self, a):
...
     _dtype = NativeBool()
     _rank  = 0
     _shape = ()
     _precision = default_precision['bool']
+    _precedence = 6
+
+    def __init__(self, *args):
+        if self.stage == 'syntactic':
+            args = handle_precedence(args, self.precedence)
+            self._args = tuple(args)
```
yguclu, line 476: Also here

pyccel/ast/core.py
```diff
@@ -402,9 +495,12 @@ class Is(Basic, PyccelAstNode):
...
     _shape = ()
     _precision = default_precision['bool']
+    _precedence = 7
 
-    def __new__(cls, lhs, rhs):
-        return Basic.__new__(cls, lhs, rhs)
+    def __init__(self, *args):
+        if self.stage == 'syntactic':
+            args = handle_precedence(args, self.precedence)
+            self._args = tuple(args)
```
yguclu, line 503: Also here

pyccel/ast/core.py
```diff
@@ -432,9 +532,12 @@ class IsNot(Basic, PyccelAstNode):
...
```
yguclu, line 540: Also here

pyccel/ast/basic.py
```diff
@@ -16,12 +16,12 @@ def fst(self):
...
-    _rank      = None
-    _dtype     = None
-    _precision = None
-    _order     = None
+    stage       = None
+    _shape      = None
+    _rank       = None
+    _dtype      = None
+    _precision  = None
+    _order      = None
```
yguclu, line 24: If we want to be consistent with how the module was written so far, shouldn't we add the line
`_precedence = None` here?

pyccel/ast/core.py
EmilyBourne: As per here: https://docs.python.org/3/reference/expressions.html#operator-precedence
17 is Subscription, slicing, call, attribute reference
So a Variable would have `precedence = 17`, as would an IndexedElement, and almost every other PyccelAstNode
In the syntactic step there are still a certain number of sympy objects. For example a `Variable` is not saved as such until its `dtype` is known. Thus in the syntactic stage it is still a (sympy) `Symbol`. This means that letting Python raise an attribute error will not work

pyccel/ast/core.py
EmilyBourne: You are right. Oops

pyccel/ast/basic.py
EmilyBourne: I had that line in at one point (that's why `_precedence` was still in the copy operator and why this file had been modified), however given that we have to handle objects that might not have a `precedence` anyway (see above, sympy `Symbol` still exists in the syntactic stage) I came to the conclusion that it wasn't necessary to add the `_precedence` property to Nodes that aren't operators

EmilyBourne: I have now checked out this file from the master branch to revert it to the original version without the additional spaces and `_precedence` in the copy operator

pyccel/ast/core.py
```diff
@@ -232,7 +232,7 @@ def __init__(self, *args):
 
         if self.stage == 'syntactic':
             args = handle_precedence(args, self.precedence, self._commutative)
-            self._args = tuple(args)
+            self._args = args
```
yguclu: I would like to make a recommendation the about coding style. When possible we should try to avoid redefining a variable that already exists (as it happens here for `args`) and create a new variable instead. Here the obvious solution actually makes the code one line shorter:
```
self._args = handle_precedence(args, self.precedence, self._commutative)
```

EmilyBourne: You are right. This came from laziness on my part originally (I didn't want to change `args` everywhere in the `__init__`) but as I have limited this step to the syntactic stage I don't need to change everything, just that line

pyccel/parser/syntactic.py
```diff
@@ -535,43 +505,33 @@ def _visit_UnaryOp(self, stmt):
...
 
         if isinstance(stmt.op, ast.Add):
-            expr = PyccelAdd(first, second)
-            return change_priority(expr)
+            return PyccelAdd(first, second)
 
         elif isinstance(stmt.op, ast.Mult):
-            expr = PyccelMul(first, second)
-            return change_priority(expr)
+            return PyccelMul(first, second)
```
nhamidn, line 519: there is a problem here, the function take in consideration only the first and the second condition.
for example this python code 

```
def f(x, y):
    while x==10 or y==10 or x == 20:
        x = x + 1
    return 0
```

converted to this : 

in fortran 

```
....

function f(x, y) result(Dummy_266)

  implicit none
  integer(kind=8)  :: Dummy_266  
  integer(kind=8), value  :: x
  integer(kind=8), value  :: y

  do while (x == 10_8  .or. y == 10_8 ) 
    x = x + 1_8
  end do
  Dummy_266 = 0_8
  return
end function

....
```
in C 

```
long f(long x, long y)
{
    long Dummy_102;
    while (x == 10 || y == 10)
    {
        x = x + 1;
    }
    Dummy_102 = 0;
    return Dummy_102;
}

```

we've made a a solution in the [bug_parentheses](https://github.com/pyccel/pyccel/tree/bug_parentheses) branch

EmilyBourne: Good catch! You are right, unlike binary operators, boolean operators can take any number of arguments.

This is actually a different bug to the one reported in issue 392. If you find something like this please open a new issue. If you plan to fix it yourself you can assign yourself the issue so everyone else knows it's being looked at.

I have used your solution to fix this.

nhamidn (approved):

rmahjoubi (approved):

yguclu (approved):

nhamidn:
The PyccelAssociativeParenthesis still not working, i just tested the master branch

for this python code : 

```
x = 0
if (x >= 0 and x <= 5) or x == 10:
    x = x + 1
```
the translated code to Fortran was like this (there is no parentheses): 

```
....
x = 0_8
if (x >= 0_8 .and. x <= 5_8 .or. x == 10_8) then
  x = x + 1_8
end if
....
```
it was working when you first opened the pull request.

EmilyBourne:
There is no way of knowing from the python ast whether there are brackets or not. So the precedence is used to put them back where they were. In python
` (x >= 0 and x <= 5) or x == 10`
Is equivalent to:
` x >= 0 and x <= 5 or x == 10`
If these things are not equivalent in the language it is translated to then I think it should be handled there

EmilyBourne:
Or we could just add brackets whenever a boolean operator contains another boolean operator for clarity

EmilyBourne:
There should be brackets for:
` x >= 0 and (x <= 5 or x == 10)`
As this is not equivalent to
`x >= 0 and x <= 5 or x == 10`

nhamidn:
we can handle it in `ccode.py` and add brackets when there are mixed AND and OR

nhamidn:
i just tested this code 

```
if (x >= 0 or x < 7) and x == 10:
        x = x + 1
```

the translated code was like this : 

```
if ((x >= 0_8 .or. x < 7_8) .and. x == 10_8) then
    x = x + 1_8
```

it look like it working for the OR operator but not for the AND

EmilyBourne:
It is working as intended for both. The intention is to add brackets when they are necessary in python.
As I said above, we have 3 cases here:
1. `(x >= 0 and x <= 5) or x == 10`
2. `x >= 0 and (x <= 5 or x == 10)`
3. `x >= 0 and x <= 5 or x == 10`

Case 2 and 3 will give the same results in python as `and` operations  are always carried out before `or` operations when there are no brackets.
The parsing step does not read the brackets (and we can't make it read them) so we add them back in where they are necessary. In case 2 they are not necessary. While in case 1 they are.

That said option 2 is a lot easier to understand than option 3. This is why C raises a warning when you don't use brackets. As a result it may be a good idea to always use brackets when mixing `and` and `or` operations for stylistic reasons. But this should be discussed in an issue. The only people who see messages in pull requests are people that participated in the Pull Request and it would be good to have other opinions too. Please copy this conversation into an issue

PR #400, opened 2020-08-25
C printer: allow functions with no return value and no arguments

### LINKED ISSUE #394: Can't translate a function with no arguments to c
**Describe the bug**
The handling of functions if the wrapper assumes that a function has arguments. As a result, we can't translate a function without arguments into c

**To Reproduce**
```python
def f():
    return 1
```

The following code is generated to handle the arguments:
```c
if (!PyArg_ParseTuple(args, "", ))
        return NULL;
```
which of course leads to the following error:
```
func_wrapper.c: In function ‘f_wrapper’:
func_wrapper.c:10:37: error: expected expression before ‘)’ token
     if (!PyArg_ParseTuple(args, "", ))
                                     ^
```

**Expected behavior**
An `if` condition should be added to the wrapper to only collect the arguments when they are expected

**Language**
C

### LINKED ISSUE #395: Can't translate a function with no return type to C
**Describe the bug**
The handling of functions if the wrapper assumes that a function returns something. As a result, we can't translate a function which does not return anything into c

**To Reproduce**
```python
@types('int')
def g(x):
    x += 1
```

The following code is generated in the wrapper:
```c
static PyObject * g_wrapper(PyObject *self, PyObject *args)
{
    long x;

    if (!PyArg_ParseTuple(args, "l", &x))
        return NULL;
    g(x)
    return Py_BuildValue("", );
}
```
which leads to an error. This error is masked by the error output generated by issue #397. When issue #397 is fixed manually and the compile command is run the error is:
```
__pyccel__/test2_wrapper.c: In function ‘g_wrapper’:
__pyccel__/test2_wrapper.c:13:30: error: expected expression before ‘)’ token
     return Py_BuildValue("", );
                              ^
__pyccel__/test2_wrapper.c:14:1: warning: control reaches end of non-void function [-Wreturn-type]
 }
 ^
```

**Expected behavior**
Functions without a return statement in python return `None`. This value is created with the expression `Py_BuildValue("")`, it is therefore simply a case of removing the extra comma

**Language**
C

### LINKED ISSUE #397: Function with no return type not correctly implemented in cwrapper
**Describe the bug**
A call to a function with no return type is not correctly implemented in cwrapper. As a result the sympy printer is used and the ending semi-colon is missed

**To Reproduce**
E.g:
```python
@types('int')
def g(x):
    x += 1
```

Gives the following error:
```bash/fortran/c
func2_wrapper.c: In function ‘g_wrapper’:
func2_wrapper.c:13:5: error: expected ‘;’ before ‘return’
     return Py_BuildValue("", );
     ^~~~~~
func2_wrapper.c:14:1: warning: control reaches end of non-void function [-Wreturn-type]
 }
 ^
```

As the generated code contains the line:
```
g(x)
```

**Expected behavior**
The following line: https://github.com/pyccel/pyccel/blob/master/pyccel/codegen/cwrapper.py#L32
needs to be changed to:
```
func_call = FunctionCall(expr, expr.arguments)
```
and the function `_print_FunctionCall` needs to be implemented in `pyccel/codegen/printer/ccode.py`

**Language**
C

### LINKED ISSUE #398: Allow user-defined functions without return arguments in C
**Describe the bug**
`_print_FunctionCall` is not implemented so user-defined functions without return arguments cannot be called in C. As per issue #397 this means that functions without return arguments cannot be exposed to python

**Test Code**
```python
@types('int')
def f(x):
    x += 4

def g():
    f(3)
```

**Language**
C

### LINKED ISSUE #399: Number or variable preceded by a negative sign not properly convert to c
**Describe the bug**
Any operand number or variable preceded by negative sign alone or at the first position in an operation 
does not preserve its negative sign after converting to c.

**To Reproduce**
```python

@types(int)
def f(x):
        x = -1
        return x
```

```c
long f(long x) {
    long Dummy_4171;
    x = (1);
    Dummy_4171 = x;
    return Dummy_4171;
}
```

**Expected behavior**
the operand 

**Language**
C

**other example**
`return -val` 
`x = -1 + x`

rakati:
This PR brings multiple improvements to the C printer:

* Allow user-defined functions with no arguments (closes #394)
* Allow functions with no return type (closes #395, closes #397)
* Allow user-defined function without return arguments (closes #398)
* Moreover, fix bug when printing unary PyccelMinus operator (closes #399)

### DIFF
--- a/pyccel/codegen/cwrapper.py
+++ b/pyccel/codegen/cwrapper.py
@@ -1,4 +1,4 @@
-from sympy.core.function       import UndefinedFunction
+from pyccel.ast.core import FunctionCall
 
 from pyccel.codegen.printing.ccode import CCodePrinter, dtype_registry
 from pyccel.ast.core import Module, Declare, Assign
@@ -33,11 +33,12 @@ def write_python_wrapper(expr, printer):
         code += "))\n        return NULL;\n    "
 
     if len(expr.results)==0:
-        func_call = UndefinedFunction(str(expr.name))(*expr.arguments)
+        func_call = FunctionCall(expr, expr.arguments)
+
         code += printer._print(func_call) + ';'
     else:
         results = expr.results if len(expr.results)>1 else expr.results[0]
-        func_call = Assign(results,UndefinedFunction(str(expr.name))(*expr.arguments))
+        func_call = Assign(results,FunctionCall(expr, expr.arguments))
         code += printer._print(func_call)
     code += '\n'
 
--- a/pyccel/codegen/cwrapper.py
+++ b/pyccel/codegen/cwrapper.py
@@ -42,7 +42,12 @@ def write_python_wrapper(expr, printer):
 
     results_dtypes = ''.join(pytype_registry[str_dtype(arg.dtype)] for arg in expr.results)
     result_names = ', '.join(res.name for res in expr.results)
-    code += "    return Py_BuildValue(\"{0}\", {1});\n".format(results_dtypes,result_names)
+    code += "    return Py_BuildValue("
+    if not expr.results: # case of function with no return value
+        code += "\"\""
+    else: # function with return value
+        code += "\"{0}\", {1}".format(results_dtypes,result_names)
+    code += ");\n"
     code += "}\n"
     return code
 
--- a/pyccel/codegen/cwrapper.py
+++ b/pyccel/codegen/cwrapper.py
@@ -22,11 +22,15 @@ def write_python_wrapper(expr, printer):
     results_decs = '\n    '.join(printer._print(i) for i in results_decs)
     code += '{0}\n    {1}\n    '.format(arg_decs, results_decs)
 
-    code += "if (!PyArg_ParseTuple(args, \""
-    code += ''.join(pytype_registry[str_dtype(arg.dtype)] for arg in expr.arguments)
-    code += "\", "
-    code += ', '.join("&" + printer._print(arg) for arg in expr.arguments)
-    code += "))\n        return NULL;\n    "
+    # Check if the function with No arguments
+    if not expr.arguments:
+        code += "if (!PyArg_ParseTuple(args, \"\"))\n        return NULL;\n    "
+    else:
+        code += "if (!PyArg_ParseTuple(args, \""
+        code += ''.join(pytype_registry[str_dtype(arg.dtype)] for arg in expr.arguments)
+        code += "\", "
+        code += ', '.join("&" + printer._print(arg) for arg in expr.arguments)
+        code += "))\n        return NULL;\n    "
 
     if len(expr.results)==0:
         func_call = UndefinedFunction(str(expr.name))(*expr.arguments)
--- a/pyccel/codegen/cwrapper.py
+++ b/pyccel/codegen/cwrapper.py
@@ -34,10 +34,11 @@ def write_python_wrapper(expr, printer):
 
     if len(expr.results)==0:
         func_call = UndefinedFunction(str(expr.name))(*expr.arguments)
+        code += printer._print(func_call) + ';'
     else:
         results = expr.results if len(expr.results)>1 else expr.results[0]
         func_call = Assign(results,UndefinedFunction(str(expr.name))(*expr.arguments))
-    code += printer._print(func_call)
+        code += printer._print(func_call)
     code += '\n'
 
     results_dtypes = ''.join(pytype_registry[str_dtype(arg.dtype)] for arg in expr.results)
--- a/pyccel/codegen/cwrapper.py
+++ b/pyccel/codegen/cwrapper.py
@@ -34,12 +34,10 @@ def write_python_wrapper(expr, printer):
 
     if len(expr.results)==0:
         func_call = FunctionCall(expr, expr.arguments)
-
-        code += printer._print(func_call) + ';'
     else:
         results = expr.results if len(expr.results)>1 else expr.results[0]
         func_call = Assign(results,FunctionCall(expr, expr.arguments))
-        code += printer._print(func_call)
+    code += printer._print(func_call)
     code += '\n'
 
     results_dtypes = ''.join(pytype_registry[str_dtype(arg.dtype)] for arg in expr.results)
--- a/pyccel/codegen/cwrapper.py
+++ b/pyccel/codegen/cwrapper.py
@@ -22,7 +22,7 @@ def write_python_wrapper(expr, printer):
     results_decs = '\n    '.join(printer._print(i) for i in results_decs)
     code += '{0}\n    {1}\n    '.format(arg_decs, results_decs)
 
-    # Check if the function with No arguments
+    # Check if the function has No arguments
     if not expr.arguments:
         code += "if (!PyArg_ParseTuple(args, \"\"))\n        return NULL;\n    "
     else:
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -12,6 +12,7 @@ from pyccel.ast.core import PyccelPow, PyccelAdd, PyccelMul, PyccelDiv, PyccelMo
 from pyccel.ast.core import PyccelEq,  PyccelNe,  PyccelLt,  PyccelLe,  PyccelGt,  PyccelGe
 from pyccel.ast.core import PyccelAnd, PyccelOr,  PyccelNot, PyccelMinus
 
+from pyccel.ast.datatypes import NativeInteger
 
 from pyccel.ast.builtins  import Range
 from pyccel.ast.core import Declare
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -175,7 +175,11 @@ class CCodePrinter(CodePrinter):
         return ' + '.join(self._print(a) for a in expr.args)
 
     def _print_PyccelMinus(self, expr):
-        return ' - '.join(self._print(a) for a in expr.args)
+        args = [self._print(a) for a in expr.args]
+
+        if len(args) == 1:
+            return '-{}'.format(args[0])
+        return ' - '.join(a for a in args)
 
     def _print_PyccelMul(self, expr):
         return ' * '.join(self._print(a) for a in expr.args)
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -164,6 +164,14 @@ class CCodePrinter(CodePrinter):
 
         return '{0} {{\n{1}\n{2}\n}}'.format(self.function_signature(expr), decs, body)
 
+    def _print_FunctionCall(self, expr):
+        func = expr.funcdef
+        # currently support only function with one or zero output
+        args = ','.join(['{}'.format(self._print(a)) for a in expr.arguments])
+        if not func.results:
+            return '{}({});'.format(func.name, args)
+        return '{}({})'.format(func.name, args)
+
     def _print_Return(self, expr):
         code = ''
         if expr.stmt:
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -148,11 +148,13 @@ class CCodePrinter(CodePrinter):
         else:
             ret_type = self._print(datatype('void'))
         name = expr.name
-
-        arg_dtypes = [self._print(i.dtype) for i in expr.arguments]
-        arg_dtypes = [dtype_registry[(dtype, arg.precision)] for dtype,arg in zip(arg_dtypes, expr.arguments)]
-        arguments  = [self._print(i) for i in expr.arguments]
-        arg_code   = ', '.join(dtype + ' ' + arg for dtype,arg in zip(arg_dtypes,arguments))
+        if not expr.arguments:
+            arg_code = 'void'
+        else:
+            arg_dtypes = [self._print(i.dtype) for i in expr.arguments]
+            arg_dtypes = [dtype_registry[(dtype, arg.precision)] for dtype,arg in zip(arg_dtypes, expr.arguments)]
+            arguments  = [self._print(i) for i in expr.arguments]
+            arg_code   = ', '.join(dtype + ' ' + arg for dtype,arg in zip(arg_dtypes,arguments))
 
         return '{0} {1}({2})'.format(ret_type, name, arg_code)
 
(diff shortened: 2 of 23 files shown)


yguclu:
@rakati Hi, thanks for your PR! You should click on the "Update branch" button to verify that your changes are compatible with the `master` branch.

saidctb (commented):
can you add tests that shows the issues  you solved with this PR

rakati:
this one of the tests that I used
now there are no errors at the conversion stage and the code generated is correct.
```

@types(int)
def simple(x):
        return x + 2

@types(int)
def no_return(y):
        y = y + 1

def no_args():
        gift = 10
        return gift

def no_args_no_return():
        d = 0
        d += 1


def call_return_simple():
        return simple(20)

def call_return_no_args():
        return no_args()

def call_no_args_no_return():
        no_args_no_return()
```
the c conversion: 
```
long simple(long x) {
    long Dummy_8084;
    Dummy_8084 = x + 2;
    return Dummy_8084;
}

void no_return(long y) {

    y = y + 1;
}

long no_args() {
    long gift;
    gift = 10;
    return gift;
}

void no_args_no_return() {
    long d;
    d = 0;
    d += 1;
}

long call_return_simple() {
    long Dummy_5461;
    Dummy_5461 = simple(20);
    return Dummy_5461;
}

long call_return_no_args() {
    long Dummy_6232;
    Dummy_6232 = no_args();
    return Dummy_6232;
}

void call_no_args_no_return() {

    no_args_no_return();
}
```
and for the wrapper.c
```
#define PY_SSIZE_T_CLEAN
#include <Python.h>

long simple(long x);
void no_return(long y);
long no_args();
void no_args_no_return();
long call_return_simple();
long call_return_no_args();
void call_no_args_no_return();

static PyObject * simple_wrapper(PyObject *self, PyObject *args)
{
    long x;
    long Dummy_8084;
    if (!PyArg_ParseTuple(args, "l", &x))
        return NULL;
    Dummy_8084 = simple(x);
    return Py_BuildValue("l", Dummy_8084);
}

static PyObject * no_return_wrapper(PyObject *self, PyObject *args)
{
    long y;

    if (!PyArg_ParseTuple(args, "l", &y))
        return NULL;
    no_return(y);
    return Py_BuildValue("");
}

static PyObject * no_args_wrapper(PyObject *self, PyObject *args)
{

    long gift;
    if (!PyArg_ParseTuple(args, ""))
        return NULL;
    gift = no_args();
    return Py_BuildValue("l", gift);
}

static PyObject * no_args_no_return_wrapper(PyObject *self, PyObject *args)
{


    if (!PyArg_ParseTuple(args, ""))
        return NULL;
    no_args_no_return();
    return Py_BuildValue("");
}

static PyObject * call_return_simple_wrapper(PyObject *self, PyObject *args)
{

    long Dummy_5461;
    if (!PyArg_ParseTuple(args, ""))
        return NULL;
    Dummy_5461 = call_return_simple();
    return Py_BuildValue("l", Dummy_5461);
}

static PyObject * call_return_no_args_wrapper(PyObject *self, PyObject *args)
{

    long Dummy_6232;
    if (!PyArg_ParseTuple(args, ""))
        return NULL;
    Dummy_6232 = call_return_no_args();
    return Py_BuildValue("l", Dummy_6232);
}

static PyObject * call_no_args_no_return_wrapper(PyObject *self, PyObject *args)
{


    if (!PyArg_ParseTuple(args, ""))
        return NULL;
    call_no_args_no_return();
    return Py_BuildValue("");
}

static PyMethodDef test_methods[] = {
    { "simple", simple_wrapper, METH_VARARGS, "" },
    { "no_return", no_return_wrapper, METH_VARARGS, "" },
    { "no_args", no_args_wrapper, METH_VARARGS, "" },
    { "no_args_no_return", no_args_no_return_wrapper, METH_VARARGS, "" },
    { "call_return_simple", call_return_simple_wrapper, METH_VARARGS, "" },
    { "call_return_no_args", call_return_no_args_wrapper, METH_VARARGS, "" },
    { "call_no_args_no_return", call_no_args_no_return_wrapper, METH_VARARGS, "" },
    { NULL, NULL, 0, NULL }
};

static struct PyModuleDef test_module = {
    PyModuleDef_HEAD_INIT,
    "test",   /* name of module */
    NULL,   /* module documentation, may be NULL */
    -1,        /* size of per-interpreter state of the module, or -1 if the module keeps state in global variables. */
    test_methods
};

PyMODINIT_FUNC PyInit_test(void)
{
    PyObject *m;

    m = PyModule_Create(&test_module);
    if (m == NULL)
        return NULL;

    return m;
}
```
should I add more tests in the test folder.

pyccel/codegen/cwrapper.py
```diff
@@ -22,23 +22,33 @@ def write_python_wrapper(expr, printer):
...
+        code += "if (!PyArg_ParseTuple(args, \"\"))\n        return NULL;\n    "
+    else:
+        code += "if (!PyArg_ParseTuple(args, \""
+        code += ''.join(pytype_registry[str_dtype(arg.dtype)] for arg in expr.arguments)
+        code += "\", "
+        code += ', '.join("&" + printer._print(arg) for arg in expr.arguments)
+        code += "))\n        return NULL;\n    "
 
     if len(expr.results)==0:
         func_call = UndefinedFunction(str(expr.name))(*expr.arguments)
```
EmilyBourne, line 36: As described in issue #397 we should not be using `UndefinedFunction` here as it is a sympy object which does not contain information which may be necessary for printing. Please change this line as described in the issue

EmilyBourne:
> this one of the tests that I used
> should I add more tests in the test folder

The tests in the test folder are run whenever something new is added to pyccel. It is therefore important that (as much as possible) everything that works has a test. This means that if we change something in pyccel, and it breaks something that was previously working we become aware of it immediately. So whenever you add something new to pyccel you should also add a test to stop someone breaking it accidentally later.

The problem with functions without return arguments though is that it is hard to test whether it did what was expected, as there is no output. They will be tested more effectively once arrays are implemented, however this is not something that we plan to add right now. They could also be tested more effectively once printing is implemented, but this is also not entirely trivial.

In the file `tests/epyccel/test_epyccel_functions.py` there are already some basic tests for functions. I suggest that you activate whichever tests should already work and add the tests you describe above.

To activate the tests for c you will first need to pull a commit from the `activate_c_math_tests` branch. You can do this easily with the command:
```
git cherry-pick 4460fb8554b4450bf95aadd9e7ad856e518d4203
```

You can then activate the tests easily by adding an argument called `language` to all the tests that need to be in multiple languages and passing the language to the epyccel command.

It may also be worth adding a test or 2 to ensure that a `TypeError` is raised if you pass arguments to a function that doesn't have any. I think that this is what should happen when you call `PyArg_ParseTuple(args, "")` so this test would help us feel more confident that we are definitely not returning `NULL`

saidctb:
as @Emily described , it needs to be in our test case which can be run in Travis https://github.com/pyccel/pyccel/pull/400/checks?check_run_id=1027211073 , so we can be sure that we don't brake them in the future

pyccel/codegen/cwrapper.py
```diff
@@ -22,23 +22,34 @@ def write_python_wrapper(expr, printer):
...
+        code += ''.join(pytype_registry[str_dtype(arg.dtype)] for arg in expr.arguments)
+        code += "\", "
+        code += ', '.join("&" + printer._print(arg) for arg in expr.arguments)
+        code += "))\n        return NULL;\n    "
 
     if len(expr.results)==0:
-        func_call = UndefinedFunction(str(expr.name))(*expr.arguments)
+        func_call = FunctionCall(expr, expr.arguments)
+
+        code += printer._print(func_call) + ';'
```
EmilyBourne, line 38: This additional semi-colon is no longer necessary

tests/epyccel/test_epyccel_functions.py
```diff
@@ -13,6 +13,90 @@ def clean_test():
...
+    assert free_gift() == c_gift()
+
+def test_func_no_args_return_type_1(language):
+    '''test return type of a function with negative return value but no args'''
+    def free_gift():
+        gift = 10
+        return gift
+
+    c_gift = epyccel(free_gift, language=language)
+    assert isinstance(free_gift(), type(c_gift()))
```
EmilyBourne, line 32: The type and value can be tested in the same function `test_func_no_args_1` to avoid translating the function `free_gift` twice. The error message will be sufficient to tell which assertion has caused the failure

tests/epyccel/test_epyccel_functions.py
```diff
@@ -13,6 +13,90 @@ def clean_test():
...
+    assert p_lose() == c_lose()
+
+def test_func_no_args_return_type_2(language):
+    '''test return type of a function with negative return value but no args'''
+    def p_lose():
+        lose = -10
+        return lose
+
+    c_lose = epyccel(p_lose, language=language)
+    assert isinstance(p_lose(), type(c_lose()))
```
EmilyBourne, line 50: See above, these 2 functions can be merged into 1

EmilyBourne (changes requested):

EmilyBourne (approved):

tests/epyccel/test_epyccel_functions.py
```diff
@@ -13,6 +13,57 @@ def clean_test():
...
+    def p_lose():
+        lose = -10
+        return lose
+
+    c_lose = epyccel(p_lose, language=language)
+    assert p_lose() == c_lose()
+    assert isinstance(p_lose(), type(c_lose()))
+    unexpected_arg = 0
+    with pytest.raises(TypeError):
+        p_lose(unexpected_arg)
```
EmilyBourne, line 40: This should be `c_lose`

EmilyBourne (approved):

tests/epyccel/test_epyccel_functions.py
rakati: yes, the names were a little bit similar. I think I need to change my way in naming

EmilyBourne: Given that they essentially represent the same thing it's not really surprising that the names are so close. I think there are several cases like this in the tests folder. I wouldn't worry about it too much. This is why we have a review process for pull requests. The more people look at the code before it is merged into the master branch, the more likely we are to catch this kind of thing

rakati: yes, so I think I need to request more reviewers like my friends from 1337, it's an occasion for them to learn more about pyccel.

pyccel/codegen/printing/ccode.py
```diff
@@ -175,7 +186,11 @@ def _print_PyccelAdd(self, expr):
         return ' + '.join(self._print(a) for a in expr.args)
 
     def _print_PyccelMinus(self, expr):
-        return ' - '.join(self._print(a) for a in expr.args)
+        args = [self._print(a) for a in expr.args]
+
+        if len(args) == 1:
+            return '-{}'.format(args[0])
+        return ' - '.join(a for a in args)
```
yguclu, line 193: Why not just `return ' - '.join(args)`?

Could you give me an example where `PyccelMinus` has more than one argument?

pyccel/codegen/cwrapper.py
```diff
@@ -22,23 +22,32 @@ def write_python_wrapper(expr, printer):
     results_decs = '\n    '.join(printer._print(i) for i in results_decs)
     code += '{0}\n    {1}\n    '.format(arg_decs, results_decs)
 
-    code += "if (!PyArg_ParseTuple(args, \""
-    code += ''.join(pytype_registry[str_dtype(arg.dtype)] for arg in expr.arguments)
-    code += "\", "
-    code += ', '.join("&" + printer._print(arg) for arg in expr.arguments)
-    code += "))\n        return NULL;\n    "
+    # Check if the function with No arguments
```
yguclu, line 25: Did you mean "has no" rather than "with No"?

pyccel/codegen/printing/ccode.py
EmilyBourne: Do you mean where `PyccelMinus` has one argument? The example is in issue #399
The unary operator `-x` is translated to `PyccelMinus(PyccelUnary(x))`

yguclu: I mean a situation where `len(args)>1`, which triggers line 193.

EmilyBourne: Then `x - y` = PyccelMinus(x,y)

EmilyBourne: Sorry I didn't spot the difference between
`return ' - '.join(args)` and
`return ' - '.join(a for a in args)`

yguclu: OK, so the `args` tuple can either contain 1 or 2 elements here. Never 3 I suppose? I just find the code a little strange, as they way it is written makes one think that `args` could have an arbitrary length.

yguclu (approved):

pyccel/codegen/printing/ccode.py
EmilyBourne: No never 3 for a binary operator (boolean operators on the other hand can contain as many elements as they want despite not appearing significantly different in construction)

yguclu:
Good job @rakati! We'll merge your PR after #404.

pyccel/codegen/printing/ccode.py
yguclu: In the official Python documentation an explicit distinction is made between unary and binary math operators. Here `PyccelMinus` covers both, is this the same in Python's `ast`?

EmilyBourne: No. Python's `ast` has a `BinaryOp` class and a `UnaryOp` class. Each of these has an operator eg Add, Sub, UAdd, USub, Not, etc.
Pyccel's treatment of Unary operators is a bit odd:
```
        if isinstance(stmt.op, ast.Not):
            Func = PyccelNot
        elif isinstance(stmt.op, ast.UAdd):
            Func = PyccelUnary
        elif isinstance(stmt.op, ast.USub):
            Func = PyccelMinus

        return Func(PyccelUnary(target))
```
I'm not sure why the second `PyccelUnary` is necessary other than to ensure that the correct precedence is used. I think it would be better to have a `PyccelUnaryMinus`. I also don't think that Not should be using `PyccelUnary` as it has a much lower precedence. This function certainly wants looking at. But it is almost certainly beyond the scope of this PR

yguclu: Yes, this is beyond the scope of this PR (in fact I am keeping the discussion as "resolved")... but it's probably useful to keep track of our conversation here. Maybe we just need to create a `PyccelUnaryMinus` class, but we could also mimic Python's `ast` and have `PyccelBinaryOp` and `PyccelUnaryOp`.

EmilyBourne: Yes, that is also a possibility. It is currently written this way as it mimics sympy (although sympy doesn't have Minus or Divide). The current method makes specifying the precedence easy though.

Yes we should create an issue

PR #404, opened 2020-08-26
Avoid name collisions for temporary variables and modules

### LINKED ISSUE #403: Variable name collisions not always handled causing travis problems
**Describe the bug**
Variables created in `codegen/printing/fcode.py` (usually output variables) are not checked for name collisions. As the random string has only 9999 different possibilities this occasionally results in 2 objects having the same name. As a result the function returns one item less than it should. This is why our tests on travis fail occasionally. See PR #401 

**To Reproduce**

The behaviour can be reproduced by decreasing the randomness of the name generator. I.e by changing the function `create_random_string` in `ast/core.py` from:
```python
def create_random_string(expr):
    import numpy as np
    try:
        randstr = str(abs(hash(expr)
                                  + np.random.randint(500)))[-4:]
    except TypeError:
        # Catch unhashable type (e.g. list, FunctionalSum)
        randstr = str(abs(np.random.randint(10000)))[-4:]
    return randstr
```
to:
```
def create_random_string(expr):
    import numpy as np
    return str(np.random.randint(10))
```

For example the following code:
```python
def functional_for_1d_range():
    a = [i for i in range(4)]
    return len(a), a[0], a[1], a[2], a[3]
```
is then translated to:
```fortran
module test

implicit none

contains

subroutine functional_for_1d_range(Dummy_5, Dummy_9, Dummy_0)

  implicit none

  integer(kind=8), intent(out) :: Dummy_5
  integer(kind=8), intent(out) :: Dummy_9
  integer(kind=8), intent(out) :: Dummy_0
  integer(kind=8) :: i
  integer(kind=8), allocatable :: a(:)
  integer(kind=8) :: Dummy_3

  allocate(a(0:4_8 - 1_8))
  Dummy_3 = 0_8
  do i = 0_8, 4_8-1_8, 1_8
    a(Dummy_3) = i
    Dummy_3 = Dummy_3 + 1_8

  end do
  Dummy_5 = size(a,1)
  Dummy_5 = a(0_8)
  Dummy_9 = a(1_8)
  Dummy_9 = a(2_8)
  Dummy_0 = a(3_8)
  return

end subroutine functional_for_1d_range

end module test
```

**Expected behavior**
We should modify the function `create_random_string` so it takes a list of forbidden values

**Language**
Fortran
The generated names are checked in parser/semantic.py however the process is not always done very neatly and could be tidied up while fixing this

**Additional context**
`parser/semantic.py` has the following 2 functions to deal with this problem:
```
        if current_name not in self._possible_names:
            self._possible_names.add(current_name)
            return current_name
        current_name = current_name + '_' + create_random_string(current_name)
        while current_name in self._possible_names:
            # Generate random name based on the original name
            current_name = current_name + create_random_string(current_name)
        self._possible_names.add(current_name)
        return current_name

    def _get_new_variable(self, obj):
        var = create_variable(obj)
        name = var.name
        while name in self._possible_names:
            var = create_variable(obj)
            name = var.name
        self._possible_names.add(name)
        return var
```
It should not be necessary to have the while loop in multiple locations. In addition `_get_new_variable` is often used when `_get_new_name` should be used

EmilyBourne:
**Fix #403: Variable name collisions not always handled causing travis problems.**

Rewrite `create_random_string` and `create_variable` functions to generate deterministic names following an incrementing pattern. Ensure that the SyntacticParser, the `SemanticParser` and the `Codegen` all use the same function to access these functions to handle counter simply

- If epyccel generated import name is in the list of imported modules, find a new name
- Remove _print_Function function. There is no Function class
- Move name and variable creation functions to BaseParser : `get_new_name`, `get_new_variable`
- Rewrite `create_random_string` and `create_variable` functions to generate deterministic names following an incrementing pattern
These functions take a prefix and a counter. The prefix defaults to 'Dummy'
- Ensure that the SyntacticParser, the `SemanticParser` and the `Codegen` all use  `get_new_name` and `get_new_variable` to create names and variables
- Create a counter for Dummy Variables. Pass this information from `SyntaxParser` to `SemanticParser`
- Save list of used names in BaseParser. Pass this information from `SyntaxParser` to `SemanticParser`
- Update `pyccel_to_sympy` function to pass arguments that are now necessary
- Create a temporary copy of the set of used names to pass to pyccel_to_sympy as it creates temporary variables which do not need to be avoided later
- Use `create_variable` to create output variables in `SyntaxParser`. Use a different prefix `Out` and a separate counter
- Correct function name `test_functional_for_2d_var_var_range` -> `test_functional_for_2d_var_range`
- Rename `create_random_string` ->  `create_incremented_string`

### DIFF
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -819,9 +819,12 @@ def int2float(expr):
 def float2int(expr):
     return expr
 
-def create_random_string(forbidden_exprs : set, prefix = 'Dummy', nDigits : int = 2):
+def create_random_string(forbidden_exprs : set, prefix = None, nDigits : int = 2):
     assert(isinstance(forbidden_exprs, set))
     import numpy as np
+
+    if prefix is None:
+        prefix = 'Dummy'
 
     max_val = 10**nDigits
 
@@ -832,10 +835,10 @@ def create_random_string(forbidden_exprs : set, prefix = 'Dummy', nDigits : int 
         name = prefix + str(np.random.randint(max_val))
     return name
 
-def create_variable(forbidden_names):
+def create_variable(forbidden_names, prefix = None):
     """."""
 
-    name = create_random_string(forbidden_names)
+    name = create_random_string(forbidden_names, prefix)
 
     return Symbol(name)
 
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -819,11 +819,13 @@ def int2float(expr):
 def float2int(expr):
     return expr
 
-def create_random_string(forbidden_exprs : set, prefix = 'Dummy_', nDigits : int = 1):
+def create_random_string(forbidden_exprs : set, prefix = 'Dummy', nDigits : int = 1):
     assert(isinstance(forbidden_exprs, set))
     import numpy as np
 
     max_val = 10**nDigits
+
+    prefix += '_'
 
     name = prefix + str(np.random.randint(max_val))
     while name in forbidden_exprs:
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -882,7 +882,7 @@ def create_variable(forbidden_names, prefix = None, counter = 1):
 
       Returns
       ----------
-      name            : Symbol
+      name            : sympy.Symbol
                         A sympy Symbol with the incremented string name
       counter         : int
                         The expected value of the next name
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -156,7 +156,7 @@ __all__ = (
     '_atomic',
 #    'allocatable_like',
     'create_variable',
-    'create_random_string',
+    'create_incremented_string',
     'extract_subexpressions',
 #    'float2int',
     'get_assigned_symbols',
@@ -819,7 +819,7 @@ def int2float(expr):
 def float2int(expr):
     return expr
 
-def create_random_string(forbidden_exprs, prefix = 'Dummy', counter = 1):
+def create_incremented_string(forbidden_exprs, prefix = 'Dummy', counter = 1):
     assert(isinstance(forbidden_exprs, set))
     import numpy as np
     nDigits = 4
@@ -841,7 +841,7 @@ def create_random_string(forbidden_exprs, prefix = 'Dummy', counter = 1):
 def create_variable(forbidden_names, prefix = None, counter = 1):
     """."""
 
-    name, counter = create_random_string(forbidden_names, prefix, counter = counter)
+    name, counter = create_incremented_string(forbidden_names, prefix, counter = counter)
 
     return Symbol(name), counter
 
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -822,7 +822,7 @@ def float2int(expr):
 def create_random_string(forbidden_exprs, prefix = 'Dummy', counter = 1):
     assert(isinstance(forbidden_exprs, set))
     import numpy as np
-    nDigits = 2
+    nDigits = 4
 
     if prefix is None:
         prefix = 'Dummy'
@@ -842,7 +842,6 @@ def create_variable(forbidden_names, prefix = None, counter = 1):
     """."""
 
     name, counter = create_random_string(forbidden_names, prefix, counter = counter)
-    print(counter)
 
     return Symbol(name), counter
 
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -819,7 +819,7 @@ def int2float(expr):
 def float2int(expr):
     return expr
 
-def create_random_string(forbidden_exprs : set, prefix = '', nDigits : int = 1):
+def create_random_string(forbidden_exprs : set, prefix = 'Dummy_', nDigits : int = 1):
     assert(isinstance(forbidden_exprs, set))
     import numpy as np
 
@@ -833,7 +833,7 @@ def create_random_string(forbidden_exprs : set, prefix = '', nDigits : int = 1):
 def create_variable(forbidden_names):
     """."""
 
-    name = create_random_string(forbidden_names, 'Dummy_')
+    name = create_random_string(forbidden_names)
 
     return Symbol(name)
 
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -819,22 +819,21 @@ def int2float(expr):
 def float2int(expr):
     return expr
 
-def create_random_string(expr):
+def create_random_string(forbidden_exprs : set, prefix = '', nDigits : int = 1):
+    assert(isinstance(forbidden_exprs, set))
     import numpy as np
-    return str(np.random.randint(10))
-    try:
-        randstr = str(abs(hash(expr)
-                                  + np.random.randint(500)))[-4:]
-    except TypeError:
-        # Catch unhashable type (e.g. list, FunctionalSum)
-        randstr = str(abs(np.random.randint(10000)))[-4:]
-    return randstr
 
-def create_variable(expr):
+    max_val = 10**nDigits
+
+    name = prefix + str(np.random.randint(max_val))
+    while name in forbidden_exprs:
+        name = prefix + str(np.random.randint(max_val))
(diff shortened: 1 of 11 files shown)


EmilyBourne:
@yguclu I have pushed an empty commit to prove that this branch compiles without errors twice in a row. I can push one more empty commit if you would like but I'm not sure there is any point given that I managed to pinpoint the bug so we shouldn't be guessing about the fix this time

pyccel/ast/core.py
```diff
@@ -819,23 +819,31 @@ def int2float(expr):
 def float2int(expr):
     return expr
 
-def create_random_string(expr):
+def create_random_string(forbidden_exprs, prefix = 'Dummy', counter = 1):
```
yguclu, line 822: The output string is not "random" anymore, correct? 😉

EmilyBourne: I have renamed it `create_incremented_string`

saidctb (approved):

pyccel/ast/core.py
```diff
@@ -819,23 +819,31 @@ def int2float(expr):
 def float2int(expr):
     return expr
 
-def create_random_string(expr):
+def create_incremented_string(forbidden_exprs, prefix = 'Dummy', counter = 1):
+    assert(isinstance(forbidden_exprs, set))
     import numpy as np
```
yguclu, line 824: We don't need Numpy anymore, correct?

pyccel/ast/core.py
```diff
@@ -819,23 +819,31 @@ def int2float(expr):
 def float2int(expr):
     return expr
 
-def create_random_string(expr):
+def create_incremented_string(forbidden_exprs, prefix = 'Dummy', counter = 1):
```
yguclu, line 822: Can you add a docstring that explains what the arguments are?

pyccel/ast/core.py
```diff
@@ -819,23 +819,31 @@ def int2float(expr):
...
+    counter += 1
+    while name in forbidden_exprs:
+        name = name_format.format(prefix=prefix, counter = counter)
+        counter += 1
+
+    forbidden_exprs.add(name)
+
+    return name, counter
+
+def create_variable(forbidden_names, prefix = None, counter = 1):
```
yguclu, line 841: Here too, please add a dosctring!

pyccel/ast/sympy_helper.py
```diff
@@ -86,7 +86,7 @@ def sympy_to_pyccel(expr, symbol_map):
         raise TypeError(str(type(expr)))
 
 #==============================================================================
-def pyccel_to_sympy(expr, symbol_map):
+def pyccel_to_sympy(expr, symbol_map, used_names):
     """
     Convert a sympy expression to a pyccel expression replacing sympy symbols with
     pyccel expressions provided in a symbol_map
```
yguclu, line 92: Please add the "Parameters" and "Returns" sections to this docstring

pyccel/codegen/printing/fcode.py
```diff
@@ -1613,7 +1613,7 @@ def _do_range(target, iterable, prolog, epilog):
         body = self._print(expr.body)
 
         return ('{prolog}'
-                '{body}\n'
+                '{body}'
```
yguclu, line 1616: Thanks for fixing this!

pyccel/epyccel.py
```diff
@@ -18,10 +18,10 @@
 #==============================================================================
 random_selector = random.SystemRandom()
 
-def random_string( n ):
+def random_string( n, prefix = '' ):
     # we remove uppercase letters because of f2py
     chars    = string.ascii_lowercase + string.digits
-    return ''.join( random_selector.choice( chars ) for _ in range(n) )
+    return prefix + ''.join( random_selector.choice( chars ) for _ in range(n) )
```
yguclu, line 24: Was it really necessary to change this function? I see no advantage when you use it.

pyccel/epyccel.py
```diff
@@ -63,8 +63,11 @@ def epyccel_seq(function_or_module,
     if isinstance(function_or_module, FunctionType):
         pyfunc = function_or_module
         code = get_source_function(pyfunc)
-        tag = random_string(8)
-        module_name = 'mod_{}'.format(tag)
+
+        module_name = random_string(prefix='mod_', n=8)
+        while module_name in sys.modules.keys():
+            module_name = random_string(prefix='mod_', n=8)
```
yguclu, line 69: As a follow up to my previous comment, with the original `random_string` function this would read
```
module_name = 'mod_' + random_string(8)
while module_name in sys.modules.keys():
    module_name = 'mod_' + random_string(8)
```
which is totally equivalent and just as clear.

pyccel/parser/base.py
```diff
@@ -231,6 +232,7 @@ def __init__(self,
         self._metavars  = OrderedDict()
         self._namespace = Scope()
 
+        self._used_names = None
```
yguclu, line 235: Given that you changed this class, I think this is a good opportunity to update its docstring, which has not been updated in a while!

pyccel/parser/base.py
```diff
@@ -334,6 +338,28 @@ def blocking(self):
     def show_traceback(self):
         return self._show_traceback
 
+    @property
+    def used_names(self):
```
yguclu, line 342: Please add a one-line docstring

pyccel/parser/base.py
```diff
@@ -334,6 +338,28 @@ def blocking(self):
...
+        return self._used_names
+
+    def get_new_name(self, current_name = None):
```
yguclu, line 345: Please add a complete docstring here

pyccel/parser/base.py
```diff
@@ -334,6 +338,28 @@ def blocking(self):
...
+            self._used_names.add(current_name)
+            return current_name
+
+        if current_name is not None:
+            new_name, self._dummy_counter = create_incremented_string(self._used_names, prefix = current_name, counter = self._dummy_counter)
+        else:
+            new_name,_ = create_incremented_string(self._used_names, prefix = current_name)
+        return new_name
+
+    def get_new_variable(self, prefix = None):
```
yguclu, line 356: Here too

yguclu (changes requested):
Great job @EmilyBourne! I would like to see just a few minor changes/additions (see comments).

pyccel/epyccel.py
EmilyBourne: No sorry, I don't remember if this was just to have everything on one line or if I just wanted to be able to determine the whole module name by fiddling with this function while I was debugging

pyccel/parser/base.py
EmilyBourne: I am not sure exactly what to add to this docstring. I have added a short description

yguclu: The docstring should be removed from the `__init__` and added to the class docstring, and we should add a description for `output_folder`...

EmilyBourne: The `output_folder` property does not seem to be used. I will remove it

pyccel/parser/base.py
yguclu: "Parameters" and "Returns" section please?

pyccel/parser/base.py
yguclu: See above! We'll create some automatic documentation for developers at some point, so...

pyccel/parser/base.py
yguclu: Great, thanks!

pyccel/parser/base.py
```diff
@@ -201,15 +202,10 @@ def loops(self):
...
+    This class contains functions and properties which are common to SyntacticParser and SemanticParser
 
-    def __init__(self,
-                 debug=False,
-                 headers=None,
-                 static=None,
-                 show_traceback=False,
-                 output_folder=''):
-        """Parser constructor.
+    Parser constructor:
```
yguclu, line 208: Here you should just write
```
Parameters
----------
```
see https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring.

yguclu (approved):

Issue #388, opened 2020-08-19
Pyccel exits without producing a readable error when using a non-implemented data type in C


EmilyBourne:
Pyccel exits without producing a readable error when using a non-implemented data type in C. This is due to the use of the dictionary `dtype_registry` in `pyccel/codegen/printing/ccode.py`.
E.g: the following code:
```
from pyccel.decorators import types

@types('bool')
def f(x):
    return x
```
produces the error:
```
Traceback (most recent call last):
  File "/home/emily/.local/bin/pyccel", line 11, in <module>
    load_entry_point('pyccel==0.9.11', 'console_scripts', 'pyccel')()
  File "/home/emily/.local/lib/python3.6/site-packages/pyccel/commands/console.py", line 251, in pyccel
    folder        = args.output)
  File "/home/emily/.local/lib/python3.6/site-packages/pyccel/codegen/pipeline.py", line 170, in execute_pyccel
    fname = codegen.export(fname, language=language)
  File "/home/emily/.local/lib/python3.6/site-packages/pyccel/codegen/codegen.py", line 258, in export
    code = self.doprint(**settings)
  File "/home/emily/.local/lib/python3.6/site-packages/pyccel/codegen/codegen.py", line 247, in doprint
    code = printer(self.expr, parser=self.parser, **settings)
  File "/home/emily/.local/lib/python3.6/site-packages/pyccel/codegen/printing/ccode.py", line 421, in ccode
    return CCodePrinter(settings).doprint(expr, assign_to)
  File "/home/emily/.local/lib/python3.6/site-packages/pyccel/codegen/printing/codeprinter.py", line 60, in doprint
    lines = self._print(expr).splitlines()
  File "/home/emily/.local/lib/python3.6/site-packages/sympy/printing/printer.py", line 287, in _print
    return getattr(self, printmethod)(expr, **kwargs)
  File "/home/emily/.local/lib/python3.6/site-packages/pyccel/codegen/printing/ccode.py", line 105, in _print_Module
    return '\n\n'.join(self._print(i) for i in expr.body)
  File "/home/emily/.local/lib/python3.6/site-packages/pyccel/codegen/printing/ccode.py", line 105, in <genexpr>
    return '\n\n'.join(self._print(i) for i in expr.body)
  File "/home/emily/.local/lib/python3.6/site-packages/sympy/printing/printer.py", line 287, in _print
    return getattr(self, printmethod)(expr, **kwargs)
[...]

Issue #386, opened 2020-08-18
Complex numbers in C 


MenoIy:
**Some data types don't work as argument ex {complex complex64 complex128} exemple :**

```
@types('complex')
def func(x):
    return x
```

```
ERROR at code generation stage
**pyccel**: [codegen]
    |fatal| Pyccel has encountered syntax that has not been implemented yet
```

EmilyBourne:
Hello,

It looks like you are reporting 2 issues here:

1. Not all datatypes are supported as arguments in C
2. Casting (to both native datatypes and numpy datatypes) does not work in C

In order to avoid confusion in any necessary discussion about each problem it is better to write separate issues for separate problems. This also allows the issues to be closed separately if they are not solved in the same pull request.
The fact that `float` is a valid argument but cannot be used for casting shows that solving one problem does not necessarily fix the other

MenoIy:
For the issue: **Not all datatypes are supported as arguments in C** I don't know if it's a good idea to use the  Py_complex  from <Python.h> to read arguments type complex  exemple :  

a python code like  : 
```
from pyccel.decorators import types
@types('complex')
def func(y):
     return y
```

will be something like this
```
#include <complex.h>
Py_complex Cfunc(Py_complex x)
{
        return(x);
}
static PyObject* func(PyObject* self, PyObject* args)
{
        Py_complex n;
        if (!PyArg_ParseTuple(args, "D", &n))
                return  NULL;
        Py_complex test = Cfunc(n);
        PyObject* result = Py_BuildValue("D", &test);
        return result;
}
```
but to work with this we need to include <Python.h> in both  c file and add some condition to **pyccel/codegen/cwrapper.py**   to print  **'&'**  when we are working with py_complex type in    **Py_BuildValue().**

EmilyBourne:
This special case of complex numbers needs discussing. Please split this issue into 3 by creating the following 2 new issues:
1. Some data types don't work as arguments in C ex (int8 int16 bool)
2. Casting does not work in C

This issue can then be used to discuss "Complex numbers in C"

You can edit your original message and copy and paste the relevant part of the description into the new issues. Don't forget to add the `Language:C` label

EmilyBourne:
> to work with this we need to include <Python.h> in both c file and add some condition to pyccel/codegen/cwrapper.py to print '&' when we are working with py_complex type in Py_BuildValue().

[...]

PR #408, opened 2020-08-31
Add native cast c

### LINKED ISSUE #389: Casting (to both native datatypes and numpy datatypes) does not work in C
**Casting to native data type doesn't work example :**

```
x = float(32)
y = int(32.0)
z = complex(32)
```

```
**pyccel**: [codegen] 
   |fatal| Pyccel has encountered syntax that has not been implemented yet
```

**casting to numpy data type too :**

```
w = numpy.float64(15)
```

### LINKED ISSUE #390: Some data types don't work as arguments in C ex (int8 int16 bool)

MenoIy:
Fix  precision in int
by correcting  precision of some data type and using the rights flag of PyArg_ParseTuple and Py_BuildValue

Fix #389 
Adding print function for cast opération  float()   int()     bool()


tests/epyccel/test_base.py
```diff
@@ -69,6 +69,11 @@ def test_compare_is_not_nil():
 
 def test_cast_int():
     compare_epyccel(base.cast_int, 4)
+    compare_epyccel(base.cast_float_to_int, 4.5)
```
EmilyBourne, line 72: As you are testing casting, it would be good to also add a test to show that the return type is correct (see https://github.com/pyccel/pyccel/blob/master/tests/epyccel/recognised_functions/test_math_funcs.py#L41)

EmilyBourne (changes requested):
Tests are not activated automatically for c. To test your code please activate the tests which you think should be working (probably all the tests in tests/epyccel/test_base.py, maybe others too).

In order to activate the tests, you need to add an argument called `language` to all the tests that need to be in multiple languages and then pass `language` to the epyccel command (see tests/epyccel/test_epyccel_functions.py for examples)

EmilyBourne:
There are also additional tests in other files which use the datatypes that you have implemented. These tests can also be activated. There are some in each of these files:
- tests/epyccel/test_epyccel_functions.py
- tests/epyccel/test_kind.py
- tests/epyccel/test_multiple_results.py

MenoIy:
I just activated  language c in test_base.py for  some tests,  and for the others i found new issue #409

tests/epyccel/test_base.py
```diff
@@ -5,6 +5,13 @@
 from modules import base
 from conftest       import *
 
+#To delete after solve issue (is and is not not working in c) 
+def tmp_compare_epyccel(language, f, *args):
```
EmilyBourne, line 9: You do not need to have 2 `compare_epyccel` functions. You can give the same result either by providing a default value for language (which will then have to be called using the keyword):
```
def compare_epyccel(f, *args, language='fortran'):
```
or by marking the failing tests as xfail:
```
@pytest.mark.parametrize("language",['fortran',pytest.param('c', marks = pytest.mark.xfail(reason="Issue 409: 'is'/'is not' not implemented in c")
def test_is_false(language):
    compare_epyccel(base.is_false, True, language=language)
    compare_epyccel(base.is_false, False, language=language)
```

tests/epyccel/test_kind.py
```diff
@@ -40,10 +41,39 @@ def add_real(a, b):
...
+    assert(type(cast_to_int(5.2)) == type(f(5.2)))
+
+def test_output_types_2(language):
+    @types('int')
+    def cast_to_float(a):
+        b = float(a)
+        return b
+
+    f = epyccel(cast_to_float,language= language)
+    assert(type(cast_to_float(5)) == type(f(5)))
```
EmilyBourne, line 64: This type of check raises an error in Codacy, although it makes sense in this context (comparing types of objects is rare, usually you test if the type is something specific, so codacy is not built for this). You can disable the codacy test for this line by adding the following comment to the end of the line:
```
 # pylint: disable=unidiomatic-typecheck
```

EmilyBourne (changes requested):

pyccel/codegen/printing/ccode.py
```diff
@@ -63,7 +63,9 @@
                   ('complex',4) : 'float complex',
                   ('int',4)     : 'int',
                   ('int',8)     : 'long',
-                  ('bool',1)    : 'int'}
+                  ('int',2)     : 'int',
+                  ('int',1)     : 'int',
+                  ('bool',4)    : 'int'}
```
EmilyBourne, line 68: Would you mind fixing issue #388 too? As we are using a dictionary we should either check whether the key we are looking for is in the dictionary or use a try/except block to catch the `KeyError`.
This will be needed here:
https://github.com/pyccel/pyccel/blob/f20898be8d057448d2c41dd5eb1d3ec7027d2cae/pyccel/codegen/printing/ccode.py#L201
and possibly in other places too.

When the error has been caught/when the key is not in the dictionary we need to call the following code:
```
errors.report(PYCCEL_RESTRICTION_TODO, symbol=expr,
                severity='fatal')
```

EmilyBourne (changes requested):
Not all tests in tests/epyccel/test_base.py use the `is` operator. If issue #408 is completely fixed, you should be able to activate:
- def test_eq_false
- def test_eq_true
- def test_neq_false
- def test_neq_true
- def test_not

I suspect that this will not work however, as I don't see the code, that you suggested for boolean return types, implemented:
>```pytype_printer_registry={
>    'bool' : 'PyBool_FromLong({})',
>    'int' : '{}',
>    'double' : '{}'
>}
> code = ", ".join(pytype_printer_registry['str_dtype(x)].format(x)  for x  in expr list' )```

pyccel/codegen/printing/ccode.py
```diff
@@ -105,6 +107,18 @@ def _traverse_matrix_indices(self, mat):
...
+        value = self._print(expr.arg)
+        return '(double)({0})'.format(value)
+
+    def _print_Int(self, expr):
+        value = self._print(expr.arg)
+        return '(long)({0})'.format(value)
+
+    def _print_Bool(self, expr):
+        value = self._print(expr.arg)
+        return '{} ? 1 : 0'.format(value)
```
EmilyBourne, line 120: We should differentiate between different types. E.g.:
**x is bool**
```
x = (a == 2)
if (bool(x)):
```
will give:
```
if (x ? 1 : 0)
```
when
```
if (x)
```
would be sufficient

**x is int**
```
if (bool(x)):
```
will give:
```
if (x ? 1 : 0)
```
when
```
if (x)
```
would be sufficient. Although the former has the advantage of ensuring that a boolean is always one of `0` and `1` so it may be clearer to give:
```
if (x != 0)
```
Do you have opinions on the better formulation @saidctb @yguclu ?

nhamidn, line 120: Why not just return `self._print(expr.arg)`

yguclu: C99 provides the <stdbool.h> header with the `bool` data type, and `true` and `false` constants. Maybe using them would add a bit of clarity to the printed code?

In the former case I would write
```C
bool x = (a == 2);

if (x)
{
  ...
}
```
and in the latter case I would not use implicit casting to bool in order to avoid unpleasant surprises:
```C
if (x != 0)
{
  ...
}
```

saidctb: the suggested solution of @yguclu seems good to me

pyccel/codegen/printing/ccode.py
MenoIy: I didn,t use the suggested print for boolean value 
```
   'bool' : 'PyBool_FromLong({})',
   'int' : '{}',
   'double' : '{}'
}
code = ", ".join(pytype_printer_registry['str_dtype(x)].format(x)  for x  in expr list' )
```
because  im working in complex datatypes too so i wanted to look for a  general code form to print both complpex and boolean type.

EmilyBourne:
> > Not all tests in tests/epyccel/test_base.py use the `is` operator. If issue #408 is completely fixed, you should be able to activate:
> > 
> > * def test_eq_false
> > * def test_eq_true
> > * def test_neq_false
> > * def test_neq_true
> > * def test_not
> > 
> > I suspect that this will not work however, as I don't see the code, that you suggested for boolean return types, implemented:
> > 
> > > ```
> > >    'bool' : 'PyBool_FromLong({})',
> > >    'int' : '{}',
> > >    'double' : '{}'
> > > }
> > > code = ", ".join(pytype_printer_registry['str_dtype(x)].format(x)  for x  in expr list' )```
> > > ```
> 
> I didn,t use the suggested print for boolean value
> 
>    'bool' : 'PyBool_FromLong({})',
>    'int' : '{}',
>    'double' : '{}'
> }
> code = ", ".join(pytype_printer_registry['str_dtype(x)].format(x)  for x  in expr list' )
> because im working in complex datatypes too so i wanted to look for a general code form to print both comlpex and boolean > type.

Ok. It seems that in python:
```
In [1]: 1==True
Out[1]: True

In [2]: 0==False
Out[2]: True
```
so this will do for now (especially as the use of f2py means that the fortran code does not return bool types yet either)

EmilyBourne:
This branch is currently failing as `PyParse_ArgTuple` cannot collect a `_Bool`/`bool` value (https://stackoverflow.com/a/10700727/5957195)
As a result we will need to collect an int and then convert it to a bool. This is similar to the approach that will be used for complexes (see issue #386 https://github.com/pyccel/pyccel/issues/386#issuecomment-675999987).

It seems that this issue is being bogged down by lots of boolean problems. I suggest that we let this branch fix issues #388 and #389 and be a partial fix for #390 by fixing `int8` and `int16` (this will require an additional test as currently the only test using these types also uses bools). And then booleans as arguments/return types can be fixed in the same PR as complexes given that both will use the same changes in cwrapper.

yguclu:
I agree with @EmilyBourne. Let's split this PR into two separate PRs, where one is ready for merging and the other should be a "draft" (https://github.blog/2019-02-14-introducing-draft-pull-requests/)

pyccel/codegen/cwrapper.py
```diff
@@ -4,13 +4,18 @@
...
-        }
+    ('real',8)    : 'd',
+    ('real',4)    : 'f',
+    ('complex',8) : 'D',
+    ('complex',4) : 'D',
+    ('int',4)     : 'i',
+    ('int',8)     : 'l',
+    ('int',2)     : 'i',
+    ('int',1)     : 'i',
+    ('bool',4)    : 'i'
```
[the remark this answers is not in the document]
MenoIy: Ah, that's true, I truly forgot to put back str to the new register beqause I wanted to see how string are stored in pyccel ast .
And for the bool arg there is two ways to parse python bool either by collecting as pyobject and checking it and converting it to int . Or by  reading as int directly.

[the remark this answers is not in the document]
EmilyBourne: Yes, but from python3.3, reading it directly should be done with the `p` format (4th option in the link I provided). See https://stackoverflow.com/a/10700727/5957195
This also has the advantage of allowing python to convert a bool-like object to a bool. See https://bugs.python.org/issue14705

pyccel/codegen/printing/ccode.py
```diff
@@ -223,7 +242,10 @@ def function_signature(self, expr):
             arg_code = 'void'
         else:
             arg_dtypes = [self._print(i.dtype) for i in expr.arguments]
-            arg_dtypes = [dtype_registry[(dtype, arg.precision)] for dtype,arg in zip(arg_dtypes, expr.arguments)]
+            try :
+                arg_dtypes = [dtype_registry[(dtype, arg.precision)] for dtype,arg in zip(arg_dtypes, expr.arguments)]
+            except :
+                errors.report(PYCCEL_RESTRICTION_TODO, symbol=expr,severity='fatal')
```
yguclu, line 248: See my previous comment.

yguclu (changes requested):

tests/epyccel/test_epyccel_functions.py
```diff
@@ -286,10 +286,60 @@ def g(x, y):
     f = epyccel(g)
     assert f(2,1) == g(2,1)
 
+
+def test_decorator_f15(language):
+    @types('bool', 'int8', 'int16', 'int32', 'int64')
```
EmilyBourne, line 291: This test won't work in this branch

pyccel/codegen/cwrapper.py
EmilyBourne, line 17: Please revert the boolean and string format types

EmilyBourne (changes requested):

tests/epyccel/test_epyccel_functions.py
EmilyBourne: (it does, but only because `language` is not passed to `epyccel`. This means that this test is run twice for fortran

pyccel/codegen/cwrapper.py
MenoIy: im using (dtype, prec) dict here can you tell me please default string precision

MenoIy: or should i go back to the previous py_type dict for the moment.

EmilyBourne: https://github.com/pyccel/pyccel/blob/master/pyccel/ast/core.py#L4747
Apparently string precision is set to 0 as it is a nonsense value for this type

EmilyBourne (approved):

saidctb (approved):

pyccel/codegen/printing/ccode.py
```diff
@@ -63,7 +63,9 @@
                   ('complex',4) : 'float complex',
                   ('int',4)     : 'int',
                   ('int',8)     : 'long',
-                  ('bool',1)    : 'int'}
+                  ('int',2)     : 'int',
+                  ('int',1)     : 'int',
```
yguclu, line 67: Can you explain under what circumstances we deal with integers of 1 or 2 bytes? And why do we fall back to the standard 32-bit `int` in C in this case? Do you have any unit tests for this?

tests/epyccel/test_epyccel_functions.py
```diff
@@ -286,10 +286,61 @@ def g(x, y):
...
+    assert f(False, 1, 2, 3, 4)  == f15(False, 1, 2, 3, 4)
+
+
+def test_decorator_f16(language):
+    @types('int16')
+    def f16(a):
+        b = a
+        return b
+    f = epyccel(f16, language=language)
+    assert f(1) == f16(1)
```
yguclu, line 309: The input value here is a standard Python `int`, not a `numpy.int16`.

tests/epyccel/test_epyccel_functions.py
```diff
@@ -286,10 +286,61 @@ def g(x, y):
...
+
+def test_decorator_f17(language):
+    @types('int8')
+    def f17(a):
+        b = a
+        return b
+    f = epyccel(f17, language=language)
+    assert f(1) == f17(1)
```
yguclu, line 317: The input value here is a standard Python `int`, not a `numpy.int8`.

tests/epyccel/test_epyccel_functions.py
```diff
@@ -286,10 +286,61 @@ def g(x, y):
...
+
+def test_decorator_f18(language):
+    @types('int32')
+    def f18(a):
+        b = a
+        return b
+    f = epyccel(f18, language=language)
+    assert f(1) == f18(1)
```
yguclu, line 325: The input value here is a standard Python `int`, not a `numpy.int32`.

tests/epyccel/test_epyccel_functions.py
```diff
@@ -286,10 +286,61 @@ def g(x, y):
...
+
+def test_decorator_f19(language):
+    @types('int64')
+    def f19(a):
+        b = a
+        return b
+    f = epyccel(f19, language=language)
+    assert f(1) == f19(1)
```
yguclu, line 333: This should be correct, because we assume that a Python `int` is equivalent to a `numpy.int64`.

tests/epyccel/test_epyccel_functions.py
```diff
@@ -286,10 +286,61 @@ def g(x, y):
...
+
+
 ##==============================================================================
 ## CLEAN UP GENERATED FILES AFTER RUNNING TESTS
 ##==============================================================================
 #
 #def teardown_module():
 #    clean_test()
-#
+#
```
yguclu, line 346: No newline at end of file.

tests/epyccel/test_kind.py
```diff
@@ -44,6 +45,35 @@ def add_real(a, b):
...
     assert(add_real(1.0,2.0)==epyc_add_real(1.0,2.0))
 
+def test_output_types_1(language):
+    @types('float32')
+    def cast_to_int(a):
+        b = int(a)
+        return b
+
+    f = epyccel(cast_to_int, language = language)
+    assert(type(cast_to_int(5.2)) == type(f(5.2))) # pylint: disable=unidiomatic-typecheck
```
yguclu, line 55: The input argument is a standard Python `float`, not a `numpy.float32`.

yguclu (changes requested):
I have some doubts about how you handle integers and floats of different sizes...

MenoIy:
Yeah, that's true, I really forgot about numpy types in testing, First when I start testing the différent précision of a int without exceeding the max value of each type every thing was working fine but when I start using values that exceed the type max value I got some OverflowError.
   So when I wanted to know how the python manage integer and I saw that for python Integers have unlimited precision So to avoid overflow error I start using int 32 for low precision integer.
That was my bad for jumping to the conclusion I think I should use numpy type to test

EmilyBourne:
> when I start using values that exceed the type max value I got some OverflowError.
> `OverflowError: unsigned byte integer is greater than maximum`

This is to be expected. An OverflowError is an error that says that the value exceeds the type max value.

> So when I wanted to know how the python manage integer and I saw that for python Integers have unlimited precision So to avoid overflow error I start using int 32 for low precision integer.

When the type of the argument is `int` then we are talking about python integers, when the type has a specified precision then we are talking about numpy integers so this discussion is irrelevant

> In fortran overflow are managed correctly but not in c.
> ex :
> ```
> @types('int8', 'int8')
> def func(x, y):
>     return x + y
> 
> print(func(1555 , 115))
> ```
> it work fine in fortran but get overflow in c when parsing value.
When I try to compile this code in fortran I get a compiler error:
```
Error: Type mismatch in argument ‘x’ at (1); passed INTEGER(8) to INTEGER(1)
```
And if I manually change the type to the correct 1555_1 then I get the compiler error:
```
Error: Integer too big for its kind at (1). This check can be disabled with the option -fno-range-check
```
If I remove the print and use ipython then I get:
```
In [3]: test.func(1555, 115)
Out[3]: -122

In [6]: np.int8(1555)
Out[6]: 19

In [8]: np.int8(115+19)
Out[8]: -122
```
So it seems that there is the same "problem" in Fortran. In my opinion the reaction of the C code is perhaps better than the reaction of the Fortran code as it is at least obvious that the results will be wrong.

>So @yguclu should I assume that value for low precision integer are always either numpy type or dont exceed max value and work with there equivalent data type (short , char ....) or manage them in the new wrapper?

If the user doesn't want to use numpy types then they won't specify `int8` as the argument type, they will simply specify `int`. You cannot specify the precision of a python int therefore it doesn't make sense to pass a python int as an argument to something with the precision specified. It is quite uncommon to use such a low precision as it is susceptible to this kind of error. It is therefore only used if the user is worried about memory consumption and is sure that the values are sufficiently small. If we ignore this input and use a larger int then they have lost the memory advantage that they specifically asked for.

Using char for int8 and short for int16 sounds like a good idea to me

yguclu:
Could we print integers in C with a specified precision? See <[stdint.h](https://en.cppreference.com/w/c/types/integer)>

MenoIy:
From <stdint.h>code
'''
typedef signed char                int8_t;
typedef short int                int16_t;
typedef int                        int32_t;
# if __WORDSIZE == 64
typedef long int                int64_t;
'''
we can do it after all stdint  macro are just  basic C types

yguclu:
I think the contents of <stdint.h> are platform specific, in order to ensure that the resulting macro types are platform-independent. This means that our printed C code would be portable across different architectures. Moreover, I think that using these macro types would make for a very readable C code.

Anyway, this goes beyond the scope of this PR. We should discuss about it at our next meeting.

yguclu (approved):

PR #415, opened 2020-09-03
Unpack tuple index

saidctb:
In python writing `a[1,2]` is equivalent to `a[(1,2)]`, and this PR addresses this issue
 where the `__getitem__` method of `IndexedVariable`  can handle both cases


EmilyBourne:
Do you have a test to show what this is supposed to fix?

pyccel/ast/core.py
```diff
@@ -4593,7 +4593,13 @@ def __init__(
     def __getitem__(self, *args):
 
         if self.shape and len(self.shape) != len(args):
-            raise IndexError('Rank mismatch.')
+
+            raise IndexException('Rank mismatch.')
+
+        if len(args) == 1 and isinstance(args[0], (Tuple, tuple, list)):
```
EmilyBourne, line 4599: Should `args[0]` not be a `PythonTuple` for the syntax `a[(1,2)]`?

EmilyBourne (changes requested):
Can you add a test please

saidctb:
> Do you have a test to show what this is supposed to fix?

this issue only occurs when using the ast of Pyccel directly, which is what we do in Psydac

pyccel/ast/core.py
saidctb: this PR doesn't address that case, it's a bug in the IndexedVariable which is used in another project Psydac, but I'm not sure if the `PythonTuple` case can be addressed in the same way

EmilyBourne:
> > Do you have a test to show what this is supposed to fix?
> 
> this issue only occurs when using the ast of Pyccel directly, which is what we do in Psydac
So the example is:
```
from pyccel.ast.core import IndexedVariable
A = IndexedVariable('A', dtype='int', shape=(4,5))
A[(1,2)]
A[1,2]
```
?

EmilyBourne (approved):

pyccel/ast/core.py
EmilyBourne: It seems I have treated this case quite clumsily in `SemanticParser` (and possibly elsewhere) by calling `__getitem__` directly. We should probably use this PR to fix that.
https://github.com/pyccel/pyccel/blob/master/pyccel/parser/semantic.py#L900

EmilyBourne: https://github.com/pyccel/pyccel/blob/master/pyccel/ast/core.py#L2813

EmilyBourne: https://github.com/pyccel/pyccel/blob/master/pyccel/codegen/printing/fcode.py#L2272
https://github.com/pyccel/pyccel/blob/master/pyccel/codegen/printing/fcode.py#L2311
https://github.com/pyccel/pyccel/blob/master/pyccel/codegen/printing/fcode.py#L2319

saidctb:
> > > Do you have a test to show what this is supposed to fix?
> > 
> > 
> > this issue only occurs when using the ast of Pyccel directly, which is what we do in Psydac
> > So the example is:
> 
> ```
> from pyccel.ast.core import IndexedVariable
> A = IndexedVariable('A', dtype='int', shape=(4,5))
> A[(1,2)]
> A[1,2]
> ```
> 
> ?

yes

pyccel/ast/core.py
EmilyBourne: https://github.com/pyccel/pyccel/blob/master/pyccel/parser/semantic.py#L903
https://github.com/pyccel/pyccel/blob/master/pyccel/parser/semantic.py#L1772
https://github.com/pyccel/pyccel/blob/master/pyccel/parser/semantic.py#L1773

EmilyBourne: I think that's all of them ^^'

saidctb: ah yes you are right we should probably use `A[i]` instead of `A.__getitem__(i)`

yguclu (changes requested):
If these changes are needed by Psydac, we should update Pyccel's version so that we can use this version of Pyccel to run the tests.

pyccel/ast/core.py
```diff
@@ -4593,7 +4593,13 @@ def __init__(
     def __getitem__(self, *args):
 
         if self.shape and len(self.shape) != len(args):
-            raise IndexError('Rank mismatch.')
+
+            raise IndexException('Rank mismatch.')
```
EmilyBourne, line 4597: Why `IndexException` instead of `IndexError`? This gives me:
```
NameError: name 'IndexException' is not defined
```

yguclu: I've just rolled it back to `IndexError`

yguclu (approved):

EmilyBourne (approved):

yguclu:
@saidctb Can you please update the PR description with some more details? This is ready to be merged, we just need a proper commit message.

PR #419, opened 2020-09-07
Allow casting to bool within an if statement

### LINKED ISSUE #418: _visit_Is and _visit_IsNot in the semantic parser does not check if there is a bool() method in the expression
**Describe the bug**
_visit_Is and _visit_IsNot in the semantic parser does not check if there is a bool() method in the expression.

**To Reproduce**
For example this python code : 
```
def f():
    a = False
    x = 6
    if a is bool(x):
        x = x - 1
    return False
```

lead to this error : 
```
  File ".../pyccel/codegen/printing/fcode.py", line 1998, in _print_Is
    if a.dtype is NativeBool() and b.dtype is NativeBool():
AttributeError: 'bool' object has no attribute 'dtype'
```
after checking the expression that is passed to _print_Is function, i found it like that : 

`Is(a, bool(x))`

**Expected behavior**
the expression must be like that : 

`Is(a, Bool(x))`

**Language** 
In both C and Fortran

EmilyBourne:
Allow casting to bool within an if statement. Fixes #418 

- Ensure members of Is statement are visited.
- Ensure error messages clearly explain restrictions
- Create _handle_is_operator function to remove Is/IsNot duplication

### DIFF
--- a/pyccel/errors/messages.py
+++ b/pyccel/errors/messages.py
@@ -73,6 +73,7 @@ PYCCEL_RESTRICTION_UNARY_OPERATOR = 'Invert unary operator is not covered by Pyc
 PYCCEL_RESTRICTION_TRY_EXCEPT_FINALLY = 'Uncovered try/except/finally statements by Pyccel'
 PYCCEL_RESTRICTION_RAISE = 'Uncovered raise statement by Pyccel'
 PYCCEL_RESTRICTION_YIELD = 'Uncovered yield statement by Pyccel'
+PYCCEL_RESTRICTION_IS_LHS = 'Only variables are allowed as lhs for is statement'
 PYCCEL_RESTRICTION_IS_RHS = 'Only booleans and None are allowed as rhs for is statement'
 PYCCEL_RESTRICTION_IMPORT = 'Import must be inside a def statement or a module'
 PYCCEL_RESTRICTION_IMPORT_IN_DEF = 'Only From Import is allowed inside a def statement'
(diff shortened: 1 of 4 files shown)


nhamidn:
i forgot to mention that the _visit_IsNot also contain by this error

EmilyBourne:
> i forgot to mention that the _visit_IsNot also contain by this error

You are right the methods for the 2 functions are identical apart from the class name. I have created a function called by both of them to reduce duplication and stop problems existing in both being fixed only for one of the 2.

nhamidn (approved):

yguclu (commented):
How should `is` and `is not` be used now?
The do not appear to be symmetric operators any more.

EmilyBourne:
> How should is and is not be used now?
> The do not appear to be symmetric operators any more.

I'm not sure I understand your question.
The code is essentially the same as before. The are only 2 differences. 
1. Both sides are now visited instead of assuming that the lhs is a Variable and the rhs is a Variable, Boolean or None.
2. The 2 operators use the same code

Do you mean that you would like expressions such as:
```
a = True
True is a
```
to be acceptable?
The code above gives the following error in the master branch:
```
pyccel: [semantic] 
 |fatal: 3| Undefined variable (True)
```
and this error in this branch:
```
pyccel: [semantic] 
 |error: 3| Only variables are allowed as lhs for is statement
```

yguclu:
> Do you mean that you would like expressions such as:
> 
> ```
> a = True
> True is a
> ```
> 
> to be acceptable?

Exactly, there is no reason to distinguish between left-hand side and right-hand side here.

I noticed that you added the error `PYCCEL_RESTRICTION_IS_LHS`, so I wrongly assumed that you had to add a restriction in order to get things working. I did not realize that the behavior in the master branch was also wrong.

yguclu (approved):

PR #420, opened 2020-09-08
WIP: Cwrapper reorganisation

EmilyBourne:
Rewrite of the cwrapper to increase modularity. This allows more complicated datatypes to be handled. Bool and Complex are implemented. Default arguments are implemented
- Create a `pyccel/codegen/printer/cwrapperprinter.py` file which has the same layout as the other printers and inherits from `CCodePrinter`
- Make `Module` into a mutable class to allow the name to be modified (it must match the name of the generated shared library file)
- Create `VariableAddress` class to handle passing the address of pointers/variables. Add necessary prints
- Create `ImaginaryUnit` class. Previously sympy was used however complex initialisation in C requires mathematical interaction with the `ImaginaryUnit`
- Remove dead code from `codegen/cwrapper.py`
- Add printing of `Complex`, `PythonComplex`, `ImaginaryUnit`, `AliasAssign`, `Nil`, `Bool`, `Real`, `Imag` in ccode.py
- Add printing of imports in `_print_Module`
- Add additional_imports property to collect necessary additional imports (e.g. stdio.h, complex.h)
- Add `find_in_dtype_registry` and `get_declare` type functions to reduce code duplication
- Change `_print_Symbol` to `_print_Variable`
- Ensure parser argument is provided to printer class. Update docstring
- Remove unused `_print_ValuedArgument` function in fcode.py add `_print_ValuedVariable`. Print valued variable if this was used in python
- Allow us to know if an argument is keyword only (no way to enforce this in the c wrapper currently)
- Activate tests with bool arguments/return values
- Activate tests with complex arguments/return values
- Activate tests with default arguments
- Add tests for bool/complex default arguments
- Mark tests for bool/complex default arguments as xfail in fortran


pyccel/ast/core.py
```diff
@@ -2386,7 +2386,6 @@ def __init__(
                 name = name[0]
             else:
                 name = DottedName(*name)
-
```
EmilyBourne, line 2389: You should try and checkout meaningless changes before committing. It makes it easier to see which files have been modified

pyccel/codegen/printing/ccode.py
```diff
@@ -10,7 +10,7 @@
 from pyccel.ast.core import Assign, datatype, Variable, Import
 from pyccel.ast.core import CommentBlock, Comment
 
-
+from pyccel.ast.cwrapper import PyccelPyObject, PyArg_ParseTupleNode #PyBuildValueNode
```
EmilyBourne, line 13: We want to avoid having cwrapper things in this file

pyccel/codegen/printing/ccode.py
```diff
@@ -63,6 +63,7 @@
                   ('complex',4) : 'float complex',
                   ('int',4)     : 'int',
                   ('int',8)     : 'long',
+                  ('pyobject', 0) : 'PyObject',
```
EmilyBourne, line 66: If we add a getter in the class for collecting objects from this dictionary then we can extend the function in the cwrapper. That also means that we only need to write the try/except once in the getter rather than every time we use the dictionary

pyccel/codegen/printing/ccode.py
```diff
@@ -113,6 +114,14 @@ def _print_While(self,expr):
         code = code + "\n %s" % self._print(expr.body) + "\n}"
         return (code)
 
+    def _print_PyArg_ParseTupleNode(self, expr):
```
EmilyBourne, line 117: We should have access to this function from CWrapper._print_FunctionDef even if it is defined in CWraapperCodePrinter. And that way we can't accidentally use it while writing normal C

pyccel/codegen/printing/ccode.py
```diff
@@ -195,6 +204,9 @@ def _print_Declare(self, expr):
     def _print_NativeBool(self, expr):
         return 'bool'
 
+    def _print_PyccelPyObject(self, expr):
```
EmilyBourne, line 207: See above

pyccel/codegen/printing/ccode.py
```diff
@@ -222,10 +234,14 @@ def function_signature(self, expr):
...
-            arg_dtypes = [self._print(i.dtype) for i in expr.arguments]
-            arg_dtypes = [dtype_registry[(dtype, arg.precision)] for dtype,arg in zip(arg_dtypes, expr.arguments)]
-            arguments  = [self._print(i) for i in expr.arguments]
-            arg_code   = ', '.join(dtype + ' ' + arg for dtype,arg in zip(arg_dtypes,arguments))
+            #to change for arguments with ranks * ** etc...
+            #arg_dtypes = [self._print(i.dtype) for i in expr.arguments]
+            #arg_dtypes = [dtype_registry[(dtype, arg.precision)] for dtype,arg in zip(arg_dtypes, expr.arguments)]
+            #arguments  = [self._print(i) for i in expr.arguments]
+            #arg_code   = ', '.join(dtype + ' ' + arg for dtype,arg in zip(arg_dtypes,arguments))
+            argument_declare = [Declare(i.dtype, i) for i in expr.arguments]
```
EmilyBourne, line 242: Why are there declarations in the function signature? Does this not lead to semi colons in the wrong place?

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -1,10 +1,12 @@
 from pyccel.codegen.printing.ccode import CCodePrinter
-from pyccel.ast.core import Variable
+from pyccel.ast.core import Variable , Assign, FunctionDef
 from pyccel.ast.datatypes import NativeInteger, NativeReal, NativeComplex, NativeBool, NativeString
-from pyccel.ast.cwrapper import PyccelPyObject, PyArg_ParseTupleNode, PyBuildValueNode
-
+from pyccel.ast.cwrapper import PyccelPyObject, PyArg_ParseTupleNode #PyBuildValueNode
```
EmilyBourne, line 4: What happened to PyBuildValueNode?

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -15,7 +17,7 @@
         NativeReal(): 'd',
         NativeComplex():'c',
         NativeBool():'p',
-        NativeString():'s',
+        NativeString():'s',
```
EmilyBourne, line 20: Trailing whitespace

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -28,7 +30,8 @@ def get_new_name(self, used_names, requested_name):
             used_names.add(requested_name)
             return requested_name
         else:
-            return create_incremented_string(used_names, prefix=requested_name)
+            return 'random_name'
+       #     return create_incremented_string(used_names, prefix=requested_name)
```
EmilyBourne, line 34: It is important to use the incremented string function as this prevents name collisions (it is not random)

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -57,49 +60,51 @@ def _print_FunctionDef(self, expr):
             if cast_func is not None:
                 # TODO: Add other properties
                 collect_var = Variable(dtype=collect_type,
-                        name=get_new_name(used_names, a.name+"_tmp"))
+                        name=self.et_new_name(used_names, a.name+"_tmp"))
```
EmilyBourne, line 63: Missing g

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -57,49 +60,51 @@ def _print_FunctionDef(self, expr):
...
         if len(expr.results)==0:
             func_call = FunctionCall(expr, expr.arguments)
         else:
             results = expr.results if len(expr.results)>1 else expr.results[0]
             func_call = Assign(results,FunctionCall(expr, expr.arguments))
         wrapper_body.append(func_call)
-
         #TODO: Loop over results to carry out necessary casts and collect Py_BuildValue type string
-
+        name = expr.name.name+"_wrapper"
```
EmilyBourne, line 81: We should use get_new_name in case of name collisions

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -57,49 +60,51 @@ def _print_FunctionDef(self, expr):
...
         function_defs = '\n'.join(self._print(f) for f in expr.funcs)
 
-        #TODO: Print ModuleDef (see cwrapper.py L69)
+        module_def = ('static struct PyModuleDef {mod_name}_module = {{\n'
+                '   PyModuleDef_HEAD_INIT,\n'
+                '   \"{mod_name}\",   /* name of module */\n'
+                '   NULL, /* module documentation, may be NULL */\n'
+                '   -1,       /* size of per-interpreter state of the module,*/\n'
+                '   {mod_name}_methods\n'
+                '}};\n\n'.format(mod_name = expr.name))
```
EmilyBourne, line 101: The indentation should be added automatically. I think it's in the doprint function.
The comments can probably be turned into python comments.
I can't think of a reason to prefer c comments in the wrapper. Can you @MenoIy ?

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -136,4 +141,4 @@ def cwrappercode(expr, assign_to=None, **settings):
         ``(*a)`` instead of ``a``.
     """
 
-    return CWrapperCodePrinter(settings).doprint(expr, assign_to)
```
EmilyBourne: The CWrapperCodePrinter should print a module not a Codegen object

pyccel/codegen/printing/ccode.py
EmilyBourne: The line below removes the semi-colon and the new method avoids repetition

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -0,0 +1,144 @@
...
+       #     return create_incremented_string(used_names, prefix=requested_name)
+
+    def get_PyArgParseType(self, dtype):
+        #TODO: Depends on type, rank, etc
+        if dtype is NativeBool():
+            return NativeInteger(), lambda arg, tmp: Assign(arg, Bool(tmp))
+        return dtype, None
+
+    def _print_FunctionDef(self, expr):
+        used_names = set([a.name for a in expr.arguments] + [r.name for r in expr.results])
```
EmilyBourne, line 43: The function name should also be in the used names

pyccel/codegen/printing/ccode.py
MenoIy: this is printer for those type we can put it in wrapper file   later it was just to make sure that the what is wrote in the new wrapper work .

pyccel/codegen/printing/ccode.py
MenoIy: yeah thats true i just put it in ccode for the time being (test purposes only)

pyccel/codegen/printing/ccode.py
MenoIy: I used declare in  function signature for args with rank greater than 1 exemple 
```
void *somefunc(int * x,   char * str)
```

pyccel/codegen/printing/cwrappercode.py
MenoIy: I just commented PyBuildValueNode  so I can run the code to test  pyargs_parsetuple

pyccel/codegen/printing/cwrappercode.py
MenoIy: yes I know I didn't know to create that function so I used random name for the time being

pyccel/codegen/printing/cwrappercode.py
MenoIy: ok i'll do it

pyccel/codegen/printing/cwrappercode.py
MenoIy: yeah thats true ill do it

MenoIy:
for pyargs_parsetuple and py_buildvalue   are they going to  inherit from  functionDef or Functioncall   in ast.core     or they will be independent ?

pyccel/codegen/printing/cwrappercode.py
MenoIy: Yes I was going to correct this one I was  little confused on how  this class work

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: The function already exists it just needs importing from ast.core if it hasn't been yet

pyccel/ast/cwrapper.py
```diff
@@ -0,0 +1,51 @@
...
+#
+    'PyccelPyObject',
+    'PyArg_ParseTupleNode',
+    'PyBuildValueNode'
+)
+
+class PyArg_ParseTupleNode(Basic):
+
+    def __new__(cls, python_func_args, type_keys='', parse_args=None):
+        return Basic.__new__(cls, python_func_args, type_keys, parse_args)
```
EmilyBourne, line 16: Using `__new__` and `self._args` is still found in lots of places in our code, however we dig into this question a few weeks ago and realised there is no reason to do it this way. We should therefore prefer the method used in the `Variable` class, where everything is a mutable object and is stored in a variable which has a descriptive name

MenoIy:
@EmilyBourne  What do think of  this 
1-In the 2 main loop for py_argsparse and py_buildvalue we are already collecting variable (with and without cast function) so why not collect those cast functions in a list
2-Loop on the list with cast_func and print each one before printing the python function
exemple of a cast function :
casting to bool  and from bool
```
_Bool pyint_to_bool(py int)
{
}
pyobject  bool_to_pybool(_Bool)
{
}
```
that will work for other data types  like complex too 

For those cast functions i was thinking they can either be instance of functiondef class or printed manuals

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -119,7 +119,7 @@ def _print_FunctionDef(self, expr):
                 wrapper_body.append(cast_func(a , collect_var))
             else :
                 res_args.append(a)
-            build_keys += pytype_registry[a.dtype]
+            build_keys += pytype_parse_registry [a.dtype]
```
EmilyBourne: I have called it pytype_**parse**_registry because it should be different from the pytype_build_registry. The code `p` used for collecting booleans doesn't exist for pybuild so as a minimum this value will have to be changed

EmilyBourne:
> for pyargsparsetuple and pybuildvaluenode class are they going to inherit from functionDef or Functioncall in ast.core or they will be independent ?

After our discussion yesterday I believe we decided this was a good idea. However I am not sure what if anything we gain by doing this?

EmilyBourne:
> @EmilyBourne What do think of this
> 1-In the 2 main loop for py_argsparse and py_buildvalue we are already collecting variable (with and without cast function) so why not collect those cast functions in a list
> 2-Loop on the list with cast_func and print each one before printing the python function
> exemple of a cast function :
> casting to bool and from bool
> 
> ```
> _Bool pyint_to_bool(py int   x)
> {
>      _Bool res ;
>      res =  x != 0
>      return res;
> }
> pyobject * bool_to_pybool(_Bool   x)
> {
>       pyobject * res;
>       res =   x ? Py_True: Py_False;
>       return res;
> }
> ps this one will create a python boolean type with value true or fase
> ```
> 
> that will work for other data types like complex too
> 
> For those cast functions i was thinking they can either be instance of functiondef class or printed manuals

> 1-In the 2 main loop for py_argsparse and py_buildvalue we are already collecting variable (with and without cast function) so why not collect those cast functions in a list

This sounds like a good idea. I think the cast_functions list should be a property of the `CWrapperCodePrinter` class so that we are sure we have collected the casts for all functions. It should also be a set rather than a list to avoid duplicates
I don't know if it's worth it for one liners such as int->bool cast though. It takes as much space to write `b = (i != 0)` as it does to write `b = pyint_to_bool(i)`. It will definitely be useful for complexes though

EmilyBourne:
All tests are now passing. We can now begin to work on issues #386 and #390 to test the method with more complicated cases.
We will therefore need to activate the appropriate tests

pyccel/codegen/printing/cwrappercode.py
MenoIy: The functions responsible for returning the variable type and the cast function should be enough to make sure what type we are trying to build or parse.

Example for boolean type parsing is done with 'p' and building is done with 'O' so when parsing we return (bool , cast_pyint_to_bool) and when we are building value  return (pyobject, cast_bool_to_pybool)

EmilyBourne: Ah ok I see. 👍

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -0,0 +1,236 @@
...
+        else:
+            return create_incremented_string(used_names, prefix=requested_name)
+
+    def get_PyArgParseType(self, dtype):
+        #TODO: Depends on type, rank, etc
+        if dtype is NativeBool():
+            return NativeInteger(), lambda arg, tmp: Assign(arg, Bool(tmp))
+        return dtype, None
+
+    def get_PyBuildeValue(self, dtype):
```
EmilyBourne, line 43: Typo

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -0,0 +1,236 @@
...
+        if requested_name not in used_names:
+            used_names.add(requested_name)
+            return requested_name
+        else:
+            return create_incremented_string(used_names, prefix=requested_name)
+
+    def get_PyArgParseType(self, dtype):
+        #TODO: Depends on type, rank, etc
+        if dtype is NativeBool():
+            return NativeInteger(), lambda arg, tmp: Assign(arg, Bool(tmp))
```
EmilyBourne, line 40: This needs to be a NativeInteger of precision 4. Should this function receive a Variable and return a Variable and a function instead of receiving a dtype and returning a dtype and a function?

MenoIy: collecting variable and function is a good idea  but we need to pass used_name list and the  arg name too in the function args

MenoIy: or we can keep the code as is it and return a third paramete    precision  with dtype and function .

EmilyBourne: The arg name should be a property of the variable

EmilyBourne: I think it's better to return a Variable. Who knows what other information we may need to add later (e.g. I think optional variables will be pointers) and we don't want to add too many return arguments. Plus the less we do in the for loop the more chance we have of reducing that code to 1 line

pyccel/ast/cwrapper.py
```diff
@@ -119,3 +119,58 @@ def flags(self):
...
+        self._name = name
+        self._arguments = arguments
+        self._ret = ret
+        self._results = results
+        self._cast_type = cast_type
+        body_ = ''
+        #TODO I dont know if the build of body shoudl be done here or in the cwrapper
+        #TODO this is just a tmp way of printing this shoudlbe improved later
+        if self._cast_type == 'pyint_to_bool':
+            body_ += '{} = {} != 0;\n'.format(self.results[0].name, self._arguments[0].name)
```
EmilyBourne, line 141: The printing should avoid code duplication. It should therefore make use of the other functions eg:
`Assign(self.results[0], Bool(self._arguments[0]))`
(although I assume that this is what your comment refers to)

pyccel/ast/cwrapper.py
```diff
@@ -119,3 +119,58 @@ def flags(self):
...
+
+        elif self.cast_type == 'bool_to_pyobj': 
+            body_ += '{} = {} != 0 ? Py_True : Py_False;\n'.format(self.results[0].name, self._arguments[0].name)
```
EmilyBourne, line 144: We need to implement `_print_IfTernaryOperator` to be able to use this syntax

pyccel/ast/cwrapper.py
```diff
@@ -119,3 +119,58 @@ def flags(self):
...
+        self._body = body_
```
EmilyBourne, line 145: I think these should be instances of the `FunctionDef` class rather than being a subclass. As I see it, the reasons to use a class are:
1. To add some special logic in the initialisation (here the logic would be the same if this were an instance rather than a class)
2. To benefit from a common functionality which should be handled in the same way for multiple things (here the common functionality is all entirely described by the `FunctionDef` class)
3. To avoid code repetition (each if option will only be called once so there is no repetition eliminated)

I don't see that we have gained any of these advantages

pyccel/ast/cwrapper.py
```diff
@@ -119,3 +119,58 @@ def flags(self):
...
+    def name(self):
+        return self._name
+
+    @property
+    def arguments(self):
+        return self._arguments
+
+    @property
+    def results(self):
+        return self._results
```
EmilyBourne, line 164: These properties exist with the same definition in the super-class, they therefore do not need to be redefined

pyccel/ast/cwrapper.py
```diff
@@ -119,3 +119,58 @@ def flags(self):
...
+    def ret(self):
+        return self._ret
+
+    @property
+    def cast_type(self):
+        return self._cast_type
+    
+    @property
+    def body(self):
+        return self._body
```
EmilyBourne, line 176: See above

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -26,25 +26,50 @@
...
-        if dtype is NativeBool():
-            return NativeInteger(), lambda arg, tmp: Assign(arg, Bool(tmp))
-        return dtype, None
-
-    def get_PyBuildeValue(self, dtype):
-        #TODO: Depends on type, rank, etc
-        if dtype is NativeBool():
-            return NativeInteger(), lambda arg, tmp: Assign(arg, Bool(tmp))
-        return dtype, None
+            incremented_name, _ = create_incremented_string(used_names, prefix=requested_name)
```
EmilyBourne, line 36: By the way to get rid of Codacy's warning about unused variables you have to name your variable `_` . It is **very** rare that we need to disable codacy's warnings, especially outside of the `tests/` folder. If you're not sure how to get of a warning, ask

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -110,21 +141,25 @@ def _print_FunctionDef(self, expr):
...
         for a in expr.arguments:
-            collect_type, cast_func = self.get_PyArgParseType(a.dtype)
+            collect_var, cast_func = self.get_PyArgParseType(used_names, a)
             if cast_func is not None:
                 # TODO: Add other properties
-                collect_var = Variable(dtype=collect_type,
-                        name=self.get_new_name(used_names, a.name+"_tmp"))
+                cast_func = self.pop_cast_function(cast_func)        
+                if not cast_func in self._cast_functions_set:
+                    code += self._print(cast_func)
```
EmilyBourne, line 157: This prints the definition of the function inside the current function. Are you sure that is what we want? If so what is the point in `self._cast_functions_set`?

EmilyBourne: Ah! No it is printed before the function. However this will lead to a not very neat code:
```
#include <Python.h>

_Bool pyint_to_bool(...) {...}

PyObject* user_func(...) {...}

_Complex pyobject_to_complex(...) {...}

PyObject* user_func2(...) {...}

PyObject* user_func3(...) {...}

module
```

It would be better to group the casting functions together.

This can be done by printing them in `_print_Module`

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -110,21 +141,25 @@ def _print_FunctionDef(self, expr):
...
+                    self._cast_functions_set.add(cast_func)
                 wrapper_vars.append(collect_var)
                 parse_args.append(collect_var)
-                wrapper_body_translations.append(cast_func(a, collect_var))
+                cast_func_call = FunctionCall(cast_func, [collect_var])
+                wrapper_body_translations.append(AliasAssign(a, cast_func_call))
```
EmilyBourne, line 162: Why do you use an AliasAssign?

pyccel/ast/cwrapper.py
```diff
@@ -0,0 +1,176 @@
...
+    def arguments(self):
+        return self._arguments
+
+    @property
+    def results(self):
+        return self._results
+
+    @property
+    def ret(self):
+        return self._ret
```
EmilyBourne, line 168: Why does ret need to be a property? Could it not be appended to the body at the end of the initialisation?

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: `AliasAssign` is necessary when `a` is a pointer. This is the case when `a` is a `PyccelObject`. However it should not be used systematically otherwise we risk having incorrect code (e.g.)
```
def f(a = None):
    print(a)
```
will have a signature:
```
void f(_Bool* a);
```
Calling `AliasAssign(a, cast_func_call)` will give:
```
_Bool* a;
a = pyint_to_bool(a_tmp);
```
instead of :
```
_Bool* a;
*a = pyint_to_bool(a_tmp);
```

To know whether we need to use Assign or AliasAssign we need to know the return type of cast_func. The simplest place to check this is in `get_cast_function`, `get_PyArgParseType` or `get_PyBuildValue`. These functions would therefore return the expression required to cast to the correct type as before rather than returning the function that will be used to translate the expression

pyccel/ast/cwrapper.py
MenoIy: yes thats true

pyccel/ast/cwrapper.py
MenoIy: I knew it that there is something to  there  ,I totaly forgot about this class

EmilyBourne: It's not even implemented in fortran so that might be why

pyccel/ast/cwrapper.py
MenoIy: I used it as a sub class   so i can add __hash__   __eq__ methods  and because  I didn't know how  things will be later on the body part .

pyccel/codegen/printing/cwrappercode.py
MenoIy: ok I understand

pyccel/ast/cwrapper.py
EmilyBourne: Ah ok. If we use a dictionary then do we still need the `__hash__` and `__eq__` methods?
Even if the body changes it will still be unique to each conversion won't it?

MenoIy: if we use dict we can  forget about __hash__  and __eq__    bqz we can use the cast_type (string) as a key

pyccel/ast/cwrapper.py
MenoIy: Yes, its true like I said in the comment didn't know how to implement the body of the cast_function class so I put it aside until I find a better way

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -51,22 +52,47 @@ def get_cast_function(self, used_names, cast_type, from_variable, to_variable):
...
-        cast_function = CastFunction(cast_function_name, cast_type,
-                            cast_function_arg, cast_function_ret, cast_function_result)
+        cast_function_local_vars = []
+ 
+        #switch case of cast_type
+        if cast_type == 'pyint_to_bool':
+            cast_function_body = [Assign(cast_function_result[0], Bool(cast_function_arg[0]))]
+        elif cast_type == 'bool_to_pyobj':
+            cast_function_body = [If((Bool(cast_function_arg[0]),
+                [Assign(cast_function_result[0], Symbol('Py_True'))]),
```
EmilyBourne, line 65: Can we use Constant or Variable instead of Symbol? Symbol is a sympy object and we tend to have problems when sympy objects still exist at the codegen step

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -51,22 +52,47 @@ def get_cast_function(self, used_names, cast_type, from_variable, to_variable):
...
             collect_type = NativeInteger()
             collect_var = Variable(dtype=collect_type, precision=4,
                 name = self.get_new_name(used_names, variable.name+"_tmp"))
-
             cast_function = self.get_cast_function(used_names, 'pyint_to_bool', variable, collect_var)
-
+            return collect_var , cast_function
+        if variable.dtype is NativeComplex():
+            collect_type = PyccelPyObject()
+            collect_var = Variable(dtype=collect_type, rank = 1,
```
EmilyBourne, line 93: Why is this an array? Shouldn't it just be a pointer?

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -75,15 +101,26 @@ def get_PyBuildValue(self, used_names, variable):
...
+                name = self.get_new_name(used_names, variable.name+"_tmp"))
+            cast_function = self.get_cast_function(used_names, 'complex_to_pycomplex', variable, collect_var)
             return collect_var , cast_function
         return variable, None
 
     def _print_PyccelPyObject(self, expr):
         return 'pyobject'
 
+    def _print_FuncCall(self, expr):
```
EmilyBourne, line 117: Is this function necessary? Does FuncCall not print correctly if it inherits from FunctionCall ?

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -51,22 +52,47 @@ def get_cast_function(self, used_names, cast_type, from_variable, to_variable):
...
-        cast_function_name = self.get_new_name(used_names, cast_type)
-        cast_function = CastFunction(cast_function_name, cast_type,
-                            cast_function_arg, cast_function_ret, cast_function_result)
+        cast_function_local_vars = []
+ 
+        #switch case of cast_type
+        if cast_type == 'pyint_to_bool':
+            cast_function_body = [Assign(cast_function_result[0], Bool(cast_function_arg[0]))]
+        elif cast_type == 'bool_to_pyobj':
+            cast_function_body = [If((Bool(cast_function_arg[0]),
```
EmilyBourne, line 64: I think IfTernaryOperator inherits from If so you should already be able to use it even if it doesn't print in a compact way

MenoIy: ok I understand

pyccel/codegen/printing/cwrappercode.py
MenoIy: Yes , Only if create a property getter  in the functioncall class that return  the name of the function  and use it directly in the printer .

pyccel/codegen/printing/cwrappercode.py
MenoIy: I used this one until I found out that the printer of function_call use the printer of the variable to print arguments.
So what happen when we pass a variable  'X'  pointer as arg it will be printed like this 
```
some_function((*X))
```

EmilyBourne: Ah, we need to add a test that compares the function definition arguments to the provided arguments to see if we should be passing a pointer or a value. I think this is already done for the return arguments.
When we have a problem like this we should fix the problem rather than working round it by pretending that we have an array instead of a pointer. Pointers have not been used in C up till now so any problems are just because the syntax hasn't been fully implemented

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: Is it not enough to save the name in the `_func` attribute?

That said, we probably can't use `FuncCall` without defining a `FunctionDef` as that will require access to the `FunctionDef` to get the types of the arguments/return values to ensure pointers are handled correctly

pyccel/codegen/printing/cwrappercode.py
MenoIy: I didn't know that before I just used rank = 1 believing it mean pointer.

EmilyBourne: Ah ok. The `rank` is the number of dimensions. So a matrix (2d-array) has rank 2, a vector (1d-array) has rank 1, and a scalar (0d-array) has rank 0. In C arrays are represented using pointers so the distinction is currently not very clear in our code. But if we were writing C++ and using the `vector` structure instead of C-style arrays then the declaration of `Variable(dtype='int', 'a', rank = 1)` would be `std::vector<int> a` whereas the declaration of `Variable(dtype='int', 'a', is_pointer = True)` would be `int* a`

pyccel/ast/cwrapper.py
```diff
@@ -0,0 +1,143 @@
...
+        self._name = name
+        self._args = args
+
+    @property
+    def name(self):
+        return self._name
+
+    @property
+    def args(self):
+        return self._args
```
EmilyBourne, line 140: I think this class is confusing matters when we are passing pointers. I have added some logic to the `__init__` method of `FunctionCall` to handle pointers correctly, however to know how to handle them we need to know the signature of the function that we are calling. Unfortunately I think that means we will have to create a `FunctionDef` for every function that we use. We can add them to the bottom of this file with `Py_True` and `Py_False`

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -0,0 +1,351 @@
...
+            incremented_name, _ = create_incremented_string(used_names, prefix=requested_name)
+            return incremented_name
+
+    def find_in_dtype_registry(self, dtype, prec):
+        try :
+            return dtype_registry[(dtype, prec)]
+        except KeyError:
+            return CCodePrinter.find_in_dtype_registry(self, dtype, prec)
+
+    def get_cast_function(self, used_names, cast_type, from_variable, to_variable):
```
EmilyBourne, line 53: The set `used_names` contains all the names used in the current function. Here we are creating a new function. Variables local to this function cannot have a name collision with variables local to another function. Therefore `used_names` should not be passed. We should use the set `self._global_names` to avoid collisions with names in the global namespace

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -0,0 +1,351 @@
...
+
+        elif cast_type == 'pycomplex_to_complex':
+            real_part = Variable(dtype = NativeReal(),
+                            name = self.get_new_name(used_names, 'real_part'))
+            imag_part = Variable(dtype = NativeReal(),
+                            name = self.get_new_name(used_names, 'imag_part'))
+            cast_function_local_vars += [real_part, imag_part]
+            cast_function_body = [Assign(real_part, FuncCall('PyComplex_RealAsDouble', cast_function_arg))]
+            cast_function_body += [Assign(imag_part, FuncCall('PyComplex_ImagAsDouble', cast_function_arg))]
+            cast_function_body += [Assign(cast_function_result[0], FuncCall('__builtin_complex', [real_part, imag_part]))]
```
EmilyBourne, line 79: The function `__builtin_complex` is gcc specific. It does not necessarily exist on other compilers. Complex numbers should therefore be initialised in a different way. E.g. with maths and by defining `_print_ImaginaryUnit`
See : https://stackoverflow.com/a/6418824/5957195

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -0,0 +1,351 @@
...
+
+        elif cast_type == 'complex_to_pycomplex':
+            real_part = Variable(dtype = NativeReal(),
+                            name = self.get_new_name(used_names, "real_part"))
+            imag_part = Variable(dtype = NativeReal(),
+                            name = self.get_new_name(used_names, "imag_part"))
+            cast_function_local_vars += [real_part, imag_part]
+            cast_function_body = [Assign(real_part, FuncCall('__real', cast_function_arg))]
+            cast_function_body += [Assign(imag_part, FuncCall('__imag', cast_function_arg))]
```
EmilyBourne, line 88: I cannot find a definition of `__real` and `__imag` online. Can you provide a reference?

I can find a reference to `__real__` and `__imag__` but these are gcc specific functions which we therefore should not use. Instead we should prefer `creal`, `cimag` (see https://en.wikipedia.org/wiki/C_mathematical_functions#complex.h)

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -0,0 +1,351 @@
...
+        wrapper_name = self.get_new_name(used_names.union(self._global_names), expr.name.name+"_wrapper")
+        self._function_wrapper_names[expr.name] = wrapper_name
+        self._global_names.add(wrapper_name)
+        used_names.add(wrapper_name)
+
+        # Collect local variables
+        wrapper_vars = [a for a in expr.arguments] + [r for r in expr.results]
+        python_func_args = Variable(dtype=PyccelPyObject(),
+                                 name=self.get_new_name(used_names, "args"),
+                                 is_pointer=True)
```
EmilyBourne, line 199: We have this code a lot. I wonder if we should create a small helper function :
```
def get_new_PyObject(name, used_names):
    return Variable(dtype=PyccelPyObject(),
                                 name=self.get_new_name(used_names, name),
                                 is_pointer=True)
```
To reduce code duplication/make the code easier to read?

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -0,0 +1,351 @@
...
+        # TODO (After PR 422): Handle optional args
+        for a in expr.arguments:
+            collect_var, cast_func = self.get_PyArgParseType(used_names, a)
+            if cast_func is not None:
+                wrapper_vars.append(collect_var)
+                cast_func_call = FunctionCall(cast_func, [collect_var])
+                if cast_func.results[0].is_pointer:
+                    wrapper_body_translations.append(AliasAssign(a, cast_func_call))
+                else:
+                    wrapper_body_translations.append(Assign(a, cast_func_call))
```
EmilyBourne, line 229: This code is repeated here and for Build. Could `get_PyArgParseType` return the object in `wrapper_body_translations`.

The collection of the additional variables for `wrapper_vars` could be done using set logic:
`wrapper_vars = set(wrapper_vars).union(parse_args)`
although we will have to find a way to preserve the order

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -0,0 +1,351 @@
...
+        res_args = []
+        for a in expr.results :
+            collect_var, cast_func = self.get_PyBuildValue(used_names, a)
+            if cast_func is not None:
+                wrapper_vars.append(collect_var)
+                cast_func_call = FunctionCall(cast_func, [a])
+                if cast_func.results[0].is_pointer:
+                    wrapper_body.append(AliasAssign(collect_var, cast_func_call))
+                else:
+                    wrapper_body.append(Assign(collect_var, cast_func_call))
```
EmilyBourne, line 256: See above

pyccel/ast/cwrapper.py
EmilyBourne: Once this is fixed then we should have a working pull request. The only remaining work will then be to tidy up the code

pyccel/ast/cwrapper.py
```diff
@@ -0,0 +1,143 @@
...
+#--------- CONSTANTS ----------
+    'Py_True',
+    'Py_False'
+)
+
+class PyccelPyObject(DataType):
+    _name = 'pyobject'
+
+#TODO: Is there an equivalent to static so this can be a static list of strings?
+class PyArgKeywords(Basic):
```
EmilyBourne, line 28: Docstring needed

pyccel/ast/cwrapper.py
```diff
@@ -0,0 +1,143 @@
...
+    (NativeReal(), 8)    : 'd',
+    (NativeReal(), 4)    : 'f',
+    (NativeComplex(), 4) : 'O',
+    (NativeComplex(), 8) : 'O',
+    (NativeBool(), 4)    : 'p',
+    (NativeString(), 0)  : 's',
+    (PyccelPyObject(), 0): 'O',
+    }
+
+class PyArg_ParseTupleNode(Basic):
```
EmilyBourne, line 56: Docstring needed

pyccel/ast/cwrapper.py
```diff
@@ -0,0 +1,143 @@
...
+
+    @property
+    def args(self):
+        return self._parse_args
+
+    @property
+    def arg_names(self):
+        return self._arg_names
+
+class PyBuildValueNode(Basic):
```
EmilyBourne, line 113: Docstring needed

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -0,0 +1,356 @@
...
+            incremented_name, _ = create_incremented_string(used_names, prefix=requested_name)
+            return incremented_name
+
+    def find_in_dtype_registry(self, dtype, prec):
+        try :
+            return dtype_registry[(dtype, prec)]
+        except KeyError:
+            return CCodePrinter.find_in_dtype_registry(self, dtype, prec)
+
+    def get_cast_function(self, used_names, cast_type, from_variable, to_variable):
```
EmilyBourne, line 53: Docstring needed

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -0,0 +1,356 @@
...
+                                    arguments  = cast_function_arg,
+                                    body       = cast_function_body,
+                                    results    = cast_function_result,
+                                    local_vars = cast_function_local_vars)
+
+        self._cast_functions_dict[cast_type] = cast_function
+
+        return cast_function
+
+    def get_PyArgParseType(self, used_names, variable):
```
EmilyBourne, line 102: Docstring needed

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -0,0 +1,356 @@
...
+        if variable.dtype is NativeComplex():
+            collect_type = PyccelPyObject()
+            collect_var = Variable(dtype=collect_type, is_pointer=True,
+                name = self.get_new_name(used_names, variable.name+"_tmp"))
+            cast_function = self.get_cast_function(used_names, 'pycomplex_to_complex', collect_var, variable)
+            return collect_var , cast_function
+
+        return variable, None
+
+    def get_PyBuildValue(self, used_names, variable):
```
EmilyBourne, line 120: Docstring needed

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -0,0 +1,356 @@
...
+            collect_type = PyccelPyObject()
+            collect_var = Variable(dtype=collect_type, is_pointer=True,
+                name = self.get_new_name(used_names, variable.name+"_tmp"))
+            cast_function = self.get_cast_function(used_names, 'complex_to_pycomplex', variable, collect_var)
+            return VariableAddress(collect_var), cast_function
+
+        return variable, None
+
+    def _print_PyccelPyObject(self, expr):
+        return 'pyobject'
```
EmilyBourne, line 139: Is this function used?

EmilyBourne: Yes it is used here: https://github.com/pyccel/pyccel/pull/420/files/b34803ddda5c4f6d96630d143397ce8ec67fdf2b#diff-d4275df7f6b7ec4eddfc1206284ccf89R203

pyccel/codegen/printing/cwrappercode.py
MenoIy: the difference  between __real __real__  is just in gcc version , and to use creal and cimag we need to include <complex.h>

EmilyBourne: >the difference between __real __real__ is just in gcc version 

Do you have a reference for this? I get no useful results when searching for `__real` on google

EmilyBourne: > to use creal and cimag we need to include <complex.h>

I think it is to be expected that we will need to use `<complex.h>` when handling complex numbers. It is not the only case where an extra include needs to be added (e.g. `<stdio.h>` for print)

MenoIy: https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.pdf     //  Lexical Elements -- Keywords
https://help.semmle.com/QL/learn-ql/cpp/introduce-libraries-cpp.html  I had another link I just need to find it .  


> > the difference between __real **real** is just in gcc version
> 
> Do you have a reference for this? I get no useful results when searching for `__real` on google

EmilyBourne: > https://help.semmle.com/QL/learn-ql/cpp/introduce-libraries-cpp.html I had another link I just need to find it .

This link is fine. It has `__real` marked as a GNU extension (GNU are the makers of gcc). This means that this will not necessarily work with other compilers and should be avoided

pyccel/codegen/printing/cwrappercode.py
MenoIy: If the variable passed in the cast function have the same name as the element of this function this will create a problem 
```
@types('complex')
def func(x):
    real_part = x
    return real_part
```
the C code will be 
```
PyObject *complex_to_pycomplex(double _Complex real_part)
{
    double real_part;
    etc...........
```

EmilyBourne: You mean if the argument has the same name as an element of the function. We can do:
```
PyObject *complex_to_pycomplex(double _Complex real_arg)
{
    double real_part;
    etc...........
}

double _Complex real_part;
complex_to_pycomplex(real_part);
```
The cast function should always be the same unless the user's function is called `complex_to_pycomplex` I think.

I would be inclined to define all the cast functions in `ast/cwrapper.py` and then just use `get_cast_function` to return the right one

EmilyBourne: That way it is clear that the definition is separate from the call, even if that means that all the possible conversions exist as objects in the python memory (this is not a huge problem, python is memory hungry anyway. We tend to prefer readability where possible)

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -139,7 +148,7 @@ def _print_PyArg_ParseTupleNode(self, expr):
         pyarg = expr.pyarg
         pykwarg = expr.pykwarg
         flags = expr.flags
-        args = ', '.join(['{}'.format(self._print(a)) if a.is_pointer else '&{}'.format(self._print(a)) for a in expr.args])
+        args = ', '.join(['{}'.format(self._print(VariableAddress(a))) for a in expr.args])
```
MenoIy: PyArg_ParseTupleNode shoudl always get the address of the variable (pointer or not)

pyccel/codegen/printing/cwrappercode.py
MenoIy: yes that's a good idea

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: Yes sorry I must have missed this one

EmilyBourne: Ah sorry I see what you mean

PR #422, opened 2020-09-09
Add is and is not to C printer

### LINKED ISSUE #409: (is and is not) not working in c and fortran
**Just found another issue in C and fortran  (is and is not )**
**in C** 
```
@types('bool')
def is_t(x):
    if x is True:
        return 1
    return 0
```

pyccel: [codegen]
 |fatal| Pyccel has encountered syntax that has not been implemented yet. Please create an issue at https://github.com/pyccel/pyccel/issues and provide a small example of your problem. Do not forget to specify your target language (Is(x, True))

**and in Fortran when we compare to something different than bool or None**

```
@types('int')
def test(x):
    return x is 5
```

pyccel: [codegen]
 |fatal| Only booleans and None are allowed as rhs for is statement (Is(x, 5))

### LINKED ISSUE #426: Is operator with both operands are None leads to an error
**`is` operator with both operands are `None` leads to an error**
This problem is due to testing only one of the operands if it's `None` and working with the second without testing it, and trying to access the wrong attributes, this results to an `AttributeError`.  

**To Reproduce**
```
@types()
def none_is_none():
    if None is None:
        z = 1
```

**Error message**
```
  File "/home/pyccel/pyccel/parser/semantic.py", line 737, in <genexpr>
    ls = tuple(self._visit(i, **settings) for i in expr)
  File "/home/pyccel/pyccel/parser/semantic.py", line 723, in _visit
  File "/home/parser/semantic.py", line 2716, in _visit_Is
    return self._handle_is_operator(Is, expr, **settings)
  File "/home/pyccel/pyccel/parser/semantic.py", line 2700, in _handle_is_operator
    if not var1.is_optional:
AttributeError: 'Nil' object has no attribute 'is_optional
```

**Language**
C, Fortran

**Additional context**
Maybe should be fixed within this PR #422

nhamidn:
This pull request is about adding is and is not to C printer. closes #409

- Manage bool-bool comparison in C.

- Raise error for comparison of primitive immutable types.

- Improve None / None comparison. Fixes #426 

- Add more tests.


EmilyBourne:
Is this PR still a draft?

nhamidn:
> Is this PR still a draft?

No i'm gonna mark this PR as ready for review, but tests cannot be activated for C because of issue #390

pyccel/errors/messages.py
```diff
@@ -73,7 +73,7 @@
 PYCCEL_RESTRICTION_TRY_EXCEPT_FINALLY = 'Uncovered try/except/finally statements by Pyccel'
 PYCCEL_RESTRICTION_RAISE = 'Uncovered raise statement by Pyccel'
 PYCCEL_RESTRICTION_YIELD = 'Uncovered yield statement by Pyccel'
-PYCCEL_RESTRICTION_IS_RHS = 'Only booleans and None are allowed as rhs for is statement'
+PYCCEL_RESTRICTION_IS_RHS = 'Only booleans, integers and None are allowed as rhs for is statement'
```
rmahjoubi, line 76: Maybe you should improve this error message, since we made `is` symmetrical there is no difference between rhs and lhs

pyccel/parser/semantic.py
```diff
@@ -2697,6 +2697,8 @@ def _handle_is_operator(self, IsClass, expr, **settings):
             var1, var2 = var2, var1
 
         if isinstance(var2, Nil):
+            if isinstance(var1, Nil):
+                return IsClass(expr.lhs, expr.rhs)
```
EmilyBourne, line 2701: This will create `IsClass(Nil,Nil)`. Is that done on purpose? If yes then how is it evaluated

nhamidn: the None comparison is implemented only in fortran, so i decided to raise the error in the fortran code generation stage.

EmilyBourne: Ok. When the None comparison is implemented in C it will be in a way that will allow it to be symmetrical so this makes sense

yguclu:
Please resolve the conflicts with master and update this branch...

EmilyBourne (approved):

pyccel/errors/messages.py
```diff
@@ -73,11 +73,12 @@
 PYCCEL_RESTRICTION_TRY_EXCEPT_FINALLY = 'Uncovered try/except/finally statements by Pyccel'
 PYCCEL_RESTRICTION_RAISE = 'Uncovered raise statement by Pyccel'
 PYCCEL_RESTRICTION_YIELD = 'Uncovered yield statement by Pyccel'
-PYCCEL_RESTRICTION_IS_RHS = 'Only booleans and None are allowed as rhs for is statement'
+PYCCEL_RESTRICTION_IS_RHS = 'Only booleans, integers and None are allowed as rhs/lhs for is statement'
```
yguclu, line 76: Shouldn't we remove `RHS` from the name of this variable?

pyccel/errors/messages.py
```diff
@@ -73,11 +73,12 @@
...
 PYCCEL_RESTRICTION_IMPORT = 'Import must be inside a def statement or a module'
 PYCCEL_RESTRICTION_IMPORT_IN_DEF = 'Only From Import is allowed inside a def statement'
 PYCCEL_RESTRICTION_IMPORT_STAR = 'import * not allowed'
 PYCCEL_RESTRICTION_OPTIONAL_NONE = 'Variables cannot be equal to None unless they are optional arguments and None is the default value'
+PYCCEL_RESTRICTION_NONE = 'Cannot compare None to None'
```
yguclu, line 81: I don't understand this error.
```
In [1]: None is None                                                                                                               
Out[1]: True

In [2]: None == None                                                                                                               
Out[2]: True
```

tests/epyccel/modules/base.py
```diff
@@ -42,6 +42,20 @@ def compare_is_not_int(a, b):
...
+        c = True
+    return c
+
+@types('int', 'int')
+def compare_isnot_int_int(a, b):
+    c = False
+    if a is not b:
+        c = True
+    return c
+
```
yguclu, line 58: Do we have an example where we use `is` to compare objects of different types?

Even if it makes little sense, I think we should implement it: we know a priori that `is` will return `False` in that case.

pyccel/codegen/printing/fcode.py
```diff
@@ -2000,11 +2000,13 @@ def _print_Is(self, expr):
...
-        if a.dtype is NativeBool() and b.dtype is NativeBool():
+        if (a.dtype is NativeBool() and b.dtype is NativeBool()):
             return '{} .eqv. {}'.format(lhs, rhs)
 
-        errors.report(PYCCEL_RESTRICTION_IS_RHS, symbol=expr,
-            severity='fatal')
+        if (a.dtype is NativeInteger() and b.dtype is NativeInteger()):
+            return '{} .eq. {}'.format(lhs, rhs)
+
+        raise NotImplementedError("Only booleans, integers and None are currently supported for comparison")
```
yguclu, line 2009: Why not floats? And as I mentioned before, why not returning `False` whenever `a is b` has different types for `a` and `b`?

saidctb (approved):

pyccel/codegen/printing/fcode.py
```diff
@@ -2018,8 +2020,10 @@ def _print_IsNot(self, expr):
         if a.dtype is NativeBool() and b.dtype is NativeBool():
             return '{} .neqv. {}'.format(lhs, rhs)
 
-        errors.report(PYCCEL_RESTRICTION_IS_RHS, symbol=expr,
-            severity='fatal')
+        if (a.dtype is NativeInteger() and b.dtype is NativeInteger()):
+            return '{} .ne. {}'.format(lhs, rhs)
+
+        raise NotImplementedError("Only booleans, integers and None are currently supported for comparison")
```
yguclu, line 2026: Same comment as above..

yguclu (changes requested):
If we really care about implementing the `is` operator, I think we can do it without all the restrictions that appear in the proposed implementation. Some discussions may be needed.

pyccel/errors/messages.py
nhamidn: this error, manage the issue https:#426

tests/epyccel/modules/base.py
nhamidn: yes i agree, im gonna implement it

pyccel/errors/messages.py
EmilyBourne: Although Is(Nil, Nil) is a special case which we need to catch, there is nothing stopping us from returning BooleanTrue() when we encounter it instead of raising an error

nhamidn: yes thats true, returning BooleanTrue() is the best way

pyccel/codegen/printing/fcode.py
EmilyBourne: We didn't think it was a good idea to add floats as this would be quite difficult to achieve if we want the same results as python. The `Is` operator is intrinsically linked to how python stores values in memory. For floats they are usually but not always stored separately. So
```
> a = 3.5
> b = 3.5
> a is b
False
> b = a
> a is b
True
```

EmilyBourne: If the user wants a comparison operator then they should use `==` for floats. `Is` is something different which we can't implement entirely

EmilyBourne, line 2009: Why do we use a `NotImplementedError` instead of calling `errors.report`? Can't we use the message stored in `PYCCEL_RESTRICTION_IS_RHS`? If we want to keep this message then it should be modified to indicate that these types are only supported *for the is operator*, these restrictions do not exist for comparison operators in general

saidctb: we have the same problem with integers
> a = 275 
> b = 275          
>a is b           
 False

saidctb: I think we should do as numba does by comparing between the two values even if it's does not give us the same result as in python

EmilyBourne: Hmm, it seems you are right. On my computer `is` returns the same thing as `==` for integers between -5 and 256

saidctb: https://stackoverflow.com/questions/15171695/whats-with-the-integer-cache-maintained-by-the-interpreter
you can find here an explanation of the caching system

EmilyBourne: In that case integers and floats should be handled in the same way. I.e, either:
1. By implementing `is` as `==` despite that not being correct behaviour
2. Raise an error saying that pyccel does not handle python's `id` function and therefore cannot implement `is` unless one of the types is either `bool` or `None`
Personally I prefer the second option. What do you think?

EmilyBourne: >https://stackoverflow.com/questions/15171695/whats-with-the-integer-cache-maintained-by-the-interpreter
you can find here an explanation of the caching system

Interesting so -5 and 256 is not just anecdotal then. So technically we could also implement `is` for `np.uint_8` (although we would first have to implement unsigned ints ^^')

saidctb: the second option seems okay for me

yguclu: I also agree, we cannot fully replicate the behavior of `is` in Python, which also changes between interpreters! For example PyPy behaves differently than CPython: https://doc.pypy.org/en/latest/cpython_differences.html#object-identity-of-primitive-values-is-and-id

yguclu: I think that the rules in PyPy ([here](https://doc.pypy.org/en/latest/cpython_differences.html#object-identity-of-primitive-values-is-and-id)) make a lot of sense: for all primitive types `is` is equivalent to `==`, and for other types (arrays for example) it checks object identity (which we can do a priori given that we compile the code statically). What do you guys think?

tests/epyccel/modules/base.py
```diff
@@ -42,6 +42,20 @@ def compare_is_not_int(a, b):
         c = True
     return c
 
+@types('int', 'int')
+def compare_is_int_int(a, b):
+    c = False
+    if a is b:
+        c = True
+    return c
```
yguclu, line 50: Is there a reason why you don't just write the following line?
```python
return a is b
```

tests/epyccel/modules/base.py
```diff
@@ -143,3 +157,26 @@ def if_0_real(a):
         return True
     else:
         return False
+
+
+@types('int','float')
+def is_types(x,y):
+    if x is y:
+        return True
+    return False
```
yguclu, line 166: Same comment: `return x is y`

yguclu (changes requested):

pyccel/codegen/printing/fcode.py
saidctb: seems good to me

nhamidn: do i need to make `is` equivalent to `==` for the the implemented primitive immutable types in this pull request ?

saidctb: yes @nhamidn

EmilyBourne: Personally I am not a fan of this method. All our tests work in the same way: they generate a fortran file and they check that the output of the fortran file and the original python code is the same. If we translate `is` to `==` we will not be able to use this method. To me this feels like an indication that we are doing something contrary to the way pyccel works.

If we do decide to use this method then we will need to either write tests and pass the expected result explicitly, or write 2 functions 1 with `is` to be translated and 1 with `==` for the comparison. The latter method allows us to run automated tests with random values that are more likely to catch corner cases.

If this method were implementing something new then I don't think it would bother me, but as the user has the option to use the `==` operator then if they have chosen to use `is` there should be a reason for that (even if that reason is simply because they are unaware of the difference). Personally I would prefer that we raise an error "The function id() is not implemented for basic types `int`, `float`. `str`. Please use `==` for comparisons"

That said if everyone else thinks it is better to implement `is` following PyPy's rules then I'm happy to go with the majority

yguclu: @EmilyBourne I think you're right here.

I would like to follow PyPy behavior (which is also consistent with Numba's implementation as far as I understand), but this makes little sense as we are using CPython to create our wrappers.

And given that CPython does not handle `id(a)` in a meaningful way when `a` is an immutable primitive type, the only sensible option that remains is exactly what you suggest.

Hence we should raise the error
```python
'Cannot translate "is" comparison, because function "id()" is not implemented for basic types "int", "float", and "str". Please use "==" for comparison instead.'
```

yguclu (changes requested):
Please add error message for `is` comparison of primitive immutable types, as suggested by @EmilyBourne.

pyccel/codegen/printing/fcode.py
EmilyBourne:
```
'Cannot translate "is" comparison, because function "id()" is not implemented for basic types "int", "float", "complex" and "str". Please use "==" for comparison instead.'
```
I missed complex

pyccel/parser/semantic.py
```diff
@@ -2719,8 +2719,11 @@ def _handle_is_operator(self, IsClass, expr, **settings):
...
             return IsClass(var1, var2)
 
-        if ((var1.is_integer or isinstance(var1.dtype, NativeInteger)) and
-            (var2.is_integer or isinstance(var2.dtype, NativeInteger))):
+        lst = ["string", "complex", "real", "int"]
+        if (str(var1.dtype) in lst and str(var2.dtype) in lst):
+            errors.report(PYCCEL_RESTRICTION_PRIMITIVE_IMMUTABLE, symbol=expr,
+            bounding_box=(self._current_fst_node.lineno, self._current_fst_node.col_offset),
+            severity='error', blocker=self.blocking)
             return IsClass(var1, var2)
```
yguclu: I am OK with raising this error for an expression like `a is b`, but I think we can definitely handle the cases
```python
x is x      # True
x is not x  # False
```
where `x` is a **variable** of primitive immutable type, not an expression or a literal constant.

yguclu (changes requested):

pyccel/codegen/printing/ccode.py
```diff
@@ -422,7 +430,29 @@ def _print_IsNot(self, expr):
 
             return '{} != {}'.format(lhs, rhs)
         else:
-            raise NotImplementedError
+            errors.report(PYCCEL_RESTRICTION_IS_ISNOT,
+                          symbol=expr, severity='fatal')
+
+    def _print_Is(self, expr):
```
EmilyBourne, line 436: Is it possible to combine the two functions `_print_Is` and `_print_IsNot` into one, similarly to what is done in parser/semantic.py? This means you only need to update one function when making changes and guarantees that the two functions use the same method. You could pass the printed form of the operator (`==` for `is`, `!=` for `is not`) as an argument to avoid requiring several if/else blocks

pyccel/parser/semantic.py
```diff
@@ -2698,17 +2698,35 @@ def _handle_is_operator(self, IsClass, expr, **settings):
...
+            elif IsClass == IsNot:
+                return BooleanTrue()
+
         if ((var1.is_Boolean or isinstance(var1.dtype, NativeBool)) and
             (var2.is_Boolean or isinstance(var2.dtype, NativeBool))):
             return IsClass(var1, var2)
 
-        errors.report(PYCCEL_RESTRICTION_IS_RHS,
+        lst = ["string", "complex", "real", "int"]
+        if (str(var1.dtype) in lst and str(var2.dtype) in lst):
```
EmilyBourne, line 2723: You have already confirmed that `var1.dtype==var2.dtype`, therefore if one is in `lst` then so is the other

pyccel/parser/semantic.py
```diff
@@ -2698,17 +2698,35 @@ def _handle_is_operator(self, IsClass, expr, **settings):
...
+                return BooleanFalse()
+            elif IsClass == IsNot:
+                return BooleanTrue()
+
         if ((var1.is_Boolean or isinstance(var1.dtype, NativeBool)) and
             (var2.is_Boolean or isinstance(var2.dtype, NativeBool))):
             return IsClass(var1, var2)
 
-        errors.report(PYCCEL_RESTRICTION_IS_RHS,
+        lst = ["string", "complex", "real", "int"]
```
EmilyBourne, line 2722: Please use `NativeString()`, `NativeComplex()`, `NativeReal()`, `NativeInt()` instead of converting to strings

tests/errors/semantic/non_blocking/PRIMITIVE_IMMUTABLE_RESTRICTION_IS.py
```diff
@@ -0,0 +1,3 @@
+@types('int', 'int')
+def compare_is_int_int(a, b):
+    return a is b
```
EmilyBourne, line 3: We should probably add tests for the other types too

EmilyBourne (changes requested):

pyccel/parser/semantic.py
EmilyBourne: I agree

pyccel/codegen/printing/ccode.py
yguclu: This would be useful indeed.

tests/errors/semantic/non_blocking/PRIMITIVE_IMMUTABLE_RESTRICTION_IS.py
yguclu: I agree.

pyccel/parser/semantic.py
nhamidn: i agree

tests/epyccel/test_base.py
```diff
@@ -99,6 +99,29 @@ def test_if_0_real(language):
...
+            pytest.mark.xfail(reason="Strings are not yet implemented for C language"),
+            pytest.mark.c]),
+        pytest.param("fortran", marks = pytest.mark.fortran)
+    ]
+)
+def test_same_string(language):
+    compare_epyccel(base.is_same_string, language=language)
+    compare_epyccel(base.isnot_same_string, language=language)
+
+def test_same_comlex(language):
```
yguclu, line 121: Typo: missing `p` in `test_same_complex`

yguclu (changes requested):
Good job!
One typo needs fixing..

pyccel/codegen/printing/ccode.py
```diff
@@ -412,17 +412,33 @@ def _print_Imag(self, expr):
         else:
             return '0'
 
-    def _print_IsNot(self, expr):
+    def _print_is_operator(self, Op, expr):
```
EmilyBourne, line 415: We normally try to avoid starting function names with `_print` if it is not followed by the name of a class, so they don't accidentally catch an instance of something. Could you rename this please

pyccel/parser/semantic.py
```diff
@@ -2699,6 +2699,12 @@ def _handle_is_operator(self, IsClass, expr, **settings):
         var1 = self._visit(expr.lhs)
         var2 = self._visit(expr.rhs)
 
+        if (var1 == var2):
```
EmilyBourne, line 2702: @saidctb Do you know how `==` is implemented for instances of `Variables`? Is `==` sufficient or should we use `is` here?

EmilyBourne, line 2702: This should only compare Variable instances. I think this will return `True` for the following case when it should raise an error:
```
In [7]: a+b is a+b
Out[7]: False
```

EmilyBourne: I have just run the test and this is indeed translated to `True`

EmilyBourne (changes requested):

pyccel/parser/semantic.py
nhamidn: yes you are right

pyccel/codegen/printing/ccode.py
yguclu: You're right, this utility function should not have a name starting with `_print`

pyccel/parser/semantic.py
yguclu: @EmilyBourne Good catch!

EmilyBourne (changes requested):
Function name still needs changing

nhamidn:
> Function name still needs changing

i don't know how i forgot to change the function name

yguclu (approved):

EmilyBourne (approved):

PR #427, opened 2020-09-15
Raise error if number of arguments doesn't match decorator

### LINKED ISSUE #421: No error is raised if the wrong number of argument types is provided
**Describe the bug**
When the number of arguments in the `types` decorator does not match the number of function arguments pyccel does not complain.

### Too many arguments in `types` decorator

**To Reproduce**
```python
@types('int', 'int')
def f(x):
    print(x)
```

This case is not too problematic. The code compiles correctly and the additional argument is discarded. However it is probable that the user is unaware of the problem and has simply forgotten to add a function argument or remove a `types` argument. It is therefore not necessarily the last `types` argument that the user wishes to ignore

**Expected behavior**
As a minimum we should raise a warning here. We may even want to raise an error

### Too few arguments in `types` decorator

**To Reproduce**
```python
@types('int')
def f(x,y):
    print(x)
```
In this first case the code compiles as y is unused. It generates the following code:
```
module args

implicit none

contains

subroutine f(x)

  implicit none

  integer(kind=8), value :: x

  print *, x

end subroutine f

end module args
```
which is unusable.

The second case, where the second argument is used is:
```python
@types('int')
def f(x,y):
    print(x)
    print(y)
```
This raises the following error:
```
pyccel: [semantic] 
 |fatal: 3| Undefined variable (y)
```
While an error is raised it is not immediately obvious to the user why `y` is undefined

**Expected behavior**
We should raise an error explaining that the number of arguments in the function and the `types` decorator don't match

nhamidn:
Raise error when the number of arguments in the types decorator/header does not match the number of function arguments. Fixes #421 

- A warning is raised when the number of arguments of the function < the number of arguments of types decorator/header.

- An error is raised when when the number of arguments of the function > the number of arguments of types decorator/header.

- Add tests for error and warnings.


pyccel/parser/syntactic.py
```diff
@@ -657,6 +659,13 @@ def _visit_FunctionDef(self, stmt):
 
                 i = i+1
 
+            if (len(arguments) != args_number):
+                msg = 'The number of arguments in the function ({}) {} and the types decorator ({}) {} don\'t match.'.format(len(arguments), arguments, args_number, decorators['types'].args)
+                if (len(arguments) < args_number):
+                    errors.report(msg, severity='warning')
+                else:
+                    errors.report(msg, severity='error')
```
EmilyBourne, line 667: Severity should be fatal

EmilyBourne, line 667: Bounding box and symbol should be provided to help the user find the source of the problem

EmilyBourne:
Does this PR handle the problem for argument types declared in the `@types` decorator or also in the `#$ header`? (I have not tested the header case to confirm that it has the same problem)

saidctb:
can you add some tests and tests if it works for recursive functions ?

nhamidn:
> Does this PR handle the problem for argument types declared in the `@types` decorator or also in the `#$ header`? (I have not tested the header case to confirm that it has the same problem)

this PR handle the problem for argument types declared in the `@types` decorator, how can i test the header and see if it has the same error ?

nhamidn:
> can you add some tests and test if it works for recursive functions ?

Tests like checking if the error/warning is raised ?
i just tested a recursive function and it fail

for example this python code : 
```

@types('int')
def rec(x):
    x = x + 1
    if (x < 10):
        x = rec(x)
    return x

```

when i try to convert it to fortran i got this error : 

```
  File ".../pyccel/ast/core.py", line 2989, in __init__
    self._dtype       = func.results[0].dtype if len(func.results) == 1 else NativeTuple()
AttributeError: 'Symbol' object has no attribute 'dtype'
```

EmilyBourne:
> > can you add some tests and test if it works for recursive functions ?
> 
> Tests like checking if the error/warning is raised ?

Exactly. Tests like these can be added by adding a file in the tests/errors/ or tests/warnings folder. The test_errors.py file wile test whether an error is raised in the correct stage. Unfortunately there is no system checking if the correct error message is raised

EmilyBourne:
> how can i test the header and see if it has the same error ?

There are several examples using a header instead of a types decorator in the tests folder. There are several in  `tests/codegen/fcode/scripts/functions.py`

pyccel/parser/syntactic.py
EmilyBourne: Ah, sorry I have just realised that we only use the information in the types decorator for the semantic step. Therefore it is not blocking in the syntactic step. The severity is therefore 'error' as you had before, not 'fatal'

EmilyBourne: The difference between the 2 is that 'fatal' stops the current stage from completing. This is useful, for example, if we have not collected the `dtype` information and we will need it in this stage.
'error' allows the current stage to complete but doesn't move on to the next stage. So in this case that would allow all discoverable errors to be found in the syntactic stage but would not carry out the semantic stage where we will have problems due to a lack of information about the `dtype`

EmilyBourne:
> when i tried to convert it to fortran i got this error :
> 
> ```
>   File ".../pyccel/ast/core.py", line 2989, in __init__
>     self._dtype       = func.results[0].dtype if len(func.results) == 1 else NativeTuple()
> AttributeError: 'Symbol' object has no attribute 'dtype'
> ```
Recursive functions require the result type to be provided as well as the argument types. See `tests/pyccel/scripts/hope_benchmarks_decorators/hope_fib.py` for an example with a types decorator and `tests/pyccel/scripts/hope_benchmarks/hope_fib.py` for an example with a header.

This case should also raise a neat error indicating that the user needs to provide a results type. I will open an issue

pyccel/parser/syntactic.py
nhamidn: thank you for the information now i understand the difference between the error and fatal severity, im gonna use error as severity.

EmilyBourne:
I have just found this function https://github.com/pyccel/pyccel/tree/master/pyccel/ast#L3566:
`def is_compatible_header(self, header):` in `class FunctionDef`
Apparently it is never called.
Either this branch should supersede it and therefore remove it or it should use it I think

pyccel/parser/semantic.py
```diff
@@ -2315,12 +2314,22 @@ def _visit_FunctionDef(self, expr, **settings):
...
         if header is None:
             if cls_name:
                 header = self.get_header(cls_name +'.'+ name)
+                args_number -= 1
             else:
                 header = self.get_header(name)
 
+        if header:
+            if (args_number != len(header.dtypes)):
+                msg = 'The number of arguments in the function {} ({}) does not match the the number of types in decorator/header ({}).'.format(name ,args_number, len(header.dtypes))
```
EmilyBourne, line 2327: `does not match the the`
you have `the` twice

pyccel/parser/syntactic.py
```diff
@@ -636,7 +636,6 @@ def _visit_FunctionDef(self, stmt):
             ls = decorators['types'].args
             while i<len(ls) :
                 arg = ls[i]
-
```
EmilyBourne, line 639: Please try to avoid adding/removing lines in a file where no other changes are made. All it does is add an extra file for reviewers

EmilyBourne (approved):

EmilyBourne (approved):

EmilyBourne (approved):

pyccel/parser/semantic.py
```diff
@@ -2315,12 +2314,22 @@ def _visit_FunctionDef(self, expr, **settings):
...
             else:
                 header = self.get_header(name)
 
+        if header:
+            if (args_number != len(header.dtypes)):
+                msg = 'The number of arguments in the function {} ({}) does not match the number of types in decorator/header ({}).'.format(name ,args_number, len(header.dtypes))
+                if (args_number < len(header.dtypes)):
+                    errors.report(msg, symbol=expr.arguments, severity='warning')
+                else:
+                    errors.report(msg, symbol=expr.arguments, severity='fatal')
```
yguclu, line 2331: Could you please explain why do you raise only a warning if the number of function arguments is smaller than what given in the decorator/header? I understand that in this case the decorator/header contains enough information to correctly describe all function arguments, but in practice this situation will typically happen when the user modifies an existing function but forgets to update its decorator/header...

nhamidn: i don't think it good to stop Pyccel from completing the job in this situation, if we raise an error the user will need to modify the code and then execute Pyccel again, given that the types in the header/decorator goes respectively with the function arguments i think that a warning is enough to pay the attention of the user, and if we raise an error it will result a bad user experience.

yguclu: OK, I can agree with that.

PR #428, opened 2020-09-15
Cwrapper rewrite

### LINKED ISSUE #386: Complex numbers in C

### LINKED ISSUE #390: Some data types don't work as arguments in C ex (int8 int16 bool)

### LINKED ISSUE #413: Default arguments are not fully supported in c
**Describe the bug**
Although default arguments are provided in generated code they are not provided in the wrapper. As a result the function cannot be called without providing all arguments

**Test Code**
```python
@types('real','real')
def f(x = 1.5, y = 2.5):
    return x+y

# ...
def g():
    a = f(19.2,6.7)
    b = f(4.5)
    c = f(y = 8.2)
    d = f()
    #return a,b,c,d
```

This generates:
```c
double f(double x, double y)
{
    double Out_0001;
    Out_0001 = x + y;
    return Out_0001;
}

void g(void)
{
    double a;
    double b;
    double c;
    double d;
    a = f(19.2,6.7);
    b = f(4.5,2.5);
    c = f(1.5,8.2);
    d = f(1.5,2.5);
    /*return a,b,c,d*/
}
```
with the following wrapper for `f`:
```c
static PyObject * f_wrapper(PyObject *self, PyObject *args)
{
    double x;
    double y;
    double Out_0001;
    if (!PyArg_ParseTuple(args, "dd", &x, &y))
        return NULL;
    Out_0001 = f(x,y);
    return Py_BuildValue("d", Out_0001);
}
```

**Language**
C

**Additional Information**
The following link describes how to add optional arguments:
https://pythonextensionpatterns.readthedocs.io/en/latest/parsing_arguments.html#variable-number-of-arguments
Following this logic the wrapper should read:
```c
static PyObject * f_wrapper(PyObject *self, PyObject *args)
{
    double x;
    double y;
    double Out_0001;
    x = 1.5
    y = 2.5
    if (!PyArg_ParseTuple(args, "|dd", &x, &y))
        return NULL;
    Out_0001 = f(x,y);
    return Py_BuildValue("d", Out_0001);
}
```
however with this syntax the function cannot be called with the syntax `f(y = 8.2)` from python. To allow this more complicated case we must follow the instructions here:
https://pythonextensionpatterns.readthedocs.io/en/latest/parsing_arguments.html#variable-number-of-arguments-and-keyword-arguments
Which should give the following code:
```c
static PyObject * f_wrapper(PyObject *self, PyObject *args, PyObject *kwargs)
{
    static char *kwlist[] = {
        "x",
        "y",
        NULL
    };
    double x;
    double y;
    double Out_0001;
    x = 1.5
    y = 2.5
    if (! PyArg_ParseTupleAndKeywords(args, kwargs, "|dd",
                                      kwlist, &x, &y)) {
        return NULL;
    }
    Out_0001 = f(x,y);
    return Py_BuildValue("d", Out_0001);
}
```

EmilyBourne:
Rewrite of the cwrapper to increase modularity. This allows more complicated datatypes to be handled. Bool is implemented as argument/return value (fixes #390). Complex is implemented as argument/return value (fixes #386). Default arguments are implemented (fixes #413)

- Create `pyccel/codegen/printer/cwrappercode.py` file which has the same layout as the other printers and inherits from `CCodePrinter`
- Create `ast/cwrapper.py` file which contains any nodes or objects which are created for the `CWrapperCodePrinter`
- Make Module into a mutable class to allow the name to be modified (it must match the name of the generated shared library file)
- Create `VariableAddress` class to handle passing the address of pointers/variables. Add necessary prints
- Create `ImaginaryUnit` class. Previously sympy was used however complex initialisation in C requires mathematical interaction with the `ImaginaryUnit`
- Remove dead code from codegen/cwrapper.py
- Add printing of `Complex`, `PythonComplex`, `ImaginaryUnit`, `AliasAssign`, `Nil`, `Bool`, `Real`, `Imag` in `ccode.py`
- Add printing of imports in `_print_Module`
- Add additional_imports property to collect necessary additional imports (e.g. stdio.h, complex.h)
- Add `find_in_dtype_registry` and `get_declare` type functions to reduce code duplication
- Change `_print_Symbol` to `_print_Variable`
- Ensure parser argument is provided to printer class. Update docstring
- Remove unused `_print_ValuedArgument` function in `fcode.py` and add `_print_ValuedVariable`. Print valued variable if this was used in python
- Allow us to know if an argument is keyword only (no way to enforce this in the c wrapper currently)
- Activate tests with bool arguments/return values
- Activate tests with complex arguments/return values
- Activate tests with default arguments
- Add tests for bool/complex default arguments
- Mark tests for bool/complex default arguments as xfail in fortran

cc: @MenoIy


saidctb (approved):

yguclu:
@EmilyBourne You forgot to mention that you've created a new file, **pyccel/ast/cwrapper.py**, which contains some new AST nodes.

pyccel/ast/core.py
```diff
@@ -1850,6 +1860,17 @@ def __new__(
...
+
+    def __init__(
+        self,
+        name,
+        variables,
+        funcs,
+        interfaces=[],
+        classes=[],
+        imports=[],
+        ):
```
yguclu, line 1873: Would you mind updating the docstring of this class? It appears to be outdated.

pyccel/ast/core.py
```diff
@@ -1850,6 +1860,17 @@ def __new__(
         classes=[],
         imports=[],
         ):
+        return Basic.__new__(cls)
```
yguclu, line 1863: Given that no arguments are passed to `Basic.__new__` apart from `cls`, we should rewrite this function as
```
def __new__(cls, *args, **kwargs):
    return Basic.__new__(cls)
```

pyccel/ast/core.py
```diff
@@ -2345,7 +2365,8 @@ def __init__(
         cls_parameters=None,
         order='C',
         precision=0,
-        is_argument=False
+        is_argument=False,
+        is_kwonly=False
         ):
```
yguclu, line 2362: Could you update the docstring of this class too?

And I have a similar comment as above about `__new__`:
```python
def __new__(cls, *args, **kwargs):
    return Basic.__new__(cls)
```

pyccel/ast/core.py
```diff
@@ -2907,8 +2942,11 @@ class ValuedArgument(Basic):
     >>> n
     n=4
     """
+    def __new__(cls, *args, **kwargs):
+        kwargs.pop('kwonly', False)
```
yguclu, line 2938: Why are you removing `kwonly` from the keyword arguments? Maybe you wanted to write `kwargs.setdefault('kwonly', False)` instead?

pyccel/ast/core.py
```diff
@@ -2907,8 +2942,11 @@ class ValuedArgument(Basic):
...
+        return Basic.__new__(cls, *args, **kwargs)
```
yguclu, line 2939: I think we need to have a general discussion about which parameters we pass to this constructor.
For example, here it seems natural to only pass `cls`.

EmilyBourne:
We need to dereference

pyccel/ast/core.py
EmilyBourne: I am removing it because Basic does not accept kwargs. This is due to me using the Basic.__new__ call incorrectly. I have fixed this now

pyccel/ast/core.py
EmilyBourne: 👍

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -0,0 +1,385 @@
...
+
+        if variable.dtype is NativeComplex():
+            collect_type = PyccelPyObject()
+            collect_var = Variable(dtype=collect_type, is_pointer=True,
+                name = self.get_new_name(used_names, variable.name+"_tmp"))
+            cast_function = self.get_cast_function_call('pycomplex_to_complex', collect_var)
+            body = [Assign(variable, cast_function)]
+            if isinstance(variable, ValuedVariable):
+                # Decrement PyObject counter which was incremented by PyComplex_FromDoubles
+                body += [FunctionCall(Py_DECREF, [collect_var])]
```
EmilyBourne, line 114: This leads to the argument being derefenced if it is provided which leads to a segmentation fault

EmilyBourne: This has been solved by initialising the PyObject to None and adding an `If`` statement to test whether PyObject points to an argument or not. If it does the argument is collected. If not the default argument is used.

This also reduces the number of casts necessary

pyccel/codegen/printing/ccode.py
```diff
@@ -413,6 +413,18 @@ def _print_Imag(self, expr):
...
+        if Nil() in expr.args:
+            lhs = VariableAddress(expr.lhs) if isinstance(expr.lhs, Variable) else expr.lhs
+            rhs = VariableAddress(expr.rhs) if isinstance(expr.rhs, Variable) else expr.rhs
+
+            lhs = self._print(lhs)
+            rhs = self._print(rhs)
+
+            return '{} != {}'.format(lhs, rhs)
+        else:
+            raise NotImplementedError
```
EmilyBourne, line 426: I have left the equivalent `_print_Is` statement for PR #422

yguclu (changes requested):

pyccel/ast/core.py
```diff
@@ -2917,27 +2959,48 @@ def __new__(cls, expr, value):
...
+        return self._kwonly
+
     def _sympystr(self, printer):
         sstr = printer.doprint
 
         argument = sstr(self.argument)
         value = sstr(self.value)
         return '{0}={1}'.format(argument, value)
 
+class VariableAddress(Basic, PyccelAstNode):
```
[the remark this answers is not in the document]
EmilyBourne: In C it corresponds to the memory address. It is necessary to handle pointers correctly. In Fortran pointers are handled as though they were variables

pyccel/ast/core.py
```diff
@@ -1808,12 +1818,12 @@ class Module(Basic):
     variables: list
         list of the variables that appear in the block.
```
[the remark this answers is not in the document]
EmilyBourne: Oops

pyccel/ast/core.py
```diff
@@ -2893,7 +2924,16 @@ class Argument(Symbol, PyccelAstNode):
     n
     """
 
-    pass
+    def __new__(cls, *args, **kwargs):
+        kwargs.pop('kwonly', False)
+        return Symbol.__new__(cls, *args, **kwargs)
+
+    def __init__(self, *args, **kwargs):
+        self._kwonly = kwargs.pop('kwonly', False)
```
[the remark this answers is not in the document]
yguclu: I suspect that you want `kwonly` to always be a keyword argument, in which case it should be preceded by `*` as I wrote.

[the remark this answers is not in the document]
EmilyBourne: Ah, you are right sorry I missed that

yguclu (approved):

pyccel/ast/core.py
```diff
@@ -2927,10 +2927,10 @@ class Argument(Symbol, PyccelAstNode):
...
     """
 
-    def __new__(cls, *args, *, kwonly=False, **assumptions):
-        return Symbol.__new__(cls, *args, **assumptions)
+    def __new__(cls, name, *, kwonly=False, **assumptions):
+        return Symbol.__new__(cls, name, **assumptions)
 
-    def __init__(self, *args, *, kwonly=False, **assumptions):
+    def __init__(self, name, *, kwonly=False, **assumptions):
         self._kwonly = kwonly
```
yguclu: Finally! This is what I wrote in my comment! 😜

EmilyBourne: Not enough copy-pasting

Issue #359, opened 2020-07-10
Handle multiple assignment to same allocatable variable


yguclu:
In Python every assignment `x = expr` declares a new variable `x`, to which it assigns the value of `expr`.

When we generate Fortran code we map this Python expression to a Fortran assignment, and if necessary we precede it with the declaration of `x` (if it was never used before) and with a memory allocation (if `expr` is a Numpy array constructor and `x` was not declared with `@stack_array('x')`).

In all cases we cannot allow the user to write `x = expr` (i.e. we raise a PyccelError) if the two sides of the assignment have either

1. different `dtype`
2. different `rank`

Moreover, even if `x` and `expr` have same `dtype` and `rank`, the case where the value of `expr` is a Numpy array requires additional care. For the situations that require memory reallocation we raise a PyccelWarning:

3. if Numpy array `x` already exists
4. if Numpy array `x` is created inside of a loop
5. if Numpy array `x` is created inside of an if/else statement

For these situations we will add some Fortran code that checks if `x` was already allocated, and if the runtime shape of `x` and `expr` is different. Memory reallocation will only happen if needed, but Pyccel will not be able to predict that (hence the PyccelWarning).

In conclusion we need 3 new AST node classes:
1. AllocateNode
2. DeallocateNode (already exists under name of "Delete")
3. CheckReallocationNode (which uses nodes 1 and 2)

For now in semantic we will create a node of type 3, which then creates nodes of types 1 and 2 at the printing stage.

yguclu:
Let's try to define a concrete example. We would like to translate the Python program
```
import numpy as np

i = 5
j = 7

a = np.ones (i, dtype=int)
b = np.zeros(j, dtype=int)

# Raises an error: incompatible redefinition
# file pyccel/parser/semantic.py, line 1466
b = 3 * a
```
to the Fortran program
```
program prog_ex_program

implicit none

integer(kind=8) :: i
integer(kind=8) :: j
integer(kind=8), allocatable :: a(:)
integer(kind=8), allocatable :: b(:)

i = 5_8
j = 7_8
allocate(a(0:i - 1_8))
a(:) = 1
allocate(b(0:j - 1_8))
b(:) = 0

! Raises an error: incompatible redefinition
! file pyccel/parser/semantic.py, line 1466
if (any(shape(b) /= shape(a))) then
    print *, "PYCCEL RUNTIME WARNING: array b is being reallocated"
    deallocate(b)
    allocate(b(0:size(a,1)-1))
[...]

PR #437, opened 2020-09-21
Fix broken codegen tests

### LINKED ISSUE #245: Automatic array allocation via type inference
The following Python code
```python
@types('real[:], real[:], real[:]')
def array_real_loopdiff(x, y, out):
    dxy = x - y
    for k in range(len(x)):
        out[k] = dxy[k]
```

generates Fortran code where `real(kind=8), allocatable :: dxy (:) ` is not started with zero indexing, namely

```Fortran
subroutine array_real_loopdiff(x, y, out) 

  implicit none
  real(kind=8), intent(in)  :: x (0:)
  real(kind=8), intent(in)  :: y (0:)
  real(kind=8), intent(inout)  :: out (0:)
  real(kind=8), allocatable :: dxy (:) 
  integer(kind=8) :: k  

  dxy = x - y
  do k = 0, size(x,1) - 1_8, 1
    out(k) = dxy(k)
  end do

end subroutine
```

This leads to failure. Have added according case in `tests/epyccel/test_arrays.py` in branch `test_fffi` in d5cd435fc71fe18d4cbf185994bb2771a3335250.

### LINKED ISSUE #436: Several codegen tests fail in the semantic step
**Describe the bug**
As the codegen tests test only if the code did not fail they are not checking whether the sequence of operations is actually possible. As a result 5 tests are currently failing but pass on jenkins

**To Reproduce**
The following files in tests/codegen/fcode/scripts/ fail in the semantic stage:
- context.py
- decorators_elemental.py
- decorators_types.py
- functions.py
- imports.py (due to functions.py)

This means that we have broken behaviour which is hidden

EmilyBourne:
Ensure that codegen tests fail if there is a problem during their treatment, even if the problem creates a neat error (fixes #436 ). Fix broken tests. Improve `broadcast` function by supposing that mathematical expressions are valid. As a result when both shapes are unknown in a dimension they are supposed to be the same, when only one shape is unknown then it is supposed that it is the same as the known shape (fixes #245).

- Stop codegen tests passing when errors have been raised
- Print filename with error (useful for errors in imported files)
- Allow broadcast function to handle PyccelArraySize
- Correct handling of elemental functions
- Add end to end test for elemental
- Do not complain about dtype differences in an AugAssign
- Assume that unknown sizes on the same axis are the same size
- When one of the sizes is unknown, prefer the known size

### DIFF
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -206,6 +206,10 @@ def broadcast(shape_1, shape_2):
             new_shape.append(e1)
         elif isinstance(e1, PyccelArraySize) and isinstance(e2, PyccelArraySize):
             new_shape.append(e1)
+        elif isinstance(e1, PyccelArraySize):
+            new_shape.append(e2)
+        elif isinstance(e2, PyccelArraySize):
+            new_shape.append(e1)
         else:
             msg = 'operands could not be broadcast together with shapes {} {}'
             msg = msg.format(shape_1, shape_2)
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -204,6 +204,8 @@ def broadcast(shape_1, shape_2):
             new_shape.append(e2)
         elif e2 == 1:
             new_shape.append(e1)
+        elif isinstance(e1, PyccelArraySize) and isinstance(e2, PyccelArraySize):
+            new_shape.append(e1)
         else:
             msg = 'operands could not be broadcast together with shapes {} {}'
             msg = msg.format(shape_1, shape_2)
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -262,7 +262,7 @@ class PyccelOperator(Expr, PyccelAstNode):
             if None in ranks:
                 self._rank  = None
                 self._shape = None
-            elif all(not (sh is None or isinstance(sh, PyccelArraySize)) for tup in shapes for sh in tup):
+            elif all(sh is not None for tup in shapes for sh in tup):
                 if len(args) == 1:
                     shape = args[0].shape
                 else:
@@ -317,7 +317,7 @@ class PyccelDiv(PyccelOperator):
             self._rank  = None
             self._shape = None
 
-        elif all(not (sh is None or isinstance(sh, PyccelArraySize)) for tup in shapes for sh in tup):
+        elif all(sh is not None for tup in shapes for sh in tup):
             shape = broadcast(args[0].shape, args[1].shape)
 
             for a in args[2:]:
@@ -353,7 +353,7 @@ class PyccelBooleanOperator(Expr, PyccelAstNode):
             self._rank  = None
             self._shape = None
 
-        elif all(not (sh is None or isinstance(sh, PyccelArraySize)) for tup in shapes for sh in tup):
+        elif all(sh is not None for tup in shapes for sh in tup):
             shape = broadcast(args[0].shape, args[1].shape)
             for a in args[2:]:
                 shape = broadcast(shape, a.shape)
--- a/tests/pyccel/scripts/decorators_elemental.py
+++ b/tests/pyccel/scripts/decorators_elemental.py
@@ -13,6 +13,5 @@ b = square(a)
 print(b)
 
 xs = array([1., 2., 3.])
-ys = zeros_like(xs)
 ys = square(xs)
 print(ys)
--- a/tests/pyccel/scripts/decorators_elemental.py
+++ b/tests/pyccel/scripts/decorators_elemental.py
@@ -1,0 +1,18 @@
+from pyccel.decorators import types, elemental
+from numpy import array
+from numpy import zeros_like
+
+@elemental
+@types(float)
+def square(x):
+    s = x*x
+    return s
+
+a = 2.0
+b = square(a)
+print(b)
+
+xs = array([1., 2., 3.])
+ys = zeros_like(xs)
+ys = square(xs)
+print(ys)
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -830,7 +830,7 @@ class Rand(Function, NumpyNewArray):
         code_init = 'call random_number({0})'.format(lhs_code)
         stmts.append(code_init)
 
-        return '\n'.join(stmts)
+        return '\n'.join(stmts) + '\n'
 
 #==============================================================================
 class NumpyRandint(Function, NumpyNewArray):
@@ -963,7 +963,10 @@ class Full(Application, NumpyNewArray):
             code_init = '{0} = {1}'.format(lhs_code, init_value)
             stmts.append(code_init)
 
-        return '\n'.join(stmts)
+        if len(stmts) == 0:
+            return ''
+        else:
+            return '\n'.join(stmts) + '\n'
 
 #==============================================================================
 class Empty(Full):
--- a/pyccel/codegen/printing/fcode.py
+++ b/pyccel/codegen/printing/fcode.py
@@ -1071,7 +1071,7 @@ class FCodePrinter(CodePrinter):
                 if lhs_name in vars_dict:
                     stack_array = vars_dict[lhs_name].is_stack_array
 
-            return rhs.fprint(self._print, expr.lhs, stack_array) + '\n'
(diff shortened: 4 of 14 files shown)


pyccel/ast/core.py
```diff
@@ -204,6 +204,12 @@ def broadcast(shape_1, shape_2):
             new_shape.append(e2)
         elif e2 == 1:
             new_shape.append(e1)
+        elif isinstance(e1, PyccelArraySize) and isinstance(e2, PyccelArraySize):
+            new_shape.append(e1)
+        elif isinstance(e1, PyccelArraySize):
+            new_shape.append(e2)
+        elif isinstance(e2, PyccelArraySize):
+            new_shape.append(e1)
```
EmilyBourne, line 212: Should warnings be raised here?

yguclu: I am not sure. Can you give me a Python example where we enter in one of these cases?

tests/pyccel/scripts/decorators_elemental.py
```diff
@@ -0,0 +1,18 @@
...
+    s = x*x
+    return s
+
+a = 2.0
+b = square(a)
+print(b)
+
+xs = array([1., 2., 3.])
+ys = zeros_like(xs)
+ys = square(xs)
```
yguclu, line 17: It appears that Pyccel can only process a function with an array result if it's elemental. But then we should not need to manually call `zeros_like` in order to allocate the array, Pyccel should be able to do it automatically. This is related to issue #359

yguclu (approved):

yguclu (approved):

pyccel/ast/core.py
EmilyBourne: This actually occurs in lots of places, for example in most of our array tests in epyccel/test_arrays.
E.g:
```
@types( 'int32[:]', 'int32[:]' )
def array_int32_1d_add( x, y ):
    x[:] += y
```
For this calculation we are assuming that x and y have the same size. Up till now this has been done by setting the size to `None` with the modification above this would be explicit. I think we will need it to be explicit in order to carry out this operation in C. In addition once this assumption is explicit then we know the size of c in the following function:
```
@types('int[:]','int[:]')
def print_add(a, b):
    c = a + b
    print(c)
```

EmilyBourne: I am not sure what should happen if we pass the function arrays of different sizes though. In numpy this would raise a `ValueError`. A short test shows that calling the following (translated):
```
a = np.ones(3)
b = np.ones(4)
print_add(a,b)
```
gives
```
2.0, 2.0, 2.0
```
and the following:
```
a = np.ones(4)
b = np.ones(3)
print_add(a,b)
```
gives
```
2.0, 2.0, 2.0, 1.0
```

tests/pyccel/scripts/decorators_elemental.py
EmilyBourne: I don't think we could return the result of an elemental call currently with epyccel, but I have made the modifications necessary to allow the allocation when it is called from fortran code

EmilyBourne: The allocation is done in the same place for both `a=b+c` and `ys = square(xs)` so they should be fixed in the same way when #359 is fixed

yguclu (approved):

tests/pyccel/scripts/decorators_elemental.py
yguclu: Perfect, that's all we need for now! Thanks

pyccel/ast/core.py
yguclu: We don't get any warnings from the Fortran compiler?

EmilyBourne: The fortran compiler has the same problem we do: it doesn't know the size of the arrays that will be passed to the function. There is nothing inherently wrong with the function:
```
subroutine test_add(a, b)

  implicit none

  real(kind=8), intent(in) :: a(0:)
  real(kind=8), intent(in) :: b(0:)
  real(kind=8), allocatable :: c(:)

  allocate(c(0:size(a, 1) - 1_8))
  c = a + b
  print *, c

end subroutine test_add
```
And I don't think it looks at the contents of the function when compiling the call

yguclu: Could you paste the generated Fortran code to this discussion?

EmilyBourne: Even with `-Wall` I don't get any errors and the code runs

yguclu: I see, then we have a potential memory bug. Try compiling with `-fcheck=all` and run the code again...

EmilyBourne: With `-fcheck=all` I have a runtime error:
```
Fortran runtime error: Array bound mismatch for dimension 1 of array 'b' (3/4)
```

EmilyBourne: I am not sure how we should handle this case. A very complicated but effective method would be to create "assumptions" for the arguments. These assumptions could then be checked during the translation or in the wrapper (when we no longer use f2py, or in the f2py_XX.f90 file).
E.g. for:
```
@types('int[:]','int[:]')
def print_add(a, b):
    c = a + b
    print(c)
```
We would generate the assumption `shape(a,1)==shape(b,1)`

If this is not true when we call the function then we can raise an error. In the cwrapper it is possible to raise python errors e.g. `ValueError` as numpy does

This is beyond the scope of this pull request though. Currently this PR reproduces the same behaviour as before it simply makes it explicit

yguclu: This makes a lot of sense in Fortran: one should run the test suite in debug mode, where the code is compiled with lots of compile- and runtime-checks. And then recompile it in release mode (with lots of optimization and few checks) for the production runs.

yguclu: So I think we can be happy of this fact: where Numpy gives a runtime error we also get a runtime error in Fortran, provided that we add the proper compilation flags to the compiler.

EmilyBourne: If we implement `assert` then you're solution is also possible. So for the following codes:
**Code A**
```
import numpy as np

@types('real[:]','real[:]')
def test_add(a,b):
    c = a + b
    print(c)

x = np.ones(3)
y = np.ones(4)

test_add(a,b)
```

**Code B**
```
import numpy as np

@types('real[:]','real[:]')
def test_add(a,b):
    c = a + b
    print(c)
```
we would generate an assumption `shape(a,1)==shape(b,1)` for the function `test_add`. We then have 2 options:
1. Code A generates a pyccel error: `"The function test_add is called with incompatible arguments"`. Code B compiles but the wrapper will raise a `ValueError`
2. Code A and B generate:
```
module prog

implicit none

contains

subroutine test_add(a, b)

  implicit none

  real(kind=8), intent(in) :: a(0:)
  real(kind=8), intent(in) :: b(0:)
  real(kind=8), allocatable :: c(:)

  ! TODO: How to disable this check in release?
  if (shape(a,1) != shape(a,2)) then stop 'Assertion Error'

  allocate(c(0:size(a, 1) - 1_8))
  c = a + b
  print *, c

end subroutine test_add

end module prog
```

I suspect the best answer is a hybrid where we do option 2 but we still check the assumptions in the cwrapper to raise python errors rather than having a Fortran runtime error which could cause problems for an ipython instance for example.
If you are happy with this method then I can create an issue which can be tackled when someone has lots of free time

EmilyBourne: We already add the necessary flag (https://github.com/pyccel/pyccel/blob/master/pyccel/codegen/utilities.py#L59):
```
        if debug:
            flags += " -fbounds-check"
```

yguclu: > ```python
>         if debug:
>             flags += " -fbounds-check"
> ```

I think that's an old deprecated flag. The new way to do this is `-fcheck=bounds`.

yguclu: > I suspect the best answer is a hybrid where we do option 2 but we still check the assumptions in the cwrapper to raise python errors rather than having a Fortran runtime error which could cause problems for an ipython instance for example.
> If you are happy with this method then I can create an issue which can be tackled when someone has lots of free time

Yes, I think that makes sense. But a deeper discussion may be needed. Please create a dedicated issue where you can copy-paste from this thread as much as you want.

yguclu (approved):

PR #445, opened 2020-09-28
Pow and mod operators

### LINKED ISSUE #394: Can't translate a function with no arguments to c

### LINKED ISSUE #412: PyccelMod and PyccelPow are not implemented in c
**Describe the bug**
PyccelMod and PyccelPow are not implemented in c. This is because they are slightly more complicated than other basic operators as they require the `math.h` header

**Test Code**
```python
a = 3
b = 2
c = a % b
d = a ** b
```

**Language**
C

rakati:
Implement of PyccelMod and PyccelPow:

- Add printers for both operators in C (closes #412);
- Fix casting in the Fortran printer of these operators;
- Fix missing minus in the pow exponent operand;
- Add the necessary tests.

Co-authored-by: EmilyBourne <louise.bourne@gmail.com>


pyccel/codegen/printing/ccode.py
```diff
@@ -208,8 +208,37 @@ def _print_PyccelNot(self, expr):
...
+        if expr.args[0].dtype is NativeInteger():
+            base = self._print(PythonFloat(expr.args[0]))
+        else:
+            base = self._print(expr.args[0])
+        if expr.args[1].dtype is NativeInteger():
+            e = self._print(PythonFloat(expr.args[1]))
+        else:
+            e = self._print(expr.args[1])
+        if expr.dtype is NativeInteger():
+            return '(long)pow({}, {})'.format(base, e)
```
EmilyBourne, line 237: Please use cast functions to cast

pyccel/codegen/pipeline.py
```diff
@@ -240,7 +240,7 @@ def get_module_dependencies(parser, mods=(), folders=()):
         # TODO: stop at object files, do not compile executable
         #       This allows for properly linking program to modules
         #
-        if not (language == "c" and codegen.is_module):
+        if language == "fortran":
```
EmilyBourne, line 243: This will stop c programs compiling

EmilyBourne (changes requested):
Do `**` and `%` work for complexes in python? If so then complexes should be added

nhamidn:
i think an error or a warning must be raised for this example : 

```
def modulo_bool():
    x = 10
    return x % False
```

pyccel/codegen/printing/ccode.py
EmilyBourne: Sorry, I've just realised that it is difficult (impossible?) to use the casting function here as it is around the entire object

EmilyBourne:
> Do `**` and `%` work for complexes in python? If so then tests for complexes should be added

The modulo operator for complexes raises:
```
<ipython-input-1-1c439d23f603> in <module>()
----> 1 1+2j%5+2j

TypeError: can't mod complex numbers.
```

However `**` exists for complex numbers

pyccel/codegen/pipeline.py
rakati: yes, We talk about this. I just forget to restore it back.

pyccel/codegen/printing/ccode.py
EmilyBourne, line 237: Shouldn't the c type depend on the precision as determined from the operators? Why not use the `find_in_dtype_registry` function?

rakati:
> i think an error or a warning must be raised for this example :
> 
> ```
> def modulo_bool():
>     x = 10
>     return x % False
> ```
Yes, zero division error.
It is undefined behavior in c.
So this case needs specific handling I mean each modulo operation we should check
for zero-division error. which is a big change.
I think we should discuss that in our meeting.

tests/epyccel/test_epyccel_pow.py
```diff
@@ -0,0 +1,110 @@
...
+
+    f = epyccel(pow_c_i, language=language)
+    b = complex(rand(), rand())
+    e = randint(10)
+    assert(isclose(f(b, e), pow_c_i(b, e), rtol=1e-15, atol=1e-15))
+    assert(isclose(f(-b, e), pow_c_i(-b, e), rtol=1e-15, atol=1e-15))
+    assert(isclose(f(b, -e), pow_c_i(b, -e), rtol=1e-15, atol=1e-15))
+    assert(isclose(f(-b, -e), pow_c_i(-b, -e), rtol=1e-15, atol=1e-15))
+
+def test_pow_c_r(language):
```
EmilyBourne, line 99: Could you add a `test_pow_r_c` to check complex exponents without a complex base

EmilyBourne (changes requested):
Please fix codacy changes

tests/epyccel/test_epyccel_pow.py
rakati: the test fails. we should cast base to complex.
I tried to use PythonComplex but it doesn't work.
any suggestion?
or should I cast it manually?

EmilyBourne: As discussed multiple times before it is better to avoid manual casting wherever possible.
The solution depends on why it is not working

EmilyBourne: Are you sure you wrote the test correctly? It passes on my computer

rakati: yes, the test is correct. Fortran passed it but c no.
I just pushed the new test case.

EmilyBourne: Your tests are passing on travis

EmilyBourne (approved):

tests/epyccel/test_epyccel_pow.py
rakati: the worst thing to have is this situation.
the test fails from time to time.
I am now testing this special cases 
cpow(-(0.8848429441339932+0.7027477315056795j), 7)

EmilyBourne: This is true. Have you modified the test at all since it failed? Do you remember what the error message said?

rakati: No, the worst. there is a false result.
```language = 'c'

    def test_pow_c_i(language):
        @types('complex', 'int')
        def pow_c_i(x, y):
            return x ** y
    
        f = epyccel(pow_c_i, language=language)
        b = complex(rand(), rand())
        e = randint(10)
        assert(isclose(f(b, e), pow_c_i(b, e), rtol=1e-15, atol=1e-15))
        assert(isclose(f(-b, e), pow_c_i(-b, e), rtol=1e-15, atol=1e-15))
>       assert(isclose(f(b, -e), pow_c_i(b, -e), rtol=1e-15, atol=1e-15))
E       assert False
E        +  where False = isclose((3094888.768776493-1342386.774745199j), (3094888.7687764983-1342386.774745202j), rtol=1e-15, atol=1e-15)
E        +    where (3094888.768776493-1342386.774745199j) = <built-in function pow_c_i>((0.18802209340404008+0.00855573871939952j), -9)
E        +    and   (3094888.7687764983-1342386.774745202j) = <function test_pow_c_i.<locals>.pow_c_i at 0x7f989e895b80>((0.18802209340404008+0.00855573871939952j), -9)

test_epyccel_pow.py:96: AssertionErro```
the error I mentioned yesterday was when I tried to use casting with PythonComplex.
I thought the false result was because we didn't use casting.

Comment (EmilyBourne) [reply tests/epyccel/test_epyccel_pow.py:99]: If you think the fix is to convert the base to a complex then you need to run:
`git cherry-pick a3fe3ea20f484fefbad158992915a8f80886d3dc`
in order to fix the problem you had before with `PythonComplex`. But I don't think that this is necessary to solve the problem as there are currently tests passing which do not do the cast.

>cpow(-(0.8848429441339932+0.7027477315056795j), 7)

This is not a `test_pow_r_c` or a `test_pow_i_c`. Which is failing?

Comment (EmilyBourne) [reply tests/epyccel/test_epyccel_pow.py:99]: Ah yes, that looks like a precision error. You are probably right that converting to complex might fix it.
Luckily the error message gives you the input values so you can test this failing case:
`(0.18802209340404008+0.00855573871939952j), -9`

File: tests/epyccel/test_epyccel_mod.py [hunk 4]
```diff
@@ -0,0 +1,68 @@
...
+from conftest import *
+
+def test_modulo_int_int(language):
+    @types(int, int)
+    def modulo_i_i(x, y):
+        return x % y
+
+    f = epyccel(modulo_i_i, language=language)
+    x = randint(1e6)
+    y = randint(1e6) + 1
```
yguclu, line 16: It's a bit strange to choose `x` and `y` this way. I would suggest to choose `y`  much smaller than `x`, so that the result of the modulo operation is less obvious.

By the way, can you explain why you're adding `1` to a random number?

tests/epyccel/test_epyccel_mod.py
```diff
@@ -0,0 +1,68 @@
...
+    assert isinstance(f(x, y), type(modulo_i_i(x, y)))
+
+def test_modulo_real_real(language):
+    @types('real', 'real')
+    def modulo_r_r(x, y):
+        return x % y
+
+    f = epyccel(modulo_r_r, language=language)
+    x = uniform(high=1e6)
+    y = uniform(high=1e6) + 1
```
yguclu, line 28: I have the same comment as above.

tests/epyccel/test_epyccel_pow.py
```diff
@@ -0,0 +1,123 @@
...
+    assert isinstance(f(x, y), type(f_call(x, y)))
+
+def test_pow_real_real(language):
+    @types('real', 'real')
+    def pow_r_r(x, y):
+        return x ** y
+
+    f = epyccel(pow_r_r, language=language)
+    x = rand()
+    y = rand()
```
yguclu, line 31: Here we are only testing with a base `0<= x < 1` and an exponent `0 <= y < 1`. Is there a reason for restricting the test to small positive numbers?

tests/epyccel/test_epyccel_pow.py
yguclu: Try increasing `rtol`  to `1e-14`...

tests/epyccel/test_epyccel_pow.py
```diff
@@ -0,0 +1,123 @@
...
+    def pow_r_c(x, y):
+        return x ** y
+
+    f = epyccel(pow_r_c, language=language)
+    b = rand()
+    e = complex(rand(), rand())
+    assert(isclose(f(b, e), pow_r_c(b, e), rtol=1e-15, atol=1e-15))
+    assert(isclose(f(-b, e), pow_r_c(-b, e), rtol=1e-15, atol=1e-15))
+    assert(isclose(f(b, -e), pow_r_c(b, -e), rtol=1e-15, atol=1e-15))
+    assert(isclose(f(-b, -e), pow_r_c(-b, -e), rtol=1e-15, atol=1e-15))
```
yguclu, line 123: No newline at end of file.

yguclu (changes requested):

tests/epyccel/test_epyccel_mod.py
EmilyBourne: The `+1` is to avoid having `y=0`. `x % 0` results in a `ZeroDivisionError`

tests/epyccel/test_epyccel_pow.py
EmilyBourne: > Try increasing `rtol` to `1e-14`...

This seems to be sufficient, at least for the example above

tests/epyccel/test_epyccel_pow.py
rakati: I think I need to use vim for that.

EmilyBourne: Are you sure? Can you not just add an empty line at the end of the file?

tests/epyccel/test_epyccel_pow.py
rakati: @EmilyBourne  I don't think so, c convert to the correct type.
check this [implicit conversion](https://en.cppreference.com/w/c/language/conversion)
the missing casting will just raise some warnings.
rtol 1e-14 works great.

EmilyBourne: > the missing casting will just raise some warnings.

If it raises warnings then it would be good to get rid of those. I don't seem to have warnings on my computer, but it has been known to keep quiet about casting

tests/epyccel/test_epyccel_pow.py
rakati: I used a small range to avoid overflow, and I think this range will be enough and fast in testing.
as you know pow operation consummate time especially for real numbers.
I forgot to add negative values for some tests, I will add them.

tests/epyccel/test_epyccel_pow.py
rakati: vim always adds a new line to the end of the file and removes trailing whitespace automatically by default.
he will save a lot of time.
or I will config my vscode to that.

saidctb (approved):

tests/epyccel/test_epyccel_pow.py
rakati: negative cases show me a lot of interesting cases, where the c and Fortran behave in a different manner.

nhamidn (approved):

tests/epyccel/test_epyccel_pow.py
yguclu: Well, I only use vim so I am not impartial here 😆

tests/epyccel/test_epyccel_pow.py
yguclu: Can we test at least those cases where Python, C and Fortran behave the same way? For example, if `M > 1` is a real number and`N > 1` an integer (not too big to avoid overflows):
1. Real base `-M < x < M`, and positive integer exponent `y = 0, 1, 2, ..., N` (be careful to avoid the case `0**0` though);
2. Positive real base `0 < x < M` (not zero), and real exponent `-M < y < M`.

rakati: yah, those are interesting cases I will add them.

EmilyBourne (approved):

yguclu (changes requested):
Very good! Please resolve the conflict with master so that we can proceed with the merge.

yguclu (approved):

yguclu:
There is one failing test on Windows.

I have restarted the build already but the error persists, it may be avoidable by using a larger error tolerance.

rakati:
> There is one failing test on Windows.
> 
> I have restarted the build already but the error persists, it may be avoidable by using a larger error tolerance.

yes, the error is in the pow test on math functions.
I reduced the rtol.

yguclu:
Codacy complains about some unused imports. I don't know why it didn't complain before

yguclu (approved):

yguclu (approved):

EmilyBourne:
> Codacy complains about some unused imports. I don't know why it didn't complain before

That's my fault sorry. I noticed some unused imports in a different branch that codacy wasn't flagging up. Apparently that test had been turned off so I turned it back on

mostafaMamoni (approved):

yguclu (approved):

PR #446, opened 2020-09-30
Use correct precision when printing integers

### LINKED ISSUE #326: Accessing an element of an IndexedBase after the Semantic step does not return a pyccel object
This occurs in tests/codegen/cross.py and is one of the reasons we cannot do:
```
def _print_Integer:
     return "{0}_{1}".format(str(expr.p), expr.precision)
```

EmilyBourne:
Use the precision property to print the correct precision. Casting functions are used where possible so this modification only needs to be done in one place. References to sympy object `IndexedBase` are removed.
A .codacy.yml file is also added to stop checking the tests folder for duplication. Duplication here is normal

- Use IndexedVariable instead of sympy's IndexedBase
- Print integer with correct precision. Fixes #326
- In ast/builtins.py, rename `Bool` -> `PythonBool` and `Int` -> `PythonInt`
- Use casting function in PyccelDiv (Fortran & C) and PyccelMod (Fortran)
- Use casting functions for NumpySqrt (Fortran) and MathSqrt (Fortran)
- Use casting functions where possible for PyccelFloorDiv (Fortran), provide precision elsewhere
- Use default precision for casting functions in C
- Remove print functions for Sympy objects
- [CODACY] Do not check for duplication in tests
- [COVERAGE] Add test to ensure allocatables are not used in an expression


saidctb (approved):

pyccel/codegen/printing/ccode.py
```diff
@@ -317,9 +319,10 @@ def _print_PyccelMul(self, expr):
         return ' * '.join(self._print(a) for a in expr.args)
 
     def _print_PyccelDiv(self, expr):
-        args = [self._print(a) for a in expr.args]
+        args = expr.args
         if all(a.dtype is NativeInteger() for a in expr.args):
-            return ' / '.join('real({})'.format(self._print(a)) for a in args)
+            args = [PythonFloat(a) for a in args]
+        args = [self._print(a) for a in args]
         return  ' / '.join(self._print(a) for a in args)
```
yguclu, line 326: In this function `args` is redefined multiple times, but there is no reason to do so. Moreover, the printer processes the elements of `args` twice. Instead, you could write it as follows:
```python
def _print_PyccelDiv(self, expr):

    if all(a.dtype is NativeInteger() for a in expr.args):
        args = [PythonFloat(a) for a in expr.args]
    else:
        args = expr.args

    return  ' / '.join(self._print(a) for a in args)
```

yguclu (changes requested):

pyccel/codegen/printing/ccode.py
```diff
@@ -108,11 +108,13 @@ def _traverse_matrix_indices(self, mat):
...
         value = self._print(expr.arg)
-        return '(double)({0})'.format(value)
+        type_name = self.find_in_dtype_registry('real', default_precision['real'])
+        return '({0})({1})'.format(type_name, value)
 
     def _print_Int(self, expr):
         value = self._print(expr.arg)
-        return '(long)({0})'.format(value)
+        type_name = self.find_in_dtype_registry('int', default_precision['int'])
+        return '({0})({1})'.format(type_name, value)
```
yguclu, line 117: I wonder if class `ast.builtins.Int` should really be called `PythonInt` to mimic the distinction that we currently have between `ast.numbers.Float` and `ast.builtins.PythonFloat`.

yguclu (approved):

pyccel/codegen/printing/ccode.py
EmilyBourne: It looks like most of the time we are writing `from ast.builtins import Int as PythonInt` anyway so I think that's a good idea

yguclu (approved):

pyccel/codegen/printing/ccode.py
yguclu: Thanks!

yguclu (approved):

pyccel/ast/datatypes.py
```diff
@@ -76,8 +76,9 @@
                                  'int32':('int',4),
                                  'int64':('int',8),
                                  'int'  :('int', default_precision['int']),
+                                 'pythonint'  :('int', default_precision['int']),
```
yguclu, line 79: Shouldn't `PythonInt` replace `Int` here?

EmilyBourne: I think there are a few `int` s in the code too. We don't want `Int`, but `int` is a little less troublesome.
I tried this on my machine and I had `KeyError`s in the functions `test_array_int_2d_C_initialization`, `test_array_int_2d_F_initialization`, `test_array_real_1d_complex_3d_expr`, `test_array_real_2d_C_complex_3d_expr`, `test_array_real_2d_F_complex_3d_expr` and `test_array_kwargs_full` in `tests/epyccel/test_arrays.py`

EmilyBourne: Actually I have just done a little debugging on this and apparently `int` is used when we pass a `NativeInteger` to `process_dtype`. So either this is a non-issue. Or we should modify `process_dtype` to handle this case separately using `default_precision` instead of `dtype_and_precision_registry`

EmilyBourne: Whichever we choose I will need to do the same for `NativeBool` before the PR can be merged

EmilyBourne: It is also used when the user uses a string to specify the dtype of a numpy array. E.g:
```
tmp = array([[1, 2, 3], [4, 5, 6]], dtype='int', order='F')
```
We therefore need to retain these entries

yguclu: I see, so there seems to be some confusion in the way we use `dtype_and_precision_registry`. Should we meet and discuss about it?

EmilyBourne: The dictionary `dtype_and_precision_registry` is used to set the dtype and precision of a type specified in the grammar (so str->DataType) and in the function `process_dtype` in `ast/numpyext.py` to collect the dtype from the provided argument (so str->DataType (`tmp = array([[1, 2, 3], [4, 5, 6]], dtype='int', order='F')`), CastFunction->DataType (`tmp = array([[1, 2, 3], [4, 5, 6]], dtype=int, order='F')` ) or DataType->DataType (`tmp = array([[1, 2, 3], [4, 5, 6]], order='F')`)
This seems fairly consistent to me and means that it is completely normal to have int  in the list as well as PythonInt

PR #434, opened 2020-09-21
Activate math tests in C

saniQQ:
Implement Math and Numpy functions support in C language (closes #381):
 - Create printer method for Python Math functions:
       - We now support all functions that are available in C math standard library;
       - Other functions in Python Math library will be implemented after resolution of #431 in new PR;
 - Create printer method for Python Numpy functions:
       - Same as Math functions, not all functions are supported yet;
       - Numpy functions support only scalars, the array support will be added after resolution of #312 in new PR.

Moreover:
- Add some Math functions that already exist in the Fortran language to the Fortran printer;
- Add tests for both Numpy and Math functions.

NOTE: complex numbers are not covered in this PR.

PR #448, opened 2020-10-01
Add print function to C printer

rakati:
Add print function to C printer and improve Fortran print:

- Python instruction `print(x)` is now correctly translated to C if  `x` is of type `bool`, `int`, `str`, `complex`, or `float`;
- Special characters in string are now supported for both Fortran and C;
- Fix newline inside a string in Fortran;
- Add unit tests.

Notes:
- Support for printing variables of other types will be added when they are implemented in C (arrays, lists, tuples, ...).


pyccel/codegen/printing/ccode.py
```diff
@@ -209,7 +210,45 @@ def _print_PyccelNot(self, expr):
...
+        args_format = []
+        end = '\\n'
+        sep = ' '
+        for f in expr.expr:
+            if isinstance(f, ValuedVariable):
+                if f.name == 'sep'      :   sep = str(f.value)
+                elif f.name == 'end'    :   end = str(f.value)
+            elif f.dtype is NativeBool():
+                args_format.append("%s")
+                args.append('({}) ? "True" : "False"'.format(self._print(f)))
```
EmilyBourne, line 230: Ideally it would be better to use `IfTernaryOperator` (the print method needs to be implemented)

pyccel/codegen/printing/ccode.py
```diff
@@ -209,7 +210,45 @@ def _print_PyccelNot(self, expr):
...
+            elif f.dtype is NativeInteger():
+                args_format.append("%ld")
```
EmilyBourne, line 232: `%ld` is for printing objects of type `long` if we use an `int32` or similar then we will have a different precision and we will no longer have a long

pyccel/codegen/printing/ccode.py
```diff
@@ -209,7 +210,45 @@ def _print_PyccelNot(self, expr):
...
+                args.append(self._print(f))
+            elif f.dtype is NativeReal():
+                args_format.append("%lf")
```
EmilyBourne, line 235: See above `%lf` is for `double`

pyccel/codegen/printing/ccode.py
```diff
@@ -209,7 +210,45 @@ def _print_PyccelNot(self, expr):
...
+                args.append(self._print(f))
+            elif f.dtype is NativeString():
+                args_format.append("%s")
+                args.append(self._print(f))
+            elif f.dtype is NativeComplex():
+                args_format.append("(%f + %fj)")
```
EmilyBourne, line 241: If you make the `if` block in this `for` loop into a function returning the arg_format and the arg then you could call the function recursively to set the format of the real and imaginary parts of the complex.

EmilyBourne, line 241: Why do we use `%f` for a real here, but `%lf` for a real above?

pyccel/codegen/printing/ccode.py
```diff
@@ -209,7 +210,45 @@ def _print_PyccelNot(self, expr):
...
+                complex_n = self._print(f)
+                args.append('creal({}), cimag({})'.format(complex_n, complex_n))
```
EmilyBourne, line 243: It would be better to use the printer of the functions `Real` and `Imag` (see https://github.com/pyccel/pyccel/blob/master/pyccel/ast/cwrapper.py#L242-L243)

pyccel/codegen/printing/ccode.py
```diff
@@ -209,7 +210,45 @@ def _print_PyccelNot(self, expr):
...
+            else :
+                raise ("{} type is not supported currently".format(f.dtype))
+
+        args_format = sep.join(args_format)
+        args_format += end
+        args_format = '"{}"'.format(args_format)
+        code = ', '.join([args_format, *args])
+        return self._get_statement("printf({})".format(code)) + '\n'
```
EmilyBourne, line 251: The `'\n'` is added in `_print_CodeBlock`

pyccel/codegen/printing/ccode.py
rakati: I am confused now. As far as I know, we use long for `NativeInteger`.
I was thinking of using a `dtype_registry` with `arg_format`. but we use `arg_format` only in `_print_print`.
what do you suggest to use?

saidctb: yes, I think we always have a `long` but the definition of `long` changes from an architecture to another it's sometimes `int64` and sometimes `int32`

rakati: yes, these changes will affect precision. but I think `printf` has nothing to do with that.

EmilyBourne: I don't think that's right. I think we use long for (`NativeInteger`, precision=8). The default precision for integers is 8 (on most machines) so we almost always use `long`, however if we use numpy types then this is no longer the case. E.g. the following code:
```
@types('int16', 'int16')
def test_add(x, y):
    return x + y
```
is translated to:
```
short int test_add(short int x, short int y)
{
    short int Out_0001;
    Out_0001 = x + y;
    return Out_0001;
}
```
If you use `%ld` to print `x` then I think you should get a warning.

Now that PR #434 has been merged, if you merge the master branch in then you should be able to see this warning if you create tests using numpy types

EmilyBourne: A dictionary sounds like a good idea to me. If it is only relevant for the function `_print_Print` then you can define it in that function rather than at the global scope

rakati: In this case, we have another type that we should support with `%hd`.
I will try to improve the code format for easy including other types.

EmilyBourne: You can take a look at the `dtype_registry` dictionary to see the currently implemented types. The first argument is one of `_print(NativeInteger())`, `_print(NativeReal())`, `_print(NativeBool())`, or `_print(NativeComplex())`, the second is the precision.

If you call pyccel with the argument `--flags="-Wall"` in your test then you can make sure that an error is raised if the format string is wrong

pyccel/codegen/printing/ccode.py
rakati: done.

EmilyBourne: Personally I think this is dangerous. It may break any code using the words `True` or `False`, e.g:
```
True_bool = True
```
in c:
```
True_bool = True;
```
After preprocessing:
```
1_bool = 1;
```
I am against adding even more restrictions to the naming which is already difficult (see issue #227 )

EmilyBourne: > but we can't use it here, it has no meaning to construct an IfTernaryOperator from a value or a condition.

I'm not entirely sure what you mean but looking at the code it looks like `IfTernaryOperator` is implemented in a strange way, probably because fortran does not have this facility. Does your implementation break the usage of `IfTernaryOperator`? I'm not sure where it is tested for fortran. For c it is used when collecting boolean arguments (see ast/cwrapper.py:L223)

EmilyBourne: P.S. If you edit your comment instead of adding a new comment I do not receive an alert and am therefore unlikely to reply. Edit should be used for typos

EmilyBourne: > probably because fortran does not have this facility

This is not true, it can be implemented in fortran using `merge`. I would therefore be in favour of modifying `IfTernaryOperator` so it no longer inherits from `If` and keeps its original form after `_visit_Assign` in parser/semantic.py. However this is completely out of the scope of this pull request. I will create an issue for it.

Given that fact, I am happy with your original code

EmilyBourne: > Does your implementation break the usage of `IfTernaryOperator` ?

If so then your implementation of `_print_IfTernaryOperator` should be removed from this PR. You could move it to a new branch which can then be used by whoever wants to fix the `IfTernaryOperator`

rakati: `True_bool = True;` will be `True_bool = 1;` not `1_bool = 1;.`
defines don't process subwords.
to see that try to compile your example by `gcc -E code.c`. This will output the preprocessed source code.

rakati: okay. I will remove it.

EmilyBourne: > defines don't process subwords.

You are right, sorry. In that case I would be interested to know what @yguclu and @saidctb think about using macros for this as `True` and `False` will be reserved anyway

saidctb: I agree to use  macros in this case because it will make the code more readable

tests/pyccel/runtest_print.py
```diff
@@ -0,0 +1,45 @@
+# from math import (inf, nan)
```
EmilyBourne, line 1: Please save pyccel test files in the `scripts/` folder

tests/pyccel/test_pyccel.py
```diff
@@ -442,3 +439,6 @@ def test_multiple_results():
 #------------------------------------------------------------------------------
 def test_elemental():
     pyccel_test("scripts/decorators_elemental.py")
+
+def test_print(language):
+    pyccel_test("runtest_print.py", language=language, output_dtype=str)
```
EmilyBourne, line 444: It is better to use a dtype for each line to avoid problems with the precision of floats and complexes. If you group the outputs by type then you can use `[str]*nlines` to avoid typing the same thing over and over (see `test_expressions`)

pyccel/codegen/printing/ccode.py
```diff
@@ -368,6 +380,47 @@ def _print_PyccelPow(self, expr):
...
+        sep = ' '
+        for f in expr.expr:
+            if isinstance(f, ValuedVariable):
+                if f.name == 'sep'      :   sep = str(f.value)
+                elif f.name == 'end'    :   end = str(f.value)
+            else:
+                try:
+                    args_format.append(type_to_format[(self._print(f.dtype), f.precision)])
+                except KeyError:
+                    raise ("{} type is not supported currently".format(f.dtype))
```
EmilyBourne, line 410: Please use `errors.report` to raise errors. This will ensure that we raise a `PyccelError` rather than a `KeyError` and it will potentially allow us to provide more information about the location of the problem in the future

tests/pyccel/runtest_print.py
rakati: done.

tests/pyccel/test_pyccel.py
rakati: yes, I am using now dtype. but I made some changes to how strings being compared.
I did this change to be able to test the Fortran print strings.

pyccel/codegen/printing/ccode.py
rakati: done.

tests/pyccel/test_pyccel.py
```diff
@@ -110,7 +107,9 @@ def get_value(string, regex, conversion):
 def compare_pyth_fort_output_by_type( p_output, f_output, dtype=float ):
 
     if dtype is str:
-        assert(p_output.strip()==f_output.strip())
+        p_list = p_output.split()
+        f_list = f_output.split()
```
EmilyBourne, line 111: Is the splitting still needed even with the concatenations used above?

tests/pyccel/test_pyccel.py
```diff
@@ -324,8 +323,9 @@ def test_folder_imports():
     compare_pyth_fort_output(pyth_out, fort_out)
 
 #------------------------------------------------------------------------------
-def test_funcs(language):
-    pyccel_test("scripts/runtest_funcs.py", language = language)
+# it won't work until PR 462 goes through
+# def test_funcs(language):
+#     pyccel_test("scripts/runtest_funcs.py", language = language)
```
EmilyBourne, line 328: Please deactivate this test for c (by removing the `language` argument) or mark it as `xfail`. Don't remove the fortran test

tests/pyccel/test_pyccel.py
```diff
@@ -442,3 +442,19 @@ def test_multiple_results():
 #------------------------------------------------------------------------------
 def test_elemental():
     pyccel_test("scripts/decorators_elemental.py")
+
+def test_print_strings(language):
+    types = [str]
```
EmilyBourne, line 447: This does not need to be a list if all return types are the same

EmilyBourne (changes requested):

pyccel/codegen/printing/ccode.py
```diff
@@ -281,6 +285,14 @@ def _print_If(self, expr):
...
         return "\n".join(lines)
 
+    def _print_IfTernaryOperator(self, expr):
+        condition, if_exp   = expr.args[0]
+        _, else_exp         = expr.args[1]
+        condition           = self._print(condition)
+        if_exp              = self._print(if_exp)[:-1]
+        else_exp            = self._print(else_exp)[:-1]
+        return "({}) ? ({}) : ({});".format(condition, if_exp, else_exp)
+
```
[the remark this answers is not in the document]
rakati: yes, I am sorry I forgot to remove it.

tests/pyccel/test_pyccel.py
rakati: okay.

tests/pyccel/test_pyccel.py
rakati: yes, Fortran adds spaces at the beginning of the printed line so without split string will not match.

EmilyBourne: The original command used `strip()` which removes leading and trailing whitespace. This should therefore deal with spaces added at the beginning of the line while still letting you test the whitespace within the line

tests/pyccel/scripts/expressions.py
```diff
@@ -14,15 +16,15 @@
 a10 = 1.*10j
 
 b1 = 1-1-2-4-5
-b2 = 1-1-2j-4.-5
-b3 = 1-1-2j-5
+# b2 = 1-1-2j-4.-5
+# b3 = 1-1-2j-5
```
EmilyBourne, line 20: Please mark failing tests as `xfail` rather than removing them. Removing the test makes it easier for us to forget that this is broken.
Worse, if this is only broken for one language then it can be broken for another language in the future without us noticing

EmilyBourne (changes requested):

pyccel/codegen/printing/ccode.py
```diff
@@ -285,14 +285,6 @@ def _print_If(self, expr):
             lines.append("%s\n}" % var)
         return "\n".join(lines)
 
-    def _print_IfTernaryOperator(self, expr):
-        condition, if_exp   = expr.args[0]
-        _, else_exp         = expr.args[1]
-        condition           = self._print(condition)
-        if_exp              = self._print(if_exp)[:-1]
-        else_exp            = self._print(else_exp)[:-1]
-        return "({}) ? ({}) : ({});".format(condition, if_exp, else_exp)
```
EmilyBourne: Is this code in a branch somewhere? If not then it can be copy/pasted into the issue so it doesn't need to be written twice

rakati: the `IfTernaryOperator` will not inherit anymore from the `if` class. so I think this code will not be useful.

EmilyBourne: Ah good point :+1:

tests/pyccel/scripts/expressions.py
rakati: If I mark this as `xfail` then other expressions will not be tested.
and in the future, someone may break these passed tests without knowing about them, instead of having a few expressions fails, so the problem gets bigger, 
maybe separate them from `expression.py` then `xfail` them in c.
what do you think?

EmilyBourne: That's true but I think most (all?) of the basic things have tests in the `epyccel` folder. If everything is tested there then an `xfail` is sufficient. If not then I agree the best thing would be to separate them or move these tests to `epyccel` so the tests can be more granular

tests/pyccel/test_pyccel.py
rakati: compare string result is tricky. not like numbers.
actually code should be something similar to that.
```python
if dtype is str:
        p_list = p_output.split("\n")
        f_list = f_output.split("\n")
        # get firsl lines to compare
        p = p_list[0]
        f = f_list[0]
        # return the rest of the output to compare in next compare type
        p_output = "\n".join(p_list[1:] if len(p_list) > 1 else "")
        f_output = "\n".join(f_list[1:] if len(f_list) > 1 else "")
        assert(p.strip() == f.strip())
```
this can be easy to work with in case of printing one line of string.
and also can be a complex thing if you have a string with newlines inside, you should know the numbers of lines in the resulting output.
but what about this case.
```python
       print("hello world", 2020)
       #output in Fortran
       "hello world                  2020"
```
here for correct output, we should split with space. but if we do that then we should now how many words in the string.
so splitting this by whitespace and do one list comparison is better.

tests/pyccel/scripts/expressions.py
rakati: done.

tests/pyccel/test_pyccel.py
EmilyBourne: Ok. I thought the output was already split at `\n` but it seems not so then you are right that `split()` is needed

tests/pyccel/test_pyccel.py
```diff
@@ -336,13 +335,27 @@ def test_bool():
...
-    pyccel_test("scripts/expressions.py",
-                output_dtype = types)
+            [float]*2 + [float]*2 + [int]*5 + [bool]*9
+    pyccel_test("scripts/expressions.py", language=language,
+                pyccel_commands="--libs m", output_dtype = types)
+
+#------------------------------------------------------------------------------
+@pytest.mark.parametrize( 'language', (
+        pytest.param("c", marks = [
+            pytest.mark.xfail(reason="Bug: need to be fixed in c"),
```
EmilyBourne, line 349: Could you open an issue for this please

tests/pyccel/test_pyccel.py
```diff
@@ -442,3 +455,21 @@ def test_multiple_results():
...
+#------------------------------------------------------------------------------
+def test_print_strings(language):
+    types = str
+    pyccel_test("scripts/print_strings.py", language=language, output_dtype=types)
+
+#------------------------------------------------------------------------------
+@pytest.mark.parametrize( 'language', (
+        pytest.param("c", marks = pytest.mark.c),
+        pytest.param("fortran", marks = [
+            pytest.mark.xfail(reason="formated string not implemented in fortran"),
```
EmilyBourne, line 468: See above

EmilyBourne (approved):

EmilyBourne (approved):

pyccel/codegen/printing/fcode.py
```diff
@@ -1193,7 +1195,18 @@ def _print_BooleanFalse(self, expr):
         return '.False.'
 
     def _print_String(self, expr):
-        formatted_str = expr.arg.replace("'","''")
+        # characters speciaux are not handled only newline
```
yguclu, line 1198: Should this comment read this instead?
```python
# special characters are not handled, only newline
```

pyccel/codegen/printing/fcode.py
```diff
@@ -1193,7 +1195,18 @@ def _print_BooleanFalse(self, expr):
...
+        formatted_str = expr.arg.replace("'","''")\
+                                .replace('\a', '\\a')\
+                                .replace('\b', '\\b')\
+                                .replace('\f', '\\f')\
+                                .replace('\r', '\\r')\
+                                .replace('\t', '\\t')\
+                                .replace('\v', '\\v')
```
yguclu, line 1205: By using the [`ACHAR`](https://gcc.gnu.org/onlinedocs/gfortran/ACHAR.html) Fortran function and this [ASCII table](https://en.wikipedia.org/wiki/Escape_sequences_in_C#Table_of_escape_sequences) it should be possible to handle all the special characters listed here. One first splits the string into substrings (either normal strings, or special characters), then replaces the special characters with their ASCII equivalent, and finally concatenates all substrings as done below for `new_line`.

Please try!

yguclu (changes requested):

pyccel/codegen/printing/fcode.py
```diff
@@ -1193,7 +1195,18 @@ def _print_BooleanFalse(self, expr):
...
+        if "\n" in formatted_str:
+            s_list = formatted_str.split("\n")
+            code = " // new_line('') // ".join("'{}'".format(s) for s in s_list)
```
EmilyBourne, line 1208: Does this work if `\n` is at the start of the line?

rakati: yes, it works.
python ex:
```python
print("\nazul")
```
Fortran code result
```fortran
program prog_sp

implicit none

print *, '' // new_line('') // 'azul'

end program prog_sp
```

rakati: done

pyccel/codegen/printing/fcode.py
rakati: done

tests/pyccel/scripts/print_sp_and_end.py
```diff
@@ -0,0 +1,18 @@
+# pylint: disable=missing-function-docstring, missing-module-docstring/
+#  ------------------------string with sep and end ---------------------------
+
+print('ahoy', 'world', sep=' ')
```
EmilyBourne, line 4: This may be beyond the scope of this PR. But it may also be a 1 line fix so I'm going to ask anyway:
Does the concatenation method not work for sep?
I.e:
```
(' // '+self._print(sep)+' // ').join(sub_strs)
```

rakati: hh, I thought it was unnecessary to add sep and end for Fortran.
the method you provide works only for strings. presence of a non string variable will cause a problem.
anyway. I will add support to that. it is easy.

EmilyBourne: > hh, I thought it was unnecessary to add sep and end for Fortran.
That's why I said it's beyond the scope of this PR
> the method you provide works only for strings. presence of a non string variable will cause a problem.
That's true

EmilyBourne:
I think the failure on windows is due to windows and linux handling consecutive delimiters differently. Try using `re.split`
See:
https://stackoverflow.com/a/6478890/5957195

rakati:
> I think the failure on windows is due to windows and linux handling consecutive delimiters differently. Try using `re.split`
> See:
> https://stackoverflow.com/a/6478890/5957195

I am not sure about that. because I see different outputs.
I don't know if Fortran print behaves in a different manner on windows, or that something related to special characters.
But I will give it a try.

EmilyBourne (approved):

saidctb (approved):

yguclu (approved):

Issue #433, opened 2020-09-17
Optional arguments not working in C


EmilyBourne:
**Describe the bug**
Optional arguments are arguments with a default value of `None`. They are currently not implemented in C

**Test Code**
Provide code which does not currently work but which should do when this issue is fixed:
```python
@types ('int')
def f (x = None):
    if x is None:
        return x+2
    else:
        return 0
```

**Proposed Solution**
A None value can be represented by a `NULL` pointer. This requires several changes to the code

1. Creation of VariableAddress objects should be moved to the codegen stage. This allows the codegen stage to decide which properties require this class (currently `is_pointer` and `is_optional`)
I suggest the creation of a function `stored_in_c_pointer` to determine whether a VariableAddress is needed. This function can also be used in `print_Variable` to decide whether to print the value access (`*`) or not

2. Temporary variables will have to be created if the object being passed as an optional argument is not stored in a pointer. Eg
```
@types ('int')
def f (x = None):
    if x is None:
        return x+2
    else:
        return 0

def h ():
    return 7

def g ():
    y = f (8)
    z = f (y+4)
    a = f (h ())
```

3. Add the collection of optional variables to the cwrapper. The variables should be collected into a PyObject* whose default value is PyNone. The c compatible variable should be a pointer initialised to NULL if PyNone is collected. Otherwise the required memory should be allocated and the value should be collected using a pyobject_to_XX function. The allocatedmemory will need to be freed after the function call

**Language**
C
This does not work in fortran either but for fortran it cannot be fixed without fixing issue #370

EmilyBourne:
@MenoIy do you want to have a go at this as you are familiar with the cwrapper?

PR #461, opened 2020-10-05
Translate Python functions with optional arguments to C

MenoIy:
Optional arguments are arguments with a default value of None. With this PR we implement them in C and solve #433 .
The management of optional arguments is added to the Cwrapper. Detailed changes:

- Create function **stored_in_c_pointer** to determine whether a VariableAddress is needed; the function takes a Variable as argument and returns True if it's a pointer or an optional variable;
- Move the creation of VariableAddress objects from the ast to codegen stage;
- Create Python constant, variable **Py_None** representing Python's None;
- Construct **PyArg_ParseTupleNode** flags from the collecting variable instead of the C variable;
- Add the collection of  optional arguments in **get_PyArgParseType** as Pyobject with default value **Py_None**;
- Change the collect order in **get_PyArgParseType** to prioritize optional arguments;
- Create function **optional_element_management**  responsible of optional arguments (creating tmp variable and the assignment body);
- Create function  **get_collect_function_call** that returns the needed collect or cast function;
- Create CPython function responsible of the collection of the value from a pyobject (_PyLong_AsLong_,  _PyFloat_AsDouble_, _pybool_to_bool_);
- Create List __to_free_PyObject_list_  to collect new pyobject created inside the wrapper function;
- Create CPython function **Py_DECREF**  to free the pyobject variable at the end of the wrapper;
- Comment line where optional argument implies optional result;
- Change condition in **is_printer**: should verify type before accessing property;
- Create function **PyType_Check** that create CPython function responsible of verifying the argument type;
- Create function **PyErr_SetString** to return string error when **PyType_Check** returns False;
- Update and create tests for optional arguments and mark them as xfail in Fortran.


pyccel/codegen/printing/ccode.py
```diff
@@ -286,19 +286,27 @@ def _print_FunctionDef(self, expr):
...
 
+    def stored_in_c_pointer(self, a):
+        if not isinstance(a, Variable):
+            return False
+        return a.is_pointer or a.is_optional
+
     def _print_FunctionCall(self, expr):
         func = expr.funcdef
+         # Ensure the correct syntax is used for pointers
+        args = [VariableAddress(a) if self.stored_in_c_pointer(a) else a for a in expr.arguments]
```
EmilyBourne, line 297: `VariableAddress` is not needed for every argument. It should only be needed occasionally.
Eg if the function signature is
`void f(int a)`
Then we need to print `Variable`s in the function call not `VariableAddress`
However if the function call is
`void f(int* a)`
Then we need to pass a `VariableAddress`

pyccel/codegen/printing/ccode.py
```diff
@@ -286,19 +286,27 @@ def _print_FunctionDef(self, expr):
...
         # currently support only function with one or zero output
-        args = ','.join(['{}'.format(self._print(a)) for a in expr.arguments])
+        args = ','.join(['{}'.format(self._print(a)) for a in args])
         if not func.results:
             return '{}({});'.format(func.name, args)
         return '{}({})'.format(func.name, args)
 
     def _print_Return(self, expr):
         code = ''
+        args = [VariableAddress(a) if self.stored_in_c_pointer(a) else a for a in expr.expr]
```
EmilyBourne, line 306: See above, not all return values are stored in pointers

pyccel/codegen/printing/ccode.py
```diff
@@ -443,7 +455,7 @@ def _handle_is_operator(self, Op, expr):
         a = expr.args[0]
         b = expr.args[1]
 
-        if (a.dtype is NativeBool() and b.dtype is NativeBool()):
+        if (Nil() not in expr.args and a.dtype is NativeBool() and b.dtype is NativeBool()):
```
EmilyBourne, line 458: Maybe it would be simpler to reverse the order of these `If` statements rather than testing for `Nil()` twice

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -302,6 +381,11 @@ def _print_FunctionDef(self, expr):
 
         # Call PyBuildNode
         wrapper_body.append(AliasAssign(wrapper_results[0],PyBuildValueNode(res_args)))
+
+        # Call free function for python type
+        wrapper_body += [FunctionCall(Py_DECREF, [i]) for i in self._to_free_PyObject_list]
```
EmilyBourne, line 386: What object is being freed here?

pyccel/parser/semantic.py
```diff
@@ -603,7 +603,7 @@ def _infere_type(self, expr, **settings):
             d_var['cls_base'      ] = expr.cls_base
             d_var['is_pointer'    ] = expr.is_pointer
             d_var['is_polymorphic'] = expr.is_polymorphic
-            d_var['is_optional'   ] = expr.is_optional
+            #d_var['is_optional'   ] = expr.is_optional
```
EmilyBourne, line 606: I think we can remove this line completely rather than just commenting it

tests/epyccel/test_base.py
```diff
@@ -73,12 +73,12 @@ def test_not(language):
     compare_epyccel(base.not_val, False, language=language)
 
 @pytest.mark.xfail(reason="f2py does not support optional arguments https://github.com/numpy/numpy/issues/4013")
```
EmilyBourne, line 75: See function `test_same_string` for an example of how to mark tests as failing in one language only

tests/epyccel/test_epyccel_optional_args.py
```diff
@@ -0,0 +1,73 @@
+import pytest
+
+from pyccel.epyccel import epyccel
+from pyccel.decorators import types
+from conftest       import *
+
+#------------------------------------------------------------------------------
+@pytest.mark.xfail(reason="f2py does not support optional arguments https://github.com/numpy/numpy/issues/4013")
```
EmilyBourne, line 8: See above

tests/epyccel/test_epyccel_optional_args.py
```diff
@@ -0,0 +1,73 @@
...
+    f = epyccel(f1, language = language)
+
+    # ...
+    assert f(2) == f1(2)
+    assert f() == f1()
+    assert f(None) == f1(None)
+    assert f(0) == f1(0)
+    # ...
+#------------------------------------------------------------------------------
+@pytest.mark.xfail(reason="f2py does not support optional arguments https://github.com/numpy/numpy/issues/4013")
```
EmilyBourne, line 25: See above

tests/epyccel/test_base.py
```diff
@@ -73,12 +73,12 @@ def test_not(language):
...
-def test_compare_is_nil():
-    compare_epyccel(base.is_nil, True, None)
+def test_compare_is_nil(language):
+    compare_epyccel(base.is_nil, True, None, language=language)
 
 @pytest.mark.xfail(reason="f2py does not support optional arguments https://github.com/numpy/numpy/issues/4013")
```
EmilyBourne, line 79: See above

tests/epyccel/test_epyccel_optional_args.py
```diff
@@ -0,0 +1,73 @@
...
+    f = epyccel(f2, language = language)
+
+    # ...
+    assert f(2.0) == f2(2.0)
+    assert f() == f2()
+    assert f(None) == f2(None)
+    assert f(0.0) == f2(0.0)
+    # ...
+#------------------------------------------------------------------------------
+@pytest.mark.xfail(reason="f2py does not support optional arguments https://github.com/numpy/numpy/issues/4013")
```
EmilyBourne, line 42: See above

tests/epyccel/test_epyccel_optional_args.py
```diff
@@ -0,0 +1,73 @@
...
+
+    f = epyccel(f3, language = language)
+
+    # ...
+    assert f(complex(1, 2.2)) == f3(complex(1, 2.2))
+    assert f() == f3()
+    assert f(None) == f3(None)
+    # ...
+#------------------------------------------------------------------------------
+@pytest.mark.xfail(reason="f2py does not support optional arguments https://github.com/numpy/numpy/issues/4013")
```
EmilyBourne, line 58: See above

EmilyBourne (changes requested):

pyccel/codegen/printing/ccode.py
EmilyBourne: Whether you pass `VariableAddress` or not depends on the signature not the type that you are passing
Suppose you have the following variables:
```
int a;
int* p;
```
We need variable when we should use `a` or `*p`, and `VariableAddress` where we should use `&a` or `p`. If the 2 are not interchangeable then you are either in an `AliasAssign` or you're probably doing something wrong

If the function being called is:
`void f(int c)`
Then the function call should be:
`f(a);`
or
`f(*p);`
Similarly if the function is:
`void g(int* d)`
Then the function call should be:
`g(&a);`
Or
`g(p);`
Currently you are doing:
`f(a)`, `g(a)`, `f(p)`, `g(p)`

EmilyBourne:
Please add a test which uses an optional variable to call another function and a test which calls a function with an optional argument from c

pyccel/codegen/printing/cwrappercode.py
MenoIy: When returning a complex number  we use function PyComplex_FromDouble https://docs.python.org/2/c-api/complex.html#c.PyComplex_FromDoubles to create the python type complex .
Then we return  a pytObject  with PyBuild_value  that create a new pyobject from given arg . so the variable that we created with PyComplex_FromDouble is not freed  
```
@types('complex')
def f(c):
    return c


PyObject *f_wrapper(PyObject *self, PyObject *args, PyObject *kwargs)
{
    double complex c;
    double complex Out_0001;
    PyObject *c_tmp;
    PyObject *Out_0001_tmp;
    PyObject *result;
    static char *kwlist[] = {
        "c",
        NULL
    };

    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "O", kwlist, &c_tmp))
    {
        return NULL;
    }
    c = pycomplex_to_complex(c_tmp);
    Out_0001 = f(c);
    Out_0001_tmp = complex_to_pycomplex(Out_0001);   // Pycomplex created with  PyComplex_FromDouble 
    result = Py_BuildValue("O", Out_0001_tmp);  // new Pyobject created from Out_0001_tmp 
    Py_DECREF(Out_0001_tmp);   // should free Out_0001_tmp 
    return result; 
}

```

EmilyBourne: Ok that makes sense :+1

EmilyBourne:
You will need to add the following line to the top of the test file that you have created:
```
# pylint: disable=missing-function-docstring, missing-module-docstring/
```

tests/epyccel/test_epyccel_optional_args.py
```diff
@@ -0,0 +1,135 @@
...
+)
+def test_f6(language):
+    import modules.Module_4 as mod
+
+    modnew = epyccel(mod, language = language)
+
+    # ...
+    assert mod.call_optional_1() == modnew.call_optional_1()
+    assert mod.call_optional_2(None) == modnew.call_optional_2(None)
+    assert mod.call_optional_2(0) == modnew.call_optional_2(0)
```
EmilyBourne, line 134: Given that the function returns either `0` or `a`, you should provide something other than 0 as the value of a. Otherwise you don't know if the right loop was called or not

EmilyBourne (changes requested):

EmilyBourne:
This PR does not fix the second point in the issue, but I am happy to leave that for another PR if you prefer. If that is the case then please create a new issue about "temporary objects as optional arguments". It will then be easier to identify the issue from the title

tests/epyccel/test_epyccel_functions.py
```diff
@@ -1,18 +1,11 @@
...
 import pytest
 import numpy as np
 
 from pyccel.epyccel import epyccel
 from pyccel.decorators import types
-from conftest       import *
-
-
-def clean_test():
-    shutil.rmtree('__pycache__', ignore_errors=True)
```
EmilyBourne, line 14: I suspect if we use `ignore_errors = True` in conftest this may fix the windows problems. Hopefully

MenoIy: '''    
 if f in  ("__pyccel__", "__epyccel__"):
            shutil.rmtree( file_name )                              here ?  <----------------

'''

EmilyBourne: yes

EmilyBourne:
I will approve this issue when:
1. The remaining codacy problem is fixed
2. Either an issue has been created for "temporary objects as optional arguments" or the issue has been fixed and tested here

pyccel/codegen/printing/ccode.py
```diff
@@ -542,8 +547,21 @@ def stored_in_c_pointer(self, a):
...
+        for a, f in zip(expr.arguments, func.arguments):
+            if isinstance(a, Variable) and self.stored_in_c_pointer(f):
+                args.append(VariableAddress(a))
+            elif f.is_optional and not isinstance(a, Nil):
+                tmp_var_name = self._parser.get_new_name('tmp')
+                tmp_var = Variable(name = tmp_var_name, dtype = f.dtype)
+                self._additional_declare.append(tmp_var)
+                assign = Assign(tmp_var, a)
+                args.append(VariableAddress(tmp_var))
+                self._additional_code += self._print(assign) + '\n'
```
EmilyBourne, line 561: Can we make a function for these lines, e.g:
```
def create_tmp_var(init_val, match_var)
```
that way we will avoid repetition in the future

pyccel/codegen/printing/ccode.py
```diff
@@ -542,8 +547,21 @@ def stored_in_c_pointer(self, a):
...
-        args = [VariableAddress(a) if isinstance(a, Variable) and self.stored_in_c_pointer(f) else a
-            for a, f in zip(expr.arguments, func.arguments)]
+
+        args = []
+        for a, f in zip(expr.arguments, func.arguments):
+            if isinstance(a, Variable) and self.stored_in_c_pointer(f):
+                args.append(VariableAddress(a))
+            elif f.is_optional and not isinstance(a, Nil):
+                tmp_var_name = self._parser.get_new_name('tmp')
+                tmp_var = Variable(name = tmp_var_name, dtype = f.dtype)
```
EmilyBourne, line 557: I recommend using `f.clone` if this is possible, that way we are sure that `Variable` includes all the necessary information (e.g. shape, rank, is_pointer, etc)

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -1,5 +1,7 @@
 # coding: utf-8
 # pylint: disable=R0201
+# pylint: disable=wildcard-import
```
EmilyBourne, line 3: We would like to avoid disabling errors in the `pyccel` folder. I know that `from pyccel.errors.messages import *` is used in several files, but the long term aim is to remove this syntax not to hide errors if this syntax is used for other imports.

If you want to disable a pytest error for 1 line only rather than the whole file you can add a comment to the end of that line

In this case though I think that there are few enough messages used that we can stop using this syntax. The only object that need to be imported from errors.messages is:
- PYCCEL_RESTRICTION_TODO

EmilyBourne (changes requested):

EmilyBourne:
Please add tests for the second point in the issue

pyccel/codegen/printing/ccode.py
MenoIy: I don't think we need to copy all informations

pyccel/codegen/printing/cwrappercode.py
MenoIy: done.

pyccel/codegen/printing/ccode.py
MenoIy: done.

pyccel/codegen/printing/ccode.py
```diff
@@ -544,6 +544,14 @@ def stored_in_c_pointer(self, a):
             return False
         return a.is_pointer or a.is_optional
 
+    def create_tmp_var(self, init_val, match_var):
+        tmp_var_name = self._parser.get_new_name('tmp')
+        tmp_var = Variable(name = tmp_var_name, dtype = match_var.dtype)
+        self._additional_declare.append(tmp_var)
+        assign = Assign(tmp_var, init_val)
+        self._additional_code += self._print(assign) + '\n'
+        return VariableAddress(tmp_var)
```
EmilyBourne, line 553: I think you should return a `Variable`, you don't know what this function will be used for in the future and it is possible that a `VariableAddress` is not required

tests/epyccel/modules/Module_5.py
```diff
@@ -0,0 +1,88 @@
+from pyccel.decorators import types
```
EmilyBourne, line 1: Missing pylint comment

pyccel/codegen/printing/ccode.py
EmilyBourne: I think this is a problem with the `clone` function, I'm not sure `is_optional` should be copied. However the additional information that is copied is not yet implemented in C, so maybe this is good enough for now.

EmilyBourne (approved):

saidctb (approved):

tests/epyccel/test_epyccel_optional_args.py
yguclu: Now this function returns `2` or `a`, hence this issue is resolved.

yguclu (approved):

PR #462, opened 2020-10-06
Generate a header file for C files

### LINKED ISSUE #430: Imports not working for C in the case of a FunctionCall
Imports not working for C when calling a function from the main program

**To Reproduce**
```

@types('int,int')
def rec(b,y):
    return b + y

z = rec(2,4)

```

the generated Program in C 
```
#include "test"
#include "stdlib.h"
int main()
{
    long z;
    z = rec(2,4);
    return 0;
} 
```
Which fail in the compilation stage

```
'test' file not found
#include "test"
```
the import of the function should have the .c extension 
`#include "test.c"`

**Language**
C

### LINKED ISSUE #431: C modules should generate a .h file
**Describe the bug**
The translation of a module to C leads to the generation of a .c file containing the function definitions. It is bad practice to include this file in other files. Instead we should include a .h file containing the function signatures. Currently this .h file is not generated

**Language**
C

### LINKED ISSUE #442: Warnings are being raised multiple times for C language
while i'm working on issue #421 i found that warnings in C language are being raised 2 times.
for example this python code: 

```
import numpy
import numpy

@types('int,int')
def func(a,b):
    return a + b
```
generate multiple warnings : 

```
pyccel: [codegen] 
 |warning| Duplicated import  (numpy)

pyccel: [codegen] 
 |warning| Duplicated import  (numpy) 
```
this is because that `errors.check()` is called in both `python_wrapper.py` and  `pipeline.py`

the solution is tot call `errors.check()` in `python_wrapper.py` and let `pipeline.py` manage it

EmilyBourne:
Create `ModuleHeader` class to handle printing of .h file. Pass  includes, libs and libdirs to `setuptools.Extension` explicitly

- Create `ModuleHeader` object
- Add mechanism to Codegen for printing header file
- Write basic `_print_ModuleHeader` function (fixes #431)
- Add .h in `_print_Import` instead of in the `_additional_imports` set (fixes #430)
- Compile c modules to make the .o file available for linking
- Do not error with unused command line arguments
- Move filename setter to before 'read_file' to collect invalid python syntax correctly
- Set filename for codegen step
- Stop failure when filename is None
- Always call errors.reset after errors.check (fixes #442 )
- Pass includes, libs and libdirs to `setuptools.Extension` explicitly
- Handle `DottedName` imports in c for projects
- Add `compile_c` function to `test_pyccel.py`
- Activate as many pyccel tests as possible (all without collisions, arrays, or multiple results)


yguclu (approved):

pyccel/codegen/cwrapper.py
```diff
@@ -1,10 +1,47 @@
...
+        compiler,
+        include = '',
+        libs = '',
+        libdirs = '',
+        flags = ''):
+    """
+    Create the code for the setup file which uses python setuptools
+    to compile a c file and generate the corresponding shared
+    library file
+    """
```
yguclu, line 20: Would you mind adding `Parameters` and `Returns` to the docstring?

tests/pyccel/test_pyccel.py
```diff
@@ -66,6 +66,29 @@ def compile_pyccel(path_dir,test_file, options = ""):
...
+    command = [shutil.which("gcc"), "-O3", "%s.c" % root]
+    if isinstance(dependencies, list):
+        for d in dependencies:
+            d = insert_pyccel_folder(d)
+            command.append(d[:-3]+".o")
+            command.append("-I"+os.path.dirname(d))
+    elif isinstance(dependencies, str):
+        dependencies = insert_pyccel_folder(dependencies)
+        command.append(dependencies[:-3]+".o")
+        command.append("-I"+os.path.dirname(dependencies))
```
yguclu, line 84: There is some repeated code here. We could write:
```python
deps = dependencies if hasattr(dependencies, '__iter__') else [dependencies]

for d in deps:
    d = insert_pyccel_folder(d)
    command.append(d[:-3]+".o")
    command.append("-I"+os.path.dirname(d))
```

yguclu (changes requested):

tests/pyccel/test_pyccel.py
EmilyBourne: A string has a `__iter__` method. I have used:
```
deps = [dependencies] if isinstance(dependencies, str) else dependencies
```
That way the code works for tuples too

yguclu (approved):

pyccel/codegen/printing/ccode.py
```diff
@@ -250,18 +254,38 @@ def _print_Complex(self, expr):
...
         return self._print(PyccelAssociativeParenthesis(PyccelAdd(expr.real_part,
                         PyccelMul(expr.imag_part, ImaginaryUnit()))))
 
     def _print_ImaginaryUnit(self, expr):
         return '_Complex_I'
 
+    def _print_ModuleHeader(self, expr):
+        name = expr.module.name
+        # TODO: Add classes and interfaces
+        funcs = ';\n\n'.join(self.function_signature(f) for f in expr.module.funcs)
```
rakati, line 267: I think here if the module contains only one function, the `';\n\n'` will not be added.

EmilyBourne: Yes, I added the semi colon for the final function in the assembly in the return. But I think you're right that it's not very clear that way. I will change it

pyccel/codegen/printing/ccode.py
```diff
@@ -250,18 +254,38 @@ def _print_Complex(self, expr):
...
     def _print_ImaginaryUnit(self, expr):
         return '_Complex_I'
 
+    def _print_ModuleHeader(self, expr):
+        name = expr.module.name
+        # TODO: Add classes and interfaces
+        funcs = '\n\n'.join('{};'.format(self.function_signature(f)) for f in expr.module.funcs)
+
+        # Print imports last to be sure that all additional_imports have been collected
+        imports = [*expr.module.imports, *map(Import, self._additional_imports)]
```
rakati, line 270: Maybe this can be a separate issue.
this code in python
```python
@types('int,int')
def rec(b,y):
    return b + y
z = rec(2,4)
```
generated in c `prog_pro.c`
```c
#include <pro.h>
#include <stdlib.h>
int main()
{
    long z;

    z = rec(2,4);
    return 0;
}
```
and `pro.h`
```c
#include <stdlib.h>

long rec(long b, long y)
{
    long Out_0001;
    Out_0001 = b + y;
    return Out_0001;
}
```
here we don't need `<stdlib.h>` but it is imported in the two files, so we need to remove it until time needed.
and for the local header file imported we need to use the C convention for including local  `#include "pro.h"` instead of `#include <pro.h>`  (it doesn't lead to error in compiling but it is better to use C standard coding style).

EmilyBourne: >here we don't need `<stdlib.h>`

Which file doesn't need `stdlib.h`? `prog_pro.c`, `pro.h`, or `pro.c`? The second file that you have copied above is `pro.c` not `pro.h`. 
As a first attempt I have left all the same imports in `pro.h` as in `pro.c` as there is no simple way to know which ones are needed for the header and which are needed for the body of the function.

> we need to use the C convention for including local headers `#include "pro.h"` instead of `#include <pro.h>`

I agree but pyccel doesn't save any information in the `Import` class apart from the name. So currently at the codegen stage we don't know if the header is local or global. I think we should open a new issue for this as it may require modifications in several places, especially if we can import user-provided fortran files to the python (I think this was possible at one point, but I have no idea if it still works, @saidctb ?). If the user provides a file then we don't know if it is local or global.
Unless you have an idea for a quick fix?

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -433,7 +433,7 @@ def _print_Module(self, expr):
 
         # Print imports last to be sure that all additional_imports have been collected
         imports  = [Import(s) for s in self._additional_imports]
-        imports += [Import('Python.h')]
+        imports += [Import('Python')]
         imports  = '\n'.join(self._print(i) for i in imports)
 
         return ('#define PY_SSIZE_T_CLEAN\n'
```
rakati, line 439: It will be good if we include the header file instead of rewriting the functions signatures.

pyccel/codegen/printing/ccode.py
rakati: > Which file doesn't need `stdlib.h`? `prog_pro.c`, `pro.h`, or `pro.c`? The second file that you have copied above is `pro.c` not `pro.h`.
> As a first attempt I have left all the same imports in `pro.h` as in `pro.c` as there is no simple way to know which ones are needed for the header and which are needed for the body of the function.

the header file `.h` will include all necessary libraries and macros, for the file with the functions declaration need only to include the header (which is the header itself contains necessary includes).

EmilyBourne: Ah! I see what you mean. I will do that

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: I do not understand your comment. We do include the header file

pyccel/codegen/printing/ccode.py
EmilyBourne: done

pyccel/codegen/printing/cwrappercode.py
rakati: I mean in the generated wrapper file. instead of writing module functions signature, we only need to include the header generated.
I will give a detailed example.

pyccel/codegen/printing/ccode.py
rakati: great Emily. but we still need have to include the `_additional_imports`  in the `moduleHeader`.
to see that try to use a math function.

pyccel/codegen/printing/cwrappercode.py
rakati: python module `pro.py`
```python
@types('int,int')
def rec(b,y):
    from math import abs
    return abs(b) + y
```
generated C files.
`pro.c`
```c
#include <pro.h>
long rec(long b, long y)
{

    long Out_0001;

    Out_0001 = fabs((double)(b)) + y;
    return Out_0001;
}
```
`pro.h`
```c
#ifndef pro_H // pro_H need to be PRO_H
#define pro_H // same here

#include <stdlib.h>
// we need also to include <math.h>

long rec(long b, long y);

#endif // pro_H
```
`pro_wrapper.c`
```c
#define PY_SSIZE_T_CLEAN
#include <stdlib.h> // we don't need this
#include <Python.h>

// here instead of writing signature we only need to include pro.h
long rec(long b, long y);
// we need to use separator only when the item printed



PyObject *rec_wrapper(PyObject *self, PyObject *args, PyObject *kwargs)
{

    long b;
    long y;
// more code
```

EmilyBourne: Ah sorry, I had not realised we were talking about the wrapper file. I decided to handle the wrapper file differently as at some point in the near future it will also be used for fortran. When that happens it makes more sense to handle the difference in the function signatures in `cwrappercode.py` than to generate a `.h` file for a fortran function using `ccode.py` which is supposed to be pure c

pyccel/codegen/printing/ccode.py
EmilyBourne: Oops you are right, I fixed it

EmilyBourne: I have created an issue for the `""` vs `<>` problem. Are you happy to resolve this conversation now?

rakati: hhh, we went back to the same point.
I see that we need to get the `_additional_imports`  from `.c` file printer to `.h` printer, which currently I am no sure if we can do that.

rakati: of course, I am happy with your work.
you did well. these are just structuring stuff.

EmilyBourne: If the imports aren't needed in the header then I think it's better if they're only included in the `.c` file. That way a minimal number of additional things are imported into other files. That will help to avoid name collisions. I don't know if there's any way of avoiding a double import of something like `complex.h` right now though. I will open an issue

pyccel/codegen/printing/ccode.py
```diff
@@ -250,18 +254,38 @@ def _print_Complex(self, expr):
...
+        imports = '\n'.join(self._print(i) for i in imports)
+
+        return ('#ifndef {name}_H\n'
+                '#define {name}_H\n\n'
+                '{imports}\n\n'
+                #'{classes}\n\n'
+                '{funcs}\n\n'
+                #'{interfaces}\n\n'
+                '#endif // {name}_H').format(
+                        name    = name,
```
rakati, line 280: use `name,.upper()`.
I will approve this PR when you resolve this.

pyccel/codegen/printing/ccode.py
```diff
@@ -250,18 +254,38 @@ def _print_Complex(self, expr):
...
+        name = expr.module.name
+        # TODO: Add classes and interfaces
+        funcs = '\n\n'.join('{};'.format(self.function_signature(f)) for f in expr.module.funcs)
+
+        # Print imports last to be sure that all additional_imports have been collected
+        imports = [*expr.module.imports, *map(Import, self._additional_imports)]
+        imports = '\n'.join(self._print(i) for i in imports)
+
+        return ('#ifndef {name}_H\n'
+                '#define {name}_H\n\n'
```
rakati, line 274: see below

pyccel/codegen/printing/ccode.py
EmilyBourne: I wondered about this. I didn't use `name.upper()` as I was worried about name clashes if we have a `test.py` and a `Test.py` file. But it's probably not worth ignoring the convention just in case the user has bad names. This won't work in fortran anyway as fortran is not case sensitive.
I have changed to `name.upper()`

pyccel/codegen/printing/ccode.py
rakati: I see what you mean. for C style all includes, structures, and macros are in the `.h` file.
the `.c` files will include just the header file.

> I don't know if there's any way of avoiding a double import of something like complex.h right now though. I will open an issue

If you revise the header file you will see that it is wrapped with `ifndef` ... `endif`.
so there will be no problem if we import that header many times.
and the code will be well organized.
in the future, we may add only one function in a file and create one header for all functions within these files.
what do you think.

rakati (approved):

pyccel/codegen/printing/ccode.py
EmilyBourne: This is really a question of convention. In my coding classes I was taught to limit the number of includes in the header so something isn't included unless it needs to be / isn't included without you being aware.
E.g.
I write a file my printer.h which contains an implementation of a function `printf` that I want to use in the rest of my code for some reason (e.g to always write to stderr, I don't have a good example in mind, but I'm sure there is one).
I then import `printer.h` in my file where I also import `BigClass.h`, `BigClass.h` has a debug method which uses `printf` from stdio. If `stdio.h` was only imported in `BigClass.c` where it is needed then this will not cause a problem, but if it is imported in `BigClass.h` to limit the number of imports in `BigClass.c` then my `main.c` will contain 2 implementations of `printf` which will be a problem

> in the future, we may add only one function in a file and create one header for all functions within these files.
what do you think.

1 function per `.c` file and 1 `.h` file for all `.c` files? That is not standard practice in c I can't see any advantages to doing it this way. Did you have any in mind? It makes name clashes more likely and you have to be more careful with the order in which you construct your `.h` file so any structures containing structures are declared in the right order

PR #479, opened 2020-10-13
Support passing function as argument in C

### LINKED ISSUE #444: Allow function to be passed as argument in C
**Describe the feature**
Currently functions can't be passed as argument in C.

**Test Code**

E.g
```
    #$ header function f()
    def f():
        return 0

    #$ header high_order((int)())
    def high_order(function):
        x = function()
        return x

    x = high_order(f)
```

E.g
```
    #$ header function f(int)
    def f(x):
        return x + 1

    #$ header function high_order((int)(int), int)
    def high_order(function, x):
        x =  function(x)
        return x

    x = high_order(f, 0)
```

**Language**
C

rmahjoubi:
**Allow functions to be passed as arguments in C**

- Provide grammar to parse a function that takes a function as an argument.
- Add class FunctionAddress to represent a function that is being passed as an argument.
- Add class FuncAddressDeclare used in code generation in C to declare a pointer to function. 
- Add class ValuedAddressDeclare used to represent a valued function passed as an argument.
- Add tests for a function that is being passed as an argument.
- Temporarily parse  arguments as PythonObject in Cwrapper.

Fixes #444


pyccel/ast/core.py
```diff
@@ -2673,6 +2673,18 @@ def is_ndarray(self):
...
+                        allocatable=self._allocatable,
+                        is_stack_array=self._is_stack_array,
+                        is_pointer=self._is_pointer,
+                        is_target=self._is_target,
+                        is_polymorphic=self._is_polymorphic,
+                        is_optional=self._is_optional,
+                        shape=self._shape, cls_base=self._cls_base,
+                        order=self.order, precision=self.precision,
+                        is_argument=self._is_argument,
+                        is_kwonly=self._is_kwonly)
```
EmilyBourne, line 2687: Why is this function necessary? The `clone` function does the same thing.
If it is necessary then the function needs a new name. We are not renaming a variable but creating another Variable with the same properties (cloning the Variable). It is important that this distinction is clear as renaming something would not invalidate other existing references to the object, but cloning will

pyccel/ast/core.py
```diff
@@ -4298,6 +4375,62 @@ def execute(self):
...
+        if not isinstance(static, bool):
+            raise TypeError('Expecting a boolean for static attribute')
+
+        return Basic.__new__(
+            cls,
+            variable,
+            intent,
+            value,
+            static,
+            )
```
EmilyBourne, line 4404: For new classes we would like to avoid using `__new__` as the `_args` tuple is hard to maintain. Please use `__init__` (see `Variable` class for how to handle `__new__` in this case)

pyccel/ast/cwrapper.py
```diff
@@ -122,16 +122,22 @@ def __init__(self, python_func_args, python_func_kwargs, c_func_args, parse_args
...
+                self._flags += 'O'
+            else:
+                self._flags += pytype_parse_registry[(c_func_args[i].dtype, c_func_args[i].precision)]
             i+=1
         if i < len(c_func_args):
             self._flags += '|'
         while i < len(c_func_args):
-            self._flags += pytype_parse_registry[(c_func_args[i].dtype, c_func_args[i].precision)]
+            if isinstance(c_func_args[i], FunctionAddress):
+                self._flags += 'O'
```
EmilyBourne, line 134: Why can't we add a `FunctionAddress` to the `pytype_parse_registry`?

pyccel/ast/headers.py
```diff
@@ -189,29 +189,83 @@ def create_definition(self):
...
+                        if isinstance(dtype, str):
+                            try:
+                                dtype = datatype(dtype)
+                            except ValueError:
+                                dtype = DatatypeFactory(str(dtype), ("_name"))()
+                                is_pointer = True
+                        arg = Variable(dtype, '',
+                                    allocatable=allocatable, is_pointer=is_pointer,
+                                    rank=rank, shape=shape ,order = order, precision = precision,
+                                    is_argument=True)
```
EmilyBourne, line 216: It looks like a lot of the code here is similar to what is in the else. Could it be extracted to a new function to avoid repetition?

pyccel/ast/headers.py
```diff
@@ -189,29 +189,83 @@ def create_definition(self):
...
+                        if isinstance(dtype, str):
+                            try:
+                                dtype = datatype(dtype)
+                            except ValueError:
+                                dtype = DatatypeFactory(str(dtype), ("_name"))()
+                                is_pointer = True
+                        ret = Variable(dtype, '',
+                                    allocatable=allocatable, is_pointer=is_pointer,
+                                    rank=rank, shape=shape ,order = order, precision = precision,
+                                    is_argument=True)
```
EmilyBourne, line 240: See above

pyccel/codegen/printing/ccode.py
```diff
@@ -385,6 +386,22 @@ def get_declare_type(self, expr):
...
+            msg = 'Multiple output arguments is not yet supported in c'
+            errors.report(msg+'\n'+PYCCEL_RESTRICTION_TODO, symbol=expr,
+                severity='fatal')
+        else:
+            ret_type = self._print(datatype('void')) + ' '
+        name = expr.name
+        if not expr.arguments:
+            arg_code = 'void'
+        else:
+            arg_code = ', '.join('{}'.format(self._print_FuncAddressDeclare(i)) if isinstance(i, FunctionAddress) else '{0}{1}'.format(self.get_declare_type(i), i) for i in expr.arguments)
```
EmilyBourne, line 402: This line doesn't fit on my screen on github. It needs to be split into a multi-line object

pyccel/codegen/printing/ccode.py
```diff
@@ -424,8 +441,11 @@ def function_signature(self, expr):
         if not expr.arguments:
             arg_code = 'void'
         else:
-            arg_code = ', '.join('{0}{1}'.format(self.get_declare_type(i), i) for i in expr.arguments)
-        return '{0}{1}({2})'.format(ret_type, name, arg_code)
+            arg_code = ', '.join('{}'.format(self.function_signature(i)) if isinstance(i, FunctionAddress) else '{0}{1}'.format(self.get_declare_type(i), i) for i in expr.arguments)
```
EmilyBourne, line 444: This also doesn't fit on screen

pyccel/codegen/printing/ccode.py
```diff
@@ -516,10 +536,12 @@ def _print_MathSqrt(self, expr):
...
 
-    def _print_FunctionDef(self, expr):
+    def _print_FunctionAddress(self, expr):
+        return expr.name
 
-        decs  = [Declare(i.dtype, i) for i in expr.local_vars]
-        decs += [Declare(i.dtype, i) for i in expr.results]
+    def _print_FunctionDef(self, expr):
+        decs  = [Declare(i.dtype, i) if isinstance(i, Variable) else FuncAddressDeclare(i) for i in expr.local_vars]
+        decs += [Declare(i.dtype, i) if isinstance(i, Variable) else FuncAddressDeclare(i.name, i.arguments, i.results) for i in expr.results]
```
EmilyBourne, line 544: Why is the constructor call different between these 2 lines?

pyccel/parser/semantic.py
```diff
@@ -2375,34 +2374,51 @@ def _visit_FunctionDef(self, expr, **settings):
...
                     if additional_args:
                         args += additional_args
 
                     args.append(a_new)
-                    self.insert_variable(a_new, name=str(a_new.name))
-
+                    if isinstance(a_new, FunctionAddress):
+                        self.insert_function(a_new)
+                    else:
+                        self.insert_variable(a_new, name=str(a_new.name))
```
EmilyBourne, line 2421: Is this required a lot? If so would it be worth modifying `insert_variable` instead so it starts with:
```
if isinstance(var, FunctionAddress):
    return self.insert_function(var)
```
?

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -104,6 +104,9 @@ def get_PyArgParseType(self, used_names, variable):
             These handle casting collect_var to variable if necessary
         """
 
+        if isinstance(variable, FunctionAddress):
+            return variable, []
```
EmilyBourne, line 108: Does this not give an incompatible type error when you try to compile?

EmilyBourne (changes requested):

pyccel/ast/core.py
rmahjoubi: this function is necessary for fortran printer. as you mentioned `rename` function creates another variable with the same properties in order to do that with the function `clone` the variable properties needs to be copied before passing them to the function. as for the name of the function it should be changed since cloning is other than renaming. I will remove it until it being used in the fortran printer.

EmilyBourne: > the variable properties needs to be copied before passing them

Why do you say that? The clone method should copy all properties unless you provide a different value in a kwarg

rmahjoubi: yes you are right that where i was mistaking, clone function does the same thing

tests/epyccel/modules/highorder_functions.py
```diff
@@ -0,0 +1,102 @@
+#$ header function f1(int)
```
EmilyBourne, line 1: Missing pylint line at start of test file

pyccel/parser/semantic.py
rmahjoubi: it's only used once, should I modify `insert_variable` however ?

EmilyBourne: If you think it might be needed every time we use `insert_variable` then I would modify it. If this will be the only place it will ever be needed then leave it

yguclu:
@rmahjoubi You need to fix the conflicts with the master branch in order to update this branch

pyccel/codegen/printing/cwrappercode.py
rmahjoubi: it doesn't, why do you suspect that ?

EmilyBourne: The second argument of this return handles casting the variable collected from the argument parsing and translating it to a type that makes sense in pure C. So in this case it should translate a `PyObject*` to a function pointer, e.g. `int(*f)(int)`

I think you decided to leave this translation for another PR but it would be nice if we could raise an error when trying to use the epyccelized function as if you pass a `PyObject*` as a function pointer I suspect that you will have either a compiler problem, or a segmentation fault. You will know more about this than me I think.
In order to raise a python error instead you need to use `PyErr_SetString` and return Nil(). You can search for `PyErr_SetString` in `codegen/printing/cwrappercode.py` for an example of how to raise errors

rmahjoubi: `        if isinstance(variable, FunctionAddress):
            body = [PyErr_SetString('PyExc_NotImplementedError', '"can not pass a function as argument"'), Return([Nil()])]
            return variable, body
`
do you mean something like this?

EmilyBourne: Yes exactly that looks perfect :+1:

pyccel/ast/core.py
```diff
@@ -3743,6 +3745,72 @@ def __str__(self):
...
+        name,
+        arguments,
+        results,
+        body,
+        is_optional=False,
+        is_pointer=False,
+        is_kwonly=False,
+        is_argument=False,
+        *args,
+        **kwargs
```
EmilyBourne, line 3760: It seems odd to add `*args` here this means that you have to specify `is_optional`, `is_pointer`, `is_kwonly`, and `is_argument` in order to provide additional args. I think it would make more sense to have these 4 as keyword only arguments so:
```
def __init__(
        self,
        name,
        arguments,
        results,
        body,
        *args,
        is_optional=False,
        is_pointer=False,
        is_kwonly=False,
        is_argument=False,
        **kwargs)
```

pyccel/ast/core.py
```diff
@@ -3743,6 +3745,72 @@ def __str__(self):
                 args   = ', '.join(self.args),
                 result = result)
 
+class FunctionAddress(FunctionDef):
```
EmilyBourne, line 3748: Missing class docstring

pyccel/ast/core.py
```diff
@@ -3743,6 +3745,72 @@ def __str__(self):
...
+
+    @property
+    def is_kwonly(self):
+        return self._is_kwonly
+
+    @property
+    def is_optional(self):
+        return self._is_optional
+
+class ValuedFunctionAddress(FunctionAddress):
```
EmilyBourne, line 3801: Missing class docstring

pyccel/ast/core.py
```diff
@@ -4327,6 +4395,62 @@ def execute(self):
 
 # TODO: Should Declare have an optional init value for each var?
 
+class FuncAddressDeclare(Basic):
```
EmilyBourne, line 4398: Missing class docstring

EmilyBourne: @yguclu any idea why these are not raising codacy errors?

tests/epyccel/modules/highorder_functions.py
```diff
@@ -0,0 +1,104 @@
...
+def test_real_3():
+    x = high_real_3(f8)
+    return x
+
+#$ header function test_valuedarg_1()
+def test_valuedarg_1():
+    x = high_valuedarg_1(2)
+    return x
+
+#$ header function test_real_real_int()
```
EmilyBourne, line 101: Does this also work with the `@types` decorator?

EmilyBourne (changes requested):

tests/epyccel/modules/highorder_functions.py
rmahjoubi: do you mean this test specifically ? it does in general, i will add a couple of tests with `@type` decorator

EmilyBourne: No I meant in general I hadn't seen a test so I wasn't sure. In that case yes, please add tests :)

EmilyBourne (approved):

pyccel/ast/core.py
yguclu: @EmilyBourne Not really

pyccel/ast/headers.py
```diff
@@ -179,6 +179,31 @@ def create_definition(self):
         funcs = []
         dtypes = []
 
+        def build_argument(var_name, dc):
```
yguclu, line 182: Can you please add a comment line that explains what this function does? It will be very useful for future developers looking into this code.

pyccel/codegen/printing/ccode.py
```diff
@@ -474,6 +476,24 @@ def get_declare_type(self, expr):
...
+                severity='fatal')
+        else:
+            ret_type = self._print(datatype('void')) + ' '
+        name = expr.name
+        if not expr.arguments:
+            arg_code = 'void'
+        else:
+            arg_code = ', '.join('{}'.format(self._print_FuncAddressDeclare(i))
+                        if isinstance(i, FunctionAddress) else '{0}{1}'.format(self.get_declare_type(i), i)
+                        for i in expr.arguments)
```
yguclu, line 494: Are you handling here a function as argument, which itself has a function argument?

pyccel/parser/semantic.py
```diff
@@ -2426,7 +2443,8 @@ def _visit_FunctionDef(self, expr, **settings):
             body = self._visit(expr.body)
 
             # ISSUE 177: must update arguments to get is_target
```
yguclu, line 2445: @EmilyBourne @saidctb This comment refers to an issue that was closed a long time ago, does it still carry a meaning?

yguclu (changes requested):

pyccel/ast/cwrapper.py
```diff
@@ -114,7 +114,7 @@ def __init__(self, python_func_args, python_func_kwargs, c_func_args, parse_args
             raise TypeError('Python func args should be a Variable')
         if not isinstance(python_func_kwargs, Variable):
             raise TypeError('Python func kwargs should be a Variable')
-        if not isinstance(c_func_args, list) and any(not isinstance(c, Variable) for c in c_func_args):
+        if not isinstance(c_func_args, list) and any(not isinstance(c, (Variable, FunctionAddress)) for c in c_func_args):
```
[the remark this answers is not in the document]
rmahjoubi: it's should be an `or`. however we pass a `c_func_args` sympy tuple and not a list. I think I should perform a test to check if `c_func_args` is iterable insead

pyccel/codegen/printing/ccode.py
rmahjoubi: yes, except that it's not fully supported, I still don't properly extract information about argument types and return types of the most inner function argument

yguclu: Good. Can you please add a comment line (starting with `# TODO:`) with this explanation?

pyccel/ast/cwrapper.py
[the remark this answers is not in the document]
yguclu: It makes sense to use tuples in all situations where we don't plan to modify the list. Let me know what you find out please!

pyccel/parser/semantic.py
EmilyBourne: I don't think so

pyccel/ast/cwrapper.py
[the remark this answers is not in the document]
rmahjoubi: I checked, we don't need to change the list, and I don't think we will in this context, we only use it to get informations about python arguments that will be parsed

pyccel/ast/cwrapper.py
```diff
@@ -114,7 +116,7 @@ def __init__(self, python_func_args, python_func_kwargs, c_func_args, parse_args
             raise TypeError('Python func args should be a Variable')
         if not isinstance(python_func_kwargs, Variable):
             raise TypeError('Python func kwargs should be a Variable')
-        if not isinstance(c_func_args, list) and any(not isinstance(c, Variable) for c in c_func_args):
+        if not isinstance(c_func_args, (tuple, Tuple)) or any(not isinstance(c, (Variable, FunctionAddress)) for c in c_func_args):
```
EmilyBourne, line 119: `tuple` should be sufficient. `Tuple` is a sympy object which we should probably remove at some point. It's therefore better to not add additional unnecessary references to it

pyccel/ast/headers.py
```diff
@@ -179,6 +179,32 @@ def create_definition(self):
...
+            shape = None
+            if rank >1:
+                order = dc['order']
+
+            if isinstance(dtype, str):
+                try:
+                    dtype = datatype(dtype)
+                except ValueError:
+                    dtype = DataTypeFactory(str(dtype), ("_name"))()
+                    is_pointer = True
```
EmilyBourne, line 200: What case is this treating? Why is it always a pointer?

pyccel/parser/syntax/headers.py
```diff
@@ -96,6 +123,7 @@ def expr(self):
         d_var['allocatable'] = len(trailer)>0
         d_var['is_pointer'] = False
         d_var['precision']  = precision
+        d_var['isfunc'] = False
```
EmilyBourne, line 126: All the other keys use underscores to separate words (ie `is_func`). I think it would be good to keep the same convention here

EmilyBourne (changes requested):

pyccel/ast/cwrapper.py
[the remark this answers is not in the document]
yguclu: Then it seems that we don't need to check if `c_func_args` is a list, and the correct check here should be
```python
if any (not isinstance(c, (Variable, FunctionAddress)) for c in c_func_args)
```
or (a little easier to read)
```python
if not all(isinstance(c, (Variable, FunctionAddress)) for c in c_func_args)
```

@EmilyBourne @rmahjoubi What do you think?

pyccel/ast/cwrapper.py
yguclu: I made a comment to the outdated code, because it seemed natural to continue the same discussion. sorry for the confusion

yguclu: My point there was that we probably don't even need to check the type of `c_func_args`...

pyccel/ast/cwrapper.py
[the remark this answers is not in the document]
EmilyBourne: I agree, if you iterate over something that is not iterable then the error should be enough to tell what the problem is. If we'd rather keep it as it is then we should update the error message to specify a tuple of Variables rather than a list of Variables

pyccel/ast/headers.py
rmahjoubi: it's been there before, i only moved it another function, I don't know why we set `is_pointer` to true. this is the case of a user defined type i guess?

EmilyBourne: Maybe, but I'm not sure why that would make it a pointer (in the Fortran sense). Do you know @saidctb

saidctb: this is the case of a user defined class , it's not used anymore for sure

rmahjoubi: should i remove it and raise an error instead ?

EmilyBourne: I would leave it as we plan to add classes at some point so there's no point re writing what we already have. I'm not sure why a class as an argument would have to be a pointer though

yguclu (approved):

pyccel/ast/headers.py
rmahjoubi: in the C case, i think it makes sense, in this example:
```
class p:
    def __init__(self):
        self.x = 0

def f(p):
     p.x = 1

cls = p()
f(cls)
```

in c we can have somthing like this:
```
 typedef struct{
     int x;
     void(*init)(int*);
 } p;

int f(p* cls)
{
   cls->x = 6;
}
```
if instead we used `int f(p cls)` this wont affect the attributes of cls

EmilyBourne: In C lots of things need to be passed as pointers. In your example `cls` should be passed as a pointer anyway in order to avoid costly copies. This is why we created the `stored_in_c_pointer` function.
However things with the `is_pointer` property will me pointers in all languages.
In Fortran (and therefore pyccel as pyccel was originally designed with only Fortran in mind) a pointer is something which points at something else and therefore requires the `pointer` attribute in Fortran. There is no reason for this to be the case for all classes.
You could also imagine the case of C++ where we would use references rather than pointers for your example

EmilyBourne: If we can't think of a reason why this needs to be a pointer I would like to remove this line to avoid potential errors once classes are added. It's not really relevant to this PR, but if we don't do it now then we will forget. Do you agree @saidctb ?

saidctb: yes fine by me

saidctb (approved):

pyccel/ast/headers.py
EmilyBourne: Then I will approve once this line is removed

yguclu: @rmahjoubi Your PR is ready to be merged as soon as you apply the requested change!

rmahjoubi: @EmilyBourne It's good now

yguclu (approved):

EmilyBourne (approved):

PR #484, opened 2020-10-19
Fix compilation error when using some Numpy functions with integer args

### LINKED ISSUE #373: Several Fortran functions need casting to real
If an integer is provided as an argument to certain functions then the compilation fails as fortran only accepts `real` or `complex` values. The argument should be cast to a real in the printing step.

The following is an example:
```
from pyccel.decorators import external

@external
def test_tanh():
    from numpy import tanh
    return tanh(2)

test2=epyccel(test_tanh)
test2()
```

The error generated by the compiler is:
```
/home/emily/Code/test/tmp/__epyccel__/__pyccel__/mod_ywlidiyr.f90:17:20:

   Dummy_2958 = tanh(2_8)
                    1
Error: ‘x’ argument of ‘tanh’ intrinsic at (1) must be REAL or COMPLEX
```

This was previously discussed in issue #201

rakati:
- Add casting to real for math functions in Fortran;
- Improve tests of Numpy functions by:
   - raise the range of value tests,
   - testing negative args and mix of positive and negative args,
   - testing integer numbers and mix of real and integer numbers,
   - testing return type for different types of args;
- Add a new class for the absolute function in `numpyext.py` to fix the wrong return type.
closes #373


pyccel/ast/numpyext.py
```diff
@@ -1185,11 +1186,11 @@ class NumpyArctanh(NumpyUfuncUnary) : pass
 
 #=======================================================================================
 
-class NumpyAbs(NumpyUfuncUnary):
+class NumpyAbsolute(NumpyUfuncUnary):
```
EmilyBourne, line 1189: What is the difference between `NumpyAbs` and `NumpyAbsolute`? Why do we need both classes?
According to https://numpy.org/doc/stable/reference/generated/numpy.absolute.html:
> numpy.absolute
> np.abs is a shorthand for this function.

pyccel/codegen/printing/fcode.py
```diff
@@ -2242,6 +2250,11 @@ def _print_MathPow(self, expr):
         base_c = self._print(base)
         e_c    = self._print(e)
         return '{} ** {}'.format(base_c, e_c)
+    def _print_NumpyFabs(self, expr):
+        arg = expr.args[0]
+        code_args = self._print(arg)
+        code = 'fabs({})'.format(code_args)
+        return self._get_statement(code)
```
EmilyBourne, line 2257: Why can't this function be handled by `_print_NumpyUfuncBase`? The main difference that I see is that the argument is not converted to a real. Is this unnecessary for fabs and necessary for almost every other NumpyUfuncBase or have I missed something?

EmilyBourne (changes requested):
Sorry I put my review comments in the wrong place. Please see https://github.com/pyccel/pyccel/commit/12a19b92d92d12678534dfa6e74e83f5ec4f86b7#comments

pyccel/ast/numpyext.py
yguclu: @rakati I have the same question here...

yguclu (changes requested):
You've added an impressive amount of tests, good job!

But maybe you need to test complex arguments too?

And there are a couple of open questions regarding `numpy.fabs` and `numpy.absolute`...

pyccel/codegen/printing/fcode.py
yguclu: Same question here...

pyccel/ast/numpyext.py
rakati: I was getting a different return type from abs and absolute functions.
so as a solution I separate them. but after your comment, I did a small search and I found that abs and absolute is quite the same. I think I made a mistake somewhere, I will check that again.

pyccel/codegen/printing/fcode.py
rakati: yes, you are right, we don't need it. I add it just for testing purposes. I forgot to delete it.
maybe it is the problem of different return types I got. I will make sure of it.

pyccel/ast/numpyext.py
EmilyBourne: Maybe you were using `abs` and `numpy.absolute` rather than `numpy.abs` and `numpy.absolute`?

rakati: I just realize the problem
```python
import numpy as np

type(np.fabs(479092))
# output >> numpy.float64

type(np.absolute(479092))
# output >> numpy.int64
```
I didn't pay attention that it is `fabs` function not `abs`.

pyccel/ast/numpyext.py
```diff
@@ -1185,11 +1186,11 @@ class NumpyArctanh(NumpyUfuncUnary) : pass
...
     def __init__(self, x):
         self._shape     = x.shape
         self._rank      = x.rank
-        self._dtype     = NativeInteger() if x.dtype is NativeInteger() else NativeReal()
+        self._dtype     = x.dtype
```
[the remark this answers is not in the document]
rakati: there are no tests for complex numbers.
I think the support of complex numbers will be in a separate PR for both Fortran and C printers.
what do you think?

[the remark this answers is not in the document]
EmilyBourne: I think the lack of tests for complexes is an oversight on my part. I think I copied the first half of the tests over from `test_math_funcs.py`. I think they are already supported but just not tested.
I am happy to leave the tests for another PR as there will probably be lots to add again, but I would revert this line which would break any existing support (I think this has already been done)

pyccel/codegen/printing/fcode.py
```diff
@@ -2215,7 +2216,13 @@ def _print_NumpyUfuncBase(self, expr):
             func_name = numpy_ufunc_to_fortran[type_name]
         except KeyError:
             errors.report(PYCCEL_RESTRICTION_TODO, severity='fatal')
-        code_args = ', '.join(self._print(i) for i in expr.args)
+        args = []
+        for arg in expr.args:
+            if arg.dtype is NativeInteger():
+                args.append(self._print(NumpyFloat(arg)))
+            else :
+                args.append(self._print(arg))
```
EmilyBourne, line 2224: This could be re-written with a list comprehension as:
```
args = [self_print(NumpyFloat(a) if a.dtype is NativeInteger() else a) for a in expr.args]
```
The line is a bit long so it might want breaking into 2, ie.:
```
args = [self_print(NumpyFloat(a) if a.dtype is NativeInteger() \
                            else a) for a in expr.args]
```
but list comprehension is much faster than a loop and append

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -1,12 +1,15 @@
...
+# pylint: disable=missing-function-docstring, missing-module-docstring, unidiomatic-typecheck/
+import sys
 import pytest
-from numpy.random import rand, randint
+from numpy.random import rand, randint, uniform
 from numpy import isclose
 
 from pyccel.decorators import types
 from pyccel.epyccel import epyccel
 from conftest import *
```
EmilyBourne, line 9: This import should no longer be required. It does not exist in the master branch. You probably need to merge the master branch into this branch (ideally, please do this in about 20 minutes once PR 498 has gone through)

EmilyBourne (changes requested):

pyccel/codegen/printing/fcode.py
rakati: done

EmilyBourne (approved):

yguclu (approved):

PR #485, opened 2020-10-20
Implement IfTernaryOperator in C and Fortran

### LINKED ISSUE #452: IfTernaryOperator is strangely implemented
**Describe the bug**
`IfTernaryOperator` currently inherits from `If`. As a result it is expected to have a body that contains complete statements, however this is not what an `IfTernaryOperator` is. As this concept exists in both fortran and c there is no reason to not implement it correctly.

**Current behaviour**
Consider the following code:
```python
b = True
a = 1 if b else 2
c = 25 + (a if b else 4)
```
After the syntactic stage line 2 contains:
`Assign( Symbol(a), IfTernaryOperator((b,1),(True,2)) )`
After the semantic stage this becomes:
`IfTernaryOperator( (b, Assign(Variable(a), 1)), (True, Assign(Variable(a), 2) )`
This is then printed using the `If` printer. It can no longer be printed as a ternary operator

After the syntactic stage line 3 contains:
`Assign( Symbol(a), IfTernaryOperator((b,1),(True,2)) )`
It then crashes at the semantic stage as an `IfTernaryOperator` does not inherit from `PyccelAstNode`

**Expected behavior**
When fixed line 2 should generate:
**Fortran**:
```fortran
a = merge(1,2,b)
```
**c**:
```c
a = b ? 1 : 2;
```

**Additional Comments**
If the `IfTernaryOperator` is implemented as a ternary operator then it no longer needs to inherit from `If` and does not need to store `True` as the condition for the last branch

MenoIy:
Fixes #452 
- Change class **IfTernaryOperator** to inherit from PyccelAstNode
- Change constructor of **IfTernaryOperator** to accept arguments (`cond`, `value_true`, `value_false`)
- Collect PyccelAstNode info from `value_true` and `value_false`
- Remove the code that transforms [Assign, IfTernaryOperator]  to a normal If
- Add Error and typing management to class
- Add C printer
- Add Fortran printer
- Since Fortran `merge` function doesn't accept different types, cast values to IfTernaryOperator expression type in Fortran printer

### DIFF
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -5203,8 +5203,13 @@ class IfTernaryOperator(Basic, PyccelAstNode):
         self._args = [cond, first, second]
         if self.stage == 'syntactic':
             return
+
+        if isinstance(first , Nil) or isinstance(second, Nil):
+            raise TypeError('None is not implemeted for Ternary Operator')
+        if first.shape != second.shape :
+            raise TypeError('results in Ternary Operator should be the same shape')
         if isinstance(first.dtype, NativeString) ^ isinstance(second.dtype, NativeString):
-             raise TypeError('Only one of the condition results is type string')
+            raise TypeError('Only one of the Ternary Operator results is type string')
         _tmp_list = [NativeBool(), NativeInteger(), NativeReal(), NativeComplex(), NativeString()]
         if first.dtype not in _tmp_list :
             raise TypeError('cannot determine the type of {}'.format(first.dtype))
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -5192,8 +5192,8 @@ class IfTernaryOperator(Basic, PyccelAstNode):
     >>> from pyccel.ast.core import Assign, IfTernaryOperator
     >>> n = Symbol('n')
     >>> x = 5 if n > 1 else 2
-    >>> IfTernaryOperator([PyccelGt(n > 1),  5,  2])
-    IfTernaryOperator([PyccelGt(n > 1),  5,  2])
+    >>> IfTernaryOperator(PyccelGt(n > 1),  5,  2)
+    IfTernaryOperator(PyccelGt(n > 1),  5,  2)
     """
     def __init__(self, *args):
         self._args = []
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -5178,9 +5178,6 @@ class If(Basic):
 
 
 class IfTernaryOperator(Basic, PyccelAstNode):
-    _rank  = 0
-    _shape = ()
-
     """Represents a IfTernaryOperator statement in the code.
 
     Parameters
@@ -5195,8 +5192,8 @@ class IfTernaryOperator(Basic, PyccelAstNode):
     >>> from pyccel.ast.core import Assign, IfTernaryOperator
     >>> n = Symbol('n')
     >>> x = 5 if n > 1 else 2
-    >>> IfTernaryOperator([PyccelGt(n > 1),  CodeBlock([5]),  CodeBlock([2])])
-    IfTernaryOperator([PyccelGt(n > 1),  CodeBlock([5]),  CodeBlock([2])])
+    >>> IfTernaryOperator([PyccelGt(n > 1),  5,  2])
+    IfTernaryOperator([PyccelGt(n > 1),  5,  2])
     """
     def __init__(self, *args):
         self._args = []
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -5208,10 +5208,13 @@ class IfTernaryOperator(Basic, PyccelAstNode):
             return
         if isinstance(first.dtype, NativeString) ^ isinstance(second.dtype, NativeString):
              raise TypeError('Only one of the condition results is type string')
-        _tmp_list = [NativeBool(), NativeInteger(), NativeReal(), NativeComplex(), NativeString()]
-        _tmp_elem = max([first, second], key = lambda x : _tmp_list.index(x.dtype))
-        self._dtype = _tmp_elem.dtype
-        self._precision = _tmp_elem.precision
+        _tmp_list = [NativeBool(), NativeInteger(), NativeReal(), NativeComplex()]
+        if first.dtype not in _tmp_list :
+            raise TypeError('cannot determine the type of {}'.format(first.dtype))
+        if second.dtype not in _tmp_list :
+            raise TypeError('cannot determine the type of {}'.format(second.dtype))
+        self._dtype = max([first.dtype, second.dtype], key = lambda x : _tmp_list.index(x))
+        self._precision = max([first.precision, second.precision])
 
 
 
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -5205,7 +5205,7 @@ class IfTernaryOperator(Basic, PyccelAstNode):
             return
 
         if isinstance(first , Nil) or isinstance(second, Nil):
-            raise TypeError('None is not implemeted for Ternary Operator')
+            raise NotImplementedError('None is not implemeted for Ternary Operator')
         if first.shape != second.shape :
             raise TypeError('results in Ternary Operator should be the same shape')
         if isinstance(first.dtype, NativeString) ^ isinstance(second.dtype, NativeString):
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -5203,11 +5203,8 @@ class IfTernaryOperator(Basic, PyccelAstNode):
         self._args = [cond, first, second]
         if self.stage == 'syntactic':
             return
-
         if isinstance(first , Nil) or isinstance(second, Nil):
             raise NotImplementedError('None is not implemeted for Ternary Operator')
-        if first.shape != second.shape :
-            raise TypeError('results in Ternary Operator should be the same shape')
         if isinstance(first.dtype, NativeString) ^ isinstance(second.dtype, NativeString):
             raise TypeError('Only one of the Ternary Operator results is type string')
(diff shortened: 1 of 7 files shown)


pyccel/ast/core.py
```diff
@@ -5177,15 +5177,56 @@ def bodies(self):
...
+        cond = args[0]
+        first = args[1]
+        second = args[2]
+        self._args = [cond, first, second]
+        if self.stage == 'syntactic':
+            return
 
-    pass
+        if isinstance(first , Nil) or isinstance(second, Nil):
+            raise TypeError('None is not implemeted for Ternary Operator')
```
EmilyBourne, line 5208: Could you use a `NotImplementedError` instead of a `TypeError`. We catch `NotImplementedError`s so they raise a neat error, but `TypeError`s should occur when a developer is using the code wrong as `TypeError`s will lead to back traces

pyccel/ast/core.py
```diff
@@ -5177,15 +5177,56 @@ def bodies(self):
...
+    """
     def __init__(self, *args):
-        for arg in self.args:
-            if len(arg[1].body)!=1:
-                raise TypeError('IfTernary body must be of length 1')
+        self._args = []
+        cond = args[0]
+        first = args[1]
+        second = args[2]
+        self._args = [cond, first, second]
```
EmilyBourne, line 5203: Why did you decide to use a list containing the three arguments rather than saving each into their own variable?

pyccel/ast/core.py
```diff
@@ -5177,15 +5177,56 @@ def bodies(self):
...
+    Examples
+    --------
+    >>> from sympy import Symbol
+    >>> from pyccel.ast.core import Assign, IfTernaryOperator
+    >>> n = Symbol('n')
+    >>> x = 5 if n > 1 else 2
+    >>> IfTernaryOperator([PyccelGt(n > 1),  5,  2])
+    IfTernaryOperator([PyccelGt(n > 1),  5,  2])
+    """
     def __init__(self, *args):
```
EmilyBourne, line 5198: Can an `IfTernaryOperator` be initialised with an arbitrary number of arguments? I would have thought that it would require exactly 3

pyccel/ast/core.py
```diff
@@ -5177,15 +5177,56 @@ def bodies(self):
...
+            raise TypeError('cannot determine the type of {}'.format(first.dtype))
+        if second.dtype not in _tmp_list :
+            raise TypeError('cannot determine the type of {}'.format(second.dtype))
+        self._dtype = max([first.dtype, second.dtype], key = lambda x : _tmp_list.index(x))
+        self._precision = max([first.precision, second.precision])
+        if None in [first.rank, second.rank]:
+            self._rank = None
+            self.shape = None
+        else :
+            self._shape = broadcast(first.shape, second.shape)
```
EmilyBourne, line 5224: It is interesting that you chose to use `broadcast` and yet you require the shape of `first` and `second` to be the same. Could this restriction be relaxed?

pyccel/codegen/printing/fcode.py
```diff
@@ -2058,6 +2066,20 @@ def _print_If(self, expr):
 
         return ''.join(lines)
 
+    def _print_IfTernaryOperator(self, expr):
+        args = expr.body
+        cond = self._print(PythonBool(args[0]))
+        first = args[1]
+        second = args[2]
+        try :
+            cast_func = python_builtin_datatypes[str_dtype(expr.dtype)]
```
EmilyBourne, line 2075: Does this mean that we can't use an `IfTernaryOperator` for strings in Fortran?

EmilyBourne (changes requested):

pyccel/ast/core.py
MenoIy: Actually, I started using lists at first because I was not sure about what I should do. And I finished by keeping it Normally it should be  3 arguments.

pyccel/ast/core.py
MenoIy: normally it should be 3 arguments for the format   _value_if_cond__true if _cond_ else _value_cond_false_

pyccel/ast/core.py
MenoIy: True, I didn't pay attention to _broadcast__  function. checking shape is not needed.

pyccel/codegen/printing/fcode.py
MenoIy: this should only cast only when types are different. for string, I did a check in the class when only one of ifternary result is a string  i raise an error otherwise if both results are strings it should work

pyccel/ast/core.py
EmilyBourne: In that case I think it would be more readable if we use `self._cond` `self._first` `self._second`

EmilyBourne: Or maybe even `self._cond`, `self._value_true`, `self._value_false`

pyccel/ast/core.py
EmilyBourne: In that case I think it would be better if we have:
`def __init__(self, cond, value_true, value_false):`
Currently if someone misunderstands how to use the `IfTernaryOperator` then they might provide extra args and not realise that they are unused. For example if they wanted to construct a case like:
`a if b else c if d else e`
then allowing a variable number of arguments might make them believe that they can do `IfTernaryOperator([b, a, d, c, e])`
It also means that we have to read the function body in order to know which arguments need to be passed

pyccel/codegen/printing/fcode.py
EmilyBourne: Ok, but this line is still called when handling Strings isn't it? So the error will still be raised. Please add a test with strings to check

pyccel/ast/core.py
```diff
@@ -5177,15 +5177,61 @@ def bodies(self):
...
+
+        if self.stage == 'syntactic':
+            return
+        if isinstance(value_true , Nil) or isinstance(value_false, Nil):
+            raise NotImplementedError('None is not implemeted for Ternary Operator')
+        if isinstance(value_true.dtype, NativeString) ^ isinstance(value_false.dtype, NativeString):
+            raise TypeError('Only one of the Ternary Operator results is type string')
+        _tmp_list = [NativeBool(), NativeInteger(), NativeReal(), NativeComplex(), NativeString()]
+        if value_true.dtype not in _tmp_list :
+            raise TypeError('cannot determine the type of {}'.format(value_true.dtype))
```
EmilyBourne, line 5211: As explained above please use a `NotImplementedError`

pyccel/ast/core.py
```diff
@@ -5177,15 +5177,61 @@ def bodies(self):
...
+        if value_false.dtype not in _tmp_list :
+            raise TypeError('cannot determine the type of {}'.format(value_false.dtype))
```
EmilyBourne, line 5213: As explained above please use a `NotImplementedError`

pyccel/ast/core.py
```diff
@@ -5177,15 +5177,61 @@ def bodies(self):
...
+        self._cond = cond
+        self._value_true = value_true
+        self._value_false = value_false
+
+        if self.stage == 'syntactic':
+            return
+        if isinstance(value_true , Nil) or isinstance(value_false, Nil):
+            raise NotImplementedError('None is not implemeted for Ternary Operator')
+        if isinstance(value_true.dtype, NativeString) ^ isinstance(value_false.dtype, NativeString):
+            raise TypeError('Only one of the Ternary Operator results is type string')
```
EmilyBourne, line 5208: Here it might be better to actually use `errors.raise` instead of a `NotImplementedError` as this is not something that we could implement in the future but is actually a pyccel restriction

EmilyBourne: E.g https://github.com/pyccel/pyccel/blob/master/pyccel/ast/core.py#L223

tests/epyccel/test_epyccel_IfTernaryOperator.py
```diff
@@ -0,0 +1,132 @@
...
+    # ...
+    assert f(6) == f5(6)
+    assert f(4) == f5(4)
+    assert f(5) == f5(5)
+    # ...
+#------------------------------------------------------------------------------
+def test_f6(language):
+    @types('int')
+    def f6(x):
+        a = x if x < 0 else 1 if x < 5 else complex(0, 1) if x == 5 else 6.5
```
yguclu, line 80: My brain is having a hard time parsing this line! :sweat_smile: 

Can you please add a comment line above this, where you add parentheses to make it understandable? I think it reads
```python
        a = x if x < 0 else (1 if x < 5 else (complex(0, 1) if x == 5 else 6.5))
```

yguclu (changes requested):

pyccel/ast/core.py
```diff
@@ -5177,15 +5177,61 @@ def bodies(self):
...
+    """
+    def __init__(self, cond, value_true, value_false):
+        self._cond = cond
+        self._value_true = value_true
+        self._value_false = value_false
+
+        if self.stage == 'syntactic':
+            return
+        if isinstance(value_true , Nil) or isinstance(value_false, Nil):
+            errors.report('None is not implemeted for Ternary Operator', severity='fatal')
```
EmilyBourne, line 5206: Typo `implemeted` -> `implemented`

pyccel/ast/core.py
```diff
@@ -5177,15 +5177,61 @@ def bodies(self):
...
+        if isinstance(value_true.dtype, NativeString) or isinstance(value_false.dtype, NativeString):
+            raise TypeError('Strings are not supported by Ternary Operator')
```
EmilyBourne, line 5208: As above, this should be an `errors.report`

tests/epyccel/test_epyccel_IfTernaryOperator.py
yguclu: Thanks!

yguclu (approved):

EmilyBourne (approved):

PR #486, opened 2020-10-21
Raise warning for undefined decorators

### LINKED ISSUE #348: Unrecognised decorators are ignored but no warning is raised
As a USER I would like to know if something in my code is not translated. E.g:
```
from mod import xfail

@xfail
def f():
    pass
```
This should raise a warning `xfail` not used

nhamidn:
Raise a warning if Pyccel encounters an unknown decorator (not defined in `pyccel.decorators`) and hence ignores it. Fixes #348

### DIFF
--- a/pyccel/ast/utilities.py
+++ b/pyccel/ast/utilities.py
@@ -218,7 +218,8 @@ def builtin_import(expr):
     if source == 'pyccel.decorators':
         funcs = [f[0] for f in inspect.getmembers(pyccel_decorators, inspect.isfunction)]
         for target in expr.target:
-            if str(target) not in funcs:
+            tmp = str(target).split()
+            if tmp[0] not in funcs:
                 errors = Errors()
                 errors.report("{} does not exist in pyccel.decorators".format(target),
                         symbol = expr, severity='error')
--- a/pyccel/ast/utilities.py
+++ b/pyccel/ast/utilities.py
@@ -218,8 +218,8 @@ def builtin_import(expr):
     if source == 'pyccel.decorators':
         funcs = [f[0] for f in inspect.getmembers(pyccel_decorators, inspect.isfunction)]
         for target in expr.target:
-            tmp = str(target).split()
-            if tmp[0] not in funcs:
+            search_target = target.name if isinstance(target, AsName) else target
+            if search_target not in funcs:
                 errors = Errors()
                 errors.report("{} does not exist in pyccel.decorators".format(target),
                         symbol = expr, severity='error')
--- a/pyccel/ast/utilities.py
+++ b/pyccel/ast/utilities.py
@@ -218,7 +218,7 @@ def builtin_import(expr):
     if source == 'pyccel.decorators':
         funcs = [f[0] for f in inspect.getmembers(pyccel_decorators, inspect.isfunction)]
         for target in expr.target:
-            search_target = target.name if isinstance(target, AsName) else target
+            search_target = target.name if isinstance(target, AsName) else str(target)
             if search_target not in funcs:
                 errors = Errors()
                 errors.report("{} does not exist in pyccel.decorators".format(target),
--- a/pyccel/errors/messages.py
+++ b/pyccel/errors/messages.py
@@ -60,6 +60,9 @@ WRONG_NUMBER_OUTPUT_ARGS = 'Number of output arguments does not match number of 
 INDEXED_TUPLE = 'Tuples must be indexed with constant integers for the type inference to work'
 LIST_OF_TUPLES = 'Cannot create list of non-homogeneous tuples'
 
+UNDEFINED_DECORATOR = 'Decorator not used'
+UNDEFINED_DECORATORS = 'Decorators not used'
+
 UNDEFINED_LAMBDA_VARIABLE = 'Unknown variable(s) in lambda function'
 UNDEFINED_LAMBDA_FUNCTION = 'Unknown function in lambda function'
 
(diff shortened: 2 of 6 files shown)


tests/warnings/semantic/UNDEFINED_DECORATOR.py
```diff
@@ -0,0 +1,4 @@
+
```
EmilyBourne, line 1: Missing pylint statement for test file

pyccel/ast/utilities.py
```diff
@@ -218,7 +218,8 @@ def builtin_import(expr):
     if source == 'pyccel.decorators':
         funcs = [f[0] for f in inspect.getmembers(pyccel_decorators, inspect.isfunction)]
         for target in expr.target:
-            if str(target) not in funcs:
+            tmp = str(target).split()
+            if tmp[0] not in funcs:
```
EmilyBourne, line 222: Can you give an example where split is required?

pyccel/errors/messages.py
```diff
@@ -60,6 +60,9 @@
 INDEXED_TUPLE = 'Tuples must be indexed with constant integers for the type inference to work'
 LIST_OF_TUPLES = 'Cannot create list of non-homogeneous tuples'
 
+UNDEFINED_DECORATOR = 'Decorator not used'
+UNDEFINED_DECORATORS = 'Decorators not used'
```
EmilyBourne, line 64: I'm not sure it's worth having 2 error messages just for this, I would just use:
`Decorator(s) not used`

EmilyBourne (changes requested):

pyccel/ast/utilities.py
nhamidn: for example:

`from pyccel.decorators import types as func_types`

` types as func_types` is considered as an import

EmilyBourne: Hmm, I think this would be clearer if you used:
```
target = target.target if isinstance(target, AsName) else target
```

nhamidn: this will make pyccel look for `func_types` in `pyccel.decorators`

in order to check for the imports properly i need to build an import dictionary which i will use to check if the import is defined or not.

EmilyBourne: > this will make pyccel look for `func_types` in `pyccel.decorators`
Ah, sorry then you should use:
`search_target = target.name if isinstance(target, AsName) else target`
> in order to check for the imports properly i need to build an import dictionary which i will use to check if the import is defined or not.
I am not sure I understand why this is necessary

tests/warnings/semantic/UNDEFINED_DECORATOR.py
```diff
@@ -0,0 +1,5 @@
+# pylint: disable=missing-function-docstring, missing-module-docstring/
+
+@toto
```
EmilyBourne, line 3: You can change this line to:
```
@toto # pylint: disable=undefined-variable
```
to disable `undefined-variable` for this line. If there was an `undefined-decorator` pylint flag then I would say disable it for the whole file, but in this case I'd rather just disable it for this line

pyccel/ast/utilities.py
nhamidn: ah sorry its not necessary.

pyccel/ast/utilities.py
```diff
@@ -218,7 +218,8 @@ def builtin_import(expr):
     if source == 'pyccel.decorators':
         funcs = [f[0] for f in inspect.getmembers(pyccel_decorators, inspect.isfunction)]
         for target in expr.target:
-            if str(target) not in funcs:
+            search_target = target.name if isinstance(target, AsName) else target
+            if search_target not in funcs:
```
EmilyBourne, line 222: Are you sure `str()` is no longer needed?

nhamidn: you are right `target` must be casted to `str`

pyccel/parser/semantic.py
```diff
@@ -2295,6 +2296,18 @@ def _visit_FunctionDef(self, expr, **settings):
...
+        keys = decorators.keys()
+        not_used = []
+        for value in keys:
+            if value not in def_decorators.__all__:
+                not_used.append(value)
+
+        if len(not_used) == 1:
+            errors.report(UNDEFINED_DECORATORS, symbol=''.join(not_used), severity='warning')
+        elif len(not_used) > 1:
+            errors.report(UNDEFINED_DECORATORS, symbol=', '.join(not_used), severity='warning')
```
EmilyBourne, line 2309: This separation is no longer needed

EmilyBourne (changes requested):

EmilyBourne (approved):

pyccel/parser/semantic.py
```diff
@@ -2295,6 +2296,16 @@ def _visit_FunctionDef(self, expr, **settings):
         is_private   = expr.is_private
 
         header = expr.header
+
+        keys = decorators.keys()
+        not_used = []
+        for value in keys:
+            if value not in def_decorators.__all__:
+                not_used.append(value)
```
yguclu, line 2304: This can be rewritten in a more compact way (also much faster) using a list comprehension:
```python
not_used = [d for d in decorators if d not in def_decorators.__all__]
```
Please note that when you loop over a dictionary you automatically get an iterator over its keys, so there is no need to use the `keys()` method.

yguclu (changes requested):

yguclu (approved):

PR #488, opened 2020-10-21
Handle negative indexes for arrays in fortran

### LINKED ISSUE #248: Conversion of negative indexes in arrays
Currently, `a[-1]` is translated literally to `a(-1)` in Fortran, causing an out-of-bounds warning, and probably not intended behavior. Would it make sense to change it to `a(ubound(a,dim))`? See https://stackoverflow.com/questions/20986269/how-do-i-access-last-item-in-an-array-in-fortran/20986963

EmilyBourne:
Handle negative constants as indexes for arrays. Also handle other indexes that could be negative if the new `allow_negative_index` decorator is specified. This fixes #248 

- Handle indexes which are negative constants by subtracting them from the shape
- Add `allow_negative_index` decorator and save property to `Variable`
- Add modulo operator when the index is not an integer and the user has added the `allow_negative_index` decorator for this variable
- Add tests

### DIFF
--- a/pyccel/parser/syntactic.py
+++ b/pyccel/parser/syntactic.py
@@ -632,6 +632,13 @@ class SyntaxParser(BasicParser):
             for i, arg in enumerate(args):
                 args[i] = str(arg).replace("'", '')
             decorators['stack_array'] = tuple(args)
+
+        if 'allow_negative_index' in decorators:
+            args = list(decorators['allow_negative_index'].args)
+            for i, arg in enumerate(args):
+                args[i] = str(arg).replace("'", '')
+            decorators['allow_negative_index'] = tuple(args)
+
         # extract the types to construct a header
         if 'types' in decorators:
             types = []
--- a/pyccel/parser/syntactic.py
+++ b/pyccel/parser/syntactic.py
@@ -630,13 +630,13 @@ class SyntaxParser(BasicParser):
         if 'stack_array' in decorators:
             args = list(decorators['stack_array'].args)
             for i, arg in enumerate(args):
-                args[i] = str(arg).replace("'", '')
+                args[i] = str(arg)
             decorators['stack_array'] = tuple(args)
 
         if 'allow_negative_index' in decorators:
             args = list(decorators['allow_negative_index'].args)
             for i, arg in enumerate(args):
-                args[i] = str(arg).replace("'", '')
+                args[i] = str(arg)
             decorators['allow_negative_index'] = tuple(args)
 
         # extract the types to construct a header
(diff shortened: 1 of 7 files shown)


pyccel/parser/syntactic.py
```diff
@@ -632,6 +632,13 @@ def _visit_FunctionDef(self, stmt):
             for i, arg in enumerate(args):
                 args[i] = str(arg).replace("'", '')
             decorators['stack_array'] = tuple(args)
+
+        if 'allow_negative_index' in decorators:
+            args = list(decorators['allow_negative_index'].args)
+            for i, arg in enumerate(args):
+                args[i] = str(arg).replace("'", '')
```
nhamidn, line 639: i didn't understand why using `.replace` here

EmilyBourne: You are right. I copied and pasted from the `stack_array` decorator. I suspect it is a hangover from the old parser

nhamidn (approved):

pyccel/parser/syntactic.py
```diff
@@ -630,8 +630,15 @@ def _visit_FunctionDef(self, stmt):
         if 'stack_array' in decorators:
             args = list(decorators['stack_array'].args)
             for i, arg in enumerate(args):
-                args[i] = str(arg).replace("'", '')
+                args[i] = str(arg)
             decorators['stack_array'] = tuple(args)
```
yguclu, line 634: If there is no need to perform any substitution in the arguments' names, we should rewrite this as
```python
decorators['stack_array'] = tuple(str(a) for a in decorators['stack_array'].args)
```

pyccel/parser/syntactic.py
```diff
@@ -630,8 +630,15 @@ def _visit_FunctionDef(self, stmt):
...
+
+        if 'allow_negative_index' in decorators:
+            args = list(decorators['allow_negative_index'].args)
+            for i, arg in enumerate(args):
+                args[i] = str(arg)
+            decorators['allow_negative_index'] = tuple(args)
```
yguclu, line 640: I have the same comment as above. We should rewrite this as
```python
decorators['allow_negative_index'] = tuple(str(a) for a in decorators['allow_negative_index'].args)
```

yguclu (changes requested):

yguclu (approved):

yguclu:
This should be the next PR to go through. So please abort the test builds and update the branch from master.

yguclu:
Never mind, I've just done it myself :wink:

PR #500, opened 2020-10-23
Make parallel epyccel work on functions

### LINKED ISSUE #451: epyccel crashes when called in parallel
**Describe the bug**
the epyccel function crashes when called in parallel

E.g

```
cat > test_epyccel.py
from mpi4py import MPI
comm = MPI.COMM_WORLD

def f():
    return 1

g = epyccel(f, comm=comm)
```
and when executed with  `mpirun -n 1 python3  test_epyccel.py`
it crashed with this error message

```
  File "test_epyccel.py", line 8, in <module>
    g = epyccel(f, comm=comm)
  File "/home/said/phd/pyccel/pyccel/epyccel.py", line 185, in epyccel
    fmod_path = os.path.abspath(fmod.__file__)
AttributeError: 'fortran' object has no attribute '__file__'
```

yguclu:
This closes #451


saidctb:
do we have an example that tests compiling modules in parallel ?, if yes we can add a tests for functions also

EmilyBourne (commented):
Can we add a test for this?

EmilyBourne:
I'm not sure why this is the only branch with this problem but the windows tests keep failing for `test_numpy_funcs.py`. I think the tolerance will need to be increased to convince it to pass

tests/epyccel/test_parallel_epyccel.py
```diff
@@ -0,0 +1,81 @@
...
+from pyccel.epyccel import epyccel
+
+@pytest.fixture(params=[
+    pytest.param('fortran', marks = pytest.mark.fortran),
+    pytest.param('c'      , marks = [pytest.mark.c,
+        pytest.mark.xfail(message='Arrays not implemented in C')])
+    ]
+)
+def language(request):
+    return request.param
```
EmilyBourne, line 15: This is not necessary. Pytest finds this automatically in the conftest.py file

pyccel/epyccel.py
```diff
@@ -181,28 +237,47 @@ def epyccel( inputs, **kwargs ):
...
-            fmod      = epyccel_seq( inputs, **kwargs )
-            fmod_path = os.path.abspath(fmod.__file__)
-            fmod_name = fmod.__name__
+            try:
+                mod, fun = epyccel_seq( python_function_or_module, **kwargs )
+                mod_path = os.path.abspath(mod.__file__)
+                mod_name = mod.__name__
+                fun_name = python_function_or_module.__name__ if fun else None
+                success  = True
+            except:
```
EmilyBourne, line 246: Codacy is complaining about this as it is very bad practice. Is this supposed to catch everything (even typos in the python code) and broadcast it or is this supposed to catch something more specific?

pyccel/epyccel.py
```diff
@@ -181,28 +237,47 @@ def epyccel( inputs, **kwargs ):
...
+                exc_info = sys.exc_info()
```
EmilyBourne, line 247: Does this collect information about the exception? Would it be useful to also add the information contained in the exception itself?

EmilyBourne (changes requested):

tests/epyccel/test_parallel_epyccel.py
yguclu: But I need to say that C fails...

pyccel/epyccel.py
yguclu: Yes, it should catch anything! I just need to broadcast the exception information, whatever it is.

pyccel/epyccel.py
yguclu: This catches the type, the value and the traceback. `exc_info` is a tuple.

tests/epyccel/test_parallel_epyccel.py
EmilyBourne: Ah ok, sorry. It's a shame that we don't have any tests for this that don't need arrays. But I suppose this should be a non-issue soon

pyccel/epyccel.py
EmilyBourne: Then I think we should add a comment explaining why this is an exception and deactivate the pylint error for this line

yguclu: Should I write `except Exception`?

EmilyBourne: maybe. That doesn't catch everything though. I think system.exit() isn't caught

EmilyBourne: I would just do something like:
```
# error handling carried out after broadcast to prevent deadlocks
except: # pylint: disable=bare-except
```

EmilyBourne (approved):

PR #501, opened 2020-10-26
Separate Python3 built-in functions from Numpy's functions

### LINKED ISSUE #258: Properly separate Python3 built-in functions from Numpy's and Math's functions
The confusion is evident in [pyccel.ast.utilities](https://github.com/pyccel/pyccel/blob/65fcab425c17bcb24097f2d11561c686633c34fe/pyccel/ast/utilities.py#L26), where the `math_functions` dictionary is used for importing functions that come from both `numpy` and `math`. 

For example, `abs` is a Python3 built-in function (in which case it calls the magic method `__abs__` of its argument) and also a Numpy function, but does not exist in `math`.

We should improve `pyccel.ast` and make sure that each of the functions that we support exists in one of the following modules:
- [x] pyccel.ast.builtins
- [x] pyccel.ast.numpy_ext
- [x] pyccel.ast.math_ext
- [x] pyccel.ast.itertools_ext

To avoid name clashes we must make sure that we use different names in each module, by prepending `Python`, `Numpy`, `Math`, or `Itertool` to the class names.

rakati:
Separate Python3 built-in functions from Numpy's functions:
- prefix numpy classes in numpyext.py by Numpy keyword.
- Add Numpy prefix to necessary printer methods.


EmilyBourne:
Looks like you missed a couple of uses. Codacy should be able to help you with this. It might be worth running the pylint command locally:
```
pylint --disable=all --enable=undefined-variable .
```
As codacy only checks the files that you have changed rather than the whole codebase

EmilyBourne:
Is this PR ready for review? We should try to merge it quickly in case changes are made in other branches involving the classes that you've renamed

pyccel/ast/numpyext.py
```diff
@@ -140,7 +138,7 @@ def __new__(cls, arg, dtype=None, order='C'):
         if dtype is None:
             dtype = arg.dtype
         dtype, prec = process_dtype(dtype)
-
+        print("numpy new array dtype", dtype)
```
EmilyBourne, line 141: Forgotten debugging comment?

EmilyBourne (changes requested):

EmilyBourne (approved):

EmilyBourne (approved):

pyccel/codegen/printing/fcode.py
```diff
@@ -1146,9 +1151,8 @@ def _print_Assign(self, expr):
 
         code = ''
         if (expr.status == 'unallocated') and not (expr.like is None):
-            stmt = ZerosLike(lhs=lhs_code, rhs=expr.like)
-            code += self._print(stmt)
-            code += '\n'
+            errors.report("Numpy array dtype=object not supported ", symbol=expr,
+                    severity='fatal')
```
yguclu, line 1155: Can you explain this change?

EmilyBourne: We discussed this. We thought that given the if condition in the line above, this case was designed to handle:
```
np.zeros_like(None)
Out[3]: array(0, dtype=object)
```
Actually looking at the line again, it seems that I misread it though. There is a `not` in the condition:
```
if (expr.status == 'unallocated') and not (expr.like is None):
```

EmilyBourne: The line has been wrong since you updated the constructor of ZerosLike but our tests never enter this line (probably because `status` is not reliably set). It raises a codacy error. But I don't know what it's supposed to do

yguclu: If we don't know what these lines were supposed to do, I suggest we revert this changes and add a comment:
```python
# TODO [YG, 30.10.2020]: verify what this check does and removed it if obsolete
```

What kind of Codacy error do we get here?

EmilyBourne: The codacy error is about the arguments. The constructor of `ZerosLike` is:
```
def __new__(cls, a, dtype=None, order='K', subok=True)
```
If we are lucky then Codacy will realise that this isn't actually a change (as with the examples in ast/parallel/mpi.py) but if we aren't then this will keep raising an error and will block the merge

rakati: Unexpected keyword argument 'rhs' in constructor call
`stmt = ZerosLike(lhs=lhs_code, rhs=expr.like)`

yguclu: What if we simply comment those lines out and add a TODO comment?

rakati: I agree with you, those lines will help in future implementation.
but I think we should also keep raising an error if so, to help in locating the problem.

yguclu: I see your point, but the error that you are raising is not correct, because there is a `not` in this condition:
```python
if (expr.status == 'unallocated') and not (expr.like is None):
```
So I'd rather just comment out the code to make Codacy happy for now.

PR #502, opened 2020-10-26
Support passing function as argument in Fortran

### LINKED ISSUE #477: Functions as arguments in Fortran
**Describe the feature**
Currently functions can't be passed as argument in Fortran.

**Test Code**

E.g
```
    #$ header function f()
    def f():
        return 0

    #$ header high_order((int)())
    def high_order(function):
        x = function()
        return x

    x = high_order(f)
```

E.g
```
    #$ header function f(int)
    def f(x):
        return x + 1

    #$ header function high_order((int)(int), int)
    def high_order(function, x):
        x =  function(x)
        return x

    x = high_order(f, 0)
```

**Language**
Fortran

rmahjoubi:
**Allow functions to be passed as arguments in Fortran**

-Add Interface attribute to FunctionDef class.
-Add code generation for functions interfaces.
-Create function_signature method which handles both functions definitions and procedures of an interface.
-Separately compile the module which contains functions that takes functions as arguments.
-Add tests for functions that takes functions as arguments. 

Fixes #477


tests/pyccel/test_pyccel.py
```diff
@@ -179,6 +184,7 @@ def compare_pyth_fort_output( p_output, f_output, dtype=float ):
 #------------------------------------------------------------------------------
 def pyccel_test(test_file, dependencies = None, compile_with_pyccel = True,
         cwd = None, pyccel_commands = "", output_dtype = float,
+        compile_dep_with_pyccel = True,
```
EmilyBourne, line 187: I'm not sure if this is necessary. I am happy to compile the dependencies with pyccel every time we compile the program with pyccel

tests/pyccel/test_pyccel.py
```diff
@@ -377,6 +388,14 @@ def test_expressions(language):
             [float]*3 + [int]*2 + [float]*2 + [int]*5 + [complex] + [bool]*9
     pyccel_test("scripts/expressions.py", language=language,
                 output_dtype = types)
+#------------------------------------------------------------------------------
+def test_highorder():
+    pyccel_test("scripts/runtest_highorder_functions.py",
+            dependencies = "scripts/highorder_functions.py",
+            compile_dep_with_pyccel = False,
+            output_dtype = [int,int,float,float,float,int,float,
+                float, int])
```
EmilyBourne, line 397: This test should work in fortran and c

pyccel/codegen/printing/fcode.py
```diff
@@ -1408,32 +1412,68 @@ def _print_FunctionDef(self, expr):
...
+        else:
+            for i in _default_methods:
+                # because we may have a class Point with init: Point___init__
+                if i in name:
+                    name = name.replace(i, _default_methods[i])
+
+        signature = self.function_signature(expr, name)
+        decs = OrderedDict()
+        functions = expr.functions
+        func_interfaces = '\n'.join(self._print(i) for i in expr.interfaces)
```
EmilyBourne, line 1451: Has this been tested? I am not sure if we are allowed to write an interface in a function

pyccel/codegen/printing/fcode.py
```diff
@@ -1383,14 +1383,18 @@ def _print_FunctionDef(self, expr):
 
         for i,arg in enumerate(expr.arguments):
             if expr.arguments_inout[i]:
-                dec = Declare(arg.dtype, arg, intent='inout')
-
+                if isinstance(arg, FunctionAddress):
+                    dec = FuncAddressDeclare(arg, intent='inout')
```
EmilyBourne, line 1387: Is this possible?

pyccel/codegen/printing/fcode.py
```diff
@@ -1319,43 +1332,29 @@ def _print_F2PYFunctionDef(self, expr):
...
-        decs = OrderedDict()
         args_decs = OrderedDict()
-        # ... local variables declarations
+
+        for j, i in enumerate(expr.results):
+            if not i.name:
+                i.rename('out_{}'.format(j))
+        for j, i in enumerate(expr.arguments):
+            if not i.name:
+                i.rename('in_{}'.format(j))
```
EmilyBourne, line 1355: Is this still valid after the changes made in branch `devel_444` ? I thought the `rename` function was removed?

EmilyBourne:
I have had a look at `pyccel/codegen/printing/fcode.py` and everything in `tests/pyccel/` as I know that these are new. But I'm not sure what is new everywhere else. I had a quick look, but I don't think you've merged the changes from `devel_444` recently so some of my comments would have been about things that have been fixed there. If you want us to review this branch properly while we're waiting for the last things to be sorted on PR #479 you can modify this PR to ask to merge `devel_477` into `devel_444`, that way it's obvious what is new. You can then point it back at master once PR #479 is merged

tests/pyccel/test_pyccel.py
rmahjoubi: we are already running the same tests in epyccel  for c, I think there is no need to run them twice .?

EmilyBourne: So then do you think we should remove this test once the epyccel version can be activated for fortran? That sounds ok to me

tests/pyccel/test_pyccel.py
rmahjoubi: the dependency in this case is the module containing the functions that takes functions as arguments, compiling it with pyccel will lead to error at shared library generation stage

pyccel/codegen/printing/fcode.py
rmahjoubi: can you please tell me why we are not allowed to write an interface in a function. if this is the case could you also give an alternative solution.

tests/pyccel/test_pyccel.py
EmilyBourne: Sorry I said that backwards. I meant:
 I am happy to compile the dependencies with gfortran every time `compile_with_pyccel = False`

pyccel/codegen/printing/fcode.py
rmahjoubi: maybe in c if you like to change to which function the pointer is pointing to and then  return it, here i don't think it is. I'am gonna remove it

pyccel/codegen/printing/fcode.py
EmilyBourne: I don't know if we are allowed or not. That is why I am asking if it has been tested. If we are not allowed then it would be due to the fortran standard. It is sometimes quite picky about where objects can be placed (e.g. all variables must be declared at the start of a function, you can have a function in a function, but not  a function in a function in a function).
If you have tested it and it works then we should be ok.

pyccel/codegen/printing/fcode.py
EmilyBourne: But here we are translating from fortran. If you do:
```
def f(a):
    a  = new_func

b = old_func
f(b)
```
Then b is not modified

pyccel/codegen/printing/fcode.py
EmilyBourne: Sorry, this is a real rename function

rmahjoubi: this is an actual rename function, i just added this method recently, it gives names to the dummy arguments of the functions inside the interface

tests/pyccel/test_pyccel.py
rmahjoubi: I agree

EmilyBourne: I have added it to the list of things to do for issue #507

pyccel/codegen/printing/fcode.py
rmahjoubi: what i mean is for:
```
def f(a):
    a = new_func
    return a
```
we should have the ability to change a inside the function
this is kind of irrelevant to your question

tests/pyccel/test_pyccel.py
rmahjoubi: yes this makes sense

pyccel/codegen/printing/fcode.py
rmahjoubi: yes I think it's allowed

EmilyBourne: Does that mean you've tested it?

rmahjoubi: yes, all the tests uses a generated code that contains interfaces inside functions

EmilyBourne: Ok, perfect

pyccel/codegen/printing/fcode.py
EmilyBourne, line 1355: Why do we need to rename the variables?

pyccel/codegen/printing/fcode.py
EmilyBourne: If you mean:
```
def f(a):
    a = new_func
    return a
```
Then for fortran this would be translated by something like:
```
def f(a):
    a = new_func
    Out_0001 = a
    return Out_0001
```
so a is still an `in` not an `inout`. If you want to be able to do this, we may need to pass by value. I don't know if this is possible for procedures?
@yguclu  do you know what the standard says about this?

pyccel/ast/f2py.py
```diff
@@ -121,7 +124,7 @@ def as_static_function(func, name=None):
     int_kind = Variable('int', 'f2py_array_dimension').precision
 
     for a in args:
-        if a.rank > 1 and a.order == 'C':
+        if rank and a.rank > 1 and a.order == 'C':
```
EmilyBourne, line 127: Is `rank` correctly defined here?

pyccel/codegen/printing/fcode.py
yguclu: I don't think that the `intent` keyword makes any sense for argument functions in Fortran...

yguclu: I've just checked, specifying the `intent` attribute for a dummy procedure is not allowed. It cannot be passed by `value` either.

pyccel/codegen/printing/fcode.py
rmahjoubi: when a function is passed as an argument, we leave the name of it variables empty, as we only need them in case of fortran.
so these are just dummy variables of the prototype. we need to rename them so we can properly calculate `is_procedure`, they are also needed to print the call of a function argument with multiple returns.

pyccel/ast/f2py.py
rmahjoubi: it's not correctly defined. do you think it's good to have `FunctionAddress` inherits from `Variable` so we can avoid this kind of checks, in this case for a `FunctionAddress` the rank would be 0, it's as if we are treating it as a c pointer.

EmilyBourne: I think we already discussed this and decided that it was not quite a `Variable`. That said given the number of places in the code where you need to do `ifinstance(a, (Variable, FunctionAddress))` I am not sure I agree. What do you think @rmahjoubi @yguclu ? As a minimum even if it doesn't inherit from `Variable` I think it should inherit from `PyccelAstNode`

rmahjoubi: I think so

EmilyBourne: My comment here originally was not about whether `rank` is correctly defined with regards to `FunctionAddress` but whether the variable `rank` in this line has meaning.
The previous time that you use rank you define it (https://github.com/pyccel/pyccel/pull/502/files#diff-0243435837f0679be53be6bdf9816fa4b1f5d7cd3410affd18f0e7508a9726f0R77-R80) however we are now in a new loop and `rank` still has the value of the last argument of the previous loop.
This could be why your tests are failing

rmahjoubi: in the recent commit, i got rid of the variable `rank` completely, now i'am using `a.rank` only where a is the variable.

EmilyBourne: :+1: yes I saw, I posted my comment a few seconds before your commit ^^'
We can resolve this conversation once the issue about FunctionAddress and Variable has been created

rmahjoubi: I just created it

pyccel/ast/core.py
```diff
@@ -3303,6 +3316,10 @@ def functions(self):
     def hide(self):
         return self.functions[0].hide or self._args[2]
 
+    @property
+    def is_argument(self):
+        return self._args[3]
```
EmilyBourne, line 3321: Can we use this PR to make Instance use variables instead of storing everything in a big tuple? The more things there are in `self._args` the harder it gets to read the code

EmilyBourne (approved):

pyccel/ast/core.py
rmahjoubi: Yes sure

pyccel/parser/semantic.py
```diff
@@ -2274,17 +2274,18 @@ def _visit_Return(self, expr, **settings):
...
+        cls_name        = expr.cls_name
+        hide            = False
+        kind            = 'function'
+        decorators      = expr.decorators
+        funcs           = []
+        sub_funcs       = []
+        func_interfaces  = []
+        is_pure         = expr.is_pure
+        is_elemental    = expr.is_elemental
+        is_private      = expr.is_private
```
yguclu, line 2288: I assume that you changed all these lines to align the `=` signs after you added one line:
```python
func_interfaces = []
```
But the result is not exactly what you intended, because you have an extra space just before the equal sign!

rmahjoubi: yea, I didn't notice it

pyccel/codegen/printing/fcode.py
```diff
@@ -1408,32 +1402,65 @@ def _print_FunctionDef(self, expr):
             sig = 'elemental {}'.format(sig)
 
         arg_code  = ', '.join(self._print(i) for i in chain( expr.arguments, out_args ))
+
+        imports = ''.join(self._print(i) for i in expr.imports)
+        prelude = ''.join(self._print(i) for i in args_decs.values())
+        parts = ["{}({}) {}\n".format(sig, arg_code, func_end,),
+                imports,
+                "implicit none\n",
```
yguclu, line 1410: This line can be removed: since there are no local variables in a function definition, there is no danger of an implicit type declaration (Fortran-77 style)

pyccel/codegen/printing/fcode.py
```diff
@@ -1408,32 +1402,65 @@ def _print_FunctionDef(self, expr):
             sig = 'elemental {}'.format(sig)
 
         arg_code  = ', '.join(self._print(i) for i in chain( expr.arguments, out_args ))
+
+        imports = ''.join(self._print(i) for i in expr.imports)
+        prelude = ''.join(self._print(i) for i in args_decs.values())
+        parts = ["{}({}) {}\n".format(sig, arg_code, func_end,),
```
yguclu, line 1408: I suggest to remove `\n` from this string, in order to make the printed interface a little more compact.

pyccel/codegen/printing/fcode.py
```diff
@@ -1225,6 +1225,15 @@ def _print_String(self, expr):
     def _print_Interface(self, expr):
         # ... we don't print 'hidden' functions
         name = self._print(expr.name)
+        if expr.is_argument:
+            sigs = ''
+            for f in expr.functions:
+                self._handle_fortran_specific_a_prioris(list(f.arguments) + list(f.results))
+                sigs += self.function_signature(f, f.name) + '\n'
+                sigs += 'end %s\n' % sigs.split()[0]
```
yguclu, line 1233: It would be nice to separate the functions' signatures with an empty line, and to properly indent them.

@EmilyBourne Should we add the `interface` keyword to the `indent_code` function?

pyccel/ast/core.py
```diff
@@ -2782,6 +2782,18 @@ def clone(self, name, new_class = None, **kwargs):
             is_optional=kwargs.pop('is_optional',self.is_optional),
             cls_base=kwargs.pop('cls_base',self.cls_base),
             )
+    def rename(self, newname):
+        """
+        Rename the Variable name
+        newname.
```
yguclu, line 2788: This docstring is a bit confusing. Can we simply write "Change variable name"?

pyccel/codegen/printing/fcode.py
rmahjoubi: This function, is used in printing both function definition and interfaces, should i return a dictionary and then decide how to print in the calling method ?

tests/pyccel/scripts/highorder_functions.py
```diff
@@ -0,0 +1,90 @@
...
+@types()
+def f10():
+    y = 0
+    if y:
+        pass
+
+#$ header function high_int_1((int)(int), int)
+def high_int_1(function, a):
+    x = function(a)
+    return x
```
yguclu, line 49: `function` is a reserved keyword in Fortran. I am a bit surprised that the generated code actually compiles!

yguclu (changes requested):
Very good job @rmahjoubi! I've left a few comments scattered around..

pyccel/codegen/printing/fcode.py
yguclu: Sorry, I thought this was only used for the interface (given its name). Yes. that's a good idea.

pyccel/codegen/printing/fcode.py
EmilyBourne: Yes I think so

yguclu (approved):

EmilyBourne (approved):

tests/pyccel/scripts/highorder_functions.py
yguclu: I stand corrected. The Fortran standard defines many **keywords**, but it does **not** define any **reserved words**. Hence calling a function `function` is totally legal, but it's definitely not a good practice.

PR #505, opened 2020-10-26
Map Python imports to C imports

### LINKED ISSUE #503: The OpenMP Fortran library (omp_lib) is imported in the generated C code
**Describe the bug**
`omp_lib` is imported in the generated C code when using an OpenMP function.

**To Reproduce**
```

@types(int)
def set_num_threads(n):
    from pyccel.stdlib.internal.openmp import omp_set_num_threads
    omp_set_num_threads(n)
```

generate this code in C:
```
#include <test.h>
#include <stdlib.h>

void set_num_threads(long n)
{
    #include <omp_lib.h>


    omp_set_num_threads(n);
}
```

and this python code: 
```
n = 5

from pyccel.stdlib.internal.openmp import omp_set_num_threads
omp_set_num_threads(n)
```

generate this code in C: 

```
#include <omp.h>
#include <omp_lib.h>
#include <stdlib.h>
int main()
{
    long n;
    n = 5;

    omp_set_num_threads(n);
    return 0;
}
```
Both of those generated codes gave a compilation error because of `omp_lib` is unrecognized.
**Expected behavior**
`omp_lib` must not be used for C language, and the import must not be printed inside the `FunctionDef.`
For the first case the imported printed inside the `FunctionDef `because this method use the import so removing the printing of the import from the `_print_FunctionDef `method will solve this problem.
But for the `omp_lib` i didnt found when pyccel add it to the list of imports.

**Language**
C

This issue block the activation of OpenMP tests for C language.

EmilyBourne:
Create a dictionary which maps known module names (eg numpy, omp_lib, etc) onto their c equivalent (eg. None, omp, etc). Fixes #503

### DIFF
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -411,8 +411,11 @@ class CCodePrinter(CodePrinter):
         else:
             source = self._print(expr.source)
 
+        # Get with a default value is not used here as it is
+        # slower and on most occasions the import will not be in the
+        # dictionary
         if source in import_dict:
-            source = import_dict[source]
+            source = import_dict[source] # pylint: disable=consider-using-get
 
         if source is None:
             return ''
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -189,6 +189,8 @@ dtype_registry = {('real',8)    : 'double',
                   ('int',1)     : 'char',
                   ('bool',4)    : 'bool'}
 
+import_dict = {'numpy'   : None,
+               'omp_lib' : 'omp' }
 
 class CCodePrinter(CodePrinter):
     """A printer to convert python expressions to strings of c code"""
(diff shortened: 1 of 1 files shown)


nhamidn:
when declaring an **OpenMP** import inside a function shouldn't we move the `omp.h` import outside of the function or to the header file in the case of shared library ?

this python code : 

```
@types(int)
def set_num_threads(n):
    from pyccel.stdlib.internal.openmp import omp_set_num_threads
    omp_set_num_threads(n)
```

generate this code : 

```
#include <test.h>
#include <stdlib.h>

void set_num_threads(long n)
{
    #include <omp.h>


    omp_set_num_threads(n);
}
```

nhamidn (approved):

yguclu:
@EmilyBourne In the description you mention mapping Numpy to `None`, but I don't see this in the changes...

yguclu (approved):

EmilyBourne:
> @EmilyBourne In the description you mention mapping Numpy to `None`, but I don't see this in the changes...

I removed it as `numpy` is not added to the imports and will therefore not reach the codegen stage. But I decided to keep it in the comments to explain how this would handle the case where we save the import as it is needed in some languages (e.g. math) but do not need to print it in c. I am not sure that this is necessary though as libraries such as math/numpy are usually handled differently as it is not because they are in one module in python that this is necessarily the case in c.
Having this option lets us keep math/numpy/etc in the imports if we want to later though, which would presumably simplify python printing

PR #519, opened 2020-10-29
Multiple results cannot be returned in C #443

### LINKED ISSUE #443: Multiple results cannot be returned in C
**Describe the bug**
A C function can only return 1 argument. It would be nice if it could return multiple arguments

**Test Code**
```python
@pure
def const_int_float():
    return 1, 3.4
```

**Proposed Implementation**
The function above would be translated as follows:
```c
void const_int_float(int* result_1, float* result_2)
{
    (*result_1) = 1;
    (*result_2) = 3.4;
}
```

**Language**
C

MenoIy:
Fixes #443 
The approach used to solve the issue : 
**PART 1 :**
- As in Fortran we add the return to the function arguments as an address
- Create a temporary list of additional arguments.
 -> When a printing function definition we create a list of temporary variables as pointer corresponding to the function returns.
 example :
```
def func()
     return 1, 2.5  , 3
```
 will create the given list :
```
[long *tmp1,  double * tmp2,  long *tmp3]
```
this list of arguments will be used when printing the function signature. 
Assign each function results to the corresponding temporary variable at the end of the function definition.
Change the type of function return to void.
Not printing the  return
For example the function above will be:
```
void func(long *tmp_ret, double *tmp_ret_0001, long *tmp_ret_0002)
{
   // add list of temporary variable to function signature .
   // function type is void now.
    long Out_0001;
    double Out_0002;
    long Out_0003;

    Out_0001 = 1;
    Out_0002 = 2.5;
    Out_0003 = 3;

    // Assigning each result to it given temporary variable.
    // No return is printed
    (*tmp_ret) = Out_0001;
    (*tmp_ret_0001) = Out_0002;
    (*tmp_ret_0002) = Out_0003;
}
```
 ->  When printing function signature is headers or another file we collect the temporary list of variables from  funcdef results .
 -> When Assign  Assigning function call to multiple variable
 The additional args list is cleared after the printing of the function signature.
example:
```
x, y, z = func()
```
we change the assign code block to a normal function call  and passe the left side variable address to the function call
```
func(&x, &y, &x)
```
**PART 2:**
When calling  print on a function with multiple returns 
```
print(func())
```
we create temporary variables to hold function return and passe them to the print function we the forme (x1, x2. .....)
```
int tmp1;
int tmp2;
int tmp3;

func(&tmp1, &tmp2, &tmp3)
printf("(%ld, %f, %ld)", tmp1, tmp2, tmp3);
```

### DIFF
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -673,7 +673,7 @@ class CCodePrinter(CodePrinter):
         self._additional_declare.clear()
 
         if len(expr.results) > 1:
-            self._additional_args = [a.clone(name = self._parser.get_new_name('return'), is_pointer =True) for a in expr.results]
+            self._additional_args = [a.clone(name = self._parser.get_new_name('tmp_ret'), is_pointer =True) for a in expr.results]
             body += '\n'.join(self._print_Assign(Assign(a, b)) for  a, b in zip(self._additional_args, expr.results))
 
         sep = self._print(SeparatorComment(40))
@@ -752,7 +752,8 @@ class CCodePrinter(CodePrinter):
         args = [VariableAddress(a) if self.stored_in_c_pointer(a) else a for a in expr.expr]
         if expr.stmt:
             code += self._print(expr.stmt)+'\n'
-        code +='return {0};'.format(self._print(args[0]))
+        if len(args) == 1:
+            code +='return {0};'.format(self._print(args[0]))
         return code
 
     def _print_Nil(self, expr):
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -223,6 +223,7 @@ class CCodePrinter(CodePrinter):
         self._parser = parser
         self._additional_code = ''
         self._additional_declare = []
+        self._additional_args = []
 
     def _get_statement(self, codestring):
         return "%s;" % codestring
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -477,20 +477,21 @@ class CCodePrinter(CodePrinter):
         end = '\n'
         sep = ' '
         for f in expr.expr:
-            arg_format, arg = '', ''
             if isinstance(f, ValuedVariable):
                 if f.name == 'sep'      :   sep = str(f.value)
                 elif f.name == 'end'    :   end = str(f.value)
             elif isinstance(f, FunctionCall) and isinstance(f.dtype, NativeTuple):
                 tmp_list = self.extract_function_call_results(f)
+                tmp_arg_format_list = []
                 for a in tmp_list:
                     arg_format, arg = self.get_print_format_and_arg(a)
+                    tmp_arg_format_list.append(arg_format)
+                    args.append(arg)
+                args_format.append('({})'.format(', '.join(tmp_arg_format_list)))
                 assign = Assign(tmp_list, f)
                 self._additional_code += self._print(assign) + '\n'
             else:
                 arg_format, arg = self.get_print_format_and_arg(f)
-
-            if arg_format and arg:
                 args_format.append(arg_format)
                 args.append(arg)
 
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -438,8 +438,8 @@ class CCodePrinter(CodePrinter):
                                .replace("'", "\\'")
         return '"{}"'.format(format_str)
 
-    def _print_Print(self, expr):
-        self._additional_imports.add("stdio")
+
+    def get_print_format_and_arg(self, var):
         type_to_format = {('real',8)    : '%.12lf',
                           ('real',4)    : '%.12f',
                           ('complex',8) : '(%.12lf + %.12lfj)',
@@ -450,6 +450,20 @@ class CCodePrinter(CodePrinter):
                           ('int',1)     : '%c',
                           ('bool',4)    : '%s',
                           ('string', 0) : '%s'}
+        try:
+            arg_format = type_to_format[(self._print(var.dtype), var.precision)]
+        except KeyError:
+            errors.report("{} type is not supported currently".format(var.dtype), severity='fatal')
+        if var.dtype is NativeComplex():
+            arg = self._print(NumpyReal(var)), self._print(NumpyImag(var))
+        elif var.dtype is NativeBool():
+            arg = '{} ? "True" : "False"'.format(self._print(var))
+        else:
+            arg = self._print(var)
+        return arg_format, arg
+
+    def _print_Print(self, expr):
+        self._additional_imports.add("stdio")
         args_format = []
         args = []
         end = '\n'
@@ -459,17 +473,9 @@ class CCodePrinter(CodePrinter):
                 if f.name == 'sep'      :   sep = str(f.value)
                 elif f.name == 'end'    :   end = str(f.value)
             else:
-                try:
-                    args_format.append(type_to_format[(self._print(f.dtype), f.precision)])
-                except KeyError:
-                    errors.report("{} type is not supported currently".format(\
-                        f.dtype), severity='fatal')
-                if f.dtype is NativeComplex():
-                    args.extend([self._print(NumpyReal(f)), self._print(NumpyImag(f))])
-                elif f.dtype is NativeBool():
-                    args.append('{} ? "True" : "False"'.format(self._print(f)))
-                else:
-                    args.append(self._print(f))
+                arg_format, arg = self.get_print_format_and_arg(f)
+                args_format.append(arg_format)
+                args.append(arg)
 
         args_format = sep.join(args_format)
         args_format += end
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -828,6 +828,9 @@ class CCodePrinter(CodePrinter):
         return "{0} {1}= {2};".format(lhs_code, op, rhs_code)
 
     def _print_Assign(self, expr):
+        if not isinstance(expr.lhs, Variable) and isinstance(expr.rhs, FunctionCall):
+            self._additional_args = [VariableAddress(a) for a in expr.lhs]
+            return '{};'.format(self._print(expr.rhs))
         lhs = self._print(expr.lhs)
         rhs = self._print(expr.rhs)
         return '{} = {};'.format(lhs, rhs)
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -11,7 +11,7 @@ from pyccel.ast.core import DottedName
 from pyccel.ast.core import PyccelAdd, PyccelMul, String
 
 from pyccel.ast.datatypes import default_precision
-from pyccel.ast.datatypes import NativeInteger, NativeBool, NativeComplex, NativeReal
+from pyccel.ast.datatypes import NativeInteger, NativeBool, NativeComplex, NativeReal, NativeTuple
 
 
 from pyccel.ast.numpyext import NumpyFloat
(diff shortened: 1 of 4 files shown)


pyccel/codegen/printing/ccode.py
```diff
@@ -449,6 +450,28 @@ def _print_Print(self, expr):
...
+            arg = '{} ? "True" : "False"'.format(self._print(var))
+        else:
+            arg = self._print(var)
+        return arg_format, arg
+
+    def extract_function_call_results(self, expr):
+        tmp_list = []
+        for a in expr.funcdef.results :
+            tmp_var = self.create_tmp_var(a)
+            tmp_list.append(tmp_var)
```
EmilyBourne, line 469: This would be faster with a list comprehension:
```
tmp_list = [self.create_tmp_var(a) for a in expr.funcdef.results]
```

pyccel/codegen/printing/ccode.py
```diff
@@ -457,18 +480,20 @@ def _print_Print(self, expr):
             if isinstance(f, ValuedVariable):
                 if f.name == 'sep'      :   sep = str(f.value)
                 elif f.name == 'end'    :   end = str(f.value)
+            elif isinstance(f, FunctionCall) and isinstance(f.dtype, NativeTuple):
```
EmilyBourne, line 483: Are (inhomogenous) tuples supported now? Or just for the print stage?

pyccel/codegen/printing/ccode.py
```diff
@@ -668,6 +698,11 @@ def _print_FunctionDef(self, expr):
         decs += [Declare(i.dtype, i) for i in self._additional_declare]
         decs  = '\n'.join(self._print(i) for i in decs)
         self._additional_declare.clear()
+
+        if len(expr.results) > 1:
+            self._additional_args = [a.clone(name = self._parser.get_new_name('tmp_ret'), is_pointer =True) for a in expr.results]
```
EmilyBourne, line 703: Why do the variables need renaming?

EmilyBourne, line 703: When you clone you create a new object which can break references to this variable elsewhere (they will not be updated to be pointers). How are you working round this problem?

EmilyBourne (changes requested):

pyccel/codegen/printing/ccode.py
MenoIy: No this is only for the print stage .

pyccel/codegen/printing/ccode.py
```diff
@@ -668,6 +698,12 @@ def _print_FunctionDef(self, expr):
...
+            body += '\n'.join(self._print_Assign(Assign(a, b)) for  a, b in zip(self._additional_args, expr.results))
+            body += '\n' + self._print_Return(Return([1]))
```
EmilyBourne, line 705: Why do you always return 1?

pyccel/codegen/printing/ccode.py
EmilyBourne: I have just realised that we cannot do operations with these variables so this is not a problem :+1:

pyccel/codegen/printing/ccode.py
MenoIy: what I mean by that is that function with multiple results is always true 
```
def func():
    return 0, 0
bool(func())  ===> True
```

pyccel/codegen/printing/ccode.py
MenoIy: it's not renaming I'm creating a list of new temp variables

pyccel/codegen/printing/ccode.py
MenoIy: And maybe I need to make it return boolean better

EmilyBourne: Hmm, personally I am not a fan of doing return True when it cannot be False, but it can sometimes be useful so I am fine with you leaving it.
But if you want to do this then I would either:
1. return True (bools are implemented in c, why not use them?)
2. return 0 (integer code for no error)

pyccel/codegen/printing/ccode.py
EmilyBourne: Ok. Can you post a code snippet so I can see what this generates? I think results are already temporary variables so I'm worried you are going to get something like:
```
def f():
    a = 1
    b = 3.0
    return a, b
```
```
int f(int* tmp_ret_1, float* tmp_ret_2)
{
    int a = 1;
    int b = 3.0;
    int Out_1 = a;
    float Out_2 = b;
    (*tmp_ret_1) = Out_1;
    (*tmp_ret_2) = Out_2;
}
```

MenoIy: i didn't really understand  but the generated code  is like the one that you write 
```int f(int* tmp_ret_1, float* tmp_ret_2)
{
    int a = 1;
    int b = 3.0;
    int Out_1 = a;
    float Out_2 = b;
    (*tmp_ret_1) = Out_1;
    (*tmp_ret_2) = Out_2;
}
```

EmilyBourne: The variables `Out_X` are created because in fortran all returns need to have the same name.

In c or python we can do:
```
def f(b):
    a = 1
    c = 2
    if b:
        return a
    else:
        return c
```
but in fortran this is not possible. So these `Out_X` variables were created so that the code becomes:
```
def f(b):
    a = 1
    c = 2
    if b:
        Out_1 = a
        return Out_1
    else:
        Out_1 = c
        return Out_1
```
These `Out_X` variables are not used anywhere else They are just temporary variables for the output.
Your `tmp_ret_X` variables do the same job. So you will be doing the same thing twice. The code above would generate the following if you used similar logic with 1 return:
```
def f(b):
    a = 1
    c = 2
    if b:
        Out_1 = a
        (*tmp_ret_1) = Out_1
        return tmp_ret_1
    else:
        Out_1 = c
        (*tmp_ret_1) = Out_1
        return tmp_ret_1
```
It shouldn't be necessary to create a new temporary variable. You should be able to use the existing `Out_X` variables. So the example from before should create:
```
int f(int* Out_1, float* Out_2)
{
    int a = 1;
    int b = 3.0;
    (*Out_1) = a;
    (*Out_2) = b;
}
```
which is cleaner

EmilyBourne: The difficulty is how to implement this ^^'. Ideally there should be some way of knowing that the `Out_X` variables are part of a multiple return, then you can just add them to `stored_in_c_pointer`

MenoIy: in this case, is it safe to use Variable class setter to change the is_pointer property?

EmilyBourne: I think in this case it is maybe ok

pyccel/codegen/printing/ccode.py
```diff
@@ -693,20 +688,18 @@ def _print_NumpyRandint(self, expr):
...
         self._additional_declare.clear()
 
-        if len(expr.results) > 1:
-            self._additional_args = [a.clone(name = self._parser.get_new_name('tmp_ret'), is_pointer =True) for a in expr.results]
-            body += '\n'.join(self._print_Assign(Assign(a, b)) for  a, b in zip(self._additional_args, expr.results))
-            body += '\n' + self._print_Return(Return([BooleanTrue()]))
-
         sep = self._print(SeparatorComment(40))
-
+        self._additional_args = []
```
EmilyBourne, line 702: This will not work if you have a function in a function

pyccel/codegen/printing/ccode.py
```diff
@@ -573,11 +571,8 @@ def function_signature(self, expr):
         if len(expr.results) == 1:
             ret_type = self.get_declare_type(expr.results[0])
         elif len(expr.results) > 1:
-            ret_type = self._print(datatype('bool')) + ' '
-            if not self._additional_args :
-                self._additional_args = [a.clone(name = a.name, is_pointer =True) for a in expr.results]
-            args += self._additional_args
-            self._additional_args.clear()
+            ret_type = self._print(datatype('int')) + ' '
```
EmilyBourne, line 574: Why did you change the return type back?

pyccel/codegen/printing/ccode.py
```diff
@@ -693,20 +688,18 @@ def _print_NumpyRandint(self, expr):
 
     def _print_FunctionDef(self, expr):
 
+        if len(expr.results) > 1:
+            self._additional_args = expr.results
```
EmilyBourne, line 692: This will not work in the case of a function in a function. I suggest using a list of lists so that this line becomes:
`self._additional_args.append(expr.results)`

pyccel/codegen/printing/ccode.py
MenoIy: I changed to int return 0 for no error because If a function with multiple results  always true I don't see were it can help it's like writing if True

pyccel/codegen/printing/ccode.py
EmilyBourne: I suggest initialising `self._additional_args` in `__init__`.
Then here you can do:
`self._additional_args.pop()`
This combined with the correction at line 692 should work

pyccel/codegen/printing/ccode.py
EmilyBourne: Ah ok, I didn't spot the return 0 :+1:

pyccel/codegen/printing/ccode.py
MenoIy: Just to make it clear what do you mean by function in a function ?

EmilyBourne: In python:
```
def f():
    def g():
        return 1, False
    if True:
        return g()
    else:
        return 2, True
```
In C:
```
int f(int* Out_1, bool* Out_2)
{
    int g(int* Out_3, bool* Out_4)
    {
        (*Out_3) = 1;
        (*Out_4) = 0;
        return 0;
    }
    if (1)
    {
        g(Out_1, Out_2);
        return 0;
    }
    else
    {
        (*Out_1) = 2;
        (*Out_2) = 1;
        return 0;
    }
}
```

pyccel/codegen/printing/ccode.py
```diff
@@ -689,7 +689,7 @@ def _print_NumpyRandint(self, expr):
     def _print_FunctionDef(self, expr):
 
         if len(expr.results) > 1:
-            self._additional_args = expr.results
+            self._additional_args.append(expr.results)
```
EmilyBourne, line 692: You need to initialise `self._additional_args` in the `__init__` method

MenoIy: it's already set to empty in __init__   line  227

EmilyBourne: Ah, my bad. Sorry

EmilyBourne:
I think you can mark this PR as ready for review again now

EmilyBourne (approved):

yguclu:
@MenoIy Is the PR's description up to date after the various changes you've made during the review process? And could you clarify in the description what is actually new compared to master?

MenoIy:
@yguclu  yes I already updated the description during the review change's.

yguclu (approved):

pyccel/codegen/printing/ccode.py
```diff
@@ -666,14 +688,19 @@ def _print_NumpyRandint(self, expr):
 
     def _print_FunctionDef(self, expr):
 
+        if len(expr.results) > 1:
+            self._additional_args.append(expr.results)
```
yguclu, line 692: I've just realized that `_additional_args` is a list of lists. Do you need this to handle nested functions?

EmilyBourne: Yes. A list was used, but I recommended using a list of lists for nested functions

yguclu (approved):

PR #520, opened 2020-10-29
Clean up code to reduce generated warnings

### LINKED ISSUE #489: Libraries should only be provided when the linker stage is carried out
**Describe the bug**
When compiling, libraries should only be provided when the linker stage is carried out. Therefore they should not be provided if the `-c` flag is provided. This is currently not respected and is leading to warnings with clang (see travis. When compiled with c there are warnings due to -lm)

EmilyBourne:
Reduce the number of warnings generated by the tests. Modify `compare_epyccel` function to avoid pyccelising the same function multiple times

- Raise an error for arrays as arguments in C
- Only provide libraries when linking (fixes #489)
- Provide language to compile_files to stop passing -J argument
- Only provide library directories when providing libraries
- Put argument for J with flag
- Mark C tests with missing c marker
- Modify compare_epyccel to be a class to avoid translating the same function multiple times
- Do not create function body if it will always raise an error to avoid unreachable-code warnings (functions as arguments)

### DIFF
--- a/tests/epyccel/test_higherorder_functions.py
+++ b/tests/epyccel/test_higherorder_functions.py
@@ -3,81 +3,83 @@ import pytest
 import modules.highorder_functions as mod
 from pyccel.epyccel import epyccel
 
-@pytest.mark.c
-def test_int_1():
-    modnew = epyccel(mod, language = "c", verbose = True)
+@pytest.fixture( params=[
+        pytest.param("fortran", marks = [
+            pytest.mark.fortran,
+            pytest.mark.skip]),
+        pytest.param("c", marks = [
+            pytest.mark.c]
+        )
+    ]
+)
+def language(request):
+    return request.param
+
+def test_int_1(language):
+    modnew = epyccel(mod, language = language)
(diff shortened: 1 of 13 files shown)


tests/epyccel/test_higherorder_functions.py
```diff
@@ -1,70 +1,81 @@
...
 import modules.highorder_functions as mod
 from pyccel.epyccel import epyccel
 
+@pytest.mark.c
 def test_int_1():
     modnew = epyccel(mod, language = "c", verbose = True)
 
     x_expected = mod.test_int_1()
     x = modnew.test_int_1()
     assert x == x_expected
```
yguclu, line 12: Given that all these tests could be run with Fortran in the near future, wouldn't it be better to have `language` as a parameter, and set it equal to `c` using a fixture? We could also have a `fortran` option with `pytest.mark.skip`.

yguclu (changes requested):

yguclu (approved):

Issue #491, opened 2020-10-21
C source code for Numpy arrays


saniQQ:
**Description**

As a DEVELOPER, in order to support Numpy arrays, I need an equivalent C data structure as part of the Pyccel internal library.

**Acceptance criteria**

- [x] Implement C data structure for multi-dimensional Numpy arrays (and array slices, which don't own memory)
- [x] Implement memory allocation and de-allocation
- [x] Implement `__setitem__` and `__getitem__` methods for scalar indices
- [x] Implement `__setitem__` method for Python `slice` input in the case of a scalar right-hand-side
- [x] Implement `__getitem__`method for Python `slice` inputs: create array slice
- [x] Make sure that indexing works on array slices
- [x] Store C source files somewhere in Pyccel's library
- [x] Write unit-tests in C

******

**implementing numpy array in C**

the arrays are represented by a one dimensional  array (continuous in the memory)
******

**the struct**
```C
typedef struct s_ndarray
{
    /* raw data buffer*/
    t_ndarray_type *data;
    /* number of dimmensions */
    int nd;
    /* shape 'size of each dimmension' */
    int *shape;
    /* strides 'number of bytes to skip to get the next element' */
    int *strides;
    /* type of the array elements */
    int type; // TODO : make it into an enum
    int lenght;
    int is_slice;
} t_ndarray;
```
**the union of the data buffer**
```C
typedef union s_ndarr_type
{
    /* data --- change it later to shorter names or use anonymous union*/
    char *raw_data;
    int *int_nd;
    float *float_nd;
    double *double_nd;

}  t_ndarray_type;

```
**the main functions prototype**
```C
t_ndarray *init_array(char *temp, int nd, int *shape, int type);
t_ndarray *make_slice(t_ndarray *p, ...);
int get_index(t_ndarray *arr, ...);
int free_array(t_ndarray *dump);
```

saniQQ:
still need to add a function that iterate the array and apply a function to it elements 

if we use a void* as the parameter we need to have a wrapper function (ex : a generic log function that then we can use it to cast to the proper type)
```int array_iterate(t_ndarray *arr, void (*f)(void *));```
or we can have multiple function each treat the corresponding type 
```int array_iterate_int(t_ndarray *arr, void (*f)(int));```
```int array_iterate_float(t_ndarray *arr, void (*f)(float));```

Issue #517, opened 2020-10-27
Fix codacy subprocess call with shell=True errors


EmilyBourne:
**Describe the bug**
Codacy has the following error:

> B602: Test for use of popen with shell equals true
> 
> Python possesses many mechanisms to invoke an external executable. However, doing so may present a security issue if appropriate care is not taken to sanitize any user provided or variable input.
> 
> This plugin test is part of a family of tests built to check for process spawning and warn appropriately. Specifically, this test looks for the spawning of a subprocess using a command shell. This type of subprocess invocation is dangerous as it is vulnerable to various shell injection attacks. Great care should be taken to sanitize all input in order to mitigate this risk. Calls of this type are identified by a parameter of ‘shell=True’ being given.
> 
> Additionally, this plugin scans the command string given and adjusts its reported severity based on how it is presented. If the command string is a simple static string containing no special shell characters, then the resulting issue has low severity. If the string is static, but contains shell formatting characters or wildcards, then the reported issue is medium. Finally, if the string is computed using Python’s string manipulation or formatting operations, then the reported issue has high severity. These severity levels reflect the likelihood that the code is vulnerable to injection.

**Acceptance Criteria**
- [x] Investigate how Popen handles things like "-I/home/user/code/folder with spaces/"
- [x] Store flags/libs/libdirs/etc in lists
- [x] Create cmd as a list of non-empty strings
- [x] Use verbose and join to print the expected command
- [x] Remove `shell=True` argument from subprocess calls

EmilyBourne:
This is important as the shell opened with `shell=True` may not match the shell which called python. In this case, it is harder to python to correctly preserve the environment variables. Notably, this probably prevents the use of the pytest action in our github windows tests

EmilyBourne:
This is necessary to allow #516 to find failures on PR #593 (marked as xfail for now)

EmilyBourne:
> Investigate how Popen handles things like "-I/home/user/code/folder with spaces/"

`Popen` does not need to quote these items as it assembles the `char* argc[]` object which is passed to the executable directly. Each element of the list is indexed in `argc`. This therefore achieves the same aim as the quotes but without requiring them

PR #522, opened 2020-10-30
ND-Arrays C code

saniQQ:
Implement N-dimensional arrays (Numpy-style) in C:

- Struct `t_ndarray` represents an N-dim. array: it contains all the information needed to access a flat memory storage with multiple indices and arbitrary strides, and it supports multiple datatypes (int, float, double, double complex);
- Struct `t_slice` represents a slice, i.e. a range of values (start, stop, step) for an index;
- Function `new_slice` creates a slice, given (start, stop, step);
- Function `array_create` creates an empty array (not initialised);
- Function `array_slicing` creates a sliced view of an N-dim. array, given the array and N slices;
- Function `get_index` handles indexing (computes index in multi-dimensional array);
- Add unit test for C code, which is run by Pytest using a custom conftest file.

Source and header files are located in pyccel/stdlib/ndarrays. This solves #491.


tests/ndarrays/conftest.py
```diff
@@ -0,0 +1,118 @@
...
+        # Run the exe that corresponds to the .c file and capture the output.
+        test_exe = os.path.splitext(str(self.fspath))[0]
+        rootdir = str(self.config.rootdir)
+        test_exe = os.path.relpath(test_exe)
+        if  sys.platform.startswith("win"):
+            ndarray_path =  rootdir + "\\pyccel\\stdlib\\ndarrays\\"
+        else :
+            ndarray_path = rootdir + "/pyccel/stdlib/ndarrays/"
+        comp_cmd = "gcc "+test_exe+".c "+ndarray_path+"*.c -I "+ndarray_path+" -o "+test_exe
+        subprocess.run(comp_cmd, check= 'TRUE' ,shell = 'TRUE')
```
EmilyBourne, line 49: When possible it is better to not use `shell = True` (see issue #517 ). To avoid this problem you need to pass a list instead of a string for the command. Something like:
```
comp_cmd = [shutil.which("gcc"), self.fspath, os.path.join(ndarray_path,"*.c"), "-I", ndarray_path, "-o", test_exe]
```

tests/ndarrays/conftest.py
```diff
@@ -0,0 +1,118 @@
...
+        """
+        # Run the exe that corresponds to the .c file and capture the output.
+        test_exe = os.path.splitext(str(self.fspath))[0]
+        rootdir = str(self.config.rootdir)
+        test_exe = os.path.relpath(test_exe)
+        if  sys.platform.startswith("win"):
+            ndarray_path =  rootdir + "\\pyccel\\stdlib\\ndarrays\\"
+        else :
+            ndarray_path = rootdir + "/pyccel/stdlib/ndarrays/"
+        comp_cmd = "gcc "+test_exe+".c "+ndarray_path+"*.c -I "+ndarray_path+" -o "+test_exe
```
EmilyBourne, line 48: I believe it is encouraged to use `shutil` to get the absolute path of the executable

tests/ndarrays/conftest.py
```diff
@@ -0,0 +1,118 @@
...
+        super().__init__(name, parent)
+        self.test_result = test_result
+
+    @classmethod
+    def from_parent(cls, name, parent, test_result):
+        return super().from_parent(name=name, parent=parent, test_result=test_result)
+
+    def runtest(self):
+        """The test has already been run. We just evaluate the result."""
+        # print(self.)
```
EmilyBourne, line 97: Can this comment be removed now?

tests/ndarrays/conftest.py
```diff
@@ -0,0 +1,118 @@
...
+        if self.test_result["condition"] == "FAIL":
+            raise CTestException(self, self.name)
+
+    def reportinfo(self):
+        """"Called to display header information about the test case."""
+        return self.fspath, self.test_result["line_number"] - 1 , self.name
```
EmilyBourne, line 103: Why do you need to do -1 on the line_number ?

tests/ndarrays/test_ndarrays.c
```diff
@@ -0,0 +1,278 @@
...
+#include <unistd.h>
+
+
+#define m_assert(X, Y, dscr) my_assert(X , Y, dscr, __func__, __FILE__, __LINE__)
+
+#define my_assert(X , Y, dscr, __func__, __FILE__, __LINE__) _Generic((X), double: assert_double,\
+							float: assert_float,\
+							int: assert_int,\
+							double complex : assert_complex_double,\
+							default: assert_ns)(X , Y, dscr, __func__, __FILE__, __LINE__)
```
EmilyBourne, line 11: Why do you need 2 macros? Is it not sufficient to do:
```
#define my_assert(X , Y, dscr) _Generic((X), double: assert_double,\
							float: assert_float,\
							int: assert_int,\
							double complex : assert_complex_double,\
							default: assert_ns)(X , Y, dscr, __func__, __FILE__, __LINE__)
```
?

tests/ndarrays/test_ndarrays.c
```diff
@@ -0,0 +1,278 @@
...
+#define getname(X) #X
+
+void assert_double(double v1 , double v2, const char *dscr, const char * func, const char *file, int line)
+{
+
+	if (v1 != v2)
+	{
+		printf("[FAIL] %s:%d:%s\n", file, line, func);
+		printf("[INFO] %s:%f != %s:%f\n", getname(v1), v1, getname(v2),v2);
```
EmilyBourne, line 20: Does `getname(v1)` just print `v1`? Why is the `getname` function helpful here?

tests/ndarrays/test_ndarrays.c
```diff
@@ -0,0 +1,278 @@
...
+
+	printf("[FAIL] %s:%d:%s\n", file, line, func);
+	printf("[INFO] not supported type\n");
+	printf("[DSCR] %s\n", dscr);
+	return ;
+}
+
+int	test_indexing_int(void)
+{
+	int m_1[] = {2, 3, 5, 5, 6, 7, 10, 11, 12, 260, 6, 8, 8, 0, 45, 0, 1, 0, 0, 1, 200, 33, 5, 57, 62, 70, 103, 141, 122, 26, 36, 82, 8, 10, 4115, 22, 1, 11, 1, 19};
```
EmilyBourne, line 77: This line does not fit on my screen. Please split it

tests/ndarrays/test_ndarrays.c
```diff
@@ -0,0 +1,278 @@
...
+	int m_1_shape[] = {5, 8};
+	t_ndarray *x;
+	int index;
+	int c_index;;
```
EmilyBourne, line 81: There are 2 semi-colons here

tests/ndarrays/test_ndarrays.c
```diff
@@ -0,0 +1,278 @@
...
+	int value;
+	int c_value;;
```
EmilyBourne, line 83: There are 2 semi-colons here

tests/ndarrays/test_ndarrays.c
```diff
@@ -0,0 +1,278 @@
...
+	value = x->data->int_nd[index];
+	c_value = 103;
+	m_assert(value , c_value, "testing the value");
+	free_array(x);
+	return (1);
+}
+
+int	test_indexing_double(void)
+{
+	double m_1[] = {2, 3, 5, 5, 6, 7, 10, 11, 12, 260, 6.34, 8, 8.002, 0.056, 45, 0.1, 1.02, 0.25, 0.00005, 1, 200, 33, 5, 57, 62, 70, 103.009, 141, 122, 26.50, 36.334, 82, 8.44002, 10.056, 4115, 22.1, 1.1102, 011.25, 1.01110005, 19};
```
EmilyBourne, line 101: This line doesn't fit on my screen either

tests/ndarrays/test_ndarrays.c
```diff
@@ -0,0 +1,278 @@
...
+	printf("[PASS] %s:%d:%s\n", file, line, func);
+	printf("[DSCR] %s\n", dscr);
+}
+void assert_ns(float v1 , float v2, const char *dscr, const char * func, const char *file, int line)
+{
+
+	printf("[FAIL] %s:%d:%s\n", file, line, func);
+	printf("[INFO] not supported type\n");
+	printf("[DSCR] %s\n", dscr);
+	return ;
```
EmilyBourne, line 72: This line does nothing

tests/ndarrays/test_ndarrays.c
```diff
@@ -0,0 +1,278 @@
...
+	index = 3 * x->strides[0] + 2 * x->strides[1];
+	c_index = 26;
+	m_assert(index , c_index, "testing the strides"); //testing the strides
+	m_assert(get_index(x, 3, 2) , c_index, "testing the indexing function"); //testing the indexing function
+	// testing the value with the index [0, 0]
+	value = x->data->int_nd[index];
+	c_value = 103;
+	m_assert(value , c_value, "testing the value");
+	free_array(x);
+	return (1);
```
EmilyBourne, line 96: Why return 1? Is this to signify true? If so why not use a boolean? Or to signify no problems? Why not 0, the standard error code that you have used in the main?
Although I am not sure that it makes sense to return anything. This doesn't even indicate that the assertions were successful

tests/ndarrays/test_ndarrays.c
```diff
@@ -0,0 +1,278 @@
...
+	value = x->data->double_nd[index];
+	c_value = 103.009;
+	m_assert(value , c_value, "testing the value");
+	free_array(x);
+	return (1);
+}
+
+int	test_indexing_complex_double(void)
+{
+	double complex m_1[] = {0.37 + 0.588*I,  0.92689451+0.57106791*I, 0.93598206+0.30289964*I,  0.54404246+0.09516331*I, 0.02827254+0.00432899*I,  0.06873651+0.24810741*I, 0.94040543+0.43508215*I,  0.58532094+0.67890618*I, 0.68742283+0.64951155*I,  0.15372315+0.89699101*I};
```
EmilyBourne, line 125: This line doesn't fit on my screen

tests/ndarrays/test_ndarrays.c
```diff
@@ -0,0 +1,278 @@
...
+	int m_1_shape[] = {5, 2};
+	t_ndarray *x;
+	int index;
+	int c_index;;
+	double complex value;
+	double complex c_value;;
```
EmilyBourne, line 131: There are two semi-colons

tests/ndarrays/test_ndarrays.c
```diff
@@ -0,0 +1,278 @@
...
+	return (1);
+}
+
+int	test_indexing_complex_double(void)
+{
+	double complex m_1[] = {0.37 + 0.588*I,  0.92689451+0.57106791*I, 0.93598206+0.30289964*I,  0.54404246+0.09516331*I, 0.02827254+0.00432899*I,  0.06873651+0.24810741*I, 0.94040543+0.43508215*I,  0.58532094+0.67890618*I, 0.68742283+0.64951155*I,  0.15372315+0.89699101*I};
+	int m_1_shape[] = {5, 2};
+	t_ndarray *x;
+	int index;
+	int c_index;;
```
EmilyBourne, line 129: There are two semi-colons

tests/ndarrays/test_ndarrays.c
```diff
@@ -0,0 +1,278 @@
...
+*/
+
+int	test_slicing_int(void)
+{
+	int m_1[] = {2, 3, 5, 5, 6, 7, 10, 11, 12, 260, 6, 8, 8, 0, 45, 0, 1, 0, 0, 1, 200, 33, 5, 57, 62, 70, 103, 141, 122, 26, 36, 82, 8, 10, 4115, 22, 1, 11, 1, 19};
+	int m_1_shape[] = {8, 5};
+	t_ndarray *x;
+	t_ndarray *slice;
+	int index;
+	int c_index;;
```
EmilyBourne, line 159: There are two semi-colons

tests/ndarrays/test_ndarrays.c
```diff
@@ -0,0 +1,278 @@
...
+	int value;
+	int c_value;;
```
EmilyBourne, line 161: There are two semi-colons

tests/ndarrays/test_ndarrays.c
```diff
@@ -0,0 +1,278 @@
...
+}
+
+int	test_slicing_double(void)
+{
+	double m_1[] = {2, 3, 5, 5, 6, 7, 10, 11, 12, 260, 6.34, 8, 8.002, 0.056, 45, 0.1, 1.02, 0.25, 0.00005, 1, 200, 33, 5, 57, 62, 70, 103.009, 141, 122, 26.50, 36.334, 82, 8.44002, 10.056, 4115, 22.1, 1.1102, 011.25, 1.01110005, 19};
+	int m_1_shape[] = {8, 5};
+	t_ndarray *x;
+	t_ndarray *slice;
+	int index;
+	int c_index;;
```
EmilyBourne, line 196: There are two semi-colons

tests/ndarrays/test_ndarrays.c
```diff
@@ -0,0 +1,278 @@
...
+	double value;
+	double c_value;;
```
EmilyBourne, line 198: There are two semi-colons

tests/ndarrays/test_ndarrays.c
```diff
@@ -0,0 +1,278 @@
...
+
+int	test_slicing_complex_double(void)
+{
+	double complex m_1[] = {0.37 + 0.588*I,  0.92689451+0.57106791*I, 0.93598206+0.30289964*I,  0.54404246+0.09516331*I, 0.02827254+0.00432899*I,
+			  0.06873651+0.24810741*I, 0.94040543+0.43508215*I,  0.58532094+0.67890618*I, 0.68742283+0.64951155*I,  0.15372315+0.89699101*I};
+	int m_1_shape[] = {2, 5};
+	t_ndarray *x;
+	t_ndarray *slice;
+	int index;
+	int c_index;;
```
EmilyBourne, line 234: There are two semi-colons

tests/ndarrays/test_ndarrays.c
```diff
@@ -0,0 +1,278 @@
...
+	double complex value;
+	double complex c_value;;
```
EmilyBourne, line 236: There are two semi-colons

tests/ndarrays/test_ndarrays.c
```diff
@@ -0,0 +1,278 @@
...
+	return (1);
+}
+
+/* 
+**	slicing tests
+*/
+
+int	test_slicing_int(void)
+{
+	int m_1[] = {2, 3, 5, 5, 6, 7, 10, 11, 12, 260, 6, 8, 8, 0, 45, 0, 1, 0, 0, 1, 200, 33, 5, 57, 62, 70, 103, 141, 122, 26, 36, 82, 8, 10, 4115, 22, 1, 11, 1, 19};
```
EmilyBourne, line 154: This line doesn't fit on my screen

tests/ndarrays/test_ndarrays.c
```diff
@@ -0,0 +1,278 @@
...
+	value = x->data->int_nd[get_index(x, 1, 2)];
+	m_assert(value , c_value, "testing slice assignment");
+	free_array(x);
+	free_array(slice);
+	return (1);
+}
+
+int	test_slicing_double(void)
+{
+	double m_1[] = {2, 3, 5, 5, 6, 7, 10, 11, 12, 260, 6.34, 8, 8.002, 0.056, 45, 0.1, 1.02, 0.25, 0.00005, 1, 200, 33, 5, 57, 62, 70, 103.009, 141, 122, 26.50, 36.334, 82, 8.44002, 10.056, 4115, 22.1, 1.1102, 011.25, 1.01110005, 19};
```
EmilyBourne, line 191: This line doesn't fit on my screen

tests/ndarrays/test_ndarrays.c
```diff
@@ -0,0 +1,278 @@
...
+	c_index = 5;
+	for (int i = 0; i < slice->shape[0]; i++)
+	{
+		for (int j = 0; j < slice->shape[1]; j++)
+		{
+			value = slice->data->int_nd[get_index(slice, i, j)];
+			c_value = m_1[c_index];
+			c_index+=2;
+            if (value != c_value)
+			    m_assert(value , c_value, "testing slice values");
```
EmilyBourne, line 176: Strange indenting. Are you mixing tabs and spaces?

tests/ndarrays/test_ndarrays.c
```diff
@@ -0,0 +1,278 @@
...
+	test_indexing_double();
+	test_indexing_int();
+	test_indexing_complex_double();
+
+	/* slicing tests */
+	test_slicing_double();
+	test_slicing_int();
+	test_slicing_complex_double();
+
+	write(1, "ALL the tests PASSed.\n", 22);
```
EmilyBourne, line 276: Is this printed even if a test fails?

pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -0,0 +1,124 @@
...
+	}
+	a->strides = malloc(nd * sizeof(int));
+	for (int i = 0; i < a->nd; i++)
+	{
+		a->strides[i] = 1;
+		for (int j = i + 1; j < a->nd; j++)
+			a->strides[i] *= a->shape[j];
+	}
+	a->data = malloc(sizeof(t_ndarray_type));
+	a->data->raw_data = calloc(a->lenght , a->type);
```
EmilyBourne, line 30: @yguclu Do you think we should be using `calloc` or `malloc` here? I would expect `malloc` to be used for `np.empty` and `calloc` to be used for `np.zeros` as per the numpy description of `empty`:
> `empty`, unlike `zeros`, does not set the array values to zero,
> and may therefore be marginally faster.  On the other hand, it requires
> the user to manually set all the values in the array, and should be
> used with caution.

pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -0,0 +1,124 @@
...
+	if (temp)
+		memcpy(a->data->raw_data, temp, a->lenght * a->type);
```
EmilyBourne, line 32: Have you considered how to manage `np.full`?

pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -0,0 +1,124 @@
...
+	return (a);
+}
+
+/* 
+** deallocation
+*/
+
+int free_array(t_ndarray *dump)
```
EmilyBourne, line 40: If everything is going to be a pointer then I wonder if it would be better to pass a pointer to a pointer. That way you can set the pointer to NULL.
Ie:
```
int free_array(t_ndarray **dump)
{
	if (!(*dump)->is_slice)
	{
		free((*dump)->data->raw_data);
		(*dump)->data->raw_data = NULL;
	}
	free((*dump)->data);
	(*dump)->data = NULL;
	free((*dump)->shape);
	(*dump)->shape = NULL;
	free((*dump)->strides);
	(*dump)->strides = NULL;
	free(*dump);
	*dump = NULL;
	return (1);
}
```

pyccel/stdlib/ndarrays/ndarrays.h
```diff
@@ -0,0 +1,66 @@
...
+/* allocations */
+t_ndarray *init_array(char *temp, int nd, int *shape, int type);
+
+/* dumping data */
+// int array_value_dump(t_ndarray *arr);
+// int array_data_dump(t_ndarray *arr);
+
+/* slicing */
+t_slice *slice_data(int start, int end, int step);
+t_ndarray *make_slice(t_ndarray *p, ...);
```
EmilyBourne, line 55: Why does `make_slice` need to return a pointer?

pyccel/stdlib/ndarrays/ndarrays.h
```diff
@@ -0,0 +1,66 @@
...
+	/* number of dimmensions */
+	int	nd;
+	/* shape 'size of each dimmension' */
+	int *shape;
+	/* strides 'number of bytes to skip to get the next element' */
+	int *strides;
+	/* type of the array elements */
+	int type; // TODO : make it into an enum
+	int lenght;
+	int is_slice;
```
EmilyBourne, line 41: Why not use a bool?

pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -0,0 +1,124 @@
...
+	slice->type = p->type;
+	slice->shape = malloc(sizeof(int) * p->nd);
+	memcpy(slice->shape, p->shape, sizeof(int) * p->nd);
+	slice->strides = malloc(sizeof(int) * p->nd);
+	memcpy(slice->strides, p->strides, sizeof(int) * p->nd);
+	slice->is_slice = 1;
+	va_start(va, p);
+	while ((slice_data = va_arg(va, t_slice*)))
+	{
+		slice->shape[i] = (slice_data->end - slice_data->start + (slice_data->step / 2)) / slice_data->step;
```
EmilyBourne, line 92: Why does step need dividing by 2?

EmilyBourne: Is this to round up?

pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -0,0 +1,124 @@
...
+		start += slice_data->start * p->strides[i];
+		slice->strides[i] *= slice_data->step;
+		i++;
+		free(slice_data);
```
EmilyBourne, line 96: What happens if the user wants to use the same slice date multiple times rather than creating a temporary?

pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -0,0 +1,124 @@
...
+int get_index(t_ndarray *arr, ...)
+{
+	va_list va;
+	int index;
+	va_start(va, arr);
+
+	index = 0;
+	for (int i = 0; i < arr->nd; i++)
+	{
+			index += va_arg(va, int) * arr->strides[i];
```
EmilyBourne, line 120: strange indentation

pyccel/stdlib/ndarrays/ndarrays.h
```diff
@@ -0,0 +1,66 @@
...
+typedef struct s_slice_data{
+	int start;
+	int end;
+	int step;
+}	t_slice;
+
+typedef struct s_ndarray
+{
+	/* raw data buffer*/
+	t_ndarray_type *data;
```
EmilyBourne, line 31: Why does this need to be a pointer? Can we have a `t_ndarray` with `data==NULL` or where `data` contains multiple objects of type `t_ndarray_type`?

EmilyBourne (changes requested):

pyccel/stdlib/ndarrays/ndarrays.h
```diff
@@ -0,0 +1,66 @@
...
+# include <stdio.h>
+# include <stdarg.h>
+
+typedef union	u_ndarr_type
+{
+	/*
+	*** data --- change it later to shorter names or use anonymous array
+	*/
+	char	*raw_data;
+	int		*int_nd;
```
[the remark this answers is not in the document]
saniQQ: vscode changed the tabsize and i did not notice

pyccel/stdlib/ndarrays/ndarrays.h
```diff
@@ -0,0 +1,66 @@
...
+	float	*float_nd;
+	double	*double_nd;
+    double complex *complex_double;
```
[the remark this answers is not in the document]
saniQQ: fixed

pyccel/stdlib/ndarrays/ndarrays.h
```diff
@@ -0,0 +1,66 @@
...
+	t_ndarray_type *data;
+	/* number of dimmensions */
+	int	nd;
+	/* shape 'size of each dimmension' */
+	int *shape;
+	/* strides 'number of bytes to skip to get the next element' */
+	int *strides;
+	/* type of the array elements */
+	int type; // TODO : make it into an enum
+	int lenght;
```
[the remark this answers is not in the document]
saniQQ: fixed

pyccel/stdlib/ndarrays/ndarrays.h
```diff
@@ -0,0 +1,66 @@
...
+/* dumping data */
+// int array_value_dump(t_ndarray *arr);
+// int array_data_dump(t_ndarray *arr);
+
+/* slicing */
+t_slice *slice_data(int start, int end, int step);
+t_ndarray *make_slice(t_ndarray *p, ...);
+
+/* free */
+int free_array(t_ndarray *dump);
```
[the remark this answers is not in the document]
saniQQ: the free_array handles the slice case

pyccel/stdlib/ndarrays/ndarrays.h
```diff
@@ -0,0 +1,66 @@
...
+
+/* indexing */
+int get_index(t_ndarray *arr, ...);
+
+/* other funcs */
+t_ndarray *mat_product(t_ndarray *mat1, t_ndarray *mat2);
```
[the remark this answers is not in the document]
saniQQ: yeah it should be in a separate header and file

pyccel/stdlib/ndarrays/ndarrays.h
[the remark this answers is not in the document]
EmilyBourne: I mean to delete a `t_slice*` object but it looks like this doesn't contain pointers so it is not needed :+1:

tests/ndarrays/conftest.py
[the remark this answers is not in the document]
saniQQ: i just used put the relative path of self.fspath in test_exe so i don't have to deal with the C:\ in windows it needed to be parsed and changed to /c/ or C:\\  if its not clear i can change it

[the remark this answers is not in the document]
EmilyBourne: I'm not sure I understand. You receive an absolute path to a X.c file. But the path needs to be changed so `C:\` is replaced with `C:\`?
Or you need to add the `C:\`? There are lots of functions in the `os.path` module which should help you with this kind of problem (e.g. os.path.abspath)

tests/ndarrays/conftest.py
```diff
@@ -0,0 +1,118 @@
...
+
+        """
+        # Run the exe that corresponds to the .c file and capture the output.
+        test_exe = os.path.splitext(str(self.fspath))[0]
+        rootdir = str(self.config.rootdir)
+        test_exe = os.path.relpath(test_exe)
+        if  sys.platform.startswith("win"):
+            ndarray_path =  rootdir + "\\pyccel\\stdlib\\ndarrays\\"
+        else :
+            ndarray_path = rootdir + "/pyccel/stdlib/ndarrays/"
```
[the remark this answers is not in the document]
saniQQ: os.path.join does not add double backslash in windows and when the shell read the command he treat it as a backslash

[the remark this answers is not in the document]
EmilyBourne: Are you sure? What error does this raise? Remember that `print("\\")` will give `\` in python
Do you want to chat about os.path?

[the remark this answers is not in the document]
saniQQ: this is fixed when not using the shell = true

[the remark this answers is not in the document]
EmilyBourne: Ah perfect :+1:

tests/ndarrays/test_ndarrays.c
saniQQ: i just wanted to have the variable name and the value next to each other e.g:  `INFO : get_index(x, 3, 1):25 != c_index:26`
i did fixed it to work how i wanted

tests/ndarrays/test_ndarrays.c
saniQQ: fixed

tests/ndarrays/test_ndarrays.c
saniQQ: fixed

tests/ndarrays/test_ndarrays.c
saniQQ: i will remove it && and will use 0 in the future to indicate true

EmilyBourne: 0 to indicate no errors ^^
0 to indicate true is just confusing

saniQQ: ah okey make sens

tests/ndarrays/test_ndarrays.c
saniQQ: i removed it the test info are good alone

pyccel/stdlib/ndarrays/ndarrays.c
saniQQ: yeah i will split init_array to have an array_empty function

pyccel/stdlib/ndarrays/ndarrays.c
saniQQ: i will a new array_empty then init the data

pyccel/stdlib/ndarrays/ndarrays.h
saniQQ: will resolve when fix the pointers for everything

pyccel/stdlib/ndarrays/ndarrays.c
saniQQ: to make this work i changed the make_slice function to get just a struct not a pointer >>> we will need to pass the exact slice as the number of dimension because i can't check if the pointer is NULL

pyccel/stdlib/ndarrays/ndarrays.h
saniQQ: fixed

pyccel/stdlib/ndarrays/ndarrays.h
saniQQ: i did change it to an anonymous union

pyccel/stdlib/ndarrays/ndarrays.c
saniQQ: i changed it so that the function slice_data return a struct not a pointer 
now we need to give the make_slice function the slices_data for all dimensions

pyccel/stdlib/ndarrays/ndarrays.h
```diff
@@ -0,0 +1,66 @@
...
+	/* raw data buffer*/
+	t_ndarray_type *data;
+	/* number of dimmensions */
+	int	nd;
+	/* shape 'size of each dimmension' */
+	int *shape;
+	/* strides 'number of bytes to skip to get the next element' */
+	int *strides;
+	/* type of the array elements */
+	int type; // TODO : make it into an enum
```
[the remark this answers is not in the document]
saniQQ: fixed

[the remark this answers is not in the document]
saniQQ: and added type_size

pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -0,0 +1,118 @@
...
+        start += slice_data.start * p.strides[i];
+        slice.strides[i] *= slice_data.step;
+        i++;
+    }
+    va_end(va);
+    slice.raw_data = p.raw_data + start * p.type_size;
+    slice.length = 1;
+    for (int i = 0; i < slice.nd; i++)
+            slice.length *= slice.shape[i];
+    return (slice);
```
EmilyBourne, line 98: I disagree with codacy here. As we have a free function which frees shape there shouldn't be a problem. I suspect this is due to using a C++ static code analysis for c code

pyccel/stdlib/ndarrays/ndarrays.h
```diff
@@ -0,0 +1,66 @@
...
+    int end;
+    int step;
+}               t_slice;
+
+enum e_types{
+        nd_int,
+        nd_float,
+        nd_double,
+        nd_complex_double
+            };
```
EmilyBourne, line 22: Why is the closing brace indented?

pyccel/stdlib/ndarrays/ndarrays.h
```diff
@@ -0,0 +1,66 @@
...
+    /* number of dimmensions */
+    int             nd;
+    /* shape 'size of each dimmension' */
+    int             *shape;
+    /* strides 'number of bytes to skip to get the next element' */
+    int             *strides;
+    /* type of the array elements */
+    enum e_types    type; // TODO : make it into an enum
+    int             type_size; // TODO : make it into an enum
```
EmilyBourne, line 42: These TODOs can be removed

pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -0,0 +1,118 @@
+#include "ndarrays.h"
+
+/* 
+** allocation
+*/
+
+t_ndarray array_init(char *temp, int nd, int *shape, enum e_types type, int type_size)
```
EmilyBourne, line 7: Why do we need to pass the type_size? Can `array_init` not define this using the type?

EmilyBourne (changes requested):

pyccel/stdlib/ndarrays/ndarrays.c
saniQQ: yeah and i tested it and there is no leaks if we call the free function

pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -0,0 +1,118 @@
...
+    int i = 0;
+    int start = 0;
+
+    slice.nd = p.nd;
+    slice.type = p.type;
+    slice.type_size = p.type_size;
+    slice.shape = malloc(sizeof(int) * p.nd);
+    memcpy(slice.shape, p.shape, sizeof(int) * p.nd);
+    slice.strides = malloc(sizeof(int) * p.nd);
+    memcpy(slice.strides, p.strides, sizeof(int) * p.nd);
```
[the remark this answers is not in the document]
saniQQ: for the strides it is still needed because the slice strides depend on the base array so if i removed it i will just need to recalculate them with a for loop and i think using memcpy here is better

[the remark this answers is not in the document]
EmilyBourne: Ah you're right sorry I misread the `*=`

tests/ndarrays/conftest.py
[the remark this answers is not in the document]
EmilyBourne: Therefore here I suggest using:
```
ndarray_path = os.path.join(rootdir, "pyccel", "stdlib", "ndarrays")
```

pyccel/stdlib/ndarrays/ndarrays.c
EmilyBourne: I don't want to mark this conversation as resolved as it is important that reviewers can see it. In addition codacy errors prevent us from merging branches so we need to decide whether to disable this error here, in general, or if we need a different static analysis tool

tests/ndarrays/conftest.py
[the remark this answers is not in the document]
EmilyBourne: If this is fixed I am happy to approve this PR

EmilyBourne:
Can this PR be marked as ready for review?

EmilyBourne (approved):

pyccel/stdlib/ndarrays/ndarrays.h
```diff
@@ -0,0 +1,68 @@
...
+/* allocations */
+t_ndarray   array_create(int nd, int *shape, enum e_types type);
+t_ndarray   array_init(char *temp, int nd, int *shape, enum e_types type);
+t_ndarray   array_ones(int nd, int *shape, enum e_types type);
+t_ndarray   array_zeros(int nd, int *shape, enum e_types type);
+// t_ndarray   array_full(void *c, int nd, int *shape, enum e_types type);
+
+/* slicing */
+t_slice     slice_data(int start, int end, int step);
+t_ndarray   slice_make(t_ndarray p, ...);
```
yguclu, line 60: I make this comment here, but it applies to multiple places in the code. As I mentioned at one of our meetings, there is some confusion with the use of the word "slice". In order to be consistent with Python, we should call **slice** only the `t_slice` object, and maybe the function that creates it. Hence:
```c
t_slice  slice(int start, int end, int step);  // Also OK to call it "new_slice" or "create_slice"
```
Regarding the action of **slicing** an array, it results in an array **view**. So I suggest we write
```c
t_ndarray  array_slicing(t_ndarray p, ...); // Also OK to call it "sliced_array"
```
Suggestions are welcome!

pyccel/stdlib/ndarrays/ndarrays.h
```diff
@@ -0,0 +1,68 @@
...
+    bool            is_slice;
+}               t_ndarray;
+
+/* functions prototypes */
+
+/* allocations */
+t_ndarray   array_create(int nd, int *shape, enum e_types type);
+t_ndarray   array_init(char *temp, int nd, int *shape, enum e_types type);
+t_ndarray   array_ones(int nd, int *shape, enum e_types type);
+t_ndarray   array_zeros(int nd, int *shape, enum e_types type);
```
yguclu, line 55: Given that we are in the process of adding an `Allocate` node to the Pyccel AST, it seems preferable to have `array_init`, `array_ones` and `array_zeros` only initialize the array, without allocating it.

@saidctb @EmilyBourne What do you think?

tests/ndarrays/test_ndarrays.c
```diff
@@ -0,0 +1,446 @@
...
+    /* array_ones tests */
+    test_array_ones_int();
+    test_array_ones_double();
+    test_array_ones_cdouble();
+    /* array_zeros tests */
+    test_array_zeros_int();
+    test_array_zeros_double();
+    test_array_zeros_cdouble();
+    return (0);
+}
```
yguclu, line 446: Missing newline at end of file

yguclu (changes requested):

pyccel/stdlib/ndarrays/ndarrays.h
EmilyBourne: I'm not sure I agree. While the addition of the `Allocate` node will make it very easy to print:
```
x = array_create(...)
array_init(x, ...)
```
we usually try to have code that resembles the python code, unless the changes make the result more natural in the target language. I'm not sure that this is the case here.

If we did do this though, then the initialisation is potentially simple enough to not require a function.
I would probably agree that:
```
x = array_create(...)
memcpy(arr.raw_data, temp, arr.buffer_size);
```
looks more like standard c code.

pyccel/stdlib/ndarrays/ndarrays.h
saniQQ: i will go with new_slice 
and array_slicing "to match the other functions"

pyccel/stdlib/ndarrays/ndarrays.h
yguclu: Now that I think about it, it makes a lot of sense to leave the initialization to the C printer... In many practical cases we will need some non-trivial initialization anyway. If we take this path, the simple initialization functions provided here would be only used in the unit tests.

saidctb: we need an empty which is for the array creation, and a function for filling the array with a value to use it for `np.full`, and all the other will use a combination of the two

EmilyBourne: `empty` is equivalent to `array_create` but I agree that we should use the same naming convention

`array_ones` can use the full method, but I think we should keep `array_zeros` as you can allocate 0 much faster with memset than with a loop

saniQQ: for np.full we either need an iterator function to use it on the sliced arrays or we just use the C printer to create the needed loops (from the C printer we can manage the different types without using the _generic macro)

EmilyBourne: `full` creates an array so you don't need to worry about slices. That means you can just iterate over the flattened array

saniQQ: is it better to write it in the C printer or create a generic function for it which mean we need a full function for each type the call them with the _generic macro

yguclu: After PR #521 the nodes `NumpyFull`, `NumpyEmpty`, `NumpyZeros`, `NumpyOnes`, etc. will not take care of memory allocation any more, because an `Allocate` node (which we can rename however we want) will be created at the semantic stage. So I hardly see the need for more than the C functions for allocation and deallocation, and maybe a few utility functions later on to help with the loops.

This discussion is becoming a bit confusing to me. Are you guys reconsidering PR #521 altogether?

saidctb: we wanted to make it compatible with the PR #521 which means we need to separate between array allocation and initialization , in here we have `array_ones` and `array_zeros` instead we can just add a`array_full` which initializes the array to whatever value you give it  with  the exception of array_zeros but the rest should use `array_full`, so then the allocation node in PR #521 will use `array_create`

yguclu: OK, then I suggest that we change the names as follows: `array_create` creates a new multi-dimensional array, while `array_fill` just fills it with a scalar value.
Any further optimization can be done at the printing stage.

pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -45,49 +45,28 @@ t_ndarray   array_create(int nd, int *shape, enum e_types type)
...
 {
-    t_ndarray arr;
-
-    arr = array_create(nd, shape, type);
-    bzero(arr.raw_data, arr.buffer_size);
-    return (arr);
+    if (c == 0)
+        memset(arr.raw_data, 0, arr.buffer_size);
+    for (int i = 0; i < arr.length; i++)
+        arr.nd_cdouble[i] = c;
```
yguclu, line 69: Same comment

tests/ndarrays/test_ndarrays.c
```diff
@@ -184,8 +187,9 @@ int test_slicing_int(void)
     int value;
     int c_value;
 
-    x = array_init((char *)m_1, 2, m_1_shape, nd_int);
-    slice = slice_make(x, slice_data(1, 2, 1), slice_data(0, 5, 2));
+    x = array_create(2, m_1_shape, nd_int);
+    memcpy(x.raw_data, m_1, x.buffer_size);
+    slice = array_slicing(x, new_slice(1, 2, 1), new_slice(0, 5, 2));
```
yguclu, line 192: Can we use a different name instead of `slice`? What about `xview`, or simply `y`?

tests/ndarrays/test_ndarrays.c
```diff
@@ -223,8 +227,10 @@ int test_slicing_double(void)
     int c_index;
     double value;
     double c_value;
-    x = array_init((char *)m_1, 2, m_1_shape, nd_double);
-    slice = slice_make(x, slice_data(1, 2, 1), slice_data(0, 5, 2));
+
+    x = array_create(2, m_1_shape, nd_double);
+    memcpy(x.raw_data, m_1, x.buffer_size);
+    slice = array_slicing(x, new_slice(1, 2, 1), new_slice(0, 5, 2));
```
yguclu, line 233: Same comment as above

tests/ndarrays/test_ndarrays.c
```diff
@@ -259,8 +265,9 @@ int test_slicing_cdouble(void)
     double complex value;
     double complex c_value;
 
-    x = array_init((char *)m_1, 2, m_1_shape, nd_cdouble);
-    slice = slice_make(x, slice_data(1, 2, 1), slice_data(0, 5, 2));
+    x = array_create(2, m_1_shape, nd_cdouble);
+    memcpy(x.raw_data, m_1, x.buffer_size);
+    slice = array_slicing(x, new_slice(1, 2, 1), new_slice(0, 5, 2));
```
yguclu, line 270: Same comment

yguclu (changes requested):

tests/ndarrays/test_ndarrays.c
saniQQ: will use the `xview`

tests/ndarrays/test_ndarrays.c
saniQQ: will use the `xview`

tests/ndarrays/test_ndarrays.c
saniQQ: will use the `xview`

pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -45,49 +45,28 @@ t_ndarray   array_create(int nd, int *shape, enum e_types type)
...
 {
-    t_ndarray arr;
-
-    arr = array_create(nd, shape, type);
-    memcpy(arr.raw_data, temp, arr.buffer_size);
-    return (arr);
+    if (c == 0)
+        memset(arr.raw_data, 0, arr.buffer_size);
+    for (int i = 0; i < arr.length; i++)
+        arr.nd_int[i] = c;
```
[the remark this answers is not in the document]
saniQQ: fixed

yguclu (approved):

saidctb (approved):

EmilyBourne (approved):

PR #523, opened 2020-10-30
Use pythonic syntax in OpenMP pragmas

mostafaMamoni:
Improve OpenMP pragmas to use Python's keywords (e.g. `#$ omp for` instead of `#$ omp do`) where appropriate:

- Modify grammar of OpenMP pragmas;
- Write unit tests that work for both C and Fortran.


pyccel/codegen/printing/fcode.py
```diff
@@ -1690,6 +1690,7 @@ def _print_OMP_Parallel_Construct(self, expr):
 
     def _print_Omp_End_Clause(self, expr):
         omp_expr = str(expr.txt)
+        omp_expr = omp_expr.replace("for", "do")
```
EmilyBourne, line 1693: Can `omp_expr` contain any user defined words? E.g. the names of variables? If so what happens if the variable is called "my_for_var"?

nhamidn: No, we can't have variables in the `END Clause`, variables must be declared at the start of the` loop construct `

EmilyBourne:
This looks ok to me. It looks like you just need to update the tests in internal/ to use your new omp syntax

EmilyBourne (approved):

rakati (approved):

nhamidn (approved):

yguclu:
This PR looks good to me! But changes to our pragma syntax will break the user code, so we have to think twice before merging them into the master branch.

In this case I think it is best if we keep this PR on hold while we completely specify our own grammar for Pyccel OpenMP pragmas. Once we have agreed on that, we should create a file with some minimal documentation and then include any required changes into this PR (for example, removing `#$ omp end for` at the end of a for loop).

nhamidn:
> This PR looks good to me! But changes to our pragma syntax will break the user code, so we have to think twice before merging them into the master branch.
> 
> In this case I think it is best if we keep this PR on hold while we completely specify our own grammar for Pyccel OpenMP pragmas. Once we have agreed on that, we should create a file with some minimal documentation and then include any required changes into this PR (for example, removing `#$ omp end for` at the end of a for loop).

i agree.

nhamidn:
i think this PR can be merged or closed, i have merged the change to my branch so im using the new syntax.

yguclu (approved):

PR #524, opened 2020-11-02
Properly separate Python3 built-in functions from Numpy and Itertools (part-2)

### LINKED ISSUE #258: Properly separate Python3 built-in functions from Numpy's and Math's functions

### LINKED ISSUE #514: Remove large dictionaries from ast.utilities
**Description**

As a DEVELOPER, I want a code base that is easy to maintain. To this end, each Python module should only import a minimum number of objects from other modules.

Currently `ast.utilities` needs to import lots of objects just to populate some dictionaries. Instead it should import the dictionaries from the appropriate files (e.g. `numpy_functions` from `ast.numpy_ext`).

**Acceptance Criteria**
- [x] Move dictionaries to the appropriate files
- [x] Have module `ast.utilities` import dictionaries instead of creating them
- [x] Remove many useless import statements from `ast.utilities`

rakati:
-  Move dictionaries from `ast.utilities` to the appropriate files.
-  Have module `ast.utilities` import dictionaries instead of creating them.
-  Remove many useless import statements from `ast.utilities`.
-  Create new classes in `ast.builtins` and remove linking to `numpyext `classes.
-  Move itertools functions to `ast.itertoolsext` and change imports of these functions from `ast.core` to `ast.itertoolsext`.

closes #258 
closes #514


pyccel/ast/itertoolsext.py
```diff
@@ -0,0 +1,21 @@
+
+from sympy import Tuple
+from .basic     import Basic
+
+__all__ = (
+    'Product',
+)
+
+class Product(Basic):
+    """Represents a Product stmt."""
```
EmilyBourne, line 10: This comment could be clearer. It sounds like this represents a product statements (i.e. `a*b`) but I don't think this is the case right?

pyccel/ast/numpyext.py
```diff
@@ -1320,3 +1319,80 @@ class NumpyInt64(NumpyInt):
...
+    'empty'     : NumpyEmpty,
+    'zeros'     : NumpyZeros,
+    'ones'      : NumpyOnes,
+    'full_like' : NumpyFullLike,
+    'empty_like': NumpyEmptyLike,
+    'zeros_like': NumpyZerosLike,
+    'ones_like' : NumpyOnesLike,
+    'array'     : NumpyArray,
+    # ...
+    'shape'     : Shape,
```
EmilyBourne, line 1338: Was Shape missed in the last PR?

tests/pyccel/test_pyccel.py
```diff
@@ -253,267 +253,266 @@ def teardown(path_dir = None):
...
+
+# #------------------------------------------------------------------------------
+# def test_import_collisions():
+#     pyccel_test("scripts/import_syntax/collisions4.py",
+#             dependencies = ["scripts/import_syntax/user_mod.py", "scripts/import_syntax/user_mod2.py"])
+
+# #------------------------------------------------------------------------------
+# def test_numpy_kernels_compile():
+#     cwd = get_abs_path(".")
+#     compile_pyccel(os.path.join(cwd, "scripts/numpy/"), "numpy_kernels.py")
```
EmilyBourne, line 488: Were all these comments for debugging?
To run one particular test you can use the command:
```
pytest test_pyccel.py -k test_multiple_results
```

pyccel/ast/numpyext.py
rakati: No, It was there.
check [this](https://github.com/pyccel/pyccel/pull/524/files#diff-045bf48321280ff9b7f386d34e81f62aa08ce3e66405adba5f5675d123aeb725L69-L70).

tests/pyccel/test_pyccel.py
rakati: I forgot about this temporary change.

pyccel/ast/itertoolsext.py
rakati: Yes. poor description.
what about now?

pyccel/ast/numpyext.py
EmilyBourne: I mean why is it not `NumpyShape`

tests/pyccel/scripts/runtest_multiple_results.py
```diff
@@ -50,8 +50,8 @@ def print_multiple():
 
 print_multiple()
 
-print(min(f3()))
-print(max(f3()))
+# print(min(f3()))
+# print(max(f3()))
```
EmilyBourne, line 54: Why is this no longer working?

pyccel/ast/builtins.py
```diff
@@ -433,6 +434,64 @@ def __new__(cls, *args):
...
+
+
+    def fprint(self, printer, lhs=None):
+        """Fortran print."""
+
+        rhs_code = printer(self.arg)
+        if lhs:
+            lhs_code = printer(lhs)
+            return '{0} = sum({1})'.format(lhs_code, rhs_code)
+        return 'sum({0})'.format(rhs_code)
```
EmilyBourne, line 485: Why do we need a `fprint` function here rather than using the standard method of creating a `_print_PythonSum` function in codegen/printing/fcode.py?

pyccel/ast/builtins.py
```diff
@@ -457,3 +516,18 @@ def python_builtin_datatype(name):
 
     return None
 
+builtin_functions_dict = {
+    'abs'      : PythonAbs,  # TODO: create a built-in Abs
```
EmilyBourne, line 520: Is this TODO still valid?

pyccel/ast/builtins.py
```diff
@@ -457,3 +516,18 @@ def python_builtin_datatype(name):
...
+    'enumerate': Enumerate,
+    'int'      : PythonInt,
+    'float'    : PythonFloat,
+    'complex'  : PythonComplex,
+    'bool'     : PythonBool,
+    'sum'      : PythonSum,
+    'len'      : Len,
+    # 'max'      : PythonMax,
+    # 'min'      : PythonMin,
+    'not'      : Not,   # TODO [YG, 20.05.2020]: do not use Sympy's Not
```
EmilyBourne, line 532: Is this TODO still valid?

pyccel/ast/builtins.py
```diff
@@ -457,3 +516,18 @@ def python_builtin_datatype(name):
 
     return None
 
+builtin_functions_dict = {
+    'abs'      : PythonAbs,  # TODO: create a built-in Abs
+    'range'    : Range,
+    'zip'      : Zip,
+    'enumerate': Enumerate,
```
EmilyBourne, line 523: I assume you are leaving the renaming for another PR? In this case why is this PR marked as closing issue #258 ?

pyccel/ast/itertoolsext.py
```diff
@@ -0,0 +1,28 @@
...
+    def __new__(cls, *args):
+        if not isinstance(args, (tuple, list, Tuple)):
+            raise TypeError('args must be an iterable')
+        elif len(args) < 2:
+            return args[0]
+        return Basic.__new__(cls, *args)
+
+    @property
+    def elements(self):
+        "get expression's elements"
```
EmilyBourne, line 27: It is usual to use triple quotes for function doc-strings

EmilyBourne (changes requested):
This PR is marked as closing issue #258. I thought it was only a partial fix to the outstanding points in #258?

pyccel/ast/numpyext.py
rakati: I asked myself the same question, and I found the answer.
now I forget.
I will reconsider that.

tests/pyccel/scripts/runtest_multiple_results.py
rakati: they were linked to numpy min and max.
But I see that builtin min and max, do not behave the same as numpy min, max.
so we need to create a separate class for them.

pyccel/ast/numpyext.py
EmilyBourne: Is it because Shape is a function rather than a class?

pyccel/ast/builtins.py
rakati: no, I will remove the comment.

pyccel/ast/builtins.py
rakati: yes. we still use the sympy not.

tests/pyccel/scripts/runtest_multiple_results.py
EmilyBourne: When the argument is a tuple (as here) I think they behave in the same way. Would it be possible to implement that at least so as to not risk breaking code that currently works?
What are the main differences that need to be implemented?

pyccel/ast/builtins.py
rakati: I think we don't need to rename the python syntax things.

pyccel/ast/itertoolsext.py
rakati: a brand new way of docstrings coming with pyccel-rakati.

pyccel/ast/builtins.py
EmilyBourne: I think it's not as serious as there are no name collisions for these objects. But the aim is to have a new naming convention, so there is no reason for this to not include these classes

tests/pyccel/scripts/runtest_multiple_results.py
rakati: builtin max and min can take multiple arguments or an iterator or strings. while np.min or np.max takes only one arg (array).

EmilyBourne: So then could we raise a `NotImplementedError` for all those cases except `iterator` to reproduce the current behaviour?

pyccel/ast/builtins.py
rakati: ok. I will take care of that.

pyccel/ast/numpyext.py
rakati: yes. I thought about that.

rakati: we should respect name convention

pyccel/ast/builtins.py
```diff
@@ -8,13 +8,14 @@
 """
 
 from sympy import Symbol, Function, Tuple
-from sympy import Expr
+from sympy import Expr, Not
 from sympy import sympify
 from sympy.tensor import Indexed, IndexedBase
 
+import pyccel.ast.core as Core
```
EmilyBourne, line 15: Is this how you get around your circular import problem?

rakati: yes, I can't use this
```python
from .core import Variable
```
now it works fine. but Travis-ci failed with the Linux version

yguclu: I think that your solution to the circular import problem fools Codacy, but it does not work with Python 3.5...

yguclu: Apparently Python 3.5 supports relative circular imports: see the last bullet point [here](https://docs.python.org/3.5/whatsnew/3.5.html#other-language-changes). But you now use an absolute import to stop Codacy from complaining.

pyccel/ast/builtins.py
rakati: I don't know why exactly. but I see you guys used it in many Numpy classes (see `numpyext.py`).

EmilyBourne: In numpy it is usually done if the object can't be printed without a `lhs` (e.g. NumpyOnes).
Hopefully we can remove this separation after PR #521 goes through.
If you can't think of a reason then I would definitely move it to codegen/printing/fcode.py which is where we would normally search for it

tests/pyccel/scripts/runtest_multiple_results.py
rakati: I add now min and max to work like numpy max and min, but they need to handle a lot of cases and create tests for them.
which is out of scope in this PR. 
we need to create an issue for that.

yguclu (approved):
Excellent job! Thanks for taking care of this.

PR #525, opened 2020-11-04
Use iso_c types for fortran to ensure compatibility for the cwrapper

mostafaMamoni:
- [x] Use iso_c_binding to specify precision

- [x] Mark all structures and functions in the f2py_X.F90 file with bind (c)

### DIFF
--- a/pyccel/codegen/printing/fcode.py
+++ b/pyccel/codegen/printing/fcode.py
@@ -277,7 +277,7 @@ class FCodePrinter(CodePrinter):
                                             name=name)
 
         imports = ''.join(self._print(i) for i in expr.imports)
-        imports += 'use ISO_C_BINDING'
+        imports += 'use ISO_C_BINDING\n'
         decs    = ''.join(self._print(i) for i in expr.declarations)
         body    = ''
 
@@ -330,7 +330,7 @@ class FCodePrinter(CodePrinter):
         self._handle_fortran_specific_a_prioris(self.parser.get_variables(self._namespace))
         name    = 'prog_{0}'.format(self._print(expr.name)).replace('.', '_')
         imports = ''.join(self._print(i) for i in expr.imports)
-        imports += 'use ISO_C_BINDING'
+        imports += 'use ISO_C_BINDING\n'
         body    = self._print(expr.body)
 
         # Print the declarations of all variables in the namespace, which include:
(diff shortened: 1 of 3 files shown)


pyccel/codegen/printing/fcode.py
```diff
@@ -269,14 +270,14 @@ def _handle_fortran_specific_a_prioris(self, var_list):
...
         self._handle_fortran_specific_a_prioris(self.parser.get_variables(self._namespace))
-
         name = self._print(expr.name)
         name = name.replace('.', '_')
         if not name.startswith('mod_') and self.prefix_module:
             name = '{prefix}_{name}'.format(prefix=self.prefix_module,
                                             name=name)
 
         imports = ''.join(self._print(i) for i in expr.imports)
+        imports += 'use ISO_C_BINDING'
```
EmilyBourne, line 280: This should end with a `\n`

pyccel/codegen/printing/fcode.py
```diff
@@ -327,9 +328,9 @@ def _print_Module(self, expr):
 
     def _print_Program(self, expr):
         self._handle_fortran_specific_a_prioris(self.parser.get_variables(self._namespace))
-
         name    = 'prog_{0}'.format(self._print(expr.name)).replace('.', '_')
         imports = ''.join(self._print(i) for i in expr.imports)
+        imports += 'use ISO_C_BINDING'
```
EmilyBourne, line 333: This should end with a `\n`

EmilyBourne (changes requested):
This looks good to me but it seems to be failing for almost every test on travis. Are you sure f2py knows how to handle iso_c_binding?

yguclu:
> This looks good to me but it seems to be failing for almost every test on travis. Are you sure f2py knows how to handle iso_c_binding?

I agree, most likely there is a problem with f2py. @mostafaMamoni Do the tests run on your machine?

yguclu:
If indeed f2py does not understand the `use iso_c_binding` instruction, that would mean that this PR and #527 need each other to work. Hence I suggest that we merge this branch into [`ebourne_remove_f2py`](https://github.com/pyccel/pyccel/tree/ebourne_remove_f2py).

EmilyBourne:
This branch is now merged into PR #527 as the 2 are inter-dependant

PR #527, opened 2020-11-04
Remove f2py

EmilyBourne:
Remove f2py from pyccel.
- This requires a stricter handling of array input arguments. If a copy would be required then the input is not accepted (e.g. passing a C-ordered array when an F-ordered array was requested).
- Returning strings requires allocation of memory (a string is an array of chars). Thus this problem is similar to the problem of returning an array. This is beyond the scope of this PR, so support for returning strings has been temporarily dropped.
- The `C_BOOL` type has `kind=4`, this breaks `omp_set_dynamic` and `omp_set_nested` as they require `kind=1` (@saidctb presumably we need to add kind information to the header?)

### Commit summary
- Create a dictionary mapping types and kind parameter to their iso_c_binding name (cc: @mostafaMamoni )
- Use dictionary to print kind (cc: @mostafaMamoni )
- Remove f2py call for faily.f90. The c extension method (without f2py) does not require this
- Rename `f2py_` objects `bind_c_`
- Remove f2py comments for default values
- Save original function in `BindCFunctionDef`. This helps identify which arguments are shape arguments
- Add numpy getter functions, properties, and data type
- Do not print functions with functions as arguments in the wrapper file. These functions cannot be called from python so they do not need to be here
- Add necessary numpy libraries to `setup_X.py` file
- Pass object files not source files to `setup_X.py` file. This allows languages other than c/c++ to be used
- Ensure correct omp flags are passed for each platform (cc: @nhamidn )
- Add error printing for wrapping stage. As it becomes more complicated it can also raise errors
- [CODACY] Do not use `settings = {}` in `Printer.__init__` functions
- Duplicate `get_declare_type` function as `CCodePrinter` raises an error for unsupported array argument
- Reorder `codegen/printing/cwrappercode.py` file to group `_print_X` functions
- Set current numpy acceptable version to 1.19. Deprecation messages will be raised for things deprecated after this version
- Add `target_language` property to `CWrapperCodePrinter`
- Add `use ISO_C_BINDING` in interface printing
- Pass arguments by value in the wrapper unless they are optional
- Add `ISO_C_BINDING` precision to `BooleanTrue` and `BooleanFalse`
- Add `bind(c)` label to functions in the cwrapper (cc: @mostafaMamoni )
- Add a dictionary to map fortran flags to their c equivalent. This is necessary when compiling files in fortran but compiling the wrapper in c
- Add function to locate gfortran library in system
- Add support for capital letters in function names
- Revert array tests with boolean arguments. For f2py compatibility we had been passing `int` arguments and converting `bool` results to test booleans. This is no longer necessary
- Correct incorrect types decorators in `test_array`
- Remove tests passing F-ordered arrays to C-order functions and vice-versa. This would require a copy
- Skip tests for non-implemented concepts
- Remove xfail marks
- Use epyccel to test `highorder_functions`. The tests are therefore removed from `test_pyccel`


.travis.yml
```diff
@@ -144,10 +144,5 @@ jobs:
       - dlltool -d msmpi.def -l libmsmpi.a -D msmpi.dll
       - cd -
 
-    before_script:
-      # Call f2py once (it will fail) to clear some cache
-      - echo -e "subroutine foo()\nend subroutine\n" > faily.f90
-      - ${PYTHON} -m numpy.f2py -c -m faily faily.f90 || true
-      - rm faily.f90
```
yguclu, line 151: I am so happy that we could finally remove this hack! :smiley:

pyccel/ast/core.py
```diff
@@ -3179,7 +3179,7 @@ def __new__(cls, func, args, current_function=None):
         if not isinstance(args, (tuple, list, Tuple)):
             raise TypeError('> expecting an iterable')
 
-        # add the messing argument in the case of optional arguments
```
yguclu, line 3182: :rofl:

pyccel/ast/core.py
```diff
@@ -3924,8 +3924,21 @@ def rename(self, newname):
...
 
-class F2PYFunctionDef(FunctionDef):
-    pass
+class BindCFunctionDef(FunctionDef):
+    def __new__(cls, *args, original_function, **kwargs):
+        return FunctionDef.__new__(cls, *args, **kwargs)
+
+    def __init__(self, *args, original_function, **kwargs):
+        self._original_function = original_function
+        return FunctionDef.__init__(self, *args, **kwargs)
```
yguclu, line 3933: The `__init__` function does not return anything... I am surprised that Codacy does not complain about it.

EmilyBourne: Functions that don't return anything actually return `None`. This is probably why it doesn't see it as a problem

EmilyBourne:
@saniQQ @rakati I have asked for your reviews as I know you are working on the array things for C. I have added various numpy access functions to allow us to collect the information from arrays for fortran. You will need the same things for C. Please let me know if this section needs more documentation or if you need additional numpy access functions

nhamidn (approved):

mostafaMamoni (approved):

pyccel/ast/core.py
yguclu: I know, but writing `return None` is definitely not idiomatic Python. `__init__` functions should never contain a return statement.

yguclu: By the way, could you please add a short docstring to this important class?

pyccel/ast/datatypes.py
```diff
@@ -59,8 +59,31 @@
...
+        4 : 'C_FLOAT',
+	    8 : 'C_DOUBLE',
+        16 : 'C_LONG_DOUBLE'},
+    "complex" : {
+        4 : 'C_FLOAT_COMPLEX',
+	    8 : 'C_DOUBLE_COMPLEX',
+        16 : 'C_LONG_DOUBLE_COMPLEX'},
+    "logical" : {
+		4 : "C_BOOL"}
+}
```
yguclu, line 79: There seems some misalignment here, which makes it hard to read. Could you please improve the formatting a bit?

tests/epyccel/test_bitwise.py
```diff
@@ -49,7 +49,6 @@ def test_left_shift_b_b(language, a, b):
 @pytest.mark.parametrize( 'language', [
         pytest.param("c", marks = pytest.mark.c),
         pytest.param("fortran", marks = [
-            pytest.mark.xfail(reason="problem in wrapping issue #370"),
             pytest.mark.fortran]
```
yguclu, line 52: We don't need to assign a list to `marks` anymore, right?

tests/epyccel/test_bitwise.py
```diff
@@ -97,7 +96,6 @@ def test_bit_or_i_i(language, a, b):
...
```
yguclu, line 99: See above

tests/epyccel/test_bitwise.py
```diff
@@ -126,7 +124,6 @@ def test_bit_and_i_i(language, a, b):
...
```
yguclu, line 127: See above

tests/epyccel/test_epyccel_default_args.py
```diff
@@ -67,7 +67,6 @@ def f3(x = 1.5, y = 2.5):
 #------------------------------------------------------------------------------
 @pytest.mark.parametrize( 'language', (
         pytest.param("fortran", marks = [
-            pytest.mark.xfail(reason="f2py does not handle bool default values"),
             pytest.mark.fortran]),
```
yguclu, line 70: See above

tests/epyccel/test_epyccel_default_args.py
```diff
@@ -91,7 +90,6 @@ def f4(x = True):
 #------------------------------------------------------------------------------
 @pytest.mark.parametrize( 'language', (
         pytest.param("fortran", marks = [
-            pytest.mark.xfail(reason="Complex default value not written in a f2py readable manner"),
             pytest.mark.fortran]),
```
yguclu, line 93: See above

tests/epyccel/test_epyccel_optional_args.py
```diff
@@ -8,7 +8,6 @@
 #------------------------------------------------------------------------------
 @pytest.mark.parametrize( 'language', [
         pytest.param("fortran", marks = [
-            pytest.mark.xfail(reason="f2py does not support optional arguments"),
             pytest.mark.fortran]),
```
yguclu, line 11: I have the same comment as above.
BTW, why do we need to pass the same pytest marks to so many functions in this module?

tests/epyccel/test_higherorder_functions.py
```diff
@@ -5,8 +5,7 @@
 
 @pytest.fixture( params=[
         pytest.param("fortran", marks = [
-            pytest.mark.fortran,
-            pytest.mark.skip]),
+            pytest.mark.fortran]),
         pytest.param("c", marks = [
             pytest.mark.c]
```
yguclu, line 10: Again, we can do without the lists, right?

yguclu (changes requested):
Excellent job!

I find it quite amazing that the `codegen.printing.cwrappercode` module works for both Fortran and C with very little distinction between the two languages. You only check the language 4 times (with `if-else` blocks) in the whole module!

I have some minor comments.

tests/epyccel/test_epyccel_optional_args.py
EmilyBourne: We mark functions depending on which language they are translated to. This is actually done in every file, but when there is no language specific xfail mark, it can be handled with the fixture so it is invisible everywhere else

saniQQ (approved):

EmilyBourne:
The openmp tests are passing on this branch when the option `compiler = ifort` is added to the epyccel command

EmilyBourne:
When a compiler is specified. This compiler is used for compiling all pyccel generated files except `X_wrapper.c`. `X_wrapper.c` is compiled by python's `distutils.setup`. As a result it is compiled and linked with the compiler used to build python. This seems to match what f2py does. If we compile using:
```
f2py -c  --fcompiler=intelem
```
Then the generated file has the following extension : `cpython-37m-x86_64-linux-gnu.so`

EmilyBourne:
f2py uses intel to link. This allows intel to use Interprocedural Optimizations. We should add this as an improvement later

yguclu (approved):

PR #528, opened 2020-11-05
Extend header language by adding intent information

### LINKED ISSUE #499: Extend header language by adding intent information
In the current header language , there is no way to specify the intent of an argument
so we should add a new keyword `const` to specify that an argument is intent(in) , otherwise it's intent(inout).

E.g
`#$ header function(const int[:], float[:])` will  make the first argument as `intent(in)` and the second as `intent(inout)`

**Acceptance Criteria**
- [x] Add const to grammar
- [x] Verify that calculated intent matches grammar if `const` is specified
- [ ] Create a new class `Argument` which should inherit from `Variable` and will contain intent information

nhamidn:
This PR is about extending the header language, and it closes #499:

- Add `const` keyword to the header grammar;
- When `const` is specified, check that calculated intent is `in` (otherwise raise error);
- Add tests for new `const` keyword.


pyccel/ast/core.py
```diff
@@ -2566,6 +2570,10 @@ def __init__(
             allocatable = False
         self.allocatable = allocatable
 
+        if is_const is None:
+            is_const = False
```
EmilyBourne, line 2574: It might be good to also add a type check

pyccel/ast/core.py
```diff
@@ -5589,28 +5597,29 @@ def is_None(expr):
...
 
-def get_assigned_symbols(expr):
+def get_assigned_symbols(expr,all):
     """Returns all assigned symbols (as sympy Symbol) in the AST.
 
     Parameters
     ----------
     expr: Expression
         any AST valid expression
+    all: get all symbols if TRUE (ignore rank)
```
EmilyBourne, line 5607: I don't understand what this does. The docstring should say which type to pass. Is `all` a `bool`?
What does the `rank` have to do with anything? What happens if I pass `false`? It returns no symbols? Only some symbols?

pyccel/codegen/printing/fcode.py
```diff
@@ -939,6 +940,8 @@ def _print_Declare(self, expr):
         if intent:
             if intent == 'in' and rank == 0 and is_static is False:
                 intentstr = ', value'
+                if is_const is True:
```
EmilyBourne, line 943: No need to compare with `True`:
`if is_const:`

pyccel/parser/semantic.py
```diff
@@ -2543,6 +2547,11 @@ def _visit_FunctionDef(self, expr, **settings):
                             args_inout[i] = True
 
                         i_fa += 1
+                if (args_inout[i] is True and a.is_const is True) or (str(a) in all_assigned and a.is_const is True):
```
EmilyBourne, line 2550: What is the second condition checking for here? `is_const = True` and `a` is an array? Does this mean we cannot specify that arrays are constant?

pyccel/parser/syntax/headers.py
```diff
@@ -150,14 +152,18 @@ def __init__(self, **kwargs):
...
         super(UnionTypeStmt, self).__init__(**kwargs)
 
     @property
     def expr(self):
         l = []
-        for i in self.dtype:
+        for num, i in enumerate(self.dtype):
             l += [i.expr]
+            if self.const is not None:
+                l[num]["is_const"] = True
```
EmilyBourne, line 165: Doesn't `l` contain lists not dictionaries? I'm not sure how this works

pyccel/ast/core.py
nhamidn: yes its a `bool`,
i have added the intent(in) to argument which have `rank = 0`, so when using `get_assigned_symbols() `it get only variable with `rank != 0`, which lead to compilation error for this example:

```
#$ header function funct_c(const int, int)
def funct_c( x, a ):
    x += a
    return x
```
in this case we need to raise an error because the user defined x as a const by also he is modifying it.

EmilyBourne: So if `all == True` then it returns all symbols to which something is assigned? And if `all == False` then it only returns symbols to which something is assigned with `rank!=0`?

pyccel/parser/semantic.py
nhamidn: the second condition check if an argument with a `rank == 0` is modified so we raise an error if `const` is specified for this argument

pyccel/parser/semantic.py
```diff
@@ -2502,9 +2503,12 @@ def _visit_FunctionDef(self, expr, **settings):
 
             results_names = [str(i) for i in results]
 
-            assigned = get_assigned_symbols(body)
+            assigned = get_assigned_symbols(body, False)
             assigned = [str(i) for i in assigned]
 
+            all_assigned = get_assigned_symbols(body, True)
+            all_assigned = [str(i) for i in all_assigned]
```
EmilyBourne, line 2510: If we need the result for `all=True` and `all=False` would it not be simpler to do:
```
all_assigned = get_assigned_symbols(body, True)
assigned      = [a for a in all_assigned if a.rank >0]
```
?
This way we don't need to parse the tree twice

pyccel/ast/core.py
nhamidn: exactly

EmilyBourne: Ok. Can you update the docstring to make this clearer please. As a minimum you need to provide the expected type. It would be good if you could also specify in the docstring what the expected result is if `all = False`

pyccel/parser/semantic.py
EmilyBourne: Is the `all` argument still necessary with this improvement?

pyccel/ast/core.py
nhamidn: i didn't get what you mean by adding a type check

pyccel/parser/semantic.py
nhamidn: no, no need for the `all` argument now.

pyccel/parser/syntax/headers.py
nhamidn: it contain dictionaries

EmilyBourne: Then can you explain the line above it:
`l += [i.expr]`?

nhamidn: im not sure does this line do, from what i understand all the datatypes from the header

EmilyBourne: I think half of your sentence is missing ^^'. I didn't think you could add a list to a dictionary (hence my question). Am I wrong? I know that we can declare a dictionary from a list as follows:
```
d = dict([('a', 1), ('b', 2), ('c', 3)])
```
Does adding a list to a dictionary work in the same way?

pyccel/ast/core.py
yguclu: Checking that the type of the `is_const` argument is indeed a `bool`.

yguclu: But these two lines are a bit strange: the default value of this argument is `False`, so why should anybody pass `is_const = None` to this function?

yguclu: I see that the same (strange) approach is used for other arguments in this constructor, so I would suggest you try and clean it up a bit.

pyccel/parser/syntax/headers.py
yguclu: `i.expr` is a dictionary, `[i.expr]` is a list, and `l += [i.expr]` extends the list `l`. But this all property is written in a strange way, it's not very clear... What about this instead?
```python
l = []
for i in self.dtype:
    if self.const:
        i.expr["is_const"] = True
    l.append(i.expr)
```
or this
```python
l = [i.expr for i in self.dtype]
if self.const:
    for e in l:
        e["is_const"] = True
```

pyccel/parser/semantic.py
yguclu: I also have a question here. If you are just evaluating `bool` variables, why not writing this?
```python
if (args_inout[i] and a.is_const) or (str(a) in all_assigned and a.is_const):
```
By the way, this is equivalent to
```python
if a.is_const and (args_inout[i] or str(a) in all_assigned):
```

yguclu (changes requested):
Some minor cleanup would be appreciated.. see comments!

pyccel/parser/semantic.py
nhamidn: you are right the condition can be written in a simple and clear way

pyccel/parser/syntax/headers.py
nhamidn: Yes the `i.expr` is a dictionary and `l += [i.expr]` extends the list `l`

EmilyBourne (approved):

pyccel/parser/semantic.py
```diff
@@ -2659,7 +2659,7 @@ def _visit_FunctionDef(self, expr, **settings):
                             args_inout[i] = True
 
                         i_fa += 1
-                if (args_inout[i] is True and a.is_const is True) or (str(a) in all_assigned and a.is_const is True):
+                if (args_inout[i] or (str(a) in all_assigned)) and a.is_const:
```
yguclu, line 2662: Just a minor detail...

As far as I know the Python interpreter processes a condition in the form `a and b` by first checking `a`, and then it checks `b` only if `a == True`. Hence you should put in `a` the easiest condition to evaluate, in this case `a.is_const`.

yguclu (changes requested):

EmilyBourne:
I have just realised that we calculate whether a variable is_const, but we do not seem to save this information anywhere. Is this done on purpose?

When compiling the file `pyccel/tests/pyccel/scripts/funcs.py` with ifort (version 12.0.0) I get the following error:
```
 The specification-part of a PURE function must specify that all dummy arguments have explicitly declared INTENT(IN) except procedure arguments and arguments with the POINTER attribute.
```
I thought this branch would fix this problem. But it only fixes it if I specify `const` manually in the decorator

pyccel/parser/semantic.py
nhamidn: when i write the condition like :
`if a.is_const and (args_inout[i] or (str(a) in all_assigned)):`
the `test_higherorder_functions.py` fails saying that :
` AttributeError: 'FunctionAddress' object has no attribute 'is_const'`
so i think i need also to check for` isinstance(a, Variable) == True`

yguclu (approved):

EmilyBourne:
> I have just realised that we calculate whether a variable is_const, but we do not seem to save this information anywhere. Is this done on purpose?
> 
> When compiling the file `pyccel/tests/pyccel/scripts/funcs.py` with ifort (version 12.0.0) I get the following error:
> 
> ```
>  The specification-part of a PURE function must specify that all dummy arguments have explicitly declared INTENT(IN) except procedure arguments and arguments with the POINTER attribute.
> ```
> 
> I thought this branch would fix this problem. But it only fixes it if I specify `const` manually in the decorator

It seems that this is caused by using old compilers. The default ifort compiler on my machine was:
```
ifort (IFORT) 12.0.0 20101116
Copyright (C) 1985-2010 Intel Corporation.  All rights reserved.
```
The default gfortran version was:
```
GNU Fortran (GCC) 4.4.7 20120313 (Red Hat 4.4.7-23)
Copyright (C) 2010 Free Software Foundation, Inc.

GNU Fortran comes with NO WARRANTY, to the extent permitted by law.
You may redistribute copies of GNU Fortran
under the terms of the GNU General Public License.
For more information about these matters, see the file named COPYING
```
This version raises the same error as intel:
```
Error: Argument 'x' of pure function 'add2' at (1) must be INTENT(IN)
```

pyccel/parser/semantic.py
```diff
@@ -2656,6 +2659,12 @@ def _visit_FunctionDef(self, expr, **settings):
                             args_inout[i] = True
 
                         i_fa += 1
+                if isinstance(a, Variable):
+                    if a.is_const and (args_inout[i] or (str(a) in all_assigned)):
+                        msg = 'Argument ({}) must be an inout!'.format(a)
```
yguclu, line 2664: Sorry for this late comment, but I think that we should give a more meaningful error message to the user.

A non-Fortran programmer may not understand what `intent` means, so I recommend
```python
msg = "Cannot modify 'const' argument ({})".format(a)
```

yguclu (approved):

Issue #512, opened 2020-10-26
Handle interfaces in the cwrapper


EmilyBourne:
Description
------
As per issue #301 interfaces will be added. E.g:
```python
# Create 4 functions
@types('int','int')
@types('real','real')
@types('int','real')
@types('real','int')
def f(a,b):
    pass
```
The c wrapper needs to detect the input types and then choose the appropriate function from the interfaces. This can be done using `PyLong_Check`, `PyInt_Check`, etc (see https://github.com/pyccel/pyccel/blob/master/pyccel/ast/cwrapper.py#L340)

Acceptance Criteria
------
TBD

PR #530, opened 2020-11-09
Implement function interfaces

### LINKED ISSUE #301: Implement function interfaces
**Describe the improvement**
Pyccel will allow the user to declare multiple possible types for a function argument. This allows the same code to be used for multiple functions with different types.

**Test Code**
The syntax in python will be as follows:
```
# Create 4 functions
@types('int','int')
@types('real','real')
@types('int','real')
@types('real','int')
def f(a,b):
    pass

# Create 4 functions
#$ header function f(int,int)
#$ header function f(real,real)
#$ header function f(int,real)
#$ header function f(real,int)
def f(a,b):
    pass
```
The user can therefore choose to either specify all possible combinations of types (example 1)

Consider the following example:
```
#$ header function decr(int)
#$ header function decr(real)
def decr(x):
    y = x - 1
    return y
```

In this example Pyccel should generate one Fortran function where `x` is `integer`, and another function where `x` is `real`. Moreover, Pyccel should generate a **Fortran interface** for calling both functions with a common name. This interface should only exist in the module, not in the generated `f2py_mod.f90` file, here the cwrapper will handle the interface

E.g:
```
module function_decr_interface

implicit none

interface decr
module procedure decr_0
module procedure decr_1
end interface

contains

function decr_0(x) result(y)
  implicit none
  real(kind=8)  :: y  
  real(kind=8)  :: x  
  y = x - 1_8
  return
end function

function decr_1(x) result(y)
  implicit none
  real(kind=8)  :: y  
  real(kind=8), value  :: x
  y = x - 1_8
  return
end function

end module
```

For C 2 functions should be generated and the appropriate function should be used as required

This issue concerns only code generation not the cwrapper

### LINKED ISSUE #511: Allow function interfaces to be implemented using templates
**Description**
Pyccel will allow the user to declare multiple possible types for a function argument. This allows the same code to be used for multiple functions with different types. 

**Test Code**
```python
# Create 4 functions
@template('T', values=['int','real'])
@template('R', values=['int','real'])
@types('T','R')
def f(a,b):
    pass

# Create 2 functions
@template('T', values=['int','real'])
@types('T','T')
def f(a,b):
    pass


# Create 2 functions
#$ header template T(int|real)
#$ header function f(T,T)
def f(a,b):
    pass

# Create 4 functions
#$ header template T(int|real)
#$ header template R(int|real)
#$ header function f(T,R)
def f(a,b):
    pass
```

The user can therefore choose to either specify all possible combinations of types (see issue #301) or to create a template type which can have one of a number of values. By combining multiple template types (example 2) we can therefore create a cross product of all proposed types

rmahjoubi:
**Implementation of interfaces**

-Allow the user to declare multiple possible types for a function's arguments. This allows the same code to be used for multiple function's types..
-Give the user the ability to declare Templates which can represent different types for a single variable.
-Allow any combination of the two above.
-The user can declare a Template dedicated for a single function by decorating it using the Template decorator, this will override any  Template with the same name.
-The user can declare a Template using pragmas.
-Any combination of the two above is allowed.
-In Fortran, the needed functions are generated, and also an interface to call them with the same name.
-In C, the needed functions are generated, and the appropriate one is called.
-Add tests for interfaces.

Fixes #301 
Fixes #511


pyccel/codegen/printing/ccode.py
```diff
@@ -685,6 +685,8 @@ def _print_Rand(self, expr):
     def _print_NumpyRandint(self, expr):
         raise NotImplementedError("Randint not implemented")
 
+    def _print_Interface(self, expr):
```
MenoIy, line 688: is this printer not supposed to call the printer for every function in the interface?

rmahjoubi: It doesn't, it only prints the interfaces, and since there are none in this case, it returns an empty string

pyccel/ast/headers.py
```diff
@@ -163,7 +182,7 @@ def kind(self):
     def is_static(self):
         return self._args[4]
 
-    def create_definition(self):
+    def create_definition(self, templates = []):
```
EmilyBourne, line 185: You should be able to use a tuple to get rid of the codacy warning. Default arguments are static objects so they raise errors if they are modifiable. This is because if I do:
```
def f(a = []):
   print(a)
   a.append(1)

f()
f()
f()
```
The output will be:
```
[]
[1]
[1, 1]
```

pyccel/ast/headers.py
```diff
@@ -211,7 +230,27 @@ def build_argument(var_name, dc):
                 dtypes += [[i]]
             else:
                 raise TypeError('element must be of type UnionType or dict')
-        for args_ in product(*dtypes):
+
+        #TODO: handel the case of functions arguments
```
EmilyBourne, line 234: Typo

pyccel/ast/headers.py
```diff
@@ -211,7 +230,27 @@ def build_argument(var_name, dc):
...
+                    iterx[i] = templates[tmplt].args
+                    old_values[j['datatype']] = i
+                else:
+                    iterx[i] = [j]
+            for itery in product(*iterx):
+                itery = list(itery)
+                for j, i in enumerate(itery):
+                    if 'datatype' in i and i['datatype'] in templates_names:
+                        itery[j] = itery[old_values[i['datatype']]].copy()
+                arg_codes.append(itery)
```
EmilyBourne, line 252: Can you add some comments to this section to explain what each step is doing please

pyccel/ast/core.py
```diff
@@ -3764,6 +3711,93 @@ def __str__(self):
                 args   = ', '.join(self.args),
                 result = result)
 
+class Interface(Basic):
+
+    """Represent an Interface"""
```
EmilyBourne, line 3716: Can you add a more detailed docstring including Parameters please

pyccel/ast/core.py
```diff
@@ -3764,6 +3711,93 @@ def __str__(self):
...
+    @property
+    def name(self):
+        return self._name
+
+    @property
+    def functions(self):
+        return self._functions
+
+    @property
+    def hide(self):
```
EmilyBourne, line 3746: Can you add docstrings to some (ideally all) of these class methods please.
What is this function for?

pyccel/parser/semantic.py
```diff
@@ -362,12 +362,19 @@ def insert_class(self, cls, parent=False):
         else:
             raise TypeError('Expected A class definition ')
 
+    def insert_template(self, expr):
+        """insert a template"""
```
EmilyBourne, line 366: What are we inserting a template into?

pyccel/parser/semantic.py
```diff
@@ -362,12 +362,19 @@ def insert_class(self, cls, parent=False):
...
     def insert_header(self, expr):
         """."""
         if isinstance(expr, MethodHeader):
             self.namespace.headers[expr.name] = expr
         elif isinstance(expr, FunctionHeader):
-            self.namespace.headers[expr.func] = expr
+            if expr.func in self.namespace.headers:
+                self.namespace.headers[expr.func].append(expr)
+            else:
+                self.namespace.headers[expr.func] = [expr]
```
EmilyBourne, line 377: Is there a significant difference in how we treat functions and methods?

pyccel/parser/semantic.py
```diff
@@ -492,11 +499,21 @@ def remove_variable(self, name):
...
         """."""
         container = self.namespace
+        headers = []
         while container:
             if name in container.headers:
-                return container.headers[name]
+                headers += container.headers[name]
             container = container.parent_scope
-        return None
+        return headers
```
EmilyBourne, line 507: Are you sure we need to look for the headers in multiple scopes? Can you give an example where this is necessary?

pyccel/parser/semantic.py
```diff
@@ -492,11 +499,21 @@ def remove_variable(self, name):
...
+
+    def get_templates(self):
+        """."""
```
EmilyBourne, line 510: Please add a real docstring not a placeholder

pyccel/parser/semantic.py
```diff
@@ -492,11 +499,21 @@ def remove_variable(self, name):
...
+        container = self.namespace
+        templates = []
+        while container:
+            templates += container.templates
+            container = container.parent_scope
+        return templates
```
EmilyBourne, line 516: Does this handle name collisions in different scopes?
```
@template('T',['int','real','complex'])
...
def bla():
    @template('T',['int','real'])
    ...
```

pyccel/parser/semantic.py
```diff
@@ -2591,26 +2585,20 @@ def _visit_FunctionDef(self, expr, **settings):
...
-            for i,f in enumerate(funcs):
+            for f in funcs:
                 #TODO add new scope for the interface
-                self.namespace.sons_scopes[name+'_'+ str(i)] = self.namespace.sons_scopes[name]
-                new_funcs.append(f.clone(name+'_'+ str(i)))
+                self.insert_function(f)
 
-            funcs = Interface(name, new_funcs)
+            self.create_new_function_scope(name)
+            self.exit_function_scope()
```
EmilyBourne, line 2599: Does this serve a purpose?

pyccel/parser/syntactic.py
```diff
@@ -633,54 +642,110 @@ def _visit_FunctionDef(self, stmt):
...
+
+            if not isinstance(decorators['template'], list):
+                decorators['template'] = [decorators['template']]
+            for comb_types in decorators['template']:
+                if len(comb_types.args) != 2:
+                        msg = 'Number of Arguments provided to the types decorator is not valid'
+                        errors.report(msg,
+                                        symbol = comb_types,
+                                        bounding_box = (stmt.lineno, stmt.col_offset),
+                                        severity='error')
```
EmilyBourne, line 658: Why does `comb_types.args` have to have exactly 2 elements?

pyccel/decorators.py
```diff
@@ -38,6 +39,12 @@ def identity(f):
         return f
     return identity
 
+def template(name, types=[]):
+    """template decorator."""
+    def identity(f):
+        return f
+    return identity
```
EmilyBourne, line 46: Do you not also need to pass the function `f`? You can test this by running your test code in pure python mode

pyccel/parser/syntactic.py
```diff
@@ -633,54 +642,110 @@ def _visit_FunctionDef(self, stmt):
...
+
+                tp_name = comb_types.args[0]
+                ls = comb_types.args[1]
+                if isinstance(ls, ValuedArgument):
+                    arg_name = ls.name
+                    ls  = ls.value
+                    if not arg_name == 'types':
```
EmilyBourne, line 665: We force the `name` arg to never be a keyword arg?

pyccel/parser/syntactic.py
```diff
@@ -633,54 +642,110 @@ def _visit_FunctionDef(self, stmt):
...
-                                  severity='error')
-
-                i = i+1
-
-            txt  = '#$ header ' + name
-            txt += '(' + ','.join(types) + ')'
+                        ls = ls if isinstance(ls, (PythonTuple)) else list(ls)
+
+                while i<len(ls) :
+                    arg = ls[i]
```
EmilyBourne, line 675: Is ther a reason for using a while loop rather than a for loop?

pyccel/parser/syntactic.py
```diff
@@ -633,54 +642,110 @@ def _visit_FunctionDef(self, stmt):
...
+                decorators['types'] = [decorators['types']]
+            for comb_types in decorators['types']:
+                types = []
+                results = []
+                container = types
+                i = 0
+                ls = comb_types.args
+    
+                while i<len(ls) :
+                    arg = ls[i]
```
EmilyBourne, line 708: See above, why use a while?

pyccel/parser/syntactic.py
```diff
@@ -633,54 +642,110 @@ def _visit_FunctionDef(self, stmt):
...
+                        container = results
+                        if not arg_name == 'results':
+                            msg = 'Argument "{}" provided to the types decorator is not valid'.format(arg_name)
+                            errors.report(msg,
+                                        symbol = comb_types['types'],
+                                        bounding_box = (stmt.lineno, stmt.col_offset),
+                                        severity='error')
+                        else:
+                            ls = arg if isinstance(arg, PythonTuple) else [arg]
+                            i = -1
```
EmilyBourne, line 729: Is this why you use a while? You can use `break` for this

EmilyBourne, line 729: Is this why you use a while? To recurse?

I think this would be more legible as a function

tests/epyccel/test_generic_functions.py
```diff
@@ -0,0 +1,94 @@
+ # pylint: disable=missing-function-docstring, missing-module-docstring/
+import pytest
+import modules.generic_functions as mod
+from pyccel.epyccel import epyccel
+
+@pytest.fixture( params=[
+        pytest.param("fortran", marks = [
+            pytest.mark.fortran,
+            pytest.mark.skip]),
```
EmilyBourne, line 9: Please add a reason when you skip

tests/epyccel/test_generic_functions.py
```diff
@@ -0,0 +1,94 @@
...
+        pytest.param("c", marks = [
+            pytest.mark.c]
```
EmilyBourne, line 11: Does this PR already implement issue #512 ?

pyccel/parser/semantic.py
rmahjoubi: they are inserted to the templates of the scope

pyccel/parser/semantic.py
rmahjoubi: I don't think so, I'am going to handle them in the same manner

pyccel/parser/semantic.py
rmahjoubi: this was the way we do it before, I can't think of a reason to look for them in multiple scopes. should I check only in the current scope

pyccel/parser/semantic.py
rmahjoubi: now it's handled, a template overrides any templates with the same name in its parents scopes, although we can't test it since neither classes nor nested functions are implemented yet

pyccel/parser/semantic.py
rmahjoubi: It doesn't, I just removed it

pyccel/parser/syntactic.py
rmahjoubi: `@template('z', types=['int', 'real'])`, we are giving the choice for the second argument to be a keyword arg or not while forcing it the key to be `types` in case of keyword arg

pyccel/parser/semantic.py
EmilyBourne: The templates are inserted into the scope? Or something is inserted into the templates of the scope?
Whichever it is can you write it in the docstring please

pyccel/parser/semantic.py
EmilyBourne: I'm not sure. @saidctb ?

pyccel/parser/semantic.py
EmilyBourne: nested functions are implemented in fortran

pyccel/parser/syntactic.py
EmilyBourne: So yes?

EmilyBourne: Why?

tests/epyccel/test_generic_functions.py
rmahjoubi: No it doesn't, I am pyccelizing the modules in the tests

EmilyBourne: Ah ok. So you don't actually call any interface functions directly here :+1:

pyccel/parser/semantic.py
rmahjoubi: a scope have a list of templates, here we append the given template to that list

rmahjoubi: yes sure

pyccel/parser/semantic.py
rmahjoubi: ok perfect I will add tests in fortran

pyccel/parser/syntactic.py
rmahjoubi: I am not sure where the problem exactly is, do you mean we should force the list of types to be a keyword arg?

EmilyBourne: No I mean you have said that I am not allowed to do:
`@template(name='z', types=['int', 'real']`
But I don't understand why

rmahjoubi: yes you are right this case is not handled properly

rmahjoubi: it's now fixed

pyccel/parser/syntactic.py
rmahjoubi: Yes that's why it's used. you think I should make a function for the while block, it's only used in one place

EmilyBourne: This only works because the results have to be given by keyword and the arguments have to be given by position right? If the order was changed then you might miss some values

EmilyBourne: I'm happy to leave it as it is for now and get a second opinion from other reviewers when the branch is passing about what is more legible

pyccel/decorators.py
rmahjoubi: We don't actually need to decorate the function, passing the function `f` doesn't serve any purpose.

EmilyBourne: Hmmm, this is an interesting question that probably deserves some discussion.

On the one hand, if the template can be used for multiple functions then it does not make sense for it to be a decorator

On the other hand, if we want to translate a function rather than an entire module we need the template to be attached to the function so that we have the information when we inspect the code

EmilyBourne: Do you have a suggestion for what we could use instead? Maybe it should be an argument to the epyccel function to allow for reuse of the templates without code duplication?

pyccel/parser/semantic.py
```diff
@@ -2576,12 +2569,11 @@ def _visit_FunctionDef(self, expr, **settings):
...
 
-            if len(interfaces) == 1:
-                # insert the FunctionDef into the scope
-                # to handle the case of a recursive function
-                # TODO improve in the case of an interface
-                func = FunctionDef(name, args, results, [])
-                self.insert_function(func)
+            # insert the FunctionDef into the scope
+            # to handle the case of a recursive function
+            # TODO improve in the case of an interface
```
saidctb, line 2574: you can delete this TODO, can you confirm that the interface works when we have a recursive function ?

pyccel/decorators.py
rmahjoubi: I don't think we have such problem, since we gather all the templates pragmas/decorators in `FunctionDef`, it's the samething we do for types, basically `template` and `types` works in the same manner.
what I meant before is that we don't need the actual python decoration to happen, we just need to be able to retrieve this information from the `ast`, this is why I favored not passing the function `f`, because it requires more checks and does the same thing as this.

EmilyBourne: > I don't think we have such problem

Which problem exactly are you referring to?

> basically template and types works in the same manner

Yes, but `types` is always specific to a function

> what I meant before is that we don't need the actual python decoration to happen

You are right, apparently the code above does not cause an error 

> we just need to be able to retrieve this information from the ast

I would still double check that you can do that when you `epyccel`ize a single function rather than an entire module. You won't be able to write a test for this until issue #512 is fixed, but it would be worth checking that the case passes the semantic step. When we do:
```
import mod
f = epyccel(mod.test_func)
```
we do not have access to the entire code of `mod`. Instead we just have access to the specified function and decorators that are attached to it. If you don't pass the function then I wonder if the decorator is attached

rmahjoubi: I am not sure, i thought since something like this works:
```
#$ header function f2(int)
#$ header function f1(int)
def f2(a):
    pass
def f1(a):
    pass
```
then this will work:
```
#$ header template T(int|real)
#$ header template U(int|real)
#$ header function f(T, U)
def f(a, b):
   pass
```
because I handled them in the same way. I'm going to test if `f = epyccel(mod.f)` passes the semantics as you suggested.
As for the decorators I presumed if the user decorated a function with a template then he wanted it to be used only for that function, so it's not available for other functions. In the other hand any template pragma written in the scope of the function or it's parents scopes are available to that function. 

> If you don't pass the function then I wonder if the decorator is attached

I checked the `ast` for this, the information is preserved.

pyccel/parser/semantic.py
```diff
@@ -506,11 +511,25 @@ def remove_variable(self, name):
     def get_header(self, name):
         """."""
         container = self.namespace
+        headers = []
         while container:
             if name in container.headers:
-                return container.headers[name]
+                if isinstance(container.headers[name], list):
+                    headers += container.headers[name]
```
EmilyBourne, line 518: You could also use `headers.extend(container.headers[name])`. I tried to search to find out if there's a speed difference but I couldn't find anything so it's up to you which you prefer

pyccel/parser/semantic.py
```diff
@@ -2432,25 +2418,27 @@ def _visit_FunctionDef(self, expr, **settings):
...
-                if (args_number < len(header.dtypes)):
+        if cls_name:
+            header += self.get_header(cls_name +'.'+ name)
+            args_number -= 1
+        else:
+            header += self.get_header(name)
+
+        for hd in header:
+            if (args_number != len(hd.dtypes)):
+                msg = 'The number of arguments in the function {} ({}) does not match the number of types in decorator/header ({}).'.format(name ,args_number, len(hd.dtypes))
```
EmilyBourne, line 2440: This line doesn't fit on my screen. Please split it

pyccel/parser/semantic.py
```diff
@@ -2505,6 +2497,8 @@ def _visit_FunctionDef(self, expr, **settings):
             arguments      = expr.arguments
             header_results = m.results
 
+            if len(interfaces) > 1:
+                name = interface_name + '_' + str(i)
```
EmilyBourne, line 2501: Have you thought about formatting this so you always get a name with the same length? e.g.:
```
interface_name_01
interface_name_02
...
interface_name_12
```
instead of:
```
interface_name_1
interface_name_2
...
interface_name_12
```

pyccel/parser/syntactic.py
```diff
@@ -633,54 +643,127 @@ def _visit_FunctionDef(self, stmt):
...
+                        arg = arg.strip("'").strip('"')
+                        types.append(arg)
+                    else:
+                        msg = 'Invalid argument type of type {} passed to template decorator'.format(type(arg))
+                        errors.report(msg,
+                                    symbol = comb_types,
+                                    bounding_box = (stmt.lineno, stmt.col_offset),
+                                    severity='error')
+
+                    i += 1
```
EmilyBourne, line 702: Why can't this be a for loop?

pyccel/parser/syntactic.py
```diff
@@ -633,54 +643,127 @@ def _visit_FunctionDef(self, stmt):
...
+                        arg_name = arg.name
+                        arg  = arg.value
+                        container = results
+                        if not arg_name == 'results':
+                            msg = 'Argument "{}" provided to the types decorator is not valid'.format(arg_name)
+                            errors.report(msg,
+                                        symbol = comb_types,
+                                        bounding_box = (stmt.lineno, stmt.col_offset),
+                                        severity='error')
+                        else:
```
EmilyBourne, line 745: What can the `arg_name` be other than `results`?

tests/codegen/fcode/scripts/generic_methods.py
```diff
@@ -15,21 +15,3 @@ def __del__(self):
...
-
-p.y = a + 5
-p.y = p.x + 5
-
-p.translate(1.0, 2.0)
-
-print(p.x, p.y)
-print(a)
-
-del p
```
EmilyBourne, line 35: What error do you get in this section? Maybe we should keep a copy with the bug in the known_bugs folder ?

tests/errors/semantic/blocking/INTERFACE_WRONG_TYPES.py
```diff
@@ -0,0 +1,15 @@
+# pylint: disable=missing-function-docstring, missing-module-docstring/
+from pyccel.decorators import types
+from pyccel.decorators import template
+
+@template('z', types=['int', 'real'])
+@template('y', types=['int', 'real'])
+@types('z', 'z', 'y')
+def multi_tmplt_1(x, y, z):
```
EmilyBourne, line 8: Could you add a similar test where the types of `y` and `z` are not the same.

E.g.
```
@template('y', types=['int'])
@template('z', types=['int', 'real'])
@types('y', 'z')
def multi_tmplt_2(y, z):
    return y + z

multi_tmplt_2(1.5, 3.5)
```

EmilyBourne (changes requested):
Do we know if this code is definitely dead now: https://github.com/pyccel/pyccel/pull/530/files#diff-9ba0c438140113a543a41e16000b939989e841d4fc6a6a4112b5c8e5d8807524R2719-R2730
?

The function for determining the types looks quite complicated. Can we add

pyccel/ast/core.py
```diff
@@ -3861,6 +3810,125 @@ def __str__(self):
...
+    Parameters
+    ----------
+    name : str
+        The name of the interface.
+
+    functions : iterable
+        The functions of the interface.
+
+    hide: bool
+        if True, the interface definition will not be generated.
```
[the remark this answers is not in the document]
rmahjoubi: I believe so, but it wasn't me who added this property

pyccel/ast/core.py
```diff
@@ -3861,6 +3810,125 @@ def __str__(self):
...
+
+    @property
+    def functions(self):
+        """"Functions of the interface."""
+        return self._functions
+
+    @property
+    def hide(self):
+        """If True, the interface definition will not be generated."""
+        return self._functions[0].hide or self._hide
```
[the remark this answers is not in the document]
rmahjoubi: I'm  not sure, this code was in the class Interface before

pyccel/ast/core.py
```diff
@@ -3861,6 +3810,125 @@ def __str__(self):
...
+
+    @property
+    def cls_name(self):
+        """Class name if the interface is a method of cls_name."""
+        return self._functions[0].cls_name
+
+    @property
+    def kind(self):
+        """ 'function' or 'procedure' """
+        return self._functions[0].kind
```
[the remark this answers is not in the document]
rmahjoubi: Yes I agree this shouldn't be `function or procedure`. this code was here before I didn't know if I should remove

pyccel/ast/core.py
```diff
@@ -3861,6 +3810,125 @@ def __str__(self):
...
+
+    @property
+    def decorators(self):
+        """a list of properties."""
+        return self._functions[0].decorators
+
+    @property
+    def is_procedure(self):
+        """True if the Interface is a procedure."""
+        return self._functions[0].is_procedure
```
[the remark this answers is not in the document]
rmahjoubi: I don't think I know the unswer. this code was in this class before I didn't know what it's used for so I left it there

pyccel/codegen/printing/pycode.py
```diff
@@ -78,19 +78,24 @@ def _print_FunctionDef(self, expr):
...
-                    dec = '@{name}'.format(name=n)
-
-                code = '{dec}\n{code}'.format(dec=dec, code=code)
-        header = expr.header
-        if header:
-            header = self._print(header)
+            for n,f in decorators.items():
+                # TODO - All decorators must be stored in a list
+                if not isinstance(f, list):
+                    f = [f]
```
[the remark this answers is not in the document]
rmahjoubi: I just added this TODO, I believe we should allow for duplicated decorators of all kinds, is this suitable ?

pyccel/parser/syntactic.py
rmahjoubi: This is if the user gave the wrong keyword to the types decorators, we raise an error for this

pyccel/ast/core.py
```diff
@@ -3861,6 +3810,125 @@ def __str__(self):
...
+
+    @property
+    def is_argument(self):
+        """True if the interface is used for a function argument."""
+        return self._is_argument
+
+    @property
+    def global_vars(self):
+        """Variables wich will not be passed to the interface."""
+        return self._functions[0].global_vars
```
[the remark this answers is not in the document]
rmahjoubi: I agree with you, this code was here already I didn't add it, should I remove it?

pyccel/codegen/printing/pycode.py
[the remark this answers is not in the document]
EmilyBourne: Ah, that sounds like it makes sense. Maybe you can do the TODO already

pyccel/ast/core.py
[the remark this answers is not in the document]
EmilyBourne: @saidctb ?

pyccel/parser/syntactic.py
EmilyBourne: Ah sorry, I didn't read this properly, it's the opposite to what I thought

EmilyBourne:
Don't forget to merge the master

pyccel/ast/core.py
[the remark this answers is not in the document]
EmilyBourne: I'm not sure. Is it used anywhere?

pyccel/ast/core.py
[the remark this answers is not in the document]
saidctb: this means that the interface name is not gonna be used but rather the specific function name , it doesn't make a lot of sense now after adding C

pyccel/ast/core.py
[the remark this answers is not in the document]
saidctb: this means the global variables that are used in the interface or function

pyccel/ast/core.py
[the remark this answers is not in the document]
saidctb: I'm not sure of this stuff are used anymore in the FuctionDef or the printer to determine if a function is a procedure or not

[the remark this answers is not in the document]
EmilyBourne: Hmm, sounds like we might need an extra issue to clean up the `FunctionDef` class and get rid of all the functions we're not using anymore

pyccel/codegen/printing/pycode.py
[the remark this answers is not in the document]
rmahjoubi: This requires changing all the decorators we use, I can create an issue for this?

pyccel/ast/core.py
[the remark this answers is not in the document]
EmilyBourne: I can't think of a global variable that would be needed for an interface. Presumably there's no reason for the `global_vars` of all functions in an interface to be the same?

pyccel/ast/core.py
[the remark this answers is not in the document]
EmilyBourne: So then maybe we should remove it?

pyccel/codegen/printing/pycode.py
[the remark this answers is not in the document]
EmilyBourne: Ah ok, sorry I assumed it was simpler than that. In that case creating an issue sounds like a good idea

yguclu:
Given that we have recently started worrying about our (non-existent) documentation, I was thinking that this new development in our headers definitely deserves a bit of user documentation itself. So it's probably time to start!

@rmahjoubi Do you think you could provide a short markdown file where you show a few examples and explain how they work? This will later become part of our user documentation, or a tutorial.

pyccel/ast/core.py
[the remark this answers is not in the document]
yguclu: @EmilyBourne I don't know why this whole class is shown as being modified, I think git is just confused. @rmahjoubi made some changes to class `Interface`, but he did not rewrite it completely.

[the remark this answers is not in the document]
EmilyBourne: Ah ok. I think he moved it in the file which is probably why it's all shown up as new. So I wasn't sure what was new and what wasn't or if it was just a rewrite. But I thought this might be a good opportunity to look at the contents of the class as we are unlikely to do it later

pyccel/ast/core.py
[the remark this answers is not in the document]
yguclu: More general question: do we still want to print an `interface` block in the generated Fortran code, or not? It's not clear to me how we can actually expose the Fortran interface to the calling C code.

pyccel/ast/headers.py
```diff
@@ -89,11 +89,33 @@ def __getnewargs__(self):
...
+    def args(self):
+        "Types the template represents."
+        return self._args
+
 #==============================================================================
 class FunctionHeader(Header):
     """Represents function/subroutine header in the code.
 
-    func: str
+    name: str
```
yguclu, line 118: Can you please add the line
```
Parameters
----------
```
to this docstring, just before listing all the arguments of the class constructor?

pyccel/ast/core.py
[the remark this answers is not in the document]
EmilyBourne: We can't expose the interface itself to C. But we can still use it in the fortran code, and if the user wishes to use the generated Fortran code elsewhere then they can still find their interfaced function

pyccel/ast/headers.py
```diff
@@ -89,11 +89,33 @@ def __getnewargs__(self):
         args = (self._args[0],)
         return args
 
+#==============================================================================
+class Template(Header):
+    """Represents a Template."""
```
yguclu, line 94: Could you please expand this docstring a bit? I think it needs a "Parameters" section, and an "Examples" section would also be very useful!

pyccel/ast/core.py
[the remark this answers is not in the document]
rmahjoubi: Is there an advantage in using interfaces rather than the actual functions in Fortran?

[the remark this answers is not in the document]
EmilyBourne: Maybe readability?

pyccel/ast/core.py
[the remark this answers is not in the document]
saidctb: there is no reason when we add the multiple dispatch case, so we can remove it

pyccel/ast/headers.py
```diff
@@ -212,7 +234,31 @@ def build_argument(var_name, dc):
...
+                else:
+                    iterx[i] = [j]
+            #generate combinations using product, and replace remaining
+            #templates with its appropriates values
+            for itery in product(*iterx):
+                itery = list(itery)
+                for j, i in enumerate(itery):
+                    if 'datatype' in i and i['datatype'] in templates:
+                        itery[j] = itery[old_values[i['datatype']]].copy()
+                arg_codes.append(itery)
```
yguclu, line 260: I find this code difficult to understand, probably because of the choice of variables' names. I recommend using `i` and `j` for the integer indices, but not for the actual elements of an iterable container.

It also seems to me that the templates could be processed one by one, until they are all used. I am giving here some pseudo-code to explain what I mean:
```python
# First, process unions
...

# Next, process templates
arg_codes = (*dtypes)
for t in templates:
    arg_codes = (replace_template(c, t, a) for c in arg_codes for a in t.args)
```
Where function `replace_template` creates a new function signature by replacing a certain template `T` with one of its possible values. Please let me know if I am missing something here!

pyccel/codegen/printing/ccode.py
```diff
@@ -685,6 +685,8 @@ def _print_Rand(self, expr):
...
+        return ""
     def _print_FunctionDef(self, expr):
```
yguclu, line 690: Are we not missing an empty line before the function definition?

pyccel/codegen/printing/pycode.py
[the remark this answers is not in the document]
yguclu: I agree, please create an issue with a clear description and some examples.

pyccel/parser/base.py
```diff
@@ -127,6 +127,7 @@ def __init__(self, *, headers=None, decorators=None):
         self._classes   = OrderedDict()
         self._functions = OrderedDict()
         self._macros    = OrderedDict()
+        self._templates = {}
```
yguclu, line 130: What is the reason for now using an ordered dictionary like for every other attributes?

pyccel/parser/semantic.py
yguclu: I think the magic method `__iadd__` of a list just calls `extend`, so there should be no difference at all. I personally like to use `+` and `+=` for list concatenation.

pyccel/parser/semantic.py
yguclu: Please create a new issue if this does not work.

pyccel/parser/syntactic.py
yguclu: A Pythonic way to write this loop is:
```python
for arg in ls:
    ...
```

pyccel/parser/syntactic.py
yguclu: I also find this hard to read, because the index `i` starts from `0` and is either incremented by `1` or set to `-1`...

pyccel/codegen/printing/pycode.py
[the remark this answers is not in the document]
rmahjoubi: Ok I will create it

pyccel/parser/semantic.py
rmahjoubi: Yes I will create an issue for it

pyccel/parser/syntactic.py
rmahjoubi: I will try to change it

pyccel/parser/syntactic.py
```diff
@@ -621,8 +623,16 @@ def _visit_FunctionDef(self, stmt):
         imports      = []
 
         # TODO improve later
```
yguclu, line 625: I don't know what this TODO actually means.

@rmahjoubi If you think that your changes are a sufficient improvement here, I suggest you remove the TODO comment.

pyccel/ast/core.py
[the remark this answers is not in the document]
rmahjoubi: Ok I will remove it

pyccel/ast/core.py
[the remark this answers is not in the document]
rmahjoubi: I will create an issue for this

pyccel/parser/semantic.py
rmahjoubi: I will change it

tests/codegen/fcode/scripts/generic_methods.py
rmahjoubi: I thing there were two errors in this file, only one is related to interfaces, I am not sure I going to test it

pyccel/ast/headers.py
rmahjoubi: I think this is a better way to do it, I will try to improve this code section

pyccel/parser/base.py
rmahjoubi: yes this should be an ordered dictionary as well

rmahjoubi:
> @rmahjoubi Do you think you could provide a short markdown file where you show a few examples and explain how they work? This will later become part of our user documentation, or a tutorial.
I was thinking this myself, I will add a markdown file where I explain how these headers works

tests/codegen/fcode/scripts/generic_methods.py
EmilyBourne: That is very possible. In which case we should have 2 copies of this file. One (the one you have now) to test interfaces in functions and one (the old one, still in the errors folder) to show the other problem

pyccel/decorators.py
EmilyBourne: Actually epyccel doesn't work with header comments. Originally pyccel only used header comments, the decorators were added later in order to allow the user to epyccelize a function. So:
```
#$ header function f2(int)
#$ header function f1(int)
def f2(a):
    pass
def f1(a):
    pass
```
and
```
#$ header template T(int|real)
#$ header template U(int|real)
#$ header function f(T, U)
def f(a, b):
   pass
```
are both fine, but mean you have to use `pyccel` or epyccelize the whole module. This is the difference between decorators and header comments

> As for the decorators I presumed if the user decorated a function with a template then he wanted it to be used only for that function, so it's not available for other functions. 

That makes sense to me

> I checked the ast for this, the information is preserved.
Ok, then we should be fine

tests/errors/semantic/blocking/INTERFACE_WRONG_TYPES.py
EmilyBourne: By similar I mean also checking we get an error. I just want to make sure that the code correctly restricts the types of y

rmahjoubi: oh, yes now I see. so I should add a test in errors

EmilyBourne: Yes please :)

rmahjoubi: can you check please

EmilyBourne: Looks good thanks. You can resolve this conversation now

rmahjoubi: Ok great

pyccel/codegen/printing/pycode.py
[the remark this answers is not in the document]
rmahjoubi: Issue #549 is created for this.

pyccel/ast/core.py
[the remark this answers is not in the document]
rmahjoubi: Issue #550 is created for this.

pyccel/parser/semantic.py
rmahjoubi: Issue #551 is created for this

tests/codegen/fcode/scripts/generic_methods.py
rmahjoubi: Sorry where should I keep the one with the error that I fixed, do you mean I should mark it as xfail?

EmilyBourne: In the master branch we have the file `tests/errors/known_bugs/generic_methods.py`. It is in the errors folder so this tells us that it contains something that doesn't work but did at some point.
In this branch 1 of the 2 problems in that file are fixed. It therefore makes sense for you to create the new file `tests/codegen/fcode/scripts/generic_methods.py` so this doesn't get broken again. However if you completely remove the code for the second problem then we lose the fact that that bit should work.

So you should have 2 versions of the code.
1. The file `tests/errors/known_bugs/generic_methods.py` from the master branch in the same location which will illustrate the second problem
2. The file `tests/codegen/fcode/scripts/generic_methods.py` from this branch which shows that the first problem has been fixed

rmahjoubi: Yes, I will create the second file, for the first one is a newly introduced bug, when a class has a generic method, I will try to fix it.

rmahjoubi: Actually I already created a file tests/codegen/fcode/scripts/generic_methods.py, that shows the fixed problem.
The second problem occurs when we call the method.

EmilyBourne: You didn't create a file you moved a file and then removed some information. Creating a file to show the fixed problem is good. I just don't want to lose the original file that shows the method calling problem

pyccel/ast/core.py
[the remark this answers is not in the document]
rmahjoubi: I removed all unused methods in the class Interface

pyccel/ast/core.py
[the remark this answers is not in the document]
rmahjoubi: I removed all unused methods in this class

pyccel/ast/core.py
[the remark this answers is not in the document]
rmahjoubi: I removed it

pyccel/parser/syntactic.py
rmahjoubi: There's still the issue about storing duplicated decorators

EmilyBourne: A TODO comment should contain enough information for another developer to understand what needs doing. Either this comment needs more info. Or it should just refer to the issue. Or it should be removed and the issue is itself sufficient

rmahjoubi: Ok, I will remove it

pyccel/parser/semantic.py
EmilyBourne: @saidctb , this seems to have got lost. Do you know if we need to look for headers in multiple scopes? And if so, why?

pyccel/ast/core.py
```diff
@@ -3861,6 +3810,90 @@ def __str__(self):
...
+        is_argument = False,
+        ):
+
+        if not isinstance(name, str):
+            raise TypeError('Expecting an str')
+        if not isinstance(functions, list):
+            raise TypeError('Expecting a list')
+        self._name = name
+        self._functions = functions
+        self._hide = hide
```
EmilyBourne, line 3855: Is there any point keeping this variable if you removed the access function?

tests/codegen/fcode/scripts/generic_methods.py
rmahjoubi: Yes I see, I will try to fix the problem and then add file as it was in tests/codegen

pyccel/codegen/printing/fcode.py
```diff
@@ -296,10 +296,7 @@ def _print_Module(self, expr):
         sep = self._print(SeparatorComment(40))
         interfaces = ''
         if expr.interfaces:
-            interfaces = '\n'.join(self._print(i) for i in expr.interfaces if not i.hide)
-            for interface in expr.interfaces:
-                if not interface.hide:
-                    body += '\n'.join(''.join([sep, self._print(i), sep]) for i in interface.functions)
+            interfaces = '\n'.join(self._print(i) for i in expr.interfaces)
```
EmilyBourne, line 299: We seem to have lost the insertion of the interfaces into the body

tests/codegen/fcode/scripts/generic_methods.py
EmilyBourne: If you can fix the problem then great. If not then just keep this and put the original back in errors

pyccel/ast/core.py
rmahjoubi: I forgot it, I will remove it, thanks

EmilyBourne: Don't forget the docstring

pyccel/parser/syntactic.py
```diff
@@ -633,54 +644,126 @@ def _visit_FunctionDef(self, stmt):
...
+                    msg = 'Argument provided to the template decorator is not valid'
                     errors.report(msg,
-                                  symbol = decorators['types'],
-                                  bounding_box = (stmt.lineno, stmt.col_offset),
-                                  severity='error')
+                                    symbol = comb_types,
+                                    bounding_box = (stmt.lineno, stmt.col_offset),
+                                    severity='fatal')
+
+                i = 0
```
EmilyBourne, line 686: Unused variable?

pyccel/codegen/printing/fcode.py
rmahjoubi: I print the interfaces previous to the `contains` statement, that's why I have body and interfaces separated

EmilyBourne: Found it! Thanks :+1:

pyccel/parser/syntactic.py
rmahjoubi: yes, it didn't get detected, I just removed it

EmilyBourne: Probably there is a variable with the same name in the same scope which is used

pyccel/ast/core.py
rmahjoubi: Yes I removed it too

pyccel/ast/headers.py
```diff
@@ -268,19 +269,11 @@ def template_resolve(arg_code, Tname, typ):
...
-                        arg_codes.append(tmp_codes)
-                if len(arg_codes) > leng:
-                    arg_codes.pop(0)
-                else:
-                    arg_codes.append(arg_codes.pop(0))
+        for tmplt in templates:
+            arg_codes = [process_template(arg_code, tmplt, d_type, j) for arg_code in arg_codes\
+                    for j, d_type, in enumerate(templates[tmplt].args) if arg_code]
+        if len(arg_codes) > 1:
+            arg_codes = list(filter(None, arg_codes))
```
yguclu, line 276: I don't fully understand the following points:

1. Why we need to pass `type_index` to `process_template`;
2. Why `process_template` is allowed to return `None`;

The need to filter out any `None` makes the code more cumbersome. If some templates are not present in the original `arg_codes` list, surely they can be removed from the `templates` list before proceeding any further?

yguclu (changes requested):

pyccel/ast/headers.py
rmahjoubi: If we have `arg_codes ==  [[int, int]]` for example, we are going to parse arg_code[0] the number of types in each template. we can chose to return [int, int] from within `process_template` without touching it, but this will result in duplication of [int,int] the number of types in each template. or we can return None, but we will lose [int, int]. that's is why we return it only once if we didn't touch it, and for the rest we only return None.
In the following case when we have: `arg_codes == [[int, int], [T, int]]`, we can not remove the template T from templates, because it's used in arg_codes, but still what is stated before will occur for [int, int].
This will happen also if are processing template R for `arg_codes == [[int, T]]`.
I can retrieve all the `arg_code in arg_codes` that will not be processed, then add them the next time to arg_codes, and retrieve what's will not be processed and so on... but I don't think this will do any better

yguclu: So here you are essentially dealing with the situation where `arg1 = int | T`, which is a mix between a union and a template...

yguclu: Thanks @rmahjoubi! I cannot see a clear way to further improve this algorithm, unless we put some restrictions on the way the user can combine multiple `@types` headers, unions and templates.

I still think that it is useful to restructure the code a bit, in order to improve its future maintainability. For example, one could apply the original algorithm I had suggested to each of the signatures (one for each instance of @types), after adding a function that returns only the templates that are actually present in a signature!
```python
def find_templates(signature, available_templates):
    ''' Return a list of Template objects present in the given function signature. '''

def process_template(signature, template_name, d_type):
    ''' Create a new signature where the given template is substituted with a specific datatype. '''

new_signatures = []
for signature in signatures:
    tmp_signatures = (signature,)
    for t in find_templates(signature, templates):
        tmp_signatures = tuple(process_template(s, t.name, d_type) for s in tmp_signatures for d_type in t.args)
    new_signatures += tmp_signatures
```

This pseudo-code is maybe not complete, as one needs to determine what the type of `t` is supposed to be.

rmahjoubi: Thank you, that was very helpful. I added the restrictions, now we raise an error if a template is in a uniontype. The case of multiple headers for a function works fine, since we treat each header at a time. I went back to what you suggested before.
could you check please.

pyccel/ast/headers.py
```diff
@@ -14,6 +14,8 @@
 from .datatypes import datatype, DataTypeFactory, UnionType
 from .macros import Macro, MacroShape, construct_macro
 from .core import local_sympify
+from pyccel.errors.errors import Errors
```
yguclu, line 17: Codacy wants us to place absolute imports before relative imports. I think that makes sense, and it's probably recommended by some Python PEP out there. Could you please reorder them?

pyccel/ast/headers.py
```diff
@@ -247,20 +252,23 @@ def build_argument(var_name, dc):
...
-                    resolved = True
-                    tmp_arg_code[i] = d_type.copy()
-            if resolved or type_index == 0:
-                return tmp_arg_code
-            return None
+        def find_templates(signature, templates):
+            new_templates = {d_type['datatype']:templates[d_type['datatype']]\
+                    for d_type in signature\
+                    if 'datatype' in d_type and d_type['datatype'] in templates}
+            return new_templates
```
yguclu, line 264: Even though this is a local function, I think it's useful to add a one-line docstring to it.

pyccel/ast/headers.py
```diff
@@ -247,20 +252,23 @@ def build_argument(var_name, dc):
                         is_argument=True)
             return var
 
-        def process_template(arg_code, Tname, d_type, type_index):
-            tmp_arg_code = arg_code.copy()
-            resolved = False
+        def process_template(signature, Tname, d_type):
+            new_sig = tuple(d_type if 'datatype' in t and t['datatype'] == Tname\
+                    else t for t in signature)
+            return new_sig
```
yguclu, line 258: Even though this is a local function, I think it's useful to add a one-line docstring to it.

pyccel/ast/headers.py
```diff
@@ -269,14 +277,14 @@ def process_template(arg_code, Tname, d_type, type_index):
...
-            arg_codes = [process_template(arg_code, tmplt, d_type, j) for arg_code in arg_codes\
-                    for j, d_type, in enumerate(templates[tmplt].args) if arg_code]
-        if len(arg_codes) > 1:
-            arg_codes = list(filter(None, arg_codes))
+        signatures = list(product(*dtypes))
+        new_templates = find_templates(signatures[0], templates)
+
+        for tmplt in new_templates:
+            signatures = [process_template(s, tmplt, d_type)\
+                    for s in signatures for d_type in new_templates[tmplt].args]
```
yguclu, line 285: This looks very concise and elegant now!

I suggest you make `signatures` a tuple instead of a list, because a `tuple` is an immutable object (it has no `__setitem__` method) and hence we make it clear that we never change it in our algorithm. What we do is to define a **new** tuple at each loop cycle instead.

pyccel/errors/messages.py
```diff
@@ -129,3 +129,4 @@
 FOUND_IS_IN_ASSIGN = 'Found `is` statement in assignment [Ignored]'
 ARRAY_REALLOCATION = 'Array redefinition may cause memory reallocation at runtime'
 ARRAY_DEFINITION_IN_LOOP = 'Array definition in for loop may cause memory reallocation at each cycle. Consider creating the array before the loop'
+TEMPLATE_IN_UNIONTYPE = 'Cannot use templates in a union type'
```
yguclu, line 132: Since allowing templates inside unions makes the `@types` annotations quite confusing, I totally agree with not allowing it at all!

@EmilyBourne, @saidctb Do you also agree with me and @rmahjoubi?

yguclu (changes requested):

pyccel/errors/messages.py
EmilyBourne: I agree, but I would like a more verbose error message. I am not certain that a user who wasn't involved in the technical discussions would know what you mean by a `union type`. Maybe:
```
Only one header may be provided when using template types
```
?

pyccel/ast/headers.py
EmilyBourne: You can also change this line to `from ..errors.errors import Errors` if you prefer, but I suspect it will still want it reordered

pyccel/errors/messages.py
yguclu: By union we mean `@types('int | float')` I think, so we don't want to allow the following:
```python
@template('T', types=['float', 'complex'])
@types('int | T')
```

EmilyBourne: Ah, that makes even more sense. I didn't realise we'd kept union types. Don't templates do the same job?

yguclu: I also did not realize that at the beginning! :sweat_smile: That's why @rmahjoubi and I had a hard time understanding each other! :rofl: Unions cover only part of what a template can do, because they only allow tensor products.

yguclu: For functions that only require a single template for a single argument, a union is more handy to write. But in all other cases templates are much better, being more explicit and compact at the same time.

saidctb: how about something like this ?
```
@template('T', types=['float', 'complex'])
@types('T[;,:,:] | T[:.:]')
```

yguclu: I cannot imagine many functions that would work for 2D and 3D arrays. Unless they are elemental

yguclu: @rmahjoubi Is the following signature allowed right now?
```python
@template('T', types=['float', 'complex'])
@types('T[:,:]')
```
or does it have to be like this?
```python
@templates('T', types=['float[:,:]', 'complex[:,:]')
@types('T')
```

yguclu (approved):

pyccel/ast/headers.py
```diff
@@ -204,15 +252,41 @@ def build_argument(var_name, dc):
...
+            new_templates = {d_type['datatype']:templates[d_type['datatype']]\
+                    for d_type in signature\
+                    if 'datatype' in d_type and d_type['datatype'] in templates}
+            return new_templates
 
         for i in self.dtypes:
             if isinstance(i, UnionType):
+                for d_type in i.args:
+                    if d_type['datatype'] in templates:
+                        errors.report(TEMPLATE_IN_UNIONTYPE,
```
EmilyBourne, line 272: Do we have an object that we could provide as a symbol here to help the user identify which line is causing the error?

pyccel/errors/messages.py
saidctb: you can easily have one that computes the sum for example , so all vectorized operations + reduction

pyccel/parser/semantic.py
```diff
@@ -1023,7 +1042,9 @@ def _visit_DottedVariable(self, expr, **settings):
 
         # look for a class method
         if isinstance(expr.rhs, Application):
-
+            methods = list(first.cls_base.methods) + list(first.cls_base.interfaces)
+            if any(isinstance(method, Interface) for method in methods):
+                errors.report('Generic methods are not supported yet', severity='fatal')
```
EmilyBourne, line 1047: Can we provide the method as a symbol?

pyccel/parser/semantic.py
```diff
@@ -1048,7 +1068,9 @@ def _visit_DottedVariable(self, expr, **settings):
 
         # look for a class attribute / property
         elif isinstance(expr.rhs, Symbol) and first.cls_base:
-
+            methods = list(first.cls_base.methods) + list(first.cls_base.interfaces)
+            if any(isinstance(method, Interface) for method in methods):
+                errors.report('Generic methods are not supported yet', severity='fatal')
```
EmilyBourne, line 1073: See above `symbol = method`?

pyccel/parser/semantic.py
EmilyBourne: We should even have access to the line number in semantic with:
```
bounding_box=(self._current_fst_node.lineno, self._current_fst_node.col_offset)
```

EmilyBourne (changes requested):
Looks good to me, but would you mind making a couple of improvements to the `errors.report` calls. Ideally when we call `errors.report` we should provide either the line and column number, or a symbol, or both. This helps the user find the relevant bit of code. If you don't provide either then the only clue you have is the error message itself. Which is helpful but not necessarily enough if you have, for example, lots of templates and only one which is raising an error

pyccel/errors/messages.py
rmahjoubi: Right now It has to be like this:
```
@template('T', types=['float[:,:]', 'complex[:,:]'])
@types('T')
```

pyccel/ast/headers.py
rmahjoubi: can we use the name of the function which header causes the problem ?

EmilyBourne: That seems like it would be sufficient if you don't have the header object to print

rmahjoubi: I have it, but the problem the header contains dictionaries of types, which makes the error hard to read

EmilyBourne: Up to you then. It might be worth checking if the dictionary is printed when you print the header. If you think it's hard to read. The function should be sufficient

rmahjoubi: I checked, the dictionaries are printed, I will go with the function's name

pyccel/errors/messages.py
yguclu: > you can easily have one that computes the sum for example , so all vectorized operations + reduction

I you just compute the sum you have two options:
1. Use `numpy.sum()`, hence you don't need Pyccel;
2. Flatten the array first and then use 1D arrays, hence you don't need unions.

pyccel/parser/semantic.py
rmahjoubi: should I loop over the methods to get the one creating the problem, or the line number is enough?

pyccel/errors/messages.py
saidctb: you are thinking about a simple example, imagine having big code where different functions call each other and you need the sum , so using numpy is not an option, and flattening it also

EmilyBourne (approved):

pyccel/errors/messages.py
EmilyBourne: Why is using numpy not an option?

pyccel/parser/semantic.py
yguclu: Yes, I think you should loop over the methods. We should not worry about performance here, because we are going to raise an error anyway!

yguclu: Is this conversation resolved? So no loop to get the method?

rmahjoubi: I will add the loop

rmahjoubi: I think it's good now

yguclu (approved):

Issue #410, opened 2020-09-01
Add brackets when mixing `and` and `or` operations for stylistic reasons


nhamidn:
for this python code : 
```
if (x >= 0 and x <= 5) or x == 10:
        x = x + 1
```

the translated code was like this : 

```
if (x >= 0 && x <= 5 || x == 10)
    {
        x = x + 1;
    }
```

even if there is no brackets this code is working well, but it raises a warning when compiling the generated code in C, so we must add brackets when there are mixed Or and And.

----------------------------------------------------------

> It is working as intended for both. The intention is to add brackets when they are necessary in python.
> As I said above, we have 3 cases here:
> 1. `(x >= 0 and x <= 5) or x == 10`
> 2. `x >= 0 and (x <= 5 or x == 10)`
> 3. `x >= 0 and x <= 5 or x == 10`
> 
> Case 2 and 3 will give the same results in python as `and` operations  are always carried out before `or` operations when there are no brackets.
> The parsing step does not read the brackets (and we can't make it read them) so we add them back in where they are necessary. In case 2 they are not necessary. While in case 1 they are.
> 
> That said option 2 is a lot easier to understand than option 3. This is why C raises a warning when you don't use brackets. As a result it may be a good idea to always use brackets when mixing `and` and `or` operations for stylistic reasons. But this should be discussed in an issue. The only people who see messages in pull requests are people that participated in the Pull Request and it would be good to have other opinions too. Please copy this conversation into an issue

_Originally posted by @EmilyBourne in
https://github.com/pyccel/pyccel/pull/396#issuecomment-681101777_

EmilyBourne:
Similarly the following code:
```
@types('bool', 'int')
def compare_is_int(a, b):
    c = False
    if a is bool(b):
        c = True
    return c
```
generates
```
if (a == b != 0)
```
Which also raises the warning:
```
 warning: suggest parentheses around comparison in operand of ‘!=’
```

EmilyBourne:
> generates
> 
> ```
> if (a == b != 0)
> ```
> 
> Which also raises the warning:
> 
> ```
>  warning: suggest parentheses around comparison in operand of ‘!=’
> ```
> 
> Although this may just be due to missing parentheses in the `int` cast

An elegant way of fixing this would be to add a `precedence` property to the `Bool` class in `pyccel/ast/builtins.py`:
```
    @property
    def precedence(self):
        if isinstance(self.arg.dtype, NativeBool):
            return 17
        else:
[...]

PR #540, opened 2020-11-12
Generate C code for Numpy array operations

### LINKED ISSUE #508: Generate C code for Numpy array operations
**Description**

As a DEVELOPER, I need to translate Numpy array operations from Python to C, and by so doing use the Pyccel C standard library.

**Acceptance criteria**

- [ ] Copy C files to `__pyccel__` folder
- [x] Handle array creation
- [ ] Handle array deletion
- [x] Handle `__setitem__` with scalar and slice index
- [x] Handle `__getitem__` with scalar index
- [x] Handle `__getitem__` with slice index (creates array slice)
- [ ] Compile the generated code and the included C files
- [ ] Write unit tests

**Notes**
- This story is strictly about generating C code, not about calling it from Python

saniQQ:
- Add printing in codegen/printing/ccode.py for following classes:
     - `NumpyArray`
     -  `NumpyFull` (it handles `NumpyEmpty`, `NumpyZeros`, `NumpyOnes`)
- Add indexed variables of following types to C printer:
     - array views
     - scalars 
- Add more available types for C arrays: long, short int, float complex, bool

Closes #508


pyccel/codegen/printing/ccode.py
```diff
@@ -192,6 +197,16 @@
...
+arr_dtype_registry = {('real',8)    : 'double',
+                  ('real',4)    : 'float',
+                  ('complex',8) : 'cdouble',
+                  ('complex',4) : 'cfloat',
+                  ('int',4)     : 'int',
+                  ('int',8)     : 'long',
+                  ('int',2)     : 'sint',
+                  ('int',1)     : 'char',
+                  ('bool',4)    : 'bool'}
+
```
EmilyBourne, line 209: Why are we using different data types for the arrays?

pyccel/codegen/printing/ccode.py
```diff
@@ -586,6 +602,68 @@ def function_signature(self, expr):
...
 
+    def _print_IndexedElement(self, expr):
+        if isinstance(expr.base, IndexedVariable):
+            base = expr.base.internal_variable
+        else:
+            base = expr.base
+        base = self._print(expr.base.label)
+        inds = list(expr.indices)
+        inds = inds[::-1]
+        base_shape = Shape(expr.base)
```
EmilyBourne, line 613: Is there any reason why you chose to use `Shape(expr.base)` here instead of `expr.base.shape`?

pyccel/codegen/printing/ccode.py
```diff
@@ -586,6 +602,68 @@ def function_signature(self, expr):
         else:
             return '{0}{1}({2})'.format(ret_type, name, arg_code)
 
+    def _print_IndexedElement(self, expr):
+        if isinstance(expr.base, IndexedVariable):
+            base = expr.base.internal_variable
+        else:
+            base = expr.base
```
EmilyBourne, line 609: This is not used but I think it should save you some checks later if you did use it

pyccel/codegen/printing/ccode.py
```diff
@@ -586,6 +602,68 @@ def function_signature(self, expr):
...
+        base = self._print(expr.base.label)
+        inds = list(expr.indices)
+        inds = inds[::-1]
+        base_shape = Shape(expr.base)
+        allow_negative_indexes = (isinstance(expr.base, IndexedVariable) and \
+                expr.base.internal_variable.allows_negative_indexes)
```
EmilyBourne, line 615: This is simpler if you use your `base` variable

pyccel/codegen/printing/ccode.py
```diff
@@ -586,6 +602,68 @@ def function_signature(self, expr):
...
+            else:
+                if isinstance(ind, Slice):
+                    #setting the slice start and end to their correct value if none is provided
+                    start = ind.start
+                    end = ind.end
+                    if ind.start is None:
+                        start = 0
+                    if ind.end is None:
+                        end = expr.base.shape[i]
+                    inds[i] = Slice.__new__(Slice, start, end)
```
EmilyBourne, line 628: You don't need to call the `__new__` method explicitly. And you probably shouldn't. I don't know if the `__init__` method is called if you do this. To declare an instance of a class you should just do:
```
inds[i] = Slice(start, end)
```

pyccel/codegen/printing/ccode.py
```diff
@@ -586,6 +602,68 @@ def function_signature(self, expr):
...
+                #indices of indexedElement of len==1 shouldn't be a Tuple
+                if isinstance(ind, Tuple) and len(ind) == 1:
+                    inds[i].args = ind[0]
+                if allow_negative_indexes and not isinstance(ind, Integer):
+                    inds[i] = PyccelMod(ind, base_shape[i])
+        inds = [self._print(i) for i in inds]
+        #set dtype to the C struct types
+        dtype = self.find_in_dtype_registry(format(expr.dtype), expr.precision, array=True)
```
EmilyBourne, line 636: What is the `format` function doing here?

pyccel/codegen/printing/ccode.py
```diff
@@ -586,6 +602,68 @@ def function_signature(self, expr):
...
+        if expr.rank > 0:
+            return "array_slicing(%s, %s)" % (base, ", ".join(inds))
+        return "%s.nd_%s[get_index(%s, %s)]" % (base, dtype, base, ", ".join(inds))
+
+    def _print_Allocate(self, expr):
+        free_code = ''
+        #free the array if its already allocated
+        if  (expr.status != 'unallocated'):
+            free_code = 'if (%s.raw_data != NULL)\n' % self._print(expr.variable.name)
```
EmilyBourne, line 645: This `if` shouldn't be needed if the allocation status is `allocated`
I don't know if it can currently have this status, but it is planned for the future

pyccel/codegen/printing/ccode.py
```diff
@@ -586,6 +602,68 @@ def function_signature(self, expr):
...
+            free_code += '{\nfree_array(%s);\n}\n' % self._print(expr.variable.name)
+        self._additional_imports.add('ndarrays')
+        shape = ", ".join(str(a) for a in expr.shape)
+        dtype = self.find_in_dtype_registry(format(expr.variable.dtype), expr.variable.precision, array=True)
+        shape_name = "shape_dummy_%s" % format(expr.__hash__()%991)
```
EmilyBourne, line 650: Please use the `parser.get_new_name` function (I might have the wrong method name, it's defined in pyccel.parser.base) to get a new name which avoids name clashes

pyccel/codegen/printing/ccode.py
```diff
@@ -586,6 +602,68 @@ def function_signature(self, expr):
...
+        init_shape = "int %s[] = {%s};" % (shape_name, shape)
+        alloc_code = "{} = array_create({}, {}, nd_{});".format(expr.variable, len(expr.shape), shape_name, dtype)
+        return '{}{}\n{}'.format(free_code, init_shape, alloc_code)
+
+    def _print_Slice(self, expr):
+        # print(expr.args)
+        if expr.start is None or  isinstance(expr.start, Nil):
+            start = '-1'
```
EmilyBourne, line 658: Why not 0?

pyccel/codegen/printing/ccode.py
```diff
@@ -586,6 +602,68 @@ def function_signature(self, expr):
...
+        else:
+            start = self._print(expr.start)
+        if (expr.end is None) or isinstance(expr.end, Nil):
+            end = '-1'
+        else:
+            end = self._print(expr.end)
+        return 'new_slice({}, {}, {})'.format(start, end, 1)
```
EmilyBourne, line 665: Why do you not collect the slice step?

pyccel/codegen/printing/ccode.py
```diff
@@ -853,11 +931,34 @@ def _print_AugAssign(self, expr):
         return "{0} {1}= {2};".format(lhs_code, op, rhs_code)
 
     def _print_Assign(self, expr):
+        lhs = self._print(expr.lhs)
+        rhs = expr.rhs
+        if isinstance(rhs, (NumpyArray)):
+            dummy_array_name = "dummy_%s" % (rhs.__hash__()%991)
```
EmilyBourne, line 937: See above use standard function to avoid name clashes

pyccel/codegen/printing/ccode.py
```diff
@@ -853,11 +931,34 @@ def _print_AugAssign(self, expr):
...
+            dtype = self.find_in_dtype_registry(format(rhs.dtype), rhs.precision)
+            arg = rhs.arg
+            if rhs.rank > 1:
+                import functools
+                import operator
```
EmilyBourne, line 942: Imports go at the top of the file

pyccel/codegen/printing/ccode.py
```diff
@@ -853,11 +931,34 @@ def _print_AugAssign(self, expr):
...
+                arg = functools.reduce(operator.concat, arg)
```
EmilyBourne, line 943: What are you doing here?

pyccel/codegen/printing/ccode.py
```diff
@@ -853,11 +931,34 @@ def _print_AugAssign(self, expr):
...
+            dtype = self.find_in_dtype_registry(format(rhs.dtype), rhs.precision, array=True)
+            cpy_data = "memcpy({0}.nd_{2}, {1}, {0}.buffer_size);".format(lhs, dummy_array_name, dtype)
+            return  '%s\n%s\n' % (dummy_array, cpy_data)
+
+
+        if isinstance(rhs, (NumpyFull)):
+            code_init = ''
+            if rhs.fill_value is not None:
+                code_init = 'array_fill({0}, {1});'.format(self._print(rhs.fill_value), lhs)
+            if len(code_init) == 0:
```
EmilyBourne, line 954: Couldn't this just be `else`?

pyccel/codegen/printing/ccode.py
```diff
@@ -853,11 +931,34 @@ def _print_AugAssign(self, expr):
...
+            dummy_array_name = "dummy_%s" % (rhs.__hash__()%991)
+            dtype = self.find_in_dtype_registry(format(rhs.dtype), rhs.precision)
+            arg = rhs.arg
+            if rhs.rank > 1:
+                import functools
+                import operator
+                arg = functools.reduce(operator.concat, arg)
+            dummy_array = "%s %s[] = {%s};" % (dtype, dummy_array_name, ', '.join(self._print(i) for i in arg))
+            dtype = self.find_in_dtype_registry(format(rhs.dtype), rhs.precision, array=True)
+            cpy_data = "memcpy({0}.nd_{2}, {1}, {0}.buffer_size);".format(lhs, dummy_array_name, dtype)
```
EmilyBourne, line 946: I don't know if this is possible but can we do:
```
a = np.array([1,2,3])
b = np.array(a)
```
? If it works in numpy then does it work here?

pyccel/codegen/printing/ccode.py
EmilyBourne: I think I found why. Is this the dtype that you use in the array, that therefore cannot contain spaces? If so then why do none of these start with `nd_`?

EmilyBourne (changes requested):

pyccel/codegen/printing/ccode.py
saniQQ: i should add the nd_ i did it in the printing

EmilyBourne: Given that you never use these values without the `nd_` I don't see any reason to split it. And I think having it hear makes the difference clearer

pyccel/codegen/printing/ccode.py
saniQQ: i was passing -1 to check if the either the start or end are none (0 won't give me that ability) i did manage to handles it in line 620 this won't be needed anymore

EmilyBourne: Why do you need that ability?

pyccel/codegen/printing/ccode.py
saniQQ: either if its status is "allocated" or "unknown" we need to check if the pointer is not null then free it safely then allocate it without leaking the memory

pyccel/codegen/printing/ccode.py
saniQQ: its still not implemented in the slice class i will try to see if i can manage to add it

pyccel/codegen/printing/ccode.py
EmilyBourne: But if the status is `allocated` then the pointer can't be `null`. The `free` is still needed, just not the `if`

pyccel/codegen/printing/ccode.py
EmilyBourne: Ah sorry I didn't realise it wasn't implemented. In that case I would leave it as it is. We can add steps in a different pull request as we will need to add them in fortran too when we do add them

pyccel/codegen/printing/ccode.py
saniQQ: i needed it to return a string i will use self._print i think its better

pyccel/codegen/printing/ccode.py
saniQQ: i did copy it from the fcode i thought it will be useful to catch some checks

pyccel/codegen/printing/ccode.py
EmilyBourne: I don't think the `format` function is used anywhere in our code (although we often use `str.format`). I used ipython to get the docstring:
```
Signature: format(value, format_spec='', /)
Docstring:
Return value.__format__(format_spec)

format_spec defaults to the empty string.
See the Format Specification Mini-Language section of help('FORMATTING') for
details.
Type:      builtin_function_or_method
```
I have no idea what happens if you try to specify a format when you call this function on `expr.dtype`. If you don't specify a format then it defaults to `str(expr.dtype)`. If you just want to return a string, this is usually a clear way of doing that. Calling `str(a)` calls the method `a.__str__`. This method is implemented for datatypes and *should* return the same thing as `self._print(expr.dtype)` however there is nothing forcing these 2 to return the same thing. It is therefore better to use the same convention as in the rest of the code to be consistent and reduce the chance of breaking the behaviour in the future if we decide to add more information to the `DataType.__str__` method to help with debugging. In the rest of the code `str(expr.dtype)` is used for printing debugging methods and `self._print(expr.dtype)` is passed to the `self.find_in_dtype_registry` method

pyccel/codegen/printing/ccode.py
saniQQ: i thought the shape can be changed so i wanted to use Shape just in case because it return a PythonTuple

EmilyBourne: A tuple can't be changed. If you want to be able to change the shape then you should copy it into a list

pyccel/codegen/printing/ccode.py
saniQQ: thank for explaining i will change it

pyccel/codegen/printing/ccode.py
saniQQ: the `if` is to check if the pointer is **not** `null`

the `_print_Allocate` method is only called if the shapes is different 'that's when the status is unknown'  if the shapes are equal the free is not necessary

pyccel/codegen/printing/ccode.py
saniQQ: i was going to manage that in the C code but i am managing it from `_print_IndexedElement` now.
i just forget to remove the ifs

pyccel/codegen/printing/ccode.py
saniQQ: okey then i think i will change it to `expr.base.shape`

pyccel/codegen/printing/ccode.py
EmilyBourne: I agree. But if it **can't** be `null` then the check always returns `True` so you don't need the `if`, just the body of the `if`

pyccel/codegen/printing/ccode.py
EmilyBourne: I was just wondering what there is to manage. If you pass `None` to `start` then doesn't it default to `0`?

pyccel/codegen/printing/ccode.py
saniQQ: the "allocated" status will only be needed in the garbage collection step (from what i did understand) but in the case reallocation the status is "unknown" so we need to use `free` because we don't know if its allocated or not

pyccel/codegen/printing/ccode.py
saniQQ: ah, i just wanted to manage them in the same place

pyccel/codegen/printing/ccode.py
EmilyBourne: Ah not quite. Although currently that will often be true. Let me explain the logic

There are three possible states:
- `unallocated` - This variable has definitely not been allocated it is `null`
   Example 1:
   ```
   nd_array* a;
   // Code which does not use a
   ```
   a is `unallocated`
- `allocated` - This variable has definitely been allocated. It needs to be freed
   Example 2.
   ```
   nd_array* a = new_nd_array(...);
   ```
   a is `allocated`. If we want to allocate it with a different shape it definitely needs freeing. At the end of the function it definitely needs garbage collecting.
- `unknown` - We might have allocated this variable, but I'm not sure
   Example 3.
   ```
   nd_array* a;
   if (b) a = new_nd_array(...);
   ```
   a is `allocated` only if `b`. I do not know `b` so I do not know if a was `allocated`. Therefore the status of a is `unknown`.  If we want to allocate a with a different shape it might need freeing, or it might still be `null`. At the end of the function it might need garbage collecting or it might still be `null`.
   Example 4.
   ```
   nd_array* a;
   for (int i=0; i<n; ++i)
   {
        a = new_nd_array(...);
        // More code here
   }
   ```
   At the end of the for loop a is `allocated` if `n>0`. I don't know `n` therefore the status of a is `unknown`. As above it may or may not need freeing both for re-allocation and for garbage collection
   In addition, when `i=0` a is `unallocated`, but when `i=1` a is `allocated`. Therefore at the allocation line a may or may not be allocated. Therefore the status of a is `unknown` at this line

If you use these definitions your code should be right. The confusion is probably coming from the fact that currently we don't know if we are in an if block or a for loop so when we do:
```
 nd_array* a;
// Mystery code
a = new_nd_array(...);
```
we have no way of knowing what is in the mystery code. We might be in the case of Example 3 or 4. To guard against this we are currently setting the status to `unknown` even when it should be `allocated`. This is something we hope to fix at some point. As far as the printing is concerned though, you don't need to know if the logic was done correctly. You just check the allocation if it is `unknown`. Does that make sense?

pyccel/codegen/printing/ccode.py
EmilyBourne: I checked. That is the right function name. You can either use:
```
self.parser.get_new_name()
```
to get a new dummy name. Or you can use the `create_incremented_string` method from `ast.core` to get a similar naming scheme to what you are currently using.
In this case the call should be:
```
create_incremented_string(self.parser.used_names, prefix = 'shape_dummy')
```

saniQQ: thank you, will fix it

pyccel/codegen/printing/ccode.py
saniQQ: that's clears a lot of the confusions i had thank you

pyccel/codegen/printing/ccode.py
saniQQ: this is to flatten the array element into one list

pyccel/codegen/printing/ccode.py
```diff
@@ -8,19 +8,24 @@
...
+from pyccel.ast.core      import PyccelUnarySub, PyccelMod
 from pyccel.ast.datatypes import default_precision
 from pyccel.ast.datatypes import NativeInteger, NativeBool, NativeComplex, NativeReal, NativeTuple
 
 
 from pyccel.ast.numpyext import NumpyFloat
 from pyccel.ast.numpyext import NumpyReal, NumpyImag
+from pyccel.ast.numpyext import NumpyFull, NumpyArray
+from pyccel.ast.numpyext import Shape
+from pyccel.ast.numpyext import Tuple
```
EmilyBourne, line 23: Why are we importing `Tuple` from `pyccel.ast.numpyext` rather than directly from sympy?

saniQQ: is there a big difference with the 2 cases?

EmilyBourne: `Tuple` is defined in sympy. You can import it from `pyccel.ast.numpyext` because that file defines at this line:
https://github.com/pyccel/pyccel/blob/master/pyccel/ast/numpyext.py#L9
```
from sympy import Tuple
```
Technically there is no difference as you are importing the same object either way, but not importing it from where it is defined adds a step when trying to find the definition and dupes us into thinking that the `CCodePrinter` doesn't depend on sympy when it apparently does. We are slowly removing the sympy dependencies so hiding one is really unhelpful

EmilyBourne:
Don't forget to merge in the master branch

pyccel/codegen/printing/ccode.py
```diff
@@ -587,6 +603,67 @@ def function_signature(self, expr):
...
+        if isinstance(expr.base, IndexedVariable):
+            base = expr.base.internal_variable
+        else:
+            base = expr.base
+        inds = list(expr.indices)
+        inds = inds[::-1]
+        base_shape = base.shape
+        allow_negative_indexes = (isinstance(base, IndexedVariable) and \
+                base.internal_variable.allows_negative_indexes)
+        base = self._print(expr.base.label)
```
EmilyBourne, line 616: It would be better if you gave this variable a different name then you can keep using the `base` variable defined above
(Also shouldn't it be `self._print(base.label)`? )

pyccel/codegen/printing/ccode.py
```diff
@@ -587,6 +603,67 @@ def function_signature(self, expr):
...
+                inds[i] = PyccelMinus(base_shape[i], ind.args[0])
+            else:
+                if isinstance(ind, Slice):
+                    #setting the slice start and end to their correct value if none is provided
+                    start = ind.start
+                    end = ind.end
+                    if ind.start is None:
+                        start = 0
+                    if ind.end is None:
+                        end = expr.base.shape[1]
```
EmilyBourne, line 628: Beware of using the shape of an `IndexedVariable`. It seems to be defined as the shape of the underlying `IndexedBase`. `IndexedBase` is a sympy object to there may be data loss when saving this value. You would be better off using the base variable you defined above which contains the `internal_variable` if the object is an `IndexedVariable`

pyccel/codegen/printing/ccode.py
```diff
@@ -587,6 +603,67 @@ def function_signature(self, expr):
...
+                    inds[i].args = ind[0]
+                if allow_negative_indexes and not isinstance(ind, LiteralInteger):
+                    inds[i] = PyccelMod(ind, base_shape[i])
+        inds = [self._print(i) for i in inds]
+        #set dtype to the C struct types
+        dtype = self._print(expr.dtype)
+        dtype = self.find_in_dtype_registry(dtype, expr.precision, array=True)
+        if expr.rank > 0:
+            return "array_slicing(%s, %s)" % (base, ", ".join(inds))
+        return "%s.%s[get_index(%s, %s)]" % (base, dtype, base, ", ".join(inds))
```
EmilyBourne, line 641: Are we assuming that we are dealing with an array and never a tuple here?

pyccel/codegen/printing/ccode.py
```diff
@@ -513,8 +528,9 @@ def get_declare_type(self, expr):
         prec  = expr.precision
         rank  = expr.rank
         dtype = self.find_in_dtype_registry(dtype, prec)
-
         if rank > 0:
+            if expr.is_ndarray:
+                return 't_ndarray '
             errors.report(PYCCEL_RESTRICTION_TODO, symbol="rank > 0",severity='fatal')
```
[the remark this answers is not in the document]
saniQQ: yeah, the error for tuples is raised before but i did leave it there so we can separate the declarations of (tuple, lists) from numpy arrays

pyccel/codegen/printing/ccode.py
```diff
@@ -858,7 +934,30 @@ def _print_Assign(self, expr):
             self._temporary_args = [VariableAddress(a) for a in expr.lhs]
             return '{};'.format(self._print(expr.rhs))
         lhs = self._print(expr.lhs)
-        rhs = self._print(expr.rhs)
+        rhs = expr.rhs
+        if isinstance(rhs, (NumpyArray)):
+            if rhs.rank == 0:
+                raise NotImplementedError(expr.lhs + "=" + expr.rhs)
+            dummy_array_name, _ = create_incremented_string(self._parser.used_names, prefix = 'array_dummy')
+            dtype = self.find_in_dtype_registry(self._print(rhs.dtype), rhs.precision)
```
EmilyBourne, line 942: Why don't you need to pass `array = True`?

pyccel/codegen/printing/ccode.py
```diff
@@ -858,7 +934,30 @@ def _print_Assign(self, expr):
...
+            arg = rhs.arg
+            if rhs.rank > 1:
+                arg = functools.reduce(operator.concat, arg)
+            arg = ', '.join(self._print(i) for i in arg)
+            dummy_array = "%s %s[] = {%s};\n" % (dtype, dummy_array_name, arg)
+            dtype = self.find_in_dtype_registry(format(rhs.dtype), rhs.precision, array=True)
+            cpy_data = "memcpy({0}.nd_{2}, {1}, {0}.buffer_size);".format(lhs, dummy_array_name, dtype)
```
EmilyBourne, line 949: Do you still need the `nd_` here?

EmilyBourne (changes requested):
Can we activate some of the c tests now? E.g. some of the tests in tests/epyccel/recognised_functions/test_numpy_funcs.py

Or do we need to implement garbage collection first? Which issue is the garbage collection attached to?

pyccel/codegen/printing/ccode.py
saniQQ: here i need the C type to use it in the declaration

pyccel/codegen/printing/ccode.py
saniQQ: i need to remove it.

saniQQ:
> Can we activate some of the c tests now? E.g. some of the tests in tests/epyccel/recognised_functions/test_numpy_funcs.py
> 
> Or do we need to implement garbage collection first? Which issue is the garbage collection attached to?

we first need to be able to compile the C sources, and yes we need to implement garbage collection but we don't seem to have created an issue for that yet, in the #508 we have writing unit tests but i think before that wee need to be able to free our code

pyccel/codegen/printing/ccode.py
saniQQ: i assumed that since we are not handling the tuples yet.

pyccel/codegen/printing/ccode.py
saniQQ: fixed

pyccel/codegen/printing/ccode.py
saniQQ: i found an unexpected behaviour here when trying to assign a view

eg:
`a=np.array([[1,2,3],[4,5,6]])
b = a[0]`
so i need to handle this case first

pyccel/codegen/printing/ccode.py
EmilyBourne: Hmm, maybe we need 2 `find_in_dtype_registry` functions. Using the same name gives me the impression that they are getting an object for the same role, so I would assume that you pass `array = Ture` when you are handling an array.

In reality `array = True` indicates that you want the name of the array data member, not the dtype for the declaration

EmilyBourne:
> > Can we activate some of the c tests now? E.g. some of the tests in tests/epyccel/recognised_functions/test_numpy_funcs.py
> > Or do we need to implement garbage collection first? Which issue is the garbage collection attached to?
> 
> we first need to be able to compile the C sources, and yes we need to implement garbage collection but we don't seem to have created an issue for that yet, in the #508 we have writing unit tests but i think before that wee need to be able to free our code

Hmm ok. Then I suggest that we create a specific issue for correctly freeing the arrays

If we have no way of testing the results I would suggest creating a test in `tests/codegen/ccode/` . These tests run the codegen step, but don't test if the code compiles or gives correct results

EmilyBourne:
Is this PR still a work in progress?

saniQQ:
yes its still need the compiling part

pyccel/codegen/printing/ccode.py
EmilyBourne: What is the unuxepected behaviour?

EmilyBourne:
> yes it still need the compiling part

But it says "printing part only" in the PR name. I assumed that meant that the compilation was going in a different PR?

saniQQ:
> > yes it still need the compiling part
> 
> But it says "printing part only" in the PR name. I assumed that meant that the compilation was going in a different PR?

my bad i will change it to match the issue #508

saniQQ:
> If we have no way of testing the results I would suggest creating a test in `tests/codegen/ccode/` . These tests run the codegen step, but don't test if the code compiles or gives correct results

i will make the tests waiting for the compiling to be done

pyccel/codegen/printing/ccode.py
saniQQ: i will create another function to make it more clear

pyccel/codegen/printing/ccode.py
saniQQ: because in that case `base` is a `Variable` object and don't have a `label` attribute i just need to check some cases to _print them correctly

EmilyBourne: Ah. The `if`/`else` block creating the `base` variable is supposed to ensure that `base` is always a `Variable` I think. Maybe you just need to use `base.name`?

saniQQ: yes that's what i did

EmilyBourne:
We have now resolved all my comments, but you still have quite a lot of codacy errors. I will check that I can approve this branch once you have worked through the codacy errors (you can test this locally using pylint (pip install pylint) )

tests/codegen/ccode/scripts/arrays_create.py
```diff
@@ -1,5 +1,4 @@
-# pylint: disable=missing-function-docstring, missing-module-docstring/
-from pyccel.decorators import types
+# pylint: disable=missing-function-docstring, missing-module-docstring, unused-variable/
```
EmilyBourne, line 1: Please don't disable errors for an entire file. The only exception to this rule is disabling missing-docstring in test files

tests/codegen/ccode/scripts/arrays_slicing.py
```diff
@@ -1,4 +1,4 @@
-# pylint: disable=missing-function-docstring, missing-module-docstring/
+# pylint: disable=missing-function-docstring, missing-module-docstring, unused-variable, undefined-variable/
```
EmilyBourne, line 1: Here you can leave `unused-variable` disabled as there is no way to create a test which uses all variables with what is currently implemented

EmilyBourne: This is also true in `arrays_create`, but there should be no reason to disable `undefined-variable`

saniQQ: the undefined variables is for the pyccel decorators

tests/codegen/ccode/scripts/arrays_create.py
saniQQ: fixed

tests/codegen/ccode/scripts/arrays_slicing.py
```diff
@@ -0,0 +1,51 @@
...
+    a = 1
+    v = z[a:, 0:2] #pylint:disable=unused-variable
+
+def array_view_C_negative_indexes_literals():
+
+    from numpy import full
+    z = full((2, 5), 2)
+    v = z[-1:, 0:-3] #pylint:disable=unused-variable
+
+@allow_negative_index(y) #pylint:disable=undefined-variable
```
EmilyBourne, line 24: Why are you disabling this error? It is a very important error. You can see this by trying to execute this file in pure python mode. It does not run

tests/codegen/ccode/scripts/arrays_slicing.py
EmilyBourne: Search for examples of use of the decorator in other folders. Undefined-variable shouldn't be being raised

tests/codegen/ccode/scripts/arrays_slicing.py
saniQQ: i thought it was intended to let it work on pyccel

EmilyBourne: No. One of the main rules of pyccel is that the file must run in python in order to be an acceptable input to pyccel. Have a look in other files to see how the decorators are used

pyccel/codegen/printing/ccode.py
```diff
@@ -381,7 +397,7 @@ def _print_PyccelMod(self, expr):
         second = self._print(expr.args[1])
 
         if expr.dtype is NativeInteger():
-            return "{} % {}".format(first, second)
+            return "(({}) % {})".format(first, second)
```
EmilyBourne, line 400: Do you need all these parentheses? Does pyccel not add enough?
Do you have an example

saniQQ: if we have a python code as follows:
```
@allow_negative_index('a')
def array_negative_indexing():
      from numpy import array
      a = array([1,2,3])
      i = -1
      a[i] = a[i - 1]
```
the last line in C is translated to 
`a.nd_long[get_index(a, i % 3)] = a.nd_long[get_index(a, i - 1 % 3)];`

EmilyBourne: Normally the parentheses are added in the syntactic stage. You're solution leads to the following code:
```
a = (4-3)%2
```
being translated to:
```
a = (((4 - 3)) % 2);
```
which means there are too many parentheses and it's not particularly readable

I have had a look at where the parentheses are added and it is in the constructor of `PyccelOperator`:
https://github.com/pyccel/pyccel/blob/master/pyccel/ast/core.py#L342
Could you try moving the call to `handle_precedence` out of the `if` and see what happens for your case and the case above. This is linked to issue #410

saniQQ: sure i will try that

EmilyBourne: It seems to work for me. A similar operation should also be carried out in several other classes. But maybe we should leave this to be fixed with issue #410?

EmilyBourne (approved):

EmilyBourne:
> - arrays view (the syntax a[1:] is working but a[1] when its an array view is not i don't know if there is something to know if an array is a view or not so i can change the literals to a slice)

`a[1]` is an array view if there are slices in other dimensions. Can you not use this information to change the literal to a slice?

saniQQ:
> > * arrays view (the syntax a[1:] is working but a[1] when its an array view is not i don't know if there is something to know if an array is a view or not so i can change the literals to a slice)
> 
> `a[1]` is an array view if there are slices in other dimensions. Can you not use this information to change the literal to a slice?

i did fix that problem in the commit 5221ffa forgot to change the description

yguclu (approved):
@saniQQ Excellent job, thank you!
@EmilyBourne Thanks for the careful review of this PR!

yguclu:
About the indexing issue, it can be simplified to this: when we index a Numpy array `a` with a smaller number of indices/slices than `a.ndim`, we need to be very careful and follow Numpy's rules.

I think that this deserves a separate issue, where we collect all the information about Numpy's rules and then proceed with an implementation. For now it would be great if we could meet these two basic requirements:

1. Have the same Numpy support in the Fortran and C printers;
2. Raise a `NotImplementedError` whenever we find code that uses advanced indexing.

PR #544, opened 2020-11-13
Add  slots  to classes in ast.core

mostafaMamoni:
`__slots__` added in bitwise, pyccel, and boolean operators classes in ast.core.


pyccel/ast/basic.py
```diff
@@ -6,7 +6,7 @@
 class Basic(sp_Basic):
     """Basic class for Pyccel AST."""
     _fst = None
-
+    __slots__ = []
```
EmilyBourne, line 9: Shouldn't this be `__slots__ = ['_fst']`?

pyccel/ast/core.py
```diff
@@ -211,6 +211,7 @@ def broadcast(shape_1, shape_2):
         new_shape_1 = shape_1
 
     new_shape = []
+    __slots__ = []
```
EmilyBourne, line 214: I don't think `__slots__` make sense in functions.
See https://book.pythontips.com/en/latest/__slots__magic.html
for a description of slots

pyccel/ast/core.py
```diff
@@ -249,7 +251,7 @@ def handle_precedence(args, my_precedence):
 class PyccelBitOperator(Expr, PyccelAstNode):
     _rank = 0
     _shape = ()
-
+    __slots__ = ['_args']
```
EmilyBourne, line 254: The `_args` slot should be already defined by the `Expr` class.
See https://stackoverflow.com/questions/1816483/how-does-inheritance-of-slots-in-subclasses-actually-work
for discussions of how `__slots__` works for subclasses

pyccel/ast/core.py
```diff
@@ -337,6 +346,7 @@ def __init__(self, *args):
 
 class PyccelOperator(Expr, PyccelAstNode):
 
+    __slots__ = ['_args', '_dtype', '_rank', '_shape']
```
EmilyBourne, line 349: `PyccelAstNode` will also need a `__slots__` attribute for this to serve a purpose

pyccel/ast/basic.py
EmilyBourne: I am not sure if the `_fst = None` declaration needs declaring in the slots. It may be that it does need to be in the slots but that the current declaration needs to be removed. This needs some research

EmilyBourne (changes requested):
I suggest you read a little about `__slots__` to understand exactly what it is doing so that you're clear on where it is needed and what it should contain. Maybe you could summarise at one of the scrum meetings when you're done so everyone is clear on its use and the special cases (e.g. inheritance)

pyccel/ast/core.py
yguclu: Another good reference is here: https://stackoverflow.com/questions/472000/usage-of-slots

pyccel/ast/core.py
yguclu: By the way, we are going to have a big problem with multiple inheritance.

EmilyBourne: Is there any reason for `PyccelAstNode` to not inherit from `Basic` ? That could help with a lot of these multiple inheritance problems

yguclu: I think `Expr` allows a node to be used in an arithmetic expression, while `Basic` is, well, more "basic"....

yguclu: But since we have the basic arithmetic operators as Pyccel nodes, we should be able to do this. I'm afraid we went down a rabbit hole here :sweat_smile:

Issue #538, opened 2020-11-10
Improve stdlib header files


EmilyBourne:
**Describe the bug**
The datatypes used in the stdlib header files are not sufficiently precise. E.g. `pyccel/stdlib/internal/openmp.h` contains:
```
#$ header function omp_set_num_threads(int)
#$ header function omp_set_dynamic(bool)
```

An `int` has different precision on different platforms but `omp_set_num_threads` always requires the same precision.
The current definition results in a warning when compiling with `ifort`

A similar problem arises with `bool`. The `C_BOOL` type does not always have the same precision. On linux it has `kind=4`, however `omp_set_dynamic` expects `kind=1`. This is why the `test_modules_4` and `test_modules_4_1` tests in `tests/epyccel/test_epyccel_openmp.py` are xfailing after PR #527 . We therefore need some way to specify this for the header files which do not necessarily represent python functions

saidctb:
you can use `bool*1` to say that the precision is `1`

EmilyBourne:
> you can use `bool*1` to say that the precision is `1`

I have just tested this and it does not work, nor does `bool1`. However even if we fix the problem of specifying the bool precision we are going to run into a second problem as the user will not specify this precision. This will lead to a problem such as:
```
pyccel:
 |error [semantic]: mod_e74wjh9m29r3_e74wjh9m29r3.py [4,4]| Incompatible argument dynamic_theads (bool(kind=4)) passed to function (expected bool(kind=1)). Please cast the argument explicitly or overload the function (see https://github.com/pyccel/pyccel/blob/master/tutorial/headers.md for details) (omp_set_dynamic(dynamic_theads))
```

I am not sure how to fix this. The type assumption should only be made for pyccel internal functions. Worse, I think this is a fortran specific problem so it seems odd to handle it in the semantic stage

EmilyBourne:
`omp_set_num_threads` accepts `int32` on all systems, however gfortran is more tolerant. Intel and PGI raise warnings:
```
  /scratch/ebourne/pyccel/pyccel/codegen/compiling/compilers.py:421: UserWarning: /scratch/ebourne/pyccel/tests/epyccel/__epyccel__/__pyccel__/mod_pavzmp4tybyk_pavzmp4tybyk.f90(18): warning #6075: The data type of the actual argument does not match the definition.   [N]
      call omp_set_num_threads(n)
  -----------------------------^
  
    warnings.warn(UserWarning(err))
```
[...]

PR #545, opened 2020-11-13
Rename AST nodes corresponding to literal constants

### LINKED ISSUE #532: Rename AST nodes corresponding to literal constants
Description
--------------
As a DEVELOPER, I would like to easily distinguish between Python literals and other objects (such as variables, language constructs, operators, etc..). Therefore we should rename the classes that correspond to:

- Integer literal (e.g. `1`, `3`)
- Float literal (e.g. `2.5`, `1e-12`)
- Complex literal (e.g. `1 + 2j`)
- Boolean literal (`True` and `False`)
- String literal  (e.g. `Hello` and `world`)

And we should define them all in the same file.

Acceptance Criteria
-------------------------
- [x] Put all classes corresponding to Python literals into the same module (e.g. `ast.literals`)
- [x] Choose a unique naming scheme for all literals (e.g. `LiteralInteger`, `LiteralFloat`, etc.)
- [x] Rename all literal classes consistently

EmilyBourne:
- Rename literal classes so they always begin with `Literal`
- Rename `ast.numbers` -> `ast.literals`
- Move `String`->`LiteralString` class to `ast.literals`
- Update `_visit_ClassName` and `_print_ClassName` function names

Closes #532.


pyccel/ast/literals.py
```diff
@@ -0,0 +1,133 @@
...
+    'LiteralInteger',
+    'LiteralFloat',
+    'LiteralComplex',
+    'LiteralImaginaryUnit',
+    'LiteralString',
+    'get_default_literal_value'
+)
+
+#------------------------------------------------------------------------------
+class LiteralBooleanTrue(sp_BooleanTrue, PyccelAstNode):
```
EmilyBourne, line 24: Is this inheritance useful? Should we use this PR to also remove sympy from this file as it is not doing anything?

EmilyBourne: We need this as long as we are using `sympy.Function` as sympy "sympify"s arguments passed to the generated function. There does not seem to be a way to prevent this

yguclu: We'll create an Epic to tackle this.

pyccel/ast/literals.py
```diff
@@ -0,0 +1,134 @@
...
```
yguclu, line 24: What do you think of calling this class `LiteralTrue`, and the one below `LiteralFalse`? There is no ambiguity, I believe.

pyccel/parser/semantic.py
```diff
@@ -791,29 +789,25 @@ def _visit_CommentBlock(self, expr, **settings):
...
-    def _visit_Float(self, expr, **settings):
-        if isinstance(expr, Float):
-            return expr
-        elif isinstance(expr, sp_Float):
-            return Float(expr)
-        else:
-            raise TypeError("Float type is not sympy Float or pyccel Float")
-    def _visit_Complex(self, expr, **settings):
+    def _visit_LiteralInteger(self, expr, **settings):
+        return LiteralInteger(expr)
```
yguclu, line 793: Shouldn't we return just `expr` here?
Actually, it sounds like we only need one `_visit_Literal` function to cover all cases.

pyccel/ast/literals.py
```diff
@@ -0,0 +1,134 @@
...
+from sympy.logic.boolalg import BooleanTrue as sp_BooleanTrue, BooleanFalse as sp_BooleanFalse
+from sympy.core.expr     import Expr
+
+from .basic              import PyccelAstNode, Basic
+from .datatypes          import (NativeInteger, NativeBool, NativeReal,
+                                  NativeComplex, NativeString, default_precision)
+
+__all__ = (
+    'LiteralBooleanTrue',
+    'LiteralBooleanFalse',
```
yguclu, line 14: I wonder if `LiteralTrue` and `LiteralFalse` should be singletons.
Maybe in the future we'll need one for each value of `precision` used in the code.

yguclu (changes requested):

pyccel/ast/literals.py
EmilyBourne: Sounds good to me

pyccel/parser/semantic.py
EmilyBourne: Ah! This is how I managed to miss the cast in `_visit_Integer`. I put it in the wrong function.
I agree, I think there is a lot of overlap between the literal functions (there could be more if we didn't inherit from sympy.Integer etc)

pyccel/ast/literals.py
EmilyBourne: That makes sense, but we might have a precision problem. See issue #538
Given that we already have this precision problem I could make them singletons for now. What do you think?

yguclu: Sorry for the late reply. Yes, please make them singletons!

yguclu (approved):

pyccel/ast/literals.py
yguclu: After some discussion we have realized that using singletons for `LiteralTrue` and `LiteralFalse` will complicate greatly our code if we ever need to change their precision in the semantic stage. Hence we have decided to avoid this change, and merge this PR as it is.

PR #547, opened 2020-11-17
Save CodePrinter in Codegen to use information collected during printing stage

### LINKED ISSUE #483: Imports are included in both the .h and .c file
**Describe the bug**
Currently most imports are placed in the `.h` files. Normally the number of imports in the `.h` file should be minimised. So only the files necessary for the signatures should be imported. It may be difficult to tell which files are required so currently everything is imported in the `.h` file. The `.c` file then imports the `.h` file and any discovered imports (e.g. `math.h`, `complex.h`). These discovered imports may exist in both the `.h` and the `.c` file. This is notably the case for any function with a complex argument

**Language**
C

rakati:
Change the way we call the codeprinter to be saved in Codegen:

- Change the printer_register to map languages to its codeprinter;
- Call directly the codeprinter without using the intermediate function;
- Save the codeprinter in _printer to use its information;
- Print moduleheader first to include only libraries used by it.

Closes #483


pyccel/codegen/codegen.py
```diff
@@ -149,6 +154,23 @@ def code(self):
...
+        """ Set the current codeprinter instance"""
+        # Get language used (default language used is fortran)
+        language = settings.pop('language', 'fortran')
+
+        # Set language
+        if not language in ['fortran', 'c', 'python']:
+            raise ValueError('{} language is not available'.format(language))
+        self._language = language
+
+        # instanciat codePrinter
```
EmilyBourne, line 167: typo `instanciat`->`instantiate`

pyccel/codegen/printing/ccode.py
```diff
@@ -274,7 +274,7 @@ def _print_ModuleHeader(self, expr):
         funcs = '\n\n'.join('{};'.format(self.function_signature(f)) for f in expr.module.funcs)
 
         # Print imports last to be sure that all additional_imports have been collected
-        imports = [*expr.module.imports, *map(Import, self._additional_imports)]
+        imports = [*expr.module.imports,*map(Import, self._additional_imports)]
```
EmilyBourne, line 277: Not sure why you changed this file

pyccel/codegen/printing/pycode.py
```diff
@@ -34,9 +34,9 @@ class PythonCodePrinter(SympyPythonCodePrinter):
     ))
     _kc = {k: ''+v for k, v in _known_constants_math.items()}
 
-    def __init__(self, settings=None):
+    def __init__(self, parser=None, settings={}):
```
EmilyBourne, line 37: The default is None to make codacy happy. See here for why : https://stackoverflow.com/a/41686973/5957195

yguclu: As a rule of thumb, default values should be immutable.

pyccel/codegen/printing/ccode.py
rakati: this was revert changes.
if you follow the commits you will see that.

EmilyBourne: Ah so you changed the file then changed it back and the only difference left is the removed space. In that case please checkout this file from the master branch so it is clear that it is unchanged

EmilyBourne (approved):

yguclu (approved):

PR #548, opened 2020-11-17
Add Interfaces to the C wrapper

MenoIy:
Solve issue #512:

**A global description of the approach  used:**

- Use stdint.h macros for integer precision;
- Create check type function for python and numpy data types;
- Create function` _PyArray_ScalarAsCtype_` to collect C value from numpy types;
- Create a registry that maps each type to a given flag;
- Create a printer for the interface in the C wrapper;
- Loop on all the functions in the interfaces and collect the following information: 
    - Argument: name,  datatype, and flag;
    - Check datatype function for each arguments data types;
    - The python type object to parse value: pyobject for  rank < 1 else PyArrayObject
    - Body: body responsible for collecting values from python object,
    - Temporary variables used to extract values,
    - Returns;
- Create a mini wrapper function from the information collected above and call the corresponding interface function;
- Create a function from the collected flags and check_type functions that verify each argument and return a given check value;
- Append the call to each function of the interface to the corresponding condition.

**More details:** 
the condition of each function call is an integer 64 bits each 4 bits responsible for holding the flag type of an argument  example : 
```
def func(a, b ,c, d)  ===>  ... 0000 0000 0000 0000 aaaa bbbb cccc dddd
```
aaaa, bbbb ,cccc, dddd are the flag value mentioned in the flag_regestry.
a long can hold 64 / 4  arguments and each argument can be from 1 to 15 type

this approach is used to avoid the multiple calls to check _type function 
- a detailed example : 
a python code 
```
@template('z', types=['int', 'real'])
@template('y', types=['int', 'real'])
@types('z', 'y')
def func(z, y):
    return z
```

will generate a C code in this form : 
```
mini_call_func0_(pyobject *a, pyobject *b)    // (int  ,real)
{
         int a_t;
         float b_t;
        //collecting value from a  and b 
        if PyArray_CheckScalar(a)
                PyArray_ScalarAsCtype(a, &a_t)
        else 
                a_t = PyLong_AsLong(a)
       if PyArray_CheckScalar(b)
                PyArray_ScalarAsCtype(b, &b_t)
        else 
                b_t = PyFloat_AsDouble(b)
        res =  call func(a_t, b_t)
        //build python value
        //return python value
}
mini_call_func_1(pyobject *a, pyobject *b)
{...}
mini_call_func_2(pyobject *a, pyobject *b)
{...}
mini_call_func_3(pyobject *a, pyobject *b)
{...}
check_func((pyobject *a, pyobject *b))
{
       int flag = 0;
       if (a is int )
              flag += 2 << 4
       else  if (a is float)
              flag += 8 << 4
       else
              return 0 and an error format (a  should be either int or float)
       //same for b  but  flags are not shifted
       if (b is float)
              flag += 8
       .
       .
       return flag
}
func_wrapper()
{
       // declaring  and parsing args ;
       int check = check_func(a , b)
       if check == 0
              return null
       if check == 34 
              call func int int
       else if check == 40
              call func int float 
        .
        .
        else 
               return  null and raise error wrong combination
}
```


pyccel/codegen/printing/cwrappercode.py
```diff
@@ -330,6 +333,193 @@ def optional_element_management(self, used_names, a, collect_var):
...
+        check_func_body = [Assign(check_var, LiteralInteger(0))] + check_func_body
+        check_func_body.append(Return([check_var]))
+        # Creating check function definition
+        check_func_name = self.get_new_name(used_names.union(self._global_names), 'type_check')
+        self._global_names.add(check_func_name)
+        check_func_def = FunctionDef(name = check_func_name,
+            arguments = parse_args,
+            results = [check_var],
+            body = check_func_body,
+            local_vars = [])
```
EmilyBourne, line 469: This handles a completely different c function right? I think it would be clearer if this code went into it's own function rather than being in the interface wrapper, even if it's only used here. What do you think?

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -330,6 +333,193 @@ def optional_element_management(self, used_names, a, collect_var):
...
+        funcs_def.append(check_func_def)
+
+        # Create the wrapper body with collected informations
+        body_tmp = [((PyccelNot(check_var), [Return([Nil()])]))] + body_tmp
```
EmilyBourne, line 473: `if ( ! check_var ) return NILL` ?
Shouldn't this have an associated error message?

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -330,6 +333,193 @@ def optional_element_management(self, used_names, a, collect_var):
...
+        body_tmp.append((LiteralTrue(),
+            [PyErr_SetString('PyExc_TypeError', '"Arguments combinations don\'t exist"'),
+            Return([Nil()])]))
+        wrapper_body_translations = [If(*body_tmp)]
+
+        # Parsing Arguments
+        parse_node = PyArg_ParseTupleNode(python_func_args, python_func_kwargs, funcs[0].arguments, parse_args, keyword_list, True)
+        wrapper_body += list(default_value.values())
+        wrapper_body.append(If((PyccelNot(parse_node), [Return([Nil()])])))
```
EmilyBourne, line 482: Missing error message?

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -518,7 +687,11 @@ def _print_Module(self, expr):
             static_funcs = expr.funcs
         function_signatures = '\n'.join('{};'.format(self.function_signature(f)) for f in static_funcs)
 
-        function_defs = '\n\n'.join(self._print(f) for f in expr.funcs)
+        #TODO it should be a better way to do this :
+        funcs = expr.interfaces
+        funcs += [f for f in expr.funcs if f.name not in [item.name for s in funcs for item in s.functions]]
```
EmilyBourne, line 692: I would split this into 2 lines to make sure that python doesn't try to create the inner list on every loop. So:
```
item_names = [item.name for s in funcs for item in s.functions]
funcs += [f for f in expr.funcs if f.name not in item_names]
```

What is item_names?

EmilyBourne (changes requested):
Obviously this needs lots of tests. But I assume you're waiting for PR #530 to go through too?

Otherwise well done. This looks surprisingly neat for the amount of things it implements :)

pyccel/ast/cwrapper.py
```diff
@@ -154,12 +154,18 @@ def __init__(self, python_func_args, python_func_kwargs, c_func_args, parse_args
...
+                self._flags += pytype_parse_registry[(parse_args[i].dtype, parse_args[i].precision)]
             i+=1
         if i < len(c_func_args):
             self._flags += '|'
         while i < len(c_func_args):
-            self._flags += self.get_pytype(c_func_args[i], parse_args[i])
+            if isinstance(c_func_args[i], FunctionAddress) or is_interface:
+                self._flags += 'O'
+            else:
+                self._flags += pytype_parse_registry[(parse_args[i].dtype, parse_args[i].precision)]
```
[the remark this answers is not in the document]
MenoIy: Nothing is wrong my bad I was testing  with that code and when  I merged your pull req I kept my change and i forgot to remove it later

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: Here : https://github.com/pyccel/pyccel/pull/548/files#diff-69e3bda8e13a07d02bdfe586b58c9b581dd6497ea648cbe202e535f31c06c7aaR458
?
:+1:

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: Same thing right? PyArg_ParseTupleNode sets the error?

pyccel/ast/cwrapper.py
```diff
@@ -154,12 +154,18 @@ def __init__(self, python_func_args, python_func_kwargs, c_func_args, parse_args
         i = 0
 
         while i < len(c_func_args) and not isinstance(c_func_args[i], ValuedVariable):
-            self._flags += self.get_pytype(c_func_args[i], parse_args[i])
+            if isinstance(c_func_args[i], FunctionAddress) or is_interface:
+                self._flags += 'O'
+            else:
+                self._flags += pytype_parse_registry[(parse_args[i].dtype, parse_args[i].precision)]
```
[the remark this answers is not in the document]
MenoIy: I think I need to change the function call  and  it should no be  a staticmethod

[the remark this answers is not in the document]
EmilyBourne: I'm not sure you need to do both. The only reason it's a staticmethod is to make codacy happy as it doesn't use anything in self

pyccel/ast/cwrapper.py
```diff
@@ -372,16 +378,34 @@ def args(self):
...
+                    results   = [Variable(dtype=NativeBool(), name = 'r')])
+
+    # check for python type
     try :
-        check_type = check_type_registry[data_type]
+        check_type = python_type_check_registry[data.dtype]
     except KeyError:
-        errors.report(PYCCEL_RESTRICTION_TODO, symbol=data_type,severity='fatal')
-    func = FunctionDef(name = check_type,
+        errors.report(PYCCEL_RESTRICTION_TODO, symbol=data.dtype,severity='fatal')
```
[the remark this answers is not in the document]
MenoIy: yes already tested that's why i created the new check_type function that takes into consideration the numpy types

pyccel/codegen/printing/cwrappercode.py
MenoIy: yes, it's completely different from the c function, I will try grouping it in one function.

pyccel/codegen/printing/cwrappercode.py
MenoIy: yes

pyccel/codegen/printing/cwrappercode.py
MenoIy: https://docs.python.org/3/c-api/arg.html#c.PyArg_ParseTupleAndKeywords
on failure, it returns false and raises the appropriate exception.

pyccel/ast/cwrapper.py
```diff
@@ -482,9 +506,44 @@ def pybool_to_bool(cast_function_name):
     NativeReal() : PyFloat_AsDouble,
 }
 
-check_type_registry = {
+python_type_check_registry = {
     NativeInteger(): 'PyLong_Check',
     NativeComplex() : 'PyComplex_Check',
     NativeReal() : 'PyFloat_Check',
     NativeBool() : 'PyBool_Check',
```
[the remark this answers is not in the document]
MenoIy: for the moment  this dict is used only in the check type function in ast/cwraper  this one create a function def with the name from the dict .

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -330,6 +333,193 @@ def optional_element_management(self, used_names, a, collect_var):
...
+
+        #Collecting all functions names
+        funcs = expr.functions
+        # Save all used names
+        used_names = set(n.name for n in funcs)
+
+        # Find a name for the wrapper function
+        wrapper_name = self.get_new_name(used_names.union(self._global_names), expr.name+"_wrapper")
+        self._function_wrapper_names[expr.name] = wrapper_name
+        self._global_names.add(wrapper_name)
```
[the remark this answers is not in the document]
MenoIy: it could be something like this (```get_wrapper_name(used_names, func_name)```)

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -330,6 +333,193 @@ def optional_element_management(self, used_names, a, collect_var):
...
+        body_tmp = []
+
+        # To store the mini function responsible of collecting value and calling interfaces functions and return the builded value
+        funcs_def = []
+        default_value = {} # dict to collect all initialisation needed in the wrapper
+        check_var = Variable(dtype = NativeInteger(), name = self.get_new_name(used_names , "check"))
+        wrapper_vars[check_var.name] = check_var
+        types_dict = {} #dict to collect each variable possible type and the and the corresponding flags
+        # Managing the body of wrapper
+        # TODO split or re use exisiting functiond in the wrapper
```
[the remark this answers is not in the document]
MenoIy: I don't think. we can do it now, I'm still trying to find a good way to do it

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -330,6 +333,193 @@ def optional_element_management(self, used_names, a, collect_var):
...
+        for func in funcs :
+            mini_wrapper_func_body = []
+            res_args = []
+            mini_wrapper_func_vars = {}
+            flags = 0
+
+            # Loop for all args in every functions and create the corresponding condition and body
+            # TODO add array management like in the wrapper of simple function
```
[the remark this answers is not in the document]
MenoIy: I didn't really understand the implementation done in the c wrapp code  so I left this one for later when I finish the PR.

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -330,6 +333,193 @@ def optional_element_management(self, used_names, a, collect_var):
...
+        # Collect wrapper arguments and results
+        wrapper_args    = [python_func_selfarg, python_func_args, python_func_kwargs]
+        wrapper_results = [self.get_new_PyObject("result", used_names)]
+
+        # Collect parser arguments
+        parse_args = [Variable(dtype=PyccelPyObject(), is_pointer=True, name = self.get_new_name(used_names, a.name+"_tmp")) for a in funcs[0].arguments]
+        wrapper_vars = {a.name : a for a in parse_args}
+
+        # Collect argument names for PyArgParse
+        arg_names         = [a.name for a in funcs[0].arguments]
```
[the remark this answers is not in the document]
MenoIy: no

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -330,6 +333,193 @@ def optional_element_management(self, used_names, a, collect_var):
...
+        default_value = {} # dict to collect all initialisation needed in the wrapper
+        check_var = Variable(dtype = NativeInteger(), name = self.get_new_name(used_names , "check"))
+        wrapper_vars[check_var.name] = check_var
+        types_dict = {} #dict to collect each variable possible type and the and the corresponding flags
+        # Managing the body of wrapper
+        # TODO split or re use exisiting functiond in the wrapper
+        for func in funcs :
+            mini_wrapper_func_body = []
+            res_args = []
+            mini_wrapper_func_vars = {}
```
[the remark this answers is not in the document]
MenoIy: it was needed when i started the PR after improving  the code i left it as dict

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -330,6 +333,193 @@ def optional_element_management(self, used_names, a, collect_var):
...
+                    check = PyccelAssociativeParenthesis(PyccelOr(PyccelEq(VariableAddress(a), VariableAddress(Py_None)), check))
+                    default_value[a.name] =  self.get_default_assign(a, b)
+
+                    if not b.is_optional :
+                        assign = [Assign(b, IfTernaryOperator(PyccelNe(VariableAddress(a), VariableAddress(Py_None)),
+                            self.get_collect_function_call(b, a), b.value))]
+                    else :  # Managing optional variable
+                        tmp_var = Variable(dtype=b.dtype, name = self.get_new_name(used_names, b.name+"_tmp"))
+                        mini_wrapper_func_vars[tmp_var.name] = tmp_var
+                        assign = [Assign(tmp_var, self.get_collect_function_call(tmp_var, a)), Assign(VariableAddress(b), VariableAddress(tmp_var))]
```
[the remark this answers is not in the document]
MenoIy: AliasAssign LFS should be a pointer here b is not a pointer but optional.

pyccel/ast/cwrapper.py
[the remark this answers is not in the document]
EmilyBourne: Ah, wait it works because you just check the dtype right? But does that not mean that your check for an `int8` would be:
```
if ( PyArray_IsScalar(arg, Byte) || PyLong_Check(arg) )
```
?
Have you tried passing an `int` to the prototype above?

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -330,6 +333,193 @@ def optional_element_management(self, used_names, a, collect_var):
...
+                        assign = [If((PyccelEq(VariableAddress(a), VariableAddress(Py_None)),
+                                    [Assign(VariableAddress(b), b.value)]), (LiteralTrue(), assign))]
```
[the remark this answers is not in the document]
MenoIy: b.value  =  None it treated as NULL in C  ``` int *b;     b = NULL ```

pyccel/codegen/printing/cwrappercode.py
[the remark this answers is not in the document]
EmilyBourne: Then we have a problem don't we? Because the `parse_args` are renamed compared to `funcs[0].arguments`

pyccel/codegen/printing/cwrappercode.py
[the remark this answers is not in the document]
EmilyBourne: I'd offer to explain via zoom but I'm not going to have a connection good enough today and I'm busy tomorrow and Friday. Maybe we can try and discuss this on slack?

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -330,6 +333,193 @@ def optional_element_management(self, used_names, a, collect_var):
...
+
+                types_dict.setdefault(b, set()).add((b, check, flag_value)) # collect variable type for each arguments
```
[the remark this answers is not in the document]
MenoIy: dict.setdefault(key, default=None)
key − This is the key to be searched.
default − This is the Value to be returned in case key is not found.

in the first run, key don't exist because the dict is empty  so it will  create key : value and return an empty set()
where value = set()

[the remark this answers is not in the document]
MenoIy: no this one is collecting every argument for every function with some additional information.
so for example  : 
```
func1(int  a)
a ((a,   is_int,   flag))
func2(float a)
a((a,   is_int,   flag)  ,(a,  is_float, flag))
```

pyccel/codegen/printing/cwrappercode.py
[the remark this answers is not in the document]
EmilyBourne: I don't see that check in the `_print_AliasAssign` function. I see your argument, but in the cwrapper we treat optional arguments as something closer to a pointer so maybe it makes sense to use the pointer assignment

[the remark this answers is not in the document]
EmilyBourne: I don't think it makes much difference though, so I'm happy to leave it as you prefer

pyccel/codegen/printing/cwrappercode.py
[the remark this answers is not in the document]
EmilyBourne: ok. Is it faster if we do:
`types_dict[b] = set()`
before the loop and then just use the normal accessor? It seems more readable to me too

pyccel/codegen/printing/cwrappercode.py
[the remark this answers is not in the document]
MenoIy: don't really understand.

.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+    "files.associations": {
+        "vector": "cpp",
+        "utility": "cpp"
+    }
+}
```
EmilyBourne, line 6: Did you mean to commit this file?

pyccel/codegen/printing/cwrappercode.py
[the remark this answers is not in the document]
MenoIy: ok

pyccel/codegen/printing/cwrappercode.py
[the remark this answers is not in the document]
EmilyBourne: Do you expect `arg_names` /= `wrapper_vars.keys()` ?
> no

That is what we have, so if it is unexpected then maybe there is a problem

pyccel/codegen/printing/cwrappercode.py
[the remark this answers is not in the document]
MenoIy: it wont work 
```
dict = {}
-->func1(int  a)  
d[a]  = set()
d[a].add((a,   is_int,   flag))
--->func2(float a)
d[a] = set()
d[a].add((a,   is_float,   flag))
--------------
print(dict)
{a : (a,   is_float,   flag)}
and it should be 
{a : ((a,   is_int,   flag)  ,(a,  is_float, flag))}

```

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -330,6 +333,202 @@ def optional_element_management(self, used_names, a, collect_var):
...
+        default_value = {} # dict to collect all initialisation needed in the wrapper
+        check_var = Variable(dtype = NativeInteger(), name = self.get_new_name(used_names , "check"))
+        wrapper_vars[check_var.name] = check_var
+        types_dict = {} #dict to collect each variable possible type and the corresponding flags
+        # Managing the body of wrapper
+        # TODO split or re use exisiting functions in the wrapper
+        for func in funcs :
+            mini_wrapper_func_body = []
+            res_args = []
+            mini_wrapper_func_vars = [a for a in func.arguments]
```
EmilyBourne, line 385: Use
`func.arguments.copy()`
or
`list(func.arguments)`

.vscode/settings.json
MenoIy: 😲

pyccel/codegen/printing/cwrappercode.py
MenoIy: i think  mini_wrapper_func_vars = func.arguments is better .

EmilyBourne: As long as you don't modify the list then yes

pyccel/codegen/printing/cwrappercode.py
MenoIy: funcs = list of interfaces
s.functions = function inside interfaces 
item = function
item.name = function.name

pyccel/codegen/printing/cwrappercode.py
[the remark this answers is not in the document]
EmilyBourne: Ah yes, the loops are in the wrong order to be able to do this. You would have to do:
```
types_dict = {a:set() for a in func[0].arguments}
```

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: :+1:
I find the naming confusing. Is this equivalent to:
```
interfaces = expr.interfaces
interface_funcs = [f.name for i in interfaces for f in i.functions]
funcs = interfaces + [f for f in expr.funcs if f.name not in interface_funcs]
```
?

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -518,7 +696,12 @@ def _print_Module(self, expr):
             static_funcs = expr.funcs
         function_signatures = '\n'.join('{};'.format(self.function_signature(f)) for f in static_funcs)
 
-        function_defs = '\n\n'.join(self._print(f) for f in expr.funcs)
+        #TODO it should be a better way to do this :
+        funcs = expr.interfaces
```
EmilyBourne, line 700: Does this method create a copy of a list? If not are we in danger of modifying the `Module` contents here?

pyccel/codegen/printing/cwrappercode.py
MenoIy: yes it s equivalent

EmilyBourne: Then would you mind changing the names to something similar to make this clearer

pyccel/codegen/printing/cwrappercode.py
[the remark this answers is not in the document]
MenoIy: thats not the same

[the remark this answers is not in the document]
EmilyBourne: Meaning it doesn't get the right results?
I thought:
```

File: pyccel/codegen/printing/cwrappercode.py (hunk 4 above)
Comment (EmilyBourne) [reply pyccel/codegen/printing/cwrappercode.py]: Thanks

File: pyccel/codegen/printing/cwrappercode.py (hunk 16 above)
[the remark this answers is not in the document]
Comment (EmilyBourne) [reply pyccel/codegen/printing/cwrappercode.py]: It's just a question of syntax really. Unless there are other people who don't understand this line either then you can leave it as it is if you prefer it

File: pyccel/ast/cwrapper.py (hunk 7 above)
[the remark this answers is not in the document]
Comment (MenoIy) [reply pyccel/ast/cwrapper.py:399]: I'm checking for type with precision first in numpy_type_check_registry after that i go create a check for python type 
the numpy one need type and precision  and python on need only the type .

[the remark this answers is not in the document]
Comment (EmilyBourne) [reply pyccel/ast/cwrapper.py:399]: But if you use an `or` then only one needs to be right so a long should pass the test when you're expecting an int8 shouldn't it?
I'm not clear on why you need both of these. Maybe we should discuss this again after PR #530 is merged and you have added tests

File: pyccel/codegen/printing/cwrappercode.py (hunk 19 above)
Comment (MenoIy) [reply pyccel/codegen/printing/cwrappercode.py]: non it doesn't create a copy,  And I don't see where is a danger as long as we don't modify it.

Comment (EmilyBourne) [reply pyccel/codegen/printing/cwrappercode.py]: Ok. Now the names have been changed it seems more obvious that this is a temporary variable that is not modified :+1:

File: tests/epyccel/test_generic_functions.py [hunk 20]
```diff
@@ -105,3 +105,131 @@ def test_multi_tmplt_2(language):
...
+    assert f1(np.float32(155.2), np.float32(177.1)) == f2(np.float32(155.2), np.float32(177.1))
+    assert f1(np.float64(166.6), np.float64(255.6)) == f2(np.float64(166.6), np.float64(255.6))
+
+def test_complex_types(language):
+    f1 = epyccel(mod.complex_types , language = language, verbose=True)
+    f2 = mod.complex_types
+
+    assert f1(complex(1, 2.2), complex(1, 2.2)) == f2(complex(1, 2.2), complex(1, 2.2))
+    assert f1(np.complex(15.5, 2.0) , np.complex(10.5, 3.4)) == f2(np.complex(15.5, 2.0) , np.complex(10.5, 3.4))
+    np.isclose(f1(np.complex64(15.5 + 2.0j) , np.complex64(10.5 + 3.4j)), f2(np.complex64(15.5 + 2.0j) , np.complex64(10.5 + 3.4j)), rtol = 1e-7, atol = 1e-8)
```
EmilyBourne, line 209: There is an `assert` missing here. When you put it back I think you will find that the test fails. I am getting a `ComplexWarning` from this line: `Casting complex values to real discards the imaginary part`. I think it is because you are using the python functions `PyComplex_RealAsDouble` and `PyComplex_ImagAsDouble` to unpack the complex. But here we do not have a `PyComplex` but instead a numpy complex.

Can we use this function for unpacking numpy types?
https://docs.scipy.org/doc/numpy-1.13.0/reference/c-api.array.html#c.PyArray_ScalarAsCtype

Does this indicate a more fundamental problem with how we treat types given that we have no way to distinguish between a python and a numpy type? If so do you think this problem is limited to the cwrapper or does it have larger consequences? (e.g. for properties)
```
In [1]: import numpy as np

In [2]: a = np.int64(2)

In [3]: len(dir(a))
Out[3]: 136

In [4]: b = 2

In [5]: len(dir(b))
Out[5]: 70
```

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -462,6 +486,197 @@ def get_PyBuildValue(self, used_names, variable):
...
+        flags = (len(types_dict) - 1) * 4
+        for a in types_dict:
+            var_name = ""
+            body = []
+            types = []
+            l = list(types_dict[a])
+            l.sort(key= lambda x : x[0].precision)
+            for s in l:
+                var_name = s[0].name
+                value = s[2] << flags
```
EmilyBourne, line 652: Is this definitely correct? For your test `default_var_1` I have:
```
long type_check_0014(PyObject *x_tmp, PyObject *y_tmp)
{

    long check;

    check = 0;
    if (PyLong_Check(x_tmp) || PyArray_IsScalar(x_tmp, Long))
    {   
        check += 32;
    }
    else if (PyFloat_Check(x_tmp) || PyArray_IsScalar(x_tmp, Double))
    {   
        check += 80;
    }
    else
    {
        PyErr_SetString(PyExc_TypeError, "x must be 64 bit integer or 64 bit real");
        return 0;
    }
    if ((y_tmp == Py_None || PyLong_Check(y_tmp) || PyArray_IsScalar(y_tmp, Long)))
    {   
        check += 2;
    }
    else
    {
        PyErr_SetString(PyExc_TypeError, "y must be 64 bit integer");
        return 0;
    }
    return check;
}
```
I thought the numbers for x should be between 32 and 64?

MenoIy: each argument is stored in  4 bits and shift to the left by 4 x (number of args - order) 
example :
so here x should be  between  (1 << 4     and  15 << 4)

EmilyBourne: Ok sorry. So then x should be between 16 and 240, which is the case

pyccel/ast/cwrapper.py
```diff
@@ -495,9 +528,44 @@ def pybool_to_bool(cast_function_name):
...
+
+# Needed to check for numpy arguments type
+Numpy_Bool_ref = Variable(dtype=NativeVoid(),  name = 'Bool')
+Numpy_Byte_ref = Variable(dtype=NativeVoid(),  name = 'Byte')
+Numpy_Short_ref = Variable(dtype=NativeVoid(),  name = 'Short')
+Numpy_Int_ref = Variable(dtype=NativeVoid(),  name = 'Int')
+Numpy_Long_ref = Variable(dtype=NativeVoid(),  name = 'Long')
+Numpy_Float_ref = Variable(dtype=NativeVoid(),  name = 'Float')
+Numpy_Double_ref = Variable(dtype=NativeVoid(),  name = 'Double')
+Numpy_Complex_ref = Variable(dtype=NativeVoid(),  name = 'ComplexFloating')
```
EmilyBourne, line 546: Are these sufficiently specific? Especially `ComplexFloating`?
I found a doc here:
https://numpy.org/devdocs/reference/arrays.scalars.html
which seems to indicate that `ComplexFloating` is abstract.

tests/epyccel/test_generic_functions.py
EmilyBourne: Here is the link with a more up to date version of numpy:
https://numpy.org/doc/stable/reference/c-api/array.html?highlight=array%20api#c.PyArray_ScalarAsCtype

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -462,6 +486,202 @@ def get_PyBuildValue(self, used_names, variable):
...
+
+        wrapper_body_translations = []
+        body_tmp = []
+
+        # To store the mini function responsible of collecting value and calling interfaces functions and return the builded value
+        funcs_def = []
+        default_value = {} # dict to collect all initialisation needed in the wrapper
+        check_var = Variable(dtype = NativeInteger(), name = self.get_new_name(used_names , "check"))
+        wrapper_vars[check_var.name] = check_var
+        types_dict = {a : set() for a in funcs[0].arguments} #dict to collect each variable possible type and the corresponding flags
```
EmilyBourne, line 526: This needs to be ordered in order to have the correct type check. Use:
```
types_dict = OrderedDict((a, set()) for a in funcs[0].arguments) #dict to collect each variable possible type and the corresponding flags
```

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -462,6 +486,202 @@ def get_PyBuildValue(self, used_names, variable):
...
+
+        sep = self._print(SeparatorComment(40))
+
+        return sep + '\n'.join(CCodePrinter._print_FunctionDef(self, f) for f in funcs_def)
+
+    def _create_wrapper_check(self, check_var, parse_args, types_dict, used_names, func_name):
+        check_func_body = []
+        flags = (len(types_dict) - 1) * 4
+        debug = []
+        for a in types_dict:
```
EmilyBourne, line 644: `types_dict` must be ordered to avoid problems. See above for fix

pyccel/ast/cwrapper.py
EmilyBourne: It is not sufficient. We need `CFloat` and `CDouble`

EmilyBourne (changes requested):
A difference should be made when collecting python vs numpy types, even if the precision is the same. If we use the python functions (e.g. PyLong_To_Long) then the numpy type is first cast to the python type then back to long. Instead as we know the precision we can use `PyArray_ScalarAsCtype` to collect the value efficiently with a `memcpy`:
https://numpy.org/doc/stable/reference/c-api/array.html?highlight=array%20api#c.PyArray_ScalarAsCtype

This will allow us to collect numpy basic types without casting and will ensure that the casting does not break anything (as in the case of complexes)

EmilyBourne:
I suspect the tests aren't running because this is a draft PR. Try removing this line temporarily:
https://github.com/pyccel/pyccel/blob/master/.github/workflows/Github_pytest.yml#L8

pyccel/ast/cwrapper.py
```diff
@@ -491,14 +524,63 @@ def pybool_to_bool(cast_function_name):
...
 }
+
+
+
+# Needed to check for numpy arguments type
+Numpy_Bool_ref = Variable(dtype=NativeVoid(),  name = 'Bool')
+Numpy_Byte_ref = Variable(dtype=NativeVoid(),  name = 'Byte')
+Numpy_Short_ref = Variable(dtype=NativeVoid(),  name = 'Short')
+Numpy_Int_ref = Variable(dtype=NativeVoid(),  name = 'Int')
+Numpy_Long_ref = Variable(dtype=NativeVoid(),  name = 'Long')
```
EmilyBourne, line 558: I can't find where these values are defined. Is there an equivalent variable with the precision in the name? Numpy contains both `np.long` and `np.int64`. See:
https://numpy.org/doc/stable/user/basics.types.html

I'm not certain that a `long` has the same precision on linux and windows. If an `Int64` exists in numpy and this fixes your type collection problems on windows then we also need to update the type that we use in our code to be certain that it matches, here:
https://github.com/pyccel/pyccel/blob/master/pyccel/codegen/printing/ccode.py#L194-L202
And probably in the array code too.
@saniQQ can you say what exactly would need updating

EmilyBourne: Here is a link with a better explanation of what I think the problem is:
https://blogs.igalia.com/mshin/2019/08/26/why-should-we-use-a-precise-width-integer-type/

> Data model | sizeof(int) | sizeof(long) | sizeof(long long) | example
> -- | -- | -- | -- | --
> LP32 | 16b | 32b | 64b | Win16
> ILP32 | 32b | 32b | 64b | Win32, i386 OSX & Linux
> LP64 | 32b | 64b | 64b | x86-64 OSX & Linux
> LLP64 | 32b | 32b | 64b | Win64

MenoIy: Yes I know the problem when using PyArray_IsScalar(pyobject,  type)     when giving Long as type it fails on windows because windows64 long is 32bits size so I tried using LongLong in type it works in windows and it fails on Linux ...

EmilyBourne: Does this mean that there is no `Int64` type provided by numpy?

EmilyBourne: The solution is not to change the imprecise types we use to other imprecise types as these will indeed be prone to the same problems. My suggested solution is to instead use precise types (i.e. `int8_t`, `int16_t`, `int32_t`, `int64_t`) see:
https://en.cppreference.com/w/c/types/integer

MenoIy: To explain better the problem stop at the check type stage so it comes from PyArray_IsScalar(pyobject, type) return false when it should return true.

saniQQ: we will need to update :
- the union in the t_ndarray struct 
- the enum (this will need to update the dictionary in ccode.py too)
- the Generic Macro to use the precise types 
- and the fill functions

EmilyBourne: This will affect PR #562 as @rakati added the fill functions to `ndarrays.h` and to the Generic Macro

saniQQ: i can push the changes here so @rakati can merge them later

EmilyBourne: That sounds like a good suggestion :+1:

pyccel/codegen/printing/ccode.py
```diff
@@ -239,7 +239,7 @@ def __init__(self, parser, settings=None):
         self.known_functions.update(userfuncs)
         self._dereference = set([] if settings is None else settings.get('dereference', []))
         self.prefix_module = prefix_module
-        self._additional_imports = set(['stdlib'])
+        self._additional_imports = set(['stdlib', 'stdint'])
```
EmilyBourne, line 242: Is this the right place for this?
@rakati fixed issue #483 with PR #547. I am not sure if this works with his fix or not

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -152,9 +154,61 @@ def _get_static_function(self, used_names, function, collect_dict):
...
+
+        return check
+
+    # --------------------------------------------------------------------
     # Functions that take care of creating cast or convert type function call :
-    # -------------------------------------------------------------------
+    # --------------------------------------------------------------------
+    def _create_collecting_value_body(self, variable, collect_var, tmp_variable = None):
+        """
+        Create If block to différence between python and numpy data types in collecting value
```
EmilyBourne, line 182: There is french in this line. How about:
```
Create If block to differentiate between python and numpy data types when collecting value
```

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -243,10 +296,11 @@ def _body_optional_variable(self, tmp_variable, variable, collect_var, check_typ
         body = [(PyccelEq(VariableAddress(collect_var), VariableAddress(Py_None)),
                 [Assign(VariableAddress(variable), Nil())])]
         if check_type : # Type check
-            check = PyccelNot(PythonType_Check(variable, collect_var))
+            check = PyccelNot(PyccelAssociativeParenthesis(PyccelOr(NumpyType_Check(variable, collect_var)
+                    , PythonType_Check(variable, collect_var))))
```
EmilyBourne, line 300: You shouldn't need to add parentheses yourself now that PR #559 has been merged

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -281,10 +335,11 @@ def _body_valued_variable(self, variable, collect_var, check_type = False) :
         body = [(PyccelEq(VariableAddress(collect_var), VariableAddress(Py_None)),
                 [Assign(variable, variable.value)])]
         if check_type : # Type check
-            check = PyccelNot(PythonType_Check(variable, collect_var))
+            check = PyccelNot(PyccelAssociativeParenthesis(PyccelOr(NumpyType_Check(variable, collect_var)
+                    , PythonType_Check(variable, collect_var))))
```
EmilyBourne, line 339: You shouldn't need to add parentheses yourself now that PR #559 has been merged

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -464,6 +524,197 @@ def get_PyBuildValue(self, used_names, variable):
...
+    def _create_wrapper_check(self, check_var, parse_args, types_dict, used_names, func_name):
+        check_func_body = []
+        flags = (len(types_dict) - 1) * 4
+        for a in types_dict:
+            var_name = ""
+            body = []
+            types = []
+            l = list(types_dict[a])
+            l.sort(key= lambda x : x[0].precision)
+            for s in l:
```
EmilyBourne, line 687: Can you use more descriptive names please. `a`, `l` and `s` is too many one letter variables for me to remember what is what

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -636,7 +883,13 @@ def _print_Module(self, expr):
             static_funcs = expr.funcs
         function_signatures = '\n'.join('{};'.format(self.function_signature(f)) for f in static_funcs)
 
-        function_defs = '\n\n'.join(self._print(f) for f in expr.funcs)
+        #TODO it should be a better way to do this :
+        interfaces = expr.interfaces
+        interface_funcs = [f.name for i in interfaces for f in i.functions]
+        funcs = interfaces + [f for f in expr.funcs if f.name not in interface_funcs]
```
EmilyBourne, line 889: Do you have something in mind for this TODO? I don't see a problem here

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -647,8 +900,8 @@ def _print_Module(self, expr):
                                      '}}').format(
                                             name = f.name,
                                             wrapper_name = self._function_wrapper_names[f.name],
-                                            doc_string = f.doc_string)
-                                     for f in expr.funcs)
+                                            doc_string = f.doc_string if isinstance(f, FunctionDef) else f.functions[0].doc_string)
```
EmilyBourne, line 903: Would it not be simpler to add a `doc_string` property to `Interface`? It has getters for most of `FunctionDef`'s other properties

tests/epyccel/test_epyccel_functions.py
```diff
@@ -333,23 +333,23 @@ def f20(a):
         b = a
         return b
     f = epyccel(f20, language=language)
-    assert f(complex(1, 2.2) == f20(complex(1, 2.2)))
+    assert f(complex(1, 2.2)), f20(complex(1, 2.2))
```
EmilyBourne, line 336: Missing `np.isclose`

tests/epyccel/test_epyccel_functions.py
```diff
@@ -333,23 +333,23 @@ def f20(a):
...
 
 def test_decorator_f21(language):
     @types('complex64')
     def f21(a):
         b = a
         return b
     f = epyccel(f21, language=language)
-    assert f(complex(1, 2.2) == f21(complex(1, 2.2)))
+    assert np.isclose(f(complex(1, 2.2)), f21(complex(1, 2.2)), rtol = 1e7, atol = 1e8)
```
EmilyBourne, line 344: This tolerance is worryingly high. Does this break if you use the same tolerance as in `tests/epyccel/recognised_functions/test_numpy_funcs.py` ?

tests/epyccel/test_epyccel_functions.py
```diff
@@ -333,23 +333,23 @@ def f20(a):
...
 
 def test_decorator_f22(language):
     @types('complex128')
     def f22(a):
         b = a
         return b
     f = epyccel(f22, language=language)
-    assert f(complex(1, 2.2) == f22(complex(1, 2.2)))
+    assert f(complex(1, 2.2)), f22(complex(1, 2.2))
```
EmilyBourne, line 352: Missing `np.isclose`

EmilyBourne (changes requested):
In the file `tests/epyccel/test_generic_functions.py` I have a few comments:

1. The xfail mark at the start of the file should no longer be valid. This PR should fix issue #512 for both fortran and c. There will still be cases that will fail as you have interfaces which can take `int` or `int32` for example. This causes errors in fortran, but it is a separate issue to issue #512 

2. Most of the epyccel calls in this file are translating the entirety of `tests/epyccel/modules/generic_functions.py`. This was done for 3 reasons.:
    1. To test if headers written anywhere in the file are attached to the correct function
    2. To test if translated code can correctly select and call an interface
    3. Because there was no cwrapper when this file was written

   As the third point is no longer valid I think this file can be broken up to speed up testing. I suggest:
    1. Split the file in 2. One file containing everything with headers (meaning the whole file must be translated) and another containing everything with decorators (meaning the function can be epyccelized individually
    2. Decide how many functions are necessary to test whether translated code can correctly select and call an interface. I suspect this is a lot less than the number that are currently used and that there were so many mostly due to the lack of a cwrapper. Now the wrapper is implemented some of the tests could be simplified to call the interfaced function directly

pyccel/codegen/printing/cwrappercode.py
MenoIy: ok I will change it

tests/epyccel/test_epyccel_functions.py
MenoIy: no, it should be just == they have the same precision. I'm gonna correct it .

tests/epyccel/test_epyccel_functions.py
MenoIy: yes it will break

EmilyBourne: Do you have an explanation for why?

pyccel/codegen/printing/cwrappercode.py
MenoIy: not anymore.

tests/epyccel/test_epyccel_functions.py
EmilyBourne: Even with the same precision `==` tends to be dangerous when dealing with floating point values. But given this just makes a copy, there's a chance that you'll get away with it

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: Then the TODO can be removed?

pyccel/ast/cwrapper.py
```diff
@@ -134,9 +134,18 @@ class PyArg_ParseTupleNode(Basic):
         List of arguments into which the result will be collected
     arg_names : list of str
         A list of the names of the function arguments
+    is_interface : boolean
+        Default value False and True when working with interface functions
     """
 
-    def __init__(self, python_func_args, python_func_kwargs, c_func_args, parse_args, arg_names):
+    def __new__(cls, *args, **kwargs):
+        return Basic.__new__(cls, *args, **kwargs)
```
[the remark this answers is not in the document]
MenoIy: can you explain

tests/epyccel/test_epyccel_functions.py
MenoIy: real and imag part of the complex are either double or float.  so when casting the double value to float we lose precision  because the float is stored in 32 bit [1 sign 8 exponent 23 mantissa] and double is stored in 64 [1, 11,  52]

pyccel/codegen/printing/cwrappercode.py
MenoIy: yes

tests/epyccel/test_epyccel_functions.py
EmilyBourne: Why are we casting to a float?

MenoIy: yes why ? it was before we implementing numpy function to collect value but now that we are using memcpy it should  work, im going to simulate the behaviour to see what happen in memory.

pyccel/ast/cwrapper.py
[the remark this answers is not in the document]
EmilyBourne: When a class inherits from another class (here `PyArg_ParseTupleNode` inherits from `Basic`) it inherits all of its methods. So when we call `__new__` if we haven't redefined `__new__` then we are calling `Basic.__new__`.

Sometimes we need to implement `__new__` ourselves. For example if the arguments to `__new__` for the base class don't match the arguments for the sub-class or if we want to leave ourselves the possibility to return a different type.

This is not the case here as you are calling `Basic.__new__` with the same arguments that were passed to `PyArg_ParseTupleNode.__new__`. This means that the same set of steps would be carried out whether this function is here or not.

pyccel/codegen/printing/ccode.py
```diff
@@ -568,6 +568,8 @@ def _print_FuncAddressDeclare(self, expr):
         return '{}(*{})({});'.format(ret_type, name, arg_code)
 
     def _print_Declare(self, expr):
+        if isinstance(expr.variable.dtype, NativeInteger) :
+            self._additional_imports.add('stdint')
```
EmilyBourne, line 572: This is definitely not the right place as it will be missed if the argument is an `int`. Maybe in `find_in_dtype_registry`. But your original location may be correct. That is why ideally I would like Noureddine's opinion

MenoIy: it was nourddine who suggested that I print include  only when it needed

pyccel/ast/cwrapper.py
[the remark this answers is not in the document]
EmilyBourne: If you want a refresher on classes and inheritance it might be useful to look at `ast.operators` and maybe pick issue #564 when it appears in a sprint as there is lots of inheritance there. The inheritance for the `__new__` and `__init__` methods work in the same way as the inheritance for all the other methods with the exception that the first argument of `__new__` is a type instead of an instance of the class

pyccel/codegen/printing/ccode.py
EmilyBourne: Ah ok :+1:

MenoIy:
@saniQQ  maybe you should create a small PR for your changes  ?

tests/epyccel/test_epyccel_functions.py
MenoIy: found it  ... here we are using complex64 and passing a python complex as arguments so it passes from the pytype_cast_function and not the numpy one.

pyccel/codegen/printing/ccode.py
EmilyBourne: I still think this is causing the windows failure though. If you translate the following function. Is stdint included?
```
@types('int')
def f(a):
    print(a)
```

tests/epyccel/test_base.py
```diff
@@ -14,7 +14,7 @@ class epyccel_test:
     """
     def __init__(self, f, lang='fortran'):
         self._f  = f
-        self._f2 = epyccel(f, language=lang)
+        self._f2 = epyccel(f, language=lang, verbose=True)
```
EmilyBourne, line 17: This is no longer needed

EmilyBourne:
> @saniQQ maybe you should create a small PR for your changes ?

I have collected the relevant commits from this branch and opened PR #569 for this

EmilyBourne:
I've added a couple of bug fixes. You should be able to activate most of the fortran tests now

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -576,6 +576,7 @@ def _print_Interface(self, expr):
             # Loop for all args in every functions and create the corresponding condition and body
             for p_arg, f_arg in zip(parse_args, func.arguments):
                 collect_var,_ = self.get_PyArgParseType(used_names, f_arg, p_arg)
+                print(collect_var, type(collect_var))
```
EmilyBourne, line 579: Don't forget to remove this before you merge

MenoIy: yes I'm trying to active array tests for fortran

tests/epyccel/modules/generic_functions_2.py
```diff
@@ -94,18 +94,18 @@ def optional_var_4(x, y = None):
...
 
 
-@template('g', types=['real', 'float32', 'float64'])
+@template('g', types=['real', 'float32'])
 @types('g', 'g')
 def float_types(x, y):
     return x + y
 
-@template('g', types=['complex', 'complex64', 'complex128'])
+@template('g', types=['complex', 'complex64'])
```
EmilyBourne: I don't think you should remove these. I think it would be better to mark them as xfail with a link to issues #567 and #568

MenoIy: I didn't want to mark all test as xfail as all other types work fine in fortran the only problem is when using same type in template.   the problem is the same in C but C doesn't complaine

EmilyBourne: Yes I agree. You should just mark these 3 functions as xfail for fortran. See `test_epyccel_openmp.py` for an example of how to do this

pyccel/ast/cwrapper.py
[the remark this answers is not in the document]
EmilyBourne: This comment seems to have been lost, but it is still valid

tests/epyccel/modules/generic_functions_2.py
```diff
@@ -129,14 +129,14 @@ def mix_types_2(x, y):
     return -x
 
 
-@template('g', types=['int[:]', 'real[:]', 'complex[:]'])
+@template('g', types=['int64[:]', 'real[:]', 'complex[:]'])
```
EmilyBourne: Can you explain why this change is necessary? Do we need an issue to correctly determine the precision of `int`?

MenoIy: Im trying to see why it fail.

EmilyBourne: Ok. Good luck. The precision of an int is supposed to be different for windows I think. It's determined with `numpy.dtype(int).alignment`. I think we should have `precision = 4` for windows and `precision = 8` for linux/mac. But as long as the array is initialised with the right type then there shouldn't be a problem. Maybe `np.int != int` on windows?

tests/epyccel/test_generic_functions.py
```diff
@@ -350,6 +367,30 @@ def test_mix_int_array():
     f2(x2, a)
     assert np.array_equal( x1, x2 )
 
+@pytest.mark.xfail(sys.platform == 'win32', reason='issue #567 and #568: Duplicated types in a template and header')
+def test_mix_int_array_2():
+    f1 = epyccel(mod2.mix_int_array_2)
+    f2 = mod2.mix_int_array_2
```
EmilyBourne, line 373: Are you sure this only fails on windows?
```
@template('g', types=['int[:]', 'int32[:]', 'int64[:]'])
@types('g', 'int')
def mix_int_array_2(x, a):
```
Shouldn't it fail everywhere?

MenoIy: yes

tests/epyccel/test_generic_functions.py
```diff
@@ -95,13 +94,341 @@ def test_tmplt_tmplt_1(language):
...
+
+@pytest.mark.parametrize( 'language', (
+        pytest.param("c", marks = pytest.mark.c),
+        pytest.param("fortran", marks = [
+            pytest.mark.xfail(sys.platform == 'win32', reason='issue #567 and #568: Duplicated types in a template and header'),
+            pytest.mark.fortran]
+        )
+    )
+)
+def test_mix_types_3(language):
```
EmilyBourne, line 281: After your last modification this will also fail on all platforms

MenoIy: this one should only fail in fortran true

MenoIy: c win32

EmilyBourne (approved):

yguclu:
I don't understand a couple of points in this PR's description:

- When check type (python or numpy) before collecting value from a pyobject and append the given collect value function
- Collect the additional collect variable and temporary variable.
- add the call of check_type function to the wrapper and group each condition with it a corresponding function call
- group all the functions created mini_wrapper and check_type and the wrapper

Would you mind rephrasing them?

Also, I usually use the PR's description as the commit message for the squashed merge commit, but this is far too long and detailed. Can you please try and explain with a few complete sentences what you are doing and why?

pyccel/codegen/printing/fcode.py
```diff
@@ -2436,9 +2436,9 @@ def _print_LiteralFloat(self, expr):
         return "%s_C_DOUBLE" % printed
 
     def _print_LiteralComplex(self, expr):
-        real_str = self._print_Float(expr.real)
-        imag_str = self._print_Float(expr.imag)
-        return "({}, {})".format(real_str, imag_str)
+        real_str = self._print(expr.real)
+        imag_str = self._print(expr.imag)
+        return "cmplx({}, {}, kind = {})".format(real_str, imag_str, expr.precision)
```
yguclu, line 2441: I don't understand why we need to call the `cmplx` function in order to print a complex literal constant. `expr.real` and `expr.imag` should inherit the correct precision from `expr`, and hence they should be printed correctly. For example
```fortran
(1.0_C_DOUBLE, 2.5_C_DOUBLE)
```

tests/epyccel/modules/generic_functions.py
```diff
@@ -108,7 +83,7 @@ def tst_gen_5():
 def tst_gen_6():
     x = gen_6(5.5, 5.5)
     y = gen_6(5, 5)
-    z = gen_6(1j, 1j)
+    z = gen_6(complex(1, 2), complex(1, 2))
```
yguclu, line 86: Is a more Pythonic syntax also accepted? E.g.
```python
z = gen_6(1+2j, 1+2j)
```

tests/epyccel/modules/generic_functions_2.py
```diff
@@ -0,0 +1,171 @@
+# pylint: disable=missing-function-docstring, missing-module-docstring/
+from pyccel.decorators import types
+from pyccel.decorators import template
+
+
+@template('z', types=['int', 'real'])
+@types('z', 'z')
+def tmplt_1(x, y):
+    return x + y
```
yguclu, line 9: Given that our unit tests also work as examples, would you mind capitalize the names of the template types throughout this file? E.g.
```python
@template('T', types=['int', 'real'])
@types('T', 'T')
```
This is a common convention for generic programming in C++

yguclu (changes requested):

tests/epyccel/modules/generic_functions.py
MenoIy: yes, It completely fine I'm gonna add it in some tests

pyccel/codegen/printing/fcode.py
MenoIy: it was @EmilyBourne  who made that change I think she had some reason to do that.

EmilyBourne: This was causing problems when a function calls another and passes a complex literal constant. I have checked out the branch at the point at which I had the problem. It was caused by :
```
z = gen_6(1j, 1j)
```
for 
```
#$ header template S(int|real|complex|bool)
@types('S', 'S')
def gen_6(x, y):
    return x + y
```

The code generated was:
```
  z = gen_6((1_C_LONG_LONG, 2_C_LONG_LONG),
      (1_C_LONG_LONG, 2_C_LONG_LONG))
```
which generated the errors:
```
generic_functions.f90:562:44:

   z = gen_6((1_C_LONG_LONG, 2_C_LONG_LONG),
                                            1
Error: Syntax error in argument list at (1)
generic_functions.f90:563:6:

       (1_C_LONG_LONG, 2_C_LONG_LONG))
      1
Error: Unclassifiable statement at (1)
```

It is really due to an underlying problem which is that we don't check that the arguments are `LiteralFloat` when we create a `LiteralComplex`. Presumably one had been missed. This seems to have been corrected elsewhere though as I cannot reproduce the problem on the branch in its current state. I am therefore happy to revert these changes

MenoIy:
@yguclu  I updated the PR description you can remove the detailed example once you understand the changes and how I calculate the arguments mask.

pyccel/codegen/printing/fcode.py
EmilyBourne: I have opened PR #572 to ensure Literals are correctly handled

yguclu (approved):

PR #553, opened 2020-11-19
Create numeric literals with correct precision

mostafaMamoni:
Commit summary:

-  Pass precision keyword-only argument to `__init__` constructors of numeric literals classes.
- Implement setter method for precision property.

### DIFF
--- a/pyccel/ast/literals.py
+++ b/pyccel/ast/literals.py
@@ -154,9 +154,6 @@ class LiteralString(Basic, Literal):
             raise TypeError('arg must be of type str')
         self._string = arg
 
-    def set_precision(self, precision):
-        self._precision = precision
-
     @property
     def arg(self):
         """ Return the python string literal """
--- a/pyccel/ast/literals.py
+++ b/pyccel/ast/literals.py
@@ -36,11 +36,17 @@ class LiteralTrue(sp_BooleanTrue, Literal):
     def __init__(self, precision = default_precision['bool']):
         self._precision = precision
 
+    def set_precision(self, precision):
+        self._precision = precision
+
 #------------------------------------------------------------------------------
 class LiteralFalse(sp_BooleanFalse, Literal):
     """Represents the python value False"""
     _dtype     = NativeBool()
     def __init__(self,precision = default_precision['bool']):
+        self._precision = precision
+
+    def set_precision(self, precision):
         self._precision = precision
 
 #------------------------------------------------------------------------------
@@ -56,11 +62,17 @@ class LiteralInteger(sp_Integer, Literal):
     def __init__(self, value, *, precision = default_precision['integer']):
         self._precision = precision
 
+    def set_precision(self, precision):
+        self._precision = precision
+
 #------------------------------------------------------------------------------
 class LiteralFloat(sp_Float, Literal):
     """Represents a float literal in python"""
     _dtype     = NativeReal()
     def __init__(self, value, *, precision = default_precision['float']):
+        self._precision = precision
+
+    def set_precision(self, precision):
         self._precision = precision
 
 #------------------------------------------------------------------------------
@@ -75,6 +87,9 @@ class LiteralComplex(Basic, Literal):
         Basic.__init__(self)
         self._real_part = real
         self._imag_part = imag
+        self._precision = precision
+
+    def set_precision(self, precision):
         self._precision = precision
 
     @property
@@ -109,6 +124,9 @@ class LiteralString(Basic, Literal):
         if not isinstance(arg, str):
             raise TypeError('arg must be of type str')
         self._string = arg
+
+    def set_precision(self, precision):
+        self._precision = precision
 
     @property
     def arg(self):
--- a/pyccel/ast/literals.py
+++ b/pyccel/ast/literals.py
@@ -37,7 +37,7 @@ class LiteralTrue(sp_BooleanTrue, Literal):
         self._precision = precision
 
     def set_precision(self, precision):
-        """ Set precision the true literal """
+        """ Set precision for the true literal """
         self._precision = precision
 
 #------------------------------------------------------------------------------
@@ -48,7 +48,7 @@ class LiteralFalse(sp_BooleanFalse, Literal):
         self._precision = precision
 
     def set_precision(self, precision):
-        """ Set precision the false literal """
+        """ Set precision for the false literal """
         self._precision = precision
 
 #------------------------------------------------------------------------------
@@ -65,7 +65,7 @@ class LiteralInteger(sp_Integer, Literal):
         self._precision = precision
 
     def set_precision(self, precision):
-        """ Set precision the inetger literal """
+        """ Set precision for the inetger literal """
         self._precision = precision
 
 #------------------------------------------------------------------------------
@@ -76,7 +76,7 @@ class LiteralFloat(sp_Float, Literal):
         self._precision = precision
 
     def set_precision(self, precision):
-        """ Set precision the float literal """
+        """ Set precision for the float literal """
         self._precision = precision
 
 #------------------------------------------------------------------------------
@@ -94,7 +94,7 @@ class LiteralComplex(Basic, Literal):
         self._precision = precision
 
     def set_precision(self, precision):
-        """ Set precision the complex literal """
+        """ Set precision for the complex literal """
         self._precision = precision
 
     @property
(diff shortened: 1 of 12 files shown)


pyccel/ast/literals.py
```diff
@@ -33,44 +33,64 @@ class Literal(PyccelAstNode):
 class LiteralTrue(sp_BooleanTrue, Literal):
     """Represents the python value True"""
     _dtype     = NativeBool()
-    _precision = default_precision['bool']
+    def __init__(self, precision = default_precision['bool']):
+        self._precision = precision
+
+    def set_precision(self, precision):
```
EmilyBourne, line 39: You can use a setter function:
```
@precision.setter
def precision(self, new_precision)
```
See ast.core for an example

pyccel/ast/literals.py
```diff
@@ -33,44 +33,64 @@ class Literal(PyccelAstNode):
...
 
 #------------------------------------------------------------------------------
 class LiteralFalse(sp_BooleanFalse, Literal):
     """Represents the python value False"""
     _dtype     = NativeBool()
-    _precision = default_precision['bool']
+    def __init__(self,precision = default_precision['bool']):
+        self._precision = precision
+
+    def set_precision(self, precision):
```
EmilyBourne, line 50: See above about setter functions

pyccel/ast/literals.py
```diff
@@ -33,44 +33,64 @@ class Literal(PyccelAstNode):
...
+    _precision = default_precision['integer']
     def __new__(cls, val):
         ival = int(val)
         obj = Expr.__new__(cls, ival)
         obj.p = ival
         return obj
+    def __init__(self, value, *, precision = default_precision['integer']):
+        self._precision = precision
+
+    def set_precision(self, precision):
```
EmilyBourne, line 67: See above about setter functions

pyccel/ast/literals.py
```diff
@@ -33,44 +33,64 @@ class Literal(PyccelAstNode):
...
+        self._precision = precision
 
 #------------------------------------------------------------------------------
 class LiteralInteger(sp_Integer, Literal):
     """Represents an integer literal in python"""
     _dtype     = NativeInteger()
-    _precision = default_precision['int']
+    _precision = default_precision['integer']
```
EmilyBourne, line 58: Is this still needed?

pyccel/ast/literals.py
```diff
@@ -33,44 +33,64 @@ class Literal(PyccelAstNode):
...
 
 #------------------------------------------------------------------------------
 class LiteralFloat(sp_Float, Literal):
     """Represents a float literal in python"""
     _dtype     = NativeReal()
-    _precision = default_precision['real']
+    def __init__(self, value, *, precision = default_precision['float']):
+        self._precision = precision
+
+    def set_precision(self, precision):
```
EmilyBourne, line 78: See above about setter functions

pyccel/ast/literals.py
```diff
@@ -33,44 +33,64 @@ class Literal(PyccelAstNode):
...
+        self._precision = precision
 
 #------------------------------------------------------------------------------
 class LiteralFloat(sp_Float, Literal):
     """Represents a float literal in python"""
     _dtype     = NativeReal()
-    _precision = default_precision['real']
+    def __init__(self, value, *, precision = default_precision['float']):
```
EmilyBourne, line 75: Can you explain why you changed the default precision ( I assume these are dictionary keys have equivalent values?)

pyccel/ast/literals.py
```diff
@@ -33,44 +33,64 @@ class Literal(PyccelAstNode):
...
         return Basic.__new__(cls, real, imag)
 
-    def __init__(self, real, imag):
+    def __init__(self, real, imag, precision = default_precision['float']):
         Basic.__init__(self)
         self._real_part = real
         self._imag_part = imag
+        self._precision = precision
+
+    def set_precision(self, precision):
```
EmilyBourne, line 96: See above about setter functions

pyccel/ast/literals.py
```diff
@@ -33,44 +33,64 @@ class Literal(PyccelAstNode):
...
 class LiteralComplex(Basic, Literal):
     """Represents a complex literal in python"""
     _dtype     = NativeComplex()
-    _precision = default_precision['complex']
 
     def __new__(cls, real, imag):
         return Basic.__new__(cls, real, imag)
 
-    def __init__(self, real, imag):
+    def __init__(self, real, imag, precision = default_precision['float']):
```
EmilyBourne, line 90: Why are you using float precision for complexes?

pyccel/ast/literals.py
```diff
@@ -105,6 +125,9 @@ def __init__(self, arg):
             raise TypeError('arg must be of type str')
         self._string = arg
 
+    def set_precision(self, precision):
```
EmilyBourne, line 133: See above about setter functions

pyccel/parser/syntactic.py
```diff
@@ -50,6 +50,7 @@
 from pyccel.ast.core import _atomic
 from pyccel.ast.core import create_variable
 
+
```
EmilyBourne, line 53: Please checkout changes so wee don't see a file which wasn't changed in the review

EmilyBourne (changes requested):

pyccel/ast/literals.py
```diff
@@ -36,7 +36,12 @@ class LiteralTrue(sp_BooleanTrue, Literal):
     def __init__(self, precision = default_precision['bool']):
         self._precision = precision
 
-    def set_precision(self, precision):
+    @property
+    def precision(self):
+        return self._precision
```
EmilyBourne, line 41: This function should be inherited from `PyccelAstNode`

pyccel/ast/literals.py
```diff
@@ -33,44 +33,93 @@ class Literal(PyccelAstNode):
...
+    def precision(self, precision):
+        """ Set precision for the false literal """
+        self._precision = precision
 
 #------------------------------------------------------------------------------
 class LiteralInteger(sp_Integer, Literal):
     """Represents an integer literal in python"""
     _dtype     = NativeInteger()
-    _precision = default_precision['int']
     def __new__(cls, val):
```
EmilyBourne, line 68: You need to update the new method too I think. To confirm try and pass the `precision` to an instance

pyccel/ast/literals.py
mostafaMamoni: but ```
@precision.setter
def precision(self, new_precision)
``` won't work without it.

mostafaMamoni: ah, or i can inherit `precision` from `PyccelAstNode`.

pyccel/ast/literals.py
mostafaMamoni: what do you mean by "update"?. i searched to what can I do but couldn't find something.

EmilyBourne: Have you tried passing the precision to an instance?

pyccel/parser/syntactic.py
EmilyBourne: This can be done with the command
`git checkout master -- pyccel/parser/syntactic.py`

EmilyBourne (approved):

yguclu:
I am wondering whether the setter for the `precision` property should be added to class `PyccelASTNode`, instead of adding it to all the literal classes.

If not, I would still prefer to add it to the `Literal` base class rather than all its subclasses.

cc: @mostafaMamoni @EmilyBourne @saidctb

EmilyBourne:
> I am wondering whether the setter for the `precision` property should be added to class `PyccelASTNode`, instead of adding it to all the literal classes.
> 
> If not, I would still prefer to add it to the `Literal` base class rather than all its subclasses.
> 
> cc: @mostafaMamoni @EmilyBourne @saidctb

I agree that it should definitely be in the base class rather than the subclasses. I'm not sure that it should be added to `PyccelASTNode` as I am not sure that we should be able to modify this property for all our nodes (e.g I don't think a setter makes sense for `PyccelAdd`)

mostafaMamoni:
Great, i'm with the second option, it will reduce lot of lines and because of the setter method concerns just literal classes in this issue.

yguclu:
> > I am wondering whether the setter for the `precision` property should be added to class `PyccelASTNode`, instead of adding it to all the literal classes.
> > If not, I would still prefer to add it to the `Literal` base class rather than all its subclasses.
> > cc: @mostafaMamoni @EmilyBourne @saidctb
> 
> I agree that it should definitely be in the base class rather than the subclasses. I'm not sure that it should be added to `PyccelASTNode` as I am not sure that we should be able to modify this property for all our nodes (e.g I don't think a setter makes sense for `PyccelAdd`)

Good, let's put the setter into `Literal` then!

yguclu (approved):

PR #554, opened 2020-11-20
Pickle header files' parser

saidctb:
Speed up the parsing of Pyccel header files by pickling the syntactic parser tree to a binary file, which can be quickly loaded whenever the same header file is needed again.

A pickle file named `.pyccel` is generated in the same directory as the  `.pyh` file. It is generated again if it goes out of sync with the header file, or if the Pyccel version is changed.


saidctb:
to tests the speedup you need to execute the following command.
`pyccel dgbtrf.py --libs lapack`  in `pyccel/tests/macro/scripts/lapack`.
you can then observe that a `lapakc.pyccel` is created in `pyccel/pyccel/stdlib/internal`

EmilyBourne:
Can you explain what exactly is being pickled here? Is a pickle created every time we run pyccel? If we run pyccel on a file, then modify that file, then run pyccel again will the changes be picked up or will the picklized version of the file be used instead?

saidctb:
> Can you explain what exactly is being pickled here? Is a pickle created every time we run pyccel? If we run pyccel on a file, then modify that file, then run pyccel again will the changes be picked up or will the picklized version of the file be used instead?

only the `pyh` files are picklized  if the user changes them manually, pyccel will still load the older version, this assumes that the `pyh` are created automatically by pyccel or if you change you need to remove the `.pyccel` file

pyccel/ast/core.py
```diff
@@ -2866,21 +2868,20 @@ def rename(self, newname):
 
         self._name = newname
 
-    def __getnewargs__(self):
-        """used for Pickling self."""
-
+    def __reduce_ex__(self, i):
```
EmilyBourne, line 2871: Missing docstring

pyccel/ast/core.py
```diff
@@ -3818,32 +3818,36 @@ def is_procedure(self):
             or len(set(self.results).intersection(self.arguments)) > 0
         return flag
 
-    def __getnewargs__(self):
-        """used for Pickling self."""
 
+    def __getnewargs__(self):
```
EmilyBourne, line 3822: Missing docstring

pyccel/ast/core.py
```diff
@@ -3818,32 +3818,36 @@ def is_procedure(self):
...
+        'is_recursive':self._is_recursive,
+        'is_pure':self._is_pure,
+        'is_elemental':self._is_elemental,
+        'is_private':self._is_private,
+        'is_header':self._is_header,
+        'arguments_inout':self._arguments_inout,
+        'functions':self._functions}
+        return args, kwargs
+
+    def __reduce_ex__(self, i):
```
EmilyBourne, line 3848: missing docstring

pyccel/ast/core.py
```diff
@@ -6052,6 +6056,13 @@ def get_focus(self):
         else:
             return self.module
 
+    def __reduce_ex__(self, i):
```
EmilyBourne, line 6059: Missing docstring

pyccel/ast/headers.py
```diff
@@ -81,13 +81,8 @@ def name(self):
...
         return self._args[1]
 
-# TODO rename dtypes to arguments
-
-    def __getnewargs__(self):
-        """used for Pickling self."""
-        # TODO improve after renaming the args property
-        args = (self._args[0],)
-        return args
+    def __reduce_ex__(self, i):
```
EmilyBourne, line 84: missing docstring

pyccel/ast/headers.py
```diff
@@ -276,14 +271,13 @@ def vectorize(self,index):
                               self.kind,
                               self.is_static)
 
-    def __getnewargs__(self):
-        """used for Pickling self."""
+    def __reduce_ex__(self, i):
```
EmilyBourne, line 274: missing docstring

pyccel/parser/base.py
```diff
@@ -500,17 +503,118 @@ def insert_import(self, expr):
...
+            f = open(filename, 'rb')
+            hs, version, parser = pickle.load(f)
+            f.close()
+        except:
+            return
+
+        import hashlib
+        code = self.code.encode('utf-8')
+        if hashlib.md5(code).hexdigest() == hs and __version__ == version:
+            self.copy(parser)
```
EmilyBourne, line 596: Is there any way we can test this? Maybe a test with 2 versions of a `.pyh` and a `cp` command to create the `.pyh` file that we are working with here?

pyccel/parser/extend_tree.py
```diff
@@ -19,6 +19,9 @@ def __init__(self, s, lineno, col_offset):
         self.lineno     = lineno
         self.col_offset = col_offset
 
+    def __reduce_ex__(self, i):
```
EmilyBourne, line 22: missing docstring

pyccel/parser/syntactic.py
```diff
@@ -78,6 +78,14 @@
 #      - use OrderedDict whenever it is possible
 from pyccel.errors.messages import *
 
+def get_name(a):
```
EmilyBourne, line 81: missing docstring

EmilyBourne (changes requested):
For some reason codacy seems to have stopped using our .pylintrc file. I've turned it back on so it will probably complain about the missing docstrings itself

saidctb:
> For some reason codacy seems to have stopped using our .pylintrc file. I've turned it back on so it will probably complain about the missing docstrings itself

the issues highlighted by codacy where not created by this PR, some I don't even know how to fix properly

EmilyBourne:
> > the missing docstrings itself
> 
> the issues highlighted by codacy where not created by this PR, some I don't even know how to fix properly

Yes surprisingly it is not complaining about the missing docstrings. Just about use of a private member. To fix it you need to use the getter function, or create one if there isn't one yet. Or if there is no need for a getter then the member should not be private and the underscore should be removed from the start of the name

EmilyBourne (changes requested):
> > the missing docstrings itself
> 
> the issues highlighted by codacy where not created by this PR, some I don't even know how to fix properly

Yes surprisingly it is not complaining about the missing docstrings. Just about use of a private member. To fix it you need to use the getter function, or create one if there isn't one yet. Or if there is no need for a getter then the member should not be private and the underscore should be removed from the start of the name

pyccel/ast/core.py
```diff
@@ -195,6 +195,8 @@
     'Point': Symbol('Point')
 }
 
+#==============================================================================
+def apply(func, args, kwargs):return func(*args, **kwargs)
```
yguclu, line 199: This function needs a docstring...

Maybe what it does is clear, but it is definitely not clear why it is needed and what it is used for.

pyccel/parser/base.py
```diff
@@ -500,17 +503,118 @@ def insert_import(self, expr):
...
+                return
+            else:
+                name = name[:-4]
 
-        print ('------- namespace -------')
-        for (k, v) in self.namespace.items():
-            print ('{var} \t :: \t {dtype}'.format(var=k, dtype=type(v)))
-        print ('-------------------------')
+            name     = '{}.pyccel'.format(name)
+            filename = path+ '/' + name
```
yguclu, line 540: Here we should use `os.path.join`, which is shell-agnostic

pyccel/parser/base.py
```diff
@@ -500,17 +503,118 @@ def insert_import(self, expr):
...
+        try:
+            code = self.code.encode('utf-8')
+            hs   = hashlib.md5(code)
+            f    = open(filename, 'wb')
+            pickle.dump((hs.hexdigest(), __version__, self), f, pickle.HIGHEST_PROTOCOL)
+            f.close()
+        except:
+            return
+
+    # TODO shall we need to load the Parser too?
```
yguclu, line 559: It is not very useful to have this comment here, which looks more like a reminder to self than a proper TODO item. I suggest we have a brief discussion on the topic instead.

pyccel/parser/base.py
```diff
@@ -500,17 +503,118 @@ def insert_import(self, expr):
...
+
+            path , name = os.path.split(self.filename)
+
+            if not name.split('.')[-1] == 'pyh':
+                return
+            else:
+                name = name[:-4]
+
+            name     = '{}.pyccel'.format(name)
+            filename = path+ '/' + name
```
yguclu, line 587: As before, we should use `os.path.join` instead.

pyccel/parser/base.py
```diff
@@ -500,17 +503,118 @@ def insert_import(self, expr):
...
+            raise ValueError('Expecting a .pyccel extension')
+
+#        print('>>> home = ', os.environ['HOME'])
+        # ...
+        try:
+            f = open(filename, 'rb')
+            hs, version, parser = pickle.load(f)
+            f.close()
+        except:
+            return
```
yguclu, line 600: Here both commands, `open` and `pickle.load`, could raise an error, and you do not make sure that the file is always closed afterwards. Instead you should follow https://docs.python.org/3/tutorial/inputoutput.html#reading-and-writing-files:
```python
try:
    with open(filename, 'rb') as f:
        hs, version, parser = pickle.load(f)
except:  # What kind of exception do you expect?
    return
```

pyccel/parser/base.py
yguclu: I agree, we need to test this mechanism somehow. And we should not take for granted that it works on all platforms.

pyccel/parser/parser.py
```diff
@@ -159,15 +178,15 @@ def append_son(self, son):
 
         self._sons.append(son)
 
-    def _parse_sons(self, d_parsers, verbose=False):
+    def parse_sons(self, d_parsers, verbose=False):
         """Recursive algorithm for syntax analysis on a given file and its
         dependencies.
         This function always terminates with an OrderedDict that contains parsers
         for all involved files.
         """
```
yguclu, line 186: Given that this method is no longer private, please add the "Parameters" and "Returns" sections to its docstring.

pyccel/parser/parser.py
```diff
@@ -89,28 +107,29 @@ def module_parser(self):
         return self._module_parser
 
     def parse(self, d_parsers=None, verbose=False):
```
yguclu, line 109: Can you please add a short docstring, complete of the "Parameters" and "Returns" sections?

pyccel/parser/base.py
```diff
@@ -500,17 +512,118 @@ def insert_import(self, expr):
                         container[source] = []
                     container[source] += name
 
-    def print_namespace(self):
+    def dump(self, filename=None):
```
yguclu, line 515: It is not clear to me how the "filename" argument is used in this function. Can you explain please?

yguclu (changes requested):

pyccel/parser/base.py
saidctb: if a filename is given you can then dump the parser in that file, if not you take the same name of the file that you are parsing

pyccel/ast/core.py
saidctb: done

pyccel/ast/core.py
saidctb: done

pyccel/ast/headers.py
saidctb: done

pyccel/ast/headers.py
saidctb: done

pyccel/ast/core.py
saidctb: done

pyccel/ast/core.py
saidctb: done

pyccel/parser/extend_tree.py
saidctb: done

pyccel/ast/core.py
saidctb: it's used by pickle's methods that I have added, you can find the doc strings there

pyccel/parser/base.py
saidctb: done

pyccel/parser/base.py
saidctb: done

pyccel/parser/base.py
saidctb: I forgot to delete it, it's not relevant anymore

pyccel/parser/base.py
saidctb: done

pyccel/parser/parser.py
saidctb: done

pyccel/parser/parser.py
saidctb: done

pyccel/parser/base.py
saidctb: it can be tested if we change the version so I'm not sure how can we change the version

saidctb: the user needs to update the repository and then it will be tested

EmilyBourne: So we only recreate the file if we update the pyccel version? Is that necessary? I thought the `.pyh` files let you interact with any fortran code. So in theory a user could create their own. Given that fact I'm not sure it makes sense to link this to the pyccel version

saidctb: It's necessary if we change something on the way we treat the header files, imagine we change the syntax in the new version or we change the something the outcome of the parser in the new version, so there is no way for us the be sure

saidctb: this part of code deals with `.pyccel` files and not `-pyh` files

EmilyBourne: Yes, but the `.pyccel` files are generated from the `.pyh` files aren't they?

My point was that I might want to update the `.pyccel` file between version changes, not that we shouldn't update it when we update the version

saidctb: yes this make sure that we update them when we change the version or the pyh file

EmilyBourne: In that case we should be able to test it as I suggested. You create a file containing:
- test_file_v1.pyh
- test_file_v2.pyh
- test_script.py

In test_script.py you do:
```
cp test_file_v1.pyh test_file.pyh
pyccel test_file.pyh
# Test output expected from v1
cp test_file_v2.pyh test_file.pyh
pyccel test_file.pyh
# Test output expected from v2
```

saidctb: done

bandit.yml
```diff
@@ -7,3 +7,7 @@ skips:
   - B101  # Ignore assert statements
   - B404  # Ignore warnings about importing subprocess
   - B603  # Ignore warnings about calling subprocess.Popen without shell=True
+  - B403  # Ignore warnings about import pickle
+  - B301  # Ignore warnings about pickle.load
+  - B303  # Ignore warnings about MD2, MD4, MD5, or SHA1 hash functions
+  - W0703 # Ignore warnings about broad-except
```
EmilyBourne, line 13: Can we not add this to the bandit file please. I think that it is a very valid warning. We should never do `except:` or `except Exception:` as it catches everything including typos which makes debugging really hard

pyccel/parser/base.py
```diff
@@ -500,17 +512,112 @@ def insert_import(self, expr):
...
+        # we are only exporting the AST.
+        try:
+            code = self.code.encode('utf-8')
+            hs   = hashlib.md5(code)
+            with open(filename, 'wb') as f:
+                pickle.dump((hs.hexdigest(), __version__, self), f, pickle.HIGHEST_PROTOCOL)
+        except FileNotFoundError:
+            pass
+        except PermissionError:
+            pass
```
EmilyBourne, line 558: You can group these:
```
except (FileNotFoundError, PermissionError):
    pass
```

pyccel/parser/base.py
```diff
@@ -500,17 +512,112 @@ def insert_import(self, expr):
...
+            raise ValueError('Expecting a .pyccel extension')
 
-        print ('------- namespace -------')
-        for (k, v) in self.namespace.items():
-            print ('{var} \t :: \t {dtype}'.format(var=k, dtype=type(v)))
-        print ('-------------------------')
+        try:
+            with open(filename, 'rb') as f:
+                hs, version, parser = pickle.load(f)
+        except Exception:
```
EmilyBourne, line 593: I agree with codacy (once I switched back to our pylint config file). We should never catch `Exception` unless we are raising an error. Maybe you want `OSError` here?

EmilyBourne (changes requested):

pyccel/parser/base.py
saidctb: the exception depends on pickle , it can be what ever exception pickle raises

bandit.yml
saidctb: yes if we can solve the problem of pickle

pyccel/parser/base.py
EmilyBourne: In that case shouldn't the try/except just be around the pickle line?
https://pythontic.com/modules/pickle/exceptions#:~:text=Exceptions%20provided%20by%20the%20Python%20Pickle%20Module%3A&text=PicklingError%20is%20used%20when%20the,is%20not%20able%20to%20proceed.&text=This%20is%20exception%20pickle.,object%20hierarchie(s)%20fails.
It looks like it should be `except pickle.PickleError`

saidctb: so we either keep it, or delete the try except clause

bandit.yml
EmilyBourne: If we have a case where it is necessary we can always disable the check at that line with a comment. Disabling it for the whole project is a bad idea because it stops the error being raised for this in the future

pyccel/parser/base.py
saidctb: it can also be the fact that the file doesn't exist or that we don't have the permission

pyccel/parser/base.py
saidctb: done

bandit.yml
saidctb: done

pyccel/parser/base.py
saidctb: I added the pickle error , so it should be fine

tests/pyccel/test_pyccel.py
```diff
@@ -508,3 +509,63 @@ def test_print_strings(language):
...
+    cwd = os.path.dirname(test_file)
+    cwd = get_abs_path(cwd)
+
+    pyccel_commands = " --language="+language
+
+    compile_pyccel(cwd, test_file, pyccel_commands)
+
+
+    lang_out = get_lang_output(get_exe(test_file))
+    assert int(lang_out)
```
EmilyBourne, line 547: Shouldn't we be checking the type of `lang_out` to make sure that the right type is used?

tests/pyccel/test_pyccel.py
```diff
@@ -508,3 +509,63 @@ def test_print_strings(language):
...
+
+    with open(test_file, 'w') as f:
+        code = ("from headers import f\n"
+                "def f(x):\n"
+                "    y = x\n"
+                "    return y\n"
+                "print(f(1.))\n")
```
EmilyBourne, line 554: Is this the line that lets us check that the mechanism is working?

EmilyBourne: I checked locally with the code:
```
#$ header function f(int)
def f(x):
    y = x
    return y

print(f(1.))
```
I get a compilation error on fortran, but it works for c.

I can get it to fail if you pass: `--flags="-Wconversion -Werror` and pass `1.5` instead of `1.`. Apparently gcc is clever enough to see that that isn't really truncation

EmilyBourne: The same flags are valid for fortran

tests/pyccel/test_pyccel.py
EmilyBourne: Never mind. This is a string anyway. The test comes from the compilation

tests/pyccel/test_pyccel.py
saidctb: of course if you take an int argument and you pass it a float, it will give you that error, the test doesn't do that

EmilyBourne: I thought that the test did that if the .pyccel file wasn't updated (ie. if it fails). If that isn't the case then can you tell me what exactly tells us that this is working?

saidctb: the fact that we are not getting that error  in github actions means that the test is working

saidctb: so I create a header file and then change it to be sure that the I don't upload the old one

EmilyBourne: >the fact that we are not getting that error in github actions means that the test is working

In that case my point is valid. If the test wasn't working then we would have the code that I printed above. This code raises an error when compiled with pyccel and fortran. But it doesn't raise an error when compiled with pyccel and c. This means that we would think that the test was working when in fact we had the same problem.

If you add the flags that I indicated then the test will still pass, but it will fail for c if there is a problem

saidctb: I'm not testing if it fails to upload or not it always uploads, I'm testing if it uploads the right one after chaning the `-pyh` file

saidctb: yes but I'm never running the code that you printed

tests/pyccel/test_pyccel.py
```diff
@@ -247,7 +248,7 @@ def teardown(path_dir = None):
             shutil.rmtree( file_name )
         elif not os.path.isfile(file_name):
             teardown(file_name)
-        elif not f.endswith(".py"):
+        elif not f.endswith(".py") and not f.endswith(".pyh") and not f.endswith(".pyccel"):
```
EmilyBourne, line 251: Shouldn't we delete the `.pyccel` file after the test to be sure that we have a clean environment? That way we don't need to test if the integer case has a float header (probably harder to test)

EmilyBourne: Equally if you write the `.pyh` file each time then can't it be deleted after the test?

tests/pyccel/test_pyccel.py
saidctb: done

tests/pyccel/test_pyccel.py
saidctb: lol the whole purpose of the test is to keep the `.pyccel` file

saidctb: if the integer case has a float header, then `int(lang_out)` will raise an error, because you are doing something like
`int('1.')`

EmilyBourne: This is called after the test. Obviously the `.pyccel` file needs to be kept during the test between the `int` case and the `float` case, but should it be kept between the `float` case and a potential second run starting with the `int` case?

saidctb: it should be fine

EmilyBourne: > if the integer case has a float header, then int(lang_out) will raise an error, because you are doing something like
int('1.')

As long as it is printed in the correct format then yes. This seems to be the case for both c and fortran

EmilyBourne (approved):

yguclu (approved):

yguclu (changes requested):
@saidctb This PR looks really good, this is a very useful feature to have!

We need a bit of documentation for both users and developers. The PR description does not explain which files are created, where, and when. In fact this is not clear to me yet.

I think you should create a short .md text file that explains how this pickling works. It may take a bit of time now, but it will save you a lot of time later because you will not need to explain it any more.

yguclu:
@saidctb It appears that your changes in `pyccel/ast/core.py` are in conflict with other changes that have taken place since you created this PR.

Can you take care of fixing the conflicts? Do you need help?

EmilyBourne:
> [no-file-name-irregular-characters] Do not use `_` in a file name
!! Can we disable this? Sounds like nonsense to me

tutorial/cache_pyh_files.md
```diff
@@ -0,0 +1,33 @@
+# Picklize header files
+
+## header files
+
+A header file in pyccel is a file containing function or variable declarations, macro definitions , templates and metavariable declarations.
+the convention is to give header files names that end with `.pyh` .
```
EmilyBourne, line 6: Is this a convention or a rule? A convention is something that is usually done but doesn't have to be whereas a rule needs to be followed

tutorial/cache_pyh_files.md
```diff
@@ -0,0 +1,33 @@
...
+Header files serve two purposes:
+- link externel libraries in the targeted languages by providing their function definitions.
```
EmilyBourne, line 8: typo `externel`->`external`

tutorial/cache_pyh_files.md
```diff
@@ -0,0 +1,33 @@
...
+- accelerate the parssing process by parsing the header file instead of the original file in the case of pyccelizing multiple files.
```
EmilyBourne, line 9: Typo `parssing` -> `parsing`

tutorial/cache_pyh_files.md
```diff
@@ -0,0 +1,33 @@
...
+```python
+from header import omp_get_num_threads
+print(omp_get_num_threads())
+```
+Pyccel can compile the python file with the following command: `pyccel openmp.py --language fortran --flags="-fopenmp"`
+it will then create the executable file `openmp`
+
+## picklizing the header file
+When compiling a header file pyccel will generate in the same directory `.pyccel` file that contains the cached result of the parser, 
+This will accelerate the compiling process of big header files, by compiling them only once and storing the results for futur compilation, Pyccel will generate a new
```
EmilyBourne, line 31: Typo `futur` -> `future`

tutorial/cache_pyh_files.md
```diff
@@ -0,0 +1,33 @@
...
+
+```python
+from header import omp_get_num_threads
+print(omp_get_num_threads())
+```
+Pyccel can compile the python file with the following command: `pyccel openmp.py --language fortran --flags="-fopenmp"`
+it will then create the executable file `openmp`
+
+## picklizing the header file
+When compiling a header file pyccel will generate in the same directory `.pyccel` file that contains the cached result of the parser,
```
EmilyBourne, line 30: This is not what the PR description says. Presumably it needs updating?

yguclu: At some point I updated the PR description based on what I understood at the time. But maybe I was wrong...

tutorial/cache_pyh_files.md
saidctb: it's a rule okay I will change it

saidctb:
> > [no-file-name-irregular-characters] Do not use `_` in a file name
> 
> !! Can we disable this? Sounds like nonsense to me

yes I will disable it

tutorial/cache_pyh_files.md
saidctb: done

tutorial/cache_pyh_files.md
saidctb: done

tutorial/cache_pyh_files.md
```diff
@@ -0,0 +1,32 @@
+# Picklize Header files
```
yguclu, line 1: I think that we should name this file `headers.md`, and change the title to `Pyccel header files`. We can then skip the first subtitle altogether.

tutorial/cache_pyh_files.md
```diff
@@ -0,0 +1,32 @@
...
+
+#$ header function omp_get_num_threads() results(int)
+```
+We then create `openmp.py` file:
+
+```python
+from header import omp_get_num_threads
+print('number of threads is :', omp_get_num_threads())
+```
+Pyccel can compile the Python file with the following command: `pyccel openmp.py --language fortran --flags="-fopenmp"`
```
yguclu, line 26: Don't we pass the `-fopenmp` flag automatically?

tutorial/cache_pyh_files.md
```diff
@@ -0,0 +1,32 @@
...
+, It will then create the executable file `openmp`
+
+## Picklizing header files
+Pyccel uses the Python Module [pickle](https://docs.python.org/3/library/pickle.html) to cache the header files.\
+When compiling a header file Pyccel will generate in the same directory a `.pyccel` file that contains the cached result of the parser,\
+This will accelerate the compiling process of big header files, by compiling them only once and storing the results for future compilation, Pyccel will generate a new `.pyccel` when after the modification the header file  or downloading a new Pyccel version.
```
yguclu, line 32: We should clarify the language a bit here. For example, say that Pyccel is "processing" a file instead of "compiling" them. Otherwise, we should say at the beginning what we mean by "compiling".

yguclu (changes requested):

tutorial/cache_pyh_files.md
saidctb: it's the same the difference is in the code generation

tutorial/cache_pyh_files.md
saidctb: this is a general example pyccel doesn't know that it's an openmp

pyccel/ast/core.py
```diff
@@ -3290,34 +3311,60 @@ def is_procedure(self):
...
+
+          out : tuple
+           A tuple of two elements
+           a callablle function that can be called
+           to create the initial version of the object
+           and its arguments.
+        """
+        args, kwargs = self.__getnewargs__()
+        out = (apply, (self.__class__, args, kwargs))
+        return out
```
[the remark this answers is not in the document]
saidctb: they don't work, we need to use the low level __reduce__

tutorial/cache_pyh_files.md
saidctb: we already have another file named `headers.md`

pyccel/ast/core.py
```diff
@@ -2392,21 +2396,39 @@ def rename(self, newname):
...
+
+          i : int
+           protocol
 
+          Results
+          -------
+
+          out : tuple
+           A tuple of two elements
+           a callablle function that can be called
```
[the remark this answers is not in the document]
saidctb: done

tutorial/cache_pyh_files.md
yguclu: But if the user writes
```bash
$ pyccel --help
```
he/she will discover that we have an `--openmp` flag available.

EmilyBourne: Not in pyccel. Maybe with epyccel, but we are passing `accelerator='openmp'` everytime in `tests_epyccel_openmp.py` so I'm guessing not

tutorial/cache_pyh_files.md
EmilyBourne: I agree. For me "compiling" means using a compiler. e.g gcc
I would use either "processing" or "pyccelizing"

pyccel/ast/core.py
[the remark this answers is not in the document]
EmilyBourne: Can you explain why they don't work?

tutorial/cache_pyh_files.md
yguclu: Ah, that's true! But that is about function headers and decorators, not about header files.

Maybe this file should be named `header_files.md` then?

tutorial/cache_pyh_files.md
saidctb: yes, don't focus too much on the example. I could have used another library or created my own fortran function and then link it using the same mechanism

pyccel/ast/core.py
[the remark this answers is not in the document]
yguclu: Why don't they work for us? The low-level reduce that you wrote looks really simple to me...

tutorial/cache_pyh_files.md
saidctb: okay

tutorial/cache_pyh_files.md
yguclu: > it's the same the difference is in the code generation

?

saidctb: I mean Pyccel is a compiler , so what it does is compiling python code and generating a `.so`, even for c programs , the c compiler generates assembly code and then the assembly compiles it to generate machine code

pyccel/ast/core.py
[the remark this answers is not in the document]
saidctb: I got errors due to sympy , yes it's much better for me too

tutorial/cache_pyh_files.md
yguclu: I am going to change the explanation myself. Let me know if I write anything wrong.

saidctb: okay

tutorial/header-files.md
```diff
@@ -0,0 +1,35 @@
...
+from header import omp_get_num_threads
+print('number of threads is :', omp_get_num_threads())
+```
+Pyccel can compile the Python file with the following command: `pyccel openmp.py --language fortran --flags="-fopenmp"`
+, It will then create the executable file `openmp`
+
+## Picklizing header files
+Parsing a large Pyccel header file with hundreds of function declarations may require a significant amount of time, therefore it is important that this process is only done once when pyccelizing multiple Python source files in a large project.
+
+To this end, Pyccel uses the [pickle](https://docs.python.org/3/library/pickle.html) Python module to store the result of the parser to a `.pyccel` binary file, which is created in the same directory of the header file.
```
EmilyBourne, line 31: Should be: `same directory as the header file`

yguclu: Just fixed. Thanks

EmilyBourne (approved):

yguclu (approved):

PR #555, opened 2020-11-20
Clean C wrapper 

MenoIy:
Make the C wrapper more readable and ready for all the future features and modifications.
- get_PyArgParseType and get_PyBuildValue are responsible only for the generation of collect variable and cast function needed.
- create function _get_static_function responsible for converting function call to static function call needed for arguments with rank greater than 0 in Fortran.
- change the Type_check function name to PythonType_check to make difference between the incoming check function.
- create _body_management function responsible for calling functions that take care of body creation.
- creat multiple functions responsible for collecting value and managing error and create the body
      _body_optional_variable responsible of optional variable .
      _body_valued_variable responsible of valued variable.
      _body_array responsible of arguments with rank greater than 0.
- add documentation to each one of those functions.


pyccel/ast/cwrapper.py
```diff
@@ -372,16 +372,16 @@ def args(self):
                         ('complex',8)  : numpy_cdouble_type,
                         ('complex',16) : numpy_clongdouble_type}
 
-def PyType_Check(data_type):
+def PythonType_Check(variable, argument):
```
EmilyBourne, line 375: Docstring please

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "files.associations": {
+        "ostream": "cpp"
+    }
+}
```
EmilyBourne, line 5: Did you mean to commit this?

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -159,7 +198,170 @@ def get_cast_function_call(self, cast_type, arg):
...
+        """
+        Responsible for collecting value and managing error and create the body
+        of optional arguments in format
+                if (pyobject == Py_None){
+                    collect Null
+                }else if(Type Check == False){
+                    Print TypeError Wrong Type
+                    return Null
+                }else{
+                    assigne pyobject value to tmp variable
```
EmilyBourne, line 214: typo `assigne`->`assign`

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -159,7 +198,170 @@ def get_cast_function_call(self, cast_type, arg):
...
+        check_type : Boolean
+            True if the type is needed
+
+        Returns
+        -------
+        body : list
+            A list of statements
+        """
+        body = [(PyccelEq(VariableAddress(collect_var), VariableAddress(Py_None)),
+                [Assign(VariableAddress(variable), variable.value)])]
```
EmilyBourne, line 234: Does this translate to:
```
int* a;
a = 4;
```
? Or did I miss something

EmilyBourne: Do we need to allocate memory for this pointer?

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -159,7 +198,170 @@ def get_cast_function_call(self, cast_type, arg):
...
+        -------
+        body : list
+            A list of statements
+        """
+        body = [(PyccelEq(VariableAddress(collect_var), VariableAddress(Py_None)),
+                [Assign(variable, variable.value)])]
+        if check_type : # Type check
+            check = PyccelNot(PythonType_Check(variable, collect_var))
+            error = PyErr_SetString('PyExc_TypeError', '"{} must be {}"'.format(variable, variable.dtype))
+            body += [(check, [error, Return([Nil()])])]
```
EmilyBourne, line 276: This seems to be duplicated in a few places right?

.vscode/settings.json
MenoIy: sorry, I forgot it again.

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: Maybe we can create a function `check_for_error(condition, err_msg)`?

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -159,7 +198,170 @@ def get_cast_function_call(self, cast_type, arg):
...
+        #rank check :
+        check = PyccelNe(FunctionCall(numpy_get_ndims,[collect_var]), LiteralInteger(collect_var.rank))
+        error = PyErr_SetString('PyExc_TypeError', '"{} must have rank {}"'.format(collect_var, str(collect_var.rank)))
+        body  = [(check, [error, Return([Nil()])])]
+
+        if check_type : #Type check
+            numpy_dtype = self.find_in_numpy_dtype_registry(variable)
+            arg_dtype   = self.find_in_dtype_registry(self._print(variable.dtype), variable.precision)
+            check = PyccelNe(FunctionCall(numpy_get_type, [collect_var]), numpy_dtype)
+            info_dump = PythonPrint([FunctionCall(numpy_get_type, [collect_var]), numpy_dtype])
```
EmilyBourne, line 322: Did I leave the `info_dump` in the master branch?! Oops^^

EmilyBourne: Although, maybe some of this is useful. It would make sense to put it in SetString but we can't do that yes

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -172,68 +374,40 @@ def get_PyArgParseType(self, used_names, variable, argument):
...
         """
-        body = []
+        cast_function = None
         collect_var = variable
 
-        if variable.is_optional:
+        if variable.rank > 0:
+            collect_type = PyccelPyArrayObject()
+            collect_var = Variable(dtype= collect_type, is_pointer = True, rank = variable.rank,
+                                    order= variable.order, name=self.get_new_name(used_names, variable.name+"_tmp"))
```
EmilyBourne, line 391: Does this need to be after `is_optional`? I don't think I tested an optional array

EmilyBourne (changes requested):

pyccel/codegen/printing/cwrappercode.py
MenoIy: Optional variable here value is NULL so its translated to :  
```
int *a;
a = NULL;
```

pyccel/codegen/printing/cwrappercode.py
MenoIy: I was thinking the same but I don't think we can raise 2 errors, maybe we can change error code format

pyccel/codegen/printing/cwrappercode.py
MenoIy: this can be a good idea but this function will be working only in the functions that create body.

MenoIy: after all, i don't think if it is good to create the function the code will be less comprehensible and I prefer that we leave the check and the error creation separated for the moment as we don't know what kind of error we will encounter.

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: Ah of course. I don't think this is the first time I've made this mistake. Maybe it would be better to put `Nil()` then it's clear

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: I meant a function that returns an If not a tuple, but you might be right, and that doesn't save many (any?) lines anyway

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: Maybe just leave it as it is until strings in c are more advanced

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: Hmm, I guess it might work if a `PyArrayObject` can be set to `PyNone`. But I'm not sure that's possible. We should test. But we'll have to add whatever function is needed to case a `PyArray` to a `PyArrayObject` if it doesn't work :/

MenoIy: 1 - PyArrayObject can't be set to PyNone.
2 - I can set the default value to NULL  and check it after PyArg_ParseTupleAndKeywords
3 - If I do that if a user call function like this ```func(None)``` it won't work   PyArg_ParseTupleAndKeywords will raise a Type error.
4 - if it's called like this ```func()```I think it will work because the Pyarrayobject won't be touched in  the  PyArg_ParseTuple
and he will keep the default value NULL

EmilyBourne: That sounds right to me. We have never handled passing Py_None anyway

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -110,7 +110,9 @@ def find_in_numpy_dtype_registry(self, var):
                     severity='fatal')
 
     def get_default_assign(self, arg, func_arg):
-        if func_arg.is_optional:
+        if arg.rank > 0 :
+            return AliasAssign(arg, Nil())
```
EmilyBourne, line 114: Can you add a test please

MenoIy: ok

yguclu (approved):

tests/epyccel/test_epyccel_optional_args.py
```diff
@@ -132,3 +135,54 @@ def test_f11(language):
...
+            x[:] = x // y
+        f = epyccel(f12, language = language)
+
+        x1 = np.array( [1,2,3], dtype=np.int )
+        x2 = np.copy(x1)
+        f(x1)
+        f12(x2)
+
+        # ...
+        assert np.array_equal(x1, x2)
```
EmilyBourne, line 162: This indenting doesn't seem to be right

tests/epyccel/test_epyccel_optional_args.py
```diff
@@ -132,3 +135,54 @@ def test_f11(language):
...
+            x[:] = x // y
+        f = epyccel(f13, language = language)
+
+        x1 = np.array( [[1,2,3], [4,5,6]], dtype=np.int32, order='F' )
+        x2 = np.copy(x1)
+        f(x1)
+        f13 (x2)
+
+        # ...
+        assert np.array_equal(x1, x2)
```
EmilyBourne, line 187: This indenting doesn't seem to be right

EmilyBourne (approved):

PR #556, opened 2020-11-20
Run unit tests with GitHub Actions

EmilyBourne:
Tests working on macOS and Linux. Not yet on Windows. 

TODO:
- Set environment variables (important for codacy coverage). Use secrets. See https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets
- Find why there are no codacy files to combine. See https://github.com/marketplace/actions/codacy-coverage-reporter


saidctb:
we can use travis for windows probably and the rest with github actions

yguclu:
> we can use travis for windows probably and the rest with github actions

I agree. For now we should turn off Windows tests until Travis-CI builds are available again.

.github/workflows/main.yml
```diff
@@ -0,0 +1,97 @@
...
+      - name: Test with pytest
+        uses: ./.github/actions/pytest_run
+      - name: Collect coverage information
+        continue-on-error: True
+        uses: ./.github/actions/coverage_collection
+
+  Linux-Master:
+
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/master'
```
yguclu, line 41: Can the `if` condition be put in the strategy matrix?

yguclu: Sorry, probably not... I don't think they support sparse matrices! 😅

EmilyBourne: You can put an `includes` or `excludes` s section but for some reason you can't put an if in that

EmilyBourne:
Do you want to merge this now so we have some tests and fix the remaining things later?
Windows setup is now working, but it won't compile. I suspect missing environment variables

yguclu:
> Do you want to merge this now so we have some tests and fix the remaining things later?
> Windows setup is now working, but it won't compile. I suspect missing environment variables

Yes, let's just merge this!
Can you please rename file `main.yml` with something more meaningful?

yguclu (approved):

PR #559, opened 2020-11-24
Improve operator handling

### LINKED ISSUE #410: Add brackets when mixing `and` and `or` operations for stylistic reasons

### LINKED ISSUE #529: Fix Fortran printer of numpy.random.randint
Bug description
-----------------
The Fortran printer generates invalid code for `numpy.random.randint` if the `low` parameter is an integer literal.

To reproduce
-----------------
The Python code
```python
import numpy as np

x = np.random.randint(-10, 10)
```
is translated to
```fortran
integer(kind=8) :: x
real(kind=8) :: Dummy_0001

call random_number(Dummy_0001)

x = floor((10_8 - -10_8) * Dummy_0001 + -10_8, kind=8)
```
which cannot compile because of two consecutive signs (`- -` and `+ -`).

Expected behavior
-----------------
The last line of code should be
```fortran
x = floor((10_8 - (-10_8)) * Dummy_0001 + (-10_8), kind=8)
```
or, much better,
```fortran
x = floor((10_8 + 10_8) * Dummy_0001 - 10_8, kind=8)
```

Language
-----------------
Fortran

Additional context
-----------------
This is possibly caused by an issue with the AST node `PyccelMinus`, or its printer function.

EmilyBourne:
Create an `ast.operators` file containing all python operators. Operators in this file have a name beginning with `Pyccel`. Class inheritance is used abundantly to reduce code duplication. Precedence handling is now carried out at all stages in case instances are created on the fly (fixes problem reported in PR 410)

**Commit Summary**
- Move operators into an `ast.operators` file
- Rename `PyccelOperator`->`PyccelBinaryOperator`. Create super class `PyccelOperator`
- Create super classes `PyccelBooleanOperator`, `PyccelComparisonOperator`, `PyccelUnaryOperator`, `PyccelBitOperator`, `PyccelArithmeticOperator`
- Split `__init__` into 3 steps:
   1. `_handle_precedence`
   2. `_set_dtype`
   3. `_set_shape_rank`
- Define `_set_dtype` and `_set_shape_rank` in specialized classes
- Handle repeated unary operators (This stops ++a being written)
- Add specialised `_handle_precedence` method to `PyccelAnd`, `PyccelOr`, and `PyccelArithmeticOperator`. Fixes #410
- Rename to match naming convention `Is` -> `PyccelIs`, `IsNot` -> `PyccelIsNot`
- Simplify `sympy_to_pyccel` function using the fact that `PyccelAdd`/`PyccelMinus` is always binary¨
- Remove unnecessary `PyccelMinus` inheritance from `PyccelAdd` (`PyccelAdd` is more specialised than `PyccelMinus`)
- Use new superclasses to simplify semantic parser
- Add test to show PR fixes #529


saniQQ (approved):

nhamidn (approved):

pyccel/ast/operators.py
```diff
@@ -0,0 +1,949 @@
+"""
+Module handling alll python builtin operators
```
yguclu, line 2: Typo...

yguclu (approved):

pyccel/ast/operators.py
```diff
@@ -0,0 +1,949 @@
...
+
+                self._shape = shape
+                self._rank  = len(shape)
+            else:
+                self._rank  = max(a.rank for a in self._args)
+                self._shape = [None]*self._rank
+
+#==============================================================================
+
+class PyccelArithmeticOperator(PyccelBinaryOperator):
```
EmilyBourne, line 391: Is anything a `PyccelBinaryOperator` but not a `PyccelArithmeticOperator`?

EmilyBourne: Yes. Comparisons

yguclu: There is a `PyccelComparisonOperator`...

PR #561, opened 2020-11-25
Use equality operator for non Logical variables.

### LINKED ISSUE #560: The NOT operator does not work in fortran in the case of INTEGERS
**Describe the bug**
The python NOT operator does not work properly in the case of an integer for fortran.

**To Reproduce**
```python
x = 1

if not x:
    print("yes!")
```

the generated code is :
```bash/fortran/c
program prog_test

use ISO_C_BINDING

implicit none

integer(C_LONG_LONG) :: x

x = 1_C_LONG_LONG
if (.not. x) then
  print *, '' // 'yes!'
end if

end program prog_test
```

this generated code fail to compile.

```
   10 | if (.not. x) then
      |         1
Error: Operand of .not. operator at (1) is INTEGER(8)
```

i think if the `x` is not a `Boolean` then we need to cast `x` to `bool`.

**Language**
Fortran

nhamidn:
When the variable `x` is not a `bool`, we now print `not x` as `x == 0` in Fortran. This fixes #560.

### DIFF
--- a/pyccel/codegen/printing/fcode.py
+++ b/pyccel/codegen/printing/fcode.py
@@ -2349,6 +2349,8 @@ class FCodePrinter(CodePrinter):
 
     def _print_PyccelNot(self, expr):
         a = self._print(expr.args[0])
+        if (expr.args[0].dtype is not NativeBool()):
+            return '{} == 0'.format(a)
         return '.not. {}'.format(a)
 
     def _print_Header(self, expr):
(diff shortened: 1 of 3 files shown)


EmilyBourne (approved):

EmilyBourne (approved):

yguclu (approved):

yguclu:
When the variable `x` is not a `bool`, we now print `not x` as `x == 0` in Fortran. This fixes #560.

PR #562, opened 2020-11-25
Enable including of Pyccel internal libraries

rakati:
**Code generation mechanism:**

- Check for necessary C/Fortran internal libraries to include;
- Remove the lib folder in `__pyccel__` if it exists to avoid missing files;
- Get a fresh copy of the lib from pyccel and compile it, then return the objects files and the path to include;
- Add internal lib object files and the lib path respectively to `dep_mods` and `inc_dirs` for correct compilation.

**Commit summary:**

- Create a small library called `pyc_math` with some basic math functions that do not exist in C and Fortran languages;
- Add printers for the additional math functions in `pyc_math`;
- Add epyccel tests for these functions;
- Add pyccel test for ndarray library;
- Fix casting for literal integer in `numpy.full` for C language.


pyccel/codegen/printing/ccode.py
```diff
@@ -778,6 +778,43 @@ def _print_MathSqrt(self, expr):
...
         return 'sqrt({})'.format(code_args)
 
+    # -------------- pyc_math functions ---------------------------------------
+    def _print_MathFactorial(self, expr):
+        """ Print the equivalent code to pycthon factorial function call"""
+        # include of the pyc_math header
+        self._additional_imports.add('pyc_math')
+        arg = expr.args[0]
+        if arg.dtype is not NativeInteger() or len(expr.args) > 1:
+            raise ("factoral function accept one integer argument")
```
EmilyBourne, line 788: Can you explain why this is a C specific problem? Looking at the docstring for `math.factorial` it seems that this rule is true in the original python, so it should not matter which language we are printing in. In that case this kind of check should be done in the `MathFactorial` constructor

pyccel/codegen/printing/ccode.py
```diff
@@ -778,6 +778,43 @@ def _print_MathSqrt(self, expr):
...
+        return 'pyc_factorial({})'.format(self._print(arg))
+
+    def _print_MathGcd(self, expr):
+        """ Print the equivalent code to pycthon gcd function call"""
+        # include of the pyc_math header
+        self._additional_imports.add('pyc_math')
+        args = expr.args
+        if len(args) != 2 or not all(arg.dtype is NativeInteger() for arg in args):
+            raise ("gcd function accept two integer arguments")
```
EmilyBourne, line 797: See above. Are you sure this is a problem that only occurs in C?

pyccel/codegen/printing/ccode.py
```diff
@@ -778,6 +778,43 @@ def _print_MathSqrt(self, expr):
...
+        return 'pyc_gcd({}, {})'.format(self._print(args[0]), self._print(args[1]))
+
+    def _print_MathDegrees(self, expr):
+        """ Print the equivalent code to pycthon degrees function call"""
+        # include of the pyc_math header
+        self._additional_imports.add('pyc_math')
+        arg = expr.args[0]
+        if len(expr.args) != 1 or arg.dtype is NativeComplex():
+            raise ("degrees function accept one float or integer argument")
```
EmilyBourne, line 806: See above. Are you sure this is a problem that only occurs in C?

pyccel/codegen/printing/ccode.py
```diff
@@ -778,6 +778,43 @@ def _print_MathSqrt(self, expr):
...
+        return 'pyc_degrees({})'.format(self._print(arg))
+
+    def _print_MathRadians(self, expr):
+        """ Print the equivalent code to pycthon radians function call"""
+        # include of the pyc_math header
+        self._additional_imports.add('pyc_math')
+        arg = expr.args[0]
+        if len(expr.args) != 1 or arg.dtype is NativeComplex():
+            raise ("radians function accept one float or integer argument")
```
EmilyBourne, line 815: See above. Are you sure this is a problem that only occurs in C?

pyccel/stdlib/ndarrays/ndarrays.h
```diff
@@ -67,7 +71,11 @@ t_ndarray   array_create(int nd, int *shape, enum e_types type);
 void        _array_fill_int(int c, t_ndarray arr);
 void        _array_fill_float(float c, t_ndarray arr);
 void        _array_fill_double(double c, t_ndarray arr);
-void        _array_fill_cdouble(complex double c, t_ndarray arr);
+void        _array_fill_bool(bool c, t_ndarray arr);
+void        _array_fill_sint(short int c, t_ndarray arr);
+void        _array_fill_long(long c, t_ndarray arr);
+void        _array_fill_cfloat(float complex c, t_ndarray arr);
+void        _array_fill_cdouble(double complex c, t_ndarray arr);
```
EmilyBourne, line 78: Where are the definitions of these functions?

pyccel/stdlib/pyc_math/pyc_math.c
```diff
@@ -0,0 +1,43 @@
...
+{
+    long    res = 1;
+
+    /* ValueError: factorial() not defined for negative values */
+    if (n < 0)
+        return 0;
+    for (int i = 2; i <= n; i++)
+        res *= i;
+    return (res);
+}
```
EmilyBourne, line 15: Do we need multiple versions for different precisions? Perhaps the maximum precision is sufficient. Is `long` the largest integer handled by pyccel?

pyccel/stdlib/pyc_math/pyc_math.c
```diff
@@ -0,0 +1,43 @@
...
+/*---------------------------------------------------------------------------*/
+long        pyc_gcd (long a, long b)
+{
+    while (b) {
+        a %= b;
+        /* swap a and b*/
+        a = a ^ b;
+        b = b ^ a;
+        a = a ^ b;
+    }
```
EmilyBourne, line 25: I wonder if this is the fastest implementation of this function. How did you choose the implementation? Do you have a reference?

pyccel/stdlib/pyc_math/pyc_math.c
```diff
@@ -0,0 +1,43 @@
...
+}
+/*---------------------------------------------------------------------------*/
+long        pyc_lcm (long a, long b)
+{
+    return a / pyc_gcd(a, b) * b;
+}
+/*---------------------------------------------------------------------------*/
+double      pyc_radians(double degrees)
+{
+    return (degrees * (M_PI / 180));
```
EmilyBourne, line 36: I think we need 180.0 here. It should be promoted, but better to be on the safe side

pyccel/stdlib/pyc_math/pyc_math.h
```diff
@@ -0,0 +1,10 @@
+#ifndef     PYC_MATH_H
+#define     PYC_MATH_H
+
+long        pyc_factorial(long n);
+long        pyc_gcd (long a, long b);
+long        pyc_lcm (long a, long b);
+double      pyc_radians(double degrees);
+double      pyc_degrees(double radians);
```
EmilyBourne, line 8: Is double precision sufficient?

tests/epyccel/test_epyccel_pyc_math.py
```diff
@@ -0,0 +1,132 @@
...
+from conftest       import *
+
+from pyccel.epyccel import epyccel
+from pyccel.decorators import types
+
+# -----------------------------------------------------------------------------
+@pytest.mark.parametrize( 'language', (
+        pytest.param("c", marks = pytest.mark.c),
+        pytest.param("fortran", marks = [
+            pytest.mark.xfail(reason="gcd not implemented in fortrand"),
```
EmilyBourne, line 15: Typo `fortran` -> `fortrand`

tests/epyccel/test_epyccel_pyc_math.py
```diff
@@ -0,0 +1,132 @@
+# pylint: disable=missing-function-docstring, missing-module-docstring/
+# pylint: disable=wildcard-import
+import sys
+from numpy.random import randint, uniform
+from numpy import isclose
+from conftest       import *
```
EmilyBourne, line 6: This import shouldn't be needed

tests/epyccel/test_epyccel_pyc_math.py
```diff
@@ -0,0 +1,132 @@
...
+    def call_degrees_i(x):
+        from math import degrees
+        return degrees(x)
+
+    f = epyccel(call_degrees_i, language=language)
+    x = randint(1e6)
+
+    assert isclose(f(x), call_degrees_i(x), rtol=1e-14, atol=1e-14)
+    assert isclose(f(-x), call_degrees_i(-x), rtol=1e-14, atol=1e-14)
+# -----------------------------------------------------------------------------
```
EmilyBourne, line 132: Why did you decide to create a new test file rather than putting these tests which test parts of the `math` library with all the other tests for the math library?

tests/epyccel/test_epyccel_pyc_math.py
```diff
@@ -0,0 +1,132 @@
...
+        from math import factorial
+        return factorial(x)
+
+    f = epyccel(call_factorial, language=language)
+    x = randint(15)
+
+    assert(f(x) == call_factorial(x))
+# -----------------------------------------------------------------------------
+# New in version 3.9.
+@pytest.mark.skipif(sys.version_info < (3, 9), reason="requires python3.9 or higher")
```
EmilyBourne, line 52: Should we add github action testing for python 3.9?

tests/epyccel/test_epyccel_pyc_math.py
```diff
@@ -0,0 +1,132 @@
...
+        pytest.param("c", marks = pytest.mark.c),
+        pytest.param("fortran", marks = [
+            pytest.mark.xfail(reason="degrees not implemented in fortran"),
+            pytest.mark.fortran]
+        )
+    )
+)
+def test_call_degrees_i(language):
+    @types('real')
+    def call_degrees_i(x):
```
EmilyBourne, line 123: Is this supposed to be `types('int')`?

EmilyBourne (changes requested):

EmilyBourne:
It looks like pip isn't collecting the `pyccel/stdlib/ndarrays` folder when it installs. Maybe this is because it doesn't contain any `.py` files? I'm not sure. The fix almost certainly goes in `setup.py`

pyccel/codegen/pipeline.py
```diff
@@ -228,6 +270,19 @@ def get_module_dependencies(parser, mods=(), folders=()):
 
         dep_mods, inc_dirs = get_module_dependencies(parser)
 
+        if language == 'c':
+            # check for used internal libraries
+            for lib in internal_libs:
+                if lib in codegen._printer._additional_imports:
+                    objs, incs = get_internal_lib(lib)
```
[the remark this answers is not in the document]
rakati: yes, but we need to copy lib to the working directory.
I will try to reconstruct the code.

pyccel/codegen/pipeline.py
```diff
@@ -174,6 +179,46 @@ def handle_error(stage):
         parsers = [parser]
         module_names = [module_name]
 
+    # -------------------------------------------------------------------------
+    internal_lib_dict = {}
+
+    def get_internal_lib(import_lib):
```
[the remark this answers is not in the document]
rakati: to be easy in reading.
you may sometimes try to understand the behavior of the code in the pipeline, and you don't using the internal library so you don't need to read the whole code for internal libs.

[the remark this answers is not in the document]
EmilyBourne: That's what comments and docstrings are for. I find this harder to read if I'm just trying to check what happens to a specific (non-internal) file

pyccel/codegen/pipeline.py
```diff
@@ -174,6 +179,46 @@ def handle_error(stage):
...
+        # remove library folder from pyccel if exist then copy new one
+        lib_dest_path = os.path.join(pyccel_dirpath, import_lib)
+        if os.path.exists(lib_dest_path):
+            shutil.rmtree(lib_dest_path)
+        shutil.copytree(lib_path, lib_dest_path)
+        # get all c files in library directory
+        c_files = []
+        for e in os.listdir(lib_dest_path):
+            if e.endswith(".c"):
+                c_files.append(os.path.join(lib_dest_path, e))
```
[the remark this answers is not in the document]
rakati: yes, just more works.
do you need me to do it also for fortran?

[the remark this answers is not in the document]
EmilyBourne: How much more work is it? It looks like just an `or` to me

pyccel/codegen/printing/ccode.py
```diff
@@ -280,7 +280,7 @@ def _print_LiteralComplex(self, expr):
     def _print_PythonComplex(self, expr):
         self._additional_imports.add("complex")
         return self._print(PyccelAssociativeParenthesis(PyccelAdd(expr.real_part,
-                        PyccelMul(expr.imag_part, LiteralImaginaryUnit()))))
+                        PyccelMul(PythonFloat(expr.imag_part), LiteralImaginaryUnit()))))
```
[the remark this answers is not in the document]
rakati: we don't need to add cast the real part to get the correct result, just the img part is enough.
maybe the problem will be also in Fortran lang, we just need to find the case.
I found this case in c when using numpy full, in the ndarray.c we are using there a Generic preprocessor to recognize the type of the literal number and here the problem happens.

pyccel/stdlib/ndarrays/ndarrays.h
rakati: they are already in the `ndarray.c` file.

pyccel/codegen/printing/ccode.py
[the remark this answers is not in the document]
EmilyBourne: I suspect that this is because a `LiteralComplex` with a real part doesn't actually exist.
```
1+2j
```
Could be a `LiteralComplex`, but in fact it is read as:
```
PyccelAdd(LiteralInt(1),PyccelComplex(0,2))
```
One day we should remove the real part, but in the mean time in case it ever gets used, I think we should treat the real and imaginary parts in the same way

> maybe the problem will be also in Fortran lang, we just need to find the case.

If the problem can occur in fortran then it shouldn't be fixed in the printer

pyccel/stdlib/pyc_math/pyc_math.c
rakati: the equivalent of the python integer in c is long. why we need other precision?
maybe we will need that for numpy factorial.

EmilyBourne: Ah you are right, we can't pass numpy types to the math library :+1:

pyccel/stdlib/pyc_math/pyc_math.c
rakati: hh, I use just a straight-forward implementation for testing purpose, I forget to mention that in the desc.
we may need to add vectorization and other optimizations to get efficient code.

pyccel/stdlib/pyc_math/pyc_math.h
EmilyBourne: See above. We can't pass numpy types to math functions

pyccel/stdlib/pyc_math/pyc_math.c
rakati: M_PI is a float number, the compiler knows how to handle that.
but if this can make you happy I will add it in the fly.

pyccel/stdlib/pyc_math/pyc_math.c
EmilyBourne: I'm not sure we need (or can) add vectorization here. But there might be a more efficient algorithm. We should probably look into this before merging otherwise we will forget to go back and do it

pyccel/stdlib/pyc_math/pyc_math.h
rakati: same as a factorial case.

tests/epyccel/test_epyccel_pyc_math.py
rakati: which one?

EmilyBourne: The conftest one. Pytest looks for it automatically

tests/epyccel/test_epyccel_pyc_math.py
rakati: I know it's the same topic (math), but here we are dealing with our implementation, the kind of errors we may get are different.
I think moving the file to recognized_function dir will be enough.
what do you think?

tests/epyccel/test_epyccel_pyc_math.py
rakati: it's up to you guys. in my opinion, we need to add the features in the new python versions.

tests/epyccel/test_epyccel_pyc_math.py
rakati: yes, I will change it.

tests/epyccel/test_epyccel_pyc_math.py
EmilyBourne: Sounds like a good compromise :+1:

tests/epyccel/test_epyccel_pyc_math.py
rakati: ok, I'll remove it.

pyccel/codegen/pipeline.py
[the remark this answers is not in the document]
rakati: there are some other changes to make but not a lot. if all things are going well maybe I will finish today otherwise It may take more time.
but I think it will be a great addition to pyccel. currently, @MenoIy  still working on the wrapper. so I will work on that until Karim is done.
what do you think?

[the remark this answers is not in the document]
EmilyBourne: I just mean get the copy/paste mechanism to work, not necessarily an entire internal library. If you have time then I think that would be good. That way we are sure that the method is sufficiently general.

If you plan to work on the wrapper next then I definitely agree that it would be easier to start that after PR #548 has been merged

pyccel/codegen/printing/ccode.py
rakati: I think this check should not be in the printer because it's a common error for both languages c and Fortran.
and these errors will be raised in python checks before using pyccel.

pyccel/codegen/printing/ccode.py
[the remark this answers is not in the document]
EmilyBourne: For the real/imag part. It turns out a `LiteralComplex` with a real part does exist if you create it with a call to `complex`. Can you confirm that you don't have this problem with (or similar to fit the problem you found):
```
np.full(2,complex(5,6))
```

pyccel/codegen/pipeline.py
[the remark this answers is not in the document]
EmilyBourne: I still agree with this comment. If the function doesn't fit comfortably on my screen then it shouldn't be a function in a function

pyccel/codegen/pipeline.py
```diff
@@ -228,6 +287,26 @@ def get_module_dependencies(parser, mods=(), folders=()):
...
 
+        # Determine if the module or program requires internal libraries
+        # by iterating over internal_libs list that containes internal libs
+        # implemented inside pyccel and compare them by module includes
+        # returned by printer includes through get_additional_imports function
+        for lib in internal_libs:
+            if lib in codegen.get_printer_imports():
+                # get the include folder path and library files
+                if lib not in internal_lib_dict:
+                    internal_lib_dict[lib] = get_internal_lib(lib)
```
EmilyBourne, line 298: It looks like you have moved this to after `convert_only` to avoid compiling when the user asked you not to. However `convert_only` also means that you need to generate all necessary code. So I think the libraries should still be collected and copied before the `if convert_only: continue`. I would use `get_internal_lib` to get the internal libraries and then add them to a list of files to be compiled which can all use the same method for compiling

pyccel/codegen/printing/ccode.py
```diff
@@ -778,6 +782,33 @@ def _print_MathSqrt(self, expr):
...
+        # include of the pyc_math header
+        self._additional_imports.add('pyc_math')
+        return 'pyc_degrees({})'.format(self._print(expr.args[0]))
+
+    def _print_MathRadians(self, expr):
+        """ Print the equivalent code to pycthon radians function call"""
+        # include of the pyc_math header
+        self._additional_imports.add('pyc_math')
+        return 'pyc_radians({})'.format(self._print(expr.args[0]))
+    # -------------- end of pyc_math functions --------------------------------
```
EmilyBourne, line 811: I don't suppose all these classes with the same methods happen to inherit from a base class that would let us write one print method with a dictionary do they?

pyccel/stdlib/math/pyc_math.f90
```diff
@@ -0,0 +1,19 @@
+module math
+    implicit none
+
+        real, parameter,private :: pi = 3.1415926536
+        real, parameter, private :: e = 2.7182818285
```
EmilyBourne, line 5: I think we should save these in double precision. Most of the generated code will use double precision

pyccel/stdlib/math/pyc_math.h
```diff
@@ -0,0 +1,10 @@
+#ifndef     PYC_MATH_H
+#define     PYC_MATH_H
+
+long        pyc_factorial(long n);
+long        pyc_gcd (long a, long b);
+long        pyc_lcm (long a, long b);
+double      pyc_radians(double degrees);
+double      pyc_degrees(double radians);
```
EmilyBourne, line 8: FWI: If this gets merged after PR #548 then all the integer datatypes will need updating

pyccel/codegen/printing/ccode.py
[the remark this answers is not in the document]
rakati: for Fortran, it's already handled.
https://github.com/pyccel/pyccel/blob/master/pyccel/ast/builtins.py#L106L111

tests/epyccel/recognised_functions/test_epyccel_pyc_math.py
```diff
@@ -0,0 +1,132 @@
...
+    y = randint(1e9)
+
+    assert(f(x,y) == call_gcd(x, y))
+# -----------------------------------------------------------------------------
+@pytest.mark.parametrize( 'language', (
+        pytest.param("c", marks = pytest.mark.c),
+        pytest.param("fortran", marks = [
+            pytest.mark.xfail(reason="factorial not implemented in fortran"),
+            pytest.mark.fortran]
+        )
```
EmilyBourne, line 37: I thought you had implemented a factorial for fortran

EmilyBourne (changes requested):

pyccel/codegen/printing/ccode.py
rakati: no, they aren't.
in order to write one print for all of them, we need to create these classes:
```python
class PycMathFunction(MathFunctionBase): pass

class PycMathFunctionFloat(PycMathFunction):
    _dtype = NativeReal()
    _precision = default_precision['real']

class PycMathFunctionInt(PycMathFunction):
    _dtype = NativeInteger()
    _precision = default_precision['integer']

class PycMathFunctionBool(PycMathFunction):
    _dtype = NativeBool()
```
then inherit all classes for the functions that are in pyc_math lib. this seems to be useless.

pyccel/stdlib/math/pyc_math.f90
```diff
@@ -4,7 +4,119 @@ module pyc_math
 
         real(kind=8), parameter, private :: pi = 4.D0 * DATAN(1.D0)
 
+    interface pyc_gcd
+
```
yguclu, line 8: A note about Fortran formatting: please remove this empty line.

pyccel/stdlib/math/pyc_math.f90
```diff
@@ -4,7 +4,119 @@ module pyc_math
...
+        module procedure pyc_gcd_4
+        module procedure pyc_gcd_8
+
```
yguclu, line 11: A note about Fortran formatting: please remove this empty line.

pyccel/stdlib/math/pyc_math.f90
```diff
@@ -4,7 +4,119 @@ module pyc_math
...
+    end interface pyc_factorial
+
+    interface pyc_lcm
+
+        module procedure pyc_lcm_4
+        module procedure pyc_lcm_8
+
+    end interface pyc_lcm
     contains
```
yguclu, line 27: Please add an empty line before the `contains` directive.

.gitignore
```diff
@@ -26,7 +26,6 @@ doc/_build
 doc/source
 
 *.pyccel
-*.f90
```
yguclu, line 29: Why should we ignore all Fortran files? If they are in `__pyccel__` they are ignored already.

EmilyBourne: Presumably this line predated the `__pyccel__` folder? In any case I agree that it can be removed

pyccel/codegen/pipeline.py
```diff
@@ -15,6 +15,18 @@
 
 __all__ = ['execute_pyccel']
 
+# map internal libraries with it's folder name inside pyccel/stdlib
```
yguclu, line 18: Please change this sentence to:
```python
# map internal libraries to their folders inside pyccel/stdlib
```

pyccel/codegen/pipeline.py
```diff
@@ -15,6 +15,18 @@
...
+internal_libs = {
+    "ndarrays" : "ndarrays",
+    "pyc_math" : "math",
+}
+
+# map language to it's file extension
```
yguclu, line 24: Typo: it's -> its

pyccel/codegen/pipeline.py
```diff
@@ -196,12 +212,67 @@ def handle_error(stage):
...
+        # if recursive:
+        #     for dep in parser.sons:
+        #         # Call same function on 'dep'
+        #         pass
         #------------------------------------------------------
 
+        # Determine if the module or program requires internal libraries
+        # by iterating over internal_libs list that containes internal libs
+        # implemented inside pyccel and compare them by module includes
+        # returned by printer includes through get_additional_imports function
```
yguclu, line 224: I got lost while reading this sentence, which is 4 lines long and does not contain any commas... Could you rephrase it please?

Typo: containes -> contains
Language: compare them by -> compare them with

pyccel/codegen/printing/ccode.py
EmilyBourne: These classes already exist on my computer. But I am not clear on why `PycMathFunctionFloat` would be necessary, i.e. why you need the dtype/precision to print.

To be clear I am not suggesting the removal of the classes `MathRadians` etc, just that the printing is done from `_print_MathFunctionFloat` which would be called if we didn't have a `_print_MathRadians` function

pyccel/codegen/printing/ccode.py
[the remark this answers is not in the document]
EmilyBourne: This problem should be fixed by PR #572

pyccel/codegen/printing/ccode.py
yguclu: @rakati I don't think that we need the three intermediate classes that you wrote above. The `PycMathFunction` base class seems enough to me, and your new classes can inherit from it directly.

In the C printer you will then have just one function, called `_print_PycMathFunction`, which will print the correct function name in C using a dictionary as @EmilyBourne suggested. The body of this function will look very similar to your current `_print_MathGcd` function.

Same thing for Fortran.

pyccel/codegen/printing/fcode.py
```diff
@@ -651,6 +657,41 @@ def _print_MathFloor(self, expr):
         prec_code = self._print(prec)
         return 'floor({}, kind={})'.format(arg_code, prec_code)
 
+    # --------------------------- pyc_math functions --------------------------
+
+    def _print_MathFactorial(self, expr):
+        """ Print the equivalent code to pycthon factorial function call"""
```
yguclu, line 663: Typo: pycthon -> python

This is repeated multiple times in the file.

yguclu (commented):
Was file pyccel/stdlib/ndarrays/ndarrays.h modified at all?
According to GitHub's web diff the whole file was changed, but I don't understand why.

pyccel/codegen/pipeline.py
rakati: I think the comment is too long.
what do you think about writing this comment instead :
```
# Iterate over the internal_libs list and determine if the printer
# requires an internal lib to be included.
```

pyccel/codegen/printing/fcode.py
rakati: fixed in both fcode and ccode

pyccel/codegen/printing/ccode.py
rakati: Ok, I will do that.

rakati:
> Was file pyccel/stdlib/ndarrays/ndarrays.h modified at all?
> According to GitHub's web diff the whole file was changed, but I don't understand why.

I fixed some bugs in this file before. but @EmilyBourne did changes the whole file to use `stdint `type, and she fixed the problems also in his PR, then I merged those changes to my branch.
I will do a checkout to this file.

tests/epyccel/test_epyccel_pyc_math.py
yguclu: @EmilyBourne Yes, I think it is time to remove version 3.5 and add version 3.9

yguclu (changes requested):
You must resolve the conflicts with master before this PR can be merged.

tests/epyccel/test_epyccel_pyc_math.py
EmilyBourne: I have created PR #585 to fix this. Once this is merged these functions will be correctly tested

pyccel/codegen/pipeline.py
```diff
@@ -196,12 +212,65 @@ def handle_error(stage):
...
+        # Iterate over the internal_libs list and determine if the printer
+        # requires an internal lib to be included.
+        for lib in internal_libs:
+            if lib in codegen.get_printer_imports():
+                # get the include folder path and library files
+                if lib not in internal_libs_name:
+                    # get the library folder name
+                    lib_name = internal_libs[lib]
+                    # get path to pyccel/stdlib/lib_name
+                    import pyccel.stdlib as stdlib_folder
```
EmilyBourne, line 230: Is there any reason the import is not done at the top of the file?

pyccel/codegen/pipeline.py
```diff
@@ -196,12 +212,65 @@ def handle_error(stage):
...
+                    stdlib_path = os.path.dirname(stdlib_folder.__file__)
```
EmilyBourne, line 231: Does this return the same value on every iteration of the for loop?

pyccel/codegen/pipeline.py
```diff
@@ -196,12 +212,65 @@ def handle_error(stage):
...
+                    source_files = []
+                    for e in os.listdir(lib_dest_path):
+                        if e.endswith(lang_ext_dict[language]):
+                            source_files.append(os.path.join(lib_dest_path, e))
+
+                    # compile library source files
+                    flags = construct_flags(f90exec,
+                                            fflags=fflags,
+                                            debug=debug,
+                                            includes=[lib_dest_path])
```
EmilyBourne, line 255: Do we really need to recalculate the flags after each loop?

pyccel/codegen/pipeline.py
```diff
@@ -196,12 +212,65 @@ def handle_error(stage):
...
+                                            binary=None,
+                                            verbose=verbose,
+                                            is_module=True,
+                                            output=lib_dest_path,
+                                            language=language)
+                    except Exception:
+                        handle_error('C {} library compilation'.format(lib))
+                        raise
+                    # Add internal lib to internal_libs_name dict
+                    internal_libs_name[lib] = 1
```
EmilyBourne, line 268: Is the `1` doing anything here? If not, why not use a set? Or even a list?

pyccel/codegen/pipeline.py
```diff
@@ -196,12 +212,65 @@ def handle_error(stage):
...
+                    # add source file without extension to internal_libs_files
+                    internal_libs_files.extend([os.path.splitext(f)[0] for f in source_files])
```
EmilyBourne, line 270: You don't need to convert the generator expression to a list before calling extend. You can just do:
```
internal_libs_files.extend(os.path.splitext(f)[0] for f in source_files)
```

See: https://www.geeksforgeeks.org/python-list-comprehensions-vs-generator-expressions/

pyccel/ast/mathext.py
```diff
@@ -13,6 +13,7 @@
     'math_functions',
     # ---
     'MathFunctionBase',
+    'PycMathFunctionBase',
```
EmilyBourne, line 16: Can you explain the difference between these two classes?

pyccel/stdlib/ndarrays/ndarrays.h
```diff
@@ -1,100 +1,100 @@
...
+                /* creating an array view */
+t_ndarray   array_slicing(t_ndarray p, ...);
+
+/* free */
+int32_t         free_array(t_ndarray dump);
+
+/* indexing */
+int32_t         get_index(t_ndarray arr, ...);
+
+#endif
```
EmilyBourne, line 100: Did this file change or is it a question of line endings?

tests/epyccel/recognised_functions/test_epyccel_pyc_math.py
```diff
@@ -0,0 +1,95 @@
+# pylint: disable=missing-function-docstring, missing-module-docstring/
+# pylint: disable=wildcard-import
```
EmilyBourne, line 2: Please do not disable wildcard import errors

EmilyBourne (changes requested):

EmilyBourne:
Does this PR fix #509?

pyccel/codegen/pipeline.py
rakati: the import is needed just to get the path of `stdlib`.

pyccel/codegen/pipeline.py
rakati: yes, the includes are different on each iteration.

pyccel/codegen/pipeline.py
rakati: now you convinced me.

pyccel/codegen/pipeline.py
rakati: ok, I will change it

pyccel/stdlib/ndarrays/ndarrays.h
rakati: no. I made a change, and you made a change too.
your PR is merged before mine, that why it looks like that.
I will do a checkout from the master.

tests/epyccel/recognised_functions/test_epyccel_pyc_math.py
rakati: ok

pyccel/ast/mathext.py
rakati: MathFunctionBase is the abstract for the math functions that are already implemented in C `math.h` library.
the other one is for `pyc_math `library functions.

EmilyBourne: Hmmm, I'm not sure this separation is necessary. Especially before the codegen stage as different languages have different math functions already implemented. If I have understood correctly the only difference between `MathFunctionBase` and `PycMathFunctionBase` is that you need to import `pyc_math.h` for `PycMathFunctionBase` and `math.h` for `MathFunctionBase`.
So `_print_MathFunctionBase` would be:
```
        self._additional_imports.add('pyc_math')
        type_name = type(expr).__name__
        try:
            func_name = pyc_math_dic[type_name]
        except KeyError:
            errors.report(PYCCEL_RESTRICTION_TODO, severity='fatal')
        args = []
        for arg in expr.args:
            if arg.dtype != expr.dtype:
                cast_func = python_builtin_datatypes[str_dtype(expr.dtype)]
                args.append(self._print(cast_func(arg)))
            else:
                args.append(self._print(arg))
        code_args = ', '.join(args)
        return '{0}({1})'.format(func_name, code_args)
```
and `_print_PycMathFunctionBase` would be:
```
        self._additional_imports.add('pyc_math')
        type_name = type(expr).__name__
        try:
            func_name = pyc_math_dic[type_name]
        except KeyError:
            errors.report(PYCCEL_RESTRICTION_TODO, severity='fatal')
        args = []
        for arg in expr.args:
            if arg.dtype != expr.dtype:
                cast_func = python_builtin_datatypes[str_dtype(expr.dtype)]
                args.append(self._print(cast_func(arg)))
            else:
                args.append(self._print(arg))
        code_args = ', '.join(args)
        return '{0}({1})'.format(func_name, code_args)
```
can't these be merged to:
```
        type_name = type(expr).__name__
        try:
            func_name, import_file = pyc_math_dic[type_name]
        except KeyError:
            errors.report(PYCCEL_RESTRICTION_TODO, severity='fatal')
        self._additional_imports.add(import_file)
        args = []
        for arg in expr.args:
            if arg.dtype != expr.dtype:
                cast_func = python_builtin_datatypes[str_dtype(expr.dtype)]
                args.append(self._print(cast_func(arg)))
            else:
                args.append(self._print(arg))
        code_args = ', '.join(args)
        return '{0}({1})'.format(func_name, code_args)
```
or similar?

rakati: this seems good. there is some differentiation, but I will try to make a general structure that can handle all these cases.

pyccel/stdlib/ndarrays/ndarrays.h
EmilyBourne: Yeah, I think I might have gotten rid of the windows line endings ^^'

pyccel/codegen/pipeline.py
EmilyBourne: This feels like overkill just for the includes. But I don't think it makes sense to fix this before fixing #517

pyccel/ast/mathext.py
EmilyBourne: That would be great. If it's just one or two classes that need handling differently (e.g MathFloor) then they can still have their own specialised method. But I'm pretty sure most of these can be handled really generally using `MathFunctionInt`, `MathFunctionFloat`, etc

EmilyBourne: This seems beyond the scope of this PR. If these specific classes can't be handled by the existing `_print_MathFunctionBase` then I would suggest giving them specialised implementations (i.e. `_print_MathLcm`) and leaving the math printing improvement for another PR. It would be good if you could open an issue describing the changes that must be made. I would rather use specialised implementations than create a `PycFunctionBase` class as this makes no sense to me outside of the codegen stage

pyccel/ast/mathext.py
```diff
@@ -82,9 +84,14 @@
 # Base classes
 #==============================================================================
 class MathFunctionBase(Function, PyccelAstNode):
+    """Abstract base class for the Math Functions"""
     _shape = ()
     _rank  = 0
 
+class PycMathFunctionBase(Function, PyccelAstNode):
+    """Abstract base class for the Math Functions implemented in Pyc_math
+    libray"""
```
EmilyBourne, line 93: I'm happy to approve this PR once this class has been removed

EmilyBourne (changes requested):

pyccel/ast/mathext.py
rakati: o n t h e w a y.

pyccel/codegen/printing/ccode.py
```diff
@@ -785,71 +776,89 @@ def _print_MathFunctionBase(self, expr):
...
+            code_arg = self._print(PythonFloat(arg))
+        else:
+            code_arg = self._print(arg)
+        return "isinf({})".format(code_arg)
 
-        ------
-        Example:
-        --------
-            math.gcd(x, y) ==> pyc_gcd(x, y)
+    def _print_MathIsnan(self, expr):
```
EmilyBourne, line 838: I'm sure some of these Math functions could be combined with a well chosen superclass (e.g. MathIsVal) but I think we can leave this for now

pyccel/codegen/printing/ccode.py
```diff
@@ -785,71 +776,89 @@ def _print_MathFunctionBase(self, expr):
...
-            if arg.dtype is NativeComplex():
-                self._additional_imports.add('complex')
-            if arg.dtype is not NativeReal():
-                args.append(self._print(NumpyFloat(arg)))
-            else :
+            if arg.dtype != expr.dtype:
+                cast_func = python_builtin_datatypes[str_dtype(expr.dtype)]
+                args.append(self._print(cast_func(arg)))
+            else:
                 args.append(self._print(arg))
```
EmilyBourne, line 798: Just checking. Does this mean that all math functions (except 1 or 2 exceptions such as MathFloor) have the same output type as their arguments? That seems like a good way of looking at it :)

EmilyBourne (approved):

pyccel/codegen/printing/ccode.py
```diff
@@ -998,7 +1060,8 @@ def _print_Assign(self, expr):
         if isinstance(rhs, (NumpyFull)):
             code_init = ''
             if rhs.fill_value is not None:
-                code_init = 'array_fill({0}, {1});'.format(self._print(rhs.fill_value), lhs)
+                dtype = self.find_in_dtype_registry(self._print(rhs.dtype), rhs.precision)
+                code_init = 'array_fill(({0}){1}, {2});'.format(dtype, self._print(rhs.fill_value), lhs)
```
saniQQ, line 1064: this lead to a double cast.
I think we better leave this without a fix until we find an appropriate one.

saniQQ (changes requested):

pyccel/codegen/printing/ccode.py
EmilyBourne: Can you give code showing the double cast? Does this lead to a double cast only if `rhs.fill_value` is already a cast? Or always?

EmilyBourne: Maybe the cast could be added only if `isinstance(rhs.fill_value, Literal)` ?

rakati: I talked about that yesterday.
the test of `numpy.full` will not pass if I left it without adding cast.
because we have `literalInteger `printer don't add casting like `pythonInt` printer.

pyccel/codegen/printing/ccode.py
rakati: yes, Most of them are returning the same type of inputs.
for the others, we create a specific printer for them.

saniQQ (approved):

pyccel/stdlib/math/pyc_math.f90
```diff
@@ -0,0 +1,149 @@
+module pyc_math
+
+implicit none
+
+real(kind=8), parameter, private :: pi = 4.D0 * DATAN(1.D0)
```
yguclu, line 5: I think we should use ISO_C_BINDING to define the kind type parameter, as we do in the Fortran printer now.

@saidctb @EmilyBourne Do you agree?

pyccel/stdlib/math/pyc_math.f90
```diff
@@ -0,0 +1,149 @@
+module pyc_math
```
yguclu, line 1: All the functions in this file should be declared as `pure`, and maybe even as `elemental`.

yguclu (changes requested):

pyccel/stdlib/math/pyc_math.f90
EmilyBourne: Yes I agreee

pyccel/stdlib/math/pyc_math.f90
rakati: yes, using the elemental functions will let us use this function for NumPy too.
but I remember that @saidctb said we will recreate this library in python and use it as a header in the future.

EmilyBourne: I agree with @rakati . Pure, yes. But `elemental` does not reflect the behaviour in python

yguclu: Adding `pure` is enough.

pyccel/stdlib/math/pyc_math.f90
saidctb: yes I agree

pyccel/stdlib/math/pyc_math.f90
```diff
@@ -0,0 +1,150 @@
...
+
+use ISO_C_BINDING
+
+implicit none
+
+real(C_DOUBLE), parameter, private :: pi = 4.D0 * DATAN(1.D0)
```
EmilyBourne, line 7: You should also use `ISO_C_BINDING` on literals to avoid potential casting. I.e:
```
4.0_C_DOUBLE
```

pyccel/stdlib/math/pyc_math.f90
```diff
@@ -98,9 +99,9 @@ function pyc_lcm_4(a, b) result(lcm)
 
     implicit none
 
-    integer(kind=4), value :: a
-    integer(kind=4), value :: b
-    integer(kind=4)        :: lcm
+    integer(C_INT32_T), intent(in) :: a
+    integer(C_INT32_T), intent(in) :: b
+    integer(C_INT32_T)        :: lcm
```
yguclu, line 104: I think that passing `a` and `b` by value was OK. Why did you change this?

rakati: I was trying to use pure, but then I realize that pure function can call only pure functions.
I removed pure keyword but I forget to restore args to be passed by value.

yguclu: But is this calling a non-pure function?

rakati: yes, lcm call gcd which is not a pure function, gcd can't be pure because it calls non pure functions mod and xor.

EmilyBourne: `mod` is pure, but `xor` apparently isn't. Pyccel translates python's `^` operator to the function `IEOR`. This seems to be pure.
GCC's doc on XOR does not recommend using it https://gcc.gnu.org/onlinedocs/gfortran/XOR.html

EmilyBourne: Passing by value doesn't seem to work with all compilers in pure functions. I get the following error:
```
Variable ‘b’ can not appear in a variable definition context (assignment) at (1) in PURE procedure
```
with gfortran 7.5

rakati: Yes, pure functions need to have arguments as intent(in).
I will replace xor with ieor and make the two functions gcd and lcm pure.

yguclu: I am pretty sure that pure functions accept both `value` and `intent(in)` arguments. Please check the standard...

EmilyBourne (approved):

pyccel/stdlib/math/pyc_math.f90
EmilyBourne: I found this link for the 2003 standard:
https://j3-fortran.org/doc/year/04/04-007.pdf
Section 12.6 talks about pure functions.
> C1266 The specification-part of a pure function subprogram shall specify that all its nonpointer dummy
> data objects have INTENT(IN).

> C1267 The specification-part of a pure subroutine subprogram shall specify the intents of all its non
> pointer dummy data objects.

In theory this means that you could use `value` if you do:
```
value, intent(in)
```
I can't find anything in the standard saying this is not allowed, however a simple test with gfortran 7.5 shows that this doesn't work (at least with this compiler). I get:
```
Dummy argument ‘a’ with INTENT(IN) in variable definition context (assignment) at (1)
```
As an argument cannot be both `intent(in)` and `value` this precludes the use of `value` for `pure` functions

yguclu (changes requested):
I repeat, the scalar input arguments should be passed by `value`. This is accepted by all modern compilers, and it is part of the Fortran standard.

@EmilyBourne I have modified `pyc_math.f90` to use `value` and compiled it with GFortran 7.5 without any issues.

pyccel/stdlib/math/pyc_math.f90
yguclu: Please check a newer version of the standard.

yguclu: https://j3-fortran.org/doc/year/10/10-007.pdf, Section 12.7:
> The specification-part of a pure function subprogram shall specify that all its nonpointer dummy data objects have the INTENT (IN) or the VALUE attribute.

The `value` attribute is part of the C-Fortran interoperability interface, so it might be quite recent.

EmilyBourne: Having investigated and discussed this further, gfortran 9.3 gives the same results. Intel accepts this syntax from ifort 17.0 onwards.
In any case the error is raised at the assignment not at the declaration. This means that we can use `value` if we keep the variables `x` and `y`.

Inefficiencies caused by this approach will be easily optimised out by the compiler. But it has the advantage of ensuring that the interface matches the one that would be generated by pyccel (which encounters the same problems).

yguclu (approved):

PR #569, opened 2020-11-30
Use stdint.h for integer types in arrays

EmilyBourne:
As discussed in PR #548, we are now using stdint.h for integer types in C. This PR makes those changes for arrays in C.

cc: @saniQQ

### DIFF
--- a/tests/ndarrays/test_ndarrays.c
+++ b/tests/ndarrays/test_ndarrays.c
@@ -5,13 +5,13 @@
 #define getname(X) #X
 #define my_assert(X , Y, dscr) _Generic((X), double: assert_double,\
                             float: assert_float,\
-                            int32: assert_int,\
+                            int32_t: assert_int32,\
                             double complex : assert_cdouble,\
                             default: assert_ns)(X , Y, getname(X), getname(Y), dscr, __func__, __FILE__, __LINE__)
 
 void assert_double(double v1 , double v2,
         const char *v1_name, const char *v2_name,const char *dscr,
-        const char * func, const char *file, int32 line)
+        const char * func, const char *file, int32_t line)
 {
     if (v1 != v2)
     {
@@ -26,7 +26,7 @@ void assert_double(double v1 , double v2,
 
 void assert_float(float v1 , float v2,
         const char *v1_name, const char *v2_name,const char *dscr,
-        const char * func, const char *file, int32 line)
+        const char * func, const char *file, int32_t line)
 {
     if (v1 != v2)
     {
@@ -39,9 +39,9 @@ void assert_float(float v1 , float v2,
     printf("[DSCR] %s\n", dscr);
 }
 
-void assert_int(int32 v1 , int32 v2,
+void assert_int32(int32_t v1 , int32_t v2,
         const char *v1_name, const char *v2_name,const char *dscr,
-        const char * func, const char *file, int32 line)
+        const char * func, const char *file, int32_t line)
 {
     if (v1 != v2)
     {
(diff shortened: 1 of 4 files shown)


saniQQ (commented):
is it necessary to use precise types for index variables?

tests/ndarrays/test_ndarrays.c
```diff
@@ -1,459 +1,459 @@
...
+}
+
+int32_t test_indexing_int32(void)
+{
+    int32_t m_1[] = {2, 3, 5, 5, 6, 7, 10, 11,
+                12, 260, 6, 8, 8, 0, 45, 0,
+                1, 0, 0, 1, 200, 33, 5, 57,
+                62, 70, 103, 141, 122, 26, 36, 82,
+                8, 10, 4115, 22, 1, 11, 1, 19};
+    int32_t m_1_shape[] = {5, 8};
```
rakati, line 88: why you use a precise type shape?
I think you need t use only int, as mentioned in the header file.

tests/ndarrays/test_ndarrays.c
```diff
@@ -1,459 +1,459 @@
...
+    t_ndarray x;
+    int32_t index;
+    int32_t c_index;
```
rakati, line 91: same here

tests/ndarrays/test_ndarrays.c
```diff
@@ -1,459 +1,459 @@
...
+}
+
+int32_t test_indexing_double(void)
+{
+    double m_1[] = {2, 3, 5, 5, 6, 7, 10, 11,
+                    12, 260, 6.34, 8, 8.002, 0.056, 45, 0.1,
+                    1.02, 0.25, 0.00005, 1, 200, 33, 5, 57,
+                    62, 70, 103.009, 141, 122, 26.50, 36.334, 82,
+                    8.44002, 10.056, 4115, 22.1, 1.1102, 011.25, 1.01110005, 19};
+    int32_t m_1_shape[] = {5, 8};
```
rakati, line 117: same here

tests/ndarrays/test_ndarrays.c
```diff
@@ -1,459 +1,459 @@
...
+}
+
+int32_t test_indexing_cdouble(void)
+{
+    double complex m_1[] = {0.37 + 0.588*I,  0.92689451+0.57106791*I,
+                            0.93598206+0.30289964*I,  0.54404246+0.09516331*I,
+                            0.02827254+0.00432899*I,  0.06873651+0.24810741*I,
+                            0.94040543+0.43508215*I,  0.58532094+0.67890618*I,
+                            0.68742283+0.64951155*I,  0.15372315+0.89699101*I};
+    int32_t m_1_shape[] = {5, 2};
```
rakati, line 146: same here

rakati (commented):
> is it necessary to use precise types for index variables?

no, just simple int.
in the forloops also and shapes.

saniQQ (approved):

EmilyBourne:
@saniQQ @rakati All your questions are about precision so I will answer them together here. As mentioned in the PR description, after PR #548 we will be using precise types in all the code generated by pyccel. While technically shapes don't need to have precise types, if they don't there may be casts generated when generating the code that creates the shape. The same is also true for indexes. (I don't know on what, if any, system this might be the case as all systems I know of have int==int32_t, but there is no guarantee of this so it is better to be consistent)

tests/ndarrays/test_ndarrays.c
EmilyBourne: You are right, it seems the header still needs some fixing

rakati (approved):

yguclu (approved):

PR #571, opened 2020-11-30
Handle duplicated types in a template or a uniontype

### LINKED ISSUE #568: Duplicated types in a template
**Describe the bug**
Using a template with duplicated types, causes a generation of identical functions with the same signature.

**To Reproduce**
```python

@template('T', types=['int', 'int'])
@types('T')
def f(a):
    return a
```

```c
#include <test.h>
#include <stdlib.h>



long f_00(long a)
{

    long Out_0001;

    Out_0001 = a;
    return Out_0001;
}

long f_01(long a)
{

    long Out_0001;

    Out_0001 = a;
    return Out_0001;
}
```

**Expected behavior**
 Only one function should be generated.

**Language**
Fortran, C

rmahjoubi:
Close #568:
- Remove duplicated types in a template;
- Remove duplicated types in a uniontype;
- Raise error if function is used in template or uniontype;
- Add tests.


MenoIy:
it's working for no more duplicated  interface function created but you need to check some other tests are failing

EmilyBourne:
@MenoIy Can you check if the method used here is compatible with what you have done in PR #548. Specifically I am concerned about losing the difference between `np.float64` and `float` which require different checks and collection methods (numpy vs python). I don't know if you always check for both or if it depends on the contents of the header

MenoIy:
@EmilyBourne  Normally I always check for precision because I have a variable and I don't know if it's a python or numpy one, but when selecting the check function if the precision of this variable dtype is the same as the default precision  I check for both python and numpy type. and the value is collected by the corresponding function.  so 
```
function('real' , 'real')   is the same as function ('float64', 'float64') in mac and lunix 
```
I worried about int more than float one ...

EmilyBourne:
> @EmilyBourne Normally I always check for precision because I have a variable and I don't know if it's a python or numpy one, but when selecting the check function if the precision of this variable dtype is the same as the default precision I check for both python and numpy type. and the value is collected by the corresponding function. so
> 
> ```
> function('real' , 'real')   is the same as function ('float64', 'float64') in all sys
> ```

Ok. Then it sounds like this PR shouldn't break anything in the other. Perfect :+1:

EmilyBourne:
Is this PR ready for review?

pyccel/parser/syntax/headers.py
```diff
@@ -59,15 +61,20 @@ def expr(self):
...
         BasicStmt.__init__(self)
 
     @property
     def expr(self):
-        l = [i.expr for i in self.dtype]
-        return Template(self.name, l)
+        if any(isinstance(d_type, FuncType) for d_type in self.dtypes):
+            msg = 'Functions in a template are not supported yet'
+            errors.report(msg,
+                        severity='fatal')
```
EmilyBourne, line 73: Can you add a test in tests/errors to make sure that this is correctly raised please

pyccel/parser/syntax/headers.py
```diff
@@ -59,15 +61,20 @@ def expr(self):
...
 
+        dtypes = tuple(dict(d_type) for d_type in {tuple(t.expr.items())\
+            for t in self.dtypes})
```
EmilyBourne, line 76: I am not clear on what this does. You have:
```
tmp_dict = {tuple(t.expr.items())  for t in self.dtypes}
dtypes = tuple( dict(d_type) for d_type in tmp_dict )
```
So `tmp_dict` should be a dictionary combining the t.expr dictionaries. But then you iterate over it. Which means that you just use the keys. I am not sure how you create a dictionary from that. I am surprised this works!

Can you simplify this expression?

pyccel/parser/syntax/headers.py
```diff
@@ -167,22 +174,26 @@ def __init__(self, **kwargs):
...
-        else:
-            return l[0]
+            for d_type in dtypes:
+                d_type["is_const"] = True
+        if len(dtypes)==1:
+            return dtypes[0]
+        if any(isinstance(d_type, FuncType) for d_type in self.dtypes):
+            msg = 'Functions in a uniontype are not supported yet'
+            errors.report(msg,
+                        severity='fatal')
```
EmilyBourne, line 193: Can you add a test in tests/errors to make sure that this is correctly raised please

pyccel/parser/syntax/headers.py
```diff
@@ -167,22 +174,26 @@ def __init__(self, **kwargs):
...
+        dtypes = [dict(d_type) for d_type in {tuple(t.items())\
+            for t in dtypes}]
```
EmilyBourne, line 195: See above. I don't understand

tests/pyccel/scripts/generic_functions.py
```diff
@@ -79,6 +79,15 @@ def array_elem1(x):
 def multi_tmplt_2(y, z):
     return y + z
 
+@template('g', types=['int', 'int'])
```
EmilyBourne, line 82: Does this only handle explicitly duplicated types or also duplicated types with different names (e.g. `float64` == `real`)?

EmilyBourne (changes requested):

pyccel/parser/syntax/headers.py
rmahjoubi: `tmp_dict` in this case could be `{(('datatype', 'int'), ('precision', 4)), (('datatype', 'int'), ('precision', 8))}`, this is in order to remove any duplicated tuples that represents the types.
and then we convert the tuples back to dictionaries, `dict((('datatype', 'int'), ('precision', 4)))` would become `{'datatype': 'int', 'precision': 4}`. Do you think this is not a suitable approach ?

pyccel/parser/syntax/headers.py
rmahjoubi: Ok I will add some tests

tests/pyccel/scripts/generic_functions.py
rmahjoubi: it handles duplicated types with different names as well, I will add some tests for this case

pyccel/parser/syntax/headers.py
EmilyBourne: Ok. That makes sense. It sounds suitable I just don't think it's very readable. Would you mind breaking it into 2 lines and giving the first line a variable with a name that explains what is happening. E.g:
```
possible_dtype_prec = {tuple(t.expr.items())  for t in self.dtypes}
dtypes = tuple( dict(d_type) for d_type in possible_dtype_prec )
```

EmilyBourne (approved):

tests/pyccel/scripts/generic_functions.py
yguclu: Do we still support `real` in annotations?

yguclu (approved):

PR #572, opened 2020-12-01
Ensure literals have the right argument. Create literals with a requested precision

EmilyBourne:
Add type checks to Literal classes to ensure that they contain arguments of the correct type. Use the precision argument to translate precision modifiers of literals to a literal of the correct precision. E.g:
```
np.float32(2)
```
Is now translated to:
```
2.0_C_FLOAT
```
instead of:
```
a = Real(2_C_LONG_LONG, 4)
```
Casts are still printed for truncated values

**Commit summary**
- Use inheritance to set `Literal` precision and check its type
- Remove unused `sympy.Integer` inheritance from `LiteralInteger` (this has many knock-on effects as it makes it obvious when we are not using pyccel operators)
- Add type checks and conversions to `Literal` classes
- Use literal precision to translate expressions such as `np.float32(2)` directly to `2.0_C_FLOAT`
- Print floats with correct precision in fortran
- Fix `unexpected keyword argument : "precision"`
- ensure that `complex` function works when complex arguments are passed
- Simplify phrases such as `2+5j` to `LiteralComplex(2,5)`

### DIFF
--- a/pyccel/ast/literals.py
+++ b/pyccel/ast/literals.py
@@ -36,6 +36,10 @@ class Literal(PyccelAstNode):
         """ Set precision for a literal class"""
         self._precision = precision
 
+    @property
+    def python_value(self):
+        """ Get python literal represented by this instance """
+
 #------------------------------------------------------------------------------
 class LiteralTrue(sp_BooleanTrue, Literal):
     """Represents the python value True"""
--- a/pyccel/ast/literals.py
+++ b/pyccel/ast/literals.py
@@ -45,6 +45,10 @@ class LiteralTrue(sp_BooleanTrue, Literal):
     def __init__(self, precision = default_precision['bool']):
         Literal.__init__(self, precision)
 
+    @property
+    def python_value(self):
+        return True
+
 #------------------------------------------------------------------------------
 class LiteralFalse(sp_BooleanFalse, Literal):
     """Represents the python value False"""
@@ -53,6 +57,10 @@ class LiteralFalse(sp_BooleanFalse, Literal):
         return sp_BooleanFalse.__new__(cls)
     def __init__(self,precision = default_precision['bool']):
         Literal.__init__(self, precision)
+
+    @property
+    def python_value(self):
+        return False
 
 #------------------------------------------------------------------------------
 class LiteralInteger(Basic, Literal):
@@ -67,6 +75,10 @@ class LiteralInteger(Basic, Literal):
             raise TypeError("A LiteralInteger can only be created with an integer")
         self.p = value
 
+    @property
+    def python_value(self):
+        return self.p
+
 #------------------------------------------------------------------------------
 class LiteralFloat(sp_Float, Literal):
     """Represents a float literal in python"""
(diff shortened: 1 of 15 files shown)


pyccel/ast/literals.py
```diff
@@ -29,59 +27,114 @@ class Literal(PyccelAstNode):
...
+            if real.precision == precision:
+                self._real_part = real
+            else:
+                self._real_part = LiteralFloat(real.args[0], precision = precision)
+        elif isinstance(real, LiteralInteger):
+            self._real_part = LiteralFloat(real.p, precision = precision)
+        elif isinstance(real, (int, float)):
+            self._real_part = LiteralFloat(real, precision = precision)
+        else:
+            raise TypeError("The real part of a LiteralComplex must be an int/float/LiteralFloat")
```
yguclu, line 125: Is a `LiteralInteger` not allowed?

pyccel/ast/operators.py
```diff
@@ -447,6 +449,14 @@ class PyccelAdd(PyccelArithmeticOperator):
     """
     _precedence = 12
 
+    def __new__(cls, arg1, arg2):
+        if isinstance(arg1, (LiteralInteger, LiteralFloat)) and \
+           isinstance(arg2, LiteralComplex) and \
+           arg2.real == LiteralFloat(0):
+            return LiteralComplex(arg1, arg2.imag)
```
yguclu, line 456: As far as I understand this handles a case like `1 + 2j`, but why shouldn't it also handle `2j + 1`?

pyccel/ast/operators.py
```diff
@@ -488,6 +498,14 @@ class PyccelMinus(PyccelArithmeticOperator):
     """
     _precedence = 12
 
+    def __new__(cls, arg1, arg2):
+        if isinstance(arg1, LiteralFloat) and \
+           isinstance(arg2, LiteralComplex) and \
+           arg2.real_part == LiteralFloat(0):
+            return LiteralComplex(arg1, -float(arg2.imag_part))
```
yguclu, line 505: Above you used the properties `real` and `imag`, here you are using `real_part` and `imag_part` instead. What's the difference?

tests/epyccel/modules/complex_func.py
```diff
@@ -0,0 +1,84 @@
...
+    a = complex(2.8+7j,1)
+    return a
+
+def create_complex_literal__complex_float():
+    a = complex(2.8+7j,1.8)
+    return a
+
+def create_complex_literal__complex_complex():
+    a = complex(2.8+7j,1.5+22j)
+    return a
```
yguclu, line 39: Maybe throw in some negative numbers for good measure?

pyccel/ast/operators.py
EmilyBourne: `real_part` and `imag_part` are the properties of `PythonComplex` while `real` and `imag` are the properties of `LiteralComplex`. We should probably use the same for both. Which do you prefer?

yguclu: I prefer the shortest names: `real` and `imag`

yguclu (approved):

PR #573, opened 2020-12-01
Describe how to link Fortran libraries using header files

saidctb:
Link a Fortran static library using header files

### DIFF
--- a/tutorial/header-files.md
+++ b/tutorial/header-files.md
@@ -79,7 +79,7 @@ def print_fib(x):
     from header import fib
     print(fib(x))
 ```
-To compile this file we execute the follwing command `pyccel test_funcs.py --libs=funcs --libdir=$PWD`, this will create the shared library `test_funcs.so`
+To compile this file we execute the following command `pyccel test_funcs.py --libs=funcs --libdir=$PWD`, this will create the shared library `test_funcs.so`
 
 ## Pickling header files
 Parsing a large Pyccel header file with hundreds of function declarations may require a significant amount of time, therefore it is important that this process is only done once when pyccelizing multiple Python source files in a large project.
--- a/tutorial/header-files.md
+++ b/tutorial/header-files.md
@@ -17,13 +17,13 @@ We create the file `header.pyh` that contains an openmp function definition:
 
 #$ header function omp_get_num_threads() results(int)
 ```
-We then create `openmp.py` file:
+We then create `example.py` file:
 
 ```python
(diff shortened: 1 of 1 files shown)


tutorial/header-files.md
```diff
@@ -24,6 +25,61 @@ print('number of threads is :', omp_get_num_threads())
...
+
+```python
+from pyccel.decorators import types
+
+@types('int')
+def print_fib(x):
+    from header import fib
+    print(fib(x))
+```
+To compile this file we execute the follwing command `pyccel test_funcs.py --libs=funcs --libdir=$PWD`, this will create the shared librar`test_funcs.so`
```
EmilyBourne, line 82: Typo `librar` -> `library `

saidctb: done

yguclu: Typo: `follwing` -> `following`

tutorial/header-files.md
```diff
@@ -24,6 +25,61 @@ print('number of threads is :', omp_get_num_threads())
 ```
 Pyccel can compile the Python file with the following command: `pyccel openmp.py --language fortran --flags="-fopenmp"`
```
yguclu, line 26: I mentioned this before. Are we sure that we cannot simply use this command,
```bash
$ pyccel openmp.py --language fortran --openmp
```
which should work with different compilers on different platforms?

saidctb: yes we can use this command for sure, it will work

EmilyBourne, line 26: Do we need to use fortran for this to work? If not then can we remove `--language fortran`

yguclu: Actually Fortran is the default language, so we definitely do not need to specify the language.

saidctb: does fopenmp works for fortran only or does  it work with all compilers and os  and for Fortran and C ?
because this way the user can change however he wants

tutorial/header-files.md
saidctb: done

tutorial/header-files.md
yguclu: I think that we should provide the simplest possible example here. Moreover it is definitely confusing to call this example file `openmp.py`. Hence I recommend
```bash
$ pyccel example.py --openmp
```

saidctb: okay

yguclu (approved):

Issue #537, opened 2020-11-10
Handle functions that return Numpy arrays


yguclu:
Description
---------------
As a USER, I would like to accelerate a Python function that returns a newly created array.

This happens all the time in prototype codes, even though it might not be the most efficient approach for HPC computing.

Acceptance Criteria
--------------------------
- [x] Return contiguous Numpy arrays in C
- [x] Return contiguous Numpy arrays in Fortran
- [x] Add extensive unit tests
- [x] Check for memory leaks
- [x] Return multiple arrays from a function

EmilyBourne:
A first version of the implementation in fortran should look as follows:
- [ ] In pure fortran functions returning arrays should return an `allocatable`
- [ ] In the `bind_c_X.F90` file (see ast/cwrapper.py):
      - a `pointer` of the same size as the `allocatable` should be allocated
      - The data from the `allocatable` should be copied to the `pointer`
      - The address of the pointer should be saved in a `c_ptr` type using the `c_loc` function
- [ ] The `c_ptr` object (equivalent to `void*` in c) should be collected and translated to a numpy array in the `X_wrapper.c` file

EmilyBourne:
Blocked by #886

PR #574, opened 2020-12-01
Add simple Garbage collector to C printer / raise errors in some array assignment cases

saniQQ:
#### Printing:

1. Collect allocatable array variables in CodeBlock (repeat recursively if there is a nested CodeBlock);
2. When a `return` is found, deallocate all the array variables collected up to this point by printing one of the deallocating functions `free_allocs`/`free_pointer` for each of the arrays;
3. If a CodeBlock is a `Program` or a `FunctionDef` with no trailing `Return` node, print one of the free functions `free_allocs`/`free_pointer` with all the array variables collected up to this point;
4. Remove the last CodeBlock variables from the `_allocs` variables list.

#### Add `free_array` to ndarrays C lib:

This function is responsible for freeing an **array object**; it frees the array data buffer + shape + strides.

#### Add `free_pointer` to ndarrays C lib:

This function is responsible for freeing a **pointer** to an array object; it frees the shape + strides, but a pointer does not own the data buffer.

#### Handle pointers assignment:

- Add `alias_assign` to ndarrays C lib:
  - `alias_assign` handles the assign between 2 pointers, pointers do not own the array buffer but do own their shape and strides;
  - the ownership of the shape and strides is for handling cases where can produce dangling pointers (this is a product of how C pointers are different than python);
- Handle reassignment of an array pointer by adding a Deallocation Node in the ast before reassigning pointers (this prevents memory leak from dangling pointers).

#### Error management:

- Raise errors when Pyccel encounters unsupported usage:
  - reassign a target variable;
  - assign of allocatable.
- Add two tests to check error handling.


pyccel/codegen/printing/ccode.py
```diff
@@ -689,7 +691,7 @@ def _print_Allocate(self, expr):
         shape = ", ".join(a for a in shape)
         dtype = self._print(expr.variable.dtype)
         dtype = self.find_in_ndarray_type_registry(dtype, expr.variable.precision)
-        shape_Assign = "(int[]){" + shape + "}"
+        shape_Assign = "(int32_t[]){" + shape + "}"
```
EmilyBourne, line 694: It would be better if you collected `int` using `find_in_dtype_registry` that way it would have been automatically updated

pyccel/codegen/printing/ccode.py
```diff
@@ -972,6 +974,16 @@ def _print_Assign(self, expr):
             return '{};'.format(self._print(expr.rhs))
         lhs = self._print(expr.lhs)
         rhs = expr.rhs
+        if isinstance(rhs, IndexedElement) and isinstance(expr.lhs, Variable):
+            free_code = ''
+            for i in rhs.args:
+                if isinstance(i, Slice):
+                    free_code += 'free_array(%s);\n' % self._print(expr.lhs.name)
+                    self._additional_imports.add('ndarrays')
+                    break
```
EmilyBourne, line 983: This can be written more succinctly as:
```
if any(isinstance(i, Slice) for i in rhs.args):
    free_code += 'free_array(%s);\n' % self._print(expr.lhs.name)
    self._additional_imports.add('ndarrays')
```
One line loops are always much faster than multi-line loops in python and `break` statements tend to make code harder to understand

pyccel/codegen/printing/ccode.py
```diff
@@ -972,6 +974,16 @@ def _print_Assign(self, expr):
             return '{};'.format(self._print(expr.rhs))
         lhs = self._print(expr.lhs)
         rhs = expr.rhs
+        if isinstance(rhs, IndexedElement) and isinstance(expr.lhs, Variable):
+            free_code = ''
+            for i in rhs.args:
+                if isinstance(i, Slice):
+                    free_code += 'free_array(%s);\n' % self._print(expr.lhs.name)
```
EmilyBourne, line 981: Am I right in saying that this handles:
```
a = b[1,:]
```
by printing
```
free_array(a);
a = ...
```
?
Is the `free_array` always necessary? This looks like it is a similar case to the `Allocate` node where a `status` is required

pyccel/codegen/printing/ccode.py
```diff
@@ -1020,7 +1032,17 @@ def _print_For(self, expr):
...
+        self._allocs.append(allocs)
         for b in expr.body :
+            if isinstance(b, Allocate):
+                allocs.add(b.variable)
+            if isinstance(b, AliasAssign):
+                if isinstance(b.rhs, IndexedElement):
+                    for i in b.rhs.args:
+                        if isinstance(i, Slice):
+                            allocs.add(b.lhs)
+                            break
```
EmilyBourne, line 1045: Can you explain why this needs to be done in the code block rather than in `_print_Allocate` and `_print_AliasAssign`?

In addition this doesn't look like a c specific problem to me. If it was well implemented with a `Deallocate` node then this could also be used in fortran or other languages we may wish to add in the future

pyccel/codegen/printing/ccode.py
```diff
@@ -1183,25 +1205,37 @@ def _print_OMP_Single_Construct(self, expr):
     def _print_Omp_End_Clause(self, expr):
         return '}'
     #=====================================
+    def free_allocs(self, variables):
+        allocs = set()
+        for var in variables:
+            for v in var:
+                allocs.add(v)
```
EmilyBourne, line 1212: Where possible you should use list comprehension as it is much faster
```
allocs = set(v for var in variables for v in var)
```

pyccel/codegen/printing/ccode.py
```diff
@@ -1183,25 +1205,37 @@ def _print_OMP_Single_Construct(self, expr):
...
+        if len(allocs) > 0:
+            allocs_str = ', '.join(self._print(i.name) for i in allocs)
+            self._allocs = self._allocs[:len(self._allocs) - 2]
```
EmilyBourne, line 1215: Why do we remove the last 2 elements?

In python negative indexes are allowed and are translated to the line you have written. This line would normally be written as: `self._allocs = self._allocs[:-2]`

pyccel/codegen/printing/ccode.py
```diff
@@ -1183,25 +1205,37 @@ def _print_OMP_Single_Construct(self, expr):
...
+            return 'free_allocs(' + str(len(allocs)) + ', ' + allocs_str + ');\n'
```
EmilyBourne, line 1216: Am I right in saying that here we right something like:
```
free_allocs(3, a, b, c);
```
?
It feels more natural to me to do:
```
free_allocs(a);
free_allocs(b);
free_allocs(c);
```
Do you have any reason for preferring the former?

pyccel/errors/messages.py
```diff
@@ -123,6 +123,10 @@
 INVALID_FILE_EXTENSION = 'Wrong file extension. Expecting `py` of `pyh`, but found'
 INVALID_PYTHON_SYNTAX = 'Python syntax error'
 
+# Assignment errors
+REASSIGN_TARGET = 'can not reassign a target variable'
+ASSIGN_ALLOCATABLES = 'can not assign two allocatable variables'
```
EmilyBourne, line 128: I do not understand this message. I cannot do:
```
a = np.ones(3) # assign one allocatable variable
b = np.ones(3) # assign another allocatable variable (now I have assigned two allocatable variables)
```

EmilyBourne, line 128: If possible we  should avoid using words like `target` and `allocatable` in the error messages. These are phrases that we use internally which come from fortran, but there is no reason to expect that someone translating from python to c would understand what is meant by them and even less reason to expect that a basic python user (who does not usually look at the generated code) would understand them

pyccel/parser/semantic.py
```diff
@@ -1317,7 +1317,6 @@ def _ensure_target(self, rhs, d_lhs):
         if isinstance(rhs, IndexedElement) and rhs.rank > 0 and rhs.base.internal_variable.allocatable:
             d_lhs['allocatable'] = False
             d_lhs['is_pointer' ] = True
-
```
EmilyBourne, line 1320: Is there any reason for removing whitespace? Spacing out code can help make it more readable. I have seen several PRs recently where some people remove whitespace and others put it back

pyccel/parser/semantic.py
```diff
@@ -1534,7 +1533,6 @@ def _visit_Assign(self, expr, **settings):
 
         rhs = expr.rhs
         lhs = expr.lhs
-
```
EmilyBourne, line 1537: See above, why remove whitespace?

pyccel/parser/semantic.py
```diff
@@ -1869,7 +1867,15 @@ def _visit_Assign(self, expr, **settings):
 
         for l, r in zip(lhs,rhs):
             is_pointer_i = l.is_pointer if isinstance(l, (Variable, DottedVariable)) else is_pointer
-
+            if isinstance(l, Variable) and l.is_ndarray:
```
EmilyBourne, line 1870: `l` is always a `Variable` I think

pyccel/parser/semantic.py
```diff
@@ -1869,7 +1867,15 @@ def _visit_Assign(self, expr, **settings):
...
+                if l.is_target:
+                    errors.report(REASSIGN_TARGET,
+                          bounding_box=(self._current_fst_node.lineno, self._current_fst_node.col_offset),
+                          severity='error', symbol=l.name)
+                if isinstance(r, Variable) and r.allocatable and l.allocatable:
+                    errors.report(ASSIGN_ALLOCATABLES,
+                          bounding_box=(self._current_fst_node.lineno, self._current_fst_node.col_offset),
+                          severity='error', symbol=l.name)
             new_expr = Assign(l, r)
```
EmilyBourne, line 1879: Are you sure this is the right place for this code? I suspect any modification of the properties has already been done by this point. I expected to see this code in `_ensure_target` or `_assign_lhs_variable`

tests/epyccel/test_arrays_multiple_assignments.py
```diff
@@ -165,6 +166,69 @@ def f(c):
...
+    errors = Errors()
+
+    # epyccel should raise an Exception
+    with pytest.raises(PyccelSemanticError):
+        epyccel(f, language=language)
+
+    # Check that we got exactly 1 Pyccel error
+    assert errors.has_errors()
+    assert errors.has_warnings()
+    assert errors.num_messages() == 2
```
EmilyBourne, line 189: The assertions that are carried out here do not match what you say you are doing in the comment. To check that you have exactly 1 Pyccel error I would expect:
```
    assert errors.has_errors() ?== 1
    assert errors.has_warnings() == 0
    assert errors.num_messages() == 1
```

tests/epyccel/test_arrays_multiple_assignments.py
```diff
@@ -165,6 +166,69 @@ def f(c):
...
+    errors = Errors()
+
+    # epyccel should raise an Exception
+    with pytest.raises(PyccelSemanticError):
+        epyccel(f, language=language)
+
+    # Check that we got exactly 1 Pyccel error and  warning
+    assert errors.has_errors()
+    assert errors.has_warnings()
+    assert errors.num_messages() == 2
```
EmilyBourne, line 221: See above. Comment does not match code

tests/epyccel/test_arrays_multiple_assignments.py
```diff
@@ -165,6 +166,69 @@ def f(c):
...
+
+    # Check that we the warning is correct
```
EmilyBourne, line 223: Forgotten `we` in sentence

EmilyBourne (changes requested):
There is some code duplication around the use of `free_allocs` which looks hard to track down if we changed the definition of `free_allocs` later when we've forgotten how it was implemented. It looks to me like it would make sense to create a `Deallocate` node to help deal with this

pyccel/codegen/printing/ccode.py
saniQQ: yes, this handles assigning to an already allocated variable (`status` can be unknown) but I don't have access to the Allocate node in this case i don't think it's necessary to add the if condition all around the code we can add it implicitly to`free_array`

the `if` condition is just to optimize so we don't need to call all the free inside the function

pyccel/parser/semantic.py
saniQQ: i think it can be a DottedVariable from what I saw from the previous line

pyccel/parser/semantic.py
saniQQ: from what i tested i have all the properties that is needed by this point. 
but I will try to move them to what you suggested I makes more sense to put them there if I have all the properties.

pyccel/codegen/printing/ccode.py
saniQQ: i just wanted the code to stay minimal that's all

pyccel/errors/messages.py
saniQQ: i meant to say: assigning between to alloctables variables
is it the right way to express it?

saniQQ: i will try to make the error msgs more clear

tests/epyccel/test_arrays_multiple_assignments.py
saniQQ: fixed

tests/epyccel/test_arrays_multiple_assignments.py
saniQQ: fixed

tests/epyccel/test_arrays_multiple_assignments.py
saniQQ: fixed

pyccel/codegen/printing/ccode.py
saniQQ: i put it there so i can keep track of which CodeBlock i am working on.

pyccel/parser/semantic.py
saniQQ: i didn't notice that i removed it mb.

pyccel/parser/semantic.py
saniQQ: i didn't notice that i removed it mb.

pyccel/errors/messages.py
saniQQ: can I merge the two errors to :
"array reallocation will cause unaccessible memory"

pyccel/codegen/printing/ccode.py
```diff
@@ -227,14 +227,15 @@ class CCodePrinter(CodePrinter):
         'dereference': set()
     }
 
-    def __init__(self, parser, settings=None):
+    def __init__(self, parser, target_language, settings=None):
```
EmilyBourne, line 230: Why do you need the target language? If we are using `CCodePrinter` then the `target_language` is C.

The only place where this may not be true is in the `CWrapperCodePrinter` but if something needs to be done differently in the `CWrapperCodePrinter` then this should be done by overriding the function from `CCodePrinter` not by modifying the `CCodePrinter` class. Think of the `CWrapperCodePrinter` as a special case of a `CCodePrinter`. And the `CCodePrinter` as a printer for writing pure c (nothing that uses python functions or targets other languages)

pyccel/codegen/printing/ccode.py
```diff
@@ -813,7 +816,11 @@ def _print_FunctionDef(self, expr):
 
         if len(expr.results) > 1:
             self._additional_args.append(expr.results)
+        if self._target_language == 'c':
+            self._allocs = [i for i in expr.local_vars if i.allocatable and i.is_ndarray]
```
EmilyBourne, line 820: Can you explain why `if self._target_language == 'c':` is needed here?

pyccel/codegen/printing/ccode.py
```diff
@@ -1186,25 +1203,34 @@ def _print_OMP_Single_Construct(self, expr):
     def _print_Omp_End_Clause(self, expr):
         return '}'
     #=====================================
+    def free_allocs(self, variables):
+
+        free_str = ''
+        for i in variables:
+            free_str += '\nfree_array(%s);' % i.name
+        return free_str
```
EmilyBourne, line 1211: This can be written more simply as:
```
free_str = '\n'.join('free_array({});'.format(i.name) for i in variables)
```
This has 2 benefits:
1. The expression no longer starts with an unnecessary line break
2. Generator expressions are faster than for loops

pyccel/errors/messages.py
```diff
@@ -123,6 +123,10 @@
 INVALID_FILE_EXTENSION = 'Wrong file extension. Expecting `py` of `pyh`, but found'
 INVALID_PYTHON_SYNTAX = 'Python syntax error'
 
+# ARRAY ERRORS
+ASSIGN_ARRAYS_ONE_ANOTHER = 'assigning arrays one another NOT IMPLEMENTED'
+ARRAY_ALREADY_IN_USE = 'attempt to reallocate an array wich is being used by another variable'
```
EmilyBourne, line 128: This message is a lot clearer thanks :+1:
Please capitalize the first word of the sentence
Typo: `wich` -> `which`

pyccel/errors/messages.py
```diff
@@ -123,6 +123,10 @@
 INVALID_FILE_EXTENSION = 'Wrong file extension. Expecting `py` of `pyh`, but found'
 INVALID_PYTHON_SYNTAX = 'Python syntax error'
 
+# ARRAY ERRORS
+ASSIGN_ARRAYS_ONE_ANOTHER = 'assigning arrays one another NOT IMPLEMENTED'
```
EmilyBourne, line 127: There seems to be a word missing here. Can I suggest:
```
Arrays which own their data cannot become views on other arrays
```
Does this seem clear to you? Does it correctly describe the situation?

pyccel/parser/semantic.py
EmilyBourne: It looks like it's in the right place now :+1:

pyccel/errors/messages.py
saniQQ: yeah I think it describe it more clearly

pyccel/codegen/printing/ccode.py
saniQQ: because the cwrapper is writing the frees of the garbage collector I think I can remove this with what u requested of just overriding the free_allocs method

pyccel/codegen/printing/ccode.py
saniQQ: yeah this is better will change it

pyccel/codegen/printing/ccode.py
```diff
@@ -899,6 +905,9 @@ def _print_Return(self, expr):
         args = [VariableAddress(a) if self.stored_in_c_pointer(a) else a for a in expr.expr]
         if expr.stmt:
             code += self._print(expr.stmt)+'\n'
+        if len(self._allocs) > 0:
+            code += self.free_allocs(self._allocs)
+            code += '\n'
```
EmilyBourne, line 910: This still feels like it belongs in the `SemanticParser` to me. Could identical code be written in the semantic step if a `Deallocate` class was created?
Then `_print_Deallocate` for fortran would just be:
```
return ''
```
?

EmilyBourne (changes requested):

pyccel/ast/core.py
```diff
@@ -877,6 +877,52 @@ def __eq__(self, other):
...
+
+    # ...
+
+    @property
+    def variable(self):
+        return self._variable
+
+    def _sympystr(self, printer):
+        sstr = printer.doprint
+        return 'Deallocate({})'.format(sstr(self.variable))
```
EmilyBourne, line 918: Do we need `_sympystr`? `__str__` is useful for debugging but this seems specific to sympy

pyccel/codegen/printing/ccode.py
```diff
@@ -698,10 +698,14 @@ def _print_Allocate(self, expr):
...
         dtype = self._print(expr.variable.dtype)
         dtype = self.find_in_ndarray_type_registry(dtype, expr.variable.precision)
-        shape_Assign = "(int[]){" + shape + "}"
+        shape_dtype = self.find_in_dtype_registry('int', 4)
+        shape_Assign = "("+ shape_dtype +"[]){" + shape + "}"
         alloc_code = "{} = array_create({}, {}, {});".format(expr.variable, len(expr.shape), shape_Assign, dtype)
         return '{}\n{}'.format(free_code, alloc_code)
 
+    def _print_Deallocate(self, expr):
+        return 'free_array({})'.format(self._print(expr.variable.name))
```
EmilyBourne, line 707: Why do you print `expr.variable.name` and not `expr.variable`? This looks like it is done to manage objects which are `stored_in_c_pointer`, but this shouldn't be the case for arrays should it?

pyccel/codegen/printing/ccode.py
```diff
@@ -981,6 +988,14 @@ def _print_Assign(self, expr):
             return '{};'.format(self._print(expr.rhs))
         lhs = self._print(expr.lhs)
         rhs = expr.rhs
+        if isinstance(rhs, IndexedElement) and isinstance(expr.lhs, Variable):
+            free_code = ''
+            if any(isinstance(i, Slice) for i in rhs.args):
+                free_code += 'free_array(%s);\n' % self._print(expr.lhs.name)
```
EmilyBourne, line 994: It would be better to make this:
```
free_code += self._print(Deallocate(expr.lhs))
```
In case we ever want to change `_print_Deallocate`

pyccel/codegen/printing/fcode.py
```diff
@@ -1236,7 +1236,9 @@ def _print_Allocate(self, expr):
             code += 'end if\n'
 
         return code
-
+#-----------------------------------------------------------------------------
+    def _print_Deallocate(self, expr):
+        return ''
```
EmilyBourne, line 1241: If you want to check that you haven't missed any or deallocated twice you could implement this method:
```
return 'deallocate({})'.format(self._print(expr.variable))
```
Fortran makes lots of errors if this is not done properly ^^'

It is not necessary though (deallocate(null) doesn't work in fortran)

pyccel/parser/semantic.py
```diff
@@ -234,6 +238,16 @@ def annotate(self, **settings):
...
 
+        #check if the ast CodeBlock is a Program to add the Garbage collecting
+        cls = (Header, EmptyNode, NewLine, Comment, CommentBlock, Module)
+        body = ast.body
+        is_module = all(isinstance(i,cls) for i in body)
+        if not isinstance(body[-1], Return) and not is_module:
+            dealloc = [Deallocate(i) for i in self._allocs]
+            body += dealloc
+            ast = CodeBlock(body)
+            self._allocs = []
```
EmilyBourne, line 249: This does not look like the right place to me. `annotate` is usually only called on the `Module` or the `Program` i.e. the top layer. Normally I think you want to add `Deallocate` in `FunctionDef`.

I wonder if the `CodeBlock` should know what it's parent is?

pyccel/parser/semantic.py
EmilyBourne: There are quite a lot of removed/added whitespace ^^'

pyccel/parser/semantic.py
```diff
@@ -1402,6 +1415,13 @@ def _assign_lhs_variable(self, lhs, d_var, rhs, new_expressions, is_augassign, *
                     new_expressions.append(Allocate(lhs, shape=lhs.alloc_shape, order=lhs.order, status=status))
                 # ...
 
+                # ...
+                # Add memory deallocation for array variables
+                if lhs.is_ndarray:
+                    # Create Deallocate node
+                    self._allocs.append(lhs)
```
EmilyBourne, line 1422: Is `_allocs` a list? Shouldn't it be a list of lists to handle functions in functions?

pyccel/parser/semantic.py
```diff
@@ -2293,8 +2320,13 @@ def _visit_Return(self, expr, **settings):
         assigns = [self._visit_Assign(e) for e in assigns]
         results = [self._visit_Symbol(i, **settings) for i in return_vars]
 
-        if assigns:
-            expr  = Return(results, CodeBlock(assigns))
+        #add the Deallocate node before the Return node
+        code = []
+        code += (assigns)
+        deallocs = [Deallocate(i) for i in self._allocs]
+        code += (deallocs)
```
EmilyBourne, line 2327: code = assigns + [Deallocate(i) for i in self._allocs]

pyccel/parser/semantic.py
```diff
@@ -2482,8 +2514,13 @@ def _visit_FunctionDef(self, expr, **settings):
             # we annotate the body
             body = self._visit(expr.body)
 
-            args    = [self.get_variable(a.name) if isinstance(a, Variable) else self.get_function(str(a.name)) for a in args]
+            #checking if the Function don't have Return then call the garbage collector
+            if not isinstance(expr.body.args[-1][-1], Return):
+                dealloc = [Deallocate(i) for i in self._allocs]
+                body = CodeBlock([body, CodeBlock(dealloc)])
+            self._allocs = []
```
EmilyBourne, line 2521: What about functions in functions?

EmilyBourne (changes requested):

pyccel/codegen/printing/ccode.py
saniQQ: if the array have the attribute is_pointer == true he will be stored in `stored_in_c_pointer`

saniQQ: and the views have the is_pointer set to true

pyccel/parser/semantic.py
saniQQ: for a Program i think this is the only place I can put it if I don't know the parent from a CodeBlock

pyccel/codegen/printing/ccode.py
EmilyBourne: `stored_in_c_pointer` is a function which tells us whether the value is stored in a c pointer. So in this case if `stored_in_c_pointer` returns true then the type should be `t_ndarray*`. Is that right? If it isn't then it sounds like this needs to be corrected in `stored_in_c_pointer` not every time you use an array variable

pyccel/parser/semantic.py
EmilyBourne: Ah you are right, a Program is only created in the codegen step! Maybe this could be a function that you can call from `_visit_FunctionDef` too to avoid code duplication?

saniQQ: yeah i can make this a function and call it when needed

pyccel/parser/semantic.py
saniQQ: i was testing with C and didn't think about this.
do we have that implemented in C ?
i will try to handle nested function

EmilyBourne: I'm not sure if it's implemented yet but it will be at some point and we don't want to have to look for things like this when we do implement it

pyccel/parser/semantic.py
EmilyBourne: Ah you are probably right. But a `DottedVariable` (should) have the same properties as a Variable and it can be an ndarray too. Imagine an array as a member of a class

pyccel/parser/semantic.py
saniQQ: okey i will try to do it

pyccel/codegen/printing/ccode.py
saniQQ: okey i think i will need to exclude the array from that function because I am not using the arrays as pointer at all so I can free them

tests/pyccel/scripts/headers.pyh
```diff
@@ -0,0 +1,2 @@
+#$ header metavar ignore_at_import=True
+#$ header function f(float)
```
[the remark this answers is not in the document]
EmilyBourne: @saniQQ Whether Said replies or not, these 2 files shouldn't be part of your PR

pyccel/parser/semantic.py
EmilyBourne: Just the bit that you can use from `_visit_FunctionDef` should be enough :+1:

tests/pyccel/scripts/headers.pyh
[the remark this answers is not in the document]
saniQQ: okey i will remove them next push

pyccel/parser/semantic.py
```diff
@@ -238,18 +239,27 @@ def annotate(self, **settings):
...
+            self._ast = ast = self.garbage_collector(ast)
+        self._allocs = []
 
         return ast
 
+    def garbage_collector(self, expr):
+        """
+        Search in a CodeBlock if no trailing Return Node is present add the needed frees.
+
+        Return the same CodeBlock if a trailing Return found otherwise Return a new CodeBlock with additionals Deallocate Nodes.
```
EmilyBourne, line 255: typo `additionals`->`additionals`
grammar:
```
 Return the same CodeBlock if a trailing Return found otherwise Return a new CodeBlock with additionals Deallocate Nodes.
```
->
```
Return the same CodeBlock if a trailing Return is found otherwise Return a new CodeBlock with additional Deallocate Nodes.
```

pyccel/parser/semantic.py
```diff
@@ -238,18 +239,27 @@ def annotate(self, **settings):
...
+        """
+        if not isinstance(expr.body[-1], Return):
+            code = expr.body + [Deallocate(i) for i in self._allocs[-1]]
+            print(code)
```
EmilyBourne, line 259: Presumably this was for debugging?

pyccel/parser/semantic.py
```diff
@@ -238,18 +239,27 @@ def annotate(self, **settings):
...
         cls = (Header, EmptyNode, NewLine, Comment, CommentBlock, Module)
-        body = ast.body
-        is_module = all(isinstance(i,cls) for i in body)
-        if not isinstance(body[-1], Return) and not is_module:
-            dealloc = [Deallocate(i) for i in self._allocs]
-            body += dealloc
-            ast = CodeBlock(body)
-            self._allocs = []
+        is_module = all(isinstance(i,cls) for i in ast.body)
+        if not is_module:
```
EmilyBourne, line 245: Is this `if` needed? If it is a module then won't `self._allocs` be empty anyway?

pyccel/parser/semantic.py
saniQQ: i did a git diff to the wrong folder will fix this

pyccel/parser/semantic.py
```diff
@@ -2511,14 +2519,13 @@ def _visit_FunctionDef(self, expr, **settings):
...
             body = self._visit(expr.body)
 
-            #checking if the Function don't have Return then call the garbage collector
-            if not isinstance(expr.body.args[-1][-1], Return):
-                dealloc = [Deallocate(i) for i in self._allocs]
-                body = CodeBlock([body, CodeBlock(dealloc)])
-            self._allocs = []
+            #Call to the garbage collector
+            body = self.garbage_collector(body)
+            self._allocs = self._allocs[:-1]
```
EmilyBourne, line 2528: See : https://www.techiedelight.com/remove-last-element-from-list-python/
for a discussion about the best way to remove one element form the end of a list

pyccel/parser/semantic.py
EmilyBourne: Does this mean `DottedVariable` is missing a property?

tests/pyccel/scripts/headers.pyh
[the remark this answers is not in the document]
EmilyBourne: You managed to remove one, but tests/pyccel/scripts/test_headers.py still seems to be there

pyccel/parser/semantic.py
```diff
@@ -234,8 +239,27 @@ def annotate(self, **settings):
...
+    def garbage_collector(self, expr):
+        """
+        Search in a CodeBlock if no trailing Return Node is present add the needed frees.
+
+        Return the same CodeBlock if a trailing Return found otherwise Return a new CodeBlock with additionals Deallocate Nodes.
+        """
+        if not isinstance(expr.body[-1], Return):
+            code = expr.body + [Deallocate(i) for i in self._allocs[-1]]
+            print(code)
+            return CodeBlock(code)
```
EmilyBourne, line 260: Can't `self._allocs` be modified in this function?

tests/pyccel/scripts/headers.pyh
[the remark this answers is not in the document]
saidctb: ah yes if you tested that on your computer that will be added automaticaly so we can add them to `.gitignore`

EmilyBourne (changes requested):
Still a few comments and a lot of unnecessary whitespace changes

pyccel/parser/semantic.py
saniQQ: i didn't test with modules or headers i put the check there to be sure that its a program I will try some tests and remove it if it's unnecessary

pyccel/parser/semantic.py
saniQQ: i will change the way i emove the last element and make it part of the function

pyccel/parser/semantic.py
EmilyBourne: @saniQQ ?

saniQQ: if it should have the same properties then yeah it's missing some of the Variable Class properties, I am not so sure to when we use the DottedVariable that's why I can't know for sure if we need the nd_array property

EmilyBourne: We use it when there is a dot in the python expression (so `a.b`). As in C/C++ this happens when `a` is a class instance, but in python it can also happen when `a` is a module (e.g. numpy). It is made of a lhs (`a` in our example) and a rhs (`b` in our example). `b` is a variable so really a `DottedVariable` should have all the same properties. In reality that is hard to maintain so I think we often try to use `rhs` to get the `Variable` properties.

So maybe the line above should be:
```
if (isinstance(l, Variable) and l.is_ndarray) or \
   (isinstance(l, DottedVariable) and l.rhs.is_ndarray):
```

but this isn't very important at the moment given that classes aren't implemented yet

saniQQ:
@saidctb after trying what we discussed having separate free functions is the viable way to do it

pyccel/parser/semantic.py
saniQQ: i will  keep that in mind for next time as that piece of code now is removed

saniQQ:
after more tests, the best way I found is to handle that from pyccel so I can have a difference between a view and a pointer so I added the `is_view` property to the Variable class

pyccel/ast/core.py
```diff
@@ -877,6 +877,51 @@ def __eq__(self, other):
...
+        self._variable = variable
+
+    # ...
+
+    @property
+    def variable(self):
+        return self._variable
+
+    def __repr__(self):
+        return 'Deallocate({})'.format(self.variable.name)
```
mostafaMamoni, line 917: this doesn't print variable name, did you know it ?

pyccel/parser/semantic.py
```diff
@@ -1339,6 +1339,9 @@ def _ensure_target(self, rhs, d_lhs):
         if isinstance(rhs, IndexedElement) and rhs.rank > 0 and rhs.base.internal_variable.allocatable:
             d_lhs['allocatable'] = False
             d_lhs['is_pointer' ] = True
+            if rhs.args[0].rank != len(rhs.args) - 1 or any(isinstance(a, Slice) for a in rhs.args):
+                d_lhs['is_view' ] = True
```
EmilyBourne, line 1343: Does it make sense for an object to be both a view and a pointer?

EmilyBourne, line 1343: I don't understand what is being checked in `rhs.args[0].rank != len(rhs.args) - 1` is there an object with a clearer name than `rhs.args[0]`? Is it `rhs.base.internal_variable`?

pyccel/parser/semantic.py
```diff
@@ -1426,7 +1429,7 @@ def _assign_lhs_variable(self, lhs, d_var, rhs, new_expressions, is_augassign, *
                 # ...
                 # Add memory deallocation for array variables
                 # check if the variable is an ndarray and not astacked variable and that is a view
-                if lhs.is_ndarray and not lhs.is_stack_array and not (lhs.is_pointer and lhs.cls_base):
+                if lhs.is_ndarray and not lhs.is_stack_array and not (lhs.is_pointer and not lhs.is_view):
```
EmilyBourne, line 1432: The expression `not (lhs.is_pointer and not lhs.is_view)` might be clearer as `(lhs.is_view or not lhs.is_pointer)`

pyccel/ast/core.py
```diff
@@ -2268,6 +2321,16 @@ def is_pointer(self, is_pointer):
...
 
+    @property
+    def is_view(self):
+        return self._is_view
+
+    @is_view.setter
+    def is_view(self, is_view):
+        if not isinstance(is_view, bool):
+            raise TypeError('is_view must be a boolean.')
+        self._is_view = is_view
```
EmilyBourne, line 2332: Do we need a setter?

pyccel/parser/semantic.py
saniQQ: you can say a view is a specific case of a pointer

saniQQ: a view own his shape and strides unlike a pointer

saidctb: we don't need the `is_view` anymore @saniQQ , no ?

saniQQ: after we fixed the bug last time I had to use it so I can free the view when reassigning pointers

pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -40,7 +40,7 @@ t_ndarray   array_create(int32_t nd, int32_t *shape, enum e_types type)
             arr.type_size = sizeof(double complex);
             break;
     }
-    arr.is_slice = false;
+    arr.is_view = false;
```
EmilyBourne, line 43: I don't see any code that sets `is_view = true`. Are we missing a function that would be called from `_print_AliasAssign`?

saniQQ: i am setting it to True in the func `t_ndarray array_slicing(t_ndarray p, ...)`  line 200

pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -148,19 +148,29 @@ void   _array_fill_cdouble(double complex c, t_ndarray arr)
...
+    if (dump.shape == NULL)
+        return (0);
+    free(dump.raw_data);
+    dump.raw_data = NULL;
     free(dump.shape);
     dump.shape = NULL;
     free(dump.strides);
     dump.strides = NULL;
     return (1);
 }
```
MenoIy, line 160: do you use returns values somewhere in the code, or are planning to implement some error management later?

pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -148,19 +148,29 @@ void   _array_fill_cdouble(double complex c, t_ndarray arr)
...
+int32_t free_pointer(t_ndarray dump)
+{
+    if (dump.is_view == false || dump.shape == NULL)
+        return (0);
+    free(dump.shape);
+    dump.shape = NULL;
+    free(dump.strides);
+    dump.strides = NULL;
+    return (1);
+}
```
MenoIy, line 171: same here.

MenoIy (changes requested):

.gitignore
```diff
@@ -50,6 +50,9 @@ tests/scripts/core/*.f90
 tests/scripts/core/*.mod
 tests/scripts/core/*.o
 
+tests/pyccel/scripts/headers.pyh
+tests/pyccel/scripts/test_headers.py
```
rakati, line 54: why you ignore these files?

saniQQ: we discussed about it in the link below
https://github.com/pyccel/pyccel/pull/574#discussion_r539303729

pyccel/errors/messages.py
```diff
@@ -123,6 +123,10 @@
 INVALID_FILE_EXTENSION = 'Wrong file extension. Expecting `py` of `pyh`, but found'
 INVALID_PYTHON_SYNTAX = 'Python syntax error'
 
+# ARRAY ERRORS
+ASSIGN_ARRAYS_ONE_ANOTHER = 'Arrays which own their data cannot become views on other arrays'
```
rakati, line 127: please change `Arrays which own their data` to `Arrays that own their data`

pyccel/errors/messages.py
```diff
@@ -123,6 +123,10 @@
...
+ARRAY_ALREADY_IN_USE = 'Attempt to reallocate an array which is being used by another variable'
```
rakati, line 128: same here, `Attempt to reallocate an array that is being used by another variable`

EmilyBourne: I don't agree with this. I've googled and found:
https://www.grammarly.com/blog/which-vs-that/

The section "Which vs. That: What’s the Difference in Usage?" agrees with my gut feeling that "which" is better

pyccel/errors/messages.py
EmilyBourne: Same as below

pyccel/stdlib/ndarrays/ndarrays.c
EmilyBourne: Ah you're right. Sorry!

pyccel/stdlib/ndarrays/ndarrays.c
saniQQ: i think it might be useful to anyone that want to make changes to the generated C code

pyccel/stdlib/ndarrays/ndarrays.c
saniQQ: i think it might be useful to anyone that want to make changes to the generated C code

pyccel/codegen/printing/ccode.py
```diff
@@ -1088,6 +1088,13 @@ def _print_AliasAssign(self, expr):
 
         lhs = self._print(lhs.name)
         rhs = self._print(rhs)
+
+        # the below condition handles the case of reassinging a pointer to an array view.
+        # setting the pointer's is_view attribute to false so it can be ignored by the free_pointer function.
+        if isinstance(expr.lhs, Variable) and expr.lhs.is_ndarray \
+                and isinstance(expr.rhs, Variable) and expr.lhs.is_ndarray and expr.rhs.is_pointer:
+            return '{0} = {1};\n{0}.is_view = false;'.format(lhs, rhs)
+
```
MenoIy, line 1097: can you please provide an example and where it fails?

pyccel/parser/semantic.py
```diff
@@ -1459,6 +1459,12 @@ def _assign_lhs_variable(self, lhs, d_var, rhs, new_expressions, is_augassign, *
                             bounding_box=(self._current_fst_node.lineno, self._current_fst_node.col_offset),
                             severity='fatal', blocker=False)
 
+                elif var.is_ndarray and var.is_pointer:
+                        # we allow pointers to be reassigned multiple times
+                        # pointers reassigning need to call free_pointer func
+                        # to remove memory leaks
+                        new_expressions.append(Deallocate(var))
```
MenoIy: why did you move this condition out of if block?
```
elif d_var['shape'] != shape:   # line 1504
```

MenoIy (changes requested):

pyccel/parser/semantic.py
EmilyBourne: I don't see this modification

pyccel/codegen/printing/ccode.py
saniQQ:
``` python
a = np.array([1, 2, 3, 4, 5])
b = a[1:]
e = b # this change is to fix this problem case because we are allowing reassinging pointer freely
```

``` C
int main()
{
    t_ndarray a;
    t_ndarray b;
    t_ndarray e;



    a = array_create(1, (int32_t[]){5}, nd_int64);
    int64_t array_dummy_0001[] = {1, 2, 3, 4, 5};

    b = array_slicing(a, new_slice(1, 5, 1));
    e = b;
    e.is_view = false;  //this change is to add this line so I can ignore the variable e in the garbage collecting
                                // because it copies the information of b and b is a view
    free_array(a);
    free_pointer(b);
    free_pointer(e);
    return 0;
}
```

pyccel/parser/semantic.py
saniQQ: the reassign of a view need to be deallocated even if the shapes are the same
they can be view of different arrays
so I needed to make the check outside the if block of that checks the shape

pyccel/codegen/printing/ccode.py
EmilyBourne: Do you mean that `e` `is_ndarray` instead of `is_pointer`? That's not right :/ 

It looks to me like the actual problem is here:
https://github.com/pyccel/pyccel/blob/master/pyccel/parser/semantic.py#L1308
`rhs.is_allocatable` should be `(rhs.is_ndarray or rhs.is_pointer)`

But I thought the pointer case was handled [here](https://github.com/pyccel/pyccel/blob/master/pyccel/parser/semantic.py#L632)

saniQQ: the problem is not from pyccel side.
the problem is we can't know If the variable in C is a view or not statically so i need to add somethings to make a difference so the garbage collector can  ignore the pointers that don't own (shape/strides)

pyccel/parser/semantic.py
```diff
@@ -1430,6 +1459,24 @@ def _assign_lhs_variable(self, lhs, d_var, rhs, new_expressions, is_augassign, *
...
+                    errors.report(ARRAY_ALREADY_IN_USE,
+                        bounding_box=(self._current_fst_node.lineno,
+                            self._current_fst_node.col_offset),
+                                severity='error', symbol=var.name)
+
+                elif var.is_ndarray and var.is_pointer:
+                    # we allow pointers to be reassigned multiple times
+                    # pointers reassigning need to call free_pointer func
+                    # to remove memory leaks
+                    new_expressions.append(Deallocate(var))
```
EmilyBourne, line 1478: I wonder if we should find a way to differentiate between deallocating memory and deallocating shapes. In fortran we can call `allocate` on both `allocatable` variables and `pointer` variables. The only differences are that `allocatable` must be contiguous, and `allocatable` is automatically deallocated at the end of the function/subroutine. We will therefore need to allocate `pointer` variables for issue #537. once this is done the current implementation of `Deallocate` might prevent us from writing a `_print_Deallocate` for fortran.

What do you think?

pyccel/codegen/printing/ccode.py
EmilyBourne: So this tells C that `b` owns its shape/strides, but not the array data, but `e` owns none of it's pointers?

saniQQ: exactly

pyccel/parser/semantic.py
saniQQ: the problem with pointers is that we can't know if they own their data or not statically

saniQQ: we can have call `_print_Deallocate` only for pointers but there is no way to know whats the pointer is holding statically 
- because we are not tracking the changes that occurs to a pointer variables (don't know if this is a bug)
- we can't determine the outcome of an `if block`

pyccel/codegen/printing/ccode.py
```diff
@@ -1131,7 +1131,8 @@ def _print_AliasAssign(self, expr):
         # setting the pointer's is_view attribute to false so it can be ignored by the free_pointer function.
         if isinstance(expr.lhs, Variable) and expr.lhs.is_ndarray \
                 and isinstance(expr.rhs, Variable) and expr.lhs.is_ndarray and expr.rhs.is_pointer:
-            return '{0} = {1};\n{0}.is_view = false;'.format(lhs, rhs)
+            self._additional_imports.add("ndarrays")
```
MenoIy, line 1134: is this necessary?

MenoIy (changes requested):

pyccel/codegen/printing/ccode.py
saniQQ: should remove it 👍

pyccel/ast/core.py
```diff
@@ -877,6 +877,48 @@ def __eq__(self, other):
...
+
+    Parameters
+    ----------
+    variable : pyccel.ast.core.Variable
+        The typed variable (usually an array) that needs memory deallocation.
+
+    Notes
+    -----
+    An object of this class is immutable, although it contains a reference to a
+    mutable Variable object.
```
mostafaMamoni, line 895: you shouldn't capitalize `variable`

pyccel/ast/core.py
```diff
@@ -877,6 +877,48 @@ def __eq__(self, other):
...
+#------------------------------------------------------------------------------
+class Deallocate(Basic):
+    """
+    Represents memory deallocation (usually of an array) for code generation.
+    This is relevant to low-level target languages, such as C or Fortran,
+    where the programmer must take care of heap memory deallocation.
+
+    Parameters
+    ----------
+    variable : pyccel.ast.core.Variable
```
mostafaMamoni, line 889: you should capitalize `variable`

pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -214,11 +214,31 @@ t_ndarray array_slicing(t_ndarray p, ...)
     return (slice);
 }
 
+/*
+** assigns
+*/
+
+void        alias_assign(t_ndarray *dest, t_ndarray src)
```
MenoIy, line 221: can you add a test where this function is used?

pyccel/ast/core.py
saniQQ: I think it's correct to capitalize when we are talking about a class object

pyccel/ast/core.py
saniQQ: See above

MenoIy (approved):

mostafaMamoni (approved):

pyccel/codegen/printing/ccode.py
```diff
@@ -1119,6 +1126,13 @@ def _print_AliasAssign(self, expr):
 
         lhs = self._print(lhs.name)
         rhs = self._print(rhs)
+
+        # the below condition handles the case of reassinging a pointer to an array view.
+        # setting the pointer's is_view attribute to false so it can be ignored by the free_pointer function.
+        if isinstance(expr.lhs, Variable) and expr.lhs.is_ndarray \
+                and isinstance(expr.rhs, Variable) and expr.lhs.is_ndarray and expr.rhs.is_pointer:
```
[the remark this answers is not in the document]
saniQQ: will change it meant to check the rhs

tests/epyccel/test_arrays_multiple_assignments.py
```diff
@@ -165,6 +166,59 @@ def f(c):
...
+
+    # Check that we got exactly 1 Pyccel error
+    assert errors.has_errors() == 1
+    assert errors.num_messages() == 1
+
+    # Check that the error is correct
+    error_info = [*errors.error_info_map.values()][0][0]
+    assert error_info.symbol  == 'x'
+    assert error_info.message == ASSIGN_ARRAYS_ONE_ANOTHER
+
```
EmilyBourne, line 221: Can the tests in this file be activated for C now?

saniQQ: we are using the NumpySum / NumpyRand node are not implemented in C yet

EmilyBourne: So then none of this is tested at all?

saniQQ: they tests that I added are for raising errors in the semantic stage

EmilyBourne: I thought `test_arrays_multiple_assignments.py` tested for an error raised in the semantic stage? Can't we activate those tests? The codegen stage isn't carried out if the semantic step raises an error

EmilyBourne: Could we activate the tests in `test_pointers.py`?

saniQQ: i need to remove the shape function from the tests/epyccel/modules/pointers.py

saniQQ: the shape func is still not implemented in C

EmilyBourne: In that case could you add some new more basic tests that should work already

saniQQ: for test_pointers.py we can actually activate it now I thought that we were using the shape method node numpyshape

EmilyBourne: We are using the numpy shape method. But in pyccel that is implemented by returning either the known value of the shape, or an instance of `PyccelArraySize`

EmilyBourne: > the tests that I added are for raising errors in the semantic stage

So the tests that you added should work for any language

saniQQ: sorry, I just got confused, I activated the tests for pointers and add a test for the case of the reassign of a pointer

tests/epyccel/test_arrays_multiple_assignments.py
```diff
@@ -165,6 +166,59 @@ def f(c):
     c = np.random.random()
     assert f(c) == g(c)
 
+#==============================================================================
+def test_Reassign_to_Target(language):
```
EmilyBourne, line 170: This tests whether an error is raised in the semantic stage. This means that the codegen stage is never reached, so the language argument is pointless

tests/epyccel/test_arrays_multiple_assignments.py
```diff
@@ -165,6 +166,59 @@ def f(c):
...
+    assert errors.has_errors() == 1
+    assert errors.num_messages() == 1
+
+    # Check that the error is correct
+    error_info = [*errors.error_info_map.values()][0][0]
+    assert error_info.symbol  == 'x'
+    assert error_info.message == ARRAY_ALREADY_IN_USE
+
+#==============================================================================
+def test_Assign_Between_Allocatables(language):
```
EmilyBourne, line 196: See above. language argument does nothing

EmilyBourne (changes requested):

tests/epyccel/test_arrays_multiple_assignments.py
saniQQ: will change it

EmilyBourne (approved):
If you can think of a good way of doing it, it would be nice to have a test in tests/pyccel which generates an executable which you can test for memory leaks with valgrind

EmilyBourne:
The test `test_ones_dtype` in `tests/epyccel/recognised_functions/test_numpy_funcs.py` is also xpassing

saniQQ:
@rakati is there a problem with casting the value that are passed to array_fill all the times ?
the Generic can map to the correct function if the type is not explicitly casted

pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -192,7 +202,7 @@ t_ndarray array_slicing(t_ndarray p, ...)
     slice.shape = malloc(sizeof(int32_t) * p.nd);
     slice.strides = malloc(sizeof(int32_t) * p.nd);
     memcpy(slice.strides, p.strides, sizeof(int32_t) * p.nd);
-    slice.is_slice = true;
+    slice.is_view = true;
     va_start(va, p);
```
yguclu, line 206: Sorry to ask for this, but I must have missed it in the previous PR. I think it is **really** confusing to call `slice` an array variable, and then call `slice_data` a slice variable. Can you call them `view` (or `arr`) and `slice` please?
```C
t_ndarray view;
t_slice slice;
```
If possible, I would like to use the same names throughout this whole file. We can chat about this.

tests/codegen/ccode/scripts/arrays_pointers.py
```diff
@@ -0,0 +1,25 @@
...
+
+def pointer_to_pointer():
+
+    from numpy import array
+
+    a = array([1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
+    b = a
+    c = b #pylint:disable=unused-variable
+
+def reassinging_pointers():
```
yguclu, line 19: Typo: `reassinging` -> `reassigning` (note the swap `ng` -> `gn`)

But you can just call this function `reassign_pointers`, which is short and clear.

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -1364,7 +1364,7 @@ def create_full_shape_F(n,m):
 @pytest.mark.parametrize( 'language', (
         pytest.param("fortran", marks = pytest.mark.fortran),
         pytest.param("c", marks = [
-            pytest.mark.skip(reason="arrays not implemented"),
+            pytest.mark.skip(reason="casting to complex in not handled corectly"),
```
yguclu, line 1367: Typo: `corectly` -> `correctly`

yguclu (changes requested):
Excellent job @saniQQ!
I have a few minor comments.

pyccel/stdlib/ndarrays/ndarrays.c
saniQQ: i will try to start only using (arr , view, slice) throughout the whole file.

yguclu (approved):

pyccel/stdlib/ndarrays/ndarrays.c
yguclu: Thanks!

Issue #549, opened 2020-11-18
Duplicated decorators


rmahjoubi:
Description
----------
As a USER, I want to decorate a function with the same decorator twice, applied to different variables.

At the moment this does not work, because only the first decorator is correctly processed.

**To Reproduce**
Provide code to reproduce the behavior:
```
from pyccel.decorators import allow_negative_index

@allow_negative_index('a')
@allow_negative_index('b')
def func():
   a = np.array([1,2,3])
   b = np.array([1,2])
   c = -1
   x = a[c]
   y = b[c]
   return 0
```
In this example allow_negative_index is not enabled in the array b.

**Language**
Fortran, C

Acceptance Criteria
--------------------
- Keep all the information for the following decorators when applied multiple times:
  - `@stack_array`
  - `@allow_negative_index`
- Put mechanism in place for any future decorator applied on a variable

PR #578, opened 2020-12-02
Process multiple decorators of the same kind

MenoIy:
Process multiple decorators of the same kind applied to a function. This solves #549.

- Always collect decorators as a list;
- Loop on dict value and collect arguments in allow_negative_index and stack_array;
- Add unit tests.


EmilyBourne:
Does this PR have the right name?

EmilyBourne (approved):

yguclu (approved):

Issue #550, opened 2020-11-18
Clean up the class FunctionDef


rmahjoubi:
FunctionDef class should get cleaned up, and get rid of all the functions we're not using in this class anymore.

- [x] List everything currently in FunctionDef class
- [ ] Add docstrings to methods / properties that are definitely being used
- [x] Identify which methods / properties are obsolete and remove them

PR #583, opened 2020-12-04
Clean up FunctionDef class

mostafaMamoni:
how i cleaned up Functiondef class:
put sys.exit(property/method name) in every property/method. so every exit status of pyccel has the name of a property/method, and then every property/method still contains sys.exit(). It's definitely useless according to the current tests.


pyccel/ast/core.py
```diff
@@ -3167,6 +3168,7 @@ def local_vars(self):
 
     @property
     def global_vars(self):
+        sys.exit("global vars in functiondef\n")
         return self._global_vars
```
saidctb, line 3172: we need to keep this one, in order to have access to the global variables used in the function

pyccel/ast/core.py
```diff
@@ -3183,6 +3185,7 @@ def kind(self):
 
     @property
     def is_static(self):
+        sys.exit("is static in functiondef\n")
```
saidctb, line 3188: this can probably be removed

pyccel/ast/core.py
```diff
@@ -3238,13 +3241,15 @@ def doc_string(self):
         return ""
 
     def print_body(self):
+        sys.exit("print body in functiondef\n")
```
saidctb, line 3244: this also

pyccel/ast/core.py
```diff
@@ -3277,6 +3282,7 @@ def rename(self, newname):
         self._name = newname
 
     def vectorize(self, body , header):
```
saidctb, line 3284: we need to keep this in order to vectorize a function

saidctb (approved):

pyccel/ast/core.py
EmilyBourne: @saidctb Is the vectorizing currently implemented? It doesn't look like this function does the vectorizing. Are we sure that this is how the vectorizing will be done when it is implemented?

pyccel/ast/core.py
```diff
@@ -3374,12 +3362,10 @@ def __reduce_ex__(self, i):
 
     # TODO
     def check_pure(self):
-        sys.exit("check_pure in functiondef\n")
         raise NotImplementedError('')
 
     # TODO
     def check_elemental(self):
-        sys.exit("check_elemental in functiondef\n")
         raise NotImplementedError('')
```
EmilyBourne: @saidctb Presumably these functions can also be removed as they are not even implemented?

yguclu: I agree. These two functions must be removed!

pyccel/ast/core.py
yguclu: I am also perplexed by this method, which is doing operations that do not make much sense to me.
For example, the line
```python
self._kind = procedure
```
should read
```python
self._kind = 'procedure'
```
Further, I don't understand why `body` should be an argument to this method, it looks like the vectorization is really performed by another function that we do not have. And finally, we do not even have a `vectorize` decorator!

@saidctb I am strongly in favor of removing this function... We can copy it to some "obsolete" folder if you really care about it.

saidctb: it's okay we can delete it

pyccel/ast/core.py
saidctb: yes

EmilyBourne (approved):

EmilyBourne:
@mostafaMamoni  Does this PR fix an issue (eg #550)? If so, please mark it as such either by modifying the PR description or by linking the issue manually using the bar on the right hand side of the page

yguclu (approved):

PR #584, opened 2020-12-04
Remove fprint methods from ast.numpyext

mostafaMamoni:
* Remove fprint methods from `ast.numpyext `and `ast.core` and reimplement them in `Fcodeprinter`;
* Add a new AST node `NumpyAutoFill`, with `NumpyZeros` and `NumpyOnes` inheriting from it instead of `NumpyEmpty`, in order to:
   - Correctly print `NumpyEmpty` objects in an `Assign` node without catching instances of `NumpyZeros` and `NumpyOnes`;
   - Avoid repeating the same __new__ function as `NumpyEmpty`.


pyccel/codegen/printing/fcode.py
```diff
@@ -1089,12 +1291,53 @@ def _print_Assign(self, expr):
...
+        
+        if isinstance(rhs, (NumpyComplex, NumpyReal, NumpyArray, NumpyLinspace)):
+            rhs = self._print(rhs)
+            return '{0} = {1}\n'.format(lhs_code,rhs)
+        
+        if isinstance(rhs, (NumpyCross, NumpyDiag, NumpyWhere)):
+            return self._print(rhs, expr.lhs)
+        
+        if isinstance(rhs, PyccelArraySize):
             return rhs.fprint(self._print, expr.lhs) + '\n'
```
EmilyBourne, line 1310: Shouldn't this fprint be handled too?

pyccel/codegen/printing/fcode.py
```diff
@@ -1089,12 +1291,53 @@ def _print_Assign(self, expr):
...
+        if isinstance(rhs, NumpyRand):
+
+            stack_array = False
+            if self._current_function:
+                name = self._current_function
+                func = self.get_function(name)
+                lhs_name = expr.lhs.name
+                vars_dict = {i.name: i for i in func.local_vars}
+                if lhs_name in vars_dict:
+                    stack_array = vars_dict[lhs_name].is_stack_array
```
EmilyBourne, line 1322: This doesn't look like it should be in the printing. Was it there before? @saidctb do you know what this does?

pyccel/codegen/printing/fcode.py
```diff
@@ -1089,12 +1291,53 @@ def _print_Assign(self, expr):
...
+        if isinstance(rhs, NumpyFull):
+
+            stack_array = False
+            if self._current_function:
+                name = self._current_function
+                func = self.get_function(name)
+                lhs_name = expr.lhs.name
+                vars_dict = {i.name: i for i in func.local_vars}
+                if lhs_name in vars_dict:
+                    stack_array = vars_dict[lhs_name].is_stack_array
```
EmilyBourne, line 1335: This code looks identical to the code above. Why not create a function?

EmilyBourne (changes requested):

pyccel/codegen/printing/fcode.py
```diff
@@ -611,26 +611,199 @@ def _print_PythonImag(self, expr):
 
     #========================== Numpy Elements ===============================#
 
-    def _print_NumpySum(self, expr):
-        return expr.fprint(self._print)
+    def _print_NumpySum(self, expr, lhs=None):
```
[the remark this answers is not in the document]
EmilyBourne: This is not the only place where this is done

pyccel/codegen/printing/fcode.py
```diff
@@ -611,26 +611,199 @@ def _print_PythonImag(self, expr):
...
+        """Fortran print."""
+
+        rhs_code = self._print(expr.arg)
+        if lhs:
+            lhs_code = self._print(lhs)
+            return '{0} = sum({1})'.format(lhs_code, rhs_code)
```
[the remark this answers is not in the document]
EmilyBourne: This is not the only place where this is done

pyccel/codegen/printing/fcode.py
```diff
@@ -1089,12 +1291,53 @@ def _print_Assign(self, expr):
...
+            stack_array = False
+            if self._current_function:
+                name = self._current_function
+                func = self.get_function(name)
+                lhs_name = expr.lhs.name
+                vars_dict = {i.name: i for i in func.local_vars}
+                if lhs_name in vars_dict:
+                    stack_array = vars_dict[lhs_name].is_stack_array
+
+            return self._print_NumpyFull(rhs, lhs_code)
```
EmilyBourne, line 1337: We should never call the `_print_ClassName` functions by name. The calls should always go through a call to `_print`

pyccel/codegen/printing/fcode.py
```diff
@@ -611,26 +611,199 @@ def _print_PythonImag(self, expr):
...
+        order = expr.order
+
+        if lhs is not None:
+            lhs  = self._print(lhs)
+
+            if rank == 2:
+                alloc = 'allocate({0}(0:size({1},1)-1,0:size({1},2)-1))'.format(lhs, first)
+
+            elif rank == 1:
+                alloc = 'allocate({}(0:size({})-1)'.format(lhs, first)
```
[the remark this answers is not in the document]
mostafaMamoni: i removed this and the return statement for codacy.

pyccel/codegen/printing/fcode.py
```diff
@@ -705,7 +878,29 @@ def _print_NumpyRandint(self, expr):
...
-        return expr.fprint(self._print)
+        if expr.high is None:
+            randreal = self._print(PyccelMul(expr.low, NumpyRand()))
+        else:
+            randreal = self._print(PyccelAdd(PyccelMul(PyccelAssociativeParenthesis(PyccelMinus(expr.high, expr.low)), NumpyRand()), expr.low))
+
+        prec_code = self._print(expr.precision)
+        return 'floor({}, kind={})'.format(randreal, prec_code)
+    
+    def _print_NumpyFull(self, expr, lhs_code):
```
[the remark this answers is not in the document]
mostafaMamoni: what about the if statements here?

pyccel/codegen/printing/fcode.py
mostafaMamoni: its not in ast.numpyext

pyccel/codegen/printing/fcode.py
```diff
@@ -1089,12 +1291,53 @@ def _print_Assign(self, expr):
...
+        if isinstance(rhs, PythonInt):
+            rhs = expr.rhs.fprint(self._print)
+            return '{0} = {1}\n'.format(lhs_code,rhs)
+        
+        if isinstance(rhs, (NumpyComplex, NumpyReal, NumpyArray, NumpyLinspace)):
+            rhs = self._print(rhs)
+            return '{0} = {1}\n'.format(lhs_code,rhs)
+        
+        if isinstance(rhs, (NumpyCross, NumpyDiag, NumpyWhere)):
+            return self._print(rhs, expr.lhs)
```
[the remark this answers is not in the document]
mostafaMamoni: you're right for NumpyCross, but the others I need a confirmation from you because I will move many outside the method because of lhs argument.

pyccel/codegen/printing/fcode.py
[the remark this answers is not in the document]
saidctb: we can delet numpy cross for the moment

pyccel/codegen/printing/fcode.py
```diff
@@ -611,26 +611,199 @@ def _print_PythonImag(self, expr):
...
+
+    def _print_NumpyWhere(self, expr, lhs):
+
+        ind   = self._print(expr.index)
+        mask  = self._print(expr.mask)
+        lhs   = self._print(lhs)
+
+        stmt  = 'pack([({ind},{ind}=0,size({mask})-1)],{mask})'.format(ind=ind,mask=mask)
+        stmt  = '{lhs}(:,0) = {stmt}'.format(lhs=lhs, stmt=stmt)
+        alloc = 'allocate({}(0:count({})-1,0:0))'.format(lhs, mask)
```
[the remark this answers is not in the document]
saidctb: we can delete them for the moment

pyccel/codegen/printing/fcode.py
saidctb: this should be removed, it handles stack arrays , so that we don't allocate them

saidctb: it's probably not needed anymore

pyccel/codegen/printing/fcode.py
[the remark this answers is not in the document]
EmilyBourne: The `if` statement is included to handle the case of `NumpyEmpty`. I think there should be an `if` statement in the `Assign` to handle this case.

pyccel/codegen/printing/fcode.py
```diff
@@ -612,25 +612,186 @@ def _print_PythonImag(self, expr):
...
+        """Fortran print."""
 
-    def _print_NumpyProduct(self, expr):
-        return expr.fprint(self._print)
+        rhs_code = self._print(expr.arg)
+        return 'sum({0})'.format(rhs_code)
 
-    def _print_NumpyMatmul(self, expr):
-        return expr.fprint(self._print)
+    def _print_NumpyProduct(self, expr, lhs=None):
```
EmilyBourne, line 620: See above. `lhs` shouldn't be here

pyccel/codegen/printing/fcode.py
```diff
@@ -612,25 +612,186 @@ def _print_PythonImag(self, expr):
...
+        """Fortran print."""
+
+        rhs_code = self._print(expr.arg)
+        if lhs:
+            lhs_code = self._print(lhs)
+            return '{0} = product({1})'.format(lhs_code, rhs_code)
+        return 'product({0})'.format(rhs_code)
+
+    def _print_NumpyMatmul(self, expr, lhs=None):
```
EmilyBourne, line 629: See above. `lhs` shouldn't be here

pyccel/codegen/printing/fcode.py
```diff
@@ -612,25 +612,186 @@ def _print_PythonImag(self, expr):
...
+                         a[2]*b[0]-a[0]*b[2],
+                         a[0]*b[1]-a[1]*b[0]]
+
+        cross_product = PythonTuple(cross_product)
+        cross_product = self._print(cross_product)
+        first = self._print(expr.first)
+        order = expr.order
+
+        if lhs is not None:
+            lhs  = self._print(lhs)
```
EmilyBourne, line 695: `lhs` is no longer an argument, so you can't use it in your code

pyccel/codegen/printing/fcode.py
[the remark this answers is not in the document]
EmilyBourne: I suppose we can always collect the implementation from the history when we want to implement it. So then the following classes should be removed:
- NumpyCross
- NumpyWhere
- NumpyDiag
? @mostafaMamoni have I missed any classes which need the `lhs`?

pyccel/codegen/printing/fcode.py
```diff
@@ -705,7 +866,28 @@ def _print_NumpyRandint(self, expr):
         if expr.rank != 0:
             errors.report(FORTRAN_ALLOCATABLE_IN_EXPRESSION,
                           symbol=expr, severity='fatal')
-        return expr.fprint(self._print)
+        if expr.high is None:
+            randreal = self._print(PyccelMul(expr.low, NumpyRand()))
+        else:
+            randreal = self._print(PyccelAdd(PyccelMul(PyccelAssociativeParenthesis(PyccelMinus(expr.high, expr.low)), NumpyRand()), expr.low))
```
EmilyBourne, line 872: Is `PyccelAssociativeParenthesis` still needed here? It should be added automatically now

pyccel/codegen/printing/fcode.py
```diff
@@ -705,7 +866,28 @@ def _print_NumpyRandint(self, expr):
...
+
+        prec_code = self._print(expr.precision)
+        return 'floor({}, kind={})'.format(randreal, prec_code)
```
EmilyBourne, line 875: Does this print the correct ISO_C precision?

pyccel/codegen/printing/fcode.py
EmilyBourne: ok

pyccel/codegen/printing/fcode.py
EmilyBourne: Given that allocation is handled with the `Allocate` node that sounds right to me. @mostafaMamoni this code, and the identical code below it should be removed

pyccel/codegen/printing/fcode.py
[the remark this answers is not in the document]
EmilyBourne: So I have checked this locally. With a few small fixes, the `NumpyCross` class still works. However it is not tested at all, only a subset of `numpy.cross` is actually implemented, and I think that in the future the function will be implemented using the method in PR #562 . Given that fact, I agree that we should just remove the `NumpyCross` class for now

pyccel/codegen/printing/fcode.py
```diff
@@ -705,7 +853,21 @@ def _print_NumpyRandint(self, expr):
...
+        prec_code = self._print(iso_c_binding["integer"][expr.precision])
+        return 'floor({}, kind={})'.format(randreal, prec_code)
+
+    def _print_NumpyFull(self, expr):
+
+        # Create statement for initialization
+        if expr.fill_value is not None:
+            init_value = self._print(expr.fill_value)
+
+        return init_value
```
EmilyBourne, line 870: If `expr.fill_value is None` then `init_value` has not been initialised and this function call will crash

pyccel/codegen/printing/fcode.py
```diff
@@ -1089,23 +1256,40 @@ def _print_Assign(self, expr):
...
-                    stack_array = vars_dict[lhs_name].is_stack_array
+        if isinstance(rhs, NumpyFull):
+            stmts = []
+            if rhs.fill_value is not None:
+                code_init = '{0} = {1}'.format(lhs_code, self._print(rhs))
+                stmts.append(code_init)
+            if len(stmts) == 0:
+                return ''
+            else:
+                return '\n'.join(stmts) + '\n'
```
EmilyBourne, line 1288: This can be much simpler:
```
if isinstance(rhs, NumpyEmpty):
    return ''
```

pyccel/codegen/printing/fcode.py
```diff
@@ -1089,23 +1256,40 @@ def _print_Assign(self, expr):
...
 
-            return rhs.fprint(self._print, expr.lhs, stack_array)
+        if isinstance(rhs, (NumpyFullLike, NumpyEmptyLike,\
+						NumpyZerosLike, NumpyOnesLike)):
+            return rhs.fprint(self._print, expr.lhs)
```
EmilyBourne, line 1292: Why does this `fprint` still exist?

pyccel/codegen/printing/fcode.py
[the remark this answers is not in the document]
mostafaMamoni: No, for NumpyFullLike, NumpyOnesLike, NumpyZeroesLike and NumpyEmpyLlike will be handled as well as NumpyFull.

EmilyBourne:
The test `tests/errors/codegen/fortran/FORTRAN_ALLOCATABLE_IN_EXPRESSION.py` is failing. This was supposed to raise an error about allocatables in expressions, however it is no longer needed now that we are using an `Allocatable` node. The code:
```
import numpy as np
a = np.ones(4)*5
```
is correctly translated to:
```
program prog_FORTRAN_ALLOCATABLE_IN_EXPRESSION

use ISO_C_BINDING

implicit none

real(C_DOUBLE), allocatable :: a(:)

allocate(a(0:4_C_INT64_T - 1_C_INT64_T))
a = 1.0_C_DOUBLE * 5_C_INT64_T

end program prog_FORTRAN_ALLOCATABLE_IN_EXPRESSION
```
This test can therefore be removed.

pyccel/codegen/printing/fcode.py
mostafaMamoni: because their fprints don't exit in ast.numpyext, Is the printer of NumpyFull will be enough for those? if so they will be handled with _print_NumpyFull with some modifications.

pyccel/codegen/printing/fcode.py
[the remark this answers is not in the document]
EmilyBourne: You will need to make one additional modification to deal with stupid users ;) . If a user does:
```
a = np.empty(3)*5
```
Then [this error](https://github.com/pyccel/pyccel/blob/master/pyccel/codegen/printing/fcode.py#L2574-L2575) should be raised. This can be done by adding a specialised `_print_NumpyEmpty` function now that `NumpyOnes` and `NumpyZeros` no longer inherit from `NumpyEmpty`. You could modify `tests/errors/codegen/fortran/FORTRAN_ALLOCATABLE_IN_EXPRESSION.py` to test this

pyccel/codegen/printing/fcode.py
```diff
@@ -612,25 +612,173 @@ def _print_PythonImag(self, expr):
...
+
+        # C ordering
+        return 'matmul({1},{0})'.format(a_code, b_code)
+
+    def _print_NumpyImag(self, expr):
+        """Fortran print."""
+
+        value = self._print(expr.arg)
+        code = 'aimag({0})'.format(value)
+        return code
```
EmilyBourne, line 647: This class inherits from `PythonImag` now. Does it need a specialised `_print_NumpyImag` method or does the `_print_PythonImag` function do the same thing?

pyccel/codegen/printing/fcode.py
EmilyBourne: If we never expect `expr.fill_value` to be `None` then the `if` is unnecessary

pyccel/codegen/printing/fcode.py
EmilyBourne: Do you know if there are many `fprint` functions outside of `ast.numpyext`? If there aren't very many then it might not be worth doing this in 2 PRs

pyccel/codegen/printing/fcode.py
```diff
@@ -1089,23 +1255,23 @@ def _print_Assign(self, expr):
...
-            if self._current_function:
-                name = self._current_function
-                func = self.get_function(name)
-                lhs_name = expr.lhs.name
-                vars_dict = {i.name: i for i in func.local_vars}
-                if lhs_name in vars_dict:
-                    stack_array = vars_dict[lhs_name].is_stack_array
+        if isinstance(rhs, (NumpyWhere, NumpyCross, NumpyComplex, NumpyReal, NumpyArray, NumpyLinspace)):
+            rhs = self._print(rhs)
+            return '{0} = {1}\n'.format(lhs_code,rhs)
```
EmilyBourne, line 1268: This block isn't needed. The same calls occur if you remove this code

pyccel/codegen/printing/fcode.py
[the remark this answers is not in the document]
EmilyBourne: Ok. In fact it seems that `NumpyWhere` also doesn't need the `lhs` so there is no need to remove it. So the only classes that need removing are:
- NumpyCross
- NumpyDiag

The whole class can be deleted, not just the implementation of its printing

pyccel/codegen/printing/fcode.py
```diff
@@ -1089,12 +1291,53 @@ def _print_Assign(self, expr):
         if isinstance(rhs, (PythonRange, Product)):
             return ''
 
-        if isinstance(rhs, (NumpyArray, NumpyLinspace, NumpyDiag, NumpyCross,\
-						NumpyWhere, PyccelArraySize)):
+        if isinstance(rhs, (PythonLen, NumpyRandint)):
+            rhs_code = self._print(rhs)
+            return '{0} = {1}\n'.format(lhs_code, rhs_code)
```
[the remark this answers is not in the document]
EmilyBourne: We should be able to use the code here: https://github.com/pyccel/pyccel/blob/master/pyccel/codegen/printing/fcode.py#L1169-L1177

pyccel/codegen/printing/fcode.py
EmilyBourne: See:
https://github.com/pyccel/pyccel/blob/master/pyccel/codegen/printing/fcode.py#L1169-L1177

pyccel/codegen/printing/fcode.py
mostafaMamoni: yeah it does the same as `_print_PythonImag`, so can we remove it?.

EmilyBourne: When looking for the function to print a `NumpyImag` the `FCodePrinter` will call `_print_NumpyImag` if it exists, and if it doesn't then it will call `_print_PythonImag`

mostafaMamoni: I think i will keep the code readable and make `_print_NumpyImag` call `_print_PythonImag`, what do you think ?.

EmilyBourne: I think this is a bad idea. It goes against the pattern established elsewhere. It leads to redundant code, and if we used your method and wanted to rename a class, this would add a lot more changes. For every other class we only implement the functions that are absolutely necessary in order to reduce the amount of code

pyccel/codegen/printing/fcode.py
```diff
@@ -612,25 +610,161 @@ def _print_PythonImag(self, expr):
...
+        elif rank == 1:
+
+            lhs   = IndexedVariable(lhs)[expr.index, expr.index]
+            rhs   = IndexedVariable(expr.array)[expr.index]
+            body  = [Assign(lhs, rhs)]
+            body  = For(expr.index, PythonRange(PythonLen(expr.array)), body)
+            code  = self._print(body)
+            alloc = 'allocate({0}(0: size({1},1)-1, 0: size({1},1)-1))'.format(lhs, array)
+
+        return alloc + '\n' + code
```
EmilyBourne, line 712: Didn't you delete this class?

pyccel/codegen/printing/fcode.py
```diff
@@ -612,25 +610,161 @@ def _print_PythonImag(self, expr):
...
+
+                code = 'reshape({}, shape({}), order=[2, 1])'.format(cross_product, first)
+            else:
+
+                code = 'reshape({}, shape({})'.format(cross_product, first)
+
+        elif rank == 1:
+            code = cross_product
+
+        return code
```
EmilyBourne, line 688: Didn't you delete this class?

pyccel/codegen/printing/fcode.py
```diff
@@ -1060,6 +1206,11 @@ def _print_SymbolicAssign(self, expr):
         stmt = Comment(str(expr))
         return self._print_Comment(stmt)
 
+    def _print_NumpyReal(self, expr):
+        value = self._print(expr.arg)
+        prec = self._print(expr.precision)
```
EmilyBourne, line 1211: Don't we need ISO_C_BINDING precision?

EmilyBourne (changes requested):

pyccel/codegen/printing/fcode.py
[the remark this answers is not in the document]
mostafaMamoni: should we still need a specialized `_print_NumpyEmpty`after we're working with:
`if isinstance(rhs, NumpyEmpty):
            return ''`
for `tests/errors/codegen/fortran/FORTRAN_ALLOCATABLE_IN_EXPRESSION.py` I removed it ;).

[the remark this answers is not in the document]
EmilyBourne: Probably we do need a specialized `_print_NumpyEmpty` to make sure the error is raised. If you do:
```
a = np.empty(3)*5
```
Then you have:
```
PyccelAssign(Variable('a'), PyccelMul(NumpyEmpty(LiteralInteger(3)), LiteralInteger(5))
```
So you won't reach the if block `if isinstance(rhs, NumpyEmpty): return ''`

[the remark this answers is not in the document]
EmilyBourne: > for `tests/errors/codegen/fortran/FORTRAN_ALLOCATABLE_IN_EXPRESSION.py` I removed it ;).

You can get it back by doing:
```
git checkout master -- tests/errors/codegen/fortran/FORTRAN_ALLOCATABLE_IN_EXPRESSION.py
```

pyccel/codegen/printing/fcode.py
mostafaMamoni: this is not a class, but yes I removed `NumpyDiag` in ast.numpyast, and sorry for `lhs` argument I think its cursed for me.

pyccel/codegen/printing/fcode.py
mostafaMamoni: the same with numpydiag

pyccel/codegen/printing/fcode.py
mostafaMamoni: yes,

pyccel/codegen/printing/fcode.py
EmilyBourne: You are right, it is not a class, but it is a function that is only called on instances of the class `NumpyDiag` so if `NumpyDiag` doesn't exist anymore then this function doesn't need to either

EmilyBourne: The `lhs` being here is the reason why we said to delete this class. There is no way round it. It's not your fault ^^

pyccel/codegen/printing/fcode.py
mostafaMamoni: there some of them in builtins for python classes and their calls in fcode.

EmilyBourne: With grep I see 4 methods. So I think it can be done in this PR

pyccel/codegen/printing/fcode.py
```diff
@@ -604,47 +602,128 @@ def _print_PythonSum(self, expr):
...
+        return rhs_code
 
     def _print_NumpyFloor(self, expr):
         result_code = self._print_MathFloor(expr)
         return 'real({}, {})'.format(result_code, iso_c_binding["real"][8])
 
     # ======================================================================= #
     def _print_PyccelArraySize(self, expr):
-        return expr.fprint(self._print)
+        init_value = self._print(expr.arg)
```
EmilyBourne, line 705: Did you delete the old fprint method?

pyccel/codegen/printing/fcode.py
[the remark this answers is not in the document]
EmilyBourne: This comment is hidden but still valid

EmilyBourne (changes requested):

pyccel/codegen/printing/fcode.py
mostafaMamoni: yes, you can check the new one in fcode.

EmilyBourne: Yes I can see the new method. That is where I left the comment. But the old method (`PyccelArraySize.fprint`) is no longer needed. It is defined in ast/core.py but according to github that file hasn't been modified. Are you sure you pushed that change?

mostafaMamoni: oh, you're right i forgot a fprint in ast.core, it should be removed now.

EmilyBourne (approved):

pyccel/ast/numpyext.py
```diff
@@ -878,27 +562,10 @@ def fill_value(self):
...
-            return ''
-        else:
-            return '\n'.join(stmts) + '\n'
-
 #==============================================================================
-class NumpyEmpty(NumpyFull):
-    """ Represents a call to numpy.empty for code generation.
+class NumpyAutoFill(NumpyFull):
+    """ Abstract class for all classes which inherit from NumpyFull but
+        the fill_value is implicitly specified
```
yguclu, line 568: Why is this class needed?

mostafaMamoni: NumpyZeros, NumpyOnes was inheriting from NumpyEmpty, so in the printing NumpyEmpty could be ether NumpyZeros or NumpyOnes in every isinstance block and this not good the new implementation of printing, this class fix that conflict and prevent repeating the same `__new__` function in NumpyEmpty.

EmilyBourne: >so in the printing NumpyEmpty could be ether NumpyZeros or NumpyOnes in every isinstance block and this not good the new

Put another way, `isinstance` couldn't be used to handle the specific problem of `NumpyEmpty` in an `Assign` as it was also catching the subclasses `NumpyZeros` and `NumpyOnes`

yguclu: Thanks for the explanation. This should be added to the PR's description.

yguclu (changes requested):
Please add to the PR's description the fact that you created a new AST node class, and the reason for doing so.

tests/errors/codegen/fortran/FORTRAN_ALLOCATABLE_IN_EXPRESSION.py
```diff
@@ -1,3 +1,3 @@
 # pylint: disable=missing-function-docstring, missing-module-docstring/
 import numpy as np
-a = np.ones(4)*5
+a = np.empty(4)*5
```
saidctb, line 3: why did you change this example ?

saidctb (approved):

tests/errors/codegen/fortran/FORTRAN_ALLOCATABLE_IN_EXPRESSION.py
EmilyBourne: Now that the allocation has been separated from the printing `ones` is printed as `1` in fortran. So:
```
a = np.ones(4)*5
```
is correctly translated to:
```
allocate(a(0:4))
a = 1*5
```

saidctb: ah okay , perfect

EmilyBourne: As all the values of the allocatable in the expression are the same we don't need a temporary array so there is no problem

yguclu (approved):

PR #586, opened 2020-12-08
Finish cleaning up the class FunctionDef

### LINKED ISSUE #550: Clean up the class FunctionDef

EmilyBourne:
- Add doc-strings to FunctionDef (fixes #550);
- Collect and print function doc strings;
- Remove obsolete constructor arguments `kind` and `hide`;
- Remove `procedure` from `FunctionKind` values in header grammar;
- Add test for function doc strings;
- Add test for private decorator.

### DIFF
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -3149,46 +3149,58 @@ class FunctionDef(Basic):
 
     @property
     def name(self):
+        """ Name of the function """
         return self._name
 
     @property
     def arguments(self):
+        """ List of variables which are the function arguments """
         return self._arguments
 
     @property
     def results(self):
+        """ List of variables which are the function results """
         return self._results
 
     @property
     def body(self):
+        """ CodeBlock containing all the statements in the function """
         return self._body
 
     @property
     def local_vars(self):
+        """ List of variables defined in the function """
         return self._local_vars
 
     @property
     def global_vars(self):
+        """ List of global variables used in the function """
         return self._global_vars
 
     @property
     def cls_name(self):
+        """ String containing the name of the class to which the method belongs.
+        If the function is not a class procedure then this returns None """
         return self._cls_name
 
     @property
     def hide(self):
+        """ False if function is pyccelized, True otherwise """
         return self._hide
 
     @property
     def kind(self):
+        """ Returns function or procedure """
         return self._kind
 
     @property
     def imports(self):
+        """ List of imports in the function """
         return self._imports
 
     @property
     def decorators(self):
+        """ List of decorators applied to the function """
         return self._decorators
 
     @property
@@ -3197,42 +3209,74 @@ class FunctionDef(Basic):
 
     @property
     def templates(self):
+        """ List of templates used to determine the types """
         return self._templates
 
     @property
     def is_recursive(self):
+        """ Returns True if the function is recursive (i.e. calls itself)
+        and False otherwise """
         return self._is_recursive
 
     @property
     def is_pure(self):
+        """ Returns True if the function is marked as pure and False otherwise
+        Pure functions must not have any side effects.
+        In other words this means that the result must be the same no matter 
+        how many times the function is called
+        e.g:
+        >>> a = f()
+        >>> a = f()
+
+        gives the same result as
+        >>> a = f()
+
+        This is notably not true for I/O functions
+        """
         return self._is_pure
 
     @property
     def is_elemental(self):
+        """ returns True if the function is marked as elemental and
+        False otherwise
+        An elemental function is a function with a single scalar operator
+        and a scalar return value which can also be called on an array.
+        When it is called on an array it returns the result of the function
+        called elementwise on the array """
         return self._is_elemental
 
     @property
     def is_private(self):
+        """ True if the function should not be exposed to
+        other modules. This includes the wrapper module and
+        means that the function cannot be used in an import
+        or exposed to python """
         return self._is_private
 
     @property
     def is_header(self):
+        """ True if the function body is not provided
+        False otherwise """
         return self._is_header
 
     @property
     def arguments_inout(self):
+        """ List of variables which are the modifiable function arguments """
         return self._arguments_inout
 
     @property
     def functions(self):
+        """ List of functions within this function """
         return self._functions
 
     @property
     def interfaces(self):
+        """ List of interfaces within this function """
         return self._interfaces
 
     @property
     def doc_string(self):
+        """ The docstring of the function """
         return ""
 
     def set_recursive(self):
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -3287,27 +3287,6 @@ class FunctionDef(Basic):
 
         self._name = newname
 
-    @property
-    def is_procedure(self):
-        """Returns True if a procedure."""
-
-        flag = False
-        if len(self.results) == 1 and isinstance(self.results[0], Expr):
-            vars_ = [i for i in preorder_traversal(self.results)
-                     if isinstance(i, Variable)]
-            flag = flag or any([i.allocatable or i.rank > 0 for i in
-                               vars_])
-        else:
-            flag = flag or len(self.results) == 1 \
-                and self.results[0].allocatable
-            flag = flag or len(self.results) == 1 \
-                and self.results[0].rank > 0
-        flag = flag or len(self.results) > 1
-        flag = flag or len(self.results) == 0
-        flag = flag \
-            or len(set(self.results).intersection(self.arguments)) > 0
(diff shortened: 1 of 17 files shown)


pyccel/ast/core.py
```diff
@@ -3149,46 +3149,58 @@ def __init__(
...
 
     @property
     def hide(self):
+        """ False if function is pyccelized, True otherwise """
         return self._hide
 
     @property
     def kind(self):
+        """ Returns function or procedure """
         return self._kind
```
EmilyBourne, line 3194: @saidctb Is this property useful? It returns either 'procedure' or 'function'. So I am not sure what it does. In fortran a 'procedure' is either a `function` or a `subroutine`. In pyccel it seems that `kind` is always 'function'. It is used in the function `is_procedure`. As far as I can tell `is_procedure` returns True if the FunctionDef is a `subroutine` (in any case this is how it is used in fcode)

pyccel/ast/core.py
```diff
@@ -3149,90 +3149,135 @@ def __init__(
...
 
     @property
     def decorators(self):
+        """ List of decorators applied to the function """
         return self._decorators
 
     @property
     def headers(self):
+        """ List of headers applied to the function """
         return self._headers
```
EmilyBourne, line 3209: @saidctb  I seem to remember you saying several times that the decorators are translated to headers. Does the `FunctionDef` class need both of these methods? If so is there a more fundamental difference in their contents that I am missing (e.g. `decorators` contains user defined decorators?)

pyccel/ast/core.py
```diff
@@ -3149,90 +3149,135 @@ def __init__(
...
 
     @property
     def templates(self):
+        """ List of templates used to determine the types """
         return self._templates
```
EmilyBourne, line 3214: @rmahjoubi Is this docstring correct? Why do we need `templates` in a `FunctionDef`? It would make sense in an `Interface` but I don't understand the relevance here

pyccel/ast/core.py
```diff
@@ -3149,90 +3149,135 @@ def __init__(
...
+        called elementwise on the array """
         return self._is_elemental
 
     @property
     def is_private(self):
+        """ True if the function should not be exposed to
+        other modules. This includes the wrapper module and
+        means that the function cannot be used in an import
+        or exposed to python """
         return self._is_private
```
EmilyBourne, line 3255: @saidctb @yguclu Does this method make sense when translating from python? It seems to add the `private :: ` suffix to the object in fortran. I remember having several discussions about this but I don't remember what the conclusion was

pyccel/ast/core.py
```diff
@@ -3149,90 +3149,135 @@ def __init__(
...
     @property
     def cls_name(self):
+        """ String containing the name of the class to which the method belongs.
+        If the function is not a class procedure then this returns None """
         return self._cls_name
 
     @property
     def hide(self):
+        """ False if function is pyccelized, True otherwise """
         return self._hide
```
EmilyBourne, line 3189: @saidctb Is this description correct? If so then why do we need a `FunctionDef` object for this function?

pyccel/ast/core.py
```diff
@@ -3149,90 +3149,135 @@ def __init__(
...
+        other modules. This includes the wrapper module and
+        means that the function cannot be used in an import
+        or exposed to python """
         return self._is_private
 
     @property
     def is_header(self):
+        """ True if the function body is not provided
+        False otherwise """
         return self._is_header
```
EmilyBourne, line 3261: @saidctb Is this description correct? I suspect it is not detailed enough?

pyccel/ast/core.py
saidctb: it's not useful

pyccel/ast/core.py
saidctb: yes both

pyccel/ast/core.py
saidctb: True if ti was created from a header, by calling  `create_definition` method of the `HeaderFunctin` class

pyccel/ast/core.py
EmilyBourne: Ah I see the problem! It definitely feels odd to me but I suppose it makes sense. A `FunctionDef` represents a python function before the semantic stage so it doesn't need to have a fixed type and be split into `Interface`s yet.

If we think of it in this way does it make more sense to access the templates via the decorators and headers rather than having it be an integral part of what a `FunctionDef` is?

pyccel/ast/core.py
EmilyBourne:
````
""" True if the implementation of the function body is not provided
        False otherwise """
````
?

EmilyBourne: Are there headers without functions which don't call `create_definition`?

pyccel/ast/core.py
rmahjoubi: That's definitely better, so an issue for this ?

EmilyBourne: Yes please :)

pyccel/ast/core.py
saidctb: we use it when we want to inline a function

pyccel/ast/core.py
saidctb: probably it will be used when we introduce classes

pyccel/ast/core.py
saidctb: yes you can define a header which is never used,
this we need when we have a header file and you import from it, the imported function will be a FunctionDef and not a FunctionHeader

pyccel/ast/core.py
EmilyBourne: Is there a way of doing that currently in pyccel?

pyccel/ast/core.py
EmilyBourne: So then it is not used currently? That sounds to me like it can be removed for now?

pyccel/ast/core.py
EmilyBourne: Ok. Does my new docstring suggestion sound right to you then?

pyccel/ast/core.py
saidctb: yes you can remove it

pyccel/ast/core.py
saidctb: we had it, but it needs to be updated, it's used when the user add `@inline` decorator

pyccel/ast/core.py
saidctb: it's fine

pyccel/ast/core.py
EmilyBourne: Ah, it seems this is linked to a `@private` decorator. So maybe it is useful? @yguclu  do you have an opinion?

pyccel/ast/core.py
EmilyBourne: But we don't have an `@inline` decorator

saidctb (approved):

pyccel/ast/core.py
yguclu: This may be useful if a Python function can be translated to Fortran, but cannot be exposed back to Python through our wrapper... Do we actually process the `private` decorator at all?

EmilyBourne: > This may be useful if a Python function can be translated to Fortran, but cannot be exposed back to Python through our wrapper

Maybe, but that isn't something that the user chooses. Currently we have functions as arguments which cannot be exposed to python. In the cwrapper these function prototypes are simply not printed and the wrapper just raises a NotImplementedError [here](https://github.com/pyccel/pyccel/blob/master/pyccel/codegen/printing/cwrappercode.py#L796)

>Do we actually process the private decorator at all?

It seems that we do. If I add a private decorator then the fortran code generated is correct, but pyccel then tries to pass it to the wrapper anyway

yguclu: OK. If we want to keep this decorator then we should make sure that it is not exposed to Python. If this amounts to adding an `if` clause in the right place, feel free to fix this in this PR. Please let me know what you find out.

EmilyBourne: 2 `if`s

EmilyBourne: I have also added a very basic test

yguclu (approved):
Excellent job!

Issue #580, opened 2020-12-02
allow_negative_index decorator does not work for arguments


EmilyBourne:
**Describe the bug**
The `allow_negative_index` decorator does not work for arguments

**To Reproduce**
Provide code to reproduce the behavior:
```python
@allow_negative_index('a')
@types('int[:]','int')
def f(a,i):
    return a[i]
```

Provide the generated code, or the error message:
```fortran
module boo

use ISO_C_BINDING

implicit none

contains

!........................................
function f(a, i) result(Out_0001)

  implicit none

  integer(C_LONG_LONG) :: Out_0001
  integer(C_LONG_LONG), intent(in) :: a(0:)
  integer(C_LONG_LONG), value :: i

  Out_0001 = a(i)
  return

end function f
!........................................

end module boo
```

**Expected behavior**
```
module boo

use ISO_C_BINDING

implicit none

contains

!........................................
function f(a, i) result(Out_0001)

  implicit none

  integer(C_LONG_LONG) :: Out_0001
  integer(C_LONG_LONG), intent(in) :: a(0:)
  integer(C_LONG_LONG), value :: i

  Out_0001 = a(MODULO(i,size(a, 1)))
  return

end function f
!........................................

end module boo
```

Issue #575, opened 2020-12-01
Conversion of negative indexes in slices


EmilyBourne:
**Describe the bug**
Related to issue #248. Although negative indexes are correctly handled for arrays. Negative indexes in slices are not handled

**To Reproduce**
Provide code to reproduce the behavior:
```python
import numpy as np

a = np.zeros(10)
print(a[-5])
print(a[:-5])
```

Provide the generated code, or the error message:
```bash/fortran/c
program prog_tmp

use ISO_C_BINDING

implicit none

real(C_DOUBLE), allocatable :: a(:)

allocate(a(0:10_C_LONG_LONG - 1_C_LONG_LONG))
a = 0.0_C_DOUBLE
print *, a(10_C_LONG_LONG - 5_C_LONG_LONG)
print *, a(:(-5_C_LONG_LONG) - 1_C_LONG_LONG)

end program prog_tmp
```

**Expected behavior**
The code should be;
```
program prog_tmp

use ISO_C_BINDING

implicit none

real(C_DOUBLE), allocatable :: a(:)

allocate(a(0:10_C_LONG_LONG - 1_C_LONG_LONG))
a = 0.0_C_DOUBLE
print *, a(10_C_LONG_LONG - 5_C_LONG_LONG)
print *, a(:10_C_LONG_LONG - 5_C_LONG_LONG - 1_C_LONG_LONG)

end program prog_tmp
```

**Acceptance Criteria**
(Similar to #248 )
- [x] Preserve the negative value to the codegen stage so the PythonPrinter correctly prints the original code with the negative index in the slice
- [x] Handle negative constants in slices by subtracting them from the shape
- [x] Add modulo operator when the index is not a constant and the user has added the `allow_negative_index` decorator for this variable
- [ ] Ensure that the shape of the variable is correctly calculated

EmilyBourne:
Reopening as the last point isn't handled. If we look at this example:
```
import numpy as np
from pyccel.decorators import types

@types('float[:]')
def test(p):
    q = p[1:-1] + p[2:]
    return np.sum(q)
```
pyccel generates the error:
```
[...]

Issue #543, opened 2020-11-13
Slices should have a step


EmilyBourne:
**Describe the bug**
Steps in slices are not yet implemented however they are ready to use in c and could be added reasonably easily in fortran

**Test Code**
Provide code which does not currently work but which should do when this issue is fixed:
```python
import numpy as np

x = np.ones(10)
y = x[::2]
```

yguclu:
This is a surprise to me! A Fortran slice is almost identical to a Numpy slice, so it should not be a problem to print the step.

Issue #564, opened 2020-11-26
Allow simplification of pyccel-created nodes


EmilyBourne:
# Description
As a USER I want the generated code to be as clear as possible while also remaining as close as possible to my original code.

Currently nodes are created in pyccel in one of two cases:
1. As a direct translation of a phrase written in python
2. As a necessary additional step in order for the code to reproduce the expected behaviour

In the first case we want the code to remain similar to the original code. Thus:
```python
a = b + -c
```
should be faithfully translated to:
```fortran
a = b + (-c)
```
this is already the case.

However in the second case we want the intention to be obvious. The form of the result is less important as the code must be changed in any case. Neater code is therefore more preferable. This is demonstrated notably in the following use cases:

## Use case A:
Python code:
```python
import numpy as np
a = np.random.randint(-10,10)
```
Currently generated fortran code:
```fortran
program prog_test

use ISO_C_BINDING

implicit none

integer(C_LONG_LONG) :: a
real(C_DOUBLE) :: Dummy_0001

call random_number(Dummy_0001)

a = floor((10_C_LONG_LONG - (-10_C_LONG_LONG)) * Dummy_0001 + ( &
      -10_C_LONG_LONG), kind=8)

end program prog_test
```
Desired fortran code:
```fortran
program prog_test

use ISO_C_BINDING

implicit none

integer(C_LONG_LONG) :: a
real(C_DOUBLE) :: Dummy_0001

call random_number(Dummy_0001)

a = floor((10_C_LONG_LONG +10_C_LONG_LONG) * Dummy_0001 - 10_C_LONG_LONG), kind=8)

end program prog_test
```

## Use case B:
Python code:
```python
import numpy as np
a = np.empty(10)
```
Currently generated fortran code:
```fortran
program prog_test

use ISO_C_BINDING

implicit none

real(C_DOUBLE), allocatable :: a(:)

allocate(a(0:10_C_LONG_LONG - 1_C_LONG_LONG))

end program prog_test
```
Desired fortran code:
```fortran
program prog_test

use ISO_C_BINDING

implicit none

real(C_DOUBLE), allocatable :: a(:)

allocate(a(0:9_C_LONG_LONG))

end program prog_test
```

The steps required to achieve this are detailed in the acceptance criteria

**Acceptance Criteria**
[...]

PR #588, opened 2020-12-08
Add Slice management

MenoIy:
Solve issue #575:
- Create function slice_management to handle slice elements logic (start, end, step);
- Convert negative indexes in slices elements using the same logic in IndexedElement;
- Add tests for Slices for 1d and 2d arrays, for C and F ordering.

Solve issue #543:
- Collect all slice needed information from the parser;
- Add step arguments to the slice class, with a default value of None;
- Use PyArraySize to calculate sliced element shape;
- Add type check for Slice elements;
- Create function slice_management to manage slice property before printing;
- Add tests for slicing arrays.

Solve issue #580:
- Add allows_negative_indexes=True to arguments data dictionary in the parser;
- Add tests.


pyccel/ast/numpyext.py
```diff
@@ -362,8 +362,10 @@ def Shape(arg):
     if isinstance(arg.shape, PythonTuple):
         return arg.shape
     else:
-        return PythonTuple(*arg.shape)
-
+        if arg.order == 'F':
+            return PythonTuple(*arg.shape)
+        else :
+            return PythonTuple(*arg.shape[::-1])
```
EmilyBourne, line 368: Are you sure about this?

pyccel/parser/semantic.py
```diff
@@ -835,6 +835,9 @@ def _visit_Slice(self, expr, **settings):
 
         if args[1] is not None:
             args[1] = self._visit(args[1], **settings)
+
+        if args[2] is not None:
```
EmilyBourne, line 839: This can also be written as `if args[2]`

MenoIy:
@saniQQ  can you take a look at the slice_management in the c code printer. I have done all the calculations for the start and the end so there is no None or negative value. for positive and None step tests in C are working fine.

MenoIy:
@yguclu  yes it's a bit slow 8x second for all tests. I was thinking of grouping each tests group in one function so I can declare and fill the array used for the tests one time that will save some time on the python side.

pyccel/ast/core.py
```diff
@@ -4941,32 +4941,51 @@ class Slice(Basic):
...
-    >>> m, n = symbols('m, n', integer=True)
+    >>> m, n, p = symbols('m, n, p', integer=True)
     >>> Slice(m,n)
     m : n
     >>> Slice(None,n)
      : n
     >>> Slice(m,None)
     m :
+    >>> Slice(m, n, p)
+    m : n : p
```
rakati, line 4958: can update the variables names in the doc-string to be more comprehensive.
suggest instead of `m, n, p` use `start, end , step.`

pyccel/ast/core.py
```diff
@@ -4852,12 +4852,12 @@ def __init__(
             for a,s in zip(args, shape):
                 if isinstance(a, Slice):
```
rakati, line 4853: I know this is beyond your PR, but you already did changes in this file so can you update the doc-string of the `IndexedElement` class, the bug is already fixed, and the example does not work well.
so I suggest this example.
```python
"""
    Represents a mathematical object with indices.
    Examples
    --------
    >>> from sympy import symbols, Idx
    >>> from pyccel.ast.core import IndexedVariable, IndexedElement
    >>> i, j = symbols('i j', cls=Idx)
    >>> A = IndexedVariable('A', dtype='int')
    >>> IndexedElement(A, i, j)
    IndexedElement(A, i, j)
    >>> IndexedElement(A, i, j) == A[i, j]
    True
    """
```

MenoIy: Sorry I didn't change anything in this class. I won't remove a TODO or a bug comment that I didn't do or fixed and I don't have a slice idea what is about.

EmilyBourne: I think I fixed that bug a while ago. I didn't notice that it was in the docstring. Sorry

I agree that it has nothing to do with this PR, but it's a quick copy and paste and I'm not sure it's worth creating a separate PR just for this. Better to fix it while it's been spotted. I am happy to vouch for the fact that it is fixed.

pyccel/parser/semantic.py
```diff
@@ -2414,7 +2417,9 @@ def _visit_FunctionDef(self, expr, **settings):
                         dtype = d_var.pop('datatype')
                         if d_var['rank']>0:
                             d_var['cls_base'] = NumpyArrayClass
-
```
saniQQ, line 2417: missing whitespace

pyccel/ast/core.py
```diff
@@ -4858,6 +4852,7 @@ def __init__(
                         new_shape.append(end)
                     else:
                         new_shape.append(PyccelMinus(end, start))
+
```
saniQQ, line 4855: unnecessary whitespace change

saniQQ (changes requested):

tests/epyccel/test_arrays.py
```diff
@@ -1495,6 +1511,1171 @@ def test_array_variable_size():
...
+    f1 = arrays.array_1d_slice_7
+    f2 = epyccel(f1)
+
+    assert np.array_equal(f1(a), f2(a))
+
+def test_array_1d_slice_8():
+    a = np.empty(21, dtype=np.int)
+    for i in range(21):
+        a[i] = 2**i
+    f1 = arrays.array_1d_slice_8
```
rakati, line 1585: small improve : `a[i] = 2**i` to `a[i] = 1 << i`

saniQQ (approved):

rakati (commented):
sorry, wrong message.

pyccel/codegen/printing/ccode.py
```diff
@@ -665,31 +665,52 @@ def _print_IndexedElement(self, expr):
...
+            # variable step in slice
+            elif allow_negative_index and not isinstance(step, LiteralInteger):
+                start = IfTernaryOperator(PyccelGt(step, LiteralInteger(0)), start, end)
+                end = IfTernaryOperator(PyccelGt(step, LiteralInteger(0)), end, start)
+        else :
+            step = LiteralInteger(1)
+        return Slice(start, end, step)
+
+    def _print_PyccelArraySize(self, expr):
+        return '{}.shape[{}]'.format(expr.arg, expr.index)
```
EmilyBourne, line 712: Do we need to handle the ordering here? @saniQQ ?

tests/epyccel/test_arrays.py
```diff
@@ -1495,6 +1511,1171 @@ def test_array_variable_size():
...
+
+#==============================================================================
+# TEST : 2d array slices order F
+#==============================================================================
+
+def test_array_2d_F_slice_1():
+    a = np.empty((21,21), dtype=np.int, order = 'F')
+    for i in range(21):
+        for j in range(21):
+            a[i][j] = 2**j
```
yguclu: The standard way to index an N-dimensional Numpy array is with N indices, so you should write `a[i, j] = 2**j`. Here you have instead:
- `a`: 2D array
- `a[i]`: 1D slicing of `a`
- `a[i][j]`: get element `j` of `a[i]`

which is definitely slower.

Anyway, I suspect that the fastest way to fill this 2D array is the following:
```python
with np.nditer(a, op_flags=['readwrite']) as it:
    for x in it:
        x[...] = 2**it.multi_index[1]
```

But without going into the Numpy craziness above, I think we could simply make `a` a module variable given that it is never modified by the various functions. So we could define, for example:
```python
a_1d   = ...
a_2d_c = ...
a_2d_f = ...
```

Please have a look at these approaches and time them, and let me know what you find out.

pyccel/codegen/printing/fcode.py
```diff
@@ -2608,31 +2608,77 @@ def _print_IndexedElement(self, expr):
...
+            elif allow_negative_index and not isinstance(step, LiteralInteger):
+                if start is None :
+                    start = IfTernaryOperator(PyccelGt(step, LiteralInteger(0)),
+                        LiteralInteger(0), PyccelMinus(shape, LiteralInteger(1)))
+
+                if end is None :
+                    end = IfTernaryOperator(PyccelGt(step, LiteralInteger(0)),
+                        PyccelMinus(shape, LiteralInteger(1)), LiteralInteger(0))
+                else :
+                    end = PyccelAdd(end, LiteralInteger(1))
```
EmilyBourne, line 2675: I am confused. We have:
```
0:10
```
translated to:
```
0:11
```
but 
```
0:
```
translated to:
```
0:shape-1
```
?

pyccel/codegen/printing/fcode.py
```diff
@@ -2608,31 +2608,77 @@ def _print_IndexedElement(self, expr):
...
+
+        if end is not None and end == tmp_end:
+            end = PyccelMinus(end, LiteralInteger(1))
```
EmilyBourne, line 2678: This doesn't seem like the clearest way of doing this to me. Can you explain what is happening and why you need this `if` block?
Do you not get: `end+1-1` in the end? for variable steps in slice?
Can you not do:
`if end is not None:`
before the block above and then just not touch cases that are already correct?

pyccel/codegen/printing/fcode.py
```diff
@@ -2608,31 +2608,77 @@ def _print_IndexedElement(self, expr):
...
         inds = [self._print(i) for i in inds]
 
-        return "%s(%s)" % (base, ", ".join(inds))
+        return "%s(%s)" % (base_code, ", ".join(inds))
 
 
     def _print_Idx(self, expr):
         return self._print(expr.label)
 
+    def _slice_management(self, _slice, shape, allow_negative_index):
```
EmilyBourne, line 2639: Is there a difference between this method and the one in the ccode? If not then I wonder if it would make sense to make it a member of the slice class. For example we could do:
`slice = slice.remove_negative_indexes(shape)` for example?

What do you think @saidctb @yguclu ?

EmilyBourne: Or `slice.start_w_pos_indexes`, `slice.step_w_pos_indexes`, and `slice.end_w_pos_indexes` ?

pyccel/parser/semantic.py
```diff
@@ -830,11 +830,14 @@ def _visit_str(self, expr, **settings):
...
-        if args[0] is not None:
+        if args[0]:
             args[0] = self._visit(args[0], **settings)
 
-        if args[1] is not None:
+        if args[1]:
             args[1] = self._visit(args[1], **settings)
+
+        if args[2]:
+            args[2] = self._visit(args[2], **settings)
```
EmilyBourne, line 840: After your changes in the `Slice` class you should be able to replace:
`args[0]` -> `start`
`args[1]` -> `end`
`args[2]` -> `step`

tests/epyccel/test_arrays.py
```diff
@@ -1476,6 +1476,22 @@ def test_multiple_negative_index_3():
 
     assert np.array_equal(f1(-1, -1, -3), f2(-1, -1, -3))
 
+def test_argument_negative_index_1():
+    a = np.empty(21, dtype=np.int)
+    for i in range(21):
+        a[i] = 1 << i
```
EmilyBourne, line 1482: You should be able to make this faster by using a list comprehension:
`a = np.array([1 << i for i in range(21)], dtype=np.int)`

tests/epyccel/test_arrays.py
```diff
@@ -1476,6 +1476,22 @@ def test_multiple_negative_index_3():
...
+    f1 = arrays.test_argument_negative_index_1
+    f2 = epyccel(f1)
+    assert np.array_equal(f1(a), f2(a))
+
+def test_argument_negative_index_2():
+    a = np.empty(21, dtype=np.int)
+    for i in range(21):
+        a[i] = 1 << i
```
EmilyBourne, line 1490: See above list comprehension

tests/epyccel/test_arrays.py
```diff
@@ -1495,6 +1511,1171 @@ def test_array_variable_size():
...
     assert s1 == s2
 
+#==============================================================================
+# TEST : 1d array slices
+#==============================================================================
+
+def test_array_1d_slice_1():
+    a = np.empty(21, dtype=np.int)
+    for i in range(21):
+        a[i] = 1 << i
```
EmilyBourne, line 1521: See above list comprehension

tests/epyccel/test_arrays.py
```diff
@@ -1495,6 +1511,1171 @@ def test_array_variable_size():
...
+    f1 = arrays.array_1d_slice_1
+    f2 = epyccel(f1)
+
+    assert np.array_equal(f1(a), f2(a))
+
+def test_array_1d_slice_2():
+    a = np.empty(21, dtype=np.int)
+    for i in range(21):
+        a[i] = 1 << i
```
EmilyBourne, line 1530: See above list comprehension. There are many more unnecessary for loops in this file

pyccel/parser/semantic.py
yguclu, line 840: Can't this be written as the following?
```python
args = [(self._visit(a, **settings) if a else None) for a in expr.args]
```

yguclu (changes requested):

tests/epyccel/test_arrays.py
```diff
@@ -1495,6 +1511,1171 @@ def test_array_variable_size():
...
+
+#==============================================================================
+# TEST : 2d array slices stride order F
+#==============================================================================
+
+def test_array_2d_F_slice_stride_1():
+    a = np.empty((21,21), dtype=np.int, order='F')
+    for i in range(21):
+        for j in range(21):
+            a[i][j] = 1 << j
```
EmilyBourne, line 2262: It should also be possible to create the 2d cases with list comprehensions

pyccel/ast/core.py
```diff
@@ -4941,32 +4935,51 @@ class Slice(Basic):
...
 
-    # TODO add step
-    # TODO check that args are integers
-    # TODO add negative indices
-    def __new__(cls, start, end):
-        return Basic.__new__(cls, start, end)
+    def __new__(cls, start, end, step = None):
+        return Basic.__new__(cls, start, end, step)
+
+    def __init__(self, start, end, step = None):
```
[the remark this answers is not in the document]
yguclu: @EmilyBourne I think the situation is a bit more complicated:
```ipython
$ ipython
Python 3.6.9 (default, Oct  8 2020, 12:12:24) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.2.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: slice?                                                                  
Init signature: slice(self, /, *args, **kwargs)
Docstring:     
slice(stop)
slice(start, stop[, step])

Create a slice object.  This is used for extended slicing (e.g. a[0:10:2]).
Type:           type
Subclasses
```

pyccel/codegen/printing/fcode.py
```diff
@@ -2608,31 +2608,77 @@ def _print_IndexedElement(self, expr):
...
+            end = IfTernaryOperator(PyccelLt(end, LiteralInteger(0)),
+                        PyccelAdd(shape, end), end)
+
+        tmp_end = end #save temporary value of end
+        # steps in slices
+        if step is not None :
+            #negative step in slice
+            if isinstance(step, PyccelUnarySub) and isinstance(step.args[0], LiteralInteger):
+                end = PyccelAdd(end, LiteralInteger(1)) if end is not None else LiteralInteger(0)
+                start = start if start is not None else PyccelMinus(shape, LiteralInteger(1))
```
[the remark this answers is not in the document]
MenoIy: a = s[ : ]  here both start and end are None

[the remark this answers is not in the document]
yguclu: The arguments `start`, `stop` and `step` to the `slice` function can all be `None`, actually:
```ipython
$ ipython
Python 3.6.9 (default, Oct  8 2020, 12:12:24) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.2.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: slice(None)                                                             
Out[1]: slice(None, None, None)
```

pyccel/ast/core.py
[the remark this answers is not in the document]
MenoIy: I just kept the old change because we are always giving start and end a value or None in the parsing stage.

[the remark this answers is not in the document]
EmilyBourne: Actually I'm not sure we handle the slice function. Possibly we only handle the syntax with a colon. In that case it makes sense for `end` to be a compulsory argument.

pyccel/codegen/printing/ccode.py
```diff
@@ -665,31 +665,52 @@ def _print_IndexedElement(self, expr):
...
+            if isinstance(step, PyccelUnarySub) and isinstance(step.args[0], LiteralInteger):
+                start = shape if _slice.start is None else start
+                end = LiteralInteger(0) if _slice.end is None else end
+
+            # variable step in slice
+            elif allow_negative_index and not isinstance(step, LiteralInteger):
+                start = IfTernaryOperator(PyccelGt(step, LiteralInteger(0)), start, end)
+                end = IfTernaryOperator(PyccelGt(step, LiteralInteger(0)), end, start)
+        else :
+            step = LiteralInteger(1)
```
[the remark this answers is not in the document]
MenoIy: it was ... before I add the management of step with decorator allow_negative_index. I'm gonna remove it .

pyccel/codegen/printing/fcode.py
MenoIy: I have done this just to avoid writing end + 1 - 1

pyccel/codegen/printing/ccode.py
saniQQ: the ordering does not affect the shape of an array

pyccel/ast/core.py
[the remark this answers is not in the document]
yguclu: For what is worth, we could just ask for all three arguments (`start`, `stop` and `end`) to be passed to the constructor. And each of them could be `None` of course.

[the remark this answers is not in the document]
MenoIy: should I change the end to stop only in class or everywhere we use the slice?

pyccel/codegen/printing/fcode.py
MenoIy: slice management in fcode is different than the one in ccode. the management of slice in c code is done like the one in python. but Fortran had some differences.

pyccel/codegen/printing/fcode.py
MenoIy: this is when giving 
```
a = b[ : : step]
```
- first point .
here Fortran manage step differently than python when end and start are none
writing :  
```
a = b[ : : step] # if step > 0 this will work
a = b[ : : step] # if step < 0 this will not  work so we need to prisise the start and the end for fortran
```
- second point.
if we take a look at python slice it looks like this [ start: end [  end is excluded but in Fortran [start, end] end is included so when we have  
```
a = b [ : 2: -2] to exulde 2 I add 1.
```

pyccel/ast/core.py
```diff
@@ -4941,32 +4935,51 @@ class Slice(Basic):
...
+        if start is not None and not (isinstance(start, (Symbol, PyccelOperator)) or
+                (hasattr(start, 'dtype') and isinstance(start.dtype, NativeInteger))):
```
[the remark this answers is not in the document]
MenoIy: I'm not getting the data type of PyccelUnarySub.

pyccel/codegen/printing/fcode.py
EmilyBourne: I am happy with the first case, I am just confused about the second case.

Here we are handling a case such as `b[0:2:i]`. If `i=1` then python will create an array containing `b[0]`, and `b[1]` but with the same syntax fortran will create an array containing `b[0]`, `b[1]` and `b[2]` so we need to print `b[0:1:i]` to get the same result.

As far as I can tell this case would fall to line 2675 so we would print `b[0:3:i]`. Should this not be a ternary operator?

I think this makes sense for `b[4:2:i]` with `i=-1`,  then python will create an array containing `b[3]`, and `b[4]` but with the same syntax fortran will create an array containing `b[2]`, `b[3]` and `b[4]` so we need to print `b[4:3:i]` to get the same result.

pyccel/ast/core.py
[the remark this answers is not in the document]
EmilyBourne: Why not?

pyccel/codegen/printing/fcode.py
EmilyBourne: Is there no way to make this an `else` instead of checking if `end == tmp_end` to see if it is modified?

EmilyBourne: Ah, so it needs to be done in this order until issue #564 is fixed?

pyccel/ast/core.py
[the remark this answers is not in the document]
MenoIy:
```
hasattr(start, 'dtype') = True
start.dtype  = None
```

pyccel/codegen/printing/fcode.py
MenoIy: seem's logic. I'm gonna remove it once I remember why I put it in the first place.

pyccel/ast/core.py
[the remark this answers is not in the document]
EmilyBourne: So there is a bug in `PyccelUnarySub`?

I just did:
```
In [11]: from pyccel.ast.core import  Variable

In [13]: a = Variable('int','a')

In [14]: a
Out[14]: a

In [15]: a.shape
Out[15]: ()

In [17]: from pyccel.ast.operators import PyccelUnarySub

In [18]: b = PyccelUnarySub(a)

In [19]: b.dtype
Out[19]: NativeInteger()
```
and it seems to work. So it shouldn't be a bug in `PyccelUnarySub`. Is it possible that you have forgotten to `_visit` something?

pyccel/parser/semantic.py
yguclu: @EmilyBourne It's hard to follow the review process if my comment gets resolved without me knowing how and why. I think the developer should reply to my comment (e.g. "done in commit ...") and then the reviewer should close the conversation...

pyccel/ast/core.py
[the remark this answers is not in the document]
yguclu: I would like it everywhere a `slice` object is used. It's a bit confusing to use `end` in my opinion, given that Python calls it `stop`.

pyccel/ast/core.py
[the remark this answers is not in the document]
EmilyBourne: Never mind, I just realised that this is `ast/core.py` so you will also be checking this in the syntactic stage. In this file you will find lots of lines doing:
```
if self.stage == 'syntactic':
            return
```
This is because the `dtype` isn't known at the syntactic stage. This seems like a better approach as your approach won't work in other unexpected circumstances. Eg.:
```
import numpy as np

def get_step():
    return 1

a = np.ones(10)
b = a[4:7:get_step()]
```

pyccel/parser/semantic.py
EmilyBourne: Sorry

pyccel/parser/semantic.py
```diff
@@ -829,16 +829,11 @@ def _visit_str(self, expr, **settings):
         return repr(expr)
 
     def _visit_Slice(self, expr, **settings):
-        args = list(expr.args)
-        if args[0]:
-            args[0] = self._visit(args[0], **settings)
+        start = self._visit(expr.start) if expr.start else expr.start
+        end = self._visit(expr.end) if expr.end else expr.end
+        step = self._visit(expr.step) if expr.step else expr.step
```
yguclu, line 834: Please just end these lines with `else None`, which I think is much more explicit and clearer to read.

pyccel/codegen/printing/ccode.py
```diff
@@ -666,31 +666,53 @@ def _print_IndexedElement(self, expr):
...
+        if isinstance(step, PyccelUnarySub) and isinstance(step.args[0], LiteralInteger):
+            start = shape if _slice.start is None else start
+            end = LiteralInteger(0) if _slice.end is None else end
+
+        # variable step in slice
+        elif allow_negative_index and step and not isinstance(step, LiteralInteger):
+            start = IfTernaryOperator(PyccelGt(step, LiteralInteger(0)), start, end)
+            end = IfTernaryOperator(PyccelGt(step, LiteralInteger(0)), end, start)
+        else :
+            step = LiteralInteger(1)
```
EmilyBourne, line 709: Am I missing something or does this miss the case of a non-negative step (e.g. `2:4:2`, or `:5:i` without `@allow_negative_index`)?

pyccel/codegen/printing/fcode.py
EmilyBourne: Is there no way to make this an else instead of checking if end == tmp_end to see if it is modified?

pyccel/parser/semantic.py
yguclu: By the way, can we be sure that `expr.start` and `expr.end` are never equal to `0` (of type `int`)?
I'm just worried that we would get `None` in that case...

pyccel/codegen/printing/ccode.py
MenoIy: In C as @saniQQ  said we will always check the step value if it's negative or positive either there is a decorator or non.

EmilyBourne: That is not my point. Here you have 3 cases:
1. **Step is a negative literal**
    Here you keep the step
2. **Step is provided and the `@allow_negative_index` decorator is set**
    Here you keep the step
3. **Everything else** (e.g. `step=2`)
    Here you set the step to 1

tests/epyccel/test_arrays.py
MenoIy: done. generating an array in a module and use it for all tests is faster.

pyccel/codegen/printing/ccode.py
EmilyBourne: I suspect your `else` should be: `elif not step:`

EmilyBourne: Compare this to the original code here:
https://github.com/pyccel/pyccel/pull/588#discussion_r542381620

MenoIy: ... that's true, I changed the condition and I didn't pay attention here it should be 
```
elif step is None : 
       step = 1
```
or like you have mentioned above.

pyccel/codegen/printing/fcode.py
```diff
@@ -923,16 +923,16 @@ def _print_MacroCount(self, expr):
...
-                    elif i.end is None:
+                        if (isinstance(i.stop, (int, LiteralInteger)) and i.stop>0) or not(isinstance(i.stop, (int, LiteralInteger))):
+                            shape.append(i.stop)
+                    elif i.stop is None:
                         if (isinstance(i.start, (int, LiteralInteger)) and i.start<s-1) or not(isinstance(i.start, (int, LiteralInteger))):
-                            shape.append(s-i.start)
+                            shape.append(s - i.start)
                     else:
-                        shape.append(i.end-i.start+1)
+                        shape.append(i.stop - i.start + 1)
```
EmilyBourne, line 935: This code isn't tested (see https://app.codacy.com/gh/pyccel/pyccel/file/51246314423/coverage?bid=15415567&fileBranchId=15415567#l901)
and I'm not sure if it's used so it's not a huge problem, but this code is definitely wrong. Would you mind putting in the missing `PyccelMinus`, `PyccelAdd`, and `LiteralInteger` calls?

pyccel/codegen/printing/fcode.py
EmilyBourne: Eg.
```
if isinstance(step, PyccelUnarySub) and isinstance(step.args[0], LiteralInteger):
    # negative step in slice
    ...
elif step is None or not allow_negative_index:
    # positive step in slice
    stop = PyccelMinus(stop, LiteralInteger(1))
else:
    # general step in slice
    ...
```
?

MenoIy: yes, changing order might work.

pyccel/codegen/printing/fcode.py
EmilyBourne: Thanks :+1:

EmilyBourne (approved):

pyccel/codegen/printing/ccode.py
```diff
@@ -666,31 +666,54 @@ def _print_IndexedElement(self, expr):
...
+                        inds[i] = self._slice_elements_managements(ind, PyccelArraySize(base, i),
+                            allow_negative_indexes)
                 inds = [self._print(i) for i in inds]
                 return "array_slicing(%s, %s)" % (base_name, ", ".join(inds))
             inds = [self._print(i) for i in inds]
         else:
             raise NotImplementedError(expr)
         return "%s.%s[get_index(%s, %s)]" % (base_name, dtype, base_name, ", ".join(inds))
 
+    def _slice_elements_managements(self, _slice, shape, allow_negative_index):
```
yguclu, line 678: I have a few comments about this function:

1. Since the `self` argument is not used, we should make this a `@staticmethod`
2. Even though this is a private method, it is important enough to deserve a **docstring**
3. The function's name is not sufficiently clear. Given that a new `Slice` object is created with explicit values of `(start, stop, step)`, maybe something like `_new_slice_with_processed_arguments`?
4. The `shape` argument does not represent the proper shape of the array (which is a tuple), but only the number of elements along the given index; so we could call it `array_size` to avoid confusion.

pyccel/codegen/printing/fcode.py
```diff
@@ -2628,23 +2631,65 @@ def _print_IndexedElement(self, expr):
...
         inds = [self._print(i) for i in inds]
 
-        return "%s(%s)" % (base, ", ".join(inds))
+        return "%s(%s)" % (base_code, ", ".join(inds))
 
 
     def _print_Idx(self, expr):
         return self._print(expr.label)
 
+    def _slice_management(self, _slice, shape, allow_negative_index):
```
yguclu, line 2640: I have a similar comment as for the C code printer.

yguclu (changes requested):

pyccel/codegen/printing/ccode.py
MenoIy: is this a decent docstring? 
        """ Create new slice with informations collected from old slice and decorators 
            
        Parameters
        ----------
            _slice : Slice
                slice needed to collect (start, stop, step)
            array_size : PyccelArraySize
                call to function size()
            allow_negative_index = Bool
                True when the decorator allow_negative_index is present
        Returns
        -------
            Slice
        """

EmilyBourne: `allow_negative_index = Bool`->`allow_negative_index : Bool`

yguclu (approved):

pyccel/codegen/printing/ccode.py
yguclu: Yes, this works. Thanks

pyccel/codegen/printing/fcode.py
yguclu: Thanks for changing this.

pyccel/ast/core.py
```diff
@@ -4929,7 +4923,7 @@ def _sympystr(self, printer):
 
 
 
-class Slice(Basic):
+class Slice(Basic, PyccelOperator):
```
EmilyBourne, line 4926: @MenoIy I seem to have missed this in the review. Can you explain why you added this inheritance without moving `Slice` to `ast.operators.py` and without using any of the properties of `PyccelOperators` (notably it is important to call the `__init__` function)?

I wanted to move `Slice` in PR #637 but this inheritance stops me putting it where I was going to so I wanted to understand why it is necessary

Issue #516, opened 2020-10-27
Replace check_output with Popen


EmilyBourne:
**Describe the bug**
There are a few different ways of running subprocesses. Currently we use 2 different methods.

1. check_output:
    ```
    subprocess.check_output(cmd, stderr=subprocess.STDOUT, shell=True)
    ```
2. Popen:
    ```
    p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True)
    out, err = p.communicate
    ```

Popen should be preferred as `check_output` raises an error if the return code is not 0. As a result we cannot print the error message associated with the problem. 
With Popen the same result can be achieved with:
```
if p.returncode != 0:
    err_msg = "Failed to build module"
    if verbose:
        err_msg += "\n" + err
    raise RuntimeError(err_msg)
```

The `universal_newlines` argument of Popen also allows us to collect `str` output which does not need to be decoded

EmilyBourne:
This is necessary to find failures on PR #593 (marked as xfail for now)

PR #593, opened 2020-12-11
Extend OpenMP support by adding more pragmas

nhamidn:
This PR is about extending the support of OpenMP by adding more Construct and Directives and their clauses:

- Cover all the constructs that are inside the documentation;
- Add clauses to the new constructs;
- Support combined constructs;
- Add tests for new pragmas;
- Update the documentation to fit new changes and fixes.

### DIFF
--- a/pyccel/ast/omp.py
+++ b/pyccel/ast/omp.py
@@ -54,64 +54,60 @@ class OmpAnnotatedComment(Basic):
 class OMP_For_Loop(OmpAnnotatedComment):
     """ Represents an OpenMP Loop construct. """
     def __init__(self, txt):
-        OmpAnnotatedComment.__init__(self, txt)
+        super().__init__(txt)
 
 class OMP_Parallel_Construct(OmpAnnotatedComment):
     """ Represents an OpenMP Parallel construct. """
-    def __init__(self, txt, combined=None):
-        OmpAnnotatedComment.__init__(self, txt, combined)
+    pass
 
 class OMP_Task_Construct(OmpAnnotatedComment):
     """ Represents an OpenMP Task construct. """
     def __init__(self, txt):
-        OmpAnnotatedComment.__init__(self, txt)
+        super().__init__(txt)
 
 class OMP_Single_Construct(OmpAnnotatedComment):
     """ Represents an OpenMP Single construct. """
     def __init__(self, txt):
-        OmpAnnotatedComment.__init__(self, txt)
+        super().__init__(txt)
 
 class OMP_Critical_Construct(OmpAnnotatedComment):
     """ Represents an OpenMP Critical construct. """
     def __init__(self, txt):
-        OmpAnnotatedComment.__init__(self, txt)
+        super().__init__(txt)
 
 class OMP_Master_Construct(OmpAnnotatedComment):
     """ Represents OpenMP Master construct. """
     def __init__(self, txt):
-        OmpAnnotatedComment.__init__(self, txt)
+        super().__init__(txt)
 
 class OMP_Masked_Construct(OmpAnnotatedComment):
     """ Represents OpenMP Masked construct. """
-    def __init__(self, txt, combined=None):
-        OmpAnnotatedComment.__init__(self, txt, combined)
+    pass
 
 class OMP_Cancel_Construct(OmpAnnotatedComment):
     """ Represents OpenMP Cancel construct. """
     def __init__(self, txt):
-        OmpAnnotatedComment.__init__(self, txt)
+        super().__init__(txt)
 
 class OMP_Target_Construct(OmpAnnotatedComment):
     """ Represents OpenMP Target construct. """
-    def __init__(self, txt, combined=None):
-        OmpAnnotatedComment.__init__(self, txt, combined)
+    pass
 
 class OMP_Teams_Construct(OmpAnnotatedComment):
     """ Represents OpenMP Teams construct. """
-    def __init__(self, txt, combined=None):
-        OmpAnnotatedComment.__init__(self, txt, combined)
+    pass
 
 class OMP_Sections_Construct(OmpAnnotatedComment):
     """ Represents OpenMP Sections construct. """
     def __init__(self, txt):
-        OmpAnnotatedComment.__init__(self, txt)
+        super().__init__(txt)
 
 class OMP_Section_Construct(OmpAnnotatedComment):
     """ Represent OpenMP Section construct. """
     def __init__(self, txt):
-        OmpAnnotatedComment.__init__(self, txt)
+        super().__init__(txt)
 
 class Omp_End_Clause(OmpAnnotatedComment):
     """ Represents the End of an OpenMP block. """
     def __init__(self, txt):
-        OmpAnnotatedComment.__init__(self, txt)
+        super().__init__(txt)
--- a/pyccel/ast/omp.py
+++ b/pyccel/ast/omp.py
@@ -4,7 +4,10 @@
 # go to https://github.com/pyccel/pyccel/blob/master/LICENSE for full license details.     #
 #------------------------------------------------------------------------------------------#
 """
-This file contain the OpenMP types.
+OpenMP has several constructs and directives, and this file contains the OpenMP types that are supported.
+We represent some types with the OmpAnnotatedComment type.
+These types are detailed on our documentation:
+https://github.com/pyccel/pyccel/blob/master/tutorial/openmp.md
 """
 
 from .basic import Basic
@@ -133,7 +136,6 @@ class OMP_Teams_Construct(OmpAnnotatedComment):
     def name(self):
         """Name of the construct."""
         return 'teams'
-
 
 class OMP_Sections_Construct(OmpAnnotatedComment):
     """ Represents OpenMP Sections construct. """
--- a/pyccel/ast/omp.py
+++ b/pyccel/ast/omp.py
@@ -38,10 +38,12 @@ class OmpAnnotatedComment(Basic):
 
     @classmethod
     def is_multiline(cls):
+        """Used to check if the construct needs brackets."""
         return cls._is_multiline
 
     @property
     def name(self):
+        """Name of the construct."""
         return ''
 
     @property
@@ -67,6 +69,7 @@ class OMP_For_Loop(OmpAnnotatedComment):
 
     @property
     def name(self):
+        """Name of the construct."""
         return 'for'
 
 class OMP_Parallel_Construct(OmpAnnotatedComment):
@@ -74,6 +77,7 @@ class OMP_Parallel_Construct(OmpAnnotatedComment):
     _is_multiline = True
     @property
     def name(self):
+        """Name of the construct."""
         return 'parallel'
 
 class OMP_Task_Construct(OmpAnnotatedComment):
@@ -105,6 +109,7 @@ class OMP_Masked_Construct(OmpAnnotatedComment):
     _is_multiline = True
     @property
     def name(self):
+        """Name of the construct."""
         return 'masked'
 
 class OMP_Cancel_Construct(OmpAnnotatedComment):
@@ -117,6 +122,7 @@ class OMP_Target_Construct(OmpAnnotatedComment):
     _is_multiline = True
     @property
     def name(self):
+        """Name of the construct."""
         return 'target'
 
 
@@ -125,6 +131,7 @@ class OMP_Teams_Construct(OmpAnnotatedComment):
     _is_multiline = True
     @property
     def name(self):
+        """Name of the construct."""
         return 'teams'
 
 
--- a/pyccel/ast/omp.py
+++ b/pyccel/ast/omp.py
@@ -51,12 +51,15 @@ class OmpAnnotatedComment(Basic):
         args = (self.txt, self.combined)
         return args
 
+class OmpCombinedAnnotatedComment(OmpAnnotatedComment):
+    """ Represents a Combined construct. """
+
 class OMP_For_Loop(OmpAnnotatedComment):
     """ Represents an OpenMP Loop construct. """
     def __init__(self, txt):
         super().__init__(txt)
 
-class OMP_Parallel_Construct(OmpAnnotatedComment):
+class OMP_Parallel_Construct(OmpCombinedAnnotatedComment):
     """ Represents an OpenMP Parallel construct. """
 
 class OMP_Task_Construct(OmpAnnotatedComment):
@@ -79,7 +82,7 @@ class OMP_Master_Construct(OmpAnnotatedComment):
     def __init__(self, txt):
         super().__init__(txt)
 
-class OMP_Masked_Construct(OmpAnnotatedComment):
+class OMP_Masked_Construct(OmpCombinedAnnotatedComment):
     """ Represents OpenMP Masked construct. """
 
 class OMP_Cancel_Construct(OmpAnnotatedComment):
@@ -87,10 +90,10 @@ class OMP_Cancel_Construct(OmpAnnotatedComment):
     def __init__(self, txt):
         super().__init__(txt)
 
-class OMP_Target_Construct(OmpAnnotatedComment):
+class OMP_Target_Construct(OmpCombinedAnnotatedComment):
     """ Represents OpenMP Target construct. """
 
-class OMP_Teams_Construct(OmpAnnotatedComment):
+class OMP_Teams_Construct(OmpCombinedAnnotatedComment):
     """ Represents OpenMP Teams construct. """
 
 class OMP_Sections_Construct(OmpAnnotatedComment):
--- a/pyccel/ast/omp.py
+++ b/pyccel/ast/omp.py
@@ -58,7 +58,6 @@ class OMP_For_Loop(OmpAnnotatedComment):
 
 class OMP_Parallel_Construct(OmpAnnotatedComment):
     """ Represents an OpenMP Parallel construct. """
-    pass
 
 class OMP_Task_Construct(OmpAnnotatedComment):
     """ Represents an OpenMP Task construct. """
@@ -82,7 +81,6 @@ class OMP_Master_Construct(OmpAnnotatedComment):
 
 class OMP_Masked_Construct(OmpAnnotatedComment):
     """ Represents OpenMP Masked construct. """
-    pass
 
 class OMP_Cancel_Construct(OmpAnnotatedComment):
     """ Represents OpenMP Cancel construct. """
@@ -91,11 +89,9 @@ class OMP_Cancel_Construct(OmpAnnotatedComment):
 
 class OMP_Target_Construct(OmpAnnotatedComment):
     """ Represents OpenMP Target construct. """
-    pass
 
 class OMP_Teams_Construct(OmpAnnotatedComment):
     """ Represents OpenMP Teams construct. """
-    pass
 
 class OMP_Sections_Construct(OmpAnnotatedComment):
     """ Represents OpenMP Sections construct. """
--- a/pyccel/ast/omp.py
+++ b/pyccel/ast/omp.py
@@ -29,11 +29,20 @@ class OmpAnnotatedComment(Basic):
     OmpAnnotatedComment(parallel)
     """
     _attribute_nodes = ()
+    _is_multiline = False
 
     def __init__(self, txt, combined=None):
         self._txt = txt
         self._combined = combined
         super().__init__()
+
+    @classmethod
+    def is_multiline(cls):
+        return cls._is_multiline
+
+    @property
+    def name(self):
+        return ''
 
     @property
     def txt(self):
@@ -51,58 +60,83 @@ class OmpAnnotatedComment(Basic):
         args = (self.txt, self.combined)
         return args
 
-class OmpCombinedAnnotatedComment(OmpAnnotatedComment):
-    """ Represents a Combined construct. """
-
 class OMP_For_Loop(OmpAnnotatedComment):
     """ Represents an OpenMP Loop construct. """
     def __init__(self, txt):
         super().__init__(txt)
 
-class OMP_Parallel_Construct(OmpCombinedAnnotatedComment):
+    @property
+    def name(self):
+        return 'for'
+
+class OMP_Parallel_Construct(OmpAnnotatedComment):
     """ Represents an OpenMP Parallel construct. """
+    _is_multiline = True
+    @property
+    def name(self):
+        return 'parallel'
 
 class OMP_Task_Construct(OmpAnnotatedComment):
     """ Represents an OpenMP Task construct. """
+    _is_multiline = True
     def __init__(self, txt):
         super().__init__(txt)
 
 class OMP_Single_Construct(OmpAnnotatedComment):
     """ Represents an OpenMP Single construct. """
+    _is_multiline = True
     def __init__(self, txt):
         super().__init__(txt)
 
 class OMP_Critical_Construct(OmpAnnotatedComment):
     """ Represents an OpenMP Critical construct. """
+    _is_multiline = True
     def __init__(self, txt):
         super().__init__(txt)
 
 class OMP_Master_Construct(OmpAnnotatedComment):
     """ Represents OpenMP Master construct. """
+    _is_multiline = True
     def __init__(self, txt):
         super().__init__(txt)
 
-class OMP_Masked_Construct(OmpCombinedAnnotatedComment):
+class OMP_Masked_Construct(OmpAnnotatedComment):
     """ Represents OpenMP Masked construct. """
+    _is_multiline = True
+    @property
+    def name(self):
+        return 'masked'
 
 class OMP_Cancel_Construct(OmpAnnotatedComment):
     """ Represents OpenMP Cancel construct. """
     def __init__(self, txt):
         super().__init__(txt)
 
-class OMP_Target_Construct(OmpCombinedAnnotatedComment):
+class OMP_Target_Construct(OmpAnnotatedComment):
     """ Represents OpenMP Target construct. """
+    _is_multiline = True
+    @property
+    def name(self):
+        return 'target'
 
-class OMP_Teams_Construct(OmpCombinedAnnotatedComment):
+
+class OMP_Teams_Construct(OmpAnnotatedComment):
     """ Represents OpenMP Teams construct. """
+    _is_multiline = True
+    @property
+    def name(self):
+        return 'teams'
+
 
 class OMP_Sections_Construct(OmpAnnotatedComment):
     """ Represents OpenMP Sections construct. """
+    _is_multiline = True
     def __init__(self, txt):
         super().__init__(txt)
 
 class OMP_Section_Construct(OmpAnnotatedComment):
     """ Represent OpenMP Section construct. """
+    _is_multiline = True
     def __init__(self, txt):
         super().__init__(txt)
 
--- a/pyccel/ast/omp.py
+++ b/pyccel/ast/omp.py
@@ -36,10 +36,10 @@ class OmpAnnotatedComment(Basic):
         self._combined = combined
         super().__init__()
 
-    @classmethod
-    def is_multiline(cls):
+    @property
+    def is_multiline(self):
         """Used to check if the construct needs brackets."""
-        return cls._is_multiline
+        return self._is_multiline
 
     @property
     def name(self):
--- a/pyccel/ast/omp.py
+++ b/pyccel/ast/omp.py
@@ -97,7 +97,7 @@ class OMP_Target_Construct(OmpAnnotatedComment):
         OmpAnnotatedComment.__init__(self, txt, combined)
 
 class OMP_Teams_Construct(OmpAnnotatedComment):
-    """ Represents OpenMP Target construct. """
+    """ Represents OpenMP Teams construct. """
     def __init__(self, txt, combined=None):
         OmpAnnotatedComment.__init__(self, txt, combined)
 
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -1495,7 +1495,7 @@ class CCodePrinter(CodePrinter):
         clauses += str(expr.txt)
         omp_expr = '#pragma omp {}{}'.format(expr.name, clauses)
 
-        if expr._is_multiline:
+        if expr.is_multiline:
             if expr.combined is None:
                 omp_expr += '\n{'
             elif (expr.combined and "for" not in expr.combined):
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -457,7 +457,6 @@ class CCodePrinter(CodePrinter):
                             PyccelMul(expr.imag, LiteralImaginaryUnit()))))
 
     def _print_PythonComplex(self, expr):
-        self._additional_imports.add("complex")
         if expr.is_cast:
             value = self._print(expr.internal_var)
         else:
@@ -467,6 +466,7 @@ class CCodePrinter(CodePrinter):
         return '({0})({1})'.format(type_name, value)
 
     def _print_LiteralImaginaryUnit(self, expr):
+        self._additional_imports.add("complex")
         return '_Complex_I'
 
     def _print_ModuleHeader(self, expr):
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -33,6 +33,8 @@ from pyccel.ast.literals  import Nil
 from pyccel.ast.numpyext import NumpyFull, NumpyArray, NumpyArange
 from pyccel.ast.numpyext import NumpyReal, NumpyImag, NumpyFloat
 
+from pyccel.ast.utilities import expand_to_loops
+
 from pyccel.ast.variable import ValuedVariable
 from pyccel.ast.variable import PyccelArraySize, Variable, VariableAddress
 from pyccel.ast.variable import DottedName
@@ -517,7 +519,7 @@ class CCodePrinter(CodePrinter):
             var = self._print(e)
             if i == 0:
                 lines.append("if (%s)\n{" % self._print(c))
-            elif i == len(expr.blocks) - 1 and c is LiteralTrue():
+            elif i == len(expr.blocks) - 1 and isinstance(c, LiteralTrue):
                 lines.append("else\n{")
             else:
                 lines.append("else if (%s)\n{" % self._print(c))
@@ -1331,13 +1333,15 @@ class CCodePrinter(CodePrinter):
                 stop=stop, step=step, body=body)
 
     def _print_CodeBlock(self, expr):
-        body = []
-        for b in expr.body :
+        body_exprs, new_vars = expand_to_loops(expr, self._parser.get_new_variable, language_has_vectors = False)
+        self._additional_declare.extend(new_vars)
+        body_stmts = []
+        for b in body_exprs :
             code = self._print(b)
             code = self._additional_code + code
             self._additional_code = ''
-            body.append(code)
-        return '\n'.join(self._print(b) for b in body)
+            body_stmts.append(code)
+        return '\n'.join(self._print(b) for b in body_stmts)
 
     def _print_Idx(self, expr):
         return self._print(expr.label)
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -252,17 +252,17 @@ class CCodePrinter(CodePrinter):
         'dereference': set()
     }
 
-    def __init__(self, parser, settings=None):
+    def __init__(self, parser, **settings):
 
         if parser.filename:
             errors.set_target(parser.filename, 'file')
 
-        prefix_module = None if settings is None else settings.pop('prefix_module', None)
+        prefix_module = settings.pop('prefix_module', None)
         CodePrinter.__init__(self, settings)
         self.known_functions = dict(known_functions)
-        userfuncs = {} if settings is None else settings.get('user_functions', {})
+        userfuncs = settings.get('user_functions', {})
         self.known_functions.update(userfuncs)
-        self._dereference = set([] if settings is None else settings.get('dereference', []))
+        self._dereference = set(settings.get('dereference', []))
         self.prefix_module = prefix_module
         self._additional_imports = set(['stdlib'])
         self._parser = parser
@@ -445,11 +445,8 @@ class CCodePrinter(CodePrinter):
         value = self._print(expr.arg)
         return '({} != 0)'.format(value)
 
-    def _print_LiteralInteger(self, expr):
-        return str(expr.p)
-
-    def _print_LiteralFloat(self, expr):
-        return CodePrinter._print_Float(self, expr)
+    def _print_Literal(self, expr):
+        return repr(expr.python_value)
 
     def _print_LiteralComplex(self, expr):
         if expr.real == LiteralFloat(0):
@@ -1635,4 +1632,4 @@ def ccode(expr, parser, assign_to=None, **settings):
         For example, if ``dereference=[a]``, the resulting code would print
         ``(*a)`` instead of ``a``.
     """
-    return CCodePrinter(parser, settings).doprint(expr, assign_to)
+    return CCodePrinter(parser, **settings).doprint(expr, assign_to)
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -1495,45 +1495,20 @@ class CCodePrinter(CodePrinter):
 
     #=================== OMP ==================
 
-    def _print_OmpCombinedAnnotatedComment(self, expr):
+    def _print_OmpAnnotatedComment(self, expr):
         clauses = ''
         if expr.combined:
             clauses = ' ' + expr.combined
         clauses += str(expr.txt)
-        omp_expr = '#pragma omp'
-        if isinstance(expr, OMP_Parallel_Construct):
-            omp_expr += ' parallel{}'.format(clauses)
-        elif isinstance(expr, OMP_Masked_Construct):
-            omp_expr += ' masked{}'.format(clauses)
-        elif isinstance(expr, OMP_Target_Construct):
-            omp_expr += ' target{}'.format(clauses)
-        elif isinstance(expr, OMP_Teams_Construct):
-            omp_expr += ' teams{}'.format(clauses)
+        omp_expr = '#pragma omp {}{}'.format(expr.name, clauses)
 
-        if expr.combined is None:
-            omp_expr += '\n{'
-        else:
-            if isinstance(expr, OMP_Parallel_Construct):
-                if (expr.combined and "for" not in expr.combined):
-                    if "masked taskloop" not in expr.combined:
-                        omp_expr += '\n{'
-            elif isinstance(expr, OMP_Target_Construct):
-                if (expr.combined and "for" not in expr.combined):
-                    if "distribute" not in expr.combined:
-                        omp_expr += '\n{'
+        if expr._is_multiline:
+            if expr.combined is None:
+                omp_expr += '\n{'
+            elif (expr.combined and "for" not in expr.combined):
+                if ("masked taskloop" not in expr.combined) and ("distribute" not in expr.combined):
+                    omp_expr += '\n{'
 
-        return omp_expr
-
-    def _print_OmpAnnotatedComment(self, expr):
-        omp_expr = str(expr.txt)
-        if isinstance(expr, OMP_For_Loop):
-            omp_expr = '#pragma omp for{}'.format(omp_expr)
-        else:
-            omp_expr = '#pragma omp {}'.format(omp_expr)
-        if isinstance(expr, (OMP_Sections_Construct, OMP_Section_Construct,
-                             OMP_Task_Construct, OMP_Single_Construct,
-                             OMP_Critical_Construct, OMP_Master_Construct)):
-            omp_expr += '\n{'
         return omp_expr
 
     def _print_Omp_End_Clause(self, expr):
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -1518,7 +1518,7 @@ class CCodePrinter(CodePrinter):
     def _print_OmpAnnotatedComment(self, expr):
         omp_expr = str(expr.txt)
         if isinstance(expr, OMP_For_Loop):
-            omp_expr = '#pragma omp for'.format(omp_expr)
+            omp_expr = '#pragma omp for{}'.format(omp_expr)
         else:
             omp_expr = '#pragma omp {}'.format(omp_expr)
         if isinstance(expr, (OMP_Sections_Construct, OMP_Section_Construct,
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -11,7 +11,7 @@ import operator
 from pyccel.ast.builtins  import PythonRange, PythonFloat, PythonComplex
 
 from pyccel.ast.core      import Declare
-from pyccel.ast.core      import FuncAddressDeclare, FunctionCall
+from pyccel.ast.core      import FuncAddressDeclare, FunctionCall, FunctionDef
 from pyccel.ast.core      import Deallocate
 from pyccel.ast.core      import FunctionAddress
 from pyccel.ast.core      import Assign, datatype, Import, AugAssign
@@ -1125,7 +1125,11 @@ class CCodePrinter(CodePrinter):
         body  = self._print(expr.body)
         decs  = [Declare(i.dtype, i) if isinstance(i, Variable) else FuncAddressDeclare(i) for i in expr.local_vars]
         if len(expr.results) <= 1 :
-            decs += [Declare(i.dtype, i) if isinstance(i, Variable) else FuncAddressDeclare(i) for i in expr.results]
+            for i in expr.results:
+                if isinstance(i, Variable) and not i.is_temp:
+                    decs += [Declare(i.dtype, i)]
+                elif not isinstance(i, Variable):
+                    decs += [FuncAddressDeclare(i)]
         decs += [Declare(i.dtype, i) for i in self._additional_declare]
         decs  = '\n'.join(self._print(i) for i in decs)
         self._additional_declare.clear()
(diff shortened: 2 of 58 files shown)


EmilyBourne:
What is the status of this PR?

nhamidn:
> What is the status of this PR?

there is two pragmas left and some tests

EmilyBourne:
Have you thought about creating an `OmpAnnotatedComment` class to reduce code duplication?

nhamidn:
> Have you thought about creating an `OmpAnnotatedComment` class to reduce code duplication?

it will be good, I'm gonna do it, but not for all the directives because there are some differences

EmilyBourne:
I wonder if we should create a file for the OpenMP types (e.g. `pyccel/ast/omp.py`). It seems strange to have them in `core.py` as they are not a core element, even if they have always been there historically

nhamidn:
> I wonder if we should create a file for the OpenMP types (e.g. `pyccel/ast/omp.py`)? It seems strange to have them in `core.py` as they are not a core element, even if they have always been there historically

it's a good idea

EmilyBourne:
Don't forget to remove the `@xfail` marks on everything you've fixed

pyccel/codegen/printing/ccode.py
```diff
@@ -1466,16 +1466,92 @@ def _print_EmptyNode(self, expr):
...
 
+    def _print_OMP_Critical_Construct(self, expr):
+        omp_expr = str(expr.txt)
+        omp_expr = '#pragma omp {}\n{{'.format(omp_expr)
+        return omp_expr
+
+    def _print_OMP_Master_Construct(self, expr):
+        omp_expr = str(expr.txt)
+        omp_expr = '#pragma omp {}\n{{'.format(omp_expr)
+        return omp_expr
```
EmilyBourne, line 1496: Several of these functions have the same body. Can't we write a `_print_OmpAnnotatedComment` which will handle all these cases?

pyccel/codegen/printing/ccode.py
```diff
@@ -1466,16 +1466,92 @@ def _print_EmptyNode(self, expr):
...
+
+    def _print_OMP_Teams_Construct(self, expr):
+        clauses = ''
+        if expr.combined:
+            clauses = ' ' + expr.combined
+        clauses += str(expr.txt)
+        omp_expr = '#pragma omp teams{}'.format(clauses)
+        if expr.combined is None:
+            omp_expr += '\n{'
+        return omp_expr
```
EmilyBourne, line 1538: Similar to above, could we have an `OmpCombinedAnnotatedComment` class to group these cases?

pyccel/codegen/printing/ccode.py
```diff
@@ -1466,16 +1466,92 @@ def _print_EmptyNode(self, expr):
...
+    def _print_OMP_Target_Construct(self, expr):
+        clauses = ''
+        if expr.combined:
+            clauses = ' ' + expr.combined
+        clauses += str(expr.txt)
+        omp_expr = '#pragma omp target{}'.format(clauses)
+        if expr.combined is None:
+            omp_expr += '\n{'
+        elif (expr.combined and "for" not in expr.combined):
+            omp_expr += '\n{'
```
EmilyBourne, line 1522: Some of the combined cases make an exception for `for`, but not all. Is that intentional?

pyccel/codegen/printing/fcode.py
```diff
@@ -1860,24 +1862,85 @@ def _do_range(target, iterable, prolog, epilog):
...
+
+    def _print_OMP_Critical_Construct(self, expr):
+        omp_expr = str(expr.txt)
+        omp_expr = '!$omp {}\n'.format(omp_expr)
+        return omp_expr
+
+    def _print_OMP_Master_Construct(self, expr):
+        omp_expr = str(expr.txt)
+        omp_expr = '!$omp {}\n'.format(omp_expr)
+        return omp_expr
```
EmilyBourne, line 1890: See above `_print_OmpAnnotatedComment`

pyccel/parser/syntax/openmp.py
```diff
@@ -11,7 +11,9 @@
 from textx.metamodel import metamodel_from_file
 
 from pyccel.parser.syntax.basic import BasicStmt
-from pyccel.ast.core import OMP_For_Loop, OMP_Parallel_Construct, OMP_Single_Construct, Omp_End_Clause
+from pyccel.ast.omp import OmpAnnotatedComment, OMP_For_Loop, OMP_Parallel_Construct, OMP_Single_Construct,\
+        Omp_End_Clause, OMP_Critical_Construct, OMP_Master_Construct,\
+        OMP_Masked_Construct, OMP_Task_Construct, OMP_Cancel_Construct, OMP_Target_Construct, OMP_Teams_Construct, OMP_Sections_Construct, OMP_Section_Construct
```
EmilyBourne, line 16: The syntax used in the rest of pyccel to group lots of imports is:
```
from pyccel.ast.omp import (OmpAnnotatedComment, OMP_For_Loop, OMP_Parallel_Construct, 
                            OMP_Single_Construct, Omp_End_Clause, OMP_Critical_Construct,
                            OMP_Master_Construct, OMP_Masked_Construct, OMP_Task_Construct,
                            OMP_Cancel_Construct, OMP_Target_Construct, OMP_Teams_Construct,
                            OMP_Sections_Construct, OMP_Section_Construct)
```

pyccel/parser/syntax/openmp.py
```diff
@@ -47,81 +49,178 @@ def expr(self):
...
+        elif isinstance(stmt, OmpTargetConstruct):
+            return stmt.expr
+        elif isinstance(stmt, OmpTeamsConstruct):
+            return stmt.expr
+        elif isinstance(stmt, OmpDistributeConstruct):
+            return stmt.expr
+        elif isinstance(stmt, OmpSectionConstruct):
+            return stmt.expr
+        elif isinstance(stmt, OmpSectionsConstruct):
+            return stmt.expr
```
EmilyBourne, line 85: Why do you need so many `elif` statements?
Can you not do:
```
elif isinstance( stmt, (OmpSingleConstruct, OmpCriticalConstruct,...))
```
Or better:
```
elif isinstance(stmt, OmpAnnotatedComment) and not isinstance(stmt, X)
```

pyccel/parser/syntax/openmp.py
```diff
@@ -47,81 +49,178 @@ def expr(self):
...
         """
-        self.clauses = kwargs.pop('clauses')
+        self.clauses  = kwargs.pop('clauses')
+        self.combined = kwargs.pop('combined')
 
-        super(OmpParallelConstruct, self).__init__(**kwargs)
+        super().__init__(**kwargs)
 
     @property
     def expr(self):
```
EmilyBourne, line 100: If this property always returns the same value, it would be better to do the calculations once, either in the `__init__` or in a function called by `__init__`

pyccel/parser/syntax/openmp.py
```diff
@@ -47,81 +49,178 @@ def expr(self):
...
+        _valid_clauses = _valid_taskloop_clauses + (OmpinReduction,)
+
+        txt = self.name
+        for clause in self.clauses:
+            if isinstance(clause, _valid_clauses):
+                txt = '{0} {1}'.format(txt, clause.expr)
+            else:
+                raise TypeError('Wrong clause for OmpTaskLoopConstruct. Given : ', \
+                                type(clause))
+        return OmpAnnotatedComment(txt)
```
EmilyBourne, line 178: See above, unnecessary calculations in `expr` is called multiple times

pyccel/parser/syntax/openmp.py
```diff
@@ -47,81 +49,178 @@ def expr(self):
...
+                          OmpMergeable, \
+                          OmpinReduction, \
+                          OmpDepend)
+
+        txt = self.name
+        for clause in self.clauses:
+            if isinstance(clause, _valid_clauses):
+                txt = '{0} {1}'.format(txt, clause.expr)
+            else:
+                raise TypeError('Wrong clause for OmpTaskConstruct')
```
EmilyBourne, line 211: See above, unnecessary calculations in expr is called multiple times

pyccel/parser/syntax/openmp.py
```diff
@@ -131,60 +230,504 @@ def expr(self):
...
-            else:
-                raise TypeError('Wrong clause for OmpSingleConstruct')
+        txt = self.name
+        for clause in self.clauses:
+            if isinstance(clause, _valid_clauses):
+                txt = '{0} {1}'.format(txt, clause.expr)
+            else:
+                raise TypeError('Wrong clause for OmpSingleConstruct')
+
+        return OMP_Single_Construct(txt)
```
EmilyBourne, line 240: See above, unnecessary calculations in expr is called multiple times

pyccel/parser/syntax/openmp.py
```diff
@@ -131,60 +230,504 @@ def expr(self):
...
+        _valid_clauses = (OmpCriticalName)
+
+        txt = self.name
+        for clause in self.clauses:
+            if isinstance(clause, _valid_clauses):
+                txt = '{0} {1}'.format(txt, clause.expr)
+            else:
+                raise TypeError('Wrong clause for OmpCriticalConstruct')
+
+        return OMP_Critical_Construct(txt)
```
EmilyBourne, line 266: See above, unnecessary calculations in expr is called multiple times

pyccel/parser/syntax/openmp.py
```diff
@@ -131,60 +230,504 @@ def expr(self):
...
+            print("> OmpSimdConstruct: expr")
+
+        txt = self.name
+        for clause in self.clauses:
+            if isinstance(clause, _valid_simd_clauses):
+                txt = '{0} {1}'.format(txt, clause.expr)
+            else:
+                raise TypeError('Wrong clause for OmpSimdConstruct')
+
+        return OmpAnnotatedComment(txt)
```
EmilyBourne, line 288: See above, unnecessary calculations in expr is called multiple times. Etc below

pyccel/parser/syntax/openmp.py
```diff
@@ -131,60 +230,504 @@ def expr(self):
...
+            print("> OmpCriticalConstruct: expr")
+
+        _valid_clauses = (OmpCriticalName)
+
+        txt = self.name
+        for clause in self.clauses:
+            if isinstance(clause, _valid_clauses):
+                txt = '{0} {1}'.format(txt, clause.expr)
+            else:
+                raise TypeError('Wrong clause for OmpCriticalConstruct')
```
EmilyBourne, line 264: Several of these classes contain these same lines of code. Could these be defined as a function of a common superclass to reduce code repetition?

EmilyBourne: See `PyccelOperator` for an example of using super classes to reduce code duplication as much as possible

pyccel/parser/syntax/openmp.py
```diff
@@ -390,15 +1027,380 @@ def expr(self):
...
+                if self.ssname:
+                    txt += ' ' + self.ssname
+        return txt
 #################################################
 
 #################################################
 # whenever a new rule is added in the grammar, we must update the following
 # lists.
+
+_valid_teams_clauses = (OmpPrivate,\
```
EmilyBourne, line 1329: By the way, I don't think the line continuation character `\` is necessary in a tuple.

pyccel/parser/syntax/openmp.py
```diff
@@ -413,7 +1415,32 @@ def expr(self):
...
+               OmpThreadLimit,
+               OmpForSimd,
+               OmpMaskedTaskloop,
+               OmpPSections,
+               OmpTaskloopSimd,
+               OmpDistributeCombined,
+               OmpTargetParallel,
+               OmpTargetTeams]
 
 omp_classes = [Openmp, OpenmpStmt] + omp_directives + omp_clauses
```
EmilyBourne, line 1445: Do you know if there's a reason why these last 3 are lists? I agree with you that they should be tuples. They should be immutable

tests/epyccel/modules/openmp.py
```diff
@@ -124,7 +124,7 @@ def test_omp_in_final():
 
     #$ omp parallel
     #$ omp single
-    #$ omp task final (i >= 10)
+    #$ omp task final(i >= 10)
```
EmilyBourne, line 127: Is this change important? If I did `#$ omp task final (i >= 10)` would pyccel still understand?

EmilyBourne (changes requested):

pyccel/codegen/printing/ccode.py
nhamidn: because for example when we have `#$ omp parallel for` we know that were gonna have a for loop after the pragma so we don't need to have an `#$ omp end parallel for`

tests/epyccel/modules/openmp.py
nhamidn: i didnt notice i changed it, there is no reason for this change I was just testing

EmilyBourne: Ok :+1: I just wanted to check it wasn't something that needed documenting

pyccel/codegen/printing/ccode.py
EmilyBourne: Your answer seems to explain why you have `and "for" not in expr.combined`. My question wasn't why you have it here, but if it should be in every class with a `combined` property. E.g. [here](https://github.com/pyccel/pyccel/pull/593/files/56efb97a63fb0f4d99fe5c9bc6659bd9cb3afcfa#diff-f1038cb7408f08867011c5feb494cd9e7a5a435a3b6f6438ef6cd3bdf3c90fbeR1536)

In codegen/printing/ccode.py, sometimes you do:
```
elif (expr.combined and "for" not in expr.combined):
    omp_expr += '\n{'
```
and sometimes you do:
```
if expr.combined is None:
    omp_expr += '\n{
```
I'm not sure if this is due to a difference in the OMP statements or due to a correction that hasn't been done everywhere?

nhamidn: but a `parallel` construct can be combined with the `sections` construct, so when it's combined with the `sections` construct we need to open the bracket

nhamidn: it depend on whats inside the `combined` variable, I try to not open the bracket when there will be only one statement after the pragma

EmilyBourne: The question is about the condition

EmilyBourne: >it depend on whats inside the combined variable, I try to not open the bracket when there will be only one statement after the pragma

This is my point. Presumably all combined variables should avoid opening brackets if `"for" in expr.combined`? So why do some classes not have this condition?

nhamidn: aah i see your question, because the `for` can't be combined with any construct, and if it's the first element after a construct we know that there is a `for` thank's to textx so we need to check just if `expr.combined is not None`

nhamidn: like `#$ omp teams distribute`, here if there is any combined construct after the `teams` it needs to be a `distribute` construct, so we check it using textx if the `expr.combined is not None` then we don't open the bracket, and what come after the `distribute` is optional we need to check it

EmilyBourne: Ok. So yes it is intentional :+1:

EmilyBourne: Presumably it wouldn't matter if we did check if `"for" in expr.combined` even if it's not necessary? (Just wondering if this could help for reducing code duplication)

pyccel/parser/grammar/openmp.tx
```diff
@@ -5,86 +5,177 @@ Openmp:
...
-OmpLoopConstruct:      'for'       clauses*=OmpLoopClause;
-OmpSingleConstruct:   'single'   clauses*=OmpSingleClause;
+OmpParallelConstruct:   'parallel'        combined=ParallelCombined? clauses*=AllOmpClauses;
+OmpLoopConstruct:       'for'             clauses*=AllOmpClauses;
+OmpTaskLoopConstruct:   name='taskloop'   clauses*=AllOmpClauses;
+OmpSingleConstruct:     name='single'     clauses*=AllOmpClauses;
+OmpCriticalConstruct:   name='critical'   clauses*=AllOmpClauses;
+OmpMaskedConstruct:     name='masked'     combined=MaskedCombined? clauses*=AllOmpClauses;
+OmpSimdConstruct:       name='simd'       clauses*=AllOmpClauses;
+OmpAtomicConstruct:     name='atomic'     clauses*=AllOmpClauses; // Not all clause were added (cant test them).
```
mostafaMamoni, line 50: typos: clause -> clauses, cant -> can't

mostafaMamoni (changes requested):

pyccel/codegen/printing/ccode.py
nhamidn: but it could be `#$ omp parallel sections` this needs a structured block so we need to open the brackets

pyccel/codegen/printing/ccode.py
```diff
@@ -1466,16 +1466,92 @@ def _print_EmptyNode(self, expr):
...
+    def _print_OMP_Master_Construct(self, expr):
+        omp_expr = str(expr.txt)
+        omp_expr = '#pragma omp {}\n{{'.format(omp_expr)
+        return omp_expr
+
+    def _print_OMP_Masked_Construct(self, expr):
+        clauses = ''
+        if expr.combined:
+            clauses = ' ' + expr.combined
+        clauses += str(expr.txt)
```
mostafaMamoni, line 1502: i think this can be simplified by: `clauses  = expr.combined +  str(expr.txt) if expr.combined else str(expr.txt)`

pyccel/codegen/printing/ccode.py
```diff
@@ -1466,16 +1466,92 @@ def _print_EmptyNode(self, expr):
...
+    def _print_OMP_Task_Construct(self, expr):
+        omp_expr = str(expr.txt)
+        omp_expr = '#pragma omp {}\n{{'.format(omp_expr)
+        return omp_expr
+
+    def _print_OMP_Target_Construct(self, expr):
+        clauses = ''
+        if expr.combined:
+            clauses = ' ' + expr.combined
+        clauses += str(expr.txt)
```
mostafaMamoni, line 1517: See above.

pyccel/codegen/printing/ccode.py
```diff
@@ -1466,16 +1466,92 @@ def _print_EmptyNode(self, expr):
...
+    def _print_OMP_Cancel_Construct(self, expr):
+        omp_expr = str(expr.txt)
+        omp_expr = '#pragma omp {}\n'.format(omp_expr)
+        return omp_expr
+
+    def _print_OMP_Teams_Construct(self, expr):
+        clauses = ''
+        if expr.combined:
+            clauses = ' ' + expr.combined
+        clauses += str(expr.txt)
```
mostafaMamoni, line 1534: See above

pyccel/codegen/printing/ccode.py
EmilyBourne: Let me rephrase my 2nd question. Presumably doing:
```
def _print_OMP_Masked_Construct(self, expr):
        clauses = ''
        if expr.combined:
            clauses = ' ' + expr.combined
        clauses += str(expr.txt)
        omp_expr = '#pragma omp masked{}'.format(clauses)
        if expr.combined is None:
            omp_expr += '\n{'
        elif (expr.combined and "for" not in expr.combined):
            omp_expr += '\n{'
        return omp_expr
```
wouldn't cause any problems? I am asking as I am wondering if we could do:
```
def _print_OmpCombinedAnnotatedComment (self, expr):
        clauses = ''
        if expr.combined:
            clauses = ' ' + expr.combined
        clauses += str(expr.txt)
        omp_expr = '#pragma omp masked{}'.format(clauses)
        if expr.combined is None:
            omp_expr += '\n{'
        elif (expr.combined and "for" not in expr.combined):
            omp_expr += '\n{'
        return omp_expr
```

pyccel/codegen/printing/fcode.py
```diff
@@ -1860,24 +1862,85 @@ def _do_range(target, iterable, prolog, epilog):
...
+
+    def _print_OMP_Masked_Construct(self, expr):
+        clauses = ''
+        if expr.combined:
+            clauses = ' ' + expr.combined
+        clauses += str(expr.txt)
```
mostafaMamoni, line 1896: See above.

pyccel/codegen/printing/fcode.py
```diff
@@ -1860,24 +1862,85 @@ def _do_range(target, iterable, prolog, epilog):
...
-        ompexpr = '!$omp {}\n'.format(omp_expr)
-        return ompexpr
+        omp_expr = '!$omp {}\n'.format(omp_expr)
+        return omp_expr
+
+    def _print_OMP_Target_Construct(self, expr):
+        clauses = ''
+        if expr.combined:
+            clauses = ' ' + expr.combined.replace("for", "do")
+        clauses += str(expr.txt)
```
mostafaMamoni, line 1910: this can be simplified by: clauses = expr.combined.replace("for", "do") + str(expr.txt) if expr.combined else str(expr.txt)

pyccel/codegen/printing/fcode.py
```diff
@@ -1860,24 +1862,85 @@ def _do_range(target, iterable, prolog, epilog):
...
+        omp_expr = '!$omp target{}\n'.format(clauses)
+        return omp_expr
+
+    def _print_OMP_Teams_Construct(self, expr):
+        clauses = ''
+        if expr.combined:
+            clauses = ' ' + expr.combined.replace("for", "do")
+        clauses += str(expr.txt)
```
mostafaMamoni, line 1918: same thing here.

mostafaMamoni (changes requested):

pyccel/codegen/printing/ccode.py
nhamidn: aaah im gonna try to make a printer for the combined construct and manage the cases

pyccel/codegen/printing/ccode.py
EmilyBourne: This is shorter, but I'm not sure that it's clearer or faster

pyccel/parser/syntax/openmp.py
nhamidn: im not sure i understand what do you mean by called multiple time

EmilyBourne: I.e. If I do:
```
var = OmpTaskLoopConstruct(name=str1, clauses=str2)
tmp = var.expr
.....
f(var.expr)
```
Here I have accessed the property `var.expr` twice. So I have done the above calculation twice. If the calculation is in the `__init__` method then the calculation is only done once

nhamidn: aah i see, i moved all the calculations to the `__init__` function.

EmilyBourne: For example if you do:
```
def __init__(self, ...):
    ...
    self._expr = OmpAnnotatedComment(txt)
```
Then the property is simple:
```
@property
def expr(self):
    return self._expr
```

This is faster as there is less calculation, but it is also more memory efficient. With a `self._expr` we have `var.expr is var.expr == True` but with your current method `var.expr is var.expr == False`

pyccel/ast/omp.py
```diff
@@ -0,0 +1,117 @@
...
+
+class OMP_For_Loop(OmpAnnotatedComment):
+    """ Represents an OpenMP Loop construct. """
+    def __init__(self, txt):
+        OmpAnnotatedComment.__init__(self, txt)
+
+class OMP_Parallel_Construct(OmpAnnotatedComment):
+    """ Represents an OpenMP Parallel construct. """
+    def __init__(self, txt, combined=None):
+        OmpAnnotatedComment.__init__(self, txt, combined)
```
EmilyBourne, line 62: The signature for this `__init__` function is identical to the signature of the superclass so it does not need to be redefined. It can be inherited

pyccel/ast/omp.py
```diff
@@ -0,0 +1,117 @@
...
+    def __getnewargs__(self):
+        """Used for Pickling self."""
+
+        args = (self.txt, self.combined)
+        return args
+
+class OMP_For_Loop(OmpAnnotatedComment):
+    """ Represents an OpenMP Loop construct. """
+    def __init__(self, txt):
+        OmpAnnotatedComment.__init__(self, txt)
```
EmilyBourne, line 57: Please use `super().__init__(txt)`. I spent some time trying to make this usage the standard wherever possible in pyccel. It ensures that the superclass `__init__` is called correctly even if the class is renamed, or a class is added to the inheritance tree

pyccel/ast/omp.py
EmilyBourne: Same comment for:
- `OMP_Masked_Construct`
- `OMP_Target_Construct`
- `OMP_Teams_Construct`

pyccel/ast/omp.py
EmilyBourne: Same comment for the other classes in this file

pyccel/parser/syntax/openmp.py
```diff
@@ -31,142 +35,547 @@ def __init__(self, **kwargs):
...
-        elif isinstance(stmt, OmpLoopConstruct):
-            return stmt.expr
-        elif isinstance(stmt, OmpSingleConstruct):
+        if isinstance(stmt, (OmpEndClause, OmpParallelConstruct, OmpLoopConstruct,
+                             OmpSingleConstruct, OmpCriticalConstruct, OmpBarrierConstruct,
+                             OmpMasterConstruct, OmpMaskedConstruct, OmpTaskLoopConstruct,
+                             OmpSimdConstruct, OmpAtomicConstruct, OmpTaskWaitConstruct,
+                             OmpTaskyieldConstruct, OmpTaskConstruct, OmpFlushConstruct,
+                             OmpCancelConstruct, OmpTargetConstruct, OmpTeamsConstruct,
+                             OmpDistributeConstruct, OmpSectionConstruct, OmpSectionsConstruct)):
```
EmilyBourne, line 52: Is this all the `OmpAnnotatedComment`s?

pyccel/parser/syntax/openmp.py
```diff
@@ -31,142 +35,547 @@ def __init__(self, **kwargs):
...
         else:
             raise TypeError('Wrong stmt for OpenmpStmt')
 
+def check_get_clauses(name, valid_clauses, clauses):
+    txt = ''
+    for clause in clauses:
+        if isinstance(clause, valid_clauses):
+            txt = '{0} {1}'.format(txt, clause.expr)
+        else:
+            raise TypeError('Wrong clause', type(clause).__name__,'for', type(name).__name__)
```
EmilyBourne, line 63: Errors do not work like `print` statements. This will not concatenate your strings

pyccel/parser/syntax/openmp.py
```diff
@@ -31,142 +35,547 @@ def __init__(self, **kwargs):
...
-
-        txt = 'parallel'
-        for clause in self.clauses:
-            if isinstance(clause, _valid_clauses):
-                txt = '{0} {1}'.format(txt, clause.expr)
-            else:
-                raise TypeError('Wrong clause for OmpParallelConstruct')
-
-        return OMP_Parallel_Construct(txt)
+        return OMP_Parallel_Construct(self.txt, self.com)
```
EmilyBourne, line 102: Is there a reason why each call to `OmpParallelConstruct.expr` should create a new `OMP_Parallel_Construct` instance?

pyccel/parser/syntax/openmp.py
```diff
@@ -31,142 +35,547 @@ def __init__(self, **kwargs):
...
+
+        self.com = None
+        if isinstance(self.combined, OmpDistributeCombined):
+            self.com = self.combined.expr
+            _valid_clauses = _valid_clauses + _valid_Distribute_clauses
+            if 'simd' in self.combined.expr:
+                _valid_clauses = _valid_clauses + _valid_simd_clauses
+            if 'parallel' in self.combined.expr:
+                _valid_clauses = _valid_clauses + _valid_parallel_clauses
+                _valid_clauses = _valid_clauses + _valid_loop_clauses
```
EmilyBourne, line 497: I am surprised there are not more statements like this across your different classes. Is each combined type (e.g. `OmpDistributeCombined`, `OmpForSimd`) only associated with one construct type (e.g. `OmpTeamsConstruct`) ?
If not then I wonder if it makes sense for combined types to have a `valid_clauses` method which returns the additional clauses which are valid for that construct? That way you won't have to copy and paste if blocks if a combined type can be used by multiple constructs?

EmilyBourne (changes requested):

pyccel/parser/syntax/openmp.py
nhamidn: no those are a `textx` rules, here we still didn't create an `OmpAnnotatedComments`

EmilyBourne: Ah oui, pardon

pyccel/parser/syntax/openmp.py
nhamidn: i don't think there is a reason, I changed them to a tuple and everything is working normally.

pyccel/parser/syntax/openmp.py
nhamidn: i agree with you, but here each combined type is associated to only one construct type, but two combined types can have the same one or more elements, I'm gonna try to simplify it

pyccel/codegen/printing/ccode.py
```diff
@@ -1487,17 +1494,48 @@ def _print_EmptyNode(self, expr):
...
-        omp_expr   = str(expr.txt)
-        return '#pragma omp {}\n{{'.format(omp_expr)
+    def _print_OmpCombinedAnnotatedComment(self, expr):
+        clauses = ''
+        if expr.combined:
+            clauses = ' ' + expr.combined
+        clauses += str(expr.txt)
+        omp_expr = '#pragma omp'
+        if isinstance(expr, OMP_Parallel_Construct):
+            omp_expr += ' parallel{}'.format(clauses)
```
EmilyBourne, line 1505: Do you think `parallel`, `masked`, `target` and `teams` could be a property of the construct? E.g:
```
class OmpCombinedAnnotatedComment(OmpAnnotatedComment):
    """ Represents a Combined construct. """
    _is_multiline = False
    @classmethod
    def is_multiline(cls):
        return cls._is_multiline

    @property
    def keyword(self):
        return ''

class OMP_Teams_Construct(OmpCombinedAnnotatedComment):
    """ Represents OpenMP Teams construct. """
    _is_multiline = True
    @property
    def keyword(self):
        return 'teams'
```

Then you could do:
```
 def _print_OmpCombinedAnnotatedComment(self, expr):
        clauses = ''
        if expr.combined:
            clauses = ' ' + expr.combined
        clauses += str(expr.txt)
        omp_expr = '#pragma omp'
        if expr.is_multiline:
            omp_expr += ' {}{}'.format(expr.keyword, clauses)
            if expr.combined is None:
                omp_expr += '\n{'
            elif (expr.combined and "for" not in expr.combined):
                if "masked taskloop" not in expr.combined:
                    omp_expr += '\n{'
        return omp_expr
```

pyccel/codegen/printing/fcode.py
```diff
@@ -1859,25 +1866,44 @@ def _do_range(target, iterable, prolog, epilog):
...
+            clauses = ' ' + combined
+        omp_expr = '!$omp'
+        if isinstance(expr, OMP_Parallel_Construct):
+            omp_expr += ' parallel'
+        elif isinstance(expr, OMP_Masked_Construct):
+            omp_expr += ' masked'
+        elif isinstance(expr, OMP_Target_Construct):
+            omp_expr += ' target'
+        elif isinstance(expr, OMP_Teams_Construct):
+            omp_expr += ' teams'
```
EmilyBourne, line 1885: See above. Why not use a property?

pyccel/codegen/printing/ccode.py
```diff
@@ -1487,17 +1494,48 @@ def _print_EmptyNode(self, expr):
...
+
+    def _print_OmpAnnotatedComment(self, expr):
+        omp_expr = str(expr.txt)
+        if isinstance(expr, OMP_For_Loop):
+            omp_expr = '#pragma omp for{}'.format(omp_expr)
+        else:
+            omp_expr = '#pragma omp {}'.format(omp_expr)
+        if isinstance(expr, (OMP_Sections_Construct, OMP_Section_Construct,
+                             OMP_Task_Construct, OMP_Single_Construct,
+                             OMP_Critical_Construct, OMP_Master_Construct)):
```
EmilyBourne, line 1536: Is there a property that make these cases special? (e.g. is_multiline) if so, then I think it would be better to define the property so you don't need to update this list in every implemented language whenever you want to add a new construct

pyccel/parser/syntax/openmp.py
```diff
@@ -31,160 +35,686 @@ def __init__(self, **kwargs):
...
-        elif isinstance(stmt, OmpLoopConstruct):
-            return stmt.expr
-        elif isinstance(stmt, OmpSingleConstruct):
+        if isinstance(stmt, (OmpEndClause, OmpParallelConstruct, OmpLoopConstruct,
+                             OmpSingleConstruct, OmpCriticalConstruct, OmpBarrierConstruct,
+                             OmpMasterConstruct, OmpMaskedConstruct, OmpTaskLoopConstruct,
+                             OmpSimdConstruct, OmpAtomicConstruct, OmpTaskWaitConstruct,
+                             OmpTaskyieldConstruct, OmpTaskConstruct, OmpFlushConstruct,
+                             OmpCancelConstruct, OmpTargetConstruct, OmpTeamsConstruct,
+                             OmpDistributeConstruct, OmpSectionConstruct, OmpSectionsConstruct)):
```
EmilyBourne, line 52: Is there no way to simplify this?

pyccel/parser/syntax/openmp.py
```diff
@@ -31,160 +35,686 @@ def __init__(self, **kwargs):
...
+        com = None
+        if isinstance(self.combined, OmpForSimd):
+            _valid_clauses = _valid_clauses + _valid_loop_clauses
+            if 'simd' in self.combined.expr:
+                _valid_clauses = _valid_clauses + _valid_simd_clauses
+            com = self.combined.expr
+        if isinstance(self.combined, OmpMaskedTaskloop):
+            _valid_clauses = _valid_clauses + (OmpFilter,)
+            if 'simd' in self.combined.expr:
+                _valid_clauses = _valid_clauses + _valid_simd_clauses
```
EmilyBourne, line 89: Here you do `_valid_clauses = _valid_clauses + _valid_simd_clauses` if `'simd' in self.combined.expr` and `self.combined` is one of: `OmpForSimd`, `OmpMaskedTaskLoop`. Can other classes have `'simd' in self.combined.expr`?

pyccel/parser/syntax/openmp.py
```diff
@@ -31,160 +35,686 @@ def __init__(self, **kwargs):
...
+
+        super().__init__(**kwargs)
 
     @property
     def expr(self):
         if DEBUG:
-            print("> OmpLoopConstruct: expr")
+            print("> OmpSingleConstruct: expr")
+
+        return self._expr
```
EmilyBourne, line 210: Is there any way to manage this through inheritance? E.g.:
```
@property
def expr(self):
        if DEBUG:
            print("> {}: expr".format(type(self))))

        return self._expr
```

pyccel/parser/syntax/openmp.py
```diff
@@ -31,160 +35,686 @@ def __init__(self, **kwargs):
...
+            print("> OmpCriticalConstruct: expr")
+
+        return self._expr
+
+class OmpSimdConstruct(BasicStmt):
+    """Class representing a Simd construct."""
+    def __init__(self, **kwargs):
+        """
+        """
+        self.name = kwargs.pop('name')
```
EmilyBourne, line 241: Name is not private here, or in other classes. This means that it can be modified outside the class easily. Is this intentional?

pyccel/parser/syntax/openmp.py
```diff
@@ -31,160 +35,686 @@ def __init__(self, **kwargs):
...
+        self.clauses = kwargs.pop('clauses')
```
EmilyBourne, line 242: See above this can be modified outside the class. Is this intentional?

pyccel/parser/syntax/openmp.py
```diff
@@ -31,160 +35,686 @@ def __init__(self, **kwargs):
...
+        if DEBUG:
+            print("> OmpSectionConstruct: expr")
+
+        return self._expr
+
+class OmpDistributeConstruct(BasicStmt):
+    """Class representing a distribute construct."""
+    def __init__(self, **kwargs):
+        """
+        """
```
EmilyBourne, line 350: `__init__` functions don't need docstrings. The class docstring fulfils this role

pyccel/parser/syntax/openmp.py
```diff
@@ -31,160 +35,686 @@ def __init__(self, **kwargs):
...
+        self.name = kwargs.pop('name')
+        self.clauses = kwargs.pop('clauses')
+
+        txt = self.name
+        txt += check_get_clauses(self, _valid_sections_clauses, self.clauses)
+
+        self._expr = OMP_Sections_Construct(txt)
+
+        super().__init__(**kwargs)
```
EmilyBourne, line 319: There is still a lot of code duplication in the `__init__` functions is there any way to create properties/superclasses to group the cases and reduce this duplication?

pyccel/parser/syntax/openmp.py
```diff
@@ -31,160 +35,686 @@ def __init__(self, **kwargs):
...
+
+        com = None
+        if isinstance(self.combined, OmpDistributeCombined):
+            com = self.combined.expr
+            _valid_clauses = _valid_clauses + _valid_Distribute_clauses
+            if 'simd' in self.combined.expr:
+                _valid_clauses = _valid_clauses + _valid_simd_clauses
+            if 'parallel' in self.combined.expr:
+                _valid_clauses = _valid_clauses + _valid_parallel_clauses
+                _valid_clauses = _valid_clauses + _valid_loop_clauses
```
EmilyBourne, line 540: Do `combined` objects have `valid_clauses` ? Couldn't you do:
```
_valid_clauses += self.combined._valid_clauses
```
?

EmilyBourne (changes requested):

pyccel/parser/syntax/openmp.py
nhamidn: yes, but not all, only construct that can be combined and can have `simd` as a combined construct, so not all other classes can have it in `self.combined.expr`

pyccel/parser/syntax/openmp.py
nhamidn: no it's not intentional, im gonna make it private

pyccel/parser/syntax/openmp.py
EmilyBourne: In that case why do you need all the `isinstance` checks?
Why not just do:
```
_valid_clauses = self._valid_clauses
if 'simd' in self.combined.expr:
    _valid_clauses += _valid_simd_clauses
if 'taskloop' in self.combined.expr:
    _valid_clauses += _valid_taskloop_clauses
```
?

pyccel/parser/syntax/openmp.py
EmilyBourne: Ok. If I were you I would fix this last. If you can find a smart solution for the inheritance to avoid code duplication then it should only need doing once instead of in all these different classes

pyccel/parser/syntax/openmp.py
nhamidn: but `self.combined` can be `None`, then if I check for `if 'simd' in self.combined.expr:` I will get a python error object has no attribute 'expr'

EmilyBourne: Option 1:
Create a class `OmpParallelCombinedConstruct` for objects which have a `combined` property
Option 2:
```
_valid_clauses = self._valid_clauses
if 'simd' in self.combined:
    if 'simd' in self.combined.expr:
        _valid_clauses += _valid_simd_clauses
    if 'taskloop' in self.combined.expr:
        _valid_clauses += _valid_taskloop_clauses
```

nhamidn: im not sure i understand the first option as i know in `textx` every class represent a rule, if i work with option 2 i need to add more if to check `if 'for' in self.combined.expr` because when i do `if isinstance(self.combined, OmpForSimd):` I'm sure here that `self.combined.expr` contains `for` and the same for the others

EmilyBourne: Surely in `textx` every _imported_ class represents a rule? But the imported classes can still inherit from other classes that `textx` doesn't need to know about correctly?

EmilyBourne: As I understand it (bearing in mind that I have never worked with `textx`), `textx` constructs a metamodel from the file `pyccel/parser/grammar/openmp.tx`. This file tells `textx` which objects need to be created depending on which words are encountered. It doesn't tell `textx` anything about the classes that these objects might inherit from

EmilyBourne: >if i work with option 2 i need to add more if to check if 'for' in self.combined.expr because when i do if isinstance(self.combined, OmpForSimd): I'm sure here that self.combined.expr contains for and the same for the others

Hmm, maybe if combined objects also had `_valid_clauses`?

pyccel/parser/syntax/openmp.py
nhamidn: done

pyccel/ast/omp.py
```diff
@@ -0,0 +1,153 @@
...
+
+    @property
+    def is_multiline(self):
+        """Used to check if the construct needs brackets."""
+        return self._is_multiline
+
+    @property
+    def name(self):
+        """Name of the construct."""
+        return ''
```
EmilyBourne, line 47: Is this code ever used? If the class needs to override this property then it is probably better to leave this method out then an error is raised if it has been forgotten

EmilyBourne: Never mind, I see it is used for several classes from `OMP_Task_Construct` onwards

pyccel/parser/syntax/openmp.py
```diff
@@ -11,396 +11,962 @@
...
+    def __init__(self, **kwargs):
+        self.final = kwargs.pop('final')
+
+        super().__init__(**kwargs)
 
     @property
     def expr(self):
         if DEBUG:
-            print("> OmpEndClause: expr")
+            print("> {}: expr".format(type(self).__name__))
```
EmilyBourne, line 243: If you do not expect anything to inherit from this class then `print("> OmpEndClause: expr")` is sufficient. I suggested `print("> {}: expr".format(type(self).__name__))` so that you could use inheritance to avoid writing it for every class. If `expr` needs to be defined anyway then the old code will run faster

pyccel/parser/syntax/openmp.py
```diff
@@ -11,396 +11,962 @@
...
+    def expr(self):
+        if DEBUG:
+            print("> {}: expr".format(type(self).__name__))
+
+        return self._expr
+
+class OmpFinal(BasicStmt):
+    """Class representing a final clause"""
+    def __init__(self, **kwargs):
+        self.final = kwargs.pop('final')
```
EmilyBourne, line 236: `final` is public. Is this intentional?

EmilyBourne:
This is starting to look very clean :+1:
Good work :smiley:

pyccel/parser/syntax/openmp.py
nhamidn: it's used in the `expr` however im gonna move the calculation for the clause to the `__init__` function too

pyccel/parser/syntax/openmp.py
```diff
@@ -11,396 +11,962 @@
...
+        return self._expr
+
+def check_get_clauses(name, valid_clauses, clauses, combined = None):
+    """
+    Function to check if the clauses are correct for a given Construct.
+    """
+    txt = ''
+    for clause in clauses:
+        if isinstance(clause, valid_clauses):
+            if isinstance(clause, OmpCopyin) and isinstance(combined, OmpTargetParallel):
```
mostafaMamoni, line 79: can these checks be in one if block since the next statement is just an else?

pyccel/parser/syntax/openmp.py
```diff
@@ -11,396 +11,962 @@
...
+    """Class representing a For loop construct."""
+    def __init__(self, **kwargs):
+        super().__init__(OMP_For_Loop, _valid_loop_clauses, **kwargs)
 
-    @property
-    def expr(self):
-        if DEBUG:
-            print("> OmpParallelConstruct: expr")
+class OmpTaskLoopConstruct(OmpConstruct):
+    """Class representing a taskloop construct."""
```
mostafaMamoni, line 123: typo: taskloop -> TaskLoop

pyccel/parser/syntax/openmp.py
```diff
@@ -11,396 +11,962 @@
...
-        _valid_clauses = (OmpNumThread, \
-                         OmpDefault, \
-                         OmpPrivate, \
-                         OmpShared, \
-                         OmpFirstPrivate, \
-                         OmpCopyin, \
-                         OmpReduction, \
-                         OmpProcBind)
+class OmpSingleConstruct(OmpConstruct):
+    """Class representing a single construct."""
```
mostafaMamoni, line 133: typo: single -> Single, i think you should delete the letter 'a' in many cases because it indicates the literal meaning of the construct name and it is obvious in this case,

pyccel/parser/syntax/openmp.py
```diff
@@ -11,396 +11,962 @@
...
-    def expr(self):
-        if DEBUG:
-            print("> OmpParallelConstruct: expr")
+class OmpTaskLoopConstruct(OmpConstruct):
+    """Class representing a taskloop construct."""
+    def __init__(self, **kwargs):
+        super().__init__(OmpAnnotatedComment, (_valid_taskloop_clauses + (OmpinReduction,)), **kwargs)
+
+class OmpTaskConstruct(OmpConstruct):
+    """Class representing a Task Construct """
```
mostafaMamoni, line 128: typo: Construct -> construct.

pyccel/parser/syntax/openmp.py
```diff
@@ -11,396 +11,962 @@
...
-        txt = ''
-        for clause in self.clauses:
-            if isinstance(clause, _valid_clauses):
-                txt = '{0} {1}'.format(txt, clause.expr)
-            else:
-                raise TypeError('Wrong clause for OmpLoopConstruct. Given : ', \
-                                type(clause))
-        return OMP_For_Loop(txt)
+class OmpDistributeConstruct(OmpConstruct):
+    """Class representing a distribute construct."""
```
mostafaMamoni, line 168: typo: distribute -> Distribute

pyccel/parser/syntax/openmp.py
```diff
@@ -11,396 +11,962 @@
...
         # TODO check if variable exist in namespace
         if DEBUG:
-            print("> OmpNumThread: expr")
+            print("> {}: expr".format(type(self).__name__))
 
         thread = self.thread
         return 'num_threads({})'.format(thread)
 
+class OmpNumTeams(BasicStmt):
+    """Class representing a num_teams clause."""
```
mostafaMamoni, line 265: num_teams -> NumTeams.

pyccel/parser/syntax/openmp.py
```diff
@@ -11,396 +11,962 @@
...
+    def expr(self):
+        # TODO check if variable exist in namespace
+        if DEBUG:
+            print("> {}: expr".format(type(self).__name__))
+
+        limit = self.limit
+        return 'thread_limit({})'.format(limit)
+
+class OmpNumTasks(BasicStmt):
+    """Class representing a num_tasks clause."""
```
mostafaMamoni, line 297: num_tasks -> NumTasks.

pyccel/parser/syntax/openmp.py
```diff
@@ -11,396 +11,962 @@
...
+
+    @property
+    def expr(self):
+        if DEBUG:
+            print("> {}: expr".format(type(self).__name__))
+
+        return '{}'.format(self.name)
+
+class OmpForSimd(BasicStmt):
+    """Class representing a combined comstruct."""
```
mostafaMamoni, line 717: typo: comstruct -> construct, there many cases like this one.

mostafaMamoni (changes requested):

pyccel/parser/syntax/openmp.py
EmilyBourne: I disagree. `Class representing Single construct.` is not correct English.

Maybe `a` should be changed to `the` though, `a` implies "one of several" I don't think there are multiple types of single constructs?

pyccel/parser/syntax/openmp.py
nhamidn: there is no `NumTeams` in openmp, this class is representing a `num_teams` clause

pyccel/parser/syntax/openmp.py
nhamidn: there is no `NumTasks ` in openmp, this class is representing a `num_tasks` clause

pyccel/parser/syntax/openmp.py
mostafaMamoni: actually, i removed 'replace it with `the`', but i considered `Single` as a proper name.

pyccel/parser/syntax/openmp.py
mostafaMamoni: i think that `num_teams` for code writing and `NumTeams` for the docstring.

nhamidn: `num_theam` is the name of the clause in openmp, there is nothing named `NumTeams` in openmp, and we don't use this name in pyccel, so why would we use it ?

pyccel/parser/syntax/openmp.py
EmilyBourne: If you count `Single` as a proper name then it would have to be:
`Class representing Single`
If you keep "construct" then it needs an article, so:
`Class representing the Single construct`
I find the second option clearer

pyccel/parser/syntax/openmp.py
EmilyBourne: I agree, it would be neater to avoid the duplicate code with:
```
if isinstance(clause, valid_clauses) and \
   not (isinstance(clause, OmpCopyin) and isinstance(combined, OmpTargetParallel)):
```

EmilyBourne (changes requested):
One small typo and a few problems that codacy has flagged then I am ready to accept :+1:

mostafaMamoni (approved):

EmilyBourne (approved):

EmilyBourne:
There is an openmp test which is xpassing could you correct the mark please

nhamidn:
> There is an openmp test which is xpassing could you correct the mark please

this test is passing on Linux and windows but it's failing on Macos

EmilyBourne:
> > There is an openmp test which is xpassing could you correct the mark please
> 
> this test is passing on Linux and windows but it's failing on Macos

Ah ok. The xfail mark can be passed a condition (see [here](https://docs.pytest.org/en/latest/reference.html#pytest.mark.xfail) ) to indicate this. Do you know what is causing this problem? It's rare for the problem to only exist on mac

nhamidn:
> here

i really dont know, i have macos and the test is passing on my computer for fortran

EmilyBourne:
> > here
> 
> i really dont know, i have macos and the test is passing on my computer for fortran

That's very strange. So then we need to fix #516 and #517 to know what the problem is?

nhamidn:
> > > here
> > 
> > 
> > i really dont know, i have macos and the test is passing on my computer for fortran
> 
> That's very strange. So then we need to fix #516 and #517 to know what the problem is?

yes, I think it's related to the OpenMP library linking because the compilation fails

nhamidn:
> > > There is an openmp test which is xpassing could you correct the mark please
> > 
> > 
> > this test is passing on Linux and windows but it's failing on Macos
> 
> Ah ok. The xfail mark can be passed a condition (see [here](https://docs.pytest.org/en/latest/reference.html#pytest.mark.xfail) ) to indicate this. Do you know what is causing this problem? It's rare for the problem to only exist on mac

now the test failed on Linux and windows, this is strange I'm gonna mark it as xfail in all platforms

EmilyBourne:
The xpass that I spotted is in the C tests. It is still marked as xpass after your last push actually. So maybe there are a few odd things here

nhamidn:
> The xpass that I spotted is in the C tests. It is still marked as xpass after your last push actually. So maybe there are a few odd things here

aah okay, I'm testing right now

EmilyBourne:
You might want to also modify the windows tests while you are testing. They are not with the rest unfortunately (probably due to issue #517) but are [here](https://github.com/pyccel/pyccel/blob/master/.github/workflows/Github_pytest.yml#L57-L64)

yguclu (changes requested):
Excellent job @nhamidn!

I would love to see some more challenging examples, with timing results, but this can be done in a separate PR.

pyccel/codegen/printing/fcode.py
```diff
@@ -166,6 +167,8 @@
     'complex' : PythonComplex
 }
 
+omploop_dict = {'for' : 'do'}
+
```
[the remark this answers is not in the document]
nhamidn: aah yes, i forgot to delete it.

yguclu (approved):

PR #597, opened 2020-12-16
Fix MPI tests on Windows

krystophny:
- Fix MPI installation problems on Windows;
- Ensure that errors are correctly raised when running pyccel in parallel.


EmilyBourne:
For faster results if you're just playing around for now, it might be a good idea to temporarily comment these lines:
- **Non-windows tests**
    https://github.com/pyccel/pyccel/blob/master/.github/workflows/Github_pytest.yml#L13-L71
    https://github.com/pyccel/pyccel/blob/master/.github/workflows/Github_pytest.yml#L110-L125
- **Non-MPI tests**
    https://github.com/pyccel/pyccel/blob/master/.github/workflows/Github_pytest.yml#L92-L103

.github/actions/windows_install/action.yml
```diff
@@ -37,31 +38,34 @@ runs:
...
+        echo "MS MPI runtime and SDK installed"
+      shell: powershell
+    - name: Setup MPI
+      run: |
+        #refreshenv
+        $MSMPI_INC="C:\\Program Files (x86)\\Microsoft SDKs\\MPI\\Include\\"
+        $MSMPI_LIB64="C:\\Program Files (x86)\\Microsoft SDKs\\MPI\\Lib\\x64\\"
+        $PATH="${Env:Path};C:\\Program Files\\Microsoft MPI\\Bin"
+        [System.Environment]::SetEnvironmentVariable("Path",$PATH)
+        echo "Path=${PATH}" >> $GITHUB_ENV
```
EmilyBourne, line 53: This is the line that is supposed to ensure that the system variable is set between shells. Maybe it should be:
```
echo "Env:Path=${PATH}" >> $GITHUB_ENV
```
?

krystophny: Yes, this is a bit tricky. I've now switched to another variant according to https://stackoverflow.com/questions/60169752/how-to-update-the-path-in-a-github-action-workflow-file-for-a-windows-latest-hos

EmilyBourne: Beware that question was asked in 2011 and they seem to have changed method since then. They detected a security vulnerability in the envionment. I'm not sure when exactly but this link:
https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
makes me think it must have been very recent.

The doc is here:
https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#environment-files

EmilyBourne: Never mind, I just realised they say that in one of the answers ^^

EmilyBourne:
You probably need [this commit](https://github.com/pyccel/pyccel/commit/23eba2d945553dc764f59b0fa242b60bbdf7cda9) to get useful debugging info. But I wonder if the remaining problems are related to issue #517. In theory the generated shell should also be a powershell and should inherit the environment but I'm not sure how well that works in practice

.github/actions/windows_install/action.yml
krystophny: Hm... Now the installation and mpiexec is running, but I still get an error on the tests. Locally on my machine everything works, so could be a problem or a restriction with github actions related to network?

EmilyBourne: Do you mind if I add a couple of prints to try and see what actually failed?

I'm definitely worried that it's just that the spawned process isn't in the right shell though :/

krystophny: Of course, please go ahead. I guess I leave this whole PR now to you again, since I'm at the end of my capabilities ^^.

EmilyBourne: The error is `(<class 'KeyError'>, KeyError('MSMPI_INC'), <traceback object at 0x00000212F4802148>)`
so I think it's just a question of keeping `$MSMPI_INC` and `$MSMPI_LIB64` in the path

.github/workflows/Github_pytest.yml
```diff
@@ -101,11 +98,11 @@ jobs:
...
-      #- name: Parallel tests
-      #  run: |
-      #    mpiexec -n 4 python -m pytest epyccel/test_parallel_epyccel.py -v -m parallel -rx
-      #    mpiexec -n 4 python -m pytest epyccel -v -m parallel -rx
-      #  working-directory: ./tests
+      - name: Parallel tests
+        run: |
+          mpiexec -n 4 python -m pytest epyccel/test_parallel_epyccel.py -v -m parallel -rx
+          # mpiexec -n 4 python -m pytest epyccel -v -m parallel -rx
+        working-directory: ./tests
```
yguclu, line 105: Instead of separating serial and parallel tests in a manual fashion, could we not use `./.github/actions/pytest_run`?

pyccel/epyccel.py
```diff
@@ -240,8 +240,8 @@ def epyccel( python_function_or_module, **kwargs ):
                 fun_name = python_function_or_module.__name__ if fun else None
                 success  = True
             # error handling carried out after broadcast to prevent deadlocks
-            except: # pylint: disable=bare-except
-                exc_info = sys.exc_info()
+            except BaseException as e: # pylint: disable=broad-except
+                exc_info = e
                 success  = False
```
yguclu, line 245: Are we sure that we get traceback object this way? See https://docs.python.org/3/library/sys.html#sys.exc_info

yguclu: OK, apparently the traceback is available directly through the exception, as `e.__traceback__` (see https://docs.python.org/3/reference/datamodel.html#traceback-objects).

.github/workflows/Github_pytest.yml
EmilyBourne: No. The pytest action uses bash so we can't use it for Windows

pyccel/epyccel.py
EmilyBourne: Yes. The traceback is available. But if we do `raise sys.exc_info()` then we get a `TypeError` because `exc_info` doesn't inherit from `BaseException`

.github/workflows/Github_pytest.yml
yguclu: I see. Can you please add a comment to issue #517 about this?

pyccel/epyccel.py
yguclu: Perfect, thanks!

.github/actions/windows_install/action.yml
yguclu: So how was this problem solved?

EmilyBourne: The problem was solved by using the command discussed above to keep `$MSMPI_INC` and `MSMPI_LIB64` in the environment variables

.github/workflows/Github_pytest.yml
EmilyBourne: done

yguclu (approved):

PR #598, opened 2020-12-16
Create missing gfortran symbolic link on macOS

EmilyBourne:
Further, the `--oversubscribe` flag is passed to `mpirun` on the macOS platform.

### DIFF
--- a/.github/actions/macos_install/action.yml
+++ b/.github/actions/macos_install/action.yml
@@ -5,8 +5,8 @@ runs:
   steps:
     - name: Install MPI, OpenMP
       run: |
-        brew install gfortran
         brew install open-mpi
         brew install libomp
+        ln -s gfortran-8 gfortran
         echo "MPI_OPTS='--oversubscribe'" >> $GITHUB_ENV
       shell: bash
--- a/.github/actions/macos_install/action.yml
+++ b/.github/actions/macos_install/action.yml
@@ -8,5 +8,5 @@ runs:
         brew install open-mpi
         brew install libomp
         ln -s /usr/local/bin/gfortran-8 bin/gfortran
-        echo "MPI_OPTS='--oversubscribe'" >> $GITHUB_ENV
+        echo "MPI_OPTS=--oversubscribe" >> $GITHUB_ENV
       shell: bash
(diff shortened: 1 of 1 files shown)


.github/actions/macos_install/action.yml
```diff
@@ -5,8 +5,8 @@ runs:
   steps:
     - name: Install MPI, OpenMP
       run: |
-        brew install gfortran
         brew install open-mpi
         brew install libomp
+        ln -s gfortran-8 gfortran
```
yguclu, line 10: I think you're missing the correct location of the `gfortran-8` executable, and you are placing the symbolic link in a directory that is not in `PATH`.

yguclu: Maybe you are confusing this command with the `alias` command?

EmilyBourne: you are right. This is a problem as the tests run from a different folder. I have been trying to modify the `PATH` on the branch `test_mac_authors`. So far no luck. Do you think I should be creating the symbolic link in a different directory?

yguclu: Try this:
```bash
 ln -s /usr/local/gfortran-8 /usr/local/gfortran
```

EmilyBourne: `ln: /usr/local/gfortran: Permission denied`
I also tried:
```
mkdir bin
        ln -s gfortran-8 bin/gfortran
        echo "PATH=$(pwd)/bin:${PATH}" >> $GITHUB_ENV
        echo "MPI_OPTS='--oversubscribe'" >> $GITHUB_ENV
```
The ln call seems to work but I still get `/bin/sh: gfortran: command not found` probably due to the process spawning a new shell (see issue #517 )

.github/actions/macos_install/action.yml
```diff
@@ -7,4 +7,6 @@ runs:
       run: |
         brew install open-mpi
         brew install libomp
+        ln -s /usr/local/bin/gfortran-8 bin/gfortran
```
yguclu, line 10: Isn't the directory `/usr/local/bin/` in the search PATH? If that's the case I suggest you create the `gfortran` symbolic link in the same directory.

 On macOS `/usr/local/` is essentially reserved to Homebrew, which `/bin/` is a system directory.

EmilyBourne: Yes, this was just a bad copy/paste

yguclu (approved):

PR #599, opened 2020-12-17
Handle Duplicated header of a function

### LINKED ISSUE #567: Duplicated header of a function
**Describe the bug**
If the same header is written twice for the same function, two functions with the same signature are generated.

**To Reproduce**
```python
#$ header function f(int)
#$ header function f(int)
def f(a):
    return a
```

```c
#include <test.h>
#include <stdlib.h>



long f_00(long a)
{

    long Out_0001;

    Out_0001 = a;
    return Out_0001;
}

long f_01(long a)
{

    long Out_0001;

    Out_0001 = a;
    return Out_0001;
}
```

**Expected behavior**
Only one function should be generated.

**Language**
Fortran, C

rmahjoubi:
Close #567:
- If a function's header is duplicated, only one header is inserted to the scope, and a warning is raised;
- If a function is decorated with multiple types decorator that specify the same signature, only one is kept in the function, and a warning is raised;
- Add tests.


nhamidn:
I think it will be good if you raise a warning for the user, to inform him that he used to write two duplicated headers/decorators

nhamidn (approved):

pyccel/parser/semantic.py
```diff
@@ -2304,12 +2309,23 @@ def _visit_FunctionDef(self, expr, **settings):
...
 
         if cls_name:
-            header += self.get_header(cls_name +'.'+ name)
+            tmp_headers = self.get_header(cls_name + '.' + name)
+            for header in tmp_headers:
+                if all(header.dtypes != hd.dtypes for hd in headers):
+                    headers.append(header)
+                else:
+                    errors.report(DUPLICATED_SIGNATURE, symbol=header,
+                            severity='warning')
```
EmilyBourne, line 2318: Is there a reason why this bit needs to be duplicated in the if/else block? I can't see a difference between the two

pyccel/parser/syntactic.py
```diff
@@ -739,7 +739,12 @@ def fill_types(ls):
                 header = hdr_parse(stmts=txt)
                 if name in self.namespace.static_functions:
                     header = header.to_static()
-                headers += [header]
+                if all(header.dtypes != hd.dtypes for hd in headers):
+                    headers += [header]
+                else:
+                    errors.report(DUPLICATED_SIGNATURE, symbol=header,
+                        bounding_box=(stmt.lineno, stmt.col_offset),
+                        severity='warning')
```
EmilyBourne, line 747: This code block appears 4 times in total. Is there no way to treat them all in the same place?

EmilyBourne (changes requested):

pyccel/parser/syntactic.py
rmahjoubi: We can keep the duplicated headers and decorators, until we gather them in `visit_functiondef` in semantics.

saidctb (approved):

pyccel/parser/syntactic.py
EmilyBourne: That sounds better to me. It seems strange to not do them all at once

yguclu: So this will be handled in a separate PR? Do we have an open issue for it?

pyccel/parser/semantic.py
```diff
@@ -2370,27 +2380,27 @@ def _visit_FunctionDef(self, expr, **settings):
                     errors.report(msg, symbol=expr.arguments, severity='fatal')
 
         interfaces = []
-        if len(header) == 0:
-            # check if a header is imported from a header file
+        if len(headers) == 0:
+            # check if a headers is imported from a headers file
             # TODO improve in the case of multiple headers ( interface )
```
yguclu, line 2385: Maybe the singular `header` should be used in this comment?

pyccel/parser/semantic.py
```diff
@@ -2370,27 +2380,27 @@ def _visit_FunctionDef(self, expr, **settings):
...
 
             errors.report(FUNCTION_TYPE_EXPECTED, symbol=name,
                    bounding_box=(self._current_fst_node.lineno, self._current_fst_node.col_offset),
                    severity='error', blocker=self.blocking)
 
-        # we construct a FunctionDef from its header
-        for hd in header:
+        # we construct a FunctionDef from its headers
+        for hd in headers:
             interfaces += hd.create_definition(templates)
```
yguclu, line 2400: The comment should probably be updated: `# We construct a FunctionDef from each function header`?

EmilyBourne (approved):

pyccel/parser/syntactic.py
rmahjoubi: I now raise the error only one time

yguclu (approved):

Issue #596, opened 2020-12-16
Operators and Literals are missing `__str__` method


EmilyBourne:
**Describe the bug**
Operators and Literals do not have a `__str__` method which can make error messages difficult to read

**To Reproduce**
For example, if we look at this example:
```
import numpy as np
from pyccel.decorators import types

@types('float[:]')
def test(p):
    q = p[3:] + p[2:]
    return np.sum(q)
```
pyccel correctly generates the error:
```
 |fatal [semantic]: tmp.py [6,4]| operands could not be broadcast together with shapes (PyccelMinus(Shape(p,0), LiteralInteger(3)),) (PyccelMinus(Shape(p,0), LiteralInteger(2)),) (PyccelAdd(p[LiteralInteger(3) : ], p[LiteralInteger(2) : ]))
```
but this error is very hard to understand.

**Expected behavior**
It should instead be printed as:
```
|fatal [semantic]: tmp.py [6,4]| operands could not be broadcast together with shapes (Shape(p,0) - 3,) (Shape(p,0) - 2,) (p[3 : ] + p[2 : ])
```

**Additional context**
It should be possible to use the inheritance to avoid too much code duplication here.

PR #600, opened 2020-12-17
Add printing methods to operators and literals

EmilyBourne:
Add printing methods to operators and literals. This allows for clean printing of error messages and fixes issue #596 

**Commit Summary**
- Change inheritance order in Literals so magic methods defined in `Literal` class are correctly inherited
- Define `__repr__` method in `Literal`
- Define `_sympystr` method in `Literal` (necessary to correctly print errors for sympy objects, e.g. `Indexed` which is used at the syntactic step)
- Define `__repr__` methods in ast.operators
- Correct error leading to broadcast error being reported twice


pyccel/ast/operators.py
```diff
@@ -228,6 +234,9 @@ class PyccelUnarySub(PyccelUnary):
         The argument passed to the operator
     """
 
+    def __repr__(self):
+        return '+{}'.format(repr(self.args[0]))
```
MenoIy, line 238: '-' or '+' ?

pyccel/ast/operators.py
```diff
@@ -212,6 +215,9 @@ def _handle_precedence(self, args):
         args = tuple(PyccelAssociativeParenthesis(a) if isinstance(a, PyccelUnary) else a for a in args)
         return args
 
+    def __repr__(self):
+        return '-{}'.format(repr(self.args[0]))
```
MenoIy, line 219: shouldn't this be '+' ?

pyccel/ast/operators.py
```diff
@@ -518,6 +545,12 @@ def __new__(cls, arg1, arg2):
         else:
             return PyccelArithmeticOperator.__new__(cls, arg1, arg2)
 
+    def __repr__(self):
+        return '{} - {}'.format(repr(self.args[0]), repr(self.args[1]))
+
+    def __str__(self):
+        return '{} - {}'.format(repr(self.args[0]), repr(self.args[1]))
+
```
MenoIy, line 553: you're using this one?

MenoIy (changes requested):

pyccel/ast/operators.py
EmilyBourne: Oops you're right

pyccel/ast/operators.py
EmilyBourne: No, you're right :+1:

MenoIy (approved):

pyccel/ast/operators.py
```diff
@@ -85,7 +85,7 @@ def broadcast(shape_1, shape_2):
         else:
             msg = 'operands could not be broadcast together with shapes {} {}'
             msg = msg.format(shape_1, shape_2)
-            errors.report(msg,severity='fatal')
+            raise PyccelSemanticError(msg)
```
yguclu, line 88: Why do you need to raise an Exception instead of using `Errors.report`?

EmilyBourne: The error is caught [here](https://github.com/pyccel/pyccel/blob/master/pyccel/parser/semantic.py#L1124) so that the context information can be added (symbol information). If we use `errors.report` to raise that error then we get the same message in the output twice.
E.g.:
```
pyccel:
 |fatal [semantic]: test.py| operands could not be broadcast together with shapes (Shape(p,0) - 3,) (Shape(p,0) - 2,)
 |fatal [semantic]: test.py [6,4]| operands could not be broadcast together with shapes (Shape(p,0) - 3,) (Shape(p,0) - 2,) (p[3 : ] + p[2 : ])
```

pyccel/ast/literals.py
```diff
@@ -44,8 +44,14 @@ def precision(self, precision):
...
 
+    def __repr__(self):
+        return repr(self.python_value)
+
+    def _sympystr(self, printer):
+        return printer.doprint(self.python_value)
+
 #------------------------------------------------------------------------------
-class LiteralTrue(sp_BooleanTrue, Literal):
+class LiteralTrue(Literal, sp_BooleanTrue):
```
saidctb, line 54: why do we still have `sp_BooleanTrue`

EmilyBourne: Because nobody said we could remove it? I think you are right. It should be removed. The only thing it might be doing is ensuring that `LiteralTrue` is a singleton, but we don't necessarily want that if we want booleans with different precision. I will remove it

pyccel/ast/operators.py
yguclu: Nice, thanks!

saidctb (approved):

yguclu (approved):

PR #602, opened 2020-12-18
Add tests of memory leaks to GitHub actions

saniQQ:
Add valgrind tests on Linux and macOS to check for memory leaks and bad memory access for C programs.
Valgrind does not exist on Windows, and there does not seem to be a command-line alternative.


pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -1,259 +1,259 @@
...
+
+    va_start(va, arr);
+    index = 0;
+    for (int32_t i = 0; i < arr.nd; i++)
+    {
+        index += va_arg(va, int32_t) * arr.strides[i];
+    }
+    va_end(va);
+    return (index);
+}
```
EmilyBourne, line 259: Did this file change?

EmilyBourne:
Please update the description to explain that valgrind doesn't exist on windows

tests/ndarrays/leaks_check.py
```diff
@@ -0,0 +1,31 @@
...
+    c = array([[1, 2, 3, 4, 5, 0], [0, 6, 7, 8, 9, 10]])
+    b = a# pylint: disable=unused-variable
+    b = a[1:]
+    b = c
+
+create_array()
+array_to_pointer()
+view_assign()
+pointer_to_pointer()
+pointer_reassign()
```
EmilyBourne, line 31: Maybe you could add an array created in a program instead of only in a function too?

pyccel/stdlib/ndarrays/ndarrays.c
saniQQ: didn't even touched it I don't know why it appears in the diff

EmilyBourne: It might be a line ending thing due to windows vs fortran. You can get rid of the changes by doing:
`git checkout master -- pyccel/stdlib/ndarrays/ndarrays.c`

EmilyBourne (approved):

EmilyBourne (approved):

yguclu (approved):

.github/actions/valgrind_run/action.yml
```diff
@@ -0,0 +1,11 @@
+name: 'Pyccel valgrind script'
+
+runs:
+  using: "composite"
+  steps:
+    - name: Test with valgrind for memory leaks
+      run: |
+        pyccel --language=c leaks_check.py
+        valgrind --leak-check=full --error-exitcode=1 ./prog_leaks_check
```
EmilyBourne, line 9: Shouldn't we be compiling in debug mode to use valgrind? I think that means the command should be:
```
pyccel --language=c --flags="-g" leaks_check.py
```

yguclu: You are right. Moreover, we should probably turn off any optimization by passing the `-O0` flag.

yguclu:
It appears that a memory leak is found on macOS 10.15 (Catalina), but this problem has been reported before (e.g. [here](https://stackoverflow.com/questions/58360093/how-to-install-valgrind-on-macos-catalina-10-15-with-homebrew)).

According to [Valgrind's website](https://valgrind.org/info/platforms.html) only macOS 10.10 and 10.11 are officially supported, so we might need to download a patch...

EmilyBourne (approved):

Issue #605, opened 2020-12-18
Python join causes crash


Beliavsky:
pyccel on 
```
print("boy") # works
print(" ".join(["boy","girl"])) # does not work

```
gives

Traceback (most recent call last):
  File "/usr/local/bin/pyccel", line 8, in <module>
    sys.exit(pyccel())
  File "/usr/local/lib/python3.8/dist-packages/pyccel/commands/console.py", line 235, in pyccel
    execute_pyccel(filename,
  File "/usr/local/lib/python3.8/dist-packages/pyccel/codegen/pipeline.py", line 134, in execute_pyccel
    parser.parse(verbose=verbose)
  File "/usr/local/lib/python3.8/dist-packages/pyccel/parser/parser.py", line 94, in parse
    parser = SyntaxParser(self._filename, **self._kwargs)
  File "/usr/local/lib/python3.8/dist-packages/pyccel/parser/syntactic.py", line 133, in __init__
    self.parse(verbose=True)
  File "/usr/local/lib/python3.8/dist-packages/pyccel/parser/syntactic.py", line 147, in parse
    ast = self._visit(self.fst)
  File "/usr/local/lib/python3.8/dist-packages/pyccel/parser/syntactic.py", line 170, in _visit
    result = getattr(self, syntax_method)(stmt)
  File "/usr/local/lib/python3.8/dist-packages/pyccel/parser/syntactic.py", line 187, in _visit_Module
    body          = [self._visit(v) for v in stmt.body]
  File "/usr/local/lib/python3.8/dist-packages/pyccel/parser/syntactic.py", line 187, in <listcomp>
    body          = [self._visit(v) for v in stmt.body]
  File "/usr/local/lib/python3.8/dist-packages/pyccel/parser/syntactic.py", line 170, in _visit
    result = getattr(self, syntax_method)(stmt)
  File "/usr/local/lib/python3.8/dist-packages/pyccel/parser/syntactic.py", line 262, in _visit_Expr
    return self._visit(stmt.value)
  File "/usr/local/lib/python3.8/dist-packages/pyccel/parser/syntactic.py", line 170, in _visit
    result = getattr(self, syntax_method)(stmt)
  File "/usr/local/lib/python3.8/dist-packages/pyccel/parser/syntactic.py", line 830, in _visit_Call
    args += self._visit(stmt.args)
  File "/usr/local/lib/python3.8/dist-packages/pyccel/parser/syntactic.py", line 170, in _visit
[...]

PR #614, opened 2020-12-19
Catch DottedVariable TypeError

EmilyBourne:
Raise an error when DottedVariable receives unexpected types. This is needed to handle the following:
```
a = " ".join(["boy","girl"])
```
See issue #605 

Raise an error when a class method is called on an object which has no class methods known to pyccel. This is needed to handle the following:
```
a = " "
b = a.join(["boy","girl"])
```


saidctb (commented):
instead of try except clause, can yo just do a check if the attribut exist ?, either in the DottedVariable contructor or in the _visit_attribute

EmilyBourne:
> instead of try except clause, can you just do a check if the attribute exist ?, either in the DottedVariable constructor or in the _visit_attribute, also shouldn't this be a sematics error and not a syntacitic one 

The error is caused at the syntactic stage by trying to make the lhs of a DottedVariable be a LiteralString. We probably get the same error if we do something like `1j.imag`. Either we need to test for the types accepted by DottedVariable (but that is just duplicating the type check so a try/except is more maintainable). Or we need to accept literals as the lhs of a DottedVariable and raise a similar error in the semantic step.

This seemed like the simplest option for now

saidctb:
> > instead of try except clause, can you just do a check if the attribute exist ?, either in the DottedVariable constructor or in the _visit_attribute, also shouldn't this be a sematics error and not a syntacitic one
> 
> The error is caused at the syntactic stage by trying to make the lhs of a DottedVariable be a LiteralString. We probably get the same error if we do something like `1j.imag`. Either we need to test for the types accepted by DottedVariable (but that is just duplicating the type check so a try/except is more maintainable). Or we need to accept literals as the lhs of a DottedVariable and raise a similar error in the semantic step.
> 
> This seemed like the simplest option for now
I think a DottedVariable should accept any literal, so the error will no longer be there, and raise the error in the sematics stage, this seems simple to me , I'm not even sure that we should check anything in the syntacitc stage the DottedVariable should accept anything

saidctb:
can you also treat this case ?
```python
a = "  "
b = a.join(["boy","girl"])
```

pyccel/parser/semantic.py
```diff
@@ -1006,10 +1006,17 @@ def _visit_Symbol(self, expr, **settings):
 
     def _visit_DottedVariable(self, expr, **settings):
 
+        if isinstance(expr.lhs, Literal):
```
saidctb, line 1009: probably , you don't need this anymore. after adding the second error report

EmilyBourne: I still need this as a `Literal` currently doesn't have a `cls_base`

saidctb: can you do this `if hasattr(first, 'cls_base') and  first.cls_base is None:`, with this you can have both cases

pyccel/parser/semantic.py
```diff
@@ -1060,6 +1060,12 @@ def _visit_DottedVariable(self, expr, **settings):
 
         # look for a class method
         if isinstance(expr.rhs, Application):
+            if not hasattr(first, 'cls_base') or first.cls_base is None:
+                errors.report('Class properties of object of type {} are not known'.format(first.dtype),
```
saidctb, line 1064: can you just raise and attribute error like python  `object has no attribute`

saidctb (approved):

yguclu (approved):

PR #615, opened 2020-12-19
Handle order property

### LINKED ISSUE #335: Order should be a member of PyccelAstNode

EmilyBourne:
Set PyccelAstNode order property. Use C ordering until printing.

**Commit Summary**

- Ensure order is set for all objects which inherit from PyccelAstNode with rank>1 (fixes #335);
- Preserve C indexing until printing stage;
- Use C ordering for tuples;
- Set PyccelAstNode properties for NumpyArray and NumpyFull.

### DIFF
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -123,7 +123,6 @@ class NumpyNewArray(PyccelAstNode):
 # TODO [YG, 18.02.2020]: accept Numpy array argument
 # TODO [YG, 18.02.2020]: use order='K' as default, like in numpy.array
 # TODO [YG, 22.05.2020]: move dtype & prec processing to __init__
-# TODO [YG, 22.05.2020]: change properties to read _dtype, _prec, _rank, etc...
 class NumpyArray(Application, NumpyNewArray):
     """
     Represents a call to  numpy.array for code generation.
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -529,7 +529,7 @@ class NumpyFull(Application, NumpyNewArray):
 
         return Basic.__new__(cls, shape, dtype, order, precision, fill_value)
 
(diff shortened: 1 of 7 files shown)


saidctb (approved):

pyccel/ast/numpyext.py
```diff
@@ -583,6 +557,10 @@ def __new__(cls, shape, dtype='float', order='C'):
         order = cls._process_order(order)
 
         return Basic.__new__(cls, shape, dtype, order, precision)
+
+    def __init__(self, shape, dtype=None, order='C'):
+        NumpyFull.__init__(self, shape, None, dtype, order)
```
yguclu, line 562: Before this addition, I suppose that the `NumpyFull.__init__` was called from the constructors of `NumpyEmpty`, `NumpyZeros`, `NumpyOnes` with the wrong arguments... How come that this did not result in some errors?

EmilyBourne: `NumpyFull` didn't have an `__init__` function before. So it was the sympy constructor which was called with `*args, **kwargs`

yguclu: Right! Thanks

yguclu (approved):

PR #621, opened 2020-12-21
Fortran: print PythonLen as size or len depending on dtype

### LINKED ISSUE #619: len(string) not translated correctly
pyccel translates

```
n = len("foo")
print(n)

```
to

```
program prog_len
use ISO_C_BINDING
implicit none
integer(C_LONG_LONG) :: n
n = size('' // 'foo',0)
print *, n
end program prog_len
```
which should instead be
```
program prog_len
use ISO_C_BINDING
implicit none
integer(C_LONG_LONG) :: n
n = len("foo")
print *, n
end program prog_len
```

### LINKED ISSUE #620: Omit DIM argument to SIZE function in translated Fortran code when not needed
`n = len([2,5])`

is translated to

`n = size([2_C_LONG_LONG, 5_C_LONG_LONG],1)`

but simpler would be

`n = size([2_C_LONG_LONG, 5_C_LONG_LONG])`

For a rank-1 Fortran array, the only permitted value of the DIM argument to the SIZE intrinsic is 1.

EmilyBourne:
Fixes #619 and fixes #620

### DIFF
--- a/pyccel/codegen/printing/fcode.py
+++ b/pyccel/codegen/printing/fcode.py
@@ -603,7 +603,11 @@ class FCodePrinter(CodePrinter):
     def _print_PythonLen(self, expr):
         var = expr.arg
         idx = 1 if var.order == 'F' else var.rank
-        return 'size({},{})'.format(self._print(var), self._print(idx))
+        dtype = var.dtype
+        if dtype is NativeString():
+            return 'len({})'.format(self._print(var))
+        else:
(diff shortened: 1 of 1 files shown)


saidctb:
can you also fix  issue #620 , you just need to add the case where the rank == 1

saidctb (approved):

yguclu (approved):

Issue #603, opened 2020-12-18
pyccel says function missing type annotation, but function is annotated


Beliavsky:
For the code xfib.py

```
def Fibonacci(n: int) -> int:
    if n<=0:
        print("Incorrect input")
    # First Fibonacci number is 0
    elif n==1:
        return 0
    # Second Fibonacci number is 1
    elif n==2:
        return 1
    else:
        return Fibonacci(n-1)+Fibonacci(n-2)
 
print(Fibonacci(9))
```

`pyccel xfib.py `

gives

```
ERROR at annotation (semantic) stage
pyccel:
 |error [semantic]: xfib.py [1]| Function is missing a type annotation (Fibonacci)
 |fatal [semantic]: xfib.py [1]| Undefined variable (n)
```

EmilyBourne:
Hi,

Thank you for your interest in pyccel. Unfortunately, pyccel is unable to read python style type annotation. Instead, the type annotation is done with a comment header or a decorator. We are currently writing a quickstart file which will provide an explanation of how this is done, but in the meantime I suggest looking at the examples in the tests. 

For the comment header syntax see:
https://github.com/pyccel/pyccel/blob/master/tests/pyccel/scripts/hope_benchmarks/hope_fib.py
and for the decorator syntax see:
https://github.com/pyccel/pyccel/blob/master/tests/pyccel/scripts/hope_benchmarks_decorators/hope_fib.py

(Both these examples specify the return type, but this is only necessary for recursive functions)

saidctb:
Hi @Beliavsky,
you can use pyccel like this
```python
@types('int', results='int')
def fibonacci(n: int) -> int:
    if n<=0:
        print("Incorrect input")
    # First Fibonacci number is 0
    elif n==1:
        return 0
    # Second Fibonacci number is 1
    elif n==2:
        return 1
    else:
        return Fibonacci(n-1)+Fibonacci(n-2)
```
we plan to add the python annotations in the futur

Beliavsky:
> ```python
> @types('int', results='int')
> ```

Thank you. I am able to make it work. In your code, Fibonacci should be capitalized consistently.

yguclu:
I think it would be nice to support type hints, so I've created a separate feature request where we can have a discussion: #606.

Issue #476, opened 2020-10-12
Failed to convert the python code to C/Fortran: Function is missing a type annotation, and so on.


hongyi-zhao:
Hi,

The environment is Ubuntu 20.04 with python 3.8.3 managed by pyenv. I try to convert [this python script](https://github.com/hongyi-zhao/test/blob/master/mu.py) into C/Fortran code with the help of pyccel, but all my tries failed. See following for more info:

```
$ pyccel --language fortran --compiler gfortran -t mu.py 

ERROR at annotation (semantic) stage
pyccel: [semantic] 
 |error: mu.py [11]| Function is missing a type annotation (fermi)
 |fatal: mu.py [12,1]| Undefined variable (E)

$ pyccel --language fortran --compiler ifort -t mu.py 

ERROR at annotation (semantic) stage
pyccel: [semantic] 
 |error: mu.py [11]| Function is missing a type annotation (fermi)
 |fatal: mu.py [12,1]| Undefined variable (E)

$ pyccel --language c --compiler icc -t mu.py 

ERROR at annotation (semantic) stage
pyccel: [semantic] 
 |error: mu.py [11]| Function is missing a type annotation (fermi)
 |fatal: mu.py [12,1]| Undefined variable (E)
```
Any hints for this problem?

Regards,
HY

EmilyBourne:
Hi,

The issue seems to be that you have not told pyccel which types your function `fermi` expects as input. Unfortunately pyccel does not have an up to date documentation to point you to. The data types must be specified for each function argument. This can either be done using the pyccel decorator `types`:
```
from pyccel.decorators import types

@types('real[:]','real','real')
def fermi(E,mu,kT):
```
or using a comment:
```
#$ header function fermi(real[:], real, real)
def fermi(E,mu,kT):
```

I suggest that you look at the file https://github.com/pyccel/pyccel/blob/master/tests/pyccel/scripts/runtest_funcs.py for an example with decorators or https://github.com/pyccel/pyccel/blob/master/tests/codegen/fcode/scripts/functions.py for an example with comments

EmilyBourne:
I have had a quick look at your code, and it seems that you use a few functions that aren't implemented in pyccel:
- np.genfromtxt
- np.argmax
- np.arange (the python equivalent `range` is implemented and seems to be sufficient for your example)
- np.trapz
- np.column_stack
- np.savetxt

[...]

PR #622, opened 2020-12-21
Add quickstart.md file to documentation

mostafaMamoni:
Add quick-start guide to tutorial directory:

- Basic information on static (C/Fortran) vs. dynamically typed (Python) languages;
- Short explanation on Python objects vs. Python variables;
- How Pyccel works: type inference, assumptions/restrictions, type annotations;
- How to use Pyccel: installation, command line usage (pyccel command), interactive usage (epyccel function);
- Other features;
- Getting help.

A link to the quick-start file was added to the README file.


doc/quickstart.md
```diff
@@ -0,0 +1,196 @@
...
+- resides in memory and has a **reference count**.
+- is accessed through one or more Python variables.
+- is destroyed by the garbage collector when its reference count drops to zero.
+
+For more details about Python object, see [this](https://docs.python.org/3/tutorial/classes.html).
+
+ ###### Python variable:
+- is a reference to a Python object in memory.
+- is created with an assignment operation `x = expr`:
+  1. if the variable `x` already exists, the interpreted reduces the reference count of its object
```
EmilyBourne, line 26: typo:
`interpreted`->`interpreter`

doc/quickstart.md
```diff
@@ -0,0 +1,196 @@
...
+  2. a new variable `x` is created, which references the value of expr.
+can be destroyed with the command del `x`.
```
EmilyBourne, line 28: Missing `- ` ?

EmilyBourne:
Could you add more explanation about how to specify the types. As you can see from our issues (#603, #476) this is the usually the first problem that users encounter and is information that you **must** have in order to be able to use pyccel.

doc/quickstart.md
```diff
@@ -0,0 +1,202 @@
...
+
+ ## What is Pyccel?
+
+- static compiler for Python 3, using Fortran or C as backend language.
+- started as small open-source project in 2018 at IPP Garching.
+- public repository is now hosted on GitHub, freely available for download.
+
+## Python’s objects, variables, and garbage collection
+
+ Python is an **interpreter** language, **dynamically typed** and **garbage-collected**.
```
EmilyBourne, line 12: Typo `interpreter`->`interpreted`

interpreter is the object, interpreted is an adjective

nhamidn:
as a user i would like to know how to compile my python file using Pyccel and what are the different flags that i can use

mostafaMamoni:
> as a user i would like to know how to compile my python file using Pyccel and what are the different flags that i can use

you cannot compile a python file, you actually convert it to a C/Fortran file and then compile the generated code.
For your question i mentioned the language flag, i think its enough for a quickstart file, but i will mention the other flags by redirecting the user to pyccel usage.

doc/quickstart.md
```diff
@@ -0,0 +1,208 @@
...
+-   is created with an assignment operation `x = expr`:
+    -   if the variable `x` already exists, the interpreter reduces the reference count of its object
+    -   a new variable `x` is created, which references the value of expr.
+
+-   can be destroyed with the command del `x`.
+
+For more details about Python variable, see [this](https://www.w3schools.com/python/python_variables.asp).
+
+ ## Static typed languages
+
```
saniQQ, line 37: can you point the other differences between a statically-typed language and python 
so the user know why we need to add somethings like (garbage collector, and some error that we raise)

doc/quickstart.md
```diff
@@ -0,0 +1,208 @@
...
+-   After the installation, type `pyccel`, the usage should be shown, then all good.
+
+-   Create a Python file that contains simple lines of code to see what will happen.
+    -   To create the file `touch file_name.py`.
+    -   Use your favorite text editor to fill the file with some lines of code or just type `echo 'print("hello, world!\n")' > file_name.py` for a quick test, `cat file_name.py` to make sure that your lines in the file.
+    -   To generate the C/Fortran code form your Python code, type `pyccel file_name.py` or `pyccel file_name.py --language fortran` to generate Fortran code, and `pyccel file_name.py --language c` to generate C code.
+    -   No problems ?, So you should discover `__pyccel__` the directory that contains your generated code and some other stuff.
+
+E.g (using `@types` decorator and a recursive function with a typed return), To specify the types of the function arguments and its return, we need to import the `@types` decorator from pyccel.decorators (as you can see in the first line) and then specify the types for each function using this format:
```
saniQQ, line 55: (as you can see in the first line) 
I don't think that it is clear which line you are talking about

doc/quickstart.md
```diff
@@ -0,0 +1,210 @@
...
+
+For more details about Python object, see [this](https://docs.python.org/3/tutorial/classes.html).
+
+ ### Python variable
+
+-   is a reference to a Python object in memory.
+
+-   is created with an assignment operation `x = expr`:
+    -   if the variable `x` already exists, the interpreter reduces the reference count of its object
+    -   a new variable `x` is created, which references the value of expr.
```
EmilyBourne, line 30: Missing word ? `otherwise`

doc/quickstart.md
```diff
@@ -0,0 +1,210 @@
...
+
+-   the type of the variable can be changed in the run-time, because python is a dynamically typed language.
+
+-   can be destroyed with the command del `x`.
+
+For more details about Python variable, see [this](https://www.w3schools.com/python/python_variables.asp).
```
EmilyBourne, line 36: `about Python variable`->`about Python variables`

doc/quickstart.md
```diff
@@ -0,0 +1,210 @@
...
+
+ ## Static typed languages
+	
+A language is statically-typed if the type of a variable is known at compile-time instead of at run-time. Common examples of statically-
+typed languages include Java, C, C++, FORTRAN, Pascal and Scalam, in the other side, in python the type of a variable is known in the run-time, that's why we need to collect the garbag in the generated code, and raise some warnings/errors for the conflicts that can occur between dynamically typed languages (python) and static typed language(C/Fortran). See [this](https://en.wikipedia.org/wiki/Type_system#:~:text=In%20programming%20languages%2C%20a%20type,%2C%20expressions%2C%20functions%20or%20modules.) and [this](https://android.jlelse.eu/magic-lies-here-statically-typed-vs-dynamically-typed-languages-d151c7f95e2b#:~:text=Static%20typed%20languages,%2C%20FORTRAN%2C%20Pascal%20and%20Scala.) for more details.
```
EmilyBourne, line 41: Scalam = Scala?

EmilyBourne, line 41: `in the other side` -> `on the other hand`

EmilyBourne, line 41: `known in the run-time`->`known at run-time`

EmilyBourne, line 41: `garbag`->`garbage`

EmilyBourne, line 41: `static typed language`->`statically typed languages`

doc/quickstart.md
```diff
@@ -0,0 +1,210 @@
...
+
+ ## Installation (see [README](https://github.com/pyccel/pyccel/blob/master/README.rst) file)
+
+ ## Command line usage
+
+-   Open a terminal app, iterm or terminal for MacOs, terminal for Linux.
+
+-   After the installation, type `pyccel`, the usage should be shown, then all good.
```
EmilyBourne, line 49: `, then all good.` -> `. If this is the case then the installation has succeeded`

doc/quickstart.md
```diff
@@ -0,0 +1,210 @@
...
+
+-   Create a Python file that contains simple lines of code to see what will happen.
+    -   To create the file `touch file_name.py`.
+    -   Use your favorite text editor to fill the file with some lines of code or just type `echo 'print("hello, world!\n")' > file_name.py` for a quick test, `cat file_name.py` to make sure that your lines in the file.
+    -   To generate the C/Fortran code form your Python code, type `pyccel file_name.py` or `pyccel file_name.py --language fortran` to generate Fortran code, and `pyccel file_name.py --language c` to generate C code.
+    -   No problems ?, So you should discover `__pyccel__` the directory that contains your generated code and some other stuff.
```
EmilyBourne, line 55: `?,`->`?`

EmilyBourne, line 55: `So` is unnecessary

doc/quickstart.md
```diff
@@ -0,0 +1,210 @@
...
+
+E.g (using `@types` decorator and a recursive function with a typed return), To specify the types of the function arguments and its return, we need to import the `@types` decorator from pyccel.decorators (as you can see in `file_name.py` first line) and then specify the types for each function using this format:
+-   for the decorator `@types('1stArgType', '2ndArgType', 'NthArgType', results='return_type')` to declare arrays `@types('1stArgType[:]', '2ndArgType[:,:]', 'NthArgType[dimensions]', results='return_type')`, the expresion `[:]` means that the array has 1 dimension and 2 dimensions with `[:,:]` , So the number of dimensions of an array related with how many colons you put in the square brackets sparated with a comma, `arr[:,:,:]` means that the array `arr` has 3 dimensions and so on.
```
EmilyBourne, line 58: Typo `expresion`->`expression`

doc/quickstart.md
```diff
@@ -0,0 +1,210 @@
...
+-   for the function `def fun('1stArgType', '2ndArgType', 'NthArgType')`.
+   
+In `@types` decorator, pyccel support the following data types: real, double, float, pythonfloat, float32, float64, pythoncomplex, complex, complex64, complex128, int8, int16, int32, int64, int, pythonint, integer, bool, pythonbool.
```
EmilyBourne, line 61: `support`->`supports`

EmilyBourne (changes requested):

doc/quickstart.md
EmilyBourne: `at run-time` not `at the run-time`

doc/quickstart.md
```diff
@@ -0,0 +1,210 @@
...
+
+-   Open a terminal app, iterm or terminal for MacOs, terminal for Linux.
+
+-   After the installation, type `pyccel`, the usage should be shown, If this is the case then the installation has succeeded.
+
+-   Create a Python file that contains simple lines of code to see what will happen.
+    -   To create the file `touch file_name.py`.
+    -   Use your favorite text editor to fill the file with some lines of code or just type `echo 'print("hello, world!\n")' > file_name.py` for a quick test, `cat file_name.py` to make sure that your lines in the file.
+    -   To generate the C/Fortran code form your Python code, type `pyccel file_name.py` or `pyccel file_name.py --language fortran` to generate Fortran code, and `pyccel file_name.py --language c` to generate C code.
+    -   No problems ? you should discover `__pyccel__` the directory that contains your generated code and some other stuff.
```
EmilyBourne, line 55: Start of sentence should be capitalised `you`->`You`

doc/quickstart.md
```diff
@@ -0,0 +1,210 @@
...
+
+E.g (using `@types` decorator and a recursive function with a typed return), To specify the types of the function arguments and its return, we need to import the `@types` decorator from pyccel.decorators (as you can see in `file_name.py` first line) and then specify the types for each function using this format:
+-   for the decorator `@types('1stArgType', '2ndArgType', 'NthArgType', results='return_type')` to declare arrays `@types('1stArgType[:]', '2ndArgType[:,:]', 'NthArgType[dimensions]', results='return_type')`, the expression `[:]` means that the array has 1 dimension and 2 dimensions with `[:,:]` , So the number of dimensions of an array related with how many colons you put in the square brackets sparated with a comma, `arr[:,:,:]` means that the array `arr` has 3 dimensions and so on.
```
EmilyBourne, line 58: I would like to suggest something a bit clearer:
```
The syntax for the decorator is: `@types('1stArgType', '2ndArgType', 'NthArgType', results='return_type')`
or to declare arrays: `@types('1stArgType[:]', '2ndArgType[:,:]', 'NthArgType[dimensions]', results='return_type')`
The expression `[:]` means that the array has 1 dimension. 2 dimensions would be specified with `[:,:]`. The number of dimensions of an array is equal to the number of comma-separated colons in the square brackets. So `arr[:,:,:]` means that the array `arr` has 3 dimensions and so on.
```

mostafaMamoni: well structured, thanks

doc/quickstart.md
```diff
@@ -0,0 +1,210 @@
...
+-   The type of the variable can be changed in the run-time, because python is a dynamically typed language.
+
+-   Tan be destroyed with the command del `x`.
+
+For more details about Python variables, see [this](https://www.w3schools.com/python/python_variables.asp).
+
+ ## Static typed languages
+	
+A language is statically-typed if the type of a variable is known at compile-time instead of at run-time. Common examples of statically-
+typed languages include Java, C, C++, FORTRAN, Pascal and Scala, on the other hand, in python the type of a variable is known at the run-time, that's why we need to collect the garbage in the generated code, and raise some warnings/errors for the conflicts that can occur between dynamically typed languages (python) and statically typed languages(C/Fortran). See [this](https://en.wikipedia.org/wiki/Type_system#:~:text=In%20programming%20languages%2C%20a%20type,%2C%20expressions%2C%20functions%20or%20modules.) and [this](https://android.jlelse.eu/magic-lies-here-statically-typed-vs-dynamically-typed-languages-d151c7f95e2b#:~:text=Static%20typed%20languages,%2C%20FORTRAN%2C%20Pascal%20and%20Scala.) for more details.
```
EmilyBourne, line 41: `at run-time` not `at the run-time`

doc/quickstart.md
```diff
@@ -0,0 +1,209 @@
...
+
+ ### Python variable
+
+-   Is a reference to a Python object in memory.
+
+-   Is created with an assignment operation `x = expr`:
+    -   If the variable `x` already exists, the interpreter reduces the reference count of its object
+    -   Otherwise a new variable `x` is created, which references the value of expr.
+
+-   The type of the variable can be changed at the run-time, because python is a dynamically typed language.
```
EmilyBourne, line 32: `at the run-time`->`at run-time`

doc/quickstart.md
```diff
@@ -0,0 +1,209 @@
...
+
+For more details about Python object, see [this](https://docs.python.org/3/tutorial/classes.html).
+
+ ### Python variable
+
+-   Is a reference to a Python object in memory.
+
+-   Is created with an assignment operation `x = expr`:
+    -   If the variable `x` already exists, the interpreter reduces the reference count of its object
+    -   Otherwise a new variable `x` is created, which references the value of expr.
```
EmilyBourne, line 30: You talk about reducing the reference count. It therefore might also be good to talk about increasing the reference count too

doc/quickstart.md
```diff
@@ -0,0 +1,209 @@
...
+
+-   The type of the variable can be changed at the run-time, because python is a dynamically typed language.
+
+-   Tan be destroyed with the command del `x`.
+
+For more details about Python variables, see [this](https://www.w3schools.com/python/python_variables.asp).
+
+ ## Static typed languages
```
EmilyBourne, line 38: `Static typed languages` -> `Statically typed languages`

doc/quickstart.md
```diff
@@ -0,0 +1,209 @@
...
+	
+A language is statically-typed if the type of a variable is known at compile-time instead of the run-time. Common examples of statically-typed languages include Java, C, C++, FORTRAN, Pascal and Scala, on the other hand, in python the type of a variable is known at the run-time, that's why we need to collect the garbage in the generated code, and raise some warnings/errors for the conflicts that can occur between dynamically typed languages (python) and statically typed languages(C/Fortran). See [this](https://en.wikipedia.org/wiki/Type_system#:~:text=In%20programming%20languages%2C%20a%20type,%2C%20expressions%2C%20functions%20or%20modules.) and [this](https://android.jlelse.eu/magic-lies-here-statically-typed-vs-dynamically-typed-languages-d151c7f95e2b#:~:text=Static%20typed%20languages,%2C%20FORTRAN%2C%20Pascal%20and%20Scala.) for more details.
```
EmilyBourne, line 40: `instead of the run-time`->`instead of run-time`

EmilyBourne, line 40: `is known at the run-time`->`is known at run-time`

EmilyBourne (changes requested):

doc/quickstart.md
```diff
@@ -0,0 +1,210 @@
...
+For more details about Python object, see [this](https://docs.python.org/3/tutorial/classes.html).
+
+ ### Python variable
+
+-   Is a reference to a Python object in memory.
+
+-   Is created with an assignment operation `x = expr`:
+    -   If the variable `x` already exists, the interpreter reduces the reference count of its object
+    -   Otherwise a new variable `x` is created, which references the value of expr.
+    -   The reference count increased with an assignment operator, in argument passing or appending an object to a list.
```
EmilyBourne, line 31:
```
    -   If the variable `x` already exists, the interpreter reduces the reference count of its object
    -   Otherwise a new variable `x` is created
    -   The variable `x` is then modified to reference the object referenced by `expr` and the reference count of this object is increased
```

EmilyBourne (approved):

saniQQ (approved):

nhamidn (approved):

saidctb:
@yguclu  @EmilyBourne, should we add Python type hints to the documentation?

yguclu:
> @yguclu @EmilyBourne, should we add Python type hints to the documentation?

Yes we should.

Moreover, we should tell the user which approach we recommend for providing type information to function arguments. And which approaches may not be supported in the future.

mostafaMamoni:
Python type hints not working in C.

doc/quickstart.md
```diff
@@ -0,0 +1,221 @@
...
+
+E.g (using `@types` decorator/python type hints and a recursive function with a typed return), To specify the types of the function arguments and its return, we need to import the `@types` decorator from pyccel.decorators (as you can see in `file_name.py` first line) and then specify the types for each function argument in `@types` using the following:
+-   The syntax for the decorator is: `@types('1stArgType', '2ndArgType', 'NthArgType', results='return_type')`, or to declare arrays: `@types('1stArgType[:]', '2ndArgType[:,:]', 'NthArgType[dimensions]', results='return_type')`, The expression `[:]` means that the array has 1 dimension. 2 dimensions would be specified with `[:,:]`. The number of dimensions of an array is equal to the number of comma-separated colons in the square brackets. So `arr[:,:,:]` means that the array `arr` has 3 dimensions and so on.
+-   In the function we just use python syntax `def fun(1stArg, 2ndArg, NthArg)`.
+-   Also, You can specify the function arguments types using python type hints, `def fun(1stArg: 1stArgType, 2ndArg: 2ndArgType, NthArg: NthArgType) -> returnType`
```
saidctb, line 60: here `1stArgType` should be a string also like the `@types` arguments

doc/quickstart.md
```diff
@@ -0,0 +1,221 @@
...
+   
+In `@types` decorator, pyccel supports the following data types: real, double, float, pythonfloat, float32, float64, pythoncomplex, complex, complex64, complex128, int8, int16, int32, int64, int, pythonint, integer, bool, pythonbool.
```
saidctb, line 62: why do we have `pythonint`? int is sufficient the same for `integer` and `pythonbool`.

doc/quickstart.md
```diff
@@ -0,0 +1,221 @@
...
+  ```python
+  from pyccel.decorators import types
+
+  @types('int', results='int')
+  def factorial(n):
+	  if n == 0: return 1
+	  else : return n * factorial(n - 1)
+   ```
+  
+  Specifying the types using python test hints.
```
saidctb, line 78: using python type hints

doc/quickstart.md
```diff
@@ -0,0 +1,221 @@
...
+   }
+   /*........................................*/
+   ```
+
+   Fortran code:
+   
+   file_name.f90
+
+   ```Fortran
+   use ISO_C_BINDING
```
saidctb, line 117: where is the begining of the Module

doc/quickstart.md
```diff
@@ -0,0 +1,223 @@
...
+-   Is a reference to a Python object in memory.
+
+-   Is created with an assignment operation `x = expr`:
+    -   If the variable `x` already exists, the interpreter reduces the reference count of its object
+    -   Otherwise a new variable `x` is created, which references the value of expr.
+    -   The variable `x` is then modified to reference the object referenced by `expr` and the reference count of this object is increased
+
+-   The type of the variable can be changed at run-time, because python is a dynamically typed language.
+
+-   Tan be destroyed with the command del `x`.
```
EmilyBourne, line 35: Typo:
`Tan`->`Can`

doc/quickstart.md
```diff
@@ -0,0 +1,223 @@
...
+
+E.g (using `@types` decorator/python type hints and a recursive function with a typed return), To specify the types of the function arguments and its return, we need to import the `@types` decorator from pyccel.decorators (as you can see in `file_name.py` first line) and then specify the types for each function argument in `@types` using the following:
+-   The syntax for the decorator is: `@types('1stArgType', '2ndArgType', 'NthArgType', results='return_type')`, or to declare arrays: `@types('1stArgType[:]', '2ndArgType[:,:]', 'NthArgType[dimensions]', results='return_type')`, The expression `[:]` means that the array has 1 dimension. 2 dimensions would be specified with `[:,:]`. The number of dimensions of an array is equal to the number of comma-separated colons in the square brackets. So `arr[:,:,:]` means that the array `arr` has 3 dimensions and so on.
+-   In the function we just use python syntax `def fun(1stArg, 2ndArg, NthArg)`.
+-   Also, You can specify the function arguments types using python type hints, `def fun(1stArg: '1stArgType', 2ndArg: '2ndArgType', NthArg: 'NthArgType') -> 'returnType'`
```
EmilyBourne, line 60: It is important to specify how to describe array arguments with type hints.

```
You can also specify the function arguments types using python type hints, e.g. `...`
For arrays the syntax is the same as for the decorator and string type hints must be used to provide pyccel with information about the number of dimensions
```

EmilyBourne (changes requested):

doc/quickstart.md
```diff
@@ -0,0 +1,223 @@
...
+   
+In `@types` decorator, pyccel supports the following data types: real, double, float, pythonfloat, float32, float64, pythoncomplex, complex, complex64, complex128, int8, int16, int32, int64, int, bool.
```
saidctb, line 62: we also don't need `pythonfloat` and `pythoncomplex`

EmilyBourne (approved):

saidctb (approved):

yguclu:
Can anyone explain to me why this file is in the `doc` directory, instead of the root directory, or maybe `tutorial`?

mostafaMamoni:
> Can anyone explain to me why this file is in the `doc` directory, instead of the root directory, or maybe `tutorial`?

its just a personnel choice, But i think it belongs to Tutorial folder since it contains steps to write a compilable code for Pyccel.

yguclu:
@pyccel/pyccel-dev Please have a look at the latest version of file `tutorial/quickstart.md`.

I am not convinced that Example 3 is useful, as it does not explain how to use Pyccel, but it only shows the code that is generated... plus the Python code is not interesting at all, as it only contains some array initializations.

Instead I think we should add a short section about the interactive usage of Pyccel (with `epyccel`). What do you think?

EmilyBourne (commented):
I think it is odd that Fortran is our default language, but all our examples are in C. Especially if you want to explain the directory structure as there is an extra file generated for Fortran.

I think example 3 has some value as it is the only example which contains arrays. However it could be more useful if it was more complex. E.g. a function with a local array and an array as an argument which it then does something with. Of course this example would then have to be in Fortran.

I agree that a short section about epyccel could be useful

tutorial/quickstart.md
```diff
@@ -0,0 +1,348 @@
...
+  end program prog_test
+  ```
+
+### Interactive Usage with `epyccel`
+
+__TODO__
+
+## Other Features
+
+Also, we are working on supporting [openmp](https://en.wikipedia.org/wiki/OpenMP), [openmpi](https://en.wikipedia.org/wiki/Open_MPI), [lapack](https://en.wikipedia.org/wiki/LAPACK)/[blas](https://en.wikipedia.org/wiki/Basic_Linear_Algebra_Subprograms), [cuda](https://en.wikipedia.org/wiki/CUDA), [openacc](https://en.wikipedia.org/wiki/OpenACC), [task-based parallelism](https://en.wikipedia.org/wiki/Task_parallelism).
```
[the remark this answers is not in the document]
yguclu: I fixed the grammar in f4ec99d and expanded this section in f6efb22

tutorial/quickstart.md
```diff
@@ -0,0 +1,350 @@
...
+else:
+    y = 2    # int     : OK
+    z = 3j   # complex : Pyccel raises an error!
+```
+
+### Type Annotations
+
+When parsing a function, Pyccel needs to know the type of the input arguments in order to perform type inference, and ultimately compute the type of the output result.
+(In the case of recursive functions, the return type should also be declared.)
+Pyccel can then perform type inference on other Python code that uses that function, because the type of its result will be already known.
```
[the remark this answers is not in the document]
yguclu: I fixed the grammar in f4ec99d

yguclu:
> I think it is odd that Fortran is our default language, but all our examples are in C. Especially if you want to explain the directory structure as there is an extra file generated for Fortran.

I preferred to show C examples because:
1. Most programmers know C, but not Fortran
2. This is a quickstart file, so I prefer to describe a simpler directory structure for `__pyccel__`

yguclu:
> I think example 3 has some value as it is the only example which contains arrays. However it could be more useful if it was more complex. E.g. a function with a local array and an array as an argument which it then does something with. Of course this example would then have to be in Fortran.

I don't think the current example 3 has any value, because the Python code has none.
We can show a bubble sort maybe. And yes, this could be Fortran

yguclu:
As a side note.

This PR was discussed at length and approved by 4 people before I had a look at it. But it was definitely not OK as a quickstart guide, and so I had to rewrite it completely. But it's still not ready, so everybody's input is welcome!

If we really care about our users' documentation, we need to keep a high quality standard. We should prefer quality over quantity, especially for a quickstart file that is the first thing (or maybe the second after the README) that a user will read...

tutorial/quickstart.md
```diff
@@ -0,0 +1,353 @@
...
+}
+```
+#### Example 2: extension module
+
+If the Python file to be accelerated only contains functions and class definitions, Pyccel will treat it as a Python module instead of just a script.
+Accordingly, it will not generate a program, but rather a Python C extension module which can be imported from Python.
+
+For example, we now consider the Python module `mod.py`, which reads
+```python
+def factorial(n : int):
```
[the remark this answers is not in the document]
yguclu: We could write factorial in a recursive way...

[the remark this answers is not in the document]
yguclu: Done. I've also added an matrix-matrix multiplication example in Fortran, and a quicksort example using `epyccel`.

tutorial/quickstart.md
```diff
@@ -0,0 +1,459 @@
...
+                c[i, j] += a[i, k] * b[k, j]
+#$ omp end parallel
+
+    return 0
+```
+We now translate this file to Fortran, and compile it to a Python C extension module, using the command
+```bash
+$ pyccel mod.py --language fortran
+```
+The flag `--language fortran` may be omitted, as Pyccel uses Fortran as the default backend language.
```
EmilyBourne, line 296: Is it worth mentioning here that arrays are not yet fully supported in C?

tutorial/quickstart.md
```diff
@@ -0,0 +1,459 @@
...
+
+In addition to the `pyccel` command, the Pyccel library provides the `epyccel` Python function, whose name stands for "embedded Pyccel": given a pure Python function `f` with type annotations, `epyccel` returns a "pyccelized" function `f_fast` that can be used in the same Python session.
+For example:
+```python
+from pyccel.epyccel import epyccel
+from mod import f
+
+f_fast = epyccel(f)
+```
+In practice `epyccel` copies the contents of `f` into a temporary python file in the `__epyccel__` directory, then it calls the `pyccel` command to generate a Python C extension module that contains a single pyccelized function, and finally it imports this function and returns it to the caller.
```
EmilyBourne, line 388: I think it could be useful to add a sentence about imports here:
```
In practice `epyccel` copies the contents of `f` into a temporary python file in the `__epyccel__` directory. As a result it is important that all imports are written inside the function when using `epyccel`. Once the file has been copied, `epyccel` calls the `pyccel` command to generate a Python C extension module that contains a single pyccelized function. Then finally, it imports this function and returns it to the caller.
```

EmilyBourne (changes requested):
It looks good. Just a few comments

tutorial/quickstart.md
```diff
@@ -0,0 +1,459 @@
+# Pyccel's Quickstart Guide
+
+Pyccel is a **static compiler** for Python 3, using Fortran or C as backend language, with a focus on high-performance computing (HPC) applications.
+
+Pyccel's main goal is to accelerate the transition from **prototype** to **production** in scientific computing, where programmers usually develop their prototype code in a user-friendly interactive language like Python, but they later need a statically compiled language like Fortran/C/C++ (as well as SIMD vectorization, parallel multi-threading, MPI parallelization, GPU offloading, etc.) in order to achieve the performance required by their final application.
```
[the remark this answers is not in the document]
yguclu: We can definitely put a full stop after "scientific computing", but I'm not sure of how to break the rest. Suggestions?

tutorial/quickstart.md
yguclu: I am not sure that we could claim that they are fully supported in Fortran either! :sweat_smile:

In any case our array support in C is getting very close to Fortran, I don't think we need to be too specific here. The more we write and the more we have to maintain!

By the way, I've run the quicksort example with C, but I got a runtime error (segmentation fault).

tutorial/quickstart.md
yguclu: I agree, this is important to mention! Thanks

tutorial/quickstart.md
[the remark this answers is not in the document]
yguclu: I have improved this first part of the file with commit fdb1a58. Please let me know what you think of it

tutorial/quickstart.md
```diff
@@ -0,0 +1,459 @@
+# Pyccel's Quickstart Guide
+
+Pyccel is a **static compiler** for Python 3, using Fortran or C as backend language, with a focus on high-performance computing (HPC) applications.
```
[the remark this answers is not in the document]
yguclu: Fixed in commit fdb1a58

tutorial/quickstart.md
yguclu: I made the changes you suggested in commit 9131b26

tutorial/quickstart.md
```diff
@@ -1,10 +1,13 @@
...
+Pyccel's main goal is to resolve the main bottleneck in scientific computing, that is the transition from **prototype** to **production**.
+Programmers usually develop their prototype code in a user-friendly interactive language like Python, but their final application requires an HPC implementation and therefore a new production code.
+In most cases this is written in a statically compiled language like Fortran/C/C++, and it uses SIMD vectorization, parallel multi-threading, MPI parallelization, GPU offloading, etc.
 
-Pyccel generates very fast Fortran or C code which is **human-readable**, hence the expert programmer can easily profile the code on the target machine and further optimize it.
+We believe that this expensive process can be avoided, or at least drastically reduced, by using Pyccel to accelerate the most computationally intensive parts of the Python prototype.
+Not only the Pyccel-generated Fortran or C code is very fast, but it is **human-readable**; hence the expert programmer can easily profile the code on the target machine and further optimize it.
```
EmilyBourne, line 10: `Not only the Pyccel-generated Fortran or C code is very fast`->`Not only is the Pyccel-generated Fortran or C code very fast`

tutorial/quickstart.md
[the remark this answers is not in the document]
EmilyBourne: That looks much more readable :+1:

tutorial/quickstart.md
yguclu: So the verb goes in second place in a sentence?! :open_mouth: Are we writing in German now? :sweat_smile:

yguclu: I've fixed this in commit 47e3382

EmilyBourne: I've no idea why it works like that, but it does when you have "not only" at the start of a sentence
https://dictionary.cambridge.org/grammar/british-grammar/not-only-but-also

tutorial/quickstart.md
EmilyBourne: I have found the problem. It is linked to the `FloorDiv` operator. I am opening 2 issues about it. The example works in c if you use:
`int((lo+hi)/2)` instead of `(lo+hi)//2`

yguclu: Thanks for looking into this! I will keep the example as it is, as I am sure that we're going to fix the problem soon. Maybe this code should be added to our unit tests.

tutorial/quickstart.md
```diff
@@ -0,0 +1,462 @@
+# Pyccel's Quickstart Guide
+
+Pyccel is a **static compiler** for Python 3, using Fortran or C as a backend language, with a focus on high-performance computing (HPC) applications.
+
+Pyccel's main goal is to resolve the main bottleneck in scientific computing, that is the transition from **prototype** to **production**.
```
EmilyBourne, line 5: This line sounds strange due to the repetition of the word `main` and the grammar around `that`. I suggest:
```
Pyccel's main goal is to resolve the principal bottleneck in scientific computing: the transition from **prototype** to **production**
```

yguclu: Fixed in commit eb26349

EmilyBourne (approved):

EmilyBourne (approved):

PR #623, opened 2020-12-21
Unravel loops when necessary

EmilyBourne:
Functions are added to `pyccel.ast.utilities` to translate vector expressions into for loops. This allows expressions which are acceptable in python to be translated into languages which would not normally accept this syntax.
**Example 1**
The following code:
```python
import numpy as np

def my_sum(a : 'int[:]', b : 'int[:]', c : 'int[:]'):
    c[:] = a + b
```
is translated to the following C code:
```c
#include "test.h"
#include "ndarrays.h"
#include <stdint.h>
#include <stdlib.h>

void my_sum(t_ndarray a, t_ndarray b, t_ndarray c)
{
    int64_t i_0001;
    for (i_0001 = 0; i_0001 < c.shape[0]; i_0001 += 1)
    {
        c.nd_int64[get_index(c, i_0001)] = a.nd_int64[get_index(a, i_0001)] + b.nd_int64[get_index(b, i_0001)];
    }
}
```

The unrolling functions take an argument `language_has_vectors` which indicates whether the language has basic support for vector operations on objects with the same shape (e.g. fortran). If it doesn't then everything is unrolled. If it does then the expressions are only unrolled as necessary. Thus example 1 generates the following fortran code:
```fortran
module test

implicit none

contains

subroutine my_sum(a, b, c)

  implicit none

  integer(C_INT64_T), intent(in) :: a(0:)
  integer(C_INT64_T), intent(in) :: b(0:)
  integer(C_INT64_T), intent(inout) :: c(0:)

  c(:) = a + b

end subroutine my_sum

end module test
```

Unrolling allows for arithmetic operations with mixed C/Fortran ordering.
**Example 2**
The following code:
```python
import numpy as np

def my_mat_sum(a : 'int[:,:]', b : 'int[:,:](order=F)', c : 'int[:,:]'):
    c[:,:] = a + b
```
gives the following c code:
```c
#include "test.h"
#include <stdint.h>
#include <stdlib.h>
#include "ndarrays.h"

void my_mat_sum(t_ndarray a, t_ndarray b, t_ndarray c)
{
    int64_t i_0001;
    int64_t i_0002;
    for (i_0001 = 0; i_0001 < c.shape[0]; i_0001 += 1)
    {
        for (i_0002 = 0; i_0002 < c.shape[1]; i_0002 += 1)
        {
            c.nd_int64[get_index(c, i_0001, i_0002)] = a.nd_int64[get_index(a, i_0001, i_0002)] + b.nd_int64[get_index(b, i_0001, i_0002)];
        }
    }
}
```
and the following fortran code:
```fortran
module test

implicit none

contains

subroutine my_mat_sum(a, b, c)

  implicit none

  integer(C_INT64_T), intent(in) :: a(0:,0:)
  integer(C_INT64_T), intent(in) :: b(0:,0:)
  integer(C_INT64_T), intent(inout) :: c(0:,0:)
  integer(C_INT64_T) :: i_0001
  integer(C_INT64_T) :: i_0002

  do i_0001 = 0_C_INT64_T, size(c, 2_C_INT64_T - 0_C_INT64_T, C_INT64_T) &
      -1_C_INT64_T, 1_C_INT64_T
    do i_0002 = 0_C_INT64_T, size(c, 2_C_INT64_T - 1_C_INT64_T, &
      C_INT64_T)-1_C_INT64_T, 1_C_INT64_T
      c(i_0002, i_0001) = a(i_0002, i_0001) + b(i_0001, i_0002)
    end do
  end do

end subroutine my_mat_sum

end module test
```

Statements with the same loop boundaries are grouped together

**Example 3**
The following:
```python
import numpy as np

def operations(a : 'int[:]', b : 'int[:]', c : 'int[:]'):
    c[:] = a + b
    c[:] = a - b
    c[:] = a * b
    c[:] = a / b
```
gives the following c code:
```c
#include "test.h"
#include <stdint.h>
#include <stdlib.h>
#include "ndarrays.h"

void operations(t_ndarray a, t_ndarray b, t_ndarray c)
{
    int64_t i_0001;
    for (i_0001 = 0; i_0001 < c.shape[0]; i_0001 += 1)
    {
        c.nd_int64[get_index(c, i_0001)] = a.nd_int64[get_index(a, i_0001)] + b.nd_int64[get_index(b, i_0001)];
        c.nd_int64[get_index(c, i_0001)] = a.nd_int64[get_index(a, i_0001)] - b.nd_int64[get_index(b, i_0001)];
        c.nd_int64[get_index(c, i_0001)] = a.nd_int64[get_index(a, i_0001)] * b.nd_int64[get_index(b, i_0001)];
        c.nd_int64[get_index(c, i_0001)] = (double)(a.nd_int64[get_index(a, i_0001)]) / (double)(b.nd_int64[get_index(b, i_0001)]);
    }
}
```
and the following fortran code:
```fortran
module test

implicit none

contains

subroutine operations(a, b, c)

  implicit none

  integer(C_INT64_T), intent(in) :: a(0:)
  integer(C_INT64_T), intent(in) :: b(0:)
  integer(C_INT64_T), intent(inout) :: c(0:)

  c(:) = a + b
  c(:) = a - b
  c(:) = a * b
  c(:) = Real(a, C_DOUBLE) / Real(b, C_DOUBLE)

end subroutine operations

end module test
```

**Example 4**
The following:
```python
import numpy as np

def operations2(a : 'int[:]', b : 'int[:]', c : 'int[:]', d : 'int[:]'):
    c[:] = a + b
    c[:] = a - b
    d[:] = a * b
    c[:] = a / b
```
gives the following c code:
```c
#include "test.h"
#include <stdlib.h>
#include <stdint.h>
#include "ndarrays.h"

void operations2(t_ndarray a, t_ndarray b, t_ndarray c, t_ndarray d)
{
    int64_t i_0001;
    for (i_0001 = 0; i_0001 < c.shape[0]; i_0001 += 1)
    {
        c.nd_int64[get_index(c, i_0001)] = a.nd_int64[get_index(a, i_0001)] + b.nd_int64[get_index(b, i_0001)];
        c.nd_int64[get_index(c, i_0001)] = a.nd_int64[get_index(a, i_0001)] - b.nd_int64[get_index(b, i_0001)];
    }
    for (i_0001 = 0; i_0001 < d.shape[0]; i_0001 += 1)
    {
        d.nd_int64[get_index(d, i_0001)] = a.nd_int64[get_index(a, i_0001)] * b.nd_int64[get_index(b, i_0001)];
    }
    for (i_0001 = 0; i_0001 < c.shape[0]; i_0001 += 1)
    {
        c.nd_int64[get_index(c, i_0001)] = (double)(a.nd_int64[get_index(a, i_0001)]) / (double)(b.nd_int64[get_index(b, i_0001)]);
    }
}
```

**Example 5**
A more complicated example showing mixed rank. The following:
```python
@types('int[:,:,:]','int[:,:]','int[:]')
def grouped_expressions2(a,b,c):
    import numpy as np
    a[:] = a - c
    a[:] = a * b
    a[:] = a + b
    a[:] = a + c
    a += np.sum(b)
```
Is converted to:
```
subroutine grouped_expressions2(a, b, c)                               
  
  implicit none                                                        
  
  integer(C_INT64_T), intent(inout) :: a(0:,0:,0:)                     
  integer(C_INT64_T), intent(in) :: b(0:,0:)                           
  integer(C_INT64_T), intent(in) :: c(0:)                              
  integer(C_INT64_T) :: i_0                                            
  integer(C_INT64_T) :: i_1                                            
  integer(C_INT64_T) :: i_2                                            
  
  do i_0 = 0, size(a, 3, C_INT64_T)-1_C_INT64_T, 1_C_INT64_T
    do i_1 = 0, size(a, 2, C_INT64_T)-1_C_INT64_T, 1_C_INT64_T         
      a(:, i_1, i_0) = a(:, i_1, i_0) - c                              
    end do
    a(:, :, i_0) = a(:, :, i_0) * b                                    
    a(:, :, i_0) = a(:, :, i_0) + b
    do i_1 = 0, size(a, 2, C_INT64_T)-1_C_INT64_T, 1_C_INT64_T         
      a(:, i_1, i_0) = a(:, i_1, i_0) + c                              
    end do                                                             
  end do
  a = a + sum(b)                                                       

end subroutine grouped_expressions2
```

In addition dependencies are taken into consideration to avoid errors.
**Example 6**
The following:
```python
@types('int[:,:]','int[:]')
def dependencies(a,b):
    import numpy as np

    c = np.zeros_like(a)

    a[:] += b
    c += b*np.sum(a)

    a[:] = c[:]
```
generates the following fortran code (sum is not implemented in C)
```fortran
module dep

implicit none

contains

subroutine dependencies(a, b)

  implicit none

  integer(C_INT64_T), intent(inout) :: a(0:,0:)
  integer(C_INT64_T), intent(in) :: b(0:)
  integer(C_INT64_T), allocatable :: c(:,:)
  integer(C_INT64_T) :: i_0001
  integer(C_INT64_T) :: tmp_0001

  allocate(c(0:size(a, 2_C_INT64_T - 1_C_INT64_T, C_INT64_T) - &
      1_C_INT64_T, 0:size(a, 2_C_INT64_T - 0_C_INT64_T, C_INT64_T) - &
      1_C_INT64_T))
  c = 0_C_INT64_T
  do i_0001 = 0_C_INT64_T, size(a, 2_C_INT64_T - 0_C_INT64_T, C_INT64_T) &
      -1_C_INT64_T, 1_C_INT64_T
    a(:, i_0001) = a(:, i_0001) + b
  end do
  tmp_0001 = sum(a)
  do i_0001 = 0_C_INT64_T, size(a, 2_C_INT64_T - 0_C_INT64_T, C_INT64_T) &
      -1_C_INT64_T, 1_C_INT64_T
    c(:, i_0001) = c(:, i_0001) + b * tmp_0001
  end do
  a(:, :) = c(:, :)

end subroutine dependencies

end module dep
```

**Commit Summary**
- Allow `pyccel.ast.basic.Basic.substitute` to substitue one object for multiple objects
- Ensure that the `lhs` and `rhs` of an Assignment are `PyccelAstNode`s by casting lists to `PythonTuple`s
- Remove `EmptyNode`s from `CodeBlock`s
- Allow add_local_var to add multiple variables
- Add `is_elemental` property to `PyccelInternalFunction`
- Add functions `compatible_operation`, `insert_index`, `collect_loops`, `insert_fors`, `expand_tuple_assignments`, `expand_to_loops` to handle loop unravelling
- Call `expand_to_loops` from `_print_CodeBlock`
- Fix bug where we write `elif` instead of `else` as `LiteralTrue` used to be a singleton but is no longer
- Create `_current_class` attribute and `get_method` function in `FCodePrinter` to correctly collect current function to insert new variable
- Create `add_vars_to_namespace` function to reduce code duplication
- Unravel `CodeBlock` of `CodeBlock`s into one `CodeBlock`
- Fix bug where `else` section of an `If` was created even if it was empty
- Add tests with mixed order and mixed ranks
- Activate as many C tests as possible and update xfail messages
- Remove `stack_array` tag in `runtest_degree_in.py` as an array of unknown size cannot be allocated on the stack
- Use `b != NULL` to check if a pointer is not `NULL` to prevent `b` being unrolled into a for loop


saidctb:
do you want to implement the broadcasting rules  of numpy?

pyccel/codegen/printing/fcode.py
```diff
@@ -1176,14 +1173,22 @@ def _print_AliasAssign(self, expr):
         return self._get_statement(code) + '\n'
 
     def _print_CodeBlock(self, expr):
-        body = []
-        for b in expr.body:
+        body_exprs, new_vars = expand_to_loops(expr.body, language_has_vectors = True)
+        if self._current_function:
+            name = self._current_function
+            func = self.get_function(name)
+            func.local_vars.extend(new_vars)
```
EmilyBourne, line 1180: @saidctb This block seems to be causing trouble with `__init__` functions. I am getting the error:
```
pyccel:
 |fatal [codegen]: generic_methods.py| Undefined function (__init__)

```
Do you have any idea how to fix this?

EmilyBourne:
elemental functions can be tested for c if a print function is written for `nd_array` type

pyccel/codegen/printing/fcode.py
saidctb: there is something strange happening there, __init__ should not be there, if you can delete that part it will be great;I'm planing on deleting the scope dependency

pyccel/ast/utilities.py
```diff
@@ -187,3 +194,393 @@ def split_positional_keyword_arguments(*args):
...
+                 A boolean indicating if the operation is compatible
+    """
+    # If the shapes don't match then an index must be required
+    shapes = [a.shape for a in args if a.shape != ()]
+    shapes = set(tuple(d if isinstance(d, Literal) else -1 for d in s) for s in shapes)
+    order  = set(a.order for a in args if a.order is not None)
+    return len(shapes) == 1 and len(order) == 1
+
+#==============================================================================
+def insert_index(expr, pos, index_var, language_has_vectors):
```
saidctb: can use the function subs instead of creating a new one, this function needs to updated,  https://github.com/pyccel/pyccel/blob/master/pyccel/ast/core.py#L190, we will need to use it in multiple places, when we apply some optimizations

EmilyBourne: I don't think we can use subs because the problem is too complicated. If you have
```
@types('int[:,:]','int[:]')
def f(a,b):
    c[:,:]=a+b
```
Then first we need to find all the variables or IndexedElements (atoms? But I think you said that doesn't work) then we need to replace them with an indexed object which preserves the Variable. The indexes aren't inserted in the same place in every variable and I need to be able to spot functions in the expression to avoid grouping objects with dependencies.

In addition I need to make sure that I don't create new objects when visiting an object (e.g. new PyccelOperator because we operate recursively on the arguments) so that I can recognise if the index was inserted or not. If it wasn't then a for loop doesn't need to be added

pyccel/codegen/printing/fcode.py
EmilyBourne: Why should `__init__` not be there?
Is it not a FunctionDef?

saidctb: yes there is no function with name `__init__` , unless you are testing classes

pyccel/ast/utilities.py
saidctb: atoms probably are not working anymore, it needs to be updated, but then you can use it each time you have an assignment of a vectorized expression. where you insert the for loop, all this can be done with subs unless I'm missing something.
example:
```
c = a + b
atoms = [a,b,c]
new_atoms = [a[i], b[i], c[i]]
new_expr = expr.subs(zip(atoms, new_atoms))
```

saidctb: >In addition I need to make sure that I don't create new objects when visiting an object (e.g. new PyccelOperator because we operate recursively on the arguments) so that I can recognize if the index was inserted or not. If it wasn't then a for loop doesn't need to be added.
what do you mean by that? isn't it straightforward?  you just see if it's a vectorized assignment or not, if yes you insert a for loop otherwise you don't.

EmilyBourne: > c = a + b
atoms = [a,b,c]
new_atoms = [a[i], b[i], c[i]]
new_expr = expr.subs(zip(atoms, new_atoms))

The missing thing is mixing ranks. In my example above `a` and `c` are matrices and `b` is a vector so in fortran it would be translated to
```
c[:,i] = a[:,i] + b
```
While in c it is easier and we get
```
c[i,j]= a[i,j]+b[j]
```
So you have to work out what you're subbing and how many indices are needed

saidctb: yes, this is not an issue because you do it individually on each atom, once you get the atoms you can verify their rank and decide what to do there, and once you have your new atoms you give that to subs.
in your example, in Fortran, you will have
```
atoms = [a,b,c]
new_atoms = [a[:,i], b,  c[:,i]]
new_expr = expr.subs(zip(atoms, new_atoms))
```
and similarly with C

saidctb: if you want to make it easier for you, you can use pyccel_to_sympy, do the transformation, and then use, sympy_to_pyccel

EmilyBourne: > yes, this is not an issue because you do it individually on each atom, once you get the atoms you can verify their rank and decide what to do there, and once you have your new atoms you give that to subs.

That makes sense but how do you handle something like
```
c += b*np.sum(a)
```
? Atoms would give us `a` but we don't want it

EmilyBourne: Although the sum should probably be extracted from the loop anyway to stop it being calculated multiple times

saidctb: you can tell atoms where to stop, you don't search what is inside a function call for example, for the expression elimination you can check this from sympy https://github.com/sympy/sympy/blob/9e8f62e059d83178c1d8a1e19acac5473bdbf1c1/sympy/simplify/cse_main.py#L628-L772

saidctb: a lot of what you are looking for has already been done in Sympy, we either use the one of Sympy or rewrite the same functions which handle Pyccel nodes.

pyccel/codegen/printing/fcode.py
EmilyBourne: The codegen tests test classes so yes I am

pyccel/ast/utilities.py
EmilyBourne: > you can tell atoms where to stop

Ah I didn't realize that, in that case this might be a good answer

EmilyBourne: So the functions we need are `atoms` and `subs`. I don't see any point in converting to sympy as we are not doing anything specific to sympy (ie calculations). So it sounds like we need to implement our own `atoms` and `subs`. I would imagine that that should be done in multiple PRs. Where would you recommend starting? Do you have any suggestions of things we can use for tests?

saidctb: we need to have an attribute  `_args` that we add to the `PyccelAstNode` that we set whenever we create a Pyccel node.
this way we can create the subs algorithm, where we go through the args, without adding all the node types.
it's also enough for the `atoms`  if a Pyccel node doesn't have args it means that it is an atom.

pyccel/codegen/printing/fcode.py
EmilyBourne: I have added a `self._current_class` attribute which works similarly to `self._current_function` (See 2621994)
Does this make sense to you?

tests/epyccel/test_epyccel_openmp.py
```diff
@@ -243,7 +243,7 @@ def test_omp_matmul_2d_2d(language):
 
 @pytest.mark.parametrize( 'language', [
         pytest.param("c", marks = [
-            pytest.mark.xfail(reason="Numpy Arrays not implemented in C !"),
+            pytest.mark.xfail(reason="for reduction doesn't work in C"),
```
EmilyBourne, line 246: @nhamidn confirms that this is fixed in PR #593

nhamidn: yes it's fixed

tests/epyccel/test_epyccel_optional_args.py
```diff
@@ -1,188 +1,190 @@
...
+
+#------------------------------------------------------------------------------
+@pytest.mark.parametrize( 'language', (
+        pytest.param("fortran", marks = pytest.mark.fortran),
+        pytest.param("c", marks = [
+            pytest.mark.xfail(reason="Arrays cannot be optional. See #757"),
+            pytest.mark.c]
+        )
+    )
+)
```
EmilyBourne, line 147: This error message and the same error message below is the only change in this file. Github is reporting that the whole file changed as my computer changed the line endings from windows format to unix format

saidctb (approved):

tests/epyccel/test_epyccel_optional_args.py
yguclu: Who among us is using Windows line endings????

yguclu:
@EmilyBourne Could you add a few more examples of generated code (both C and Fortran) to the discussion here, for the various cases that you take into consideration?

tests/epyccel/test_epyccel_optional_args.py
EmilyBourne: Git blame says @MenoIy

tests/pyccel/scripts/runtest_degree_in.py
```diff
@@ -3,7 +3,6 @@
 
 @pure
 @types('int')
-@stack_array('tmp')
```
yguclu, line 6: Why did you remove this line? The length of a stack array does not need to be known at compile time, neither in Fortran nor in C.

EmilyBourne: Ah, you are right. This doesn't work for 2 reasons in C:
1. The C code uses `Variable.shape` instead of `Variable.alloc_shape`
2. The C code uses compound literals to manage stack arrays. But the size of a compound literal must be known at runtime (it is the error from this which confused me)

EmilyBourne: Actually I think the compound literal is just done wrong. As we know the rank we should know the size. I will open an issue

EmilyBourne: Fixed in ac444f0
The issue is #760

EmilyBourne:
@yguclu 
> Could you add a few more examples of generated code (both C and Fortran) to the discussion here, for the various cases that you take into consideration?

I've just realised you said discussion. I added them to the description. Should I remove them and put them in the discussion and just keep one in the description?

tests/pyccel/scripts/runtest_degree_in.py
yguclu: Thanks!

pyccel/codegen/printing/fcode.py
```diff
@@ -1189,14 +1212,16 @@ def _print_AliasAssign(self, expr):
...
+        for b in body_exprs :
             line = self._print(b)
             if (self._additional_code):
-                body.append(self._additional_code)
+                body_stmts.append(self._additional_code)
                 self._additional_code = None
-            body.append(line)
-        return ''.join(body)
+            body_stmts.append(line)
+        return ''.join(self._print(b) for b in body_stmts)
```
yguclu, line 1224: I've just realized that this function is a bit different than the one in the C printer:

1. In the C printer we set `self._additional_code = ''` (empty string), while here we set it to `None`
2. In the C printer the statements are missing the final newline character, so that has to be added when joining lines

Regarding point 1 I don't think we should store the additional code in a string, so the Fortran printer's behavior is preferable. Similarly about point 2, I think it's cleaner to have one-line statements end with `\n`. What do you think?

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -153,7 +153,7 @@ def _get_static_function(self, used_names, function, collect_dict):
                         var = Variable(dtype=NativeInteger() ,name = self.get_new_name(used_names, a.name + "_dim"))
                         body = FunctionCall(numpy_get_dim, [collect_dict[a], i])
                         if a.is_optional:
-                            body = IfTernaryOperator(VariableAddress(collect_dict[a]), body , LiteralInteger(0))
+                            body = IfTernaryOperator(PyccelIsNot(VariableAddress(collect_dict[a]),Nil()), body , LiteralInteger(0))
```
yguclu, line 156: I this a bug fix? If yes, please add a line about it in the PR's description

pyccel/ast/numpyext.py
```diff
@@ -805,6 +808,10 @@ def _set_order(self, x1, x2):
         else:
             self._order = 'C'
 
+    @property
+    def is_elemental(self):
+        return True
+
```
yguclu, line 814: Do we need this property here? The parent class `NumpyUfuncBase` already has it.

pyccel/ast/core.py
```diff
@@ -2149,17 +2156,17 @@ def rename(self, newname):
...
 
         Parameters
         ----------
         var : Variable
               The new local variable
         """
-        var.set_current_user_node(self)
-        self._local_vars = self._local_vars + (var,)
+        _ = [v.set_current_user_node(self) for v in var]
+        self._local_vars = self._local_vars + var
```
yguclu, line 2170: I suggest we change the name of this function to `add_local_vars` (plural), which is consistent with other function names in the parser and printers.

yguclu: I would also call the argument `vars` (plural) for the same reason.

pyccel/ast/core.py
```diff
@@ -2149,17 +2156,17 @@ def rename(self, newname):
...
```
yguclu, line 2169: If `self._local_vars` is a tuple or a list, this can be written as
```python
self._local_vars += var
```

pyccel/ast/utilities.py
```diff
@@ -177,3 +186,398 @@ def split_positional_keyword_arguments(*args):
...
+    >>> insert_index(expr, 0, i, language_has_vectors = False)
+    IndexedElement(c, i) := IndexedElement(a, i) + IndexedElement(b, i)
+    >>> insert_index(expr, 0, i, language_has_vectors = True)
+    c := a + b
+    """
+    if isinstance(expr, (Variable, VariableAddress)):
+        if expr.rank==0 or -pos>expr.rank:
+            return expr
+        if expr.shape[pos]==1:
+            # If there is no dimemsion in this axis, reduce the rank
```
yguclu, line 257: Typo: dimemsion -> dimension

pyccel/ast/utilities.py
```diff
@@ -177,3 +186,398 @@ def split_positional_keyword_arguments(*args):
...
+
+    elif isinstance(expr, IndexedElement):
+        base = expr.base
+        indices = list(expr.indices)
+        if -pos>expr.base.rank or not isinstance(indices[pos], Slice):
+            return expr
+
+        # Add index at the required position
+        if expr.base.shape[pos]==1:
+            # If there is no dimemsion in this axis, reduce the rank
```
yguclu, line 272: Typo: dimemsion -> dimension

pyccel/codegen/printing/fcode.py
EmilyBourne: I agree that point 2 needs to also be done for the C printer. We are starting to have empty lines in the C code more often so I think this should be fixed.
I'm not sure there is much difference for point 1. In both cases the additional code is stored in a string so that it is printed in the right order. The only difference is the value when it is empty

In any case I don't think these points should be fixed in this PR as it is already complicated enough on its own

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: Yes and no. Previously the output was:
```
a = b ?  c : d;
```
whereas now it is:
```
a = b == NULL ?  c : d;
```
The second option is clearer and allows me to find and exclude this case rather than trying to unravel it as `b` is an array

pyccel/ast/numpyext.py
EmilyBourne: Ah no! I confused `NumpyUfuncUnary` and `NumpyUfuncBase`

pyccel/ast/core.py
EmilyBourne: `vars` is a [builtin function](https://docs.python.org/3/library/functions.html#vars) in python

pyccel/ast/utilities.py
EmilyBourne: Fixed in 4f6b8d0

pyccel/ast/utilities.py
EmilyBourne: Fixed in 4f6b8d0

pyccel/ast/core.py
EmilyBourne: Fixed in 283acb5

pyccel/ast/numpyext.py
EmilyBourne: Fixed in aa02107

pyccel/ast/core.py
EmilyBourne: Fixed in d872f4f

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: I have added a line

pyccel/ast/utilities.py
```diff
@@ -177,3 +186,398 @@ def split_positional_keyword_arguments(*args):
...
+    >>> from pyccel.ast.operators import PyccelAdd
+    >>> from pyccel.ast.utilities import expand_to_loops
+    >>> a = Variable('int', 'a', shape=(4,), rank=1)
+    >>> b = Variable('int', 'b', shape=(4,), rank=1)
+    >>> c = Variable('int', 'c', shape=(4,), rank=1)
+    >>> i = Variable('int', 'i', shape=())
+    >>> d = PyccelAdd(a,b)
+    >>> expr = [Assign(c,d)]
+    >>> expand_to_loops(expr, language_has_vectors = False)
+    [For(i_0, PythonRange(0, LiteralInteger(4), LiteralInteger(1)), CodeBlock([IndexedElement(c, i_0) := PyccelAdd(IndexedElement(a, i_0), IndexedElement(b, i_0))]), [])]
```
yguclu, line 528: This example is for function `expand_to_loops`, isn't it?

pyccel/codegen/printing/fcode.py
yguclu: OK. Can you please create an issue for point 2?

pyccel/ast/utilities.py
EmilyBourne: Ah you are right, I meant to go back and fix this but I forgot. Fixed in 3b78db5

pyccel/codegen/printing/fcode.py
EmilyBourne: Done. See issue #761

yguclu: Thanks!

yguclu:
> @yguclu
> 
> > Could you add a few more examples of generated code (both C and Fortran) to the discussion here, for the various cases that you take into consideration?
> 
> I've just realised you said discussion. I added them to the description. Should I remove them and put them in the discussion and just keep one in the description?

Don't worry about it, I will not include the examples in the merge commit's message.

Issue #587, opened 2020-12-08
Create a precision node


EmilyBourne:
**Describe the bug**
In fortran the precision needs to be specified in lots of different places. It is always printed using ISO_C_BINDING. It is easy to forget this when coding in `FCodePrinter`. As a DEVELOPER I would like to be able to print the precision in the same way as I print everything else, with a call to `fprint`

**Acceptance Criteria**
- [x] Create a Precision class
- [ ] Ensure that the `precision` of a `PyccelAstNode` is of type `Precision`. This can be done by converting precision to the expected type in `PyccelAstNode.__init__`
- [ ] Implement the necessary `_print_Precision` functions
- [ ] Replace the current implementation of precision printing with a call to `self._print`

cc: @mostafaMamoni

EmilyBourne:
The dtype may also be needed in this object

PR #625, opened 2020-12-22
Make Fortran printer handle numpy.array when rank > 2

saidctb:
* Update `PythonTuple` and `PythonList`;
* Correctly print `numpy.array(expr)` to Fortran when `expr.rank > 2`;
* When necessary, use `transpose` for `rank = 2` and `reshape` for `rank > 2`;
* Use precision from `iso_c_bindings` dictionary;
* Fix bug in line splitting when len(line)>71;
* Don't split a comment line;
* Add unit tests.

### DIFF
--- a/pyccel/ast/builtins.py
+++ b/pyccel/ast/builtins.py
@@ -316,7 +316,8 @@ class PythonTuple(Expr, PyccelAstNode):
             return
         is_homogeneous = all(a.dtype is not NativeGeneric() and \
                              args[0].dtype == a.dtype and \
-                             args[0].rank  == a.rank  for a in args[1:])
+                             args[0].rank  == a.rank  and \
+                             args[0].order == a.order for a in args[1:])
         self._inconsistent_shape = not all(args[0].shape==a.shape   for a in args[1:])
         self._is_homogeneous = is_homogeneous
         if is_homogeneous:
@@ -402,10 +403,16 @@ class PythonLen(Function, PyccelAstNode):
 #==============================================================================
 class PythonList(Tuple, PyccelAstNode):
     _order = 'C'
+    _is_homogeneous = True
     """ Represent lists in the code with dynamic memory management."""
     def __init__(self, *args, **kwargs):
         if self.stage == 'syntactic':
             return
+        self._is_homogeneous = all(a.dtype is not NativeGeneric() and \
+                             args[0].dtype == a.dtype and \
+                             args[0].rank  == a.rank  and \
+                             args[0].order == a.order for a in args[1:])
+
         bools     = [a for a in args if a.dtype is NativeBool()]
         integers  = [a for a in args if a.dtype is NativeInteger()]
         reals     = [a for a in args if a.dtype is NativeReal()]
@@ -429,14 +436,16 @@ class PythonList(Tuple, PyccelAstNode):
             elif bools:
                 self._dtype     = NativeBool()
                 self._precision  = max(a.precision for a in bools)
-            else:
-                raise TypeError('cannot determine the type of {}'.format(self))
 
             shapes = [a.shape for a in args]
             self._rank = max(a.rank for a in args) + 1
             if all(sh is not None for sh in shapes):
                 self._shape = (LiteralInteger(len(args)), ) + shapes[0]
                 self._rank  = len(self._shape)
+
+    @property
+    def is_homogeneous(self):
+        return self._is_homogeneous
 
 #==============================================================================
 class PythonMap(Basic):
--- a/pyccel/ast/builtins.py
+++ b/pyccel/ast/builtins.py
@@ -422,7 +422,7 @@ class PythonList(Tuple, PyccelAstNode):
             self._dtype = NativeString()
             self._rank  = 0
             self._shape = ()
-            assert len(integers + reals + complexes) == 0
+
         else:
             if complexes:
                 self._dtype     = NativeComplex()
--- a/pyccel/ast/builtins.py
+++ b/pyccel/ast/builtins.py
@@ -344,13 +344,12 @@ class PythonTuple(Expr, PyccelAstNode):
                     raise TypeError('cannot determine the type of {}'.format(self))
 
 
-                shapes = [a.shape for a in args]
-
+                shapes     = [a.shape for a in args]
+                self._rank = max(a.rank for a in args) + 1
                 if all(sh is not None for sh in shapes):
                     self._shape = (LiteralInteger(len(args)), ) + shapes[0]
                     self._rank  = len(self._shape)
-                else:
-                    self._rank = max(a.rank for a in args) + 1
+
         else:
             self._rank      = max(a.rank for a in args) + 1
             self._dtype     = NativeGeneric()
@@ -431,13 +430,11 @@ class PythonList(Tuple, PyccelAstNode):
                 raise TypeError('cannot determine the type of {}'.format(self))
 
             shapes = [a.shape for a in args]
-
+            self._rank = max(a.rank for a in args) + 1
             if all(sh is not None for sh in shapes):
-                assert all(sh==shapes[0] for sh in shapes)
                 self._shape = (LiteralInteger(len(args)), ) + shapes[0]
                 self._rank  = len(self._shape)
-            else:
-                self._rank = max(a.rank for a in args) + 1
+
 #==============================================================================
 class PythonMap(Basic):
     """ Represents the map stmt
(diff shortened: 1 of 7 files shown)


pyccel/ast/builtins.py
```diff
@@ -431,13 +430,11 @@ def __init__(self, *args, **kwargs):
                 raise TypeError('cannot determine the type of {}'.format(self))
 
             shapes = [a.shape for a in args]
-
+            self._rank = max(a.rank for a in args) + 1
             if all(sh is not None for sh in shapes):
-                assert all(sh==shapes[0] for sh in shapes)
                 self._shape = (LiteralInteger(len(args)), ) + shapes[0]
```
EmilyBourne, line 435: Does it still make sense to save the shape in the other dimensions if it is not the same everywhere? (I am assuming that is why the assertion was removed)

pyccel/codegen/printing/fcode.py
```diff
@@ -603,13 +602,15 @@ def _print_SumFunction(self, expr):
     def _print_PythonLen(self, expr):
         var = expr.arg
         idx = 1 if var.order == 'F' else var.rank
+        prec = self._print(expr.precision)
```
EmilyBourne, line 605: This improvement is not implemented yet, you still need to use the `iso_c_binding` dictionary

pyccel/codegen/printing/fcode.py
```diff
@@ -694,21 +695,16 @@ def _print_NumpyWhere(self, expr):
...
-            arg = functools.reduce(operator.concat, expr.arg)
-            rhs_code = 'reshape({array}, {shape})'.format(
-                    array=self._print(arg), shape=self._print(Tuple(*shape)))
-        else:
-            rhs_code = self._print(expr.arg)
-
+        rhs_code = self._print(expr.arg)
         # If Numpy array is stored with column-major ordering, transpose values
+        # use reshape with order for rank > 2
         if expr.order == 'F' and expr.rank > 1:
```
EmilyBourne, line 702: I don't see the `else`. Is C ordering handled?

pyccel/codegen/printing/fcode.py
```diff
@@ -719,15 +715,16 @@ def _print_NumpyFloor(self, expr):
     # ======================================================================= #
     def _print_PyccelArraySize(self, expr):
         init_value = self._print(expr.arg)
-
+        prec       = self._print(expr.precision)
```
EmilyBourne, line 718: See above precision needs `iso_c_binding` dictionary see #587

EmilyBourne (changes requested):

pyccel/ast/builtins.py
saidctb: we can not verify if they are the same, we just assume that they are equal for the moment, and if there is a problem it will be a run time error

pyccel/codegen/printing/fcode.py
saidctb: yes it's you don't need the else

pyccel/codegen/printing/fcode.py
saidctb: done

pyccel/codegen/printing/fcode.py
saidctb: done

EmilyBourne (approved):

pyccel/codegen/printing/fcode.py
```diff
@@ -603,13 +602,15 @@ def _print_SumFunction(self, expr):
...
+
         dtype = var.dtype
         if dtype is NativeString():
             return 'len({})'.format(self._print(var))
         elif var.rank == 1:
-            return 'size({})'.format(self._print(var))
+            return 'size({}, kind={})'.format(self._print(var), prec)
         else:
-            return 'size({},{})'.format(self._print(var), self._print(idx))
+            return 'size({},{},{})'.format(self._print(var), self._print(idx), prec)
```
yguclu, line 613: In order to be consistent with line 611, shouldn't we pass `prec` through the `kind` keyword argument?

pyccel/codegen/printing/fcode.py
```diff
@@ -694,21 +695,16 @@ def _print_NumpyWhere(self, expr):
...
+        # use reshape with order for rank > 2
         if expr.order == 'F' and expr.rank > 1:
-            rhs_code = 'transpose({})'.format(rhs_code)
+            shape    = ', '.join(self._print(i) for i in expr.shape)
+            order    = [LiteralInteger(i) for i in range(1, expr.rank+1)]
+            order    = order[1:]+ order[:1]
+            order    = ', '.join(self._print(i) for i in order)
+            rhs_code = 'reshape({},[{}], order=[{}])'.format(rhs_code, shape, order)
 
         return rhs_code
```
yguclu, line 709: According to the comment `reshape` should be used for `rank > 2`, so I would expect `transpose` to be used when `rank == 2` instead.

pyccel/codegen/printing/fcode.py
```diff
@@ -719,15 +715,16 @@ def _print_NumpyFloor(self, expr):
...
             index = self._print(expr.arg.rank - expr.index)
         else:
             index = self._print(expr.index + 1)
 
-        code_init = 'size({0}, {1})'.format(init_value, index)
+        if expr.arg.rank == 1:
+            return 'size({0}, kind={1})'.format(init_value, prec)
 
-        return code_init
+        return 'size({0}, {1}, {2})'.format(init_value, index, prec)
```
yguclu, line 727: Same comment as above: in order to have line 727 consistent with line 725, shouldn't we pass `prec` through the `kind` keyword argument?

pyccel/codegen/printing/fcode.py
```diff
@@ -2820,7 +2817,7 @@ def split_pos_code(line, endpos):
         result = []
         trailing = ' &'
         for line in lines:
-            if len(line)>72 and('"' in line or "'" in line or '!' in line):
+            if len(line)>72 and ('"' in line[72:] or "'" in line[72:] or '!' in line[:72]):
                 result.append(line)
```
yguclu, line 2821: Can you explain what you are doing here?

pyccel/codegen/printing/fcode.py
saidctb: the keyword needs to be specified in line 611 otherwise it's not gonna work, the other line I prefere not because we use it alot and it will make the code less readable.

pyccel/codegen/printing/fcode.py
saidctb: we don't split a line that contains a string or a comment , where we want to split it

pyccel/codegen/printing/fcode.py
saidctb: done

pyccel/codegen/printing/fcode.py
saidctb: same as above

pyccel/codegen/printing/fcode.py
```diff
@@ -699,12 +699,15 @@ def _print_NumpyArray(self, expr):
...
-            order    = order[1:]+ order[:1]
-            order    = ', '.join(self._print(i) for i in order)
-            rhs_code = 'reshape({},[{}], order=[{}])'.format(rhs_code, shape, order)
+        if expr.order == 'F':
+            if expr.rank == 2:
+                rhs_code = 'transpose({})'.format(rhs_code)
+            elif expr.rank > 2:
+                shape    = ', '.join(self._print(i) for i in expr.shape)
+                order    = [LiteralInteger(i) for i in range(1, expr.rank+1)]
+                order    = order[1:]+ order[:1]
```
yguclu, line 708: I would have expected this line to read
```python
order = order[::-1]
```
instead. Have you tested this for `rank > 2`?

pyccel/codegen/printing/fcode.py
yguclu: Thanks

pyccel/codegen/printing/fcode.py
yguclu: Thanks

pyccel/codegen/printing/fcode.py
saidctb: yes, it's tested for rank > 2, it will give the wrong result if you inverted the order, because the arguments are already treated

yguclu: For example, have you tried something like this?
```python
import numpy as np

x = np.array([[[0, 1, 2, 3], [4, 5, 6, 7], [8, 9, 10, 11]], [[12, 13, 14, 15], [16, 17, 18, 19], [20, 21, 22, 23]]], order='F')
```
Here `x.shape = (2, 3, 4)`.

saidctb: it was tested on something like `np.array([x,y])` where x and y are 2d or 3d arrays

saidctb: done

yguclu (approved):
Good job @saidctb!

PR #626, opened 2020-12-22
Allow slices with negative step in C

saniQQ:
- Use ternary operator to handle slices with negative `step` when `allow_negative_index` decorator is used;
- Fix duplicated operator in case of literal values;
- Add tests.

### DIFF
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -674,7 +674,7 @@ class CCodePrinter(CodePrinter):
                         inds[i] = self._new_slice_with_processed_arguments(ind, PyccelArraySize(base, i),
                             allow_negative_indexes)
                 inds = [self._print(i) for i in inds]
-                return "array_slicing(%s, %s)" % (base_name, ", ".join(inds))
+                return "array_slicing(&%s, %s)" % (base_name, ", ".join(inds))
             inds = [self._print(i) for i in inds]
         else:
             raise NotImplementedError(expr)
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -725,8 +725,9 @@ class CCodePrinter(CodePrinter):
 
         # variable step in slice
         elif allow_negative_index and step and not isinstance(step, LiteralInteger):
+            og_start = start
             start = IfTernaryOperator(PyccelGt(step, LiteralInteger(0)), start, PyccelMinus(stop, LiteralInteger(1)))
-            stop = IfTernaryOperator(PyccelGt(step, LiteralInteger(0)), stop, start)
+            stop = IfTernaryOperator(PyccelGt(step, LiteralInteger(0)), stop, og_start)
 
         return Slice(start, stop, step)
 
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -674,7 +674,7 @@ class CCodePrinter(CodePrinter):
                         inds[i] = self._new_slice_with_processed_arguments(ind, PyccelArraySize(base, i),
                             allow_negative_indexes)
                 inds = [self._print(i) for i in inds]
-                return "array_slicing(&%s, %s)" % (base_name, ", ".join(inds))
+                return "array_slicing(%s, %s)" % (base_name, ", ".join(inds))
             inds = [self._print(i) for i in inds]
         else:
             raise NotImplementedError(expr)
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -702,13 +702,13 @@ class CCodePrinter(CodePrinter):
         # negative start and end in slice
         if isinstance(start, PyccelUnarySub) and isinstance(start.args[0], LiteralInteger):
             start = PyccelMinus(array_size, start.args[0])
-        elif allow_negative_index and not isinstance(start, LiteralInteger) and not isinstance(start, PyccelArraySize):
+        elif allow_negative_index and not isinstance(start, (LiteralInteger, PyccelArraySize)):
             start = IfTernaryOperator(PyccelLt(start, LiteralInteger(0)),
                             PyccelMinus(array_size, start), start)
 
         if isinstance(stop, PyccelUnarySub) and isinstance(stop.args[0], LiteralInteger):
             stop = PyccelMinus(array_size, stop.args[0])
-        elif allow_negative_index and not isinstance(stop, LiteralInteger) and not isinstance(stop, PyccelArraySize):
+        elif allow_negative_index and not isinstance(stop, (LiteralInteger, PyccelArraySize)):
             stop = IfTernaryOperator(PyccelLt(stop, LiteralInteger(0)),
                             PyccelMinus(array_size, stop), stop)
 
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -702,13 +702,13 @@ class CCodePrinter(CodePrinter):
         # negative start and end in slice
         if isinstance(start, PyccelUnarySub) and isinstance(start.args[0], LiteralInteger):
             start = PyccelMinus(array_size, start.args[0])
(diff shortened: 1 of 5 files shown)


pyccel/codegen/printing/ccode.py
```diff
@@ -702,13 +702,13 @@ def _new_slice_with_processed_arguments(_slice, array_size, allow_negative_index
         # negative start and end in slice
         if isinstance(start, PyccelUnarySub) and isinstance(start.args[0], LiteralInteger):
             start = PyccelMinus(array_size, start.args[0])
-        elif allow_negative_index and not isinstance(start, LiteralInteger):
+        elif allow_negative_index and not isinstance(start, LiteralInteger) and not isinstance(start, PyccelArraySize):
```
MenoIy, line 705: this can be :
```
isinstance(start, (LiteralInteger, PyArraySize))
```

pyccel/codegen/printing/ccode.py
```diff
@@ -702,13 +702,13 @@ def _new_slice_with_processed_arguments(_slice, array_size, allow_negative_index
...
             start = IfTernaryOperator(PyccelLt(start, LiteralInteger(0)),
                             PyccelMinus(array_size, start), start)
 
         if isinstance(stop, PyccelUnarySub) and isinstance(stop.args[0], LiteralInteger):
             stop = PyccelMinus(array_size, stop.args[0])
-        elif allow_negative_index and not isinstance(stop, LiteralInteger):
+        elif allow_negative_index and not isinstance(stop, LiteralInteger) and not isinstance(stop, PyccelArraySize):
```
MenoIy, line 711: here too

MenoIy (changes requested):

pyccel/codegen/printing/ccode.py
saniQQ: will fix

MenoIy (approved):

pyccel/codegen/printing/ccode.py
```diff
@@ -702,13 +702,13 @@ def _new_slice_with_processed_arguments(_slice, array_size, allow_negative_index
         # negative start and end in slice
         if isinstance(start, PyccelUnarySub) and isinstance(start.args[0], LiteralInteger):
             start = PyccelMinus(array_size, start.args[0])
-        elif allow_negative_index and not isinstance(start, LiteralInteger):
+        elif allow_negative_index and not isinstance(start, (LiteralInteger, PyccelArraySize)):
```
EmilyBourne, line 705: Why is this needed? What happens if I do:
```
@types('int[:]')
def f(a):
    import numpy as np
    s = np.shape(a)[0]
    b = np.zeros(s+1)
    print(b[s])
```
?

pyccel/codegen/printing/ccode.py
```diff
@@ -720,12 +720,12 @@ def _new_slice_with_processed_arguments(_slice, array_size, allow_negative_index
...
         elif isinstance(step, PyccelUnarySub) and isinstance(step.args[0], LiteralInteger):
-            start = array_size if _slice.start is None else start
+            start = PyccelMinus(array_size, LiteralInteger(1)) if _slice.start is None else start
             stop = LiteralInteger(0) if _slice.stop is None else stop
 
         # variable step in slice
         elif allow_negative_index and step and not isinstance(step, LiteralInteger):
-            start = IfTernaryOperator(PyccelGt(step, LiteralInteger(0)), start, stop)
+            start = IfTernaryOperator(PyccelGt(step, LiteralInteger(0)), start, PyccelMinus(stop, LiteralInteger(1)))
             stop = IfTernaryOperator(PyccelGt(step, LiteralInteger(0)), stop, start)
```
EmilyBourne, line 729: Has this been tested? I just want to check that there isn't a problem due to the fact that `stop` is constructed using the modified `start`

pyccel/stdlib/ndarrays/ndarrays.h
```diff
@@ -74,7 +74,7 @@ typedef struct  s_ndarray
     int32_t                 length;
     /* size of the array */
     int32_t                 buffer_size;
-    /*  */
+    /* True if the array do not own the data */
```
EmilyBourne, line 77: Grammar
`do`-> `does`

pyccel/codegen/printing/ccode.py
saniQQ: that condition will occur only in handling a slice object and because we know that an array size will always be positive we don't need to have to add a ternary operator

pyccel/codegen/printing/ccode.py
saniQQ: yes it is tested this code :
```python
@allow_negative_index('a')
def ok():
    a = np.array([1, 2, 3, 4, 5])
    j = -1
    b = a[::j]
    for j in range(5):
        print(b[j])
```

is translated to :
```C
void ok(void)
{
    t_ndarray a;
    int64_t j;
    t_ndarray b;

    a = array_create(1, (int32_t[]){5}, nd_int64);
    int64_t array_dummy_0001[] = {1, 2, 3, 4, 5};

    j = -1;
    b = array_slicing(a, new_slice((j > 0) ? 0 : a.shape[0] - 1, (j > 0) ? a.shape[0] : (j > 0) ? 0 : a.shape[0] - 1, j));
    for (j = 0; j < 5; j += 1)
    {
        printf("%ld\n", b.nd_int64[get_index(b, j)]);
    }
    free_array(a);
    free_pointer(b);
}
```
and there is a test for this case in the tests that i added

EmilyBourne: Is there really no way for this to be simpler?!
```
new_slice((j > 0) ? 0 : a.shape[0] - 1, (j > 0) ? a.shape[0] : (j > 0) ? 0 : a.shape[0] - 1, j)
```

saniQQ: i can't think of a more simpler way because the `stop` depends on the `start` and both depend on `j` which is **dynamic**.

yguclu: ### Potential bug?

I am a bit confused by the second argument, where you check two times for the same condition `y > 0`:
```c
(j > 0) ? a.shape[0] : (j > 0) ? 0 : a.shape[0] - 1
```
- If `j > 0` you enter the first case, that is `a.shape[0]`;
- If `j <= 0` you enter the second case, where you check again if `j > 0`... but this is false and hence you always enter the case `a.shape[0] - 1`... But I think we should print `0` instead!

In summary I think you should print
```c
(j > 0) ? a.shape[0] - 1 : 0
```
and the final expression should be
```c
new_slice((j > 0) ? 0 : a.shape[0] - 1, (j > 0) ? a.shape[0] - 1: 0, j)
```
Am I missing something here? Please let me know.

### Cleaner print

It seems to me that this expression is a bit clearer:
```c
(j > 0) ? new_slice(0, a.shape[0] - 1, j) : new_slice(a.shape[0] - 1, 0, j)
```
where a TernaryOperator node has two Slice nodes as arguments. What do you think?

saniQQ: yeah there was a bug i didn't notice it, i fixed it.
to make the `TernaryOperator` accept a `Slice`  as argument the `Slice` object needs to have a (`rank`, `shape`, `precision`, `dtype`)
if we want to add them to the `Slice` object , i could make the (`rank`, `shape`, `precision`) default to `None`, but i don't know what to make the `dtype`

EmilyBourne: I think it would make sense for `Slice` to be a `PyccelAstNode`. I would imagine that you would then have:
Dtype=int
Rank=0
Shape=()
Precision=default_precision ['int']
Order=None

But it might also make sense to have
Rank=1
Shape=(3,)

saniQQ: i think having the rank and shape 0 or `none` is more close to python and won't make a confusion to future devs

EmilyBourne (approved):

yguclu (approved):

pyccel/codegen/printing/ccode.py
yguclu: So I suppose this can be improved in a separate PR?

Issue #577, opened 2020-12-01
Operands could not be broadcast together with shape


EmilyBourne:
**Describe the bug**
Pyccel only recognises identical shapes. (This error is also raised twice)

**To Reproduce**
Provide code to reproduce the behavior:
```python
import numpy as np

a = np.zeros(10)
b = np.empty(2)
b[:] = a[2:4]+a[4:6]
```

Provide the generated code, or the error message:
```bash/fortran/c
ERROR at annotation (semantic) stage
pyccel:
 |fatal [semantic]: tmp.py| operands could not be broadcast together with shapes (PyccelMinus(LiteralInteger(4), LiteralInteger(2)),) (PyccelMinus(LiteralInteger(6), LiteralInteger(4)),)
 |fatal [semantic]: tmp.py [6]| operands could not be broadcast together with shapes (PyccelMinus(LiteralInteger(4), LiteralInteger(2)),) (PyccelMinus(LiteralInteger(6), LiteralInteger(4)),) (PyccelAdd(a[LiteralInteger(2) : LiteralInteger(4)], a[LiteralInteger(4) : LiteralInteger(6)]))
```

**Acceptance Criteria**
- [ ] The function [`broadcast`](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/operators.py#L49) should call [`pyccel_to_sympy`](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/sympy_helper.py#L90). This allows sympy to be used to reduce the shape expression to its simplest form which can be compared to another shape in a meaningful way. The original pyccel shape should be returned not the sympy version. The sympy version should be used for errors
- [ ] Only 1 error should be raised when shapes are incompatible. The error appears twice due to the use of `errors.report` [here](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/operators.py#L82) and [here](https://github.com/pyccel/pyccel/blob/master/pyccel/parser/semantic.py#L1122). Instead the first instance should merely raise a `PyccelError` which can be reported in the SemanticParser
- [ ] There are situations where shapes are not incompatible, but `broadcast` cannot be sure that they are indeed compatible; `broadcast` should not raise an error but simply return whatever information it can compute

PR #627, opened 2020-12-22
Correctly calculate the shape of slice 

MenoIy:
Correctly calculate the shape of a slice element when `start` and `stop` arguments are negative:

- Check slice's start and stop values and calculate the size using the base array shape;
- Convert step to positive if it's negative;
- Raise `ValueError` when the slice step is zero;
- Divide the result by the step and round it up if the step is not `None`.


saniQQ (approved):

EmilyBourne:
Could you add a test please
Eg
```
a=b[1:]+b[:-1]
```

MenoIy:
> Could you add a test please
> Eg
> 
> ```
> a=b[1:]+b[:-1]
> ```

@EmilyBourne  The new calculated shape is either PyccelSub,  PyccelAdd, or MathCeil to create a test I need issue #577, which will
translate Pyccel expressions to Sympy and then compare it.
@rmahjoubi

rmahjoubi (approved):

saniQQ (approved):

EmilyBourne (approved):

yguclu:
Why does it say that "This Issue is blocked by *Allow simplification of pyccel-created nodes*"?

EmilyBourne:
> Why does it say that "This Issue is blocked by _Allow simplification of pyccel-created nodes_"?

In order for the tests to work either issue #564 or #577 needed to be fixed. #577 simplifies the result with sympy which allows a comparison while #564 ensures that internal objects do not contain unnatural expressions which leads to the same result.

For example if we test the following:
`a[1:]+a[:-1]`
Before this PR the shapes were:
`PyccelMinus(Shape(a), 1)` and `PyccelUnary(1)`
After this PR the shapes are:
`PyccelMinus(Shape(a), 1)` and `PyccelAdd(Shape(a),PyccelUnary(1))`
#577 changes these to a sympy expression when checking if the shape is acceptable so it recognises that these are the same.
Alternatively #564 would mean that the shapes would have been simplified at creation to:
`PyccelMinus(Shape(a), 1)` and `PyccelMinus(Shape(a), 1)`

yguclu (approved):

PR #629, opened 2020-12-22
Better handling for array slicing with literal indexes

### LINKED ISSUE #566: basic indexing : creating a view with Integers + slice objects
in basic indexing, python has a specific behaviour when mixing slice objects and Integers.

**To Reproduce**
```python
a = numpy.array([[1, 2, 3], [4, 5, 6]])
x = a[1, 1:]
y = a[1:, 1:]

x >>> [5, 6]
y >>> [[5, 6]]
```

Here is the behaviour of python for the **basic indexing**:

- An integer `i` returns the same values as the slice `i:i+1`, except the dimensionality of the returned array view is reduced by 1. In particular, a selection tuple with the p-th element an integer (and all other entries being slices `:`) returns the corresponding sub-array with dimension N - 1. If N = 1 then the returned object is an array scalar. These objects are explained in Scalars.

- If the selection tuple has all entries being simple slices `:` except the p-th entry which is a slice object `i:j:k`, then the returned array has dimension N formed by concatenating the sub-arrays returned by integer indexing of elements i, i+k, …, i + (m - 1) k < j,

- Basic slicing with more than one non-: entry in the slicing tuple, acts like a repeated application of slicing using a single non-`:` entry, where the non-`:` entries are successively taken (with all other non-`:` entries replaced by `:`). Thus, x[ind1,...,ind2,:] acts like x[ind1][...,ind2,:] under basic slicing.

nhamidn:
This fixes #566:

- Pass the rank value to the `array_slicing` function to make sure we got the correct one;
- Handle literals indexes.


pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -189,14 +189,14 @@ t_slice new_slice(int32_t start, int32_t end, int32_t step)
...
-t_ndarray array_slicing(t_ndarray arr, ...)
+t_ndarray array_slicing(t_ndarray arr, int n, ...)
 {
     t_ndarray view;
     va_list  va;
     t_slice slice;
     int32_t start = 0;
 
-    view.nd = arr.nd;
+    view.nd = n;
```
saniQQ, line 199: you should modify the call to `va_start` to start from `n` not `arr`

pyccel/codegen/printing/ccode.py
```diff
@@ -673,8 +673,10 @@ def _print_IndexedElement(self, expr):
                     if isinstance(ind, Slice):
                         inds[i] = self._new_slice_with_processed_arguments(ind, PyccelArraySize(base, i),
                             allow_negative_indexes)
+                    else:
+                        inds[i] = Slice(ind, PyccelAdd(ind, LiteralInteger(1)), LiteralInteger(1))
```
saniQQ, line 677: did you test with the decorator `allow_negative_index`

saniQQ (changes requested):

EmilyBourne:
This is marked as `needs_initial_review` and `draft`. Could you remove the label until it is ready or mark it as ready please

nhamidn:
how negative indexes are handled in the case of a program? we can't use decorators. @saidctb @yguclu

MenoIy (approved):

EmilyBourne (changes requested):
Could you add tests for the c code translation please. These can go in the pyccel folder as epyccel does not work yet

pyccel/codegen/printing/ccode.py
EmilyBourne: I would also like the answer to this question which was resolved without an answer

tests/pyccel/scripts/arrays_view.py
```diff
@@ -39,8 +39,24 @@ def array_view_negative_variable_step():
...
 
+@allow_negative_index('a')
+def array_view_negative_var():
+    a = np.array([[1, 2, 3], [4, 5, 6], [7, 9, 5]])
+    v = -1
+    x = a[v,1:]
+    for i in range(2):
+        print(x[i])
+
+def array_view_positive_literal():
```
EmilyBourne, line 50: It seems like this test isn't used. Was that done on purpose?

nhamidn: Oups, i forgot to call the function

pyccel/codegen/printing/ccode.py
nhamidn: yes, i tested with the `allow_negative_index` decorator, it's working now.

EmilyBourne: I looked at an example:
```
@allow_negative_index('a')
def array_view_negative_var():    
    a = np.array([[1, 2, 3], [4, 5, 6], [7, 9, 5]])
    v = -1
    x = a[v,1:]
```
is translated to:
```
    t_ndarray a;
    int64_t v;
    t_ndarray x;
    int64_t i;

    a = array_create(2, (int32_t[]){3, 3}, nd_int64);
    int64_t array_dummy_0006[] = {1, 2, 3, 4, 5, 6, 7, 9, 5};
    memcpy(a.nd_int64, array_dummy_0006, a.buffer_size);

    v = -1;
    x = array_slicing(a, 1, new_slice((v < 0) ? 3 + v : v, (v < 0) ? 3 + v : v + 1, 1), new_slice(1, a.shape[1], 1));
```
So for the negative index the new slice is:
```
new_slice(3 + v, 3 + v, 1)
```
whereas for the positive index the new slice is:
```
new_slice(v, v + 1, 1)
```
Can you explain this?

EmilyBourne: Is it possible that this is a parenthesis issue? I.e.: `(v < 0) ? 3 + v : v + 1` is equivalent to `((v < 0) ? 3 + v : v) + 1`?

EmilyBourne: This is not the case (see issue #648). Given that fact why does `new_slice(3 + v, 3 + v, 1)` give a  slice of size 1 and not size 0?

nhamidn: `new_slice(3 + v, 3 + v, 1)` must give a slice of size 0 in this case, it should be `new_slice(3 + v, 3 + v + 1, 1)`, i think the problem is from the `_new_slice_with_processed_arguments` method which was introduced on another pull request.
@saniQQ is this a problem? using `new_slice(3 + v, 3 + v, 1)` and not `new_slice(3 + v, 3 + v + 1, 1)` for negative index ?

EmilyBourne: It doesn't appear to be a problem as it works. But I can't seem to work out how it works. `new_slice(3 + v, 3 + v, 1)` does seem to be returning a slice of size 1 but `new_slice(v,v,1)` correctly returns size 0. I have probably missed something but I think it's worth checking what before I approve

saniQQ: i think there is a bug related to the pull request that u mentioned as it should be `new_slice((v < 0) ? 3 + v : v, ((v < 0) ? 3 + v : v) + 1, 1)` there is a missing parenthesis the addition should be around the ternary
as of why they seem to have different sizes is just the fact that one is not what it's supposed to be + we are not checking for the bounds.

EmilyBourne: Ok that makes sense as the memory is shared. I did test the shape, but probably I didn't check if we were using the ndarray shape or just the shape that pyccel calculated. If we all agree that the current code is wrong then this needs fixing before we merge. It would also be useful if the code could be tested with the fix to #648 (even if it's only locally) before merging

nhamidn: i think this should be marked as resolved, I have tested using the fix of #648 and the result was :

```
void array_view_negative_var(void)
{
    t_ndarray a;
    int64_t v;
    t_ndarray x;

    a = array_create(2, (int32_t[]){3, 3}, nd_int64);
    int64_t array_dummy_0001[] = {1, 2, 3, 4, 5, 6, 7, 9, 5};

    v = -1;
    x = array_slicing(a, 1, new_slice(v < 0 ? 3 + v : v, (v < 0 ? 3 + v : v) + 1, 1), new_slice(1, a.shape[1], 1));
    free_array(a);
    free_pointer(x);
}
```

EmilyBourne (approved):

saniQQ (approved):

yguclu (approved):

PR #631, opened 2020-12-22
Support Python style type hints

saidctb:
Accept Python annotations. For example:

```python
def fib(n: int) -> int:
    if n<=1:
        return 0
    elif n==2:
        return 1
    else:
        return fib(n-1) + fib(n-2)
```

### DIFF
--- a/pyccel/parser/syntactic.py
+++ b/pyccel/parser/syntactic.py
@@ -394,7 +394,7 @@ class SyntaxParser(BasicParser):
                 else:
                     arg = Argument(a.arg, kwonly=True, annotation=annotation)
 
-            arguments.append(arg)
+                arguments.append(arg)
 
         return arguments
 
--- a/pyccel/parser/syntactic.py
+++ b/pyccel/parser/syntactic.py
@@ -657,6 +657,9 @@ class SyntaxParser(BasicParser):
                 annotated_args.append(a.argument.annotation)
 
         if all(not a==Nil() for a in annotated_args):
+            if stmt.returns:
+                returns = ValuedArgument(Symbol('results'),self._visit(stmt.returns))
+                annotated_args.append(returns)
             decorators['types'] = [Function('types')(*annotated_args)]
 
         for d in self._visit(stmt.decorator_list):
--- a/pyccel/parser/syntactic.py
+++ b/pyccel/parser/syntactic.py
@@ -649,7 +649,13 @@ class SyntaxParser(BasicParser):
         decorators = {}
 
         # add the decorator @types if the arguments are annotated
-        annotated_args = [a.annotation for a in arguments]
+        annotated_args = []
+        for a in arguments:
+            if isinstance(a, Argument):
+                annotated_args.append(a.annotation)
+            elif isinstance(a, ValuedArgument):
+                annotated_args.append(a.argument.annotation)
+
         if all(not a==Nil() for a in annotated_args):
             decorators['types'] = [Function('types')(*annotated_args)]
 
--- a/pyccel/parser/syntactic.py
+++ b/pyccel/parser/syntactic.py
@@ -372,18 +372,22 @@ class SyntaxParser(BasicParser):
         return expr
 
     def _visit_arguments(self, stmt):
-        arguments = []
+
         if stmt.vararg or stmt.kwarg:
             errors.report(VARARGS, symbol = stmt,
                     severity='fatal')
 
+        arguments       = []
         if stmt.args:
             n_expl = len(stmt.args)-len(stmt.defaults)
-            arguments += [Argument(a.arg) for a in stmt.args[:n_expl]]
-            arguments += [ValuedArgument(Argument(a.arg),self._visit(d)) for a,d in zip(stmt.args[n_expl:],stmt.defaults)]
+            positional_args        = [Argument(a.arg, annotation=self._visit(a.annotation)) for a in stmt.args[:n_expl]]
+            valued_arguments       = [ValuedArgument(Argument(a.arg, annotation=self._visit(a.annotation)),\
+                                      self._visit(d)) for a,d in zip(stmt.args[n_expl:],stmt.defaults)]
+            arguments              = positional_args + valued_arguments
         if stmt.kwonlyargs:
-            arguments += [ValuedArgument(Argument(a.arg),self._visit(d), kwonly=True) if d is not None
-                        else Argument(a.arg, kwonly=True) for a,d in zip(stmt.kwonlyargs,stmt.kw_defaults)]
+            kwonly_arguments       = [ValuedArgument(Argument(a.arg, annotation=self._visit(a.annotation)),self._visit(d), kwonly=True) if d is not None
+                                     else Argument(a.arg, kwonly=True) for a,d in zip(stmt.kwonlyargs,stmt.kw_defaults)]
+            arguments       += kwonly_arguments
 
         return arguments
 
@@ -643,6 +647,12 @@ class SyntaxParser(BasicParser):
             return container
 
         decorators = {}
+
+        # add the decorator @types if the arguments are annotated
+        annotated_args = [a.annotation for a in arguments]
+        if all(not a==Nil() for a in annotated_args):
+            decorators['types'] = [Function('types')(*annotated_args)]
+
         for d in self._visit(stmt.decorator_list):
             tmp_var = str(d) if isinstance(d, Symbol) else str(type(d))
             if tmp_var in decorators:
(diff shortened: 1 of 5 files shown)


pyccel/parser/syntactic.py
```diff
@@ -372,18 +372,22 @@ def _visit_AugAssign(self, stmt):
...
-            arguments += [Argument(a.arg) for a in stmt.args[:n_expl]]
-            arguments += [ValuedArgument(Argument(a.arg),self._visit(d)) for a,d in zip(stmt.args[n_expl:],stmt.defaults)]
+            positional_args        = [Argument(a.arg, annotation=self._visit(a.annotation)) for a in stmt.args[:n_expl]]
+            valued_arguments       = [ValuedArgument(Argument(a.arg, annotation=self._visit(a.annotation)),\
+                                      self._visit(d)) for a,d in zip(stmt.args[n_expl:],stmt.defaults)]
+            arguments              = positional_args + valued_arguments
         if stmt.kwonlyargs:
-            arguments += [ValuedArgument(Argument(a.arg),self._visit(d), kwonly=True) if d is not None
-                        else Argument(a.arg, kwonly=True) for a,d in zip(stmt.kwonlyargs,stmt.kw_defaults)]
+            kwonly_arguments       = [ValuedArgument(Argument(a.arg, annotation=self._visit(a.annotation)),self._visit(d), kwonly=True) if d is not None
```
EmilyBourne, line 388: This line barely fits on screen. Could you break it up a bit more? Or maybe there should be a function for this?

pyccel/parser/syntactic.py
```diff
@@ -643,6 +647,21 @@ def fill_types(ls):
             return container
 
         decorators = {}
+
+        # add the decorator @types if the arguments are annotated
```
EmilyBourne, line 651: If the user has provided a decorator and adds type annotations do we get a warning? Should we?

pyccel/parser/syntactic.py
```diff
@@ -643,6 +647,21 @@ def fill_types(ls):
...
+        annotated_args = []
+        for a in arguments:
+            if isinstance(a, Argument):
+                annotated_args.append(a.annotation)
+            elif isinstance(a, ValuedArgument):
+                annotated_args.append(a.argument.annotation)
```
EmilyBourne, line 657: Would this not be faster with list comprehensions. Ie:
```
annotated_args = [a.annotation if isinstance(a, Argument) else a.argument.annotation \
                  for a in arguments if isinstance(a, (Argument, ValuedArgument)]
```
or perhaps better:
```
annotated_args  = [a.annotation for a in arguments if isinstance(a, Argument)]
annotated_args += [a.argument.annotation for a in arguments if isinstance(a, ValuedArgument)]
```

pyccel/parser/syntactic.py
```diff
@@ -643,6 +647,21 @@ def fill_types(ls):
...
+
+        if all(not a==Nil() for a in annotated_args):
```
EmilyBourne, line 659: Is `__eq__` implemented for `Nil()`? I think we usually write `isinstance(a, Nil)`

tests/epyccel/modules/python_annotations.py
```diff
@@ -0,0 +1,41 @@
...
+    from numpy import full, int32
+    z = full((2,3),5,order='F', dtype=int32)
+    x[:] = (x // y) * x + z
+
+def array_real_1d_complex_3d_expr( x:'real[:]', y:'real[:]' ):
+    from numpy import full
+    z = full(3,5)
+    x[:] = (x // y) * x + z
+
+def fib(n: int) -> int:
```
EmilyBourne, line 34: Does the following work:
```
def fib(n: np.int32) -> np.int32:
```
?

tests/epyccel/test_epyccel_python_annotations.py
```diff
@@ -0,0 +1,8 @@
+# pylint: disable=missing-function-docstring, missing-module-docstring/
+from pyccel.epyccel import epyccel
+
+def test_module_1():
+    import modules.python_annotations as mod
+
+    modnew = epyccel(mod)
+    assert modnew.fib(10) == mod.fib(10)
```
EmilyBourne, line 8: Why are you only using one of the test functions you wrote?

EmilyBourne (changes requested):

pyccel/parser/syntactic.py
saidctb: it will be an interface, if it's duplicated we will get a warning

pyccel/parser/syntactic.py
saidctb: yes you are right I can break it into two lines

pyccel/parser/syntactic.py
saidctb: I didn't do it, because it becomes unreadable

tests/epyccel/modules/python_annotations.py
saidctb: not yet, it works like the @types decorator

tests/epyccel/test_epyccel_python_annotations.py
saidctb: I only test the compilation, there is no need to test the results

pyccel/parser/syntactic.py
saidctb: done

tests/epyccel/modules/python_annotations.py
EmilyBourne: Have you tried it? I think we can use `np.int32` as `dtype` for arrays so the semantics should understand it

tests/epyccel/test_epyccel_python_annotations.py
EmilyBourne: It doesn't do any harm though, just in case the types are misinterpreted. Even if I don't see how they could be

pyccel/parser/syntactic.py
```diff
@@ -372,18 +372,29 @@ def _visit_AugAssign(self, stmt):
...
-                        else Argument(a.arg, kwonly=True) for a,d in zip(stmt.kwonlyargs,stmt.kw_defaults)]
+            for a,d in zip(stmt.kwonlyargs,stmt.kw_defaults):
+                annotation = self._visit(a.annotation)
+                if d is not None:
+                    arg = Argument(a.arg, annotation=annotation)
+                    arg = ValuedArgument(arg, self._visit(d), kwonly=True)
+                else:
+                    arg = Argument(a.arg, kwonly=True, annotation=annotation)
+
+            arguments.append(arg)
```
EmilyBourne, line 397: Should this be in the for loop?

tests/epyccel/modules/python_annotations.py
saidctb: it's needs much more work because everything is converted to the header syntax , so we need to modify the headers in order to understand the language, or abandon the headers

tests/epyccel/test_epyccel_python_annotations.py
saidctb: okay

pyccel/parser/syntactic.py
saidctb: ah yes you are right

tests/epyccel/modules/python_annotations.py
saidctb: if the following works then it will automatically work here,
`#$ header function f(np.int32) results(np.int32)`
or
`@types(np.int32, np.int32)`

saidctb: what I'm doing now is convert the python type hints to the `@types` decorator and then it gets converted to a header string in the syntactic stage.
one way to solve this issue is to make the conversion in the semantics stage from the `@types` to the header language, with this we will avoid updating the header language.

EmilyBourne: ok

saidctb: I created this issue  #633 to solve this problem

pyccel/parser/syntactic.py

tests/epyccel/test_epyccel_python_annotations.py
saidctb: done

pyccel/parser/syntactic.py
yguclu: Actually I find list comprehensions very readable in general, but maybe not much so in this case.

The second suggestion of @EmilyBourne changes the order of the elements in `annotated_args` compared to the original code of @saidctb. Is that a problem?

saidctb: yes we need to keep the order

EmilyBourne (approved):

yguclu (approved):

PR #632, opened 2020-12-23
Add support of Numpy arrays as function arguments in C language

### LINKED ISSUE #650: ndarray.h is not imported if array is declared as an argument
**Describe the bug**
The header file `ndarray.h` is not imported if the array is declared as the argument to a function

**To Reproduce**
```python
import numpy as np

@types('int[:]')
def f(x):
    print(np.shape(x)[0])
```

Provide the generated code:
tmp.h:
```c
#ifndef TMP_H
#define TMP_H

#include <stdlib.h>
#include <stdint.h>

void f(t_ndarray x);

#endif // TMP_H
```
tmp.c:
```c
#include <tmp.h>
#include <stdio.h>
#include <stdlib.h>
#include <stdint.h>

void f(t_ndarray x)
{
    printf("%ld\n", x.shape[0]);
}
```

**Expected behavior**
The import of `ndarray.h` should be carried out when `t_ndarray` is written.

**Language**
C

rakati:
Add support of Numpy arrays as function arguments in C language by:
 - Create a cast function that cast numpy arrays to ndarrays.
 - Create a function that converts a numpy strides array to ndarrays strides.
 - Cast numpy array to ndarray in the wrapper.
 - Improve Deallocator to be used for free other types.
 - Fix the DottedVariable printer.
 - Add tests to make sure that casting is working correctly.
 - Activate the test that needs these changes.
closes #650 #509


pyccel/ast/cwrapper.py
```diff
@@ -285,6 +286,15 @@ def args(self):
                         arguments = [Variable(dtype=PyccelPyObject(), name = 'o', is_pointer=True)],
                         results   = [Variable(dtype=NativeReal(), name = 'r')])
 
+#-------------------------------------------------------------------
+#                           C functions
+#-------------------------------------------------------------------
+# call c free function for freeing allocated variables
+C_Free = FunctionDef(name      = 'free',
+                           body      = [],
+                           arguments = [Variable(dtype=NativeGeneric(), name = 'o', is_pointer=True)],
```
MenoIy, line 295: Can you use NativeVoid here?

MenoIy, line 295: can you use NativeVoid here?

pyccel/ast/cwrapper.py
```diff
@@ -321,36 +336,51 @@ def args(self):
                        arguments = [Variable(dtype=PyccelPyArrayObject(), name = 'o', is_pointer=True)],
                        results   = [Variable(dtype=PyccelPyObject(), name = 'i')])
 
+numpy_get_shape = FunctionDef(name      = 'PyArray_SHAPE',
+                       body      = [],
+                       arguments = [Variable(dtype=PyccelPyArrayObject(), name = 'o', is_pointer=True)],
+                       results   = [Variable(dtype=PyccelPyObject(), name = 'i')])
```
MenoIy, line 342: NativeInteger

pyccel/ast/cwrapper.py
```diff
@@ -525,12 +555,53 @@ def pybool_to_bool(cast_function_name):
...
+                       body      = cast_function_body,
+                       results   = [cast_function_result])
+
 cast_function_registry = {
     'pyint_to_bool' : pyint_to_bool,
     'bool_to_pyobj' : bool_to_pyobj,
     'pycomplex_to_complex' : pycomplex_to_complex,
     'complex_to_pycomplex': complex_to_pycomplex,
     'pybool_to_bool' : pybool_to_bool,
+    'pyccelPyArrayObject_to_ndarray' : pyccelPyArrayObject_to_ndarray,
```
MenoIy, line 604: maybe you can name it  pyarray_to_ndarray?

pyccel/codegen/printing/ccode.py
```diff
@@ -597,6 +598,8 @@ def _print_FuncAddressDeclare(self, expr):
 
     def _print_Declare(self, expr):
         declaration_type = self.get_declare_type(expr.variable)
+        if declaration_type == "t_ndarrays":
+            self._additional_imports.add('ndarrays')
```
MenoIy, line 602: do you need to re-import ndarrays here? 
```
declaration_type = self.get_declare_type(expr.variable)
```
I think its already done here

pyccel/codegen/printing/ccode.py
```diff
@@ -642,6 +645,7 @@ def function_signature(self, expr):
             return '{0}{1}({2})'.format(ret_type, name, arg_code)
 
     def _print_IndexedElement(self, expr):
+        self._additional_imports.add('ndarrays')
```
MenoIy, line 648: same above, did you try to remove it?

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -874,6 +882,12 @@ def _print_FunctionDef(self, expr):
 
         # Call free function for python type
         wrapper_body += [FunctionCall(Py_DECREF, [i]) for i in self._to_free_PyObject_list]
+        # Call free function for C type
+        if self._target_language == 'c':
+            wrapper_body += [FunctionCall(C_Free, [DottedVariable(i,
```
MenoIy, line 887: you should do the same for interfaces

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -405,7 +409,11 @@ def _body_array(self, variable, collect_var, check_type = False) :
                 error = PyErr_SetString('PyExc_NotImplementedError',
                         '"Argument does not have the expected ordering ({})"'.format(collect_var.order))
                 body += [(PyccelNot(check), [error, Return([Nil()])])]
-        body += [(LiteralTrue(), [Assign(VariableAddress(variable),
+        if self._target_language == "c":
+            cast_function = self.get_cast_function_call('pyccelPyArrayObject_to_ndarray', collect_var)
+            body += [(LiteralTrue(), [Assign(VariableAddress(variable), cast_function)])]
+        else:
+            body += [(LiteralTrue(), [Assign(VariableAddress(variable),
```
EmilyBourne, line 416: I'm not sure why this if is necessary here. Isn't this collecting the cast function? Is that not what `get_collect_function_call` is designed to do?

pyccel/codegen/printing/cwrappercode.py
EmilyBourne, line 887: Where are these objects allocated?

pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -257,3 +258,18 @@ int32_t     get_index(t_ndarray arr, ...)
     va_end(va);
     return (index);
 }
+
+/*
+** convert numpy strides to nd_arrya strdies, and return it in a new array, to
```
EmilyBourne, line 263: Typos
`nd_arry`->`nd_array`
`strdies`->`strides`

pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -257,3 +258,18 @@ int32_t     get_index(t_ndarray arr, ...)
     va_end(va);
     return (index);
 }
+
+/*
+** convert numpy strides to nd_arrya strdies, and return it in a new array, to
+** avoid the problem of difference implemation of strides in numpy and ndarray.
```
EmilyBourne, line 264: Grammar:
`difference implementation`->`different implementations`

pyccel/stdlib/ndarrays/ndarrays.h
```diff
@@ -32,17 +32,21 @@ typedef struct  s_slice
...
 }               t_slice;
 
+/*
+** Map e_types enum to numpy NPY_TYPES enum
+** ref: numpy_repo: numpy/numpy/core/include/numpy/ndarraytypes.h
+*/
 enum e_types
 {
         nd_bool,
         nd_int8,
```
EmilyBourne, line 42: Maybe it would be better to provide numbers for everything in this enum to be consistent

pyccel/stdlib/ndarrays/ndarrays.h
```diff
@@ -63,9 +67,9 @@ typedef struct  s_ndarray
     /* number of dimensions */
     int32_t                 nd;
     /* shape 'size of each dimension' */
-    int32_t                 *shape;
+    long int                *shape;
```
EmilyBourne, line 70: Why has the type changed?

pyccel/stdlib/ndarrays/ndarrays.h
```diff
@@ -63,9 +67,9 @@ typedef struct  s_ndarray
     /* number of dimensions */
     int32_t                 nd;
     /* shape 'size of each dimension' */
-    int32_t                 *shape;
+    long int                *shape;
     /* strides 'number of bytes to skip to get the next element' */
-    int32_t                 *strides;
+    long int                *strides;
```
EmilyBourne, line 72: See above. Why has the type changed. This will lead to truncation warnings on windows computers

tests/epyccel/test_array_as_func_args.py
```diff
@@ -0,0 +1,144 @@
...
+        d1 = randint(1, 15)
+        d2 = randint(1, 15)
+        x1 = uniform(np.finfo(t).max / 4, size=(d1, d2)) + uniform(np.finfo(t).max / 4, size=(d1, d2)) * 1j
+        x2 = np.copy(x1)
+        a = uniform(np.finfo(t).max / 4) + uniform(np.finfo(t).max / 4) * 1j
+
+        f1(x1, a, d1, d2)
+        f2(x2, a, d1, d2)
+
+    assert np.array_equal( x1, x2 )
```
EmilyBourne, line 144: Are there no other tests which could now be activated?

tests/epyccel/test_array_as_func_args.py
```diff
@@ -0,0 +1,144 @@
...
+    for t in int_types:
+        size = randint(1, 30)
+        x1 = randint(np.iinfo(t).max / 2, size=size, dtype=t)
+        x2 = np.copy(x1)
+        a = randint(np.iinfo(t).max / 2, dtype=t)
+
+        f1(x1, a, size)
+        f2(x2, a, size)
+
+    assert np.array_equal( x1, x2 )
```
EmilyBourne, line 36: Shouldn't this assertion be in the for loop?

tests/epyccel/test_array_as_func_args.py
```diff
@@ -0,0 +1,144 @@
...
+    for t in float_types:
+        size = randint(1, 30)
+        x1 = uniform(np.finfo(t).max / 2, size=size)
+        x2 = np.copy(x1)
+        a = uniform(np.finfo(t).max / 2)
+
+        f1(x1, a, size)
+        f2(x2, a, size)
+
+    assert np.array_equal( x1, x2 )
```
EmilyBourne, line 56: See above. Assertion should be in loop

EmilyBourne (changes requested):

pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -257,3 +258,18 @@ int32_t     get_index(t_ndarray arr, ...)
...
+long int    *numpy_to_ndarray_strides(long int *np_strides, int type_size, int nd)
+{
+    long int *ndarray_strides;
+
+    ndarray_strides = (long int*)malloc(sizeof(long int) * nd);
+    for (int i = 0; i < nd; i++)
+        ndarray_strides[i] = np_strides[i] / type_size;
+    return ndarray_strides;
+
+}
```
MenoIy, line 275: can you explain what this function does and why you need to allocate?

pyccel/ast/cwrapper.py
rakati: yes, I will

pyccel/codegen/printing/cwrappercode.py
rakati: I forgot about it.

saniQQ:
i see that you are allocating the strides so they are not pointing to the same one as python uses.
are you updating the strides after leaving the function, so when the strides are changed inside the function they are  updated in python.

EmilyBourne:
> i see that you are allocating the strides so they are not pointing to the same one as python uses.
> are you updating the strides after leaving the function, so when the strides are changed inside the function they are  updated in python.

Is it possible to change the strides in place in python?

saniQQ:
> > i see that you are allocating the strides so they are not pointing to the same one as python uses.
> > are you updating the strides after leaving the function, so when the strides are changed inside the function they are  updated in python.
> 
> Is it possible to change the strides in place in python?

yes we can have a function like 
```python
def ex(arr):
     arr.shape = (2, 4)

a = np.array([1, 2, 3, 4, 5, 6, 7, 8])
ex(arr)
a 
>>> [[1, 2, 3, 4], [5, 6, 7, 8]]
a.shape 
>>> (2, 4)
a.strides 
>>> (32, 8)

File: pyccel/codegen/printing/cwrappercode.py (hunk 6 above)
Comment (rakati) [reply pyccel/codegen/printing/cwrappercode.py:887]: in the ndarrays library, we free the ndarrays strides after finishing work with the casted numpy array.
check this function `numpy_to_ndarray_strides` located in `stdlib/ndarrays/ndarrays.c`

File: tests/epyccel/test_array_as_func_args.py (hunk 13 above)
Comment (rakati) [reply tests/epyccel/test_array_as_func_args.py:144]: no, I search a lot. but it seems like our C Numpy still poor.
we need to boost that by creating many things.
the tests I added are temporary, we will remove them in the future, after enriching the ccode.

File: tests/epyccel/test_array_as_func_args.py (hunk 14 above)
Comment (rakati) [reply tests/epyccel/test_array_as_func_args.py:36]: yap, I missed it.

File: tests/epyccel/test_array_as_func_args.py [hunk 17]
```diff
@@ -0,0 +1,143 @@
...
+import numpy as np
+from numpy.random import randint, uniform
+
+from pyccel.epyccel import epyccel
+from pyccel.decorators import types
+
+int_types = ['int8', 'int16', 'int32', 'int64']
+float_types = ['float32', 'float64']
+
+def test_array_int_1d_scalar_add(language):
```
saidctb, line 15: why did you create new tests ?, can't you just use the one we use for fortran and mark the one that doen't work xfail for the c language

rakati: no, I can't .

rakati: the current tests are using some features not implemented yet in c.

EmilyBourne: I think there are some tests at the end of test_arrays. Isn't there a fizz buzz one that should work?

EmilyBourne: It looks like the existing tests that might work are in [loops.py](https://github.com/pyccel/pyccel/blob/master/tests/epyccel/modules/loops.py)

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -87,7 +87,11 @@ def function_signature(self, expr):
     def get_declare_type(self, expr):
         dtype = self._print(expr.dtype)
         prec  = expr.precision
-        dtype = self.find_in_dtype_registry(dtype, prec)
+        if expr.rank > 0 and self._target_language == 'c' and not expr.is_pointer:
+            self._additional_imports.add("ndarrays")
```
EmilyBourne, line 91: Maybe it would be better to do:
```
if self._target_language == 'c':
    return CCodePrinter.get_declare_type(self, expr)
```
So we don't need to maintain this condition in 2 places

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: If you allocate the `shape` you also need to free this. Wouldn't using the `Deallocate` node and marking this variable as a pointer get the correct deallocation?

pyccel/ast/cwrapper.py
```diff
@@ -525,12 +555,53 @@ def pybool_to_bool(cast_function_name):
...
+    length      = DottedVariable(cast_function_result,
+                        Variable(dtype=NativeInteger(), name='length'))
+    buffer_size = DottedVariable(cast_function_result,
+                        Variable(dtype=NativeInteger(), name='buffer_size'))
+    is_view    = DottedVariable(cast_function_result,
+                        Variable(dtype=NativeBool(), name='is_view'))
+
+    cast_function_body = [Assign(nd, FunctionCall(numpy_get_ndims, [cast_function_argument])),
+                          Assign(raw_data, FunctionCall(numpy_get_data, [cast_function_argument])),
+                          Assign(shape, FunctionCall(numpy_get_shape, [cast_function_argument])),
```
EmilyBourne, line 583: Do the types match here?

saidctb:
> > > i see that you are allocating the strides so they are not pointing to the same one as python uses.
> > > are you updating the strides after leaving the function, so when the strides are changed inside the function they are  updated in python.
> > 
> > 
> > Is it possible to change the strides in place in python?
> 
> yes we can have a function like
> 
> ```python
> def ex(arr):
>      arr.shape = (2, 4)
> 
> a = np.array([1, 2, 3, 4, 5, 6, 7, 8])
> ex(arr)
> a 
> >>> [[1, 2, 3, 4], [5, 6, 7, 8]]
> a.shape 
> >>> (2, 4)
> a.strides 
> >>> (32, 8)
> ```
> 
> the strides always changes if the shape is changed and we can eve oly change the strides like :
> 
> ```python
> a = np.array([1, 2, 3, 4, 5, 6, 7, 8])
> b = a[::2]
> b.strides >>> (16,)
> b >>> [1, 3, 5, 7]
> b.strides = (8,)
> b >>> [1, 2, 3, 4]
> ```

as discussed in the meeting, we won't allow the change of the shape and strides like in this example

saidctb:
why is this PR not fixed yet?

pyccel/codegen/printing/cwrappercode.py
rakati: we can't handle all c stuff in the `CCodePrinter`. that why we create `cwrappercode ` for things related to wrapper.

pyccel/ast/cwrapper.py
rakati: now yes.

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: Then can you explain why ndarrays only needs including in the wrapper?

pyccel/ast/core.py
```diff
@@ -884,6 +884,8 @@ class Deallocate(Basic):
     variable : pyccel.ast.core.Variable
         The typed variable (usually an array) that needs memory deallocation.
 
+    free_function: bool
```
EmilyBourne, line 887: Are you sure this is the right type?

pyccel/codegen/printing/ccode.py
```diff
@@ -966,7 +973,8 @@ def _print_FunctionCall(self, expr):
          # Ensure the correct syntax is used for pointers
         args = []
         for a, f in zip(expr.args, func.arguments):
-            if isinstance(a, Variable) and self.stored_in_c_pointer(f):
+            if isinstance(a, Variable) and self.stored_in_c_pointer(f)\
+                and not isinstance(a, DottedVariable):
```
EmilyBourne, line 977: Why not Dotted?

pyccel/ast/core.py
rakati: hhh, it's the copy-paste work.

pyccel/codegen/printing/ccode.py
rakati: sorry, I forgot about it?

pyccel/codegen/printing/cwrappercode.py
rakati: we are using ndarray type, so we need to include the header.

EmilyBourne: But this is also the case outside the CWrapper isn't it? The aim is for the `CWrapperCodePrinter` to fall back to the `CCodePrinter` wherever possible as this minimises code duplication and ensures that if something changes in the condition (e.g. we decide to have an ndarray file per rank, or to split ndarray somehow) then we only need to fix it in one place. So `CWrapperCodePrinter` should only implement more complicated logic if the equivalent does not exist in the `CCodePrinter`.

This function is already larger than I would like for something which has a direct equivalent in the `CCodePrinter`. I created the function because I needed to avoid the different handling of the rank, but you can already see the danger of it not falling back to c from the fact that the cwrapper is potentially missing the `stdint` import.

If the import of `ndarrays` here is something that only needs doing in the cwrapper then the code is fine as it is. If it is specific to c then calling `CCodePrinter.get_declare_type` is safer

pyccel/ast/cwrapper.py
EmilyBourne: On my computer numpy shapes are stored with type `npy_intp`. According to [this](https://numpy.org/doc/1.17/reference/c-api.dtype.html#c.NPY_INTP), it is platform dependant. As we always used fixed precision we are not platform dependant. I think it would be safer to copy, unless I'm looking at the wrong line?

rakati: I agree with you, I see that `npy_intp`  is a platform dependant.

rakati: do the tests passed in your machine?

pyccel/codegen/printing/cwrappercode.py
rakati: I tried to remove these conditions but I faced many problems because we don't have a proper type of t_ndarray in the ast to differentiate between pyccel array object and ndarray type.

pyccel/ast/cwrapper.py
EmilyBourne: They pass on my machine, and they seem to pass on windows too according to github which is a good sign. But I still think a copy would be better. You never know if it could change in the future and that way all platforms should work.

Plus if we copy the shape then we don't need the strange logic to call `free` we can just use the garbage collector implemented and treat the object as a pointer

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: Presumably the `is_ndarray` property is not sufficient for this?

rakati: in ccode it's sufficient, but not in the wrapper, because `pyccelarrayobject` has the same properties of ndarray.
I see that we need to make the NativeGeneric type in such a way that can accept non-standard types (ndarray).
what do you think?

pyccel/ast/cwrapper.py
rakati: I wonder how you did solve this problem in Fortran?

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -877,6 +897,20 @@ def _print_FunctionDef(self, expr):
...
+            to_free = ["strides", "shape"]
+            wrapper_body += [Deallocate(
+                                DottedVariable(
+                                            NativeInteger(),
+                                                member,
+                                                lhs=i.name,
+                                                is_pointer=True
+                                            ),
+                                free_function="free",
+                                )
```
EmilyBourne, line 911: Doesn't this do the same thing as calling `Deallocate` on an array pointer?

rakati: yes, but I made some changes to the deallocate class to work for other types.

EmilyBourne: Yes I know. My question is about whether those changes are necessary?

saniQQ: If you need to free both the `strides` and `shape` i don't think it's necessary to change the Deallocation node when we discussed this earlier I thought that you need to free the `strides` **alone**.

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -877,6 +889,9 @@ def _print_FunctionDef(self, expr):
 
         # Call free function for python type
         wrapper_body += [FunctionCall(Py_DECREF, [i]) for i in self._to_free_PyObject_list]
+        # Call free function for C type
+        if self._target_language == 'c':
+            wrapper_body += [Deallocate(i) for i in expr.arguments if i.rank > 0]
```
EmilyBourne, line 894: Here (FunctionDef) you deallocate if `i.rank > 0` but above (Interface) you deallocate if `i.allocatable`. Is there any reason for this difference?

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -632,6 +638,9 @@ def _print_Interface(self, expr):
             # Building PybuildValue and freeing the allocated variable after.
             mini_wrapper_func_body.append(AliasAssign(wrapper_results[0],PyBuildValueNode(res_args)))
             mini_wrapper_func_body += [FunctionCall(Py_DECREF, [i]) for i in self._to_free_PyObject_list]
+            # Call free function for C type
+            if self._target_language == 'c':
+                mini_wrapper_func_body += [Deallocate(i) for i in func.arguments if i.allocatable]
```
saniQQ, line 643: does the function arguments have the attr `allocatable` == 'true'?

shouldn't it be set as `false` and let the argument just be a pointer?

pyccel/codegen/printing/cwrappercode.py
rakati: no, there is no difference, the variable is allocatable if rank > 0, so I check directly the rank.

EmilyBourne: Then I think you should do the same thing in both places

pyccel/codegen/printing/cwrappercode.py
rakati: any variable with rank > 0 is set as allocatable, so checking for rank > 0 or allocatable == true for detecting array is the same thing in the c wrapper, look at the deallocator print inside `cwrappercode ` it is different then ccode deallocator print.

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -578,11 +584,12 @@ def _print_Interface(self, expr):
             mini_wrapper_func_body = []
             res_args = []
             mini_wrapper_func_vars = {a.name : a for a in func.arguments}
+            local_arg_vars = [a.clone(a.name, is_pointer=True) if hasattr(a, "rank") and a.rank > 0 else a for a in func.arguments]
```
EmilyBourne, line 587: It would be better to check if `isinstance(a, Variable)` instead of `hasattr(a, "rank")` that way you check the existence of `a.rank` and `a.clone`

saniQQ: is there no way to change this from the sematic stage when we create the `functionDef` node?

rakati: the cwrapper doesn't go through the semantic stage.

EmilyBourne: > is there no way to change this from the sematic stage when we create the functionDef node?

The same `FunctionDef` is fed to `cwrappercode` and to `ccode`/`fcode`. `ccode`/`fcode` prints it in the expected way, while `cwrappercode` uses the information in the `FunctionDef`'s signature to print a wrapper. If we change the `FunctionDef` itself then we will also change how we print the code in c or fortran which is not what we want

EmilyBourne (approved):

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -782,11 +792,14 @@ def _print_FunctionDef(self, expr):
         # Save all used names
         used_names = set([a.name for a in expr.arguments] + [r.name for r in expr.results] + [expr.name.name])
 
+        # update ndarray local variables properties
+        local_arg_vars = [a.clone(a.name, is_pointer=True) if isinstance(a, Variable) and a.rank > 0 else a for a in expr.arguments]
```
saniQQ, line 796: will the variable `a` have `allocatable` == `true` and `is_pointer` == `true` ?

rakati: yes, you are right, it should not have allocatable and is_pointer both True.
I'll fix it.

saniQQ (approved):

EmilyBourne (approved):

mostafaMamoni (approved):

pyccel/ast/cwrapper.py
```diff
@@ -528,12 +548,56 @@ def pybool_to_bool(cast_function_name):
...
+    strides     = DottedVariable(NativeInteger(), 'strides',
+                                lhs=cast_function_result.name)
+    arr_type    = DottedVariable(NativeInteger(), 'type',
+                                lhs=cast_function_result.name)
+    length      = DottedVariable(NativeInteger(), 'length',
+                                lhs=cast_function_result.name)
+    buffer_size = DottedVariable(NativeInteger(), 'buffer_size',
+                                lhs=cast_function_result.name)
+    is_view     = DottedVariable(NativeInteger(), 'is_view',
+                                lhs=cast_function_result.name)
```
yguclu, line 574: Could you make this block a bit more readable please? Moreover, you are calling the same functions many times.

You may want to copy and paste this:
```python
    Int = NativeInteger()
    Gen = NativeGeneric()
    lhs = cast_function_result.name

    nd          = DottedVariable(Int,          'nd', lhs=lhs)
    raw_data    = DottedVariable(Gen,    'raw_data', lhs=lhs, rank=1)
    shape       = DottedVariable(Int,       'shape', lhs=lhs, is_pointer=True)
    type_size   = DottedVariable(Int,   'type_size', lhs=lhs)
    strides     = DottedVariable(Int,     'strides', lhs=lhs)
    arr_type    = DottedVariable(Int,    'arr_type', lhs=lhs)
    length      = DottedVariable(Int,      'length', lhs=lhs)
    buffer_size = DottedVariable(Int, 'buffer_size', lhs=lhs)
    is_view     = DottedVariable(Int,     'is_view', lhs=lhs)

```

pyccel/ast/cwrapper.py
```diff
@@ -528,12 +548,56 @@ def pybool_to_bool(cast_function_name):
                        body      = cast_function_body,
                        results   = [cast_function_result])
 
+def pyarray_to_ndarray(cast_function_name):
+    cast_function_argument = Variable(dtype=PyccelPyArrayObject(), name='o',
+                                is_pointer=True)
+    cast_function_result   = Variable(dtype=NativeInteger(), name = 'c',
+                                rank=1)
```
yguclu, line 555: Is it really necessary to split these two lines? See comment below

pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -273,3 +274,33 @@ int32_t     get_index(t_ndarray arr, ...)
...
+    ndarray_strides = (int64_t*)malloc(sizeof(int64_t) * nd);
+    for (int i = 0; i < nd; i++)
+        ndarray_strides[i] = np_strides[i] / type_size;
+    return ndarray_strides;
+
+}
+
+/*
+** copy numpy shape to nd_array shape, and return it in a new array, to
+** avoid the problem of plateform depending architecture
```
yguclu, line 295: Typo: plateform -> platform
Typo: depending -> dependent

And what do you mean by "platform-dependent architecture" anyway?

yguclu (changes requested):

pyccel/stdlib/ndarrays/ndarrays.c
rakati: I mean different system architecture 32-bit, 64-bit ...

pyccel/ast/cwrapper.py
rakati: I am trying to respect the rule of pep.
the line should not contain more than 79 characters.

pyccel/stdlib/ndarrays/ndarrays.c
EmilyBourne: Numpy shape is not saved in a fixed length type

pyccel/ast/cwrapper.py
yguclu: The rule is there to help you write better code, not worse!
How long are these lines?
```python
Arr = PyccelPyArrayObject()
Int = NativeInteger()

cast_function_argument = Variable(Arr, 'o', is_pointer=True)
cast_function_result   = Variable(Int, 'c', rank=1)
```

pyccel/stdlib/ndarrays/ndarrays.c
yguclu: Thanks! Please improve the explanation in this comment...

yguclu (approved):

PR #634, opened 2020-12-23
Delete link to obsolete documentation

saidctb:
The files in the `doc` folder are obsolete, and so is any documentation generated from them. The user can find up-to-date information in the `tutorial` directory instead. List of changes:

- Remove obsolete link to http://pyccel.readthedocs.io from README.rst;
- Remove obsolete description of automatic docs from README.rst;
- Update project URL in setup.py.


README.rst
```diff
@@ -204,39 +204,13 @@ In order to run the unit tests and to get a coverage report, four additional Pyt
   pip3 install --user pytest
   pip3 install --user coverage
 
-
-Reading the docs
-================
```
EmilyBourne, line 209: Is it worth mentioning the tutorial folder? Once quickstart is added to it it will start to be a useful documentation source

EmilyBourne:
Does this mean we can delete the `doc/` folder that codacy says is full of nasty things?

README.rst
saidctb: ŵe can update it once we add the quickstart file

saidctb: the user can easaly see the tutorial folder in the repository, that's the first thing that you check

saidctb:
> Does this mean we can delete the `doc/` folder that codacy says is full of nasty things?

I think we should keep it, because we will use it one day, once we update it

EmilyBourne:
> > Does this mean we can delete the `doc/` folder that codacy says is full of nasty things?
> 
> I think we should keep it, because we will use it one day, once we update it

We might use the folder but will we use any of its contents?

saidctb:
> > > Does this mean we can delete the `doc/` folder that codacy says is full of nasty things?
> > 
> > 
> > I think we should keep it, because we will use it one day, once we update it
> 
> We might use the folder but will we use any of its contents?

probably this we need to discuss with Ahmed about the website and how we should organize it

EmilyBourne (approved):

saidctb:
@yguclu  can you remove the documentation link from the About section?

README.rst
```diff
@@ -204,39 +204,13 @@ In order to run the unit tests and to get a coverage report, four additional Pyt
...
-   make html
-
-Then, direct your browser to ``_build/html/index.html``.
-
 Testing
 =======
 
 To test your Pyccel installation please run the script *tests/run_tests_py3.sh* (Unix), or *tests/run_tests.bat* (Windows).
 
 Continuous testing runs on Travis CI: <https://travis-ci.com/github/pyccel/pyccel>
```
yguclu, line 212: @saidctb Since we are updating this file, can you please update this line as well? Our tests do not run on Travis anymore..

README.rst
```diff
@@ -248,11 +222,6 @@ TODO
     :scale: 100%
     :target: https://travis-ci.com/pyccel/pyccel
```
yguclu, line 223: This line also needs to be updated.

README.rst
saidctb: done

README.rst
saidctb: I didn't succeed in changing the badge to point to the GitHub actions link, I found this link https://docs.github.com/en/free-pro-team@latest/actions/managing-workflow-runs/adding-a-workflow-status-badge, but it doesn't seem to work for me

yguclu: Could you post every link that you have tried so far?

yguclu: By the way, have you already tried this?
https://github.com/pyccel/pyccel/workflows/.github/workflows/Github_pytest.yml/badge.svg

yguclu: Probably better:
https://github.com/pyccel/pyccel/workflows/.github/workflows/Github_pytest.yml/badge.svg?branch=master

README.rst
yguclu: Thanks!

README.rst
EmilyBourne: If we create a new workflow which only runs on pushes to the master branch and keep the current workflow for pull requests to the master branch then we can link to the badge for the new workflow

yguclu (changes requested):
As mentioned in a previous discussion, please create a new workflow which only runs on **pushes** to the `master` branch, and keep the current workflow for **pull requests** to the `master` branch.

Then we can link to the badge for the new workflow.

.github/workflows/master.yml
```diff
@@ -0,0 +1,120 @@
...
+        uses: codacy/codacy-coverage-reporter-action@master
+        continue-on-error: True
+        with:
+          project-token: ${{ secrets.CODACY_PROJECT_TOKEN }}
+          coverage-reports: cobertura.xml
+
+  Linux-Master:
+
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/master'
```
EmilyBourne, line 39: This line is no longer needed

.github/workflows/master.yml
```diff
@@ -0,0 +1,120 @@
...
+    strategy:
+      matrix:
+        python-version: [3.7, 3.8, 3.9]
```
EmilyBourne, line 42: You can add 3.6 here and remove the `Linux` job

EmilyBourne (changes requested):
Don't forget to stop the old workflow being activated on pushes and remove the things in that workflow which are specific to the master branch

.github/workflows/master.yml
saidctb: done

.github/workflows/master.yml
saidctb: done

README.rst
```diff
@@ -1,7 +1,6 @@
 Welcome to Pyccel
 =================
-
-|build-status| |docs| |codacy|
+|codacy|
```
EmilyBourne, line 3: We still need the build-status. That is why the new workflow was added

README.rst
EmilyBourne: The quickstart file should be merged any day now anyway though. What do you think @yguclu ?

EmilyBourne:
@yguclu changing the workflows seems to have confused github. It is marking `Linux-Master` as `Required`

README.rst
```diff
@@ -1,7 +1,7 @@
 Welcome to Pyccel
 =================
-
-|build-status| |docs| |codacy|
+![.github/workflows/Github_pytes.yml](https://github.com/pyccel/pyccel/workflows/.github/workflows/Github_pytest.yml/badge.svg)
```
EmilyBourne, line 3: typo: `Github_pytes.yml`->`Github_pytest.yml`

I think it would be better to use the same syntax as before which allows us to easily put all the badges on one line and specify the image attributes in one section at the end

saidctb: I'm just testing it to see if it works, apparently not

EmilyBourne: I don't think it can work until the workflow was been run once though :/

saidctb: the problem is that it doesn't create the badge

EmilyBourne: Yes that is what I would expect. The badge should point to the result of the workflow but if the workflow has never been run then there is no result to point to. Maybe you could try making it trigger on push to anything (so it runs once for this branch) then set it back to just push to master. That way a result exists and the badge has something to try and reference

EmilyBourne: I have done what I suggested and the badge now exists as expected

EmilyBourne: I tried what I suggested above. The badge now appears here:
https://github.com/pyccel/pyccel/workflows/master_tests/badge.svg

README.rst
EmilyBourne: I still suspect that it might be useful to mention the quickstart file but as it is still not merged I will approve this PR as is

EmilyBourne (approved):

README.rst
yguclu: OK. In PR #622 we should add a link to the quickstart file in the README.

yguclu (approved):

PR #638, opened 2020-12-24
Exempt AugAssign from raising reassigning errors

saniQQ:
When applied to a Numpy array, AugAssign does not reallocate the array; hence we should not raise an error.


EmilyBourne (approved):

EmilyBourne (changes requested):
Could you add a test please. One should be enough :+1:

saniQQ:
> Could you add a test please. One should be enough 👍

i added the tests in the Codegen tests is it enough?
i added them there because C still can't handle the augassign.

EmilyBourne:
> > Could you add a test please. One should be enough +1
> 
> i added the tests in the Codegen tests is it enough?
> i added them there because C still can't handle the augassign.

I always prefer an end-to-end test. Fortran can handle the augassign and this problem is in the semantic stage so it shouldn't matter which language it's tested with

MenoIy (approved):

EmilyBourne (approved):

yguclu (approved):

PR #645, opened 2021-01-01
Keep hidden imports 

### LINKED ISSUE #644: Keep hidden imports (needed for pycode) related to issue-#643
In many cases, some imports are removed since we rely on intrinsic Fortran functions.

Let's take the following example
```python

@types( 'int[:,:](order=C)' )
def double_loop_on_2d_array_C( z ):

    from numpy import shape

    m, n = shape( z )

    for i in range( m ):
        for j in range( n ):
            z[i,j] = i-j
```

The generated code is (using devel-643):
```python
@types("int[:,:](order=C)")
def double_loop_on_2d_array_C(z):    
    m = shape(z)[0]
    n = shape(z)[1]
    for i in range(0, m, 1):
        for j in range(0, n, 1):
            z[i,j] = i - j
```

**Expected behavior**
The import of **shape** must be kept somewhere so that we can print it using **pycode**.

**Language**
python to Python

**Additional context**
Some improvements have been done on pycode in #643

MenoIy:
Keep information about all imported modules, as required by Python printer. Closes #644 
- always collect import at the semantic stage.
- add property ignore in import class.
- ignore collected import in the C and Fortran printing stage.
- search import by name and Asname

### DIFF
--- a/pyccel/codegen/printing/fcode.py
+++ b/pyccel/codegen/printing/fcode.py
@@ -369,7 +369,7 @@ class FCodePrinter(CodePrinter):
     def _print_Import(self, expr):
 
         source = ''
-        if str(expr.source) in pyccel_builtin_import_registery or expr.ignore:
+        if expr.ignore:
             return ''
 
         if isinstance(expr.source, DottedName):
--- a/pyccel/codegen/printing/fcode.py
+++ b/pyccel/codegen/printing/fcode.py
@@ -370,7 +370,7 @@ class FCodePrinter(CodePrinter):
     def _print_Import(self, expr):
 
         source = ''
-        if str(expr.source) in pyccel_builtin_import_registery:
+        if str(expr.source) in pyccel_builtin_import_registery or expr.ignore:
             return ''
 
         if isinstance(expr.source, DottedName):
--- a/pyccel/codegen/printing/pycode.py
+++ b/pyccel/codegen/printing/pycode.py
@@ -122,6 +122,12 @@ class PythonCodePrinter(SympyPythonCodePrinter):
             code += 'return {}'.format(ret)
         return code
 
+    def _print_AsName(self, expr):
+        name = self._print(expr.name)
+        target = self._print(expr.target)
+
+        return '{name} as {target}'.format(name = name, target = target)
+
     def _print_PythonTuple(self, expr):
         args = ', '.join(self._print(i) for i in expr.args)
         return '('+args+')'
--- a/pyccel/codegen/printing/pycode.py
+++ b/pyccel/codegen/printing/pycode.py
@@ -134,6 +134,9 @@ class PythonCodePrinter(SympyPythonCodePrinter):
 
     def _print_Import(self, expr):
         target = ', '.join([self._print(i) for i in expr.target])
+
+        if target == '' and expr.source is not None:
+            return 'import {source}'.format(source=self._print(expr.source))
         if expr.source is None:
             return 'import {target}'.format(target=target)
         else:
--- a/pyccel/codegen/printing/pycode.py
+++ b/pyccel/codegen/printing/pycode.py
@@ -172,7 +172,7 @@ class PythonCodePrinter(SympyPythonCodePrinter):
 
     def _print_Import(self, expr):
         source = self._print(expr.source)
-        if expr.target is None:
+        if not expr.target:
             return 'import {source}'.format(source=source)
         else:
             target = ', '.join([self._print(i) for i in expr.target])
--- a/pyccel/codegen/printing/pycode.py
+++ b/pyccel/codegen/printing/pycode.py
@@ -146,9 +146,9 @@ class PythonCodePrinter(SympyPythonCodePrinter):
         return '.'.join(self._print(n) for n in expr.name)
 
     def _print_FunctionCall(self, expr):
-        func = self._print(expr.func)
-        args = ','.join(self._print(i) for i in expr.arguments)
-        return'{func}({args})'.format(func=func, args=args)
+        func = expr.funcdef
+        args = ','.join(self._print(i) for i in expr.args)
+        return'{func}({args})'.format(func=func.name, args=args)
 
     def _print_Len(self, expr):
         return 'len({})'.format(self._print(expr.arg))
@@ -276,9 +276,13 @@ class PythonCodePrinter(SympyPythonCodePrinter):
         return 'print({0})'.format(fs)
 
     def _print_Module(self, expr):
-        code = '\n'.join(self._print(e) for e in expr.body)
-        imports = '\n'.join(self._print(i) for i in self._additional_imports)
-        return '\n'.join([imports, code])
+        body = '\n'.join(self._print(e) for e in expr.body)
+        imports  = [*expr.imports, *self._additional_imports]
+        imports  = '\n'.join(self._print(i) for i in imports)
+        return ('{imports}\n\n'
+                '{body}').format(
+                        imports = imports,
+                        body    = body)
 
     def _print_PyccelPow(self, expr):
         base = self._print(expr.args[0])
--- a/pyccel/codegen/printing/pycode.py
+++ b/pyccel/codegen/printing/pycode.py
@@ -122,6 +122,18 @@ class PythonCodePrinter(SympyPythonCodePrinter):
             code += 'return {}'.format(ret)
         return code
 
+    def _print_Program(self, expr):
+        body  = self._print(expr.body)
+        body = self._indent_codestring(body)
+        imports  = [*expr.imports, *self._additional_imports]
+        imports  = '\n'.join(self._print(i) for i in imports)
+
+        return ('{imports}\n'
+                'if __name__ == "__main__":\n'
+                '{body}\n').format(imports=imports,
+                                    body=body)
+
+
     def _print_AsName(self, expr):
         name = self._print(expr.name)
         target = self._print(expr.target)
--- a/pyccel/codegen/printing/pycode.py
+++ b/pyccel/codegen/printing/pycode.py
@@ -134,9 +134,6 @@ class PythonCodePrinter(SympyPythonCodePrinter):
 
     def _print_Import(self, expr):
         target = ', '.join([self._print(i) for i in expr.target])
-
-        if target == '' and expr.source is not None:
-            return 'import {source}'.format(source=self._print(expr.source))
         if expr.source is None:
             return 'import {target}'.format(target=target)
         else:
--- a/pyccel/codegen/printing/pycode.py
+++ b/pyccel/codegen/printing/pycode.py
@@ -114,7 +114,13 @@ class PythonCodePrinter(SympyPythonCodePrinter):
         return code
 
     def _print_Return(self, expr):
-        return 'return {}'.format(self._print(expr.expr))
+        code = ''
+        if expr.stmt:
+            code += self._print(expr.stmt)+'\n'
+        if expr.args:
+            ret = ','.join([self._print(i) for i in expr.expr])
+            code += 'return {}'.format(ret)
+        return code
 
     def _print_PythonTuple(self, expr):
         args = ', '.join(self._print(i) for i in expr.args)
--- a/pyccel/codegen/printing/pycode.py
+++ b/pyccel/codegen/printing/pycode.py
@@ -164,7 +164,7 @@ class PythonCodePrinter(SympyPythonCodePrinter):
 
     def _print_FunctionCall(self, expr):
         func = expr.funcdef
-        args = ','.join(self._print(i) for i in expr.args)
+        args = ', '.join(self._print(i) for i in expr.args)
         return'{func}({args})'.format(func=func.name, args=args)
 
     def _print_Len(self, expr):
--- a/pyccel/codegen/printing/pycode.py
+++ b/pyccel/codegen/printing/pycode.py
@@ -144,6 +144,11 @@ class PythonCodePrinter(SympyPythonCodePrinter):
         args = ', '.join(self._print(i) for i in expr.args)
         return '('+args+')'
 
+    def _print_PyccelArraySize(self, expr):
+        arg = self._print(expr.arg)
+        index = self._print(expr.index)
+        return 'shape({0})[{1}]'.format(arg, index)
+
     def _print_Comment(self, expr):
         txt = self._print(expr.text)
         return '# {0} '.format(txt)
--- a/pyccel/codegen/printing/pycode.py
+++ b/pyccel/codegen/printing/pycode.py
@@ -117,7 +117,7 @@ class PythonCodePrinter(SympyPythonCodePrinter):
         code = ''
         if expr.stmt:
             code += self._print(expr.stmt)+'\n'
-        if expr.args:
+        if expr.expr:
             ret = ','.join([self._print(i) for i in expr.expr])
             code += 'return {}'.format(ret)
         return code
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -2848,6 +2848,7 @@ class SemanticParser(BasicParser):
         return IsClass(var1, var2)
 
     def _visit_Import(self, expr, **settings):
+
         # TODO - must have a dict where to store things that have been
         #        imported
         #      - should not use namespace
@@ -2863,6 +2864,7 @@ class SemanticParser(BasicParser):
 
         if source in pyccel_builtin_import_registery:
             imports = pyccel_builtin_import(expr)
+
             def _insert_obj(location, target, obj):
                 F = self.check_for_variable(target)
 
@@ -2875,7 +2877,7 @@ class SemanticParser(BasicParser):
                     errors.report(IMPORTING_EXISTING_IDENTIFIED,
                                   bounding_box=(self._current_fst_node.lineno, self._current_fst_node.col_offset),
                                   severity='fatal')
-            expr.ignore = True
+
             if expr.target:
                 for (name, atom) in imports:
                     if not name is None:
@@ -2883,7 +2885,6 @@ class SemanticParser(BasicParser):
                             _insert_obj('variables', name, atom)
                         else:
                             _insert_obj('functions', name, atom)
-
             else:
                 _insert_obj('variables', source_target, imports)
             key = expr.source if isinstance(expr.source, AsName) else source
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -2833,7 +2833,6 @@ class SemanticParser(BasicParser):
         #      - should not use namespace
 
         container = self.namespace.imports
-
         if isinstance(expr.source, AsName):
             source        = str(expr.source.name)
             source_target = str(expr.source.target)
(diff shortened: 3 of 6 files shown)


pyccel/parser/semantic.py
```diff
@@ -446,13 +446,15 @@ def get_import(self, name):
...
         container = self.namespace
         while container:
 
-            if name in container.imports['imports']:
-                imp =  container.imports['imports'][name]
-                break
+            #case import x as y
+            for key,_ in container.imports['imports'].items():
+                if str(key) == str(name) or (isinstance(key, AsName)
+                                            and str(key.target) == str(name)):
```
EmilyBourne, line 455: Do you definitely need the `str` calls?
If not, I think this line can be simplified thanks to the magic method `AsName.__eq__`
See : https://github.com/pyccel/pyccel/blob/master/pyccel/ast/core.py#L594

pyccel/parser/semantic.py
```diff
@@ -446,13 +446,15 @@ def get_import(self, name):
...
 
         imp = None
-
         container = self.namespace
         while container:
 
-            if name in container.imports['imports']:
-                imp =  container.imports['imports'][name]
-                break
+            #case import x as y
```
EmilyBourne, line 452: You know this is going to fall pray to "You touched it you document it right" ^^

pyccel/parser/semantic.py
```diff
@@ -1052,7 +1054,9 @@ def _visit_DottedName(self, expr, **settings):
                     args  = self._handle_function_args(rhs.args, **settings)
                     func  = first[rhs_name]
                     if new_name != rhs_name:
-                        func  = func.clone(new_name)
+                        if hasattr(func, 'clone'):
+                            func  = func.clone(new_name)
```
EmilyBourne, line 1058: Did you have a code that failed here?

pyccel/parser/semantic.py
```diff
@@ -2566,7 +2570,8 @@ def _visit_FunctionDef(self, expr, **settings):
             global_vars = [v for v in self.get_variables(self.namespace.parent_scope) if v not in args + results + local_vars]
 
             # get the imports
-            imports   = self.namespace.imports['imports'].values()
+            imports   = list(self.namespace.imports['imports'].values())
+            imports += list(self.namespace.parent_scope.imports['imports'].values())
```
EmilyBourne, line 2574: If you need the parent, do you not also need the parent's parent, etc?

pyccel/parser/semantic.py
MenoIy:
```
def func():
    import numpy as np
    f = np.sqrt(4)
    import math as np
    e = np.sqrt(4)
```

pyccel/parser/semantic.py
MenoIy: I cast to str because sometimes it can be a comparison between string and symbol. a magic method should be a good thing to add.

EmilyBourne: Hmm, maybe we need to open an issue in epic #665 to get rid of `Symbol` too

pyccel/parser/semantic.py
EmilyBourne: Ok. I think this is the right behaviour. Could you add a test. If it works then you can remove the `#TODO`

mostafaMamoni (approved):

pyccel/parser/semantic.py
```diff
@@ -443,16 +443,19 @@ def get_function(self, name):
 
 
     def get_import(self, name):
-        """."""
+        """
+        Search for a Import object with the given name in the current namespace.
```
saniQQ, line 447: typo (a Import) >> an Import

pyccel/parser/semantic.py
```diff
@@ -443,16 +443,19 @@ def get_function(self, name):
...
         container = self.namespace
         while container:
 
-            if name in container.imports['imports']:
-                imp =  container.imports['imports'][name]
-                break
+            for key,_ in container.imports['imports'].items():
+                if key == name:
+                    imp = container.imports['imports'][key]
+
```
saniQQ, line 458: is it necessary to keep loping in the parent containers after finding an import.

saniQQ (changes requested):

pyccel/parser/semantic.py
MenoIy: that's true, I changed my code a lot and I forget that I removed a condition there, I will correct it.

saniQQ (approved):

pyccel/codegen/printing/pycode.py
```diff
@@ -134,16 +163,16 @@ def _print_DottedName(self, expr):
         return '.'.join(self._print(n) for n in expr.name)
 
     def _print_FunctionCall(self, expr):
-        func = self._print(expr.func)
-        args = ','.join(self._print(i) for i in expr.arguments)
-        return'{func}({args})'.format(func=func, args=args)
+        func = expr.funcdef
+        args = ','.join(self._print(i) for i in expr.args)
```
EmilyBourne, line 167: Please add a space between arguments

pyccel/parser/semantic.py
```diff
@@ -443,15 +443,20 @@ def get_function(self, name):
 
 
     def get_import(self, name):
-        """."""
+        """
+        Search for an import with the given name in the current namespace.
+        Return None if not found.
+        """
```
EmilyBourne, line 449: Thanks :+1:

pyccel/parser/semantic.py
```diff
@@ -443,15 +443,20 @@ def get_function(self, name):
...
         container = self.namespace
         while container:
 
-            if name in container.imports['imports']:
-                imp =  container.imports['imports'][name]
+            for key,_ in container.imports['imports'].items():
+                if key == name:
+                    imp = container.imports['imports'][key]
+
+            if imp:
```
EmilyBourne, line 459: Why do you need 2 ifs? Can the expression above return None?

pyccel/codegen/printing/pycode.py
```diff
@@ -114,12 +114,41 @@ def _print_FunctionDef(self, expr):
...
+        return '{name} as {target}'.format(name = name, target = target)
 
     def _print_PythonTuple(self, expr):
         args = ', '.join(self._print(i) for i in expr.args)
         return '('+args+')'
 
+    def _print_PyccelArraySize(self, expr):
+        arg = self._print(expr.arg)
+        index = self._print(expr.index)
+        return 'shape({0})[{1}]'.format(arg, index)
```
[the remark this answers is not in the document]
MenoIy: normally here we don't need to import shape because here shape is a property of x, it can be printed as dotted elements I don't think we need to import it.

pyccel/codegen/printing/pycode.py
```diff
@@ -114,12 +114,41 @@ def _print_FunctionDef(self, expr):
...
+            code += self._print(expr.stmt)+'\n'
+        if expr.args:
+            ret = ','.join([self._print(i) for i in expr.expr])
+            code += 'return {}'.format(ret)
+        return code
+
+    def _print_Program(self, expr):
+        body  = self._print(expr.body)
+        body = self._indent_codestring(body)
+        imports  = [*expr.imports, *self._additional_imports]
```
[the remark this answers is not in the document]
MenoIy: for the moment there is no additional import needed but I left this one as a safety measure for the rest of the python printer

pyccel/codegen/printing/pycode.py
```diff
@@ -114,12 +114,41 @@ def _print_FunctionDef(self, expr):
...
+        if expr.stmt:
+            code += self._print(expr.stmt)+'\n'
+        if expr.args:
+            ret = ','.join([self._print(i) for i in expr.expr])
+            code += 'return {}'.format(ret)
+        return code
+
+    def _print_Program(self, expr):
+        body  = self._print(expr.body)
+        body = self._indent_codestring(body)
```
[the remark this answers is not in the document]
MenoIy: can you explain what do you mean by this?

pyccel/parser/semantic.py
MenoIy: yes, when the import is not in the given namespaces.
here the dict is orderedDict and I need the last import 
example:
```PYTHON
import numpy as np
import math as np
x = np.sqrt(x) #here sqrt is from math math 
```
so I found a un imp in the current scop no need to search in parents cop

pyccel/codegen/printing/pycode.py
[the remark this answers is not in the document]
EmilyBourne: It could be printed as a Dotted element, but once it's been recognised I don't think it's saved as a Dotted element any more. Maybe you could add a test for this?

pyccel/parser/semantic.py
EmilyBourne: I meant can `container.imports['imports'][key]` return None? If not then can't the break go next to the assignment

pyccel/codegen/printing/pycode.py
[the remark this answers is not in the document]
EmilyBourne: The c code and the fortran code is also indented. Eg the fortran code is indented here:
https://github.com/pyccel/pyccel/blob/master/pyccel/codegen/printing/fcode.py#L260-261

However this function isn't called explicitly in fcode.py. I think it must be called by the sympy printer.  I suspect that the python code indentation could/should/is handled the same way.

If that is the case then `_indent_codestring` doesn't need to be called here

pyccel/codegen/printing/fcode.py
```diff
@@ -369,7 +369,7 @@ def _print_Program(self, expr):
     def _print_Import(self, expr):
 
         source = ''
-        if str(expr.source) in pyccel_builtin_import_registery:
+        if str(expr.source) in pyccel_builtin_import_registery or expr.ignore:
```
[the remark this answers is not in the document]
MenoIy: print_import print also additional import and in C we handle some of the math function with additional import, and they are printed in C code even if they exist in pyccel_builtin_import_register.

I may propose this solution:
1 - drop the property ignore and use a condition in _builtin_import_registery and not in additional import_

pyccel/parser/semantic.py
MenoIy: first container.imports['imports'][key] cant return None 
Second imp is initialized with None, so if the condition name == key doesn't occur. imp is still None and we need to search in the parent scope 
otherwise, we need to break.

pyccel/codegen/printing/fcode.py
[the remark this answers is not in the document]
EmilyBourne: That sounds like a good solution. Presumably, currently in Fortran, we cannot import a module which shares a name with something in `pyccel_builtin_import_registery`?
That said, we do not have this condition in C and we don't have any problems. Presumably because the import was ignored at the semantic stage. [Codacy confirms that this condition is not used in the Fortran code](https://app.codacy.com/gh/pyccel/pyccel/file/51897655523/coverage?bid=15415567&fileBranchId=15415567#l373).

I think the simpler solution would therefore be to change this condition to `if expr.ignore`.
What do you think?

pyccel/parser/semantic.py
EmilyBourne: I think I'm not being clear. My suggested code is:
```
           for key,_ in container.imports['imports'].items():
                if key == name:
                    imp = container.imports['imports'][key]
                    break
```
If `container.imports['imports'][key]` can't return None then this should be possible and it saves an if statement

tests/epyccel/test_imports.py
```diff
@@ -0,0 +1,60 @@
...
+    f = epyccel(f3, language = language)
+    x = ones(10, dtype=int)
+    assert f(x) == f3(x)
+
+def test_import_collision(language):
+    @types('int')
+    def f4(x):
+        import modules.Module_3 as mod
+        add_one = mod.add_one(x)
+        return add_one
```
EmilyBourne, line 57: This test has "collision" in the name but it does not appear to test collisions. Or have I missed something?

MenoIy: as said before the dotted element rhs translated to just a function call  add_one = add_one(x)
but in the semantic stage, we clone the function with a different name to avoid name collision.

EmilyBourne: Ah yes sorry the function and the variable have the same name :+1:

pyccel/parser/semantic.py
EmilyBourne: I assume the collision test is for this? But to fully test it I think we also need a collision with a builtin function such as sqrt, not just a user defined function

pyccel/parser/semantic.py
MenoIy: this break will only  stop the for loop, not the while loop  
I need to leave the while loop when I find an import because there is no need to search in the parent scope

pyccel/parser/semantic.py
MenoIy: 1- Numpy and Math function in pyccel , don't have either clone methods or name property
the same problem is in C and Fortran

pyccel/parser/semantic.py
```diff
@@ -443,15 +443,20 @@ def get_function(self, name):
...
+        """
 
         imp = None
-
         container = self.namespace
         while container:
 
-            if name in container.imports['imports']:
-                imp =  container.imports['imports'][name]
+            for key,_ in container.imports['imports'].items():
```
EmilyBourne, line 455: If you don't need the values of a dictionary you can use the `.keys()` method instead of the `.items()` method

pyccel/parser/semantic.py
```diff
@@ -443,15 +443,20 @@ def get_function(self, name):
...
+                if key == name:
```
EmilyBourne, line 456: This can be simplified to
```
if name in container.imports['imports']:
```

pyccel/parser/semantic.py
EmilyBourne: Yes, but I don't think they need them, so the test that you wrote above should still work shouldn't it?

MenoIy: when we remove the condition  it won't work  in C and Fortran 
the reason it was working before is that hidden imports were not taken into consideration  so there was no name clash

EmilyBourne: I didn't suggest removing the condition, just adding a test for the case that doesn't use the condition. Eg the case you gave above

pyccel/parser/semantic.py
MenoIy: 1- having import like 
```PYTHON
import numpy as np
import math as np
```
and a call
```
x = np.sqrt()
```
both of those import satisfy the condition 
``` 
if name in container.imports['imports']:
#or
key == name    #AsName(numpy, np ) == np and AsName(math, np) == np
```
so we need to take the last import created that's the reason I'm looping
is there another way to collect the last element in dict that satisfy a given condition?

pyccel/parser/semantic.py
MenoIy: only for Fortran and C?

pyccel/parser/semantic.py
EmilyBourne: A dictionary can't have multiple elements with the same key. If the key is the same then the first import will no longer exist in the dictionary after the import of the second element so you will automatically get the last object added

pyccel/parser/semantic.py
EmilyBourne: Of course! Sorry. This should be possible if the discussion below removes the for loop though

pyccel/parser/semantic.py
MenoIy: I said nothing about the same key.  here key can be an AsName "numpy as np" and "math as np" both have np as target so we can track dotted elements np.xxxx

pyccel/parser/semantic.py
EmilyBourne: Why? Do you expect it to not work for python?
As it is in the semantic stage one language should be sufficient. Python makes the most sense for this PR I think

pyccel/parser/semantic.py
EmilyBourne: By "same key" I mean that the dictionary considers it to be the same. The details of how this is done are discussed [here](https://stackoverflow.com/questions/32279666/objects-are-not-considered-the-same-in-dictionary-keys-but-eq-is-implement). This means that if the class has no `__eq__` method then it will work as you describe (`numpy as np` and `math as np` will be considered to be different keys), but if it has an `__eq__` method (as for `AsName`) then it depends on the result of the `__hash__` function. For [`AsName`](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/core.py#L600) this means that the key is essentially the target so `numpy as np` and `math as np` will be considered to be the same key. You can verify this by printing the dictionary

pyccel/parser/semantic.py
MenoIy: one problem here 
the second import will be Import('math', ['sqrt_001'])
in C and Fortran import are ignored
but in Python we need to print sqrt with the new name sqrt_001
ex  : ```e = sqrt_001(4)```

EmilyBourne: Ah ok I see the problem. That is beyond the scope :+1:

pyccel/parser/semantic.py
```diff
@@ -2883,6 +2903,9 @@ def _insert_obj(location, target, obj):
                             _insert_obj('functions', name, atom)
             else:
                 _insert_obj('variables', source_target, imports)
+            key = expr.source if isinstance(expr.source, AsName) else source
+            _insert_obj('imports', key, Import(source, expr.target, True))
```
EmilyBourne, line 2907: Are you sure you can call `_insert_obj` twice in a row for the same object given that this function raises errors if an object exists already? Could you not use your `insert_import` function here?

MenoIy: that's what I'm trying to do I need to tweak  the function a little

pyccel/codegen/printing/pycode.py
```diff
@@ -114,12 +114,41 @@ def _print_FunctionDef(self, expr):
         return code
 
     def _print_Return(self, expr):
-        return 'return {}'.format(self._print(expr.expr))
+        code = ''
+        if expr.stmt:
+            code += self._print(expr.stmt)+'\n'
+        if expr.args:
```
EmilyBourne, line 120: Shouldn't this check `expr.expr` not `expr.args`?

pyccel/codegen/printing/pycode.py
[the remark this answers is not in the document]
EmilyBourne: Did you look into this at all?

pyccel/codegen/printing/pycode.py
EmilyBourne: Or maybe the if isn't needed? I don't think `bool(expr.args)` can be anything other than `true`

MenoIy: maybe i should check the length?

EmilyBourne: `bool([])` gives `False` so just `if expr.expr` should be enough

pyccel/codegen/printing/pycode.py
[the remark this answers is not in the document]
MenoIy: yes its inherit from sympy.printing.pycode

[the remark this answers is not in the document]
EmilyBourne: I've just realised why this obviously needs to be here. Python has no end clause so you have to indent as you go otherwise there's no way to tell where the end is :+1: Sorry 🤦

EmilyBourne (approved):

yguclu (approved):

Issue #453, opened 2020-10-05
Support Python classes


yguclu:

EmilyBourne:
Cwrapper reference: https://docs.python.org/3/extending/newtypes_tutorial.html

EmilyBourne:
Additional cwrapper reference for class variable getters and setters (this mimicks the property decorator) : https://docs.python.org/3/c-api/structures.html#c.PyGetSetDef

sboof911:
I was trying to find the best way to implement the classes in c and I found this method is the more readable and close to the concept of classes.
For the implementation of the struct  that can where we can implement the `Phase 1`:
``` c
#ifndef TESTMOD_H
#define TESTMOD_H

#include <stdio.h>
#include <stdlib.h>
#include <time.h>

typedef struct A A;

struct A
{
    void    (*Constrater)(A*, double);
    void    (*Destroctor)(A*);
    void    (*function)(A*, double);
    double   a;
};

void     Class_A(A* self, double a);
void    function_A(A* self, double x); //just for the test

#endif
```
And for the implementation of the function we should declare them by this syntax.
``` c
void    Destroctor_A(A* self)
{
    free(self);
    printf("Destroctor Called\n");
}

void    Constrator_A(A* self, double a)
{
    self->a = a;
    printf("Controctor Called\n");
}

void   function_A(A* self, double x)
{
    for (int i = 0; i < 1000000000; i++) {
        self->a += (double)i / (i + 1);
    }
}

void     Class_A(A* self, double a)
{
    self->Constrater = Constrator_A;
    self->function = function_A;
    self->Destroctor = Destroctor_A;
    self->Constrater(self, a); // Call the contractor
}
```
The declaration of the functions are in a way we will avoid name collision eg: `FunctionName_ClassName`.
Here is a `main` that test that compare the compilation time between a `function pointer call` and a `function call`.
``` c
void show_time(char *str, struct timespec *tv1, struct timespec *tv2)
{
  printf("%s: %ld seconds\n", str, tv2->tv_sec-tv1->tv_sec);
//   printf("%s: %ld nano seconds\n", str, tv2->tv_nsec-tv1->tv_nsec);
}

int main(void)
{
    A* p = malloc(sizeof(A));
    Class_A(p, 0.0);
    {
        struct timespec  tv1, tv2;
        timespec_get(&tv1, TIME_UTC);
        for(long long int i = 0; i < 100; i++)
            p->function(p, 1);
[...]

PR #637, opened 2020-12-24
Create ast.variable and bug fixes

EmilyBourne:
Move classes necessary for the handling of Variables into a new file `ast.variable`
Fixes #490 (ast/core.py too large for codacy).

**Commit Summary**
- Move classes necessary for the handling of Variables into a new file `ast.variable`
- Move Slice to `ast.internals` (should this be in operators?)
- Remove `is_polymorphic` property from Variables. This property exists in the class dtype and a "polymorphic variable" doesn't make much sense
- Add `passed_from_dotted` property to `Declare`. In fortran types are declared differently when they are used for arguments which are passed due to class syntax
- General Codacy fixes

PR #646, opened 2021-01-01
Stop using DottedVariable to store FunctionCall objects (fixes basic classes)

EmilyBourne:
Currently a `DottedVariable` is used to store any dotted object. This makes no sense for FunctionCalls. A class `DottedFunctionCall` is created to hold dotted function call objects. Minor bug fixes are also carried out which fix basic classes.

Related to #453 

**Commit Summary**
- Create `DottedFunctionCall` object which inherits from `FunctionCall`
- Stop class constructors returning pointers
- Add missing `str` call required to recognise constructor method
- Give `DottedVariable` missing `allows_negative_indexes` property
- Allow `PyccelArraySize` to provide the size of a `DottedVariable`
- Do not make assumptions about Indexed base type
- Move basic class tests from codegen folder to pyccel folder to test end to end


EmilyBourne:
This change was necessary in order to clean up the variables (see PR #637). That PR is getting too big so I have split this into its own PR

MenoIy (approved):

pyccel/ast/core.py
```diff
@@ -2919,6 +2927,51 @@ def funcdef(self):
...
 
+    @property
+    def func_name(self):
+        return self._func_name
+
+    @property
+    def interface_name(self):
+        return self._interface_name
+
+class DottedFunctionCall(FunctionCall):
```
saidctb, line 2938: if `DottedFunctionCall` inherits from `FunctionCall` then `DottedVariable` should do the same thing and inherit from `Variable`

EmilyBourne: Yes, this is being done in PR #637. But that PR was getting excessively large so I extracted this part which didn't seem relevant. Would you prefer that I moved that change to this PR?

saidctb: it's up to you , if it's in another PR I guess it's okay, but probably grouping them in one PR would be good also.

EmilyBourne: Ok. Then I will use this PR for all the Dotted improvements

pyccel/ast/core.py
```diff
@@ -2468,126 +2483,19 @@ def _eval_is_positive(self):
...
-    def __init__(self, lhs, rhs):
-        if self.stage == 'syntactic':
-            return
-        self._dtype     = rhs.dtype
-        self._rank      = rhs.rank
-        self._precision = rhs.precision
-        self._shape     = rhs.shape
-        self._order     = rhs.order
+    def __init__(self, *args, lhs, **kwargs):
+        Variable.__init__(self, *args, **kwargs)
```
saidctb, line 2493: I'm not sure about this, we should be able to do something like `(a-b).sum()`

saidctb: I guess it's the same problem for `DottedFunctionCall`, when we have something like `(2*b).sum()`

saidctb: the first element can be an expression , the rest should be variables or variables + function call

EmilyBourne: I don't see the problem. I don't impose any restrictions on the lhs. So for `(a-b).shape` the dotted variable would contain the shape part with
shape=1
rank=0
dtype=int
And the lhs (which is visited separately) would contain `PyccelMinus(a,b)`

saidctb: how about the name ?

EmilyBourne: We can't test the examples you mention for 2 reasons
1. Pyccel does not handle magic methods
2. `shape` and `sum` are special cases which would not be translated to Dotted objects anyway

But perhaps the following test (I have not tried it yet) would be enough to convince you:
```
class A:
    def __init__(self):
        self.x =3

    def f(self):
        return False

def get_A():
    return A()

a = get_A().x
b = get_A().f()
```
?

EmilyBourne: > how about the name ?

The name would be `shape` (although of course shape would actually be converted to `PyccelAstNode`). The name is the name of the class variable (this seems to be necessary to easily collect the object from the `ClassDef`). The printer then just does:
```
self._print(expr.lhs) + ' % ' + self._print(expr.name)
```
However you are right that brackets would be needed if the lhs is not a singleton

saidctb: yes, I agree we don't have those cases yet, but we will handle magic methods at a certain point or the example that you gave, we just need to think about them for now and decide what to do for later , so that we won't change the dotted variable once more

saidctb: you probably need to do something like
```python
def __init__(self, *args):
    self.clone(args[-1])
```

saidctb: but then there is this case `a.b().c.d().e` , does it make sense to seperate the dotted function call from the dotted variable , I mean it should still work fine with the seperation, but do we need it ?

EmilyBourne: The example above is now working. I had to make a few changes to the parser so it would parse correctly, but no changes to `DottedVariable` or `DottedFunctionCall`.
I have added it to the tests

saidctb: ah perfect

saidctb (approved):

pyccel/ast/core.py
EmilyBourne: > but then there is this case `a.b().c.d().e` , does it make sense to seperate the dotted function call from the dotted variable , I mean it should still work fine with the seperation, but do we need it ?

I think it helps as much as having a separation between Variable and FunctionCall. The problem with keeping everything in the same object is that every time you might encounter a DottedVariable (which is everytime you might encounter a Variable) you would have to check the type and collect the Variable.
Consider
```
a.f()+a.x
```
with the inheritance here, this is handled in exactly the same way as
```
f()+x
```
which means that we don't have to maintain classes separately. But if they are not separated then you need to check all the arguments to `PyccelAdd` to check for Dotted types and then you need to retrieve the object with the relevant information. This would need doing for every class in pyccel which needs to calculate its dtype (and there are a lot)

saidctb: ah okay I see

saidctb (approved):

yguclu (approved):

PR #647, opened 2021-01-03
Remove IndexedVariable type

EmilyBourne:
An `IndexedElement` has 2 properties, `base` and `indexes`
For `a[i]`, `a`  is the base and `i`  is the index. Previously the base was an `IndexedVariable` (most of the time). An `IndexedVariable` has most of the properties of a `Variable` but none of the newer ones (e.g allow_negative_indexes) as it was not well maintained. This means that the `Variable` has to be accessed via `a.internal_variable`. This leads to the code becoming less readable without providing any benefits. This PR removes the unnecessary `IndexedVariable` type which means that the base of an `IndexedElement` will always be up to date with the `Variable` type.

For the behaviour to remain the same in the rest of the code the magic `__getitem__` method is implemented for `Variable` and updated for `TupleVariable`. Previously the `__getitem__` method of a `TupleVariable` was only used when providing 1 index to a non-homogeneous tuple. It therefore returned the internal `Variable` at that index. It must now handle the indexing for homogeneous tuples (which is the same as for a normal `Variable`) and for inhomogeneous tuples (the first index is treated as before, the additional indices are passed to the `__getitem__` method of the returned `Variable`)

**Commit Summary**
- Remove `IndexedVariable` class
- Make `IndexedElement` base into `Variable`
- Add `__getitem__` method to `Variable`
- Update `__getitem__` method of `TupleVariable`
- Remove references to `IndexedVariable`
- Remove unnecessary code at end of `_extract_indexed_from_var` function
-  Add empty slices in `IndexedElement` constructor instead of `SemanticParser`
- Simplify `IndexedElement` constructor
- Remove obsolete methods `_print_IndexedBase` and `_print_Indexed` from Fortran printer

### DIFF
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -2468,6 +2468,17 @@ class Variable(Symbol, PyccelAstNode):
         #we do this inorder to infere the type of Pow expression correctly
         return self.is_real
 
+    def __getitem__(self, *args):
+
+        if len(args) == 1 and isinstance(args[0], (Tuple, tuple, list)):
+            args = args[0]
+
+        if self.rank != len(args):
+            raise IndexError('Rank mismatch.')
+
+        obj = IndexedElement(self, *args)
+        return obj
+
 class DottedVariable(AtomicExpr, sp_Boolean, PyccelAstNode):
 
     """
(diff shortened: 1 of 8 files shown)


EmilyBourne:
The contents of this PR were originally part of PR #637 but have been separated out to make reviewing more manageable

pyccel/ast/core.py
```diff
@@ -2468,6 +2465,17 @@ def _eval_is_positive(self):
         #we do this inorder to infere the type of Pow expression correctly
         return self.is_real
 
+    def __getitem__(self, *args):
+
+        if len(args) == 1 and isinstance(args[0], (Tuple, tuple, list)):
+            args = args[0]
+
+        if self.rank != len(args):
```
saidctb, line 2473: you can have something like `x[1]` where the rank of `x` is 2 but you can't have something like `x[1,2,3]`

EmilyBourne: So you would suggest: `if self.rank > len(args)`?

In pyccel we always add `Slice(None,None)` when indexing so this condition is always satisfied. I think probably helps so a Variable is either fully indexed or not indexed at all. Do you think this handling is likely to change? @saniQQ does the C implementation of slicing require us to provide slices for all dimensions?

saidctb: yes we should only have `if self.rank > len(args)`, I don't know how it's treated in the C language, but the missing indices should automaticaly added here

EmilyBourne: Currently the extra indexes are added in the `SemanticParser`. Do you want me to move this to the constructor of `IndexedElement`? I would rather avoid doing it in the `__getitem__` method as this will lead to code duplication

saidctb: ah okay, we probably need to do it the constructor

EmilyBourne: done

saidctb (approved):

pyccel/codegen/printing/fcode.py
```diff
@@ -1152,10 +1152,7 @@ def _print_AliasAssign(self, expr):
             rhs = rhs.variable
 
         if isinstance(lhs, TupleVariable) and not lhs.is_homogeneous:
-            if isinstance(rhs, (TupleVariable, PythonTuple)):
-                return self._print(CodeBlock([AliasAssign(l, rhs[i]) for i,l in enumerate(lhs)]))
-            else:
-                return self._print(CodeBlock([AliasAssign(l, IndexedVariable(rhs)[i]) for i,l in enumerate(lhs)]))
+            return self._print(CodeBlock([AliasAssign(l, rhs[i]) for i,l in enumerate(lhs)]))
```
yguclu, line 1155: Is it necessary to index `rhs` explicitly, or can we write this instead?
```
[AliasAssign(l, r) for l, r in zip(lhs, rhs)]
```

EmilyBourne: You are right we can use zip

pyccel/parser/semantic.py
```diff
@@ -770,6 +770,7 @@ def _infere_type(self, expr, **settings):
     def _visit(self, expr, **settings):
         """Annotates the AST.
 
+        # TODO: Why is IndexedVariable mentioned here?
         IndexedVariable atoms are only used to manipulate expressions, we then,
         always have a Variable in the namespace."""
```
yguclu, line 775: Given that `IndexedVariable` will not exist anymore after this PR, I think that we should understand this comment (if possible) and then update it or remove it!

@saidctb Do you have any clue?

pyccel/ast/core.py
```diff
@@ -329,7 +327,7 @@ def _atomic(e, cls=None,ignore=()):
     atoms_ = []
     if cls is None:
         cls = (Application, DottedVariable, Variable,
-               IndexedVariable,IndexedElement)
+               IndexedElement)
```
yguclu, line 330: Can't we write this in a single line now?

pyccel/ast/core.py
```diff
@@ -2470,12 +2467,22 @@ def _eval_is_positive(self):
...
+    def __getitem__(self, *args):
+
+        if len(args) == 1 and isinstance(args[0], (Tuple, tuple, list)):
+            args = args[0]
+
+        if self.rank < len(args):
+            raise IndexError('Rank mismatch.')
+
+        obj = IndexedElement(self, *args)
+        return obj
```
yguclu, line 2479: Here too, we can write this in a single line.

pyccel/parser/semantic.py
EmilyBourne: Ah! I had forgotten about that. That definitely needs fixing before we merge. It seems like a message that would make sense in the `insert_variable` function but not in the `_visit` function. I suspect it can be removed in any case

saidctb: I have no idea about that comment

yguclu: Let's remove it then!

EmilyBourne: Git blame says it was written by Ahmed on 18/11/2018. Does everyone agree that it can be removed?

saidctb: yes you can remove it

yguclu (approved):

Issue #492, opened 2020-10-21
Strings in  IfTernaryOperator are not working in Fortran 


MenoIy:
Strings are not working in IfTernaryOperator in Fortran when they have different lengths. 
example code python 
```
@types('int')
def func(x):
    c = 'Hello' if x > 0 else 'Goodbye'
```
the generated Fortran code
```
subroutine func(x)

  implicit none

  integer(kind=8), value :: x
  character(len=280) :: c

  c = merge('' // 'Hello', '' // 'Goodbye', x > 0_8)

end subroutine func
```
this code doesn't compile and return the following error :
```
Error: Unequal character lengths (5/7) in MERGE intrinsic at (1)
```
merge function doesn't accept strings with different lengths.

Issue #493, opened 2020-10-21
Arrays in IfTernaryOperator are not working in Fortran


MenoIy:
Arrays are not working in IfTernaryOperator in Fortran when they have different shapes 
example code python 
```
@types('int')
def func(x):
    c = [1, 2, 3, 4] if x > 0 else [1,2,3]
    return 1
```
the generated Fortran code
```
function func(x) result(Out_0001)

  implicit none

  integer(kind=8) :: Out_0001
  integer(kind=8), value :: x
  integer(kind=8), allocatable :: c(:,:)

  allocate(c(0:0_8 - 1_8, 0:4_8 - 1_8))
  c = merge([1_8, 2_8, 3_8, 4_8], [1_8, 2_8, 3_8], x > 0_8)
  Out_0001 = 1_8
  return

end function func
```
this code doesn't compile and return the following error :

```
c = merge([1_8, 2_8, 3_8, 4_8], [1_8, 2_8, 3_8], x > 0_8)
Error: Different shape for arguments 'tsource' and 'fsource' for intrinsic 'merge' at (1) on dimension 1 (4 and 3)
```
merge function doesn't accept arrays with different shapes.

Issue #490, opened 2020-10-21
pyccel/ast/core.py is larger than codacy's maximum file size


EmilyBourne:
pyccel/ast/core.py is larger than codacy's maximum file size (150KB as specified here https://docs.codacy.com/faq/troubleshooting/why-is-my-file-missing/#file-size-limitation-on-codacy-cloud). It is currently 152kB.

We should either find a way to break it into pieces. Eg. we could create an arithmetic.py file for all basic arithmetic types

Or we should use the Codacy Analysis CLI locally

yguclu:
If we spend a little time on it, I'm sure we can shrink it down to something more manageable :smile:

EmilyBourne:
PR #559 will help with this issue.

I would like to suggest the creation of an `ast.variables` file which would contain `Variable` (and all classes that inherit from it), `DottedVariable`, `FunctionAddress` and `IndexedElement`. We may also want to include `Constant`, `Argument`, and `ValuedArgument` and also related helper classes e.g. `Declare`, `FunctionAddressDeclare`, `VariableAddress`
I think this change should be sufficient to fix several of our circular import problems as it should allow us to move the definition of `PyccelArraySize` although where it should be moved to is not clear

We should also go through and remove unused classes, e.g. `Concatenate`? `Del`?

yguclu:
> I would like to suggest the creation of an `ast.variables` file which would contain `Variable` (and all classes that inherit from it), `DottedVariable`, `FunctionAddress` and `IndexedElement`. We may also want to include `Constant`, `Argument`, and `ValuedArgument` and also related helper classes e.g. `Declare`, `FunctionAddressDeclare`, `VariableAddress`

I agree, your suggestion looks really good to me. Although usually splitting a module increases the chances of a circular import to happen, rather than the opposite...

PR #651, opened 2021-01-05
Set precedence of IfTernaryOperator

mostafaMamoni:
Fix IfTernaryOperator bug by making it inherit from PyccelOperator and set its precedence.

### DIFF
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -5155,7 +5155,7 @@ class If(Basic):
         return b
 
 
-class IfTernaryOperator(Basic, PyccelAstNode):
+class IfTernaryOperator(Basic, PyccelOperator):
     """Represent a ternary conditional operator in the code, of the form (a if cond else b)
 
     Parameters
@@ -5173,11 +5173,11 @@ class IfTernaryOperator(Basic, PyccelAstNode):
     >>> IfTernaryOperator(PyccelGt(n > 1),  5,  2)
     IfTernaryOperator(PyccelGt(n > 1),  5,  2)
     """
+    _precedence = 16
     def __init__(self, cond, value_true, value_false):
         self._cond = cond
         self._value_true = value_true
         self._value_false = value_false
-
         if self.stage == 'syntactic':
             return
         if isinstance(value_true , Nil) or isinstance(value_false, Nil):
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -49,6 +49,7 @@ from .datatypes import (datatype, DataType, CustomDataType, NativeSymbol,
 
 from .literals       import LiteralTrue, LiteralFalse, LiteralInteger, Nil
 from .literals       import LiteralImaginaryUnit, LiteralString, Literal
+from .literals       import Nil
 from .itertoolsext   import Product
 from .functionalexpr import GeneratorComprehension as GC
 from .functionalexpr import FunctionalFor
@@ -2053,7 +2054,6 @@ class ConstructorCall(AtomicExpr):
             return self.func.name
         else:
             return self.func
-
 
 class Void(Basic):
 
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -105,7 +105,6 @@ __all__ = (
     'IfTernaryOperator',
     'Import',
     'IndexedElement',
-    'IndexedVariable',
     'Interface',
     'Load',
     'ModOp',
@@ -255,7 +254,7 @@ def allocatable_like(expr, verbose=False):
         talk more
     """
 
-    if isinstance(expr, (Variable, IndexedVariable, IndexedElement)):
+    if isinstance(expr, (Variable, IndexedElement)):
         return expr
     elif isinstance(expr, str):
         # if the rhs is a string
@@ -307,8 +306,7 @@ def allocatable_like(expr, verbose=False):
                 if verbose:
                     print('Functions not yet available')
                 return None
-            elif isinstance(a, (Variable, IndexedVariable,
-                            IndexedElement)):
+            elif isinstance(a, (Variable, IndexedElement)):
                 return a
             elif a.is_Symbol:
                 raise TypeError('Found an unknown symbol {0}'.format(str(a)))
@@ -328,8 +326,7 @@ def _atomic(e, cls=None,ignore=()):
     seen = []
     atoms_ = []
     if cls is None:
-        cls = (Application, DottedVariable, Variable,
-               IndexedVariable,IndexedElement)
+        cls = (Application, Variable, IndexedElement)
 
     for p in pot:
         if p in seen or isinstance(p, ignore):
@@ -766,7 +763,6 @@ class Assign(Basic):
         rhs = self.rhs
         cond = isinstance(rhs, Variable) and rhs.rank > 0
         cond = cond or isinstance(rhs, IndexedElement)
-        cond = cond or isinstance(rhs, IndexedVariable)
         cond = cond and isinstance(lhs, Symbol)
         cond = cond or isinstance(rhs, Variable) and rhs.is_pointer
         return cond
@@ -968,7 +964,7 @@ class AliasAssign(Basic):
         at this point we don't know yet all information about lhs, this is why a
         Symbol is the appropriate type.
 
-    rhs : Variable, IndexedVariable, IndexedElement
+    rhs : Variable, IndexedElement
         an assignable variable can be of any rank and any datatype, however its
         shape must be known (not None)
 
@@ -2470,12 +2466,21 @@ class Variable(Symbol, PyccelAstNode):
         #we do this inorder to infere the type of Pow expression correctly
         return self.is_real
 
+    def __getitem__(self, *args):
+
+        if len(args) == 1 and isinstance(args[0], (Tuple, tuple, list)):
+            args = args[0]
+
+        if self.rank < len(args):
+            raise IndexError('Rank mismatch.')
+
+        return IndexedElement(self, *args)
+
 class DottedVariable(Variable):
 
     """
     Represents a dotted variable.
     """
-
     def __init__(self, *args, lhs, **kwargs):
         Variable.__init__(self, *args, **kwargs)
         self._lhs = lhs
@@ -2563,24 +2568,34 @@ class TupleVariable(Variable):
         Variable.__init__(self, dtype, name, *args, **kwargs)
 
     def get_vars(self):
-        if self._is_homogeneous:
-            indexed_var = IndexedVariable(self, dtype=self.dtype, shape=self.shape,
-                prec=self.precision, order=self.order, rank=self. rank)
-            args = [Slice(None,None)]*(self.rank-1)
-            return [indexed_var[[i] + args] for i in range(len(self._vars))]
-        else:
-            return self._vars
+        return tuple(self[i] for i in range(len(self._vars)))
 
     def get_var(self, variable_idx):
+        if isinstance(variable_idx, LiteralInteger):
+            variable_idx = variable_idx.p
         return self._vars[variable_idx]
 
     def rename_var(self, variable_idx, new_name):
         self._vars[variable_idx] = self._vars[variable_idx].clone(new_name)
 
     def __getitem__(self,idx):
-        if isinstance(idx, LiteralInteger):
-            idx = idx.p
-        return self.get_var(idx)
+        if self._is_homogeneous:
+            return Variable.__getitem__(self, idx)
+        else:
+            if isinstance(idx, tuple):
+                sub_idx = idx[1:]
+                idx = idx[0]
+            else:
+                sub_idx = []
+
+            if isinstance(idx, LiteralInteger):
+                idx = idx.p
+            var = self.get_var(idx)
+
+            if len(sub_idx) > 0:
+                return var[sub_idx]
+            else:
+                return var
 
     def __iter__(self):
         return self._vars.__iter__()
@@ -4644,131 +4659,6 @@ class CommentBlock(Basic):
     def header(self, header):
         self._header = header
 
-class IndexedVariable(IndexedBase, PyccelAstNode):
-
-    """
-    Represents an indexed variable, like x in x[i], in the code.
-
-    Examples
-    --------
-    >>> from sympy import symbols, Idx
-    >>> from pyccel.ast.core import IndexedVariable
-    >>> A = IndexedVariable('A'); A
-    A
-    >>> type(A)
-    <class 'pyccel.ast.core.IndexedVariable'>
-
-    When an IndexedVariable object receives indices, it returns an array with named
-    axes, represented by an IndexedElement object:
-
-    >>> i, j = symbols('i j', integer=True)
-    >>> A[i, j, 2]
-    A[i, j, 2]
-    >>> type(A[i, j, 2])
-    <class 'pyccel.ast.core.IndexedElement'>
-
-    The IndexedVariable constructor takes an optional shape argument.  If given,
-    it overrides any shape information in the indices. (But not the index
-    ranges!)
-
-    >>> m, n, o, p = symbols('m n o p', integer=True)
-    >>> i = Idx('i', m)
-    >>> j = Idx('j', n)
-    >>> A[i, j].shape
-    (m, n)
-    >>> B = IndexedVariable('B', shape=(o, p))
-    >>> B[i, j].shape
-    (m, n)
-
-    **todo:** fix bug. the last result must be : (o,p)
-    """
-
-    def __new__(
-        cls,
-        label,
-        shape=None,
-        dtype=None,
-        prec=0,
-        order=None,
-        rank = 0,
-        **kw_args
-        ):
-
-        if isinstance(label, Application):
-            label_name = type(label)
-        else:
-            label_name = str(label)
-
-        return IndexedBase.__new__(cls, label_name, shape=shape)
-
-    def __init__(
-        self,
-        label,
-        shape=None,
-        dtype=None,
-        prec=0,
-        order=None,
-        rank = 0,
-        **kw_args
-        ):
-
-        if dtype is None:
-            raise TypeError('datatype must be provided')
-        if isinstance(dtype, str):
-            dtype = datatype(dtype)
-        elif not isinstance(dtype, DataType):
-            raise TypeError('datatype must be an instance of DataType.')
-
-
-        self._dtype      = dtype
-        self._precision  = prec
-        self._rank       = rank
-        self._order      = order
-        kw_args['order'] = order
-        self._kw_args    = kw_args
-        self._label      = label
-
-    def __getitem__(self, *args):
-
-        if len(args) == 1 and isinstance(args[0], (Tuple, tuple, list)):
-            args = args[0]
-
-        if self.shape and len(self.shape) != len(args):
-            raise IndexError('Rank mismatch.')
-
-        obj = IndexedElement(self, *args)
-        return obj
-
-    @property
-    def order(self):
-        return self.kw_args['order']
-
-    @property
-    def kw_args(self):
-        return self._kw_args
-
-    @property
-    def name(self):
-        return self._args[0]
-
-    @property
-    def internal_variable(self):
-        return self._label
-
-
-    def clone(self, name):
-        cls = eval(self.__class__.__name__)
-        # TODO what about kw_args in __new__?
-        return cls(name, shape=self.shape, dtype=self.dtype,
-                   prec=self.precision, order=self.order, rank=self.rank)
-
-    def _eval_subs(self, old, new):
-        return self
-
-    def __str__(self):
-        return str(self.name)
-
-
 class IndexedElement(Expr, PyccelAstNode):
 
     """
@@ -4777,9 +4667,9 @@ class IndexedElement(Expr, PyccelAstNode):
     Examples
     --------
     >>> from sympy import symbols, Idx
-    >>> from pyccel.ast.core import IndexedVariable, IndexedElement
+    >>> from pyccel.ast.core import Variable, IndexedElement
     >>> i, j = symbols('i j', cls=Idx)
-    >>> A = IndexedVariable('A', dtype='int')
+    >>> A = Variable('A', dtype='int')
     >>> IndexedElement(A, i, j)
     IndexedElement(A, i, j)
     >>> IndexedElement(A, i, j) == A[i, j]
@@ -4795,19 +4685,8 @@ class IndexedElement(Expr, PyccelAstNode):
 
         if not args:
             raise IndexError('Indexed needs at least one index.')
-        if isinstance(base, (str, Symbol)):
-            base = IndexedBase(base)
-        elif not hasattr(base, '__getitem__') and not isinstance(base,
-                IndexedBase):
-            raise TypeError(filldedent("""
-                Indexed expects string, Symbol, or IndexedBase as base."""))
-
-        if isinstance(base, (NDimArray, Iterable, Tuple,
-                      MatrixBase)) and all([i.is_number for i in args]):
-            if len(args) == 1:
-                return base[args[0]]
-            else:
-                return base[args]
+        if not isinstance(base, Variable):
+            raise TypeError("Indexed expects Variable as base")
         return Expr.__new__(cls, base, *args, **kw_args)
 
     def __init__(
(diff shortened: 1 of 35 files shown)


EmilyBourne:
Please add a test to prevent this regressing

pyccel/codegen/printing/ccode.py
```diff
@@ -368,7 +368,7 @@ def _print_IfTernaryOperator(self, expr):
         cond = self._print(expr.cond)
         value_true = self._print(expr.value_true)
         value_false = self._print(expr.value_false)
-        return '({cond}) ? {true} : {false}'.format(cond = cond, true =value_true, false = value_false)
+        return '(({cond}) ? {true} : {false})'.format(cond = cond, true =value_true, false = value_false)
```
EmilyBourne, line 371: This problem is not specific to C. The issue mentions changing the inheritance of `IfTernaryOperator` but I do not see that here

EmilyBourne (changes requested):

pyccel/ast/core.py
```diff
@@ -5155,7 +5155,7 @@ def bodies(self):
         return b
 
 
-class IfTernaryOperator(Basic, PyccelAstNode):
+class IfTernaryOperator(Basic, PyccelOperator):
```
EmilyBourne, line 5158: Could you move this operator to `ast.operators` please

mostafaMamoni: i almost want to suggest that.

pyccel/ast/operators.py
```diff
@@ -1070,7 +1070,80 @@ class PyccelIsNot(PyccelIs):
...
+    >>> from sympy import Symbol
+    >>> from pyccel.ast.core import Assign
+	>>>	from pyccel.ast.operators import IfTernaryOperator
+    >>> n = Symbol('n')
+    >>> x = 5 if n > 1 else 2
+    >>> IfTernaryOperator(PyccelGt(n > 1),  5,  2)
+    IfTernaryOperator(PyccelGt(n > 1),  5,  2)
+    """
+    _precedence = 3
+    def __init__(self, cond, value_true, value_false):
```
EmilyBourne, line 1096: I agree with codacy. Superclass `__init__` methods should ideally always be called. `PyccelOperator.__init__` ensures that the dtype/precision/etc are set, however it does this by calling additional methods so if this Operator needs to implement it's own version of those it can do so by overriding the virtual methods

pyccel/ast/operators.py
```diff
@@ -1070,7 +1070,80 @@ class PyccelIsNot(PyccelIs):
...
+        _tmp_list = [NativeBool(), NativeInteger(), NativeReal(), NativeComplex(), NativeString()]
+        if value_true.dtype not in _tmp_list :
+            raise NotImplementedError('cannot determine the type of {}'.format(value_true.dtype))
+        if value_false.dtype not in _tmp_list :
+            raise NotImplementedError('cannot determine the type of {}'.format(value_false.dtype))
+        if value_false.rank != value_true.rank :
+            errors.report('Ternary Operator results should have the same rank', severity='fatal')
+        if value_false.shape != value_true.shape :
+            errors.report('Ternary Operator results should have the same shape', severity='fatal')
+        self._dtype = max([value_true.dtype, value_false.dtype], key = lambda x : _tmp_list.index(x))
```
EmilyBourne, line 1115: Codacy says `Lambda may not be necessary`. This is because you can just do :
```
self._dtype = max([value_true.dtype, value_false.dtype], key = _tmp_list.index)
```

pyccel/ast/operators.py
EmilyBourne: It is also `PyccelOperator.__init__` which calls the functions which ensure that the parentheses are added where necessary so the bug will persist as long as it isn't called

EmilyBourne (changes requested):
Please add a test

pyccel/ast/literals.py
```diff
@@ -217,3 +217,12 @@ def get_default_literal_value(dtype):
         raise TypeError('Unknown type')
     return value
 
+class Nil(Basic):
+
+    """
+    class for None object in the code.
+    """
+
+    def _sympystr(self, printer):
```
EmilyBourne: You should merge PR #659 in to update this function

pyccel/ast/operators.py
```diff
@@ -1070,7 +1070,91 @@ class PyccelIsNot(PyccelIs):
...
+        self._value_true = self._args[1]
+        self._value_false = self._args[2]
+        if self.stage == 'syntactic':
+            return
+        if isinstance(value_true , Nil) or isinstance(value_false, Nil):
+            errors.report('None is not implemented for Ternary Operator', severity='fatal')
+        if isinstance(value_true.dtype, NativeString) or isinstance(value_false.dtype, NativeString):
+            errors.report('Strings are not supported by Ternary Operator', severity='fatal')
+        _tmp_list = [NativeBool(), NativeInteger(), NativeReal(), NativeComplex(), NativeString()]
+        if value_true.dtype not in _tmp_list :
```
EmilyBourne, line 1108: Is this list needed or could it be simplified to `if not isinstance (value_true.dtype, DataType)` ? (@MenoIy )

pyccel/ast/operators.py
```diff
@@ -1070,7 +1070,91 @@ class PyccelIsNot(PyccelIs):
...
+        if value_false.rank != value_true.rank :
+            errors.report('Ternary Operator results should have the same rank', severity='fatal')
+        if value_false.shape != value_true.shape :
+            errors.report('Ternary Operator results should have the same shape', severity='fatal')
+
+    def _set_dtype(self):
+        """
+        Sets the dtype and precision for IfTernaryOperator
+        """
+        _tmp_list = [NativeBool(), NativeInteger(), NativeReal(), NativeComplex(), NativeString()]
```
EmilyBourne, line 1121: We should avoid duplicating things. If the list is needed then maybe it should be a class member?

pyccel/codegen/printing/fcode.py
```diff
@@ -38,10 +37,14 @@
...
+                             If, PyccelArraySize)
 
 
 from pyccel.ast.operators      import PyccelAdd, PyccelMul, PyccelDiv, PyccelMinus
+
+from pyccel.ast.operators      import PyccelUnarySub, PyccelLt, PyccelGt, IfTernaryOperator
+from pyccel.ast.core      import FunctionCall
+
 from pyccel.ast.operators      import PyccelUnarySub, PyccelLt, PyccelGt
 from pyccel.ast.core      import FunctionCall, DottedFunctionCall
```
EmilyBourne, line 49: There are duplicated imports here

pyccel/ast/operators.py
```diff
@@ -1070,7 +1070,92 @@ class PyccelIsNot(PyccelIs):
...
+    >>> IfTernaryOperator(PyccelGt(n > 1),  5,  2)
+    IfTernaryOperator(PyccelGt(n > 1),  5,  2)
+    """
+    _precedence = 3
+    def __init__(self, cond, value_true, value_false):
+        PyccelOperator.__init__(self, cond, value_true, value_false)
+
+        self._cond = self._args[0]
+        self._value_true = self._args[1]
+        self._value_false = self._args[2]
```
EmilyBourne, line 1101: Are these variables ever used?

pyccel/ast/operators.py
MenoIy: the list is needed for getting the data type of the ternary operator.

EmilyBourne: Ah yes I see. So then it is needed in the `_set_dtype` function but not here?

pyccel/codegen/printing/fcode.py
EmilyBourne: `PyccelUnarySub, PyccelLt, PyccelGt` are still duplicated

pyccel/ast/operators.py
EmilyBourne: @MenoIy ?

EmilyBourne:
Related to #490 . After this merge ast.core seems to be tested

pyccel/ast/operators.py
```diff
@@ -1070,7 +1070,90 @@ class PyccelIsNot(PyccelIs):
...
+    >>> from sympy import Symbol
+    >>> from pyccel.ast.core import Assign
+	>>>	from pyccel.ast.operators import IfTernaryOperator
+    >>> n = Symbol('n')
+    >>> x = 5 if n > 1 else 2
+    >>> IfTernaryOperator(PyccelGt(n > 1),  5,  2)
+    IfTernaryOperator(PyccelGt(n > 1),  5,  2)
+    """
+    _precedence = 3
+    _dtype_list = [NativeBool(), NativeInteger(), NativeReal(), NativeComplex(), NativeString()]
```
EmilyBourne, line 1096: If we keep this then it should be a tuple not a list as we will never want to change this object dynamically

pyccel/ast/operators.py
EmilyBourne: @mostafaMamoni maybe you can try this? I can't think of any reason why it wouldn't work

pyccel/ast/operators.py
```diff
@@ -1070,7 +1070,90 @@ class PyccelIsNot(PyccelIs):
...
+    >>> from sympy import Symbol
+    >>> from pyccel.ast.core import Assign
+	>>>	from pyccel.ast.operators import IfTernaryOperator
+    >>> n = Symbol('n')
+    >>> x = 5 if n > 1 else 2
+    >>> IfTernaryOperator(PyccelGt(n > 1),  5,  2)
+    IfTernaryOperator(PyccelGt(n > 1),  5,  2)
+    """
+    _precedence = 3
+    _dtype_list = (NativeBool(), NativeInteger(), NativeReal(), NativeComplex(), NativeString())
```
EmilyBourne, line 1096: Maybe change the name now it's no longer a list?

pyccel/ast/operators.py
MenoIy: is there any function set_type? like the one in PyccelBinaryOperator ?

MenoIy: it should be ok using set_dtype  now that it inherits from pycelOperator

pyccel/ast/operators.py
mostafaMamoni: i thought "list" in the name refers to metaphorical meaning not the type of the variable.

pyccel/ast/operators.py
mostafaMamoni: i thought "list" in the name refers to metaphorical meaning not the type of the variable.

pyccel/ast/operators.py
```diff
@@ -1070,7 +1070,87 @@ class PyccelIsNot(PyccelIs):
...
+    """
+    _precedence = 3
+
+    def __init__(self, cond, value_true, value_false):
+        super().__init__(cond, value_true, value_false)
+
+        if self.stage == 'syntactic':
+            return
+        if isinstance(value_true , Nil) or isinstance(value_false, Nil):
+            errors.report('None is not implemented for Ternary Operator', severity='fatal')
```
EmilyBourne, line 1103: You seem to have lost the error message about `NativeString`. Was that intentional?

pyccel/ast/operators.py
```diff
@@ -1070,7 +1070,87 @@ class PyccelIsNot(PyccelIs):
...
+        if value_false.rank != value_true.rank :
+            errors.report('Ternary Operator results should have the same rank', severity='fatal')
+        if value_false.shape != value_true.shape :
+            errors.report('Ternary Operator results should have the same shape', severity='fatal')
+
+    def _set_dtype(self):
+        """
+        Sets the dtype and precision for IfTernaryOperator
+        """
+        self._dtype = max([self.value_true.dtype, self.value_false.dtype], key = NativeNumeric.index)
```
EmilyBourne, line 1117: I think this should be an `if` to handle non numeric types (e.g. a class instance)

pyccel/ast/operators.py
```diff
@@ -1070,7 +1070,87 @@ class PyccelIsNot(PyccelIs):
...
+    def _set_shape_rank(self):
+        """
+        Sets the shape and rank and the order for IfTernaryOperator
+        """
+        self._shape = self.value_true.shape
+        self._rank  = self.value_true.rank
+        if self._rank is not None and self._rank > 1:
+            if self.value_false.order != self.value_true.order :
+                errors.report('Ternary Operator results should have the same order', severity='fatal')
+            self._order = self.value_true.order
```
EmilyBourne, line 1129: Does `PyccelOperator` not have a set_order method?

pyccel/ast/operators.py
```diff
@@ -1070,7 +1070,89 @@ class PyccelIsNot(PyccelIs):
...
+        if value_false.rank != value_true.rank :
+            errors.report('Ternary Operator results should have the same rank', severity='fatal')
+        if value_false.shape != value_true.shape :
+            errors.report('Ternary Operator results should have the same shape', severity='fatal')
+
+    def _set_dtype(self):
+        """
+        Sets the dtype and precision for IfTernaryOperator
+        """
+        if self in NativeNumeric:
```
EmilyBourne, line 1119: This line is wrong and the `else` block is missing

EmilyBourne (changes requested):

pyccel/ast/operators.py
```diff
@@ -1070,7 +1070,93 @@ class PyccelIsNot(PyccelIs):
...
+            self._dtype = max([self.value_true.dtype, self.value_false.dtype], key = NativeNumeric.index)
+        elif isinstance(self, Nil):
+            self._dtype = Nil()
+        elif isinstance(self, NativeString):
+            self._dtype = NativeString()
```
EmilyBourne, line 1124: If blocks should almost always have an else. This one needs one

pyccel/ast/operators.py
```diff
@@ -1070,7 +1070,93 @@ class PyccelIsNot(PyccelIs):
...
+            errors.report('Ternary Operator results should have the same shape', severity='fatal')
+
+    def _set_dtype(self):
+        """
+        Sets the dtype and precision for IfTernaryOperator
+        """
+        if self in NativeNumeric:
+            self._dtype = max([self.value_true.dtype, self.value_false.dtype], key = NativeNumeric.index)
+        elif isinstance(self, Nil):
+            self._dtype = Nil()
```
EmilyBourne, line 1122: The dtype must always inherit from `DataType` or be `None`. `Nil` is the pyccel version of none in the code, but it is not the same as none

EmilyBourne (changes requested):

pyccel/ast/operators.py
EmilyBourne: This line is still wrong

pyccel/ast/operators.py
```diff
@@ -1070,7 +1070,93 @@ class PyccelIsNot(PyccelIs):
...
+        """
+        Sets the dtype and precision for IfTernaryOperator
+        """
+        if self in NativeNumeric:
+            self._dtype = max([self.value_true.dtype, self.value_false.dtype], key = NativeNumeric.index)
+        elif isinstance(self, NativeString):
+            self._dtype = NativeString()
+        else:
+            self._dtype = None
+        self._precision = max([self.value_true.precision, self.value_false.precision])
```
EmilyBourne, line 1125: The dtype should be None only if there is a `Nil` (is this condition necessary given that pyccel does not allow Nil to be passed to an IfTernaryOperator? ) Otherwise it is known. This block does not treat all the places where it is known (e.g  if the DataType is a class type)

EmilyBourne (changes requested):

pyccel/ast/operators.py
```diff
@@ -1070,7 +1070,91 @@ class PyccelIsNot(PyccelIs):
...
+        if value_false.rank != value_true.rank :
+            errors.report('Ternary Operator results should have the same rank', severity='fatal')
+        if value_false.shape != value_true.shape :
+            errors.report('Ternary Operator results should have the same shape', severity='fatal')
+
+    def _set_dtype(self):
+        """
+        Sets the dtype and precision for IfTernaryOperator
+        """
+        if isinstance(self, NativeString):
```
EmilyBourne, line 1119: This line is not right. Self is an IfTernaryOperator, never a NativeString

pyccel/ast/operators.py
EmilyBourne: The new block handles classes incorrectly

EmilyBourne (changes requested):

pyccel/ast/operators.py
```diff
@@ -1070,7 +1070,93 @@ class PyccelIsNot(PyccelIs):
...
+            errors.report('Ternary Operator results should have the same shape', severity='fatal')
+
+    def _set_dtype(self):
+        """
+        Sets the dtype and precision for IfTernaryOperator
+        """
+        if isinstance(self.value_true , Nil) or isinstance(self.value_false, Nil):
+            errors.report('None is not implemented for Ternary Operator', severity='fatal')
+        if isinstance(self.value_true , NativeString) or isinstance(self.value_false, NativeString):
+            errors.report('String is not implemented for Ternary Operator', severity='fatal')
```
EmilyBourne, line 1122: These tests are already done in the `__init__` function. They do not need to be done twice

pyccel/ast/operators.py
```diff
@@ -1070,7 +1070,93 @@ class PyccelIsNot(PyccelIs):
...
+        else:
+            self._dtype = max([self.value_true.dtype, self.value_false.dtype], key = NativeNumeric.index)
```
EmilyBourne, line 1124: This else will catch classes too and will break

EmilyBourne (changes requested):

pyccel/ast/operators.py
```diff
@@ -1070,7 +1070,94 @@ class PyccelIsNot(PyccelIs):
...
+        if self.stage == 'syntactic':
+            return
+        if isinstance(value_true , Nil) or isinstance(value_false, Nil):
+            errors.report('None is not implemented for Ternary Operator', severity='fatal')
+        if isinstance(value_true , NativeString) or isinstance(value_false, NativeString):
+            errors.report('String is not implemented for Ternary Operator', severity='fatal')
+        assert value_true.dtype in NativeNumeric
+        assert value_false.dtype in NativeNumeric
+        if value_true.dtype not in NativeNumeric:
+            assert(value_true.dtype == value_false.dtype)
```
EmilyBourne, line 1109: These assertions don't make sense together. There should be an if/else. In addition we should raise neat errors rather than using assertions. That said these tests are carried out in the `set_dtype` function where an error is raised so these are not necessary

tests/epyccel/test_epyccel_IfTernaryOperator.py
```diff
@@ -4,7 +4,7 @@
 
 from pyccel.epyccel import epyccel
 from pyccel.decorators import types
-
+# pw suffix means With Parentheses
```
EmilyBourne, line 7: Typo `pw`->`wp`?

EmilyBourne (changes requested):

EmilyBourne (changes requested):
Only one of the 3 points in my [previous review](https://github.com/pyccel/pyccel/pull/651#discussion_r555977065) has been addressed

pyccel/ast/operators.py
```diff
@@ -1070,7 +1070,94 @@ class PyccelIsNot(PyccelIs):
...
+        super().__init__(cond, value_true, value_false)
+
+        if self.stage == 'syntactic':
+            return
+        if isinstance(value_true , Nil) or isinstance(value_false, Nil):
+            errors.report('None is not implemented for Ternary Operator', severity='fatal')
+        if isinstance(value_true , NativeString) or isinstance(value_false, NativeString):
+            errors.report('String is not implemented for Ternary Operator', severity='fatal')
+        if value_true.dtype not in NativeNumeric and value_false.dtype not in NativeNumeric:
+            errors.report('unkown data type in if ternary operator', severity='fatal')
```
EmilyBourne, line 1107: Are you sure this condition makes sense? What will happen if the dtype is a class?

pyccel/ast/operators.py
```diff
@@ -1070,7 +1070,94 @@ class PyccelIsNot(PyccelIs):
...
+    def __init__(self, cond, value_true, value_false):
+        super().__init__(cond, value_true, value_false)
+
+        if self.stage == 'syntactic':
+            return
+        if isinstance(value_true , Nil) or isinstance(value_false, Nil):
+            errors.report('None is not implemented for Ternary Operator', severity='fatal')
+        if isinstance(value_true , NativeString) or isinstance(value_false, NativeString):
+            errors.report('String is not implemented for Ternary Operator', severity='fatal')
+        if value_true.dtype not in NativeNumeric or value_false.dtype not in NativeNumeric:
```
EmilyBourne, line 1106: Shouldn't this be an xor?

pyccel/ast/operators.py
```diff
@@ -1070,7 +1070,94 @@ class PyccelIsNot(PyccelIs):
...
+            errors.report('unkown data type in IfTernaryOperator', severity='fatal')
```
EmilyBourne, line 1107: This message does not describe the problem that is tested here.

pyccel/ast/operators.py
```diff
@@ -1070,7 +1070,94 @@ class PyccelIsNot(PyccelIs):
...
+        if value_true.dtype not in NativeNumeric:
+            assert(value_true.dtype == value_false.dtype)
```
EmilyBourne, line 1109: Please raise neat errors rather than raising assertions

mostafaMamoni: actually this should be removed, it doesn't make sense after the previous statement.

EmilyBourne: That is because the previous statement is false

EmilyBourne: You should have 3 cases:
1. `value_true.dtype in NativeNumeric  and value_false.dtype in NativeNumeric`
    **OK** : no error message (this may not need an if)
    Both values are numeric and are therefore compatible
    (e.g. `value_true = 4j`, `value_false = 4`)
2. `value_true.dtype == value_false.dtype`
    **OK** : no error message (this may not need an if)
    The values are non-numeric (as this is not case 1) but they are the same dtype so they are compatible
    (e.g. both values are instances of the same class)
3. Anything else
    **ERROR** : The types are incompatible, a clear error must be raised
    We don't know how to specify a dtype for different dtypes unless both dtypes are numeric so there is a problem

pyccel/ast/operators.py
mostafaMamoni: i didn't get the point here.

EmilyBourne: See https://github.com/pyccel/pyccel/pull/651#discussion_r557135049
Think about what will happen currently if the dtypes are both class instances

pyccel/ast/operators.py
mostafaMamoni: great, i think this will fix #492 and #493, what do you think?

EmilyBourne: No it won't. Those 2 issues are about fortran specific problems. The equivalent of the `IfTernaryOperator` in Fortran (`merge` ) cannot handle objects with different sizes. So in Fortran an `IfTernaryOperator` needs to be expanded into an `If` block in these 2 cases. This does nothing related to those issues

pyccel/ast/operators.py
```diff
@@ -1103,10 +1103,9 @@ def __init__(self, cond, value_true, value_false):
             errors.report('None is not implemented for Ternary Operator', severity='fatal')
         if isinstance(value_true , NativeString) or isinstance(value_false, NativeString):
             errors.report('String is not implemented for Ternary Operator', severity='fatal')
-        if value_true.dtype not in NativeNumeric or value_false.dtype not in NativeNumeric:
-            errors.report('unkown data type in IfTernaryOperator', severity='fatal')
-        if value_true.dtype not in NativeNumeric:
-            assert(value_true.dtype == value_false.dtype)
+        if value_true.dtype != value_false.dtype:
+            if value_true.dtype not in NativeNumeric and value_false.dtype not in NativeNumeric:
```
EmilyBourne, line 1107: The following dtypes won't raise an error:
ClassType + NativeReal

EmilyBourne (changes requested):

pyccel/ast/operators.py
```diff
@@ -1070,7 +1070,93 @@ class PyccelIsNot(PyccelIs):
...
+    def _set_dtype(self):
+        """
+        Sets the dtype and precision for IfTernaryOperator
+        """
+        if self.value_true.dtype in NativeNumeric and self.value_false.dtype in NativeNumeric:
+            self._dtype = max([self.value_true.dtype, self.value_false.dtype], key = NativeNumeric.index)
+        elif self.value_true.dtype == self.value_false.dtype:
+            self._dtype = self.value_true.dtype
+        else:
+            errors.report('Ternary Operator results should have the same data type', severity='fatal')
```
EmilyBourne, line 1123: We test if the types are compatible in `__init__` this doesn't need to be done twice so this block can be simplified to:
```
if self.value_true.dtype in NativeNumeric and self.value_false.dtype in NativeNumeric:
            self._dtype = max([self.value_true.dtype, self.value_false.dtype], key = NativeNumeric.index)
else:
            self._dtype = self.value_true.dtype
```

pyccel/ast/operators.py
```diff
@@ -1070,7 +1070,93 @@ class PyccelIsNot(PyccelIs):
...
+    def value_true(self):
+        """
+        The value_if_cond_true property for IfTernaryOperator class
+        """
+        return self._args[1]
+
+    @property
+    def value_false(self):
+        """
+        The value_if_cond_flase property for IfTernaryOperator class
```
EmilyBourne, line 1154: Typo `flase`->`false`

EmilyBourne (approved):

yguclu (approved):

Issue #643, opened 2020-12-30
Improve Python Printer (pycode)


ratnania:
Right now, our Python Printer is not up-to-date. It was mainly improved so that Psydac works.

EmilyBourne:
We should allow pyccel to translate from python to python. That way we could use all our existing tests to test the python printer

ratnania:
Yes!!

EmilyBourne:
This issue can be closed when the following tasks have been completed:
- [ ] Activate python tests in `tests/pyccel/` folder
- [ ] Add openmp printing to python printer

Additional problems are described by #802 and `ValuedVariable` problems (#700, #696)

PR #652, opened 2021-01-05
Allow translating from python to python

EmilyBourne:
As explained in issue #643 the python printer is no longer up to date. Even basic things such as `return` do not work. This is because this code is never tested. In order to avoid problems in psydac we should therefore test translation to python too. This PR makes translating from python to python possible and activates some of the few tests which work.

**Commit Summary**
- Update python Import printer
-  When translating to python stop after code generation and copy file to output location
- Stop pyccel command from overwriting original python file
- Only check for shared library when not translating to python
- Activate some tests
- Group python tests in the workflow

### DIFF
--- a/.github/actions/pytest_run/action.yml
+++ b/.github/actions/pytest_run/action.yml
@@ -6,7 +6,8 @@ runs:
     - name: Test with pytest
       run: |
         python -m pytest -rx -m "not parallel and c" --ignore=ast --ignore=printers --ignore=symbolic --ignore=ndarrays
-        python -m pytest -rx -m "not parallel and not c" --ignore=ast --ignore=printers --ignore=symbolic --ignore=ndarrays
+        python -m pytest -rx -m "not parallel and not c and not python" --ignore=ast --ignore=printers --ignore=symbolic --ignore=ndarrays
+        python -m pytest -rx -m "not parallel and python" --ignore=ast --ignore=printers --ignore=symbolic --ignore=ndarrays
         python -m pytest ndarrays/ -rx
         mpiexec -n 4 ${MPI_OPTS} python -m pytest epyccel/test_parallel_epyccel.py -v -m parallel -rx
         #mpiexec -n 4 ${MPI_OPTS} python -m pytest epyccel -v -m parallel -rx
--- a/pyccel/codegen/pipeline.py
+++ b/pyccel/codegen/pipeline.py
@@ -211,8 +211,8 @@ def execute_pyccel(fname, *,
             libs.append('iomp5')
 
     # ...
-    # Construct flags for the Fortran compiler
-    if fflags is None:
+    # Construct flags for the compiler (if one is required)
+    if fflags is None and compiler:
         fflags = construct_flags(f90exec,
                                  fflags=None,
                                  debug=debug,
@@ -295,6 +295,14 @@ def execute_pyccel(fname, *,
         if errors.has_errors():
             handle_error('code generation')
             raise PyccelCodegenError('Code generation failed')
+
+        if language == 'python':
+            basename = os.path.basename(fname)
+            new_location = os.path.join(folder, basename)
+            if verbose:
+                print("cp {} {}".format(fname, new_location))
+            shutil.copyfile(fname, new_location)
+            continue
 
         #------------------------------------------------------
         # TODO: collect dependencies and proceed recursively
--- a/pyccel/commands/console.py
+++ b/pyccel/commands/console.py
@@ -234,6 +234,10 @@ def pyccel(files=None, openmp=None, openacc=None, output_dir=None, compiler=None
 
     base_dirpath = os.getcwd()
 
+    if args.language == 'python' and args.output == '':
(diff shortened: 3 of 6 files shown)


pyccel/codegen/pipeline.py
```diff
@@ -296,6 +296,14 @@ def handle_error(stage):
             handle_error('code generation')
             raise PyccelCodegenError('Code generation failed')
 
+        if language == 'python':
```
saidctb, line 299: do we have a problem in case of a program ?

EmilyBourne: Yes. The main problem being that the python printer has no `_print_Program` method. But the file is generated in the correct location so I don't think there's a problem here

saidctb: okay

saidctb (approved):

ratnania:
Some intial devs were done in the brand devel-643.
I agree with @EmilyBourne the best thing would be to integrate pycode in our tests.
Do not forget that since we are also transforming some statements like zip map etc it is important to keep all our printers up to date. I would even go further:
Python to Python to C/Fortran
This would be a good validation

EmilyBourne:
@ratnania 
> I would even go further:
> Python to Python to C/Fortran

I'm not sure I see an advantage to this. It would slow down our tests by effectively doubling the work that needs to be done and nothing would be tested that wouldn't be tested by just testing
Python to C/Fortran/Python

The only "advantage" that I can imagine is that nothing can be implemented in the Fortran/C printers without first being implemented in the python printer. While this may seem like an advantage for other projects in the pyccel group (e.g. psydac) I don't think it is an advantage in pyccel whose main aim is to translate python to a faster language to accelerate code

ratnania:
I agree. It is not a good idea to do it for all our tests. But maybe we should think about it for other libraries that will be generating Python code.
We can have some nightly builds where we perform additional tests like evaluating the compiling time on some long files...

.github/actions/pytest_run/action.yml
```diff
@@ -6,7 +6,8 @@ runs:
     - name: Test with pytest
       run: |
         python -m pytest -rx -m "not parallel and c" --ignore=ast --ignore=printers --ignore=symbolic --ignore=ndarrays
-        python -m pytest -rx -m "not parallel and not c" --ignore=ast --ignore=printers --ignore=symbolic --ignore=ndarrays
+        python -m pytest -rx -m "not parallel and not c and not python" --ignore=ast --ignore=printers --ignore=symbolic --ignore=ndarrays
```
yguclu, line 9: Could we just pass the flag `-m "not parallel and fortran"`? I think we use the `fortran` mark quite consistently now.

pyccel/commands/console.py
```diff
@@ -234,6 +234,10 @@ def pyccel(files=None, openmp=None, openacc=None, output_dir=None, compiler=None
 
     base_dirpath = os.getcwd()
 
+    if args.language == 'python' and args.output == '':
+        print("Cannot output python file to same folder as this would overwrite the original file. Please specify --output")
+        sys.exit(1)
```
yguclu, line 239: When we use `epyccel`, we should also check that the output directory is not `__epyccel__` because we would overwrite the file there. Or is that exactly what you want?

.github/actions/pytest_run/action.yml
EmilyBourne: Unfortunately not. The `fortran` mark is currently being added thanks to the `language` fixture. However no `fortran` marks have been added manually to functions that only run in one language (e.g. `test_arrays.py` currently).

If we want to do this at some stage it might also be useful to add another mark for language independent tests such as the error tests, `test_arrays_multiple_assignments.py`, etc

pyccel/commands/console.py
EmilyBourne: I didn't think it would matter if we overwrote that file given that it is a temporary created by pyccel anyway. If we want to save it in a different folder then we need to add some logic to `epyccel.py` to choose the import folder. Currently we do:
```
# Import shared library
sys.path.insert(0, epyccel_dirpath)
```
which means that we expect the importable object to be in the `__epyccel__` folder

EmilyBourne: The logic that needs adding is not particularly complicated though, so I can add it if you think it would be useful to have all 3 python files?

saidctb: I don't think `epyccel` is an issue because it's not created by the user

yguclu:
@EmilyBourne Can we condense the title of this PR, in order to use it as the title line of the squashed commit message?

.github/actions/pytest_run/action.yml
yguclu: I see, you're right. We may need to rethink how we manage multiple languages in our unit tests.

pyccel/commands/console.py
yguclu: OK then, let's keep things as they are. Thanks!

yguclu (approved):

PR #654, opened 2021-01-05
Document Pyccel container image usage

noushi:
This PR documents Pyccel container image usage based on https://github.com/noushi/pyccel-docker generated images.
As of now, images are available on Docker Hub ( https://hub.docker.com/r/pyccel/pyccel/ ) and GitHub Container Registry ( ghcr.io , private at the moment).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -24,6 +24,16 @@ Pyccel comes with a selection of **extensions** allowing you to convert calls to
 - mpi4py
 - h5py (not available yet)
 
+Pyccel Installation Methods
+***************************
+
+Pyccel can be installed / deployed in different ways:
+
+1. using Pypi
+2. using containers
+3. from source
+
+  
 Getting Started
 ===============
 
--- a/README.rst
+++ b/README.rst
@@ -24,6 +24,27 @@ Pyccel comes with a selection of **extensions** allowing you to convert calls to
 - mpi4py
 - h5py (not available yet)
 
+Getting Started
+===============
+
+Pyccel container images are available through both Docker Hub (docker.io) and GitHub Container Registry (ghcr.io).
+
+Image tags match pyccel releases.
+
+In order to implement your pyccel accelerated code, you can use a host based volume during the pyccel container creation
+
+For example::
+  docker pull pyccel/pyccel:v1.0.0
+  docker run -it -v $PWD:/data:rwz  pyccel/pyccel:v1.0.0 bash
+
+About Pyccel Container Images
+*****************************
+
+The image is :
+- based on ubuntu:latest
+- uses distro packaged python3, gcc, gfortran, blas and openmpi
+
+
 Requirements
 ============
 
@@ -55,6 +76,8 @@ Finally, Pyccel supports distributed-memory parallel programming through the Mes
 We recommend using GFortran/Gcc and Open-MPI.
 
 Pyccel also depends on several Python3 packages, which are automatically downloaded by pip, the Python Package Installer, during the installation process. In addition to these, unit tests require the *scipy*, *mpi4py*, *pytest* and *coverage* packages, while building the documentation requires Sphinx <http://www.sphinx-doc.org/>.
+
+
 
 Linux Debian/Ubuntu/Mint
 ************************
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -24,6 +24,34 @@ Pyccel comes with a selection of **extensions** allowing you to convert calls to
 - mpi4py
 - h5py (not available yet)
 
+Getting Started
+===============
```
EmilyBourne, line 28: I'm not sure the ordering of the README is quite right. I suspect the average user will use pip rather than docker, put the pip installation instructions seem to have become lost at the end of the file.

What do you think @yguclu @saidctb ?

noushi: @EmilyBourne , true, that was one purpose for a few reasons:
1. docker based pyccel is basically a one line install which is way simpler than the other installation methods (requirements being considered)
2. as such, it fits into a getting started which should be one of the first sections

we want users to easily try out pyccel, and if they like it, they can install it the traditional way.

EmilyBourne: Ok. I still think it would be worth having a line at the beginning to say that an installation with pip is possible even if the details are later.

Most people I know who do not work in IT (including me) have never used docker or have never heard of docker. For those people, having instructions which at first glance imply that installation requires docker, may not be the most encouraging thing

noushi: I understand so I suggest an introductory chapter specifying all the deployment/installation methods available and then followed by the getting started chapter.

EmilyBourne: That sounds good to me :+1:

noushi: @EmilyBourne done.
can you check?

EmilyBourne (approved):

README.rst
yguclu: > I'm not sure the ordering of the README is quite right. I suspect the average user will use pip rather than docker, put the pip installation instructions seem to have become lost at the end of the file.
> 
> What do you think @yguclu @saidctb ?

@EmilyBourne: Sorry for the late reply. @noushi: Thanks for your PR!

I am also in the crowd of people that have never used Docker, and do not see any advantage in using it for their daily work. As far as I know, none of my colleagues use it, and I suspect that many of them does not know what it is. I am sure that some people will be happy to see a Docker file available, but I don't know any.

Most likely, people who want to try Pyccel will probably use it to accelerate one file in one of their Python projects, which is installed on their machine together with all the Python packages required by it. So Pyccel will need to work in that exact environment, maybe on macOS or even Windows. This is why Pyccel needs to be platform-independent, able to use multiple compilers, and with a minimal set of package dependencies.

README.rst
```diff
@@ -24,6 +24,44 @@ Pyccel comes with a selection of **extensions** allowing you to convert calls to
...
 - h5py (not available yet)
 
+Pyccel Installation Methods
+***************************
+
+Pyccel can be installed / deployed in different ways:
+
+1. using Pypi
+2. using containers
+3. from source
```
yguclu, line 34: I find this list a bit confusing: Pyccel is always installed in one unique way, using `pip`.

The user can choose to install a version from PyPI, or the latest "trunk" version from GitHub, but this does not change anything really. In both cases `pip` will install the required Python dependencies. And in both cases we are installing the Python source files, not some pre-compiled binaries.

A separate discussion can be made for how to install the non-Python dependencies. So far we have assumed that the user can take care of the installations, by following step-by-step the simple instructions given below for the various operating systems. If we were to use `conda`, we could have it install the non-Python dependencies, but we are no conda experts and did not want to maintain that.

A Docker image bypasses all of that by including everything in the container, Pyccel + Python dependencies + non-Python dependencies + Linux operating system. But you still use `pip` to install Pyccel inside the Docker container.

So I think this should read more accurately:

> Pyccel can be installed on virtually any machine that provides Python 3, the `pip` package manager, a C/Fortran compiler, and an Internet connection. Some advanced features of Pyccel require additional non-Python libraries to be installed, for which we provide detailed instructions below.
>
> Alternatively, Pyccel can be deployed through a **Linux Docker image** that contains all dependencies, and which can be setup with any version of Pyccel.

yguclu: I can make this changes myself in this PR.

yguclu: Sorry, apparently I cannot. This PR comes from a fork of Pyccel. I will create a new PR instead.

yguclu:
I've made the appropriate changes in PR #674, so I'm closing this PR now.

PR #655, opened 2021-01-06
Remove `sympy.Function` from pyccel

EmilyBourne:
Replace `sympy.Function` with FunctionCall and a new class `PyccelInternalFunction`. This is the first step in the epic #665, as such it closes #655 

**Commit Summary**
- Add different handling to allow the use of FunctionCall in syntactic stage
- Use FunctionCall instead of sympy's Function
    - `_visit_Application`->`_visit_FunctionCall`
    - Ensure that `funcdef` property is a string at syntactic stage
    - Remove references to `sympy.Application` and `sympy.UndefinedFunction`
    - Remove references to `sympy.Function`
- Create `PyccelInternalFunction` class to replace Sympy's Function for numpy/math inheritance
- Move PyccelInternalFunction and PyccelArraySize to new file ast.internals to avoid circular imports (these are needed in lots of places)
- Add `__new__` to Basic to ensure all args and kwargs are stored to avoid sympy singleton bugs. Make `PyccelAstNode` inherit from `Basic`
    - Remove useless `__new__` calls
    - Ensure only hashable types are passed to sympy
    - Use mutable objects in addition to sympy storage where necessary
- Add `_print_LiteralFloat` to `CCodePrinter`

**Codacy Commits**
- Remove duplicate class (`NumpySqrt` and `Sqrt` exist)
- Ensure that new `__init__` methods call parent `__init__`
     - Use overridable functions instead of separate `__init__` to avoid codacy non-parent-init-called in ast.numpyext
- Return fatal `errors.report` to avoid "inconsistent-return-statements"

### DIFF
--- a/pyccel/ast/basic.py
+++ b/pyccel/ast/basic.py
@@ -4,7 +4,10 @@
 #------------------------------------------------------------------------------------------#
 
 """
-This module contains two classes. Basic that provides a python AST and PyccelAstNode which describes each PyccelAstNode
+This module contains classes from which all pyccel nodes inherit.
+They are:
+- Basic, which provides a python AST
+--PyccelAstNode which describes each PyccelAstNode
 """
 
 from sympy.core.basic import Basic as sp_Basic
--- a/pyccel/ast/basic.py
+++ b/pyccel/ast/basic.py
@@ -17,7 +17,9 @@ class Basic(sp_Basic):
     _fst = None
 
     def __new__(cls, *args, **kwargs):
-        return sp_Basic.__new__(cls, *args, *kwargs.values())
+        hashable_args  = [a if not isinstance(a, list) else tuple(a) for a in args]
+        hashable_args += [a if not isinstance(a, list) else tuple(a) for a in kwargs.values()]
+        return sp_Basic.__new__(cls, *hashable_args)
 
     def set_fst(self, fst):
         """Sets the python.ast fst."""
--- a/pyccel/ast/basic.py
+++ b/pyccel/ast/basic.py
@@ -21,7 +21,6 @@ class Basic(sp_Basic):
 
     def __new__(cls, *args, **kwargs):
         hashable_args  = [a if not isinstance(a, list) else tuple(a) for a in args]
-        hashable_args += [a if not isinstance(a, list) else tuple(a) for a in kwargs.values()]
         return sp_Basic.__new__(cls, *hashable_args)
 
     def set_fst(self, fst):
--- a/pyccel/ast/basic.py
+++ b/pyccel/ast/basic.py
@@ -7,7 +7,7 @@
 This module contains classes from which all pyccel nodes inherit.
 They are:
 - Basic, which provides a python AST
---PyccelAstNode which describes each PyccelAstNode
+- PyccelAstNode which describes each PyccelAstNode
 """
 
 from sympy.core.basic import Basic as sp_Basic
@@ -33,6 +33,8 @@ class Basic(sp_Basic):
         return self._fst
 
 class PyccelAstNode(Basic):
+    """Class from which all nodes containing objects inherit
+    """
     stage      = None
     _shape     = None
     _rank      = None
@@ -42,22 +44,31 @@ class PyccelAstNode(Basic):
 
     @property
     def shape(self):
+        """ Tuple containing the length of each dimension
+        of the object """
         return self._shape
 
     @property
     def rank(self):
+        """ Number of dimensions of the object
+        """
         return self._rank
 
     @property
     def dtype(self):
+        """ Datatype of the object """
         return self._dtype
 
     @property
     def precision(self):
+        """ Precision of the datatype of the object """
         return self._precision
 
     @property
     def order(self):
+        """ Indicates whether the data is stored in
+        row-major ('C') or column-major ('F') format.
+        This is only relevant if rank > 1 """
         return self._order
 
     def copy_attributes(self, x):
--- a/pyccel/ast/basic.py
+++ b/pyccel/ast/basic.py
@@ -16,6 +16,9 @@ class Basic(sp_Basic):
     """Basic class for Pyccel AST."""
     _fst = None
 
+    def __new__(cls, *args, **kwargs):
+        return sp_Basic.__new__(cls, *args, *kwargs.values())
+
     def set_fst(self, fst):
         """Sets the python.ast fst."""
         self._fst = fst
@@ -24,7 +27,7 @@ class Basic(sp_Basic):
     def fst(self):
         return self._fst
 
-class PyccelAstNode:
+class PyccelAstNode(Basic):
     stage      = None
     _shape     = None
     _rank      = None
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -1859,21 +1859,37 @@ class For(Basic):
 
         return Basic.__new__(cls, target, iter_obj, body, local_vars)
 
+    def __init__(
+        self,
+        target,
+        iter_obj,
+        body,
+        local_vars = [],
+        strict=True,
+        ):
+        self._target = self._args[0]
+        self._iterable = self._args[1]
+        if strict:
+            self._body = self._args[2]
+        else:
+            self._body = body
+        self._local_vars = self._args[3]
+
     @property
     def target(self):
-        return self._args[0]
+        return self._target
 
     @property
     def iterable(self):
-        return self._args[1]
+        return self._iterable
 
     @property
     def body(self):
-        return self._args[2]
+        return self._body
 
     @property
     def local_vars(self):
-        return self._args[3]
+        return self._local_vars
 
     def insert2body(self, stmt):
         self.body.append(stmt)
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -551,9 +551,6 @@ class DottedName(Basic):
     pyccel.stdlib.parallel
     """
 
-    def __new__(cls, *args):
-        return Basic.__new__(cls, *args)
-
     @property
     def name(self):
         return self._args
@@ -816,9 +813,6 @@ class Allocate(Basic):
     mutable Variable object.
 
     """
-    def __new__(cls, *args, **kwargs):
-
-        return Basic.__new__(cls)
 
     # ...
     def __init__(self, variable, *, shape, order, status):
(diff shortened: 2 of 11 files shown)


EmilyBourne:
@saidctb macro tests are failing as no return type is provided and the function is then used in an assignment. This seems logical to me so I think I should add result types. Do you have any idea how it worked before?

saidctb:
don't we need to remove `Function` from the Ast nodes and not from the syntactic stage, otherwise we will need to remove `Symbol` and `IndexedBase` also to be consistent

EmilyBourne:
> don't we need to remove `Function` from the Ast nodes and not from the syntactic stage, otherwise we will need to remove `Symbol` and `IndexedBase` also to be consistent 

I don't see any reason to keep sympy objects when we don't use any of their properties. I agree that it is not a problem in the syntactic stage but I think it is clearer when an object is modified rather than converted between 3 different objects. `IndexedElement` should be sufficient for the `IndexedBase` case (see #658). If sympy was in less places I might agree with you but at the moment it's just going to make it harder to see what still has a sympy dependence and what doesn't

saidctb (approved):

pyccel/ast/basic.py
```diff
@@ -16,6 +19,11 @@ class Basic(sp_Basic):
     """Basic class for Pyccel AST."""
     _fst = None
 
+    def __new__(cls, *args, **kwargs):
+        hashable_args  = [a if not isinstance(a, list) else tuple(a) for a in args]
+        hashable_args += [a if not isinstance(a, list) else tuple(a) for a in kwargs.values()]
+        return sp_Basic.__new__(cls, *hashable_args)
```
yguclu, line 25: I am not convinced that creating a list of input arguments (ordered) from a dictionary of keyword arguments (not necessarily ordered) is useful. Further, we could get unexpected behavior.

saidctb: ordering a dictionary should not be a problem, because the order doesn't matter from the beginning

pyccel/ast/core.py
```diff
@@ -3573,6 +3564,7 @@ def rename(self, newname):
                              self.body, cls_name=self.cls_name)
 
 
+# TODO: [EB 06.01.2021] Is this class used? What for?
```
yguclu, line 3567: Have you searched for it in the whole project with `grep`?

pyccel/ast/mathext.py
```diff
@@ -4,9 +4,8 @@
 #------------------------------------------------------------------------------------------#
 
 import math
-from sympy import Function
 
-from pyccel.ast.basic     import PyccelAstNode
+from pyccel.ast.core      import PyccelInternalFunction
```
yguclu, line 8: Shouldn't we import this class from `pyccel.ast.internals`?

pyccel/ast/core.py
EmilyBourne: Sorry I meant to go back to this. It seems to be associated with the `@python` decorator but I am not certain what that decorator does. The inheritance from `FunctionDef` probably explains why it only appears in 5 lines excluding the contents of the class

pyccel/ast/basic.py
EmilyBourne: Dictionaries are ordered for python 3.7+, I am not using the `kwargs` passed to `sp_Basic` (unless an class with args only is called using keywords I suppose) so I don't think we could get any unexpected behaviour.

I did this as sympy has internal singleton behaviour. I have not looked into it in detail, but it seems that if the `_args` attribute is the same for 2 instances of a class then they are considered to be the same object. Thus when we modify one, the other changes too which may not be what we wanted.

If anyone has a better solution for this then I would be happy to implement it (@saidctb ?).

Otherwise I can try to just call `sp_Basic.__new__(cls)` and see if anything breaks. It may be that the singleton behaviour only works for certain sympy classes

pyccel/codegen/printing/fcode.py
```diff
@@ -2612,10 +2613,10 @@ def _print_LiteralInteger(self, expr):
 
     def _print_IndexedElement(self, expr):
         base = expr.base
-        if isinstance(base, Application) and not isinstance(base, PythonTuple):
+        if isinstance(base, PyccelInternalFunction) and not isinstance(base, PythonTuple):
```
yguclu, line 2616: Can `base` be an instance of both `PyccelInternalFunction` and `PythonTuple`?

EmilyBourne: Not at the moment, but I think that it will when the `Expr` inheritance is removed

pyccel/parser/syntactic.py
```diff
@@ -917,14 +917,14 @@ def _visit_Call(self, stmt):
...
 
         if isinstance(func, Symbol):
-            f_name = func.name
-            if str(f_name) == "print":
+            f_name = str(func.name)
+            if f_name == "print":
                 func = PythonPrint(PythonTuple(*args))
             else:
-                func = Function(f_name)(*args)
+                func = FunctionCall(f_name, args)
```
yguclu, line 924: Could we not make the `Function.__call__` magic method return a `FunctionCall` object?

pyccel/ast/basic.py
EmilyBourne: > Otherwise I can try to just call sp_Basic.__new__(cls) and see if anything breaks

This does not work for any classes which use `_args` instead of mutable objects

pyccel/parser/syntactic.py
EmilyBourne: It's not possible in the syntactic stage as `Function` is a sympy method but in the semantic stage we can make a `PyccelInternalFunction.__call__` and a `FunctionDef.__call__` return a `FunctionCall`

pyccel/ast/mathext.py
yguclu: ?

pyccel/codegen/printing/fcode.py
yguclu: I cannot imagine a class that inherits from both `PyccelInternalFunction` and `PythonTuple`. Could you give me an example?

pyccel/parser/syntactic.py
yguclu: So in the syntactic stage we create objects of type `sympy.Function`... Are we sure that we still need to do this?

pyccel/ast/core.py
yguclu: @saidctb Do you know what the `@python` decorator was for? Maybe for functions that should not be translated to Fortran/C?

pyccel/codegen/printing/fcode.py
EmilyBourne: `tuple(1,2)`
which is translated to a `PythonTuple` but is a function

pyccel/parser/syntactic.py
EmilyBourne: No. That is what is in the deleted line. `f_name` is a string at this line

pyccel/ast/basic.py
EmilyBourne: It seems that removing line 24 does not break anything. I will remove this line

yguclu: Let me help understand this. We are passing a list of hash-able arguments to `sympy.Basic.__new__` in the hope that Sympy's caching mechanism will not treat them as the same object. But if the arguments are the same we are still in trouble, right?

pyccel/codegen/printing/fcode.py
yguclu: But this is not another class that inherits from both `PyccelInternalFunction` and `PythonTuple`.

The `if` condition above is `True` if `expr` is a `PyccelInternalFunction` **and not** a `PythonTuple`, but the second condition is redundant because always true when the first is true...

Instead, this should just read
```
if isinstance(base, PyccelInternalFunction):
```

yguclu: Unless `PythonTuple` is a subclass of `PyccelInternalFunction` of course... is that the case?

pyccel/parser/syntactic.py
yguclu: If we don't need `sympy.Function` anymore, then why can't we use our own `Function` here, and add it a `__call__` method?

yguclu: Sorry, I just realized that we have `FunctionDef` but not `Function` :sweat_smile:

pyccel/codegen/printing/fcode.py
EmilyBourne: Not at the moment, but I think that it will when the `Expr` inheritance is removed

Currently `PythonTuple` inherits from `PyccelAstNode` and `Expr`. `Expr` handles the `*args`, either we could re-implement that in `PythonTuple` or we could replace the `Expr` inheritance with a `PyccelInternalFunction` inheritance

pyccel/parser/syntactic.py
EmilyBourne: So a class:
```
class Function:
    def __init__(f_name):
        assert(isinstance(f_name, str))
        self._name = f_name
        
    def __call__(self, *args):
        return FunctionCall(self._name, args)
```
which is only used in the syntactic stage? I'm not sure I see what this adds. Presumably it slows things down though

pyccel/ast/basic.py
EmilyBourne: Yes I think so. But this has always been the case so we are not generating bugs here that we did not have before

pyccel/codegen/printing/fcode.py
yguclu: Thanks for explaining

pyccel/ast/basic.py
EmilyBourne: That is why I wanted to also store the kwargs. The more information is stored, the less chance there is of the `_args` matching

pyccel/parser/syntactic.py
yguclu: Yes, it's not useful if only used in the syntactic stage.

pyccel/ast/basic.py
yguclu: OK, let's keep things as they are after the last commit. In any case this confusion will be resolved as soon as we remove all dependencies on Sympy.

pyccel/ast/core.py
yguclu: I suggest we create a dedicated issue where we say that the `@python` decorator should be tested, and then either fix it or remove it.

yguclu (approved):

PR #659, opened 2021-01-06
Move Nil class from ast.core to ast.literals

mostafaMamoni:
According to wikipedia:

> In computer science, a literal is a notation for representing a fixed value in source code.

so moving Nil class to literals.py makes more sense, and it will fix a cyclic import in https://github.com/pyccel/pyccel/pull/651.

### DIFF
--- a/pyccel/ast/literals.py
+++ b/pyccel/ast/literals.py
@@ -223,6 +223,8 @@ class Nil(Basic):
     class for None object in the code.
     """
 
-    def _sympystr(self, printer):
-        sstr = printer.doprint
-        return sstr('None')
+    def __str__(self):
+        """
+        return str ob and 'None' as value
+        """
+        return 'None'
--- a/pyccel/ast/literals.py
+++ b/pyccel/ast/literals.py
@@ -224,7 +224,4 @@ class Nil(Basic):
     """
 
     def __str__(self):
-        """
-        return str ob and 'None' as value
-        """
         return 'None'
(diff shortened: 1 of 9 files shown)


pyccel/ast/literals.py
```diff
@@ -217,3 +217,12 @@ def get_default_literal_value(dtype):
         raise TypeError('Unknown type')
     return value
 
+class Nil(Basic):
+
+    """
+    class for None object in the code.
+    """
+
+    def _sympystr(self, printer):
```
EmilyBourne, line 226: Can this be changed to:
```
def __str__(self):
    return 'None'
```
Or does that break something?

mostafaMamoni: i will check

pyccel/ast/literals.py
```diff
@@ -223,6 +223,8 @@ class Nil(Basic):
...
     """
 
-    def _sympystr(self, printer):
-        sstr = printer.doprint
-        return sstr('None')
\ No newline at end of file
+    def __str__(self):
+        """
+        return str ob and 'None' as value
+        """
```
EmilyBourne, line 229: I don't think magic methods need docstrings. The `a.__str__()` method is just the method that is called when we do `str(a)` so the only docstring needed is the docstring of `str()`

EmilyBourne (approved):

yguclu (approved):

PR #669, opened 2021-01-12
Improve shape comparison in arithmetic operations

### LINKED ISSUE #577: Operands could not be broadcast together with shape

rmahjoubi:
**Improve shape comparison in arithmetic operations**

- Use sympy_to_pyccel to reduce the shape expression to its simplest form which can be compared to another shape.
- In case of uncertainty about shape compatibility; no error is raised and broadcast returns whatever information it can compute.
- Add tests.

Fixes #577

### DIFF
--- a/pyccel/ast/operators.py
+++ b/pyccel/ast/operators.py
@@ -73,20 +73,19 @@ def broadcast(shape_1, shape_2):
 
     new_shape = []
     for e1,e2 in zip(new_shape_1, new_shape_2):
-        if pyccel_to_sympy(e1, {}, set()) == pyccel_to_sympy(e2, {}, set()):
+        sy_e1 = pyccel_to_sympy(e1, {}, set())
+        sy_e2 = pyccel_to_sympy(e2, {}, set())
+        if sy_e1 == sy_e2:
             new_shape.append(e1)
-        elif pyccel_to_sympy(e1, {}, set()) == 1:
+        elif sy_e1 == 1:
             new_shape.append(e2)
-        elif pyccel_to_sympy(e2, {}, set()) == 1:
+        elif sy_e2 == 1:
             new_shape.append(e1)
-        elif pyccel_to_sympy(e1, {}, set()).is_constant() and not\
-                pyccel_to_sympy(e2, {}, set()).is_constant():
+        elif sy_e1.is_constant() and not sy_e2.is_constant():
             new_shape.append(e1)
-        elif pyccel_to_sympy(e2, {}, set()).is_constant() and not\
-                pyccel_to_sympy(e1, {}, set()).is_constant():
+        elif sy_e2.is_constant() and not sy_e1.is_constant():
             new_shape.append(e2)
-        elif not pyccel_to_sympy(e2, {}, set()).is_constant() and not\
-                pyccel_to_sympy(e1, {}, set()).is_constant():
+        elif not sy_e2.is_constant() and not sy_e1.is_constant():
             new_shape.append(e1)
         elif isinstance(e1, PyccelArraySize) and isinstance(e2, PyccelArraySize):
             new_shape.append(e1)
--- a/pyccel/ast/operators.py
+++ b/pyccel/ast/operators.py
@@ -112,6 +112,9 @@ class PyccelOperator(Expr, PyccelAstNode):
             return
         self._set_dtype()
         self._set_shape_rank()
+        # rank is None for lambda functions
+        if self._rank is not None and self._rank > 1:
+            self._set_order()
 
     @property
     def precedence(self):
@@ -159,6 +162,16 @@ class PyccelOperator(Expr, PyccelAstNode):
 
     def __str__(self):
         return repr(self)
+
+    def _set_order(self):
+        """ Sets the shape and rank
+        This is chosen to match the arguments if they are in agreement.
+        Otherwise it defaults to 'C'
+        """
+        if all(a.order == self._args[0].order for a in self._args):
+            self._order = self._args[0].order
+        else:
+            self._order = 'C'
 
 #==============================================================================
 
--- a/pyccel/ast/operators.py
+++ b/pyccel/ast/operators.py
@@ -75,8 +75,10 @@ def broadcast(shape_1, shape_2):
 
     new_shape = []
     for e1,e2 in zip(new_shape_1, new_shape_2):
-        sy_e1 = pyccel_to_sympy(e1, {}, set())
-        sy_e2 = pyccel_to_sympy(e2, {}, set())
+        used_names = set()
+        symbol_map = {}
+        sy_e1 = pyccel_to_sympy(e1, symbol_map, used_names)
+        sy_e2 = pyccel_to_sympy(e2, symbol_map, used_names)
         if sy_e1 == sy_e2:
             new_shape.append(e1)
         elif sy_e1 == 1:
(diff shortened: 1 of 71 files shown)


pyccel/ast/operators.py
```diff
@@ -70,11 +73,20 @@ def broadcast(shape_1, shape_2):
...
     for e1,e2 in zip(new_shape_1, new_shape_2):
-        if e1 == e2:
+        if pyccel_to_sympy(e1, {}, set()) == pyccel_to_sympy(e2, {}, set()):
+            new_shape.append(e1)
+        elif pyccel_to_sympy(e1, {}, set()) == 1:
+            new_shape.append(e2)
+        elif pyccel_to_sympy(e2, {}, set()) == 1:
+            new_shape.append(e1)
+        elif pyccel_to_sympy(e1, {}, set()).is_constant() and not\
+                pyccel_to_sympy(e2, {}, set()).is_constant():
```
MenoIy, line 83: can you just create two variables to store the result of pyccel_to_sympy? and use those two variables in the conditions

pyccel/ast/sympy_helper.py
```diff
@@ -128,6 +128,12 @@ def pyccel_to_sympy(expr, symbol_map, used_names):
     elif isinstance(expr, LiteralFloat):
         return sp.Float(expr)
 
+    if isinstance(expr, int):
+        return sp.Integer(expr)
+
+    elif isinstance(expr, float):
+        return sp.Float(expr)
```
MenoIy, line 135: do you need this conversion I thought that all const values in the code are instances of Literal?

rmahjoubi: some cases they are instances of python integer

MenoIy (approved):
can you please

pyccel/ast/sympy_helper.py
EmilyBourne: Do you remember which cases had integers? Probably it is there where there is a problem rather than here. We need to try and eliminate those bugs so we print with correct precision

EmilyBourne (commented):
Maybe you could add one or two 2d tests? Plus a test where incompatible shapes are added to ensure the correct error is raised?

pyccel/ast/operators.py
```diff
@@ -59,25 +59,35 @@
...
         new_shape_2 = shape_2
     else:
         new_shape_2 = shape_2
         new_shape_1 = shape_1
 
     new_shape = []
     for e1,e2 in zip(new_shape_1, new_shape_2):
-        if e1 == e2:
+        sy_e1 = pyccel_to_sympy(e1, {}, set())
+        sy_e2 = pyccel_to_sympy(e2, {}, set())
```
EmilyBourne, line 79: I think it would be a good idea to pass the same `symbol_map` and `used_names` set to these 2 calls. That way we should be able to raise an error if we do:
```
def f(k):
    a = np.ones(k)
    b = np.ones(np.shape(a)[0]+2)
    c = a + b
```
Currently we get:
`sy_e1 = tmp_size_0001`
and
`sy_e2 = tmp_size_0001`
As the operator is not conserved we cannot raise an error currently, but this example still demonstrates the danger of not using the same dictionary and set

EmilyBourne: Once this is implemented it would probably be worth modifying [these lines](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/variable.py#L220-L225) to accept any `PyccelAstNode`. The simplification of the shape is probably no longer necessary (so we could get rid of `alloc_shape`), although the `Shape` function should also be modified to always return a PyccelArrayShape rather than the element of the shape. All of these corrections are probably beyond the scope of this PR. But if you make minor changes locally (e.g. add `Variable` and `PyccelOperator` to the tuple at [this line](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/variable.py#L220) ) then you can test your implementation in a more strenuous setting. E.g. for the example above I get:
`sy_e1 = k`
`sy_e2 = k + 2`
which your current implementation considers is acceptable

EmilyBourne: I just realised there's a much simpler example that doesn't work currently and should already:
```
import numpy as np

def f(k : int):
    a = np.ones(k)
    c = a[1:] + a[2:]
```

EmilyBourne: This sympy page may be helpful for working out if you have a system of equations with no solution: https://docs.sympy.org/latest/modules/solvers/solvers.html

pyccel/ast/operators.py
```diff
@@ -59,25 +59,38 @@
...
             new_shape.append(e1)
-        elif e1 == 1:
+        elif sy_e2.is_constant() and not sy_e1.is_constant():
             new_shape.append(e2)
-        elif e2 == 1:
+        elif not sy_e2.is_constant() and not sy_e1.is_constant()\
+                and not (sy_e1 - sy_e2).is_constant():
             new_shape.append(e1)
         elif isinstance(e1, PyccelArraySize) and isinstance(e2, PyccelArraySize):
             new_shape.append(e1)
```
EmilyBourne, line 96: Are [these lines](https://github.com/pyccel/pyccel/pull/669/files#diff-138e903e490ad8bdb3f67bbb3d83aedcc36b21d41a56b594895f62adccf84debR95-R100) still needed?

EmilyBourne: [these lines](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/operators.py#L82-L87)

rmahjoubi: yes they are not used anymore

EmilyBourne (approved):

yguclu (approved):
Very nice!

PR #671, opened 2021-01-12
Ndarray starter guide

saniQQ:
- document the rules that a user needs to know to work properly with the ndarrays with some given examples.
- examples of indexing and slicing a ndarray with the generated code.

### DIFF
--- a/tutorial/ndarrays.md
+++ b/tutorial/ndarrays.md
@@ -4,7 +4,7 @@
 
 A ndarray is a fixed-size multi-dimensional container of items of the same type and size. The number of dimensions and items in an array is defined by its shape, which is a container of N non-negative integers that specify the sizes of each dimension. The type of items in the array is specified by a separate data-type object, one of which is associated with each ndarray.
 
-Different ndarrays can share the same data, so that changes made in one ndarray may be visible in another. that is, a ndarray can be a "view" to another ndarray, and the data it is referring to is taken care of by the "base" ndarray.
+Different ndarrays can share the same data, so that changes made in one ndarray may be visible in another. That is, a ndarray can be a "view" to another ndarray, and the data it is referring to is taken care of by the "base" ndarray.
 [read more](https://numpy.org/doc/stable/reference/arrays.html)
 
 ## Pyccel ndarrays ##
@@ -39,7 +39,7 @@ pyccel:
 
 Pyccel makes a difference between ndarrays that own their data and the ones they don't.
 
-Pyccel call it own garbage collecting when needed but has a set of rules to do so:
+Pyccel calls its own garbage collector when needed, but has a set of rules to do so:
 
 -   Can not reassign ndarrays with different ranks.
 
@@ -59,7 +59,7 @@ Pyccel call it own garbage collecting when needed but has a set of rules to do s
      |error [semantic]: ex.py [4]| Incompatible redefinition (|a| real(10, 20) <-> real(10,))
     ```
 
-    This limitation is due to the way Fortran alloctable can't change the rank after declaration.
+    This limitation is due to the way Fortran allocatable can't change the rank after declaration.
 
 -   Can not assign ndarrays that own their data one another.
 
@@ -164,7 +164,7 @@ Pyccel call it own garbage collecting when needed but has a set of rules to do s
 
 ### Slicing and indexing ###
 
-the indexing and slicing in Pyccel handles only the basic indexing of [numpy arrays](https://numpy.org/doc/stable/user/basics.indexing.html).
+The indexing and slicing in Pyccel handles only the basic indexing of [numpy arrays](https://numpy.org/doc/stable/user/basics.indexing.html).
 
 Some examples:
 
--- a/tutorial/ndarrays.md
+++ b/tutorial/ndarrays.md
@@ -141,7 +141,7 @@ Pyccel calls its own garbage collector when needed, but has a set of rules to do
         end program prog_ex
         ```
 
--   Can not reassign to a ndarray that has another pointer accessing his data.
+-   Can not reassign to a ndarray that has another pointer accessing its data.
 
     ```Python
     import numpy as np
@@ -160,7 +160,7 @@ Pyccel calls its own garbage collector when needed, but has a set of rules to do
       |error [semantic]: ex.py [6]| Attempt to reallocate an array which is being used by another variable (a)
     ```
 
-    This limitation is set since we need to free the previous data when are trying to reallocate the ndarray, which in this case will cause the data where the view **b** point to became inaccessible.
+This limitation is set since we need to free the previous data when we reallocate the ndarray. In this case, this will cause the data pointed to by the view **b** to became inaccessible.
 
 ### Slicing and indexing ###
 
--- a/tutorial/ndarrays.md
+++ b/tutorial/ndarrays.md
@@ -43,23 +43,23 @@ Pyccel call it own garbage collecting when needed but has a set of rules to do s
 
 - Can not reassign ndarrays with different ranks.
 
-    ```Python
-    import numpy as np
+  ```Python
+  import numpy as np
 
-    a = np.ones((10, 20))
-    #(some code...)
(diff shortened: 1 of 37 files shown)


tutorial/ndarrays.md
```diff
@@ -0,0 +1,328 @@
...
+ERROR at annotation (semantic) stage
+pyccel:
+ |error [semantic]: ex.py [5]| Incompatible types in assignment (|a| real <-> int)
+```
+
+### Memory management ###
+
+Pyccel makes a difference between ndarrays that own their data and the ones they don't.
+
+Pyccel call it own garbage collecting when needed but has a set of rules to do so:
```
MenoIy, line 42: grammar: Pyccel calls its own garbage collector when needed,

tutorial/ndarrays.md
```diff
@@ -0,0 +1,328 @@
...
+
+    _OUTPUT_ :
+
+    ```Shell
+    ERROR at annotation (semantic) stage
+    pyccel:
+     |error [semantic]: ex.py [4]| Incompatible redefinition (|a| real(10, 20) <-> real(10,))
+    ```
+
+    This limitation is due to the way Fortran alloctable can't change the rank after declaration.
```
MenoIy, line 62: typo: allocatable

tutorial/ndarrays.md
```diff
@@ -0,0 +1,328 @@
...
+        end if
+        else
+        allocate(a(0:10_C_INT64_T - 1_C_INT64_T))
+        end if
+        a = 1.0_C_DOUBLE
+
+        end program prog_ex
+        ```
+
+-   Can not reassign to a ndarray that has another pointer accessing his data.
```
MenoIy, line 144: is that mean?
```
a = new
b = a
a = new
```

tutorial/ndarrays.md
```diff
@@ -0,0 +1,328 @@
...
+    ERROR at annotation (semantic) stage
+    pyccel:
+      |error [semantic]: ex.py [6]| Attempt to reallocate an array which is being used by another variable (a)
+    ```
+
+    This limitation is set since we need to free the previous data when are trying to reallocate the ndarray, which in this case will cause the data where the view **b** point to became inaccessible.
+
+### Slicing and indexing ###
+
+the indexing and slicing in Pyccel handles only the basic indexing of [numpy arrays](https://numpy.org/doc/stable/user/basics.indexing.html).
```
MenoIy, line 167: typo: The

MenoIy (changes requested):

tutorial/ndarrays.md
saniQQ: yeah, that's what I meant.

tutorial/ndarrays.md
```diff
@@ -0,0 +1,328 @@
...
+
+    ```Shell
+    ERROR at annotation (semantic) stage
+    pyccel:
+     |error [semantic]: ex.py [5]| Arrays which own their data cannot become views on other arrays (a)
+    ```
+
+    This limitation is due to the fact that the ndarray **a** will have to go from a data owner to a pointer to the **b** ndarray data.
+
+    _NOTE_: this limitation does not include reassigning using new data with respecting the previous rule.
```
EmilyBourne, line 84: I don't understand what you mean here

tutorial/ndarrays.md
```diff
@@ -0,0 +1,328 @@
...
+        end if
+        else
+        allocate(a(0:10_C_INT64_T - 1_C_INT64_T))
+        end if
+        a = 1.0_C_DOUBLE
+
+        end program prog_ex
+        ```
+
+-   Can not reassign to a ndarray that has another pointer accessing his data.
```
EmilyBourne, line 144: `his`->`its`
An array does not have a gender ;)

tutorial/ndarrays.md
```diff
@@ -0,0 +1,328 @@
...
+
+    _OUTPUT_ :
+
+    ```Shell
+    ERROR at annotation (semantic) stage
+    pyccel:
+      |error [semantic]: ex.py [6]| Attempt to reallocate an array which is being used by another variable (a)
+    ```
+
+    This limitation is set since we need to free the previous data when are trying to reallocate the ndarray, which in this case will cause the data where the view **b** point to became inaccessible.
```
EmilyBourne, line 163: English improvement:
```
This limitation is set since we need to free the previous data when we reallocate the ndarray. In this case, this will cause the data pointed to by the view **b** to became inaccessible
```

tutorial/ndarrays.md
saniQQ: i tried to say that the limitation of reassigning a ndarray variable is not in place when the right-hand side is a newly created array (reallocation).

EmilyBourne: Ah, you mean:
```
This limitation only applies to data that is shared between arrays, not to assignments which reserve a new memory block
```
?

MenoIy (approved):

tutorial/ndarrays.md
```diff
@@ -0,0 +1,328 @@
...
+Pyccel calls its own garbage collector when needed, but has a set of rules to do so:
+
+-   Can not reassign ndarrays with different ranks.
+
+    ```Python
+    import numpy as np
+
+    a = np.ones((10, 20))
+    #(some code...)
+     a = np.ones(10)
```
rakati, line 51: please, can you remove the space from the beginning of the line, so the user can just copy the example and work without error?

tutorial/ndarrays.md
EmilyBourne: That sounds clearer :+1:

EmilyBourne (approved):

tutorial/ndarrays.md
```diff
@@ -0,0 +1,328 @@
...
+
+```Shell
+ERROR at annotation (semantic) stage
+pyccel:
+ |error [semantic]: ex.py [5]| Incompatible types in assignment (|a| real <-> int)
+```
+
+### Memory management ###
+
+Pyccel makes a difference between ndarrays that own their data and the ones they don't.
```
yguclu, line 40: Typo: they -> that

tutorial/ndarrays.md
```diff
@@ -0,0 +1,328 @@
...
+
+    _OUTPUT_ :
+
+    ```Shell
+    ERROR at annotation (semantic) stage
+    pyccel:
+     |error [semantic]: ex.py [4]| Incompatible redefinition (|a| real(10, 20) <-> real(10,))
+    ```
+
+    This limitation is due to the way Fortran allocatable can't change the rank after declaration.
```
yguclu, line 62: The construction of this sentence sounds a bit strange when I read it.
@EmilyBourne Could you please double-check?

tutorial/ndarrays.md
```diff
@@ -0,0 +1,328 @@
...
+        {
+            t_ndarray a;
+            a = array_create(1, (int32_t[]){20}, nd_double);
+            array_fill((double)1.0, a);
+            /*(some code...)*/
+            if (a.shape != NULL)
+            {
+                free_array(a); //this line handles the redefinition of the array
+            }
+            a = array_create(1, (int32_t[]){10}, nd_double);
```
yguclu, line 110: Shouldn't we reallocate the array only if its shape changes? Here we only check that `a.shape != NULL`

yguclu (changes requested):

tutorial/ndarrays.md
EmilyBourne: You are right. `allocatable` should definitely be plural, but maybe this would be better:
```
This limitation is due to the fact that the rank of Fortran allocatable objects must be specified in their declaration.
```

tutorial/ndarrays.md
saniQQ: yes, we are reallocating the array only when the shape is changed the `if` condition for the shape is just a safeguard for C.
**it can be removed.**

yguclu (approved):

PR #672, opened 2021-01-13
import ndarrays.h when declaring an array as an argument

mostafaMamoni:
add ndarrays to additional imports if the type is ndarray


mostafaMamoni:
This bug will be solved when #632 get merged.

PR #673, opened 2021-01-13
Add print_kind method to Fortran printer

mostafaMamoni:
In order to avoid printing the kind(precision) with iso_c_binding for readability and make it easier, a printer of the kind has been implemented.

### DIFF
--- a/pyccel/codegen/printing/fcode.py
+++ b/pyccel/codegen/printing/fcode.py
@@ -599,8 +599,6 @@ class FCodePrinter(CodePrinter):
 #
 #        return self._get_statement(code)
 
-    def print_kind(self, expr):
-        return iso_c_binding[self._print(expr.dtype)][expr.precision]
 
     def _print_SumFunction(self, expr):
         return str(expr)
@@ -630,6 +628,16 @@ class FCodePrinter(CodePrinter):
     def _print_PythonImag(self, expr):
         value = self._print(expr.internal_var)
         return 'aimag({0})'.format(value)
+
+    #========================== Numpy Elements ===============================#
+
+    def print_kind(self, expr):
+        """
+        Prints the kind(precision) of a literal value
+        """
+        return iso_c_binding[self._print(expr.dtype)][expr.precision]
+
+    #=========================================================================#
 
     #========================== Numpy Elements ===============================#
 
--- a/pyccel/codegen/printing/fcode.py
+++ b/pyccel/codegen/printing/fcode.py
@@ -599,13 +599,16 @@ class FCodePrinter(CodePrinter):
 #
 #        return self._get_statement(code)
 
+    def print_kind(self, dtype, precision):
+        return ""+iso_c_binding[self._print(dtype)][precision]
+
     def _print_SumFunction(self, expr):
         return str(expr)
 
     def _print_PythonLen(self, expr):
         var = expr.arg
         idx = 1 if var.order == 'F' else var.rank
-        prec = iso_c_binding["integer"][expr.precision]
+        prec = self.print_kind(expr.dtype, expr.precsiion)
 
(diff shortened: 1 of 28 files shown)


saidctb (approved):

nhamidn (approved):

EmilyBourne:
This seems a little neater than before, but the original reason for doing this was to avoid the developer having to remember that precision is a special case. This solution doesn't fix that problem. Is there a reason this PR is needed if we still plan to fix #587 at some point?

mostafaMamoni:
> This seems a little neater than before, but the original reason for doing this was to avoid the developer having to remember that precision is a special case. This solution doesn't fix that problem. Is there a reason this PR is needed if we still plan to fix #587 at some point?

I thought that our starting point is making the printing of the precision easier without using ISO_C_BINDING, this PR solved that problem and we can make it even easier for us by just passing the `expr` instead of `dtype` and `precision` something like `self.print_kind(expr)` or `self.print_precision(expr)` if you like "precision", and i think print_precision/kind as you said neater than before and not that hard to remember especially if we implemented the second proposition.

EmilyBourne:
> > This seems a little neater than before, but the original reason for doing this was to avoid the developer having to remember that precision is a special case. This solution doesn't fix that problem. Is there a reason this PR is needed if we still plan to fix #587 at some point?
> 
> I thought that our starting point is making the printing of the precision easier without using ISO_C_BINDING, this PR solved that problem and we can make it even easier for us by just passing the `expr` instead of `dtype` and `precision` something like `self.print_kind(expr)` or `self.print_precision(expr)` if you like "precision", and i think print_precision/kind as you said neater than before and not that hard to remember especially if we implemented the second proposition.

I agree that `self.print_kind(expr)` would be even better

pyccel/codegen/printing/fcode.py
```diff
@@ -599,13 +599,16 @@ def _print_Lambda(self, expr):
 #
 #        return self._get_statement(code)
 
+    def print_kind(self, expr):
+        return ""+iso_c_binding[self._print(expr.dtype)][expr.precision]
```
EmilyBourne, line 603: Why do we need the `""+`?

mostafaMamoni: we dont need it, since we get a string from ISO_C_BINDING.

EmilyBourne (approved):

pyccel/codegen/printing/fcode.py
```diff
@@ -599,13 +599,16 @@ def _print_Lambda(self, expr):
 #
 #        return self._get_statement(code)
 
+    def print_kind(self, expr):
+        return iso_c_binding[self._print(expr.dtype)][expr.precision]
+
```
yguclu, line 604: Since this function is fundamentally different than the other `_print_TYPE` methods, I think it should be moved to another location.

yguclu (changes requested):

pyccel/codegen/printing/fcode.py
yguclu: And it needs a docstring!

mostafaMamoni: what do you mean by "location" is it in another file or in the same file?.

yguclu: I mean in the same file.
This function cannot be hidden among other `_print_TYPE` methods.
It should be one of the first methods of its class instead.

yguclu (changes requested):
Since the `print_kind` function is fundamentally different than the other `_print_TYPE` methods, it should not stay hidden among them. It should be one of the first methods of the printer class instead, so please change its location accordingly.

yguclu (approved):

PR #674, opened 2021-01-14
Describe usage of Pyccel container image

yguclu:
This is a modified version of PR #654:

Describe usage of Pyccel container image usage based on https://github.com/noushi/pyccel-docker generated images.
As of now, images are available on Docker Hub (https://hub.docker.com/r/pyccel/pyccel/) and on the GitHub Container Registry (ghcr.io, private at the moment).

Co-authored-by: noushi

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -24,10 +24,20 @@ Pyccel comes with a selection of **extensions** allowing you to convert calls to
 - mpi4py
 - h5py (not available yet)
 
+Pyccel Installation Methods
+***************************
+
+Pyccel can be installed on virtually any machine that provides Python 3, the pip package manager, a C/Fortran compiler, and an Internet connection.
+Some advanced features of Pyccel require additional non-Python libraries to be installed, for which we provide detailed instructions below.
+
+Alternatively, Pyccel can be deployed through a **Linux Docker image** that contains all dependencies, and which can be setup with any version of Pyccel.
+For more information, please read the section on `Pyccel container images`_.
+
+
 Requirements
 ============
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -224,13 +236,38 @@ Then build the documentation with::
...
 
 To test your Pyccel installation please run the script *tests/run_tests_py3.sh* (Unix), or *tests/run_tests.bat* (Windows).
 
 Continuous testing runs on Travis CI: <https://travis-ci.com/github/pyccel/pyccel>
 
+
+Pyccel Container Images
+=======================
+
+Pyccel container images are available through both Docker Hub (docker.io) and GitHub Container Registry (ghcr.io).
```
EmilyBourne, line 251: the GitHub Container Registry

yguclu: This was copied from the other PR... Is "the" missing here?

EmilyBourne: Yes, for completely correct English we should have a "the" here

yguclu: And why don't we need "the" before Docker Hub?

EmilyBourne: Hmmm, maybe this is why I didn't signal it before. Registry needs "The" as there can be multiple different registries. There could also be more than one Docker Hub, but it's short enough to sound like a proper name. It might be right as it is but it sounds a little odd to me. Or we could put "the" before both

yguclu: OK, I've added "the" before "GitHub Container Registry" as you suggested.

EmilyBourne (approved):

PR #675, opened 2021-01-14
Bugfix: create view from another view

### LINKED ISSUE #636: pointer to a view to an array is allocatable
**Describe the bug**
the attribute allocatable is True when it should be false
when assigning a view to another variable

**To Reproduce**
```python
import numpy as np
a = np.array([[1, 2, 3, 4, 5], [6, 7, 8, 9, 10]])
c = a[1:]
b = c[1:]
b = a
```

```bash/fortran/c
Arrays which own their data cannot become views on other arrays (b)
```

**Expected behavior**
we should not raise an error here as b is not an allocatable

saniQQ:
Fix #636:

- Fix bug where view created from another view has wrong values of attributes `is_allocatable` & `is_pointer`;
- Fix bug where `is_target` attribute is set to true when target is pointer.

### DIFF
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -1370,7 +1370,7 @@ class SemanticParser(BasicParser):
 
             # TODO uncomment this line, to make rhs target for
             #      lists/tuples.
-            rhs.base.is_target = False if rhs.base.is_pointer else True
+            rhs.base.is_target = not rhs.base.is_pointer
 
     def _assign_lhs_variable(self, lhs, d_var, rhs, new_expressions, is_augassign, **settings):
         """
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -1364,7 +1364,7 @@ class SemanticParser(BasicParser):
             # TODO uncomment this line, to make rhs target for
             #      lists/tuples.
             rhs.is_target = True
-        if isinstance(rhs, IndexedElement) and (rhs.rank > 0 and (rhs.base.allocatable or rhs.base.is_pointer)):
+        if isinstance(rhs, IndexedElement) and rhs.rank > 0 and (rhs.base.allocatable or rhs.base.is_pointer):
             d_lhs['allocatable'] = False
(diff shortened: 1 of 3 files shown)


mostafaMamoni (approved):

pyccel/parser/semantic.py
```diff
@@ -1364,13 +1364,13 @@ def _ensure_target(self, rhs, d_lhs):
...
             rhs.is_target = True
-        if isinstance(rhs, IndexedElement) and rhs.rank > 0 and rhs.base.allocatable:
+        if isinstance(rhs, IndexedElement) and (rhs.rank > 0 and (rhs.base.allocatable or rhs.base.is_pointer)):
             d_lhs['allocatable'] = False
             d_lhs['is_pointer' ] = True
 
             # TODO uncomment this line, to make rhs target for
             #      lists/tuples.
-            rhs.base.is_target = True
+            rhs.base.is_target = True if rhs.base.allocatable else False
```
EmilyBourne, line 1373: What about stack arrays?

saniQQ: i think i will change it to be True if rhs.base.is_pointer is false

pyccel/parser/semantic.py
```diff
@@ -1364,13 +1364,13 @@ def _ensure_target(self, rhs, d_lhs):
             # TODO uncomment this line, to make rhs target for
             #      lists/tuples.
             rhs.is_target = True
-        if isinstance(rhs, IndexedElement) and rhs.rank > 0 and rhs.base.allocatable:
+        if isinstance(rhs, IndexedElement) and (rhs.rank > 0 and (rhs.base.allocatable or rhs.base.is_pointer)):
```
MenoIy, line 1367: extra parentheses

MenoIy (changes requested):
can you add some tests ?

MenoIy (approved):

EmilyBourne (approved):

yguclu (approved):

PR #676, opened 2021-01-14
Remove sympy.Tuple from pyccel's AST representation

### LINKED ISSUE #657: Remove sympy.Tuple from Pyccel
`sympy.Tuple` does not always symplify it's arguments, however the `sympify=False` argument is easily forgotten which leads to sympy objects replacing pyccel objects. `tuple` is sufficient in pyccel

EmilyBourne:
This PR removes `sympy.Tuple` from Pyccel's AST representation, because it is not necessary. Where it is used to group items a python `tuple` is sufficient. Where it may represent tuples in the translated code `PythonTuple` should be used instead. Unless `sympy.Tuple` is created with `sympify = False` it also converts its arguments to sympy objects, making it a common source of errors. This fixes #657.

### DIFF
--- a/pyccel/ast/builtins.py
+++ b/pyccel/ast/builtins.py
@@ -11,7 +11,7 @@ In this module we implement some of them in alphabetical order.
 
 """
 
-from sympy import Symbol, Tuple
+from sympy import Symbol
 from sympy import Expr, Not
 from sympy import sympify
 from sympy.tensor import Indexed, IndexedBase
@@ -402,7 +402,7 @@ class PythonLen(PyccelInternalFunction):
         return self._args[0]
 
 #==============================================================================
-class PythonList(Tuple, PyccelAstNode):
+class PythonList(PythonTuple, PyccelAstNode):
     _order = 'C'
     _is_homogeneous = True
     """ Represent lists in the code with dynamic memory management."""
@@ -507,7 +507,7 @@ class PythonRange(Basic):
 
         _valid_args = (LiteralInteger, Symbol, Indexed)
 
-        if isinstance(args, (tuple, list, Tuple)):
+        if isinstance(args, (tuple, list)):
             if len(args) == 1:
                 stop = args[0]
             elif len(args) == 2:
@@ -548,7 +548,7 @@ class PythonZip(Basic):
     """
 
     def __new__(cls, *args):
-        if not isinstance(args, (tuple, list, Tuple)):
+        if not isinstance(args, (tuple, list)):
             raise TypeError('args must be an iterable')
         elif len(args) < 2:
             raise ValueError('args must be of length > 2')
@@ -580,11 +580,11 @@ class PythonAbs(PyccelInternalFunction):
 class PythonSum(PyccelInternalFunction):
     """Represents a call to  python sum for code generation.
 
-    arg : list , tuple , PythonTuple, Tuple, List, Variable
+    arg : list , tuple , PythonTuple, List, Variable
     """
 
     def __init__(self, arg):
-        if not isinstance(arg, (list, tuple, PythonTuple, Tuple, PythonList,
+        if not isinstance(arg, (list, tuple, PythonTuple, PythonList,
                                 Variable, Expr)): # pylint: disable=undefined-variable
             raise TypeError('Unknown type of  %s.' % type(arg))
         self._dtype = arg.dtype
@@ -601,11 +601,11 @@ class PythonSum(PyccelInternalFunction):
 class PythonMax(PyccelInternalFunction):
     """Represents a call to  python max for code generation.
 
-    arg : list , tuple , PythonTuple, Tuple, List
+    arg : list , tuple , PythonTuple, List
     """
 
     def __init__(self, x):
-        if not isinstance(x, (list, tuple, PythonTuple, Tuple, PythonList)):
+        if not isinstance(x, (list, tuple, PythonTuple, PythonList)):
             raise TypeError('Unknown type of  %s.' % type(x))
         self._shape     = ()
         self._rank      = 0
@@ -618,7 +618,7 @@ class PythonMax(PyccelInternalFunction):
 class PythonMin(PyccelInternalFunction):
     """Represents a call to  python min for code generation.
 
-    arg : list , tuple , PythonTuple, Tuple, List, Variable
+    arg : list , tuple , PythonTuple, List, Variable
     """
     def __init__(self, x):
         self._shape     = ()
--- a/pyccel/ast/builtins.py
+++ b/pyccel/ast/builtins.py
@@ -302,7 +302,7 @@ class PythonInt(Expr, PyccelAstNode):
         return self._args[0]
 
 #==============================================================================
-class PythonTuple(Expr, PyccelAstNode):
+class PythonTuple(PyccelAstNode):
     """ Represents a call to Python's native tuple() function.
     """
     _iterable        = True
@@ -402,7 +402,7 @@ class PythonLen(PyccelInternalFunction):
         return self._args[0]
 
 #==============================================================================
-class PythonList(PythonTuple, PyccelAstNode):
+class PythonList(PythonTuple):
     _order = 'C'
     _is_homogeneous = True
     """ Represent lists in the code with dynamic memory management."""
--- a/pyccel/ast/builtins.py
+++ b/pyccel/ast/builtins.py
@@ -547,7 +547,7 @@ class PythonZip(Basic):
 
     def __new__(cls, *args):
         if not isinstance(args, (tuple, list)):
-            raise TypeError('args must be an iterable')
+            raise TypeError('args must be a list or tuple')
         elif len(args) < 2:
             raise ValueError('args must be of length > 2')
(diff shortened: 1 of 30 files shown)


pyccel/ast/builtins.py
```diff
@@ -402,7 +402,7 @@ def arg(self):
         return self._args[0]
 
 #==============================================================================
-class PythonList(Tuple, PyccelAstNode):
+class PythonList(PythonTuple, PyccelAstNode):
```
yguclu, line 405: Is `PythonTuple` not a subclass of `PyccelAstNode`?

EmilyBourne: You're right

pyccel/ast/core.py
```diff
@@ -2783,7 +2782,7 @@ def __init__(self, func, args, current_function=None):
         if str(current_function) == str(name):
             func.set_recursive()
 
-        if not isinstance(args, (tuple, list, Tuple)):
+        if not isinstance(args, (tuple, list)):
             raise TypeError('> expecting an iterable')
```
yguclu, line 2786: Could you please change this error message?
A general iterable object is not accepted here, it must be either `list` or `tuple`.

pyccel/ast/core.py
```diff
@@ -3084,7 +3083,7 @@ def __init__(
             raise TypeError('Expecting a boolean for header')
 
         if arguments_inout:
-            if not isinstance(arguments_inout, (list, tuple, Tuple)):
+            if not isinstance(arguments_inout, (list, tuple)):
                 raise TypeError('Expecting an iterable ')
```
yguclu, line 3087: Same comment as above

pyccel/ast/core.py
```diff
@@ -4081,27 +4080,25 @@ def __new__(
         as_lambda=False,
         nargs=1,
         ):
-        if not isinstance(module, (str, DottedName, list, tuple,
-                          Tuple)):
+        if not isinstance(module, (str, DottedName, list, tuple)):
             raise TypeError('Expecting a string or DottedName, given {0}'.format(type(module)))
```
yguclu, line 4084: This error message is also incomplete. Can we fix it in this PR?

pyccel/ast/core.py
```diff
@@ -4081,27 +4080,25 @@ def __new__(
...
 
-        if isinstance(module, (list, tuple, Tuple)):
+        if isinstance(module, (list, tuple)):
             module = DottedName(*module)
 
         if funcs:
-            if not isinstance(funcs, (str, DottedName, list, tuple,
-                              Tuple)):
+            if not isinstance(funcs, (str, DottedName, list, tuple)):
                 raise TypeError('Expecting a string or DottedName')
```
yguclu, line 4096: Same comment as above

pyccel/ast/core.py
```diff
@@ -4378,7 +4375,7 @@ def __new__(
         iterator,
         stmts=None,
         ):
-        if not isinstance(iterator, (tuple, Tuple)):
+        if not isinstance(iterator, tuple):
             raise TypeError('iterator must be a tuple')
         if not len(iterator) == 3:
             raise ValueError('iterator must be of lenght 3')
```
yguclu, line 4381: Typo: `lenght` -> `length`

pyccel/ast/builtins.py
```diff
@@ -548,7 +548,7 @@ class PythonZip(Basic):
     """
 
     def __new__(cls, *args):
-        if not isinstance(args, (tuple, list, Tuple)):
+        if not isinstance(args, (tuple, list)):
             raise TypeError('args must be an iterable')
```
yguclu, line 552: I've just noticed that this error message is also wrong...

pyccel/ast/core.py
```diff
@@ -11,7 +11,7 @@
 from sympy import sympify
 from sympy import Add as sp_Add, Mul as sp_Mul, Pow as sp_Pow
 from sympy import Eq as sp_Eq, Ne as sp_Ne, Lt as sp_Lt, Le as sp_Le, Gt as sp_Gt, Ge as sp_Ge
-from sympy import Integral, Symbol, Tuple
+from sympy import Integral, Symbol
```
yguclu, line 14: I know that this is not strictly related to this PR, but did you know that we were importing `Integral` from `sympy`?? Apparently it only appears in line 286 of this file, where `Derivative` is also used (which is imported on line 24)...

@saidctb Do you know anything about this? Is it used by Psydac?

saidctb: it is used in `allocatable_like`  but I don't think it's used anywhere we can delete it

EmilyBourne: This comment is about issue #663. That issue is for all the extra sympy objects which are imported but used infrequently enough to not require their own PR

pyccel/parser/semantic.py
```diff
@@ -9,6 +9,7 @@
 from collections import OrderedDict
 from itertools import chain
 
+from sympy import Tuple
```
yguclu, line 12: I'm confused. Isn't the goal of this PR to completely eradicate `sympy.Tuple` from Pyccel?

pyccel/complexity/arithmetic.py
```diff
@@ -63,7 +62,7 @@ def count_ops(expr, visual=None):
         a = expr.iterable.size
         ops = sum(count_ops(i, visual) for i in expr.body.body)
         return a*ops
-    elif isinstance(expr, (Tuple,PythonTuple)):
+    elif isinstance(expr, Tuple):
```
yguclu, line 65: Do we still need to use `Tuple` here?

pyccel/symbolic/lambdify.py
```diff
@@ -123,7 +123,7 @@ def lambdify(expr, args):
     return func
 
 def set_fst(expr, fst):
-    if isinstance(expr, (tuple,list,Tuple)):
+    if isinstance(expr, (tuple,list)):
```
yguclu, line 126: We are still using `Tuple` at line 58 of this file. Can we get rid of it?

pyccel/parser/semantic.py
EmilyBourne: It is used once in this file to create a temporary object in order to use the sympy atoms function. We cannot easily add or own version until we have removed most sympy strangeness. I think it should be possible after this PR.

pyccel/ast/core.py
yguclu: OK, thanks!

pyccel/complexity/arithmetic.py
EmilyBourne: The files in this folder begin with a call to pyccel_to_sympy

pyccel/parser/semantic.py
yguclu: I see. @saidctb Can you have a quick look into this?

pyccel/symbolic/lambdify.py
EmilyBourne: I believe lambdify uses sympy to create a usable temporary function. But I'm not sure

pyccel/complexity/arithmetic.py
yguclu: I see, so we need to use Sympy in certain places in order to perform specific operations. This should be clarified in the PR's title and description. Maybe change the title to "Remove sympy.Tuple from Pyccel's AST"?

pyccel/symbolic/lambdify.py
yguclu: I think it creates a Pyccel `FunctionDef` starting from `sympy.Lambda` or something else...

pyccel/complexity/arithmetic.py
EmilyBourne: I don't think this file is used by pyccel at the moment. Sympy should be kept where it is used as designed, ie to simplify equations etc

pyccel/symbolic/lambdify.py
EmilyBourne: I have just had another look at this. You are right, it should be replaced, but I don't know enough about this file to know if it should be replaced with `tuple` or `PythonTuple`. @saidctb do you know?

saidctb: this is for symbolic manipulations, so leave sympy there for the moment

pyccel/parser/semantic.py
saidctb: it's only used here, https://github.com/pyccel/pyccel/blob/ee4c3259c42ef0533fedee1f96161df0f45fbb59/pyccel/parser/semantic.py#L2605
which for sure doesn't work properly anymore 
we need to use `_atomic` for that

yguclu (approved):

pyccel/symbolic/lambdify.py
yguclu: OK then, I think we can merge this PR as it is.

pyccel/complexity/arithmetic.py
yguclu: Fine with me.

saidctb (approved):

PR #683, opened 2021-01-19
Access templates of a function via its decorators

### LINKED ISSUE #591: Access templates of a function via its decorators
**Describe the feature**
Now templates of a function are separated from its decorators and headers, which makes them an integral part of the function, it would make better sense to access templates of a function via its decorators.

mostafaMamoni:
Currently, we have two elements (`templates` which is a dictionary of `Template` objects, and `template` which is a list of decorator calls in the `decorators` dictionary) in `FunctionDef` class that have infos in the same theme (templates). As templates are not an integral part of a function, it was decided to group this information in the `decorators` dictionary (see #591).

This PR adds a `template` key to the `decorators` dictionary. This item contains the `template` list (of decorators) and the `templates` dictionary. As a result we don't need to pass templates dictionary as an argument to `FunctionDef` since it will be accessible from `decorators` dictionary through the `template` key. It therefore fixes #591.


pyccel/parser/semantic.py
```diff
@@ -2386,13 +2386,18 @@ def _visit_FunctionDef(self, expr, **settings):
         headers = []
 
         not_used = [d for d in decorators if d not in def_decorators.__all__]
-
+        not_used.remove('templates') #templates is not a decorator
```
rmahjoubi, line 2389: I'm not sure if this is quit clear, as we already have `template` in decorators which contains a list of `FunctionCall` that each represents a call to the `template` decorator, and then we have `templates` which basically has the same thing, represented in a different manner, it's a bit ambiguous.

mostafaMamoni: for me i was thinking about why we should insert a non-decorator variable into a decorators dict, and you told that's to make the code more structured. @EmilyBourne What do you think ?

EmilyBourne: I agree with Redouane. I'm not sure what you mean by "inserting a non-decorator variable into a decorators dict". `@template` is a decorator

mostafaMamoni: The new `templates` dict which non-decorator

pyccel/ast/core.py
```diff
@@ -2136,6 +2135,10 @@ def __init__(
                 if not isinstance(i, FunctionDef):
                     raise TypeError('Expecting a FunctionDef')
 
+        templates = decorators.setdefault('template', None)
+        if templates:
+            templates = templates.setdefault('templates_dict', None)
+
```
nhamidn, line 2141: where you use `templates` here after the initialization

pyccel/parser/syntactic.py
```diff
@@ -734,8 +735,13 @@ def fill_types(ls):
                     errors.report(msg,
                                 bounding_box = (stmt.lineno, stmt.col_offset),
                                 severity='warning')
-
                 templates[tp_name] = hdr_parse(stmts=txt)
+                template['templates_dict'] = templates
+                template['template_list'] = decorators['template']
```
rmahjoubi, line 740: Could you change templates_dict and template_list to something a bit more self explanatory

mostafaMamoni: any suggestions ?

pyccel/parser/syntactic.py
```diff
@@ -631,6 +631,7 @@ def _visit_FunctionDef(self, stmt):
         global_vars  = []
         headers      = []
         templates    = {}
+        template    = {}
```
rmahjoubi, line 634: `templates` is a temporarily variable, maybe it's better if it's mentioned in the name, I would use `tmp_templates` and `templates`

mostafaMamoni: i think its not, because its a dict that is manipulated any a whole if statement.

EmilyBourne:
This PR is marked ready for review but has not been approved. See description of "Ready_for_review":
> Received at least one approval. Requires review from senior developer

nhamidn (approved):

rmahjoubi (approved):

pyccel/codegen/printing/pycode.py
```diff
@@ -83,10 +83,13 @@ def _print_FunctionDef(self, expr):
...
                 '\n{imports}\n{body}\n').format(name=name, args=args,imports=imports, body=body)
-
         decorators = expr.decorators
-
         if decorators:
+            if decorators['template']:
+                #Eliminate templates_dict to avoid conflicts in the printing
+                expr.decorators['template'] = expr.decorators['template']['template_list']
+            else:
+                expr.decorators.pop('template')
```
EmilyBourne, line 92: Can you explain what the pop is for? Does it remove an empty dict? Is `template` always present in `decorators` or might this raise a `KeyError`?

pyccel/parser/syntactic.py
```diff
@@ -729,13 +730,18 @@ def fill_types(ls):
...
                 txt  = '#$ header template ' + str(tp_name)
                 txt += '(' + '|'.join(types) + ')'
-                if tp_name in templates:
+                if tp_name in tmp_templates:
                     msg = 'The template "{}" is duplicated'.format(tp_name)
                     errors.report(msg,
                                 bounding_box = (stmt.lineno, stmt.col_offset),
                                 severity='warning')
+                tmp_templates[tp_name] = hdr_parse(stmts=txt)
+                template['templates_dict'] = tmp_templates
```
EmilyBourne, line 739: Shouldn't this be in an `else`?

EmilyBourne, line 739: Does `tmp_templates` need saving to `template['templates_dict']` multiple times? Both `template['templates_dict']` and `tmp_templates` should reference the same object in memory once this assignment has been done once so `template['templates_dict']` should be updated automatically.

Why do we need both objects?

pyccel/parser/syntactic.py
```diff
@@ -729,13 +730,18 @@ def fill_types(ls):
...
+            template['template_list'] = decorators['template']
+            decorators['template'] = template
```
EmilyBourne, line 741: Could you explain the purpose of `template_list` please. Will we have something like:
```
decorators = {
...
template : {'template_list':[FunctionCall('template','T',('int32', 'int64')),
                             FunctionCall('template','S','('int','float'))],
            'templates_dict':{'T':Template(...), 'S', Template(...)
           }
}
```

EmilyBourne: I have had another look at `pycode.py` and I think I understand the purpose. Is this done so that we can retrieve the templates in decorator format for printing decorators?

If so could you change the names to reflect this. E.g. `tempate_dict`, `decorator_list`

EmilyBourne (changes requested):

pyccel/codegen/printing/pycode.py
EmilyBourne: `template` is `None` if there are no templates

pyccel/parser/syntactic.py
mostafaMamoni: Why ?

EmilyBourne: So you don't parse the header multiple times? Or is the aim to keep the last version?

mostafaMamoni: i think we don't need an else statement because the if block is raising a fatal error.

EmilyBourne: I don't see the fatal error can you link me to the line? I see a warning, but a warning doesn't stop the execution

mostafaMamoni: sorry i mentioned the wrong if error, So i think its better without else, since we warn the user that a template is duplicated, what do you think ?

EmilyBourne: ok. Keeping the last template probably makes sense :+1:

EmilyBourne (approved):

yguclu (changes requested):
@mostafaMamoni Can you please give more details in the PR's description? The variable names are not self-explanatory, and I am having a hard time following your changes. A few more comments in the code would be useful too.

pyccel/parser/semantic.py
```diff
@@ -2381,6 +2381,8 @@ def _visit_FunctionDef(self, expr, **settings):
         args_number = len(expr.arguments)
         templates = self.get_templates()
         if decorators['template']:
+            # Load templates dict from decorators dict
+            print(decorators['template']['template_dict'])
```
yguclu, line 2385: Why do you need to print this?

pyccel/codegen/printing/pycode.py
```diff
@@ -86,7 +86,7 @@ def _print_FunctionDef(self, expr):
         decorators = expr.decorators
         if decorators:
             if decorators['template']:
-                #Eliminate template_dict to avoid conflicts in the printing
+                # Eliminate template_dict because its useless in the printing
```
yguclu, line 89: Typo: its -> it is (or it's)

pyccel/parser/semantic.py
mostafaMamoni: sorry, i forget this.

yguclu (approved):

PR #687, opened 2021-01-20
Add print method for `pass` statement

nhamidn:
- Add print method for `pass` statement to C and Fortran printers;
- Add unit tests for `pass`;
- Copy codegen tests of functions from fcode to ccode folder.

### DIFF
--- a/tests/codegen/ccode/scripts/functions.py
+++ b/tests/codegen/ccode/scripts/functions.py
@@ -1,6 +1,20 @@
 # pylint: disable=missing-function-docstring, missing-module-docstring/
 # pylint: disable=unused-variable
 
+
+#$ header function fib(int) results(int)
+def fib(n):
+    if n < 2:
+        return n
+    i = fib(n-1)
+    j = fib(n-2)
+    return i + j
+
+#$ header function recu_func(int) results(int)
+def recu_func(x):
+    if x > 0:
+        x = x - 1
+
 def helloworld():
     print('hello world')
 
--- a/tests/codegen/ccode/scripts/functions.py
+++ b/tests/codegen/ccode/scripts/functions.py
@@ -1,0 +1,43 @@
+# pylint: disable=missing-function-docstring, missing-module-docstring/
+# pylint: disable=unused-variable
+
+def helloworld():
+    print('hello world')
+
+#$ header function incr(int)
+def incr(x):
+    x = x + 1
+
+#$ header function decr(int) results(int)
+def decr(x):
+    y = x - 1
+    return y
+
+#$ header function f1(int, int, int) results(int)
+def f1(x, n=2, m=3):
+    y = x - n*m
+    return y
+
+#$ header function f2(int, int) results(int)
+def f2(x, m=None):
+    if m is None:
+        y = x + 1
+    else:
+        y = x - 1
+    return y
+
+y = decr(2)
+z = f1(1)
+
+z1 = f2(1)
+z2 = f2(1, m=0)
+
+helloworld()
+
+# TODO add messages. for the moment there's a bug in Print
+print(z1)
+print(z2)
+
+def pass_fun():
+    pass
+
--- a/tests/codegen/ccode/scripts/functions.py
+++ b/tests/codegen/ccode/scripts/functions.py
@@ -14,6 +14,7 @@ def fib(n):
 def recu_func(x):
     if x > 0:
         x = x - 1
+    return x
 
 def helloworld():
     print('hello world')
--- a/tests/codegen/ccode/scripts/functions.py
+++ b/tests/codegen/ccode/scripts/functions.py
@@ -34,7 +34,6 @@ z2 = f2(1, m=0)
 
 helloworld()
 
-# TODO add messages. for the moment there's a bug in Print
 print(z1)
 print(z2)
 
--- a/tests/codegen/ccode/scripts/ifs.py
+++ b/tests/codegen/ccode/scripts/ifs.py
@@ -7,7 +7,6 @@ z = 3
 if 1<0:
     x=4
 
-z = 3
 if z < 4:
     pass
 
--- a/tests/codegen/ccode/scripts/ifs.py
+++ b/tests/codegen/ccode/scripts/ifs.py
@@ -7,6 +7,10 @@ z = 3
 if 1<0:
     x=4
 
+z = 3
+if z < 4:
+    pass
+
 if (x==2 or y==1) and z>3 :
     x = 3
     y = 4
--- a/tests/codegen/fcode/scripts/functions.py
+++ b/tests/codegen/fcode/scripts/functions.py
@@ -64,7 +64,6 @@ z2 = f2(1, m=0)
 
 helloworld()
 
-# TODO add messages. for the moment there's a bug in Print
 print(z1)
 print(z2)
 
--- a/tests/codegen/fcode/scripts/functions.py
+++ b/tests/codegen/fcode/scripts/functions.py
@@ -68,3 +68,6 @@ helloworld()
 print(z1)
 print(z2)
 
+def pass_fun():
+    pass
+
--- a/tests/epyccel/test_epyccel_functions.py
+++ b/tests/epyccel/test_epyccel_functions.py
@@ -367,6 +367,14 @@ def test_decorator_f22(language):
     f = epyccel(f22, language=language)
     assert f(complex(1, 2.2)) == f22(complex(1, 2.2))
 
+def test_pass(language):
+    @types('int')
+    def f22(a):
+        pass
+        return a
+    f = epyccel(f22, language=language)
+    assert f(3) == f22(3)
+
 ##==============================================================================
 ## CLEAN UP GENERATED FILES AFTER RUNNING TESTS
 ##==============================================================================
--- a/tests/epyccel/test_epyccel_functions.py
+++ b/tests/epyccel/test_epyccel_functions.py
@@ -367,14 +367,6 @@ def test_decorator_f22(language):
     f = epyccel(f22, language=language)
     assert f(complex(1, 2.2)) == f22(complex(1, 2.2))
 
-def test_pass(language):
-    @types('int')
-    def f22(a):
-        pass
-        return a
-    f = epyccel(f22, language=language)
-    assert f(3) == f22(3)
-
 ##==============================================================================
 ## CLEAN UP GENERATED FILES AFTER RUNNING TESTS
 ##==============================================================================
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -1044,7 +1044,7 @@ class CCodePrinter(CodePrinter):
         return code
 
     def _print_Pass(self, expr):
-        return ''
+        return '// pass'
 
     def _print_Nil(self, expr):
         return 'NULL'
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -1040,6 +1040,9 @@ class CCodePrinter(CodePrinter):
             code += 'return 0;'
         return code
 
+    def _print_Pass(self, expr):
+        return ''
+
     def _print_Nil(self, expr):
         return 'NULL'
 
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -374,8 +374,6 @@ class CCodePrinter(CodePrinter):
         lines = []
         for i, (c, e) in enumerate(expr.args):
             var = self._print(e)
-            if (var == ''):
-                break
             if i == 0:
                 lines.append("if (%s)\n{" % self._print(c))
             elif i == len(expr.args) - 1 and c is LiteralTrue():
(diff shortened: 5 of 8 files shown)


tests/codegen/ccode/scripts/ifs.py
```diff
@@ -7,6 +7,10 @@
 if 1<0:
     x=4
 
+z = 3
```
mostafaMamoni, line 10: `z` already assigned,

mostafaMamoni (changes requested):

mostafaMamoni (approved):

EmilyBourne (changes requested):
Please add end to end tests, not just codegen tests

rakati (commented):
I see that the statement with `pass` is removed if it contains only the pass keyword.
I think we should keep it there because the user wants it there.
```python
if True:
    pass
```
generate
```python
#include <stdlib.h>
int main()
{



    return 0;
}
```

nhamidn:
> I see that the statement with `pass` is removed if it contains only the pass keyword.
> I think we should keep it there because the user wants it there.
> 
> ```python
> if True:
>     pass
> ```
> 
> generate
> 
> ```python
> #include <stdlib.h>
> int main()
> {
> 
> 
> 
>     return 0;
> }
> ```

it is managed like this in fortran, what's the equivalent of it in C?

rakati:
can add while and function tests.
```python

def func():
     pass

while(something)
     pass
```

nhamidn:
> can add while and function tests.
> 
> ```python
> def func():
>      pass
> 
> while(something)
>      pass
> ```

the while will give an endless loop in both python and C, so it's an exception we cant handle it.

rakati:
> > can add while and function tests.
> > ```python
> > def func():
> >      pass
> > 
> > while(something)
> >      pass
> > ```
> 
> the while will give an endless loop in both python and C, so it's an exception we cant handle it.

I think this is a user problem, not us.
what about function?

EmilyBourne (changes requested):
Codacy is failing

tests/epyccel/test_epyccel_functions.py
```diff
@@ -367,6 +367,14 @@ def f22(a):
     f = epyccel(f22, language=language)
     assert f(complex(1, 2.2)) == f22(complex(1, 2.2))
 
+def test_pass(language):
+    @types('int')
+    def f22(a):
+        pass
+        return a
```
rakati, line 374: To resolve this Codacy error, remove `return a`, the code will not reach the return statement because of the keyword pass.

nhamidn: i see, so im gonna add a codegen test for the function

tests/codegen/ccode/scripts/functions.py
```diff
@@ -0,0 +1,43 @@
...
+
+y = decr(2)
+z = f1(1)
+
+z1 = f2(1)
+z2 = f2(1, m=0)
+
+helloworld()
+
+# TODO add messages. for the moment there's a bug in Print
```
EmilyBourne, line 37: Can you explain this TODO? Or is it copied and pasted from fcode? If it's copy pasted then can you look into it and see if there is still a problem

EmilyBourne (changes requested):
Please update the PR description

tests/codegen/ccode/scripts/functions.py
nhamidn: yes it's copied and pasted from the code, I tested it and the printing seems to be working normally in `Fortran` 
but o get a compilation error in `C`
for example this python code : 
```
#$ header function f2(int, int) results(int)
def f2(x, m=None):
    if m is None:
        y = x + 1
    else:
        y = x - 1
    return y


z1 = f2(1)
z2 = f2(1, m=0)

# TODO add messages. for the moment there's a bug in Print
print(z1)
print(z2)
```
generate this main C code which fail in the compilation : 

```
int main()
{
    int64_t z1;
    int64_t z2;


    z1 = f2(1, NULL);
    z2 = f2(1, &m);
    /*TODO add messages. for the moment there's a bug in Print*/
    printf("%ld\n", z1);
    printf("%ld\n", z2);
    return 0;
}
```
```
error: use of
      undeclared identifier 'm'
```

EmilyBourne: Ok. This error is not related to the print so could you remove the TODO message in both files please. In C it looks like there is either a problem with passing values to optional parameters. I.e:
```
#$ header function f2(int, int) results(int)
def f2(x, m=None):
    ...


z2 = f2(1, 0)
```
Or there is a problem with keyword arguments. I.e:
```
#$ header function f2(int, int) results(int)
def f2(x, m):
    ...


z1 = f2(1, m=0)
```
Or there is a problem when combining the 2. I.e.:
```
#$ header function f2(int, int) results(int)
def f2(x, m = None):
    ...


z1 = f2(1, m=0)
```

Could you work out which and open an issue for this please

nhamidn: i tested those cases, and it look like there is a problem with keyword arguments.

EmilyBourne: ok. Thanks :)

EmilyBourne (changes requested):
@rakati makes a very good point
> I see that the statement with `pass` is removed if it contains only the pass keyword.
> I think we should keep it there because the user wants it there.
> 
> ```python
> if True:
>     pass
> ```
> 
> generate
> 
> ```python
> #include <stdlib.h>
> int main()
> {
> 
> 
> 
>     return 0;
> }
> ```

This is related to the handling of if blocks in C [here](https://github.com/pyccel/pyccel/blob/master/pyccel/codegen/printing/ccode.py#L378). You are right, this is a bug once `pass` is allowed. Consider the following code:
```
def f(b : bool):
    c = 1
    if b:
        pass
    else:
        c = 2
    return c
```
This is translated to:
```
int64_t f(bool b)
{
    int64_t c;
    c = 1;

    return c;
}
```
The code without the `pass` statement is not valid python code so this bug would be introduced by this PR. It needs fixing and a test needs adding.

I will give my next review after @rakati approves

rakati (changes requested):
see this case:
```python
def walo(z: bool):
    x = 0
    if z:
        pass
        x = 1
    else:
        x = 2
    return x
```
generated code 
```python
int64_t walo(bool z)
{
    int64_t x;
    x = 0;
    if (z)
    {

        x = 1;
    }
    else if (1)
    {
        x = 2;
    }
    return x;
}
```
pass keyword usually used for debugging, so users need to see this keyword also in generated code.
I suggest instead of blanc space
```python
def _print_pass(self, expr):
    return "// pass"
``` 
and for the bug at the top, I see that the goto keyword will be the equivalent of the pass keyword in c and c++.

EmilyBourne:
> see this case:
> 
> ```python
> def walo(z: bool):
>     x = 0
>     if z:
>         pass
>         x = 1
>     else:
>         x = 2
>     return x
> ```
> 
> generated code
> 
> ```python
> int64_t walo(bool z)
> {
>     int64_t x;
>     x = 0;
>     if (z)
>     {
> 
>         x = 1;
>     }
>     else if (1)
>     {
>         x = 2;
>     }
>     return x;
> }
> ```
> 
> pass keyword usually used for debugging, so users need to see this keyword also in generated code.
> I suggest instead of blanc space
> 
> ```python
> def _print_pass(self, expr):
>     return "// pass"
> ```
> 
> and for the bug at the top, I see that the goto keyword will be the equivalent of the pass keyword in c and c++.

I think you are confusing `pass` and `break`. See https://realpython.com/python-pass/
Your example does not do what you think it does:
```
In [1]: def walo(z: bool):
   ...:     x = 0
   ...:     if z:
   ...:         pass
   ...:         x = 1
   ...:     else:
   ...:         x = 2
   ...:     return x
   ...: 

In [2]: print(walo(True))
1
```

EmilyBourne:
@rakati 
> pass keyword usually used for debugging, so users need to see this keyword also in generated code.

The `pass` keyword is useless for debugging as it does nothing. It is used when the python syntax demands that something is written but you do not want to do anything. It is usually used during development when you haven't decided what to put in the if block/function/whatever yet. For example in pyccel it is used in ast/numpyext.py for classes with no contents. E.g:
```
class NumpyFabs    (NumpyUfuncUnary): pass
```

rakati:
> @rakati
> 
> > pass keyword usually used for debugging, so users need to see this keyword also in generated code.
> 
> The `pass` keyword is useless for debugging as it does nothing. It is used when the python syntax demands that something is written but you do not want to do anything. It is usually used during development when you haven't decided what to put in the if block/function/whatever yet. For example in pyccel it is used in ast/numpyext.py for classes with no contents. E.g:
> 
> ```
> class NumpyFabs    (NumpyUfuncUnary): pass
> ```

yes, I am sorry for that, I didn't experiment with it.
@EmilyBourne what do you think about return `// pass` instead of `''`, so the user when reading the c code he will recognize where the `pass `keyword is written.

EmilyBourne:
@rakati 
> @EmilyBourne what do you think about return // pass instead of '', so the user when reading the c code he will recognize where the pass keyword is written.

This makes sense to me

tests/codegen/ccode/scripts/functions.py
```diff
@@ -0,0 +1,42 @@
...
+    print('hello world')
+
+#$ header function incr(int)
+def incr(x):
+    x = x + 1
+
+#$ header function decr(int) results(int)
+def decr(x):
+    y = x - 1
+    return y
```
rakati, line 14: great, you add more tests.
correct me if I am wrong, do we use the` result(type)` only for recursive functions.
I see this confusing even it just for tests, but the user sometimes reads the tests to know how to work with pyccel.

nhamidn: no, it's required when writing a recursive function but it's optional for other function

tests/codegen/fcode/scripts/functions.py
```diff
@@ -64,7 +64,9 @@ def f2(x, m=None):
 
 helloworld()
 
-# TODO add messages. for the moment there's a bug in Print
```
rakati, line 67: why you remove this comment, did you check if the bug mentioned here is solved or not?
you did not mention the reason for removing this todo comment in your commit https://github.com/pyccel/pyccel/pull/687/commits/f806164cd0c1e21b55e63a3e1590461e6c7eee81
It will be useful to describe the reason in the commit, so anyone who reviews the code will understand your changes.

tests/codegen/ccode/scripts/functions.py
nhamidn: @rakati

tests/codegen/ccode/scripts/functions.py
rakati: so maybe creating a recursive function test with the `result(type) `is more convenient.
so we let the uses of result type for its main purpose (as you know the programmer don't like too much typing).

tests/codegen/fcode/scripts/functions.py
nhamidn: we already discussed about it in this PR, https://github.com/pyccel/pyccel/pull/687#discussion_r561635957

tests/codegen/ccode/scripts/functions.py
nhamidn: this codegen test was only in the fcode folder so I wake it available in the ccode, I think creating tests for functions is beyond the scope of this PR, I just didn't find where I can put test for a function with the `pass` statement so I copied the tests of the function from the fcode to the ccode

rakati: For me I review the code that you added, I don't care about its source.
the important for me is to see the added changes clear and correctly.
But it's up to you if it's fine for you to copy-past inconvenient code a left it as it is, what could I say to you.

nhamidn: here i think that a non-recursive function using a `result type` must also be tested, we can also add tests for recursive function

rakati (approved):

tests/codegen/ccode/scripts/functions.py
EmilyBourne: @rakati is right that we shouldn't introduce poor/unexplainable code just because there is similar code elsewhere.
However I agree with @nhamidn here. The `result(type)` is optional for non-recursive functions, but that means that it shouldn't cause problems if it is specified so having 1 test for this is not a problem

tests/codegen/ccode/scripts/functions.py
```diff
@@ -0,0 +1,56 @@
...
+    if n < 2:
+        return n
+    i = fib(n-1)
+    j = fib(n-2)
+    return i + j
+
+#$ header function recu_func(int) results(int)
+def recu_func(x):
+    if x > 0:
+        x = x - 1
```
EmilyBourne, line 16: This function specifies results but has no results

EmilyBourne (approved):

yguclu:
I often use `pass` for clarity, and I think it is its intended usage. For example:
```python
if a < 0:
    raise ValueError('Cannot accept negative values')
elif 0 <= a < 1:
    pass   # This is the accepted range
else:
    a %= 1 # Only consider fractional part of a
```

yguclu (approved):

Issue #694, opened 2021-01-21
Add children and parents to pyccel.ast.basic.Basic


EmilyBourne:
**Describe the new feature**
Sympy has a [`subs`](https://docs.sympy.org/latest/tutorial/basic_operations.html#substitution) function which is one of the few parts of the sympy mechanism that we wish to use. As we are removing the unnecessary sympy inheritance this function must be re-written.
The original `subs` function substitutes amongst all the arguments of a class. Implementing this ourselves will allow us to have more control over which classes are considered to be atomic objects and which arguments are classes as children.
Handling this in the constructor of `pyccel.ast.basic.Basic` will ensure that this method exists for all ast nodes. It will also allow the parent to be set automatically.

This will allow the following improvements

- Identifying the parents of nodes
     - Allows us to determine the status of an `Allocate` node
- Ensures the `atoms` and `subs` functions are not susceptible to sympy based problems

EmilyBourne:
Fixed by #718

PR #697, opened 2021-01-21
Avoid printing temporary variables for scalar single return in c

mostafaMamoni:
When a function in C returns 1 scalar variable a temporary variable is created to mimic fortran. This is not necessary, to avoid that we should make `_print_return` in ccode.py prints directly the return of scalar/variable without assigning a temporary variable,


MenoIy:
> after that we will take care of not declaring unused temporary variables in another PR.

What do you mean by another PR, are you going to leave an unused temporary variable in the c code?

mostafaMamoni:
> > after that we will take care of not declaring unused temporary variables in another PR.
> 
> What do you mean by another PR, are you going to leave an unused temporary variable in the c code?

i think i miss understand emilly, i will fix in this PR.

pyccel/codegen/printing/ccode.py
```diff
@@ -953,8 +954,13 @@ def _print_FunctionDef(self, expr):
             self._additional_args.append(expr.results)
         body  = self._print(expr.body)
         decs  = [Declare(i.dtype, i) if isinstance(i, Variable) else FuncAddressDeclare(i) for i in expr.local_vars]
-        if len(expr.results) <= 1 :
-            decs += [Declare(i.dtype, i) if isinstance(i, Variable) else FuncAddressDeclare(i) for i in expr.results]
+        if len(expr.results) == 1 :
+            if isinstance(expr.results[0], Variable):
+                if expr.results[0] not in self._unnecessary_declaration:
+                    print(type(expr.results[0]))
```
rakati, line 960: you forgot to remove the print statement.

rakati:
please add some tests to the codegen.

pyccel/codegen/printing/ccode.py
```diff
@@ -953,8 +954,12 @@ def _print_FunctionDef(self, expr):
...
         body  = self._print(expr.body)
         decs  = [Declare(i.dtype, i) if isinstance(i, Variable) else FuncAddressDeclare(i) for i in expr.local_vars]
-        if len(expr.results) <= 1 :
-            decs += [Declare(i.dtype, i) if isinstance(i, Variable) else FuncAddressDeclare(i) for i in expr.results]
+        if len(expr.results) == 1 :
+            if isinstance(expr.results[0], Variable):
+                if expr.results[0] not in self._unnecessary_declaration:
+                    decs += [Declare(expr.results[0].dtype, expr.results[0])]
+            elif isinstance(expr.results[0], FunctionAddress):
+                decs += [FuncAddressDeclare(expr.results[0])]
```
rakati, line 962: We don't add function's docstring for making pylint or Codacy happy, we are doing it for anyone who wants to read our code, the same for a line comment.
It may be easy to read your code now, but it won't be in the future.
so please try to add comments on your code.

mostafaMamoni:
> please add some tests to the codegen.

i think we have enough tests.

pyccel/codegen/printing/ccode.py
mostafaMamoni: by this, we will have a code like the header files of an OS (e.g [this](https://opensource.apple.com/source/cctools/cctools-795/include/mach-o/loader.h)), if you really want to understand the code just ask me.

pyccel/codegen/printing/ccode.py
```diff
@@ -953,8 +954,12 @@ def _print_FunctionDef(self, expr):
             self._additional_args.append(expr.results)
         body  = self._print(expr.body)
         decs  = [Declare(i.dtype, i) if isinstance(i, Variable) else FuncAddressDeclare(i) for i in expr.local_vars]
-        if len(expr.results) <= 1 :
-            decs += [Declare(i.dtype, i) if isinstance(i, Variable) else FuncAddressDeclare(i) for i in expr.results]
+        if len(expr.results) == 1 :
+            if isinstance(expr.results[0], Variable):
+                if expr.results[0] not in self._unnecessary_declaration:
```
rakati, line 959: one `if` statement is enough.

pyccel/codegen/printing/ccode.py
```diff
@@ -268,6 +268,7 @@ def __init__(self, parser, settings=None):
         self._parser = parser
         self._additional_code = ''
         self._additional_declare = []
+        self._unnecessary_declaration = []
```
rakati, line 271: can you choose a meaningful name here. 
this dictionary is used to hold only declared variables in the `return` statement.

mostafaMamoni: No, it used to hold unnecessary declarations.

mostafaMamoni: we use it `_print_FunctionDef` to avoid declaring unused vars inside it.

saidctb: can't you just delete that variable from  `self._additional_declare`

mostafaMamoni: but its not in `self._additional_declare` other wise it will be declared in `_print_functiondef` in this line `decs += [Declare(i.dtype, i) for i in self._additional_declare]`, and i think this line is needed in other stuff.

mostafaMamoni: other than that i think i should store the unused variable in a normal variable instead of list since we assigned just a single scalar return or an expression to the unused variable.

saidctb: ah yes you are right

saidctb: how about deleting the variable from the Scope ?

pyccel/codegen/printing/ccode.py
```diff
@@ -953,8 +953,11 @@ def _print_FunctionDef(self, expr):
             self._additional_args.append(expr.results)
         body  = self._print(expr.body)
         decs  = [Declare(i.dtype, i) if isinstance(i, Variable) else FuncAddressDeclare(i) for i in expr.local_vars]
-        if len(expr.results) <= 1 :
-            decs += [Declare(i.dtype, i) if isinstance(i, Variable) else FuncAddressDeclare(i) for i in expr.results]
+        if len(expr.results) == 1 :
+            if isinstance(expr.results[0], Variable):
+                decs += [Declare(expr.results[0].dtype, expr.results[0])]
+            elif isinstance(expr.results[0], FunctionAddress):
+                decs += [FuncAddressDeclare(expr.results[0])]
```
MenoIy, line 960: you are printing declare anyway is that what you want?
did you mean to remove the first condition?

MenoIy (changes requested):

pyccel/codegen/printing/ccode.py
```diff
@@ -953,8 +953,11 @@ def _print_FunctionDef(self, expr):
             self._additional_args.append(expr.results)
         body  = self._print(expr.body)
         decs  = [Declare(i.dtype, i) if isinstance(i, Variable) else FuncAddressDeclare(i) for i in expr.local_vars]
-        if len(expr.results) <= 1 :
-            decs += [Declare(i.dtype, i) if isinstance(i, Variable) else FuncAddressDeclare(i) for i in expr.results]
+        if len(expr.results) == 1 :
+            if isinstance(expr.results[0], Variable):
+                decs += [Declare(expr.results[0].dtype, expr.results[0])]
```
MenoIy, line 958: in return, you have property stmt, maybe you can use is here in the condition 
```PYTHON
if isinstance(expr.results[0], Variable) and not expr.results.stmt: # here if stmt is empty it can mean that this is not a tmp return . 
```
can you try this ?

MenoIy (changes requested):

pyccel/codegen/printing/ccode.py
MenoIy: my bad don't do this :)

pyccel/codegen/printing/ccode.py
```diff
@@ -1043,6 +1044,7 @@ def _print_Return(self, expr):
             return 'return 0;'
         if len(args) == 1:
             if expr.stmt:
+                self._unused_variable = expr.stmt.body[0].rhs
```
MenoIy, line 1047: lhs?

MenoIy (changes requested):

pyccel/codegen/printing/ccode.py
mostafaMamoni: yep, sorry

pyccel/codegen/printing/ccode.py
mostafaMamoni: what do you mean by the Scope ?

pyccel/codegen/printing/ccode.py
EmilyBourne: Documentation is always important if the purpose of the code is not clear from reading it. You don't need a comment per line, but a comment per idea is usually a good idea

>  if you really want to understand the code just ask me.

This is *always* a bad answer as you have no idea when someone will have questions about this. If someone has a question in 3 years time then you may not remember any more and you may not be working on pyccel anymore. In the worst case we may not even have a way to contact you any more. This is why clear, well documented code is *very* important, especially in a big project

MenoIy (approved):

pyccel/codegen/printing/ccode.py
EmilyBourne: >how about deleting the variable from the Scope ?

These are not local variables. I don't think they are in the Scope. Instead they are simply temporary variables created from the results property of the class `FunctionDef`. The variables which should be printed could be selected and added to `self._additional_declare`, or we can assume that all are declared, and create a second set of variables for things that we thought we should declare but were deemed unnecessary later (in this case in `_print_Return`)

pyccel/codegen/printing/ccode.py
```diff
@@ -267,6 +267,7 @@ def __init__(self, parser, settings=None):
         self._additional_imports = set(['stdlib'])
         self._parser = parser
         self._additional_code = ''
+        self._unused_variable = None
```
EmilyBourne, line 270: Can you explain why there can only be one unused variable at a time? Are we sure that this won't create problems in the case of functions within functions? Or is the placement of the return statement sufficient to avoid problems here?

pyccel/codegen/printing/ccode.py
```diff
@@ -953,8 +954,11 @@ def _print_FunctionDef(self, expr):
             self._additional_args.append(expr.results)
         body  = self._print(expr.body)
         decs  = [Declare(i.dtype, i) if isinstance(i, Variable) else FuncAddressDeclare(i) for i in expr.local_vars]
-        if len(expr.results) <= 1 :
-            decs += [Declare(i.dtype, i) if isinstance(i, Variable) else FuncAddressDeclare(i) for i in expr.results]
+        if len(expr.results) == 1 :
+            if isinstance(expr.results[0], Variable) and expr.results[0] != self._unused_variable:
+                decs += [Declare(expr.results[0].dtype, expr.results[0])]
+            elif isinstance(expr.results[0], FunctionAddress):
+                decs += [FuncAddressDeclare(expr.results[0])]
```
EmilyBourne, line 961: If `_unused_variable` is to be reusable for anything other than return variables (its name does not indicate that it is only useful for returns so I assume it is general) then the big if here should be done for every declaration. Including the declarations in the first line of this code

pyccel/codegen/printing/ccode.py
```diff
@@ -953,8 +954,11 @@ def _print_FunctionDef(self, expr):
...
         decs += [Declare(i.dtype, i) for i in self._additional_declare]
```
EmilyBourne, line 962: See above. Could it be possible one day for something in `_additional_declare` to also be in `_unused_variable`?

EmilyBourne (changes requested):
`_unused_variable` is never unset. If we follow the reasoning that `_unused_variable` could be used for any unused variable that we may have in the future then this could be a problem

pyccel/codegen/printing/ccode.py
mostafaMamoni: you are right about functions within functions, So we can use a list.

mostafaMamoni: 😅 but C does not support nested functions

pyccel/codegen/printing/ccode.py
mostafaMamoni: did mean that one day `_additional_declare` could have an unused variable?

pyccel/codegen/printing/ccode.py
mostafaMamoni: its a great i idea to have something that holds unused variables and it will even better if we have the ability to detect unused variables especially in local variables, but this variable in particular is specific in holding the unused variable that is generated from the return statement, So i think i will change its name to be more clear, and for holding unused variables in general i think is to be implemented.

EmilyBourne: If you only plan for this object to hold unused variables generated by the return statement then renaming it is important in order for that to be clear.

pyccel/codegen/printing/ccode.py
EmilyBourne: This is more to do with the processing order than the syntax. If we did:
```
def _visit_FunctionDef(self, expr):
    nested_funcs = self._visit(expr.funcs)
    ... # Existing code
    return code + '\n' + nested_funcs
```
then we could still potentially translate nested functions. But the `return` always has to be the last statement in a function so if this object is only used for returns (see rename discussion below) then I don't think there is a problem

pyccel/codegen/printing/ccode.py
saidctb: they are in the Scope, because we need to infere their type in the semantic stage.

EmilyBourne: You are right, it is done [here](https://github.com/pyccel/pyccel/blob/master/pyccel/parser/semantic.py#L2545). I thought these were created in the semantic stage, sorry. Why do we need to delete them from the Scope at this point?

EmilyBourne (changes requested):
If you want to restrict the use of `_unused_variable` to this use case, I agree that it must be renamed

pyccel/codegen/printing/ccode.py
saidctb: it's the easiest thing no ? instead of ingnoring them, we can just delete them so that we don't declare them .

pyccel/codegen/printing/ccode.py
```diff
@@ -267,6 +267,7 @@ def __init__(self, parser, settings=None):
         self._additional_imports = set(['stdlib'])
         self._parser = parser
         self._additional_code = ''
+        self._tmp_var_tobereturned = None
```
EmilyBourne, line 270: This new name is not clear and seems to indicate the opposite use to what is implemented. I suggest `_unused_return_variable`

EmilyBourne (changes requested):

saidctb:
I have an idea, let's discuss this in a meeting.

pyccel/codegen/printing/ccode.py
EmilyBourne: We don't declare them from the scope. They are declared [here](https://github.com/pyccel/pyccel/blob/master/pyccel/codegen/printing/ccode.py#L956-L957)

EmilyBourne (changes requested):
It would be great if you could add tests with the flags `-Werror -Wununsed-variable`.

Please wait until after you have discussed with @saidctb and have implemented anything which comes out of that meeting, to re-request my review as I don't want to approve if the code will change significantly later

EmilyBourne:
Don't forget to fix this problem in the python printer too

saidctb (approved):

tests/epyccel/test_return_stmt.py
```diff
@@ -0,0 +1,68 @@
+# pylint: disable=missing-function-docstring, missing-module-docstring/
+from pyccel.decorators import types
+from pyccel.epyccel import epyccel
+
+def test_single_return_var_assign(language):
+    def single_return_var_assign():
+        y = 3
+        return y
+    epyc_single_return_var_assign = epyccel(single_return_var_assign, language=language)
```
EmilyBourne, line 9: This `epyccel` call is the only one which doesn't use `-Werror -Wunused-variable`. Is that intentional?

EmilyBourne (changes requested):
The problem still needs to be fixed in the python printer (1-2 line fix).
As you have added the `is_tmp` property you should ensure that it is set for all temporary variables, not just the return temporaries

tests/epyccel/test_return_stmt.py
mostafaMamoni: you are right

EmilyBourne: Why doesn't it use them?

EmilyBourne (changes requested):
3 out of 4 of my previous comments are not resolved

tests/epyccel/test_return_stmt.py
mostafaMamoni: this test if the return itself not for the temporary variable.

mostafaMamoni:
> 3 out of 4 of my previous comments are not resolved

after Said's approval or before.

pyccel/codegen/printing/ccode.py
mostafaMamoni: the thing is just how can i tell is the code is readable for someone, So by asking me i can tell that and write some clarifications.

> a comment per idea

i do agree with that, actually [this](https://github.com/pyccel/pyccel/pull/683/commits/700bc7c3452545a77cc21e9135e229c971b6972e) what i do when i wrote a line that not answering the question  "why"  from my perspective.

tests/epyccel/test_return_stmt.py
EmilyBourne: Returns are tested in pretty much every other file in epyccel. I'm not sure this is needed in addition to that. It does not test anything that isn't tested later in the file. That said it doesn't do any harm so ok

EmilyBourne:
> > 3 out of 4 of my previous comments are not resolved
> 
> after Said's approval or before ?, can you mention the topic of the comments

After:
1. Fix the python printer : done
2. Ensure that the `is_tmp` property is set for all temporary variables (e.g. everywhere get_new_variable is called if it is possible)
3. 
    a. I am waiting for an answer to the question `Do you think this makes sense or should we only allow Assign?`
    b. The return statement in C should come from the last statement (presumably Assign) in a code block, not the first one
4. Discussion about tests : Now done

pyccel/codegen/printing/ccode.py
```diff
@@ -1033,15 +1036,16 @@ def _print_Constant(self, expr):
...
+            if expr.stmt:
+                return self._print(expr.stmt)+'\n'+'return 0;'
+            return 'return 0;'
         if len(args) == 1:
-            code +='return {0};'.format(self._print(args[0]))
-        elif len(args) > 1:
-            code += 'return 0;'
-        return code
+            if expr.stmt:
+                return 'return {0};'.format(self._print(expr.stmt.body[0].rhs))
```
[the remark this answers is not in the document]
mostafaMamoni: i think we should handle both the codeBlock and the Assign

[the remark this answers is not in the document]
EmilyBourne: In that case your code won't work if `expr.stmt` is an `Assign`. Another option would be for the `expr.stmt` to always be a `CodeBlock`. Is it ever an `Assign` currently?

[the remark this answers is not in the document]
mostafaMamoni: thats what i observe its always `CodeBlock`, i just need test cases when `expr.stmt` comes as an Assign.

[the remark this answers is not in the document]
mostafaMamoni: should i handle it as in Python printer?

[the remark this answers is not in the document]
EmilyBourne: If it is currently always a `CodeBlock` then I think it makes sense for this to be the expected behaviour. So I suggest that we update `Return.__new__` to only accept a `CodeBlock` object, or to put an `Assign` argument into a `CodeBlock`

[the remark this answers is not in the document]
EmilyBourne: >should i handle it as in Python printer?

I'm not sure I see what you mean here. In the python printer you return multiple objects which is obviously not possible in c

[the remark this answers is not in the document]
EmilyBourne: Ah I have just seen what you mean. This is also possible but I think it would be easier to just ensure that `Return.stmt` always contains a `CodeBlock`, that way we don't have to maintain 2 versions everywhere despite the fact that only 1 is used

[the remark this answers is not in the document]
EmilyBourne: If we are keeping `CodeBlock` though, it is important that you collect the rhs from the **last** statement not the first. It is possible that more complicated expressions may be converted to multiple lines. E.g:
```
return [i for i in range(4)]
```
would probably be translated to something like:
```
Allocate(Dummy_0001, 4)
For(Dummy_0002, range(4), Assign(Dummy_001[Dummy_0002], Dummy_0002))
Assign(Out_0001, Dummy_0001)
```
(This is probably not an accurate translation but I can't think of a better example right now) You can see that in a case like this it is not the first line of the block which interests you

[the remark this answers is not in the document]
saidctb: yes it will be the last statement

pyccel/codegen/printing/ccode.py
```diff
@@ -1033,15 +1036,21 @@ def _print_Constant(self, expr):
...
-            code +='return {0};'.format(self._print(args[0]))
-        elif len(args) > 1:
-            code += 'return 0;'
-        return code
+            if isinstance(expr.stmt, CodeBlock):
+                for a in expr.stmt.body:
+                    if isinstance(a, Assign):
+                        b = a
+                if isinstance(b.rhs, IndexedElement):
+                    return '{0}\nreturn {1};'.format(self._print(expr.stmt), self._print(b.lhs))
```
EmilyBourne, line 1050: It is not all `IndexedElements` that are problematic, just the ones whose array needs deallocating. For example:
```
def f(a : 'int[:]'):
    return a[0]
```
Can be translated without temporary variables

pyccel/codegen/printing/ccode.py
```diff
@@ -1033,15 +1036,21 @@ def _print_Constant(self, expr):
...
+            return 'return 0;'
         if len(args) == 1:
-            code +='return {0};'.format(self._print(args[0]))
-        elif len(args) > 1:
-            code += 'return 0;'
-        return code
+            if isinstance(expr.stmt, CodeBlock):
+                for a in expr.stmt.body:
+                    if isinstance(a, Assign):
+                        b = a
```
EmilyBourne, line 1048: Do we definitely need this loop? Do we want the last `Assign`? Or the last element and printing the assignment can be avoided if the last element is an `Assign`?

pyccel/codegen/printing/ccode.py
```diff
@@ -1033,15 +1036,21 @@ def _print_Constant(self, expr):
...
+                if isinstance(b.rhs, IndexedElement):
+                    return '{0}\nreturn {1};'.format(self._print(expr.stmt), self._print(b.lhs))
+                return 'return {0};'.format(self._print(b.rhs))
+            return 'return {0};'.format(self._print(args[0]))
```
EmilyBourne, line 1052: In a complex case like the one above, we may need to print some of the `expr.stmt`. The only part which we know we can remove is the lhs of the `Assign`

EmilyBourne (changes requested):
`Dummy_XX` variables are still not marked as temporary

pyccel/ast/core.py
```diff
@@ -2159,7 +2159,7 @@ class Return(Basic):
 
     def __new__(cls, expr, stmt=None):
 
-        if stmt and not isinstance(stmt, (Assign, CodeBlock)):
+        if stmt and not isinstance(stmt, CodeBlock):
             raise TypeError('stmt should only be of type Assign')
```
EmilyBourne, line 2163: Please update this error message too

pyccel/codegen/printing/ccode.py
EmilyBourne: By this I mean the case where a one line statement in python is translated to multiple nodes (I can't think of an example which works currently, but there could easily be one in the future). E.g:
```
return [i for i in range(4)]
```
would probably be translated to something like:
```
Allocate(Dummy_0001, 4)
For(Dummy_0002, range(4), Assign(Dummy_001[Dummy_0002], Dummy_0002))
Assign(Out_0001, Dummy_0001)
```
(This is probably not an accurate translation but I can't think of a better example right now) In this case you cannot just print `return Dummy_0001`

pyccel/codegen/printing/ccode.py
```diff
@@ -1033,15 +1036,18 @@ def _print_Constant(self, expr):
...
+                return self._print(expr.stmt)+'\n'+'return 0;'
+            return 'return 0;'
         if len(args) == 1:
-            code +='return {0};'.format(self._print(args[0]))
-        elif len(args) > 1:
-            code += 'return 0;'
-        return code
+            if isinstance(expr.stmt, CodeBlock):
+                if  isinstance(expr.stmt.body[-1], Deallocate):
+                    return '{0}\nreturn {1};'.format(self._print(expr.stmt), self._print(expr.stmt.body[-2].lhs))
```
EmilyBourne, line 1047: It is possible to have more than one `Deallocate` statement in a `Return`. For example, consider the following code:
```
def mat_mat_mat_mul(a : 'int[:,:]', b : 'int[:,:]', c : 'int[:,:]'):
    d = np.empty((a.shape[0], b.shape[1]))
    e = np.empty_like((b.shape[0], c.shape[1])
    d[:,:] = np.matmul(a,b)
    e[:,:] = np.matmul(b,c)
    a[:,:] = np.matmul(d,e)
    return True
```
This will be translated to:
```
FunctionDef(args = (Variable(a), Variable(b), Variable(c)),
                     result = Variable(Out_0001),
                     body = (
                           Allocate(d,...),
                           Allocate(e,...),
                           Assign(d[:,:], NumpyMatmul(a,b))
                           Assign(e[:,:], NumpyMatmul(b,c))
                           Assign(a[:,:], NumpyMatmul(d,e))
                           Return( expr = (Out_0001,)
                                        stmt = (
                                                Assign(Out_0001, LiteralTrue()),
                                                Deallocate(d),
                                                Deallocate(e)
                                               )
                                  )
                            )
           )
```

pyccel/codegen/printing/ccode.py
EmilyBourne: @EmilyBourne  Yes we do need this loop as all `Deallocate` statements are put in `Return.stmt` and not all (or necessarily any of them) have anything to do with the value which is returned.

@mostafaMamoni If I ask a question it is because I am not certain of the answer. In this case your code above was correct. If you can justify why something is written like that then this is usually a better answer than just changing it

pyccel/codegen/printing/ccode.py
EmilyBourne: For example, consider the example given above with multiple `Deallocate` statements. The `Assign` has nothing to do with these `Deallocate` statements. So we can print `'return {0};'.format(self._print(assign_stmt.rhs))`, however this doesn't mean that we can avoid printing all the rest of `expr.stmt`

EmilyBourne (changes requested):
`Dummy_XX` variables are still not marked as temporary

pyccel/codegen/printing/ccode.py
EmilyBourne: Another example:
```
@stack_array('b')
def f(a : int):
    b = np.ones(4)
    return b[a]
```
Here there is also no `Deallocate` statement

pyccel/codegen/printing/ccode.py
EmilyBourne: I suggest you have a look at the construction of the `Return` node:
https://github.com/pyccel/pyccel/blob/master/pyccel/parser/semantic.py#L2337-L2360

That way you should have a better idea of what it can contain and in what order

mostafaMamoni:
We can't mark `Dummy_XX` variables as temporary, because all the variables are converted from Symbol objects to Variable objects in `_visit_Assign` so there is currently no way to do this.

mostafaMamoni:
> For example, consider the example given above with multiple Deallocate statements. The Assign has nothing to do with these Deallocate statements. So we can print 'return {0};'.format(self._print(assign_stmt.rhs)), however this doesn't mean that we can avoid printing all the rest of expr.stmt

@EmilyBourne do you mean that we need to print just the deallocation of the array that we returned from.

pyccel/codegen/printing/ccode.py
EmilyBourne: > @EmilyBourne do you mean that we need to print just the deallocation of the array that we returned from.

No!!! Exactly the opposite. We need to make sure that we don't lose any `Deallocate` statements otherwise we will have a memory leak.

Maybe @saniQQ can help you understand this if you are not sure why the `Deallocate` statements are important or why they are in the `Return` object

yguclu:
Hi @mostafaMamoni. What is blocking this PR?

mostafaMamoni:
> Hi @mostafaMamoni. What is blocking this PR?

it is blocked by #718.

saidctb:
I don't think this PR should be merged anymore, it adds complications that are not needed

mostafaMamoni:
> I don't think this PR should be merged anymore, it adds complications that are not needed

yes, its just about to make the code looks more pretty

saidctb:
> > I don't think this PR should be merged anymore, it adds complications that are not needed
> 
> yes, its just about to make the code looks more pretty

yes but we can still use what you have done for the temporary variables, so that we keep the information

EmilyBourne:
@yguclu 
A bit more information about your question:
> Hi @mostafaMamoni. What is blocking this PR?

During the review of this PR we noticed that the situation is more complicated than described in the issue when the function contains local arrays allocated on the heap. In this case there are deallocations which need to take place before the return statement. This can be a problem if the statement being returned uses an array operation. E.g:
```
def f(j : int):
    a = np.empty(j)
    ...
    return a.sum()
```
In this case the temporary variable is necessary so that the sum is calculated before `a` is deallocated. However in the following case the temporary variable is not necessary:
```
def g(a : 'int[:]'):
    return a.sum()
```
Currently there is no clean way to determine whether an expression contains a given variable. Once issue #694 is fixed then we can easily check whether or not the expression describing the contents of the temporary variable contains any of the variables which are deallocated. And thus whether the temporary variable is required or not

EmilyBourne:
@saidctb 
> I don't think this PR should be merged anymore, it adds complications that are not needed

I agree that it is technically unnecessary but the code is somewhat unnatural in its current form which is why we created the issue in the first place.

This PR achieves the aim with relatively few lines of code so I am not clear on which bit specifically you think is too complicated. Do you have any specific section in mind?

saidctb:
> @EmilyBourne 
> 
> > I don't think this PR should be merged anymore, it adds complications that are not needed
> 
> I agree that it is technically unnecessary but the code is somewhat unnatural in its current form which is why we created the issue in the first place.
> 
> This PR achieves the aim with relatively few lines of code so I am not clear on which bit specifically you think is too complicated. Do you have any specific section in mind?

I mean, you need to take into account the garbage collection, so we can't do it with arrays, and we are left with only scalar symbols and expression with scalar symbols

EmilyBourne:
@saidctb 
> I mean, you need to take into account the garbage collection, so we can't do it with arrays

Since PR #718 we can take into account the garbage collection in 3 lines

mostafaMamoni:
the commits of #719 get merged in this PR, the branches are fine, the bug is just in the web page.

PR #704, opened 2021-01-22
Remove sympy indexed

### LINKED ISSUE #658: Remove sympy.Indexed from Pyccel
`sympy.Indexed` is used for indexing at the syntactic stage. `IndexedElement` should be sufficient and would prevent accidental sympifying of indexes

EmilyBourne:
The sympy object `Indexed` is not needed by pyccel as `IndexedElement` is sufficient. This PR removes it and the associated class `IndexedBase`. This fixes #658


pyccel/parser/semantic.py
```diff
@@ -2069,6 +2062,7 @@ def _visit_For(self, expr, **settings):
             errors.report(INVALID_FOR_ITERABLE, symbol=expr.target,
                    bounding_box=(self._current_fst_node.lineno, self._current_fst_node.col_offset),
                    severity='error', blocker=self.blocking)
+        PyccelAstNode.stage = 'semantic'
```
MenoIy, line 2065: do you need to set the stage too semantic again here?  already done line 1996.

EmilyBourne: 1996 sets the stage to `syntactic`. This stage creates extra untreated nodes as they would appear after the syntactic stage. The only way I found to allow this to happen without it raising errors was to temporarily set the stage back to syntactic so the type checks are not carried out. This works as there are no `_visit` functions between line 1996 and here, but if you have a better solution that would a big improvement

MenoIy (approved):

pyccel/symbolic/lambdify.py
```diff
@@ -32,7 +31,7 @@ def cse(expr):
     (vars_old, stmts) = map(list, zip(*ls))
     vars_new = []
     free_gl = expr.free_symbols
-    free_gl.update(expr.atoms(IndexedBase))
+    #free_gl.update(expr.atoms(IndexedBase)) #What should this be instead?
```
yguclu, line 34: Can you explain this comment?

EmilyBourne: `expr.atoms(IndexedBase)` finds all the `IndexedBase` items in our code. We no longer have any of these items. I do not know enough about this function to know what should go here instead. I meant to ask @saidctb . Maybe it should find all objects with a `__getitem__` method?

yguclu: @saidctb Please have a look into this and create an issue if needed. Thanks!

yguclu (approved):

PR #708, opened 2021-01-25
Fix result type of floor division, math ceil function and the wrong result of math floor of negative float in C

### LINKED ISSUE #702: FloorDiv does not give an int in C
**Describe the bug**
The floor division `a//b` in C is translated to `floor(double(a)/double(b))` in order to preserve the correct rounding behaviour. However `floor` from `math.h` returns a `float`. This means that the floor division returns a floating point. This can potentially cause problems depending on how the result is used.
If the result is saved in a variable then there is no problem.
If the result is used as an index it causes a segmentation fault.
If the result is passed to the print function, the wrong result is printed

**To Reproduce**
For example:
```python
print(3//4)
```

Translates to:
```c
#include <stdlib.h>
#include <math.h>
#include <stdio.h>
int main()
{


    printf("%ld\n", floor(3.0 / 4.0));
    return 0;
}
```

which raises the warning:
```
/home/emily/.local/lib/python3.6/site-packages/pyccel/codegen/utilities.py:154: UserWarning: /home/emily/Code/test/tmp/__pyccel__/tmp.c: In function ‘main’:
/home/emily/Code/test/tmp/__pyccel__/tmp.c:8:15: warning: format ‘%ld’ expects argument of type ‘long int’, but argument 2 has type ‘double’ [-Wformat=]
     printf("%ld\n", floor(3.0 / 4.0));
             ~~^
             %f

  warnings.warn(UserWarning(output))
```
Executing the program gives the following result:
```
140728609725192
```

**Language**
C

### LINKED ISSUE #715: math.floor gives wrong result when passing negative float numbers.
**`math.floor` gives wrong values when passing negative float numbers.**

this is an example using interactive python for math.floor
```python
>> @types('real')
      def f1(x):
          from math import floor
           return floor(x)
>> cfloor = epyccel(f1, language='c')
>> cfloor(-3.2)
-3
>> math.floor(-3.2)
-4
```
We can see the problem in the conversion of this function
```python
@types('real')
def f1(x):
    from math import floor
     return floor(x)
```
the result of conversion with `pyccel `command gives:
```c
#include "t.h"
#include <stdlib.h>
#include <stdint.h>
#include <math.h>

int64_t f1(double x)
{
    int64_t Out_0001;

    Out_0001 = floor((int64_t)(x));
    return Out_0001;
}
```
here the cast is needed for casting the result of the floor, not the argument of floor function.
**Language**
C language

rakati:
Improve math function base print and the floor division print methods by fixing some bugs:
- Fix result type of `floor division` and  `math.ceil` function in C.
- Fix the wrong result of `math.floor` function when passing a negative float number.
- Add test cases for negative numbers.
- Add flags `-Werror -Wconversion` to check the conversion and the type of the results.
- Add one test of floor division with 3 operands. 
Fix #702
Fix #715


pyccel/codegen/printing/ccode.py
```diff
@@ -1070,11 +1070,18 @@ def _print_PyccelDiv(self, expr):
...
+        # the result type is float
+        need_to_cast = True
         if all(a.dtype is NativeInteger() for a in expr.args):
             args = [PythonFloat(a) for a in expr.args]
         else:
             args = expr.args
+            need_to_cast = True
         code = ' / '.join(self._print(a) for a in args)
+        if (need_to_cast):
+            return "(int64_t)floor({})".format(code)
```
nhamidn, line 1084: what if the first argument is an `int` but the second is 'real' ?

rakati: the result type is real.

pyccel/codegen/printing/ccode.py
```diff
@@ -1070,11 +1070,18 @@ def _print_PyccelDiv(self, expr):
...
         self._additional_imports.add("math")
+        # the result type of the floor division is dependent on the arguments
+        # type, if all arguments are integers the result is integer otherwise
+        # the result type is float
+        need_to_cast = True
         if all(a.dtype is NativeInteger() for a in expr.args):
             args = [PythonFloat(a) for a in expr.args]
         else:
             args = expr.args
+            need_to_cast = True
```
nhamidn, line 1081: do you mean that this must be `need_to_cast = True`

nhamidn: So we need to do casting only when the twp element are `int`

nhamidn (approved):

pyccel/codegen/printing/ccode.py
```diff
@@ -1070,11 +1070,13 @@ def _print_PyccelDiv(self, expr):
...
-        else:
-            args = expr.args
-        code = ' / '.join(self._print(a) for a in args)
+        # the result type of the floor division is dependent on the arguments
+        # type, if all arguments are integers the result is integer otherwise
+        # the result type is float
+        need_to_cast = all(a.dtype is NativeInteger() for a in expr.args)
+        code = ' / '.join(self._print(a if a.dtype is NativeReal() else PythonFloat(a)) for a in expr.args)
+        if (need_to_cast):
+            return "(int64_t)floor({})".format(code)
```
EmilyBourne, line 1079: > the result type of the floor division is dependent on the arguments type, if all arguments are integers the result is integer otherwise the result type is float

You then convert all the arguments to reals (to handle negative values (`-3//4`->`-1` but in c `-3/4`->`0`) ) so doesn't the result always need to be cast?

tests/epyccel/test_epyccel_division.py
```diff
@@ -145,6 +145,23 @@ def fdiv_i_i(x, y):
...
+    f = epyccel(fdiv_i_i_i, language=language)
+    x = randint(1e9)
+    y = randint(low=1, high= 1e3)
+    z = randint(low=1, high= 1e2)
+
+    assert (f(x, y, z) == fdiv_i_i_i(x, y, z))
+    assert (f(-x, y, z) == fdiv_i_i_i(-x, y, z))
+    assert (f(x, -y, z) == fdiv_i_i_i(x, -y, z))
+    assert (f(-x, -y, z) == fdiv_i_i_i(-x, -y, z))
+    assert isinstance(f(x, y, z), type(fdiv_i_i_i(x, y, z)))
```
EmilyBourne, line 164: I'm not sure this will be sufficient to test that we have the correct type as the returned type will always be whatever you declared it to be. Maybe there is a compiler warning that you could use to test this? E.g. `epyccel(fdiv_i_i_i, language=language, fflags='-Werror -Wconversion'`
See https://github.com/pyccel/pyccel/blob/master/tests/epyccel/test_kind.py#L34-L45

EmilyBourne:
Have you checked if `numpy.floor` and `math.floor` might have this problem too?

pyccel/codegen/printing/ccode.py
rakati: yes, you are right, we need to cast only one argument.

tests/epyccel/test_epyccel_division.py
rakati: `Wconversion` has been changed in the newer version of gcc ().By using the  flag, we will be obliged to cast all arguments to `double `, because `Wconversion`  checks also for mismatching of the arguments passed to function with the acceoted.

rakati: the [`Wconversion` ](https://gcc.gnu.org/wiki/NewWconversion)has been changed in the newer version of gcc (4.3 >=).
By using this flag, we will be obliged to cast all arguments to double, or we will mismatch in the type of arguments passed to the `floor` function.
```c
 double floor(double x)
```
so @EmilyBourne only one of your conversations will be solved.
what do you think?
for me, I see that we don't need this check, because the code will have a lot of casts and will not be more readable.

pyccel/codegen/printing/ccode.py
EmilyBourne: Never mind I've just realised I was reading this the wrong way round. I forgot that the return type of `//` depends on the argument types.

Presumably `need_to_cast` is equivalent to `expr.dtype is NativeInteger`?

tests/epyccel/test_epyccel_division.py
EmilyBourne: Hmmm, I'm not sure. You're right that the code will need cleaner if we have fewer casts, but I think this is an important bug which potentially causes big problems so maybe it is better to have better tests

rakati:
> Have you checked if `numpy.floor` and `math.floor` might have this problem too?

yes, for `math.floor`.
I find another bug in two, in the case of negative values.
I will try to add tests for negative values and fix these bugs in this PR.

pyccel/codegen/printing/ccode.py
```diff
@@ -893,26 +893,16 @@ def _print_MathFunctionBase(self, expr):
...
-                cast_func = python_builtin_datatypes[str_dtype(expr.dtype)]
-                args.append(self._print(cast_func(arg)))
+            if arg.dtype != NativeReal() and not func_name.startswith("pyc"):
+                args.append(self._print(PythonFloat(arg)))
             else:
                 args.append(self._print(arg))
         code_args = ', '.join(args)
+        if expr.dtype == NativeInteger():
+            cast_type = self.find_in_dtype_registry('int', default_precision['int'])
+            return '({0}){1}({2})'.format(cast_type, func_name, code_args)
```
EmilyBourne, line 903: Does this need doing for all math functions?

pyccel/codegen/printing/ccode.py
```diff
@@ -1096,11 +1086,14 @@ def _print_PyccelDiv(self, expr):
...
-        else:
-            args = expr.args
-        code = ' / '.join(self._print(a) for a in args)
+        # the result type of the floor division is dependent on the arguments
+        # type, if all arguments are integers the result is integer otherwise
+        # the result type is float
+        need_to_cast = all(a.dtype is NativeInteger() for a in expr.args)
+        code = ' / '.join(self._print(a if a.dtype is NativeReal() else PythonFloat(a)) for a in expr.args)
+        if (need_to_cast):
+            cast_type = self.find_in_dtype_registry('int', default_precision['int'])
```
EmilyBourne, line 1095: I wonder if it would be worth using a function like `_cast_to` (see PR #714 ) to cast the result to the correct type if necessary rather than just handling native integers? I have run a quick test on my computer and FloorDiv works for numpy types too which means that you may also want one of the following:
`(float)floor(a/b)`
`(int32_t)floor(a/b)`

tests/epyccel/test_epyccel_division.py
```diff
@@ -137,50 +137,80 @@ def test_call_fdiv_i_i(language):
...
+    fflags = "-Werror -Wconversion"
+
+    f = epyccel(fdiv_r_r, language=language, fflags=fflags)
     x = uniform(high=1e9)
     y = uniform(low=1e-14, high= 1e3)
     assert (f(x, y) == fdiv_r_r(x, y))
     assert (f(-x, y) == fdiv_r_r(-x, y))
     assert (f(x, -y) == fdiv_r_r(x, -y))
     assert (f(-x, -y) == fdiv_r_r(-x, -y))
+    assert isinstance(f(x, y), type(fdiv_r_r(x, y)))
```
EmilyBourne, line 216: Given my comment above I want to ask for 1 or 2 tests with numpy types.

pyccel/codegen/printing/ccode.py
rakati: all maths functions accept real numbers in their arguments and return real numbers as result.

EmilyBourne: ok perfect :+1:

pyccel/codegen/printing/ccode.py
rakati: No, the situation here is different, I am not casting the expression I am casting the result of the expression with an added prefix (the function name).

EmilyBourne: I'm not sure I see what you mean here.
If I do:
```
> import numpy as np
> a = np.int32(-3)
> b = np.int32(4)
> a//b
-1
> type(a//b)
np.int32
```
Won't this be translated by pyccel to:
```
int32_t a;
int32_t b;
int32_t c;
a=-3;
b=4;
c = (int64_t)floor(-3/4); // implicit-conversion
```
?
It's a much less critical problem as the type is right but it will lead to the result being cast twice when you assign it to a variable. Or have I missed something?

tests/epyccel/test_epyccel_division.py
rakati: mm, I don't know what you mean exactly.
do you mean using numpy types in the random values, or using precise types like (int8, int16 ...) in the types decorate?.

EmilyBourne: Both. Using precise types in the types decorator, which then means you need to pass random values with the matching type

pyccel/codegen/printing/ccode.py
rakati: you give a great example, I need to cast to expression precision not default precision, I forgot that.
to explain more why I can't use `_cast_to` function:
I need to add cast after adding the function `(cast)func_name([expr.args])`, the _cast_to normally used to cast an expression directly, not a string, like in my case.

tests/epyccel/test_epyccel_division.py
rakati: ok, the same as what I did with numpy array as function arguments test.

pyccel/codegen/printing/ccode.py
EmilyBourne: Ah ok that's what you meant. I think a good improvement to the `_cast_to` function would be for it to take a string in input too or to return a statement such as `'(cast){}'` that way it could be used in a lot more places. That's why I said a function _like_ `_cast_to` ;-).

If you and @saniQQ can come up with a function that works for both of you then that is probably the most reusable version. The `_cast_to` has the advantage that it handles the test of whether the cast is necessary

EmilyBourne (approved):

yguclu (approved):

Issue #690, opened 2021-01-20
Implement numpy.arange


EmilyBourne:
**Describe the feature**
We would like [np.arange](https://numpy.org/doc/stable/reference/generated/numpy.arange.html) to be implemented in pyccel.

**Acceptance criteria**
- [ ] Add new class for `numpy.arange` nodes (inherits from `NumpyNewArray`)
- [ ] Add unit test

PR #711, opened 2021-01-26
Add numpy arange

MenoIy:
Solve #690:
- Create class NumpyArange;
- Add printer for NumpyArange in C and Fortran printers;
- Add tests.

### DIFF
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -15,7 +15,7 @@ from .core           import (ClassDef, FunctionDef,
 
 from .internals      import PyccelInternalFunction
 
-from .operators      import broadcast, PyccelMinus, PyccelAdd
+from .operators      import broadcast, PyccelMinus, PyccelDiv
 
 from .builtins       import (PythonInt, PythonBool, PythonFloat, PythonTuple,
                              PythonComplex, PythonReal, PythonImag, PythonList)
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -183,17 +183,18 @@ class NumpyArange(NumpyNewArray):
 
     Parameters
     ----------
-        stop : Number
-            End of interval.
+    start : Numeric
+        Start of interval, default value 0
 
-        start : Number
-            Start of interval, default value 0
+    stop : Numeric
+        End of interval
 
-        step : Number default 1
-            Spacing between values, default value 1
+    step : Numeric
+        Spacing between values, default value 1
 
-        dtype : Datatype
-            The type of the output array
+    dtype : Datatype
+        The type of the output array, if dtype is not given,
+        infer the data type from the other input arguments.
     """
 
     def __init__(self, start, stop = None, step = None, dtype = None):
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -82,7 +82,8 @@ __all__ = (
     'Shape',
     'NumpyWhere',
     'NumpyZeros',
-    'NumpyZerosLike'
+    'NumpyZerosLike',
+    'NumpyArange'
 )
 
 #==============================================================================
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -180,11 +180,19 @@ class NumpyArange(NumpyNewArray):
     """
     Represents a call to  numpy.arange for code generation.
 
-    arg :
+    Parameters
+    ----------
         stop : Number
-        start : Number default 0
+            End of interval.
+
+        start : Number
+            Start of interval, default value 0
+
         step : Number default 1
+            Spacing between values, default value 1
+
         dtype : Datatype
+            The type of the output array
     """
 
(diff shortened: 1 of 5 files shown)


mostafaMamoni (approved):
Looks fine to me, but is it possible to make `_print_assign` as in fcode

pyccel/ast/numpyext.py
```diff
@@ -173,6 +175,67 @@ def _sympystr(self, printer):
     def arg(self):
         return self._arg
 
+#==============================================================================
+class NumpyArange(NumpyNewArray):
```
EmilyBourne, line 179: Could this inherit from `PythonRange` and `NumpyNewArray` to avoid re-implementing some of the `PythonRange` properties? Or does this create some kind of problematic diamond shaped class inheritance?

pyccel/ast/numpyext.py
```diff
@@ -173,6 +175,67 @@ def _sympystr(self, printer):
...
+
+        step : Number default 1
+            Spacing between values, default value 1
+
+        dtype : Datatype
+            The type of the output array
+    """
+
+    def __init__(self, start, stop = None, step = None, dtype = None):
+        from .mathext import MathCeil
```
EmilyBourne, line 199: Why can't this go at the top of the file with the other imports?

pyccel/codegen/printing/ccode.py
```diff
@@ -288,6 +288,32 @@ def _traverse_matrix_indices(self, mat):
...
 
+    #========================== Numpy Elements ===============================#
+    def print_NumpyArange(self, expr, lhs):
+        start  = self._print(expr.start)
+        stop   = self._print(expr.stop)
+        step   = self._print(expr.step)
+        dtype  = self.find_in_ndarray_type_registry(self._print(expr.dtype), expr.precision)
+
+        target = Variable(expr.dtype, name =  self._parser.get_new_name('s'))
+        index  = ValuedVariable(NativeInteger(), name = self._parser.get_new_name('i'), value=0)
```
EmilyBourne, line 299: Why do you need a `ValuedVariable`? These are usually used for arguments with a default value not assignments

MenoIy:
> Looks fine to me, but is it possible to make `_print_assign` as in fcode

I'd love to do that but is not like Fortran in C we need the Lhs of the assign in the loop.

pyccel/ast/numpyext.py
MenoIy: hmm, I don't think it can, at least if the PythonRange class is changed just as you did in your PR

EmilyBourne: Why not? I would have thought that would help if anything?

MenoIy: I'm not that good at inheritance, but pythonRange property returns elements stored in self._args, and in arange i don't do that.

EmilyBourne (approved):

pyccel/codegen/printing/fcode.py
```diff
@@ -741,6 +741,29 @@ def _print_NumpyFloor(self, expr):
         result_code = self._print_MathFloor(expr)
         return 'real({}, {})'.format(result_code, self.print_kind(expr))
 
+    def _print_NumpyArange(self, expr):
+        start  = self._print(expr.start)
+        step   = self._print(expr.step)
+        shape  = PyccelMinus(expr.shape[0], LiteralInteger(1))
+        target = Variable(NativeInteger(), name =  self.parser.get_new_name('i'))
```
yguclu, line 748: Why do you call this variable `target`? Is it not an index?

MenoIy: i just followed the same naming we do in For printer

pyccel/ast/numpyext.py
```diff
@@ -173,6 +176,66 @@ def _sympystr(self, printer):
...
+            End of interval.
+
+        start : Number
+            Start of interval, default value 0
+
+        step : Number default 1
+            Spacing between values, default value 1
+
+        dtype : Datatype
+            The type of the output array
```
yguclu, line 196: I have a few comments about this docstring:

- Please reorder the parameters with `start` before `stop`.

- Also, there is no need to indent them:
```
Parameters
----------
param : type
    Description.
```
- Is `Number` really a type?

- Can you be a bit more consistent in the way that you give the default value?

pyccel/codegen/printing/fcode.py
yguclu: I don't know when the `For` printer was written, but I doubt that it is a model of style.

pyccel/ast/numpyext.py
MenoIy: Can you be a bit more consistent in the way that you give the default values?
can you explain more what do you mean by this

yguclu: For `start` you write the default value once, for `step` you write it twice, for `dtype` you don't write it at all.

MenoIy: ah true sorry

yguclu (approved):

yguclu:
Very good job @MenoIy! Thanks

Issue #238, opened 2020-02-20
Support logical indexing in numpy


krystophny:
It would be nice if one could support jumpy logical indexing of arrays with boolean index arrays, e.g.
```
c = (a>0)
a[c] = 3
```

PR #714, opened 2021-01-27
Add casting to get_index when needed.

saniQQ:
- Add func `_cast_to` to help with casting to the correct type when needed;
- Raise error when passing non integer/slice to an `indexedElement`;
- Add tests for indexing with `int64` and `int32`.

### DIFF
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -843,11 +843,11 @@ class CCodePrinter(CodePrinter):
         return "%s.%s[get_index(%s, %s)]" % (base_name, dtype, base_name, ", ".join(inds))
 
     def _cast_to(self, expr, dtype, precision):
-        """ add cast to an expresion when needed
+        """ add cast to an expression when needed
         parameters
         ----------
             expr      : PyccelAstNode
-                the expression to be casted
+                the expression to be cast
             dtype     : Datatype
                 base type of the cast
             precision : integer
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -722,7 +722,7 @@ class CCodePrinter(CodePrinter):
                         inds[i] = Slice(ind, PyccelAdd(ind, LiteralInteger(1)), LiteralInteger(1))
                 inds = [self._print(i) for i in inds]
                 return "array_slicing(%s, %s, %s)" % (base_name, expr.rank, ", ".join(inds))
-            inds = [self._cast_to(i, NativeInteger(), 8) for i in inds]
+            inds = [self._cast_to(i, NativeInteger(), 8).format(self._print(i)) for i in inds]
         else:
             raise NotImplementedError(expr)
         return "%s.%s[get_index(%s, %s)]" % (base_name, dtype, base_name, ", ".join(inds))
@@ -731,9 +731,9 @@ class CCodePrinter(CodePrinter):
         """ add cast to an expresion when needed
         parameters
         ----------
-            expr :
+            expr      : PyccelAstNode
                 the expression to be casted
-            dtype : Datatype
+            dtype     : Datatype
                 base type of the cast
             precision : integer
                 precision of the base type of the cast
@@ -744,10 +744,9 @@ class CCodePrinter(CodePrinter):
                 Return string that contains the expression casted to the desired type
         """
         if (expr.dtype != dtype or expr.precision != precision):
-            cast=self.find_in_dtype_registry(self._print(expr.dtype), precision)
-            expr = self._print(expr)
-            return '({cast}){expr}'.format(cast=cast, expr=expr)
-        return self._print(expr)
+            cast=self.find_in_dtype_registry(self._print(dtype), precision)
+            return '({}){{}}'.format(cast)
+        return '{}'
     def _print_DottedVariable(self, expr):
         """convert dotted Variable to their C equivalent"""
         return '{}.{}'.format(self._print(expr.lhs), self._print(expr.name))
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -741,7 +741,7 @@ class CCodePrinter(CodePrinter):
         Return
         ------
             String
-                Return string that contains the expression casted to the desired type
+                Return format string that contains the desired cast type
         """
         if (expr.dtype != dtype or expr.precision != precision):
             cast=self.find_in_dtype_registry(self._print(dtype), precision)
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -728,7 +728,7 @@ class CCodePrinter(CodePrinter):
         return "%s.%s[get_index(%s, %s)]" % (base_name, dtype, base_name, ", ".join(inds))
 
     def _cast_to(self, expr, dtype, precision):
-        """ add cast to an expresion
+        """ add cast to an expresion when needed
         parameters
         ----------
             expr :
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -696,11 +696,17 @@ class CCodePrinter(CodePrinter):
                         inds[i] = Slice(ind, PyccelAdd(ind, LiteralInteger(1)), LiteralInteger(1))
                 inds = [self._print(i) for i in inds]
                 return "array_slicing(%s, %s, %s)" % (base_name, expr.rank, ", ".join(inds))
-            inds = [self._print(i) for i in inds]
+            inds = [self._cast_to(i, NativeInteger(), 8) for i in inds]
         else:
             raise NotImplementedError(expr)
         return "%s.%s[get_index(%s, %s)]" % (base_name, dtype, base_name, ", ".join(inds))
 
+    def _cast_to(self, expr, dtype, precision):
+        if (expr.dtype == dtype and expr.precision != precision):
+            cast=self.find_in_dtype_registry(self._print(expr.dtype), precision)
+            expr = self._print(expr)
+            return '({cast}){expr}'.format(cast=cast, expr=expr)
+        return self._print(expr)
     def _print_DottedVariable(self, expr):
         """convert dotted Variable to their C equivalent"""
         return '{}.{}'.format(self._print(expr.lhs), self._print(expr.name))
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -728,7 +728,22 @@ class CCodePrinter(CodePrinter):
         return "%s.%s[get_index(%s, %s)]" % (base_name, dtype, base_name, ", ".join(inds))
 
     def _cast_to(self, expr, dtype, precision):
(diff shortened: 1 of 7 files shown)


mostafaMamoni (approved):

pyccel/codegen/printing/ccode.py
```diff
@@ -696,11 +696,17 @@ def _print_IndexedElement(self, expr):
...
                 inds = [self._print(i) for i in inds]
                 return "array_slicing(%s, %s, %s)" % (base_name, expr.rank, ", ".join(inds))
-            inds = [self._print(i) for i in inds]
+            inds = [self._cast_to(i, NativeInteger(), 8) for i in inds]
         else:
             raise NotImplementedError(expr)
         return "%s.%s[get_index(%s, %s)]" % (base_name, dtype, base_name, ", ".join(inds))
 
+    def _cast_to(self, expr, dtype, precision):
+        if (expr.dtype == dtype and expr.precision != precision):
```
EmilyBourne, line 705: Shouldn't this be:
`(expr.dtype != dtype or expr.precision != precision)`
?

saniQQ: i wanted the function to be unable to cast to a different type. (float -> int)

EmilyBourne: That makes sense for indexing arguments, but I'm not sure it makes sense for a function called `_cast_to`.

I am thinking about cases where I might re-use this later. If I did `cast_to(obj, Real, 4)` where `obj` is an integer, I would be very surprised if I got back `expr`.

Maybe the function should be renamed? `_choose_precision`? `_enforce_precision`?
Or maybe we should check that the argument is an integer elsewhere and then leave this as a standard casting function?

What do you think?

pyccel/codegen/printing/ccode.py
```diff
@@ -696,11 +696,17 @@ def _print_IndexedElement(self, expr):
                         inds[i] = Slice(ind, PyccelAdd(ind, LiteralInteger(1)), LiteralInteger(1))
                 inds = [self._print(i) for i in inds]
                 return "array_slicing(%s, %s, %s)" % (base_name, expr.rank, ", ".join(inds))
-            inds = [self._print(i) for i in inds]
+            inds = [self._cast_to(i, NativeInteger(), 8) for i in inds]
         else:
             raise NotImplementedError(expr)
         return "%s.%s[get_index(%s, %s)]" % (base_name, dtype, base_name, ", ".join(inds))
 
+    def _cast_to(self, expr, dtype, precision):
```
EmilyBourne, line 704: docstring

pyccel/codegen/printing/ccode.py
saniQQ: if we can unsure the arguments to be integers we can leave it as a standard casting function.

EmilyBourne: I tested and it didn't raise an error :/ so it probably should be added as a warning or an error in the semantic stage. We usually say that pyccel only works on valid python code though

saniQQ: yeah it doesn't raise the error , should all `indexedElement` raise an error if the arguments are not integers?

EmilyBourne: I think that would be a good idea. In theory numpy arrays can be indexed with booleans too (see issue #238) so it is important that we raise an error if the index is something that we don't handle

pyccel/codegen/printing/ccode.py
```diff
@@ -722,11 +722,32 @@ def _print_IndexedElement(self, expr):
...
+                base type of the cast
+            precision : integer
+                precision of the base type of the cast
+
+        Return
+        ------
+            String
+                Return string that contains the expression casted to the desired type
+        """
+        if (expr.dtype != dtype and expr.precision != precision):
```
EmilyBourne, line 746: `and`->`or`?

pyccel/parser/semantic.py
```diff
@@ -944,6 +944,12 @@ def _extract_indexed_from_var(self, var, args, name):
                 bounding_box=(self._current_fst_node.lineno, self._current_fst_node.col_offset),
                 severity='error', blocker=self.blocking)
 
+        for arg in var[args]._indices:
+            if not isinstance(arg, (Slice, LiteralInteger)) and not \
+                (isinstance(arg, Variable) and isinstance(arg.dtype, (NativeBool, NativeInteger))):
```
EmilyBourne, line 949: Why does the index need to be a `Variable`. Can we not do things like:
`a[i+j]` ? As long as we have the right dtype we should be fine

EmilyBourne, line 949: We want to support boolean indexing in the future but I don't think it is supported at all currently

EmilyBourne (approved):

pyccel/codegen/printing/ccode.py
```diff
@@ -722,11 +722,32 @@ def _print_IndexedElement(self, expr):
...
+        Return
+        ------
+            String
+                Return string that contains the expression casted to the desired type
+        """
+        if (expr.dtype != dtype or expr.precision != precision):
+            cast=self.find_in_dtype_registry(self._print(expr.dtype), precision)
+            expr = self._print(expr)
+            return '({cast}){expr}'.format(cast=cast, expr=expr)
+        return self._print(expr)
```
rakati, line 750: I see that this method doesn't always add cast the expression like what you explain in the docstring.
can you update the docstring desc?
before that, I see that this function serves only this [line-725](https://github.com/pyccel/pyccel/pull/714/files#diff-f1038cb7408f08867011c5feb494cd9e7a5a435a3b6f6438ef6cd3bdf3c90fbeR725), can you explain more why we need this function, and where it can help us.

saniQQ: having a function that does the casting is better than having duplicated code.

EmilyBourne: In lots of the places where we need to cast we cannot use `_print(expr)` (e.g. when casting `floor`, calling `_print(expr)` would cause an infinite loop).
I wonder if we could make this more general, eg by returning a format string

saniQQ: Does making the function take the expr string, type, and precision as arguments fixing this? because passing the AstNode will result in an infinite loop.

EmilyBourne: Passing the `AstNode` only leads to an infinite loop if `_print(expr)` is called. Making it take a string in place of `expr` would fix this, but you can no longer do the important test `if (expr.dtype != dtype or expr.precision != precision):`
So you would have to pass `expr` and `str_expr`.

saniQQ: is there a way to do it without passing `expr` and `expr_str` at the same time? i can't think of a way.

EmilyBourne: As I mentioned above, this could be done by returning a format string. I.e:
```
if ...
    ...
    return '({cast}){{}}'
else:
    return '{}'

But you still need to manage both in the calling function

Comment (saniQQ) [reply pyccel/codegen/printing/ccode.py]: wouldn't this make it easier to forget to add to the code.

Comment (saniQQ) [reply pyccel/codegen/printing/ccode.py]: even if passing the `expr` and `expr_str` is a little bit odd but it will make it easier to not make a mistake.

Comment (EmilyBourne) [reply pyccel/codegen/printing/ccode.py]: Hmm, I hadn't considered that. Possibly. Would it not raise an error in C? Or will it default to 0?
@rakati do you have any suggestions?

Comment (EmilyBourne) [reply pyccel/codegen/printing/ccode.py]: You could make `expr_str` an optional parameter which defaults to `self._print(expr)`

Comment (saniQQ) [reply pyccel/codegen/printing/ccode.py]: will this not the same as calling `self._print(expr)` inside the function?

Comment (saniQQ) [reply pyccel/codegen/printing/ccode.py]: with returning the format i think its oket unless you want to use `_cast_to` in a complex format string as the line will be too long a call to `_cast_to` will be like `self._cast_to(i, NativeInteger(), 8).format(self._print(i))`

Comment (EmilyBourne) [reply pyccel/codegen/printing/ccode.py]: If it is misused then yes, but this will easily be spotted. In cases where an infinite loop would occur, we would need to make sure that we provided `expr_str`

Comment (EmilyBourne) [reply pyccel/codegen/printing/ccode.py]: > with returning the format I think it's okay unless you want to use _cast_to in a complex format string as the line will be too long.

You can always split the line if it becomes too long by saving the format string in a variable. In any case passing it as an argument will give almost the same line length

Comment (saniQQ) [reply pyccel/codegen/printing/ccode.py]: which one seems more intuitive?

Comment (EmilyBourne) [reply pyccel/codegen/printing/ccode.py]: Personally I don't have a preference

Comment (saniQQ) [reply pyccel/codegen/printing/ccode.py]: can we even use `self._print` as optional argument ??
i got self is not defined error.

File: pyccel/codegen/printing/ccode.py [hunk 6]
```diff
@@ -744,10 +744,9 @@ def _cast_to(self, expr, dtype, precision):
                 Return string that contains the expression casted to the desired type
```
EmilyBourne, line 744: Don't forget to update the description of the return type

pyccel/codegen/printing/ccode.py
EmilyBourne:
```
def _cast_to(self, expr, dtype, precision, expr_str = None):
    if expr_str is None:
        expr_str = self._print(expr)
```

EmilyBourne (approved):

yguclu:
I am waiting for the approval of @rakati before merging this.

rakati (approved):

pyccel/codegen/printing/ccode.py
```diff
@@ -837,11 +837,31 @@ def _print_IndexedElement(self, expr):
...
                 inds = [self._print(i) for i in inds]
                 return "array_slicing(%s, %s, %s)" % (base_name, expr.rank, ", ".join(inds))
-            inds = [self._print(i) for i in inds]
+            inds = [self._cast_to(i, NativeInteger(), 8).format(self._print(i)) for i in inds]
         else:
             raise NotImplementedError(expr)
         return "%s.%s[get_index(%s, %s)]" % (base_name, dtype, base_name, ", ".join(inds))
 
+    def _cast_to(self, expr, dtype, precision):
+        """ add cast to an expresion when needed
```
yguclu, line 846: Typo: expresion -> expression (note the missing `s`)

pyccel/codegen/printing/ccode.py
```diff
@@ -837,11 +837,31 @@ def _print_IndexedElement(self, expr):
...
+        parameters
+        ----------
+            expr      : PyccelAstNode
+                the expression to be casted
```
yguclu, line 850: I believe that past participle of "cast" is "cast". It is an irregular verb.

yguclu (changes requested):
Good job @saniQQ! Do you mind fixing a couple of typos?

pyccel/codegen/printing/ccode.py
EmilyBourne: > I believe

You are correct

yguclu (approved):

PR #716, opened 2021-01-29
Drop Sympy's Symbol

MenoIy:
- Create class PyccelSymbol
- Create functions symbols that return a tuple of instances of PyccelSymbol
- Change imports in all code
- Remove function allocatable_like  not used anymore.
- Remove unnecessary cast to a string.
- Add magic methods ```__eq__``` and  ```__hash__``` to Variable and DottedVariable

### DIFF
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -2493,7 +2493,7 @@ class ClassDef(Basic):
 
         if isinstance(name, str):
             name = PyccelSymbol(name)
-        elif not isinstance(name, PyccelSymbol):
+        else:
             raise TypeError('Function name must be PyccelSymbol or string')
 
         # attributes
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -8,6 +8,7 @@
 from collections     import OrderedDict
 
 from sympy import sympify
+from sympy import Symbol as sp_Symbol
 from sympy import Add as sp_Add, Mul as sp_Mul, Pow as sp_Pow
 from sympy import Eq as sp_Eq, Ne as sp_Ne, Lt as sp_Lt, Le as sp_Le, Gt as sp_Gt, Ge as sp_Ge
 from sympy import Integer as sp_Integer
@@ -128,11 +129,11 @@ __all__ = (
 
 #==============================================================================
 local_sympify = {
-    'N'    : Symbol('N'),
-    'S'    : Symbol('S'),
-    'zeros': Symbol('zeros'),
-    'ones' : Symbol('ones'),
-    'Point': Symbol('Point')
+    'N'    : sp_Symbol('N'),
+    'S'    : sp_Symbol('S'),
+    'zeros': sp_Symbol('zeros'),
+    'ones' : sp_Symbol('ones'),
+    'Point': sp_Symbol('Point')
 }
 
 # TODO - add EmptyStmt => empty lines
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -350,7 +350,7 @@ class AsName(Basic):
         return '{0} as {1}'.format(sstr(self.name), sstr(self.target))
 
     def __eq__(self, string):
-        if isinstance(string, (str, PyccelSymbol)):
+        if isinstance(string, str):
             return string == self.target
         else:
             return self is string
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -36,7 +36,7 @@ from .builtins  import (PythonEnumerate, PythonLen, PythonList, PythonMap,
 from .datatypes import (datatype, DataType, NativeSymbol,
                         NativeBool, NativeRange,
                         NativeTuple, is_iterable_datatype, str_dtype)
-from .internals      import Slice, Symbol
+from .internals      import Slice, PyccelSymbol
 
 from .literals       import LiteralTrue, LiteralFalse, LiteralInteger, Nil
 from .literals       import LiteralImaginaryUnit, LiteralString
@@ -243,7 +243,7 @@ def extract_subexpressions(expr):
              sp_Or, sp_Eq, sp_Ne, sp_Lt, sp_Gt,
              sp_Le, sp_Ge)
 
-    id_cls = (Symbol, IndexedElement,
+    id_cls = (PyccelSymbol, IndexedElement,
               DottedVariable, sp_Float, sp_Integer,
               sp_Rational, LiteralImaginaryUnit,sp_Boolean,
               LiteralTrue, LiteralFalse, LiteralString,
@@ -393,7 +393,7 @@ def create_incremented_string(forbidden_exprs, prefix = 'Dummy', counter = 1):
 
 def create_variable(forbidden_names, prefix = None, counter = 1):
     """This function takes a prefix and a counter and uses them to construct
-    a Symbol with a name of the form:
+    a PyccelSymbol with a name of the form:
             prefix_counter
     Where counter is formatted to fill 4 characters
     The new name is checked against a list of forbidden expressions. If the
@@ -411,8 +411,8 @@ def create_variable(forbidden_names, prefix = None, counter = 1):
 
       Returns
       ----------
-      name            : Symbol
-                        A Symbol with the incremented string name
+      name            : PyccelSymbol
+                        A PyccelSymbol with the incremented string name
       counter         : int
                         The expected value of the next name
 
@@ -420,7 +420,7 @@ def create_variable(forbidden_names, prefix = None, counter = 1):
 
     name, counter = create_incremented_string(forbidden_names, prefix, counter = counter)
 
-    return Symbol(name), counter
+    return PyccelSymbol(name), counter
 
 
 class AsName(Basic):
@@ -460,7 +460,7 @@ class AsName(Basic):
         return '{0} as {1}'.format(sstr(self.name), sstr(self.target))
 
     def __eq__(self, string):
-        if isinstance(string, (str, Symbol)):
+        if isinstance(string, (str, PyccelSymbol)):
             return string == self.target
         else:
             return self is string
@@ -508,7 +508,7 @@ class Assign(Basic):
         In the syntactic stage:
            Object representing the lhs of the expression. These should be
            singular objects, such as one would use in writing code. Notable types
-           include Symbol, and IndexedElement. Types that
+           include PyccelSymbol, and IndexedElement. Types that
            subclass these types are also supported.
         In the semantic stage:
            Variable or IndexedElement
@@ -592,7 +592,7 @@ class Assign(Basic):
         rhs = self.rhs
         cond = isinstance(rhs, Variable) and rhs.rank > 0
         cond = cond or isinstance(rhs, IndexedElement)
-        cond = cond and isinstance(lhs, Symbol)
+        cond = cond and isinstance(lhs, PyccelSymbol)
         cond = cond or isinstance(rhs, Variable) and rhs.is_pointer
         return cond
 
@@ -606,12 +606,12 @@ class Assign(Basic):
         rhs = self.rhs
         if isinstance(lhs, Variable):
             return isinstance(lhs.dtype, NativeSymbol)
-        elif isinstance(lhs, Symbol):
+        elif isinstance(lhs, PyccelSymbol):
             if isinstance(rhs, PythonRange):
                 return True
             elif isinstance(rhs, Variable):
                 return isinstance(rhs.dtype, NativeSymbol)
-            elif isinstance(rhs, Symbol):
+            elif isinstance(rhs, PyccelSymbol):
                 return True
 
         return False
@@ -789,9 +789,9 @@ class AliasAssign(Basic):
 
     Parameters
     ----------
-    lhs : Symbol
+    lhs : PyccelSymbol
         at this point we don't know yet all information about lhs, this is why a
-        Symbol is the appropriate type.
+        PyccelSymbol is the appropriate type.
 
     rhs : Variable, IndexedElement
         an assignable variable can be of any rank and any datatype, however its
@@ -799,12 +799,12 @@ class AliasAssign(Basic):
 
     Examples
     --------
-    >>> from pyccel.ast.internals import Symbol
+    >>> from pyccel.ast.internals import PyccelSymbol
     >>> from pyccel.ast.core import AliasAssign
     >>> from pyccel.ast.core import Variable
     >>> n = Variable('int', 'n')
     >>> x = Variable('int', 'x', rank=1, shape=[n])
-    >>> y = Symbol('y')
+    >>> y = PyccelSymbol('y')
     >>> AliasAssign(y, x)
 
     """
@@ -841,17 +841,17 @@ class SymbolicAssign(Basic):
 
     Parameters
     ----------
-    lhs : Symbol
+    lhs : PyccelSymbol
 
     rhs : Range
 
     Examples
     --------
-    >>> from pyccel.ast.internals import Symbol
+    >>> from pyccel.ast.internals import PyccelSymbol
     >>> from pyccel.ast.core import SymbolicAssign
     >>> from pyccel.ast.core import Range
     >>> r = Range(0, 3)
-    >>> y = Symbol('y')
+    >>> y = PyccelSymbol('y')
     >>> SymbolicAssign(y, r)
 
     """
@@ -932,7 +932,7 @@ class AugAssign(Assign):
         In the syntactic stage:
            Object representing the lhs of the expression. These should be
            singular objects, such as one would use in writing code. Notable types
-           include Symbol, and IndexedElement. Types that
+           include PyccelSymbol, and IndexedElement. Types that
            subclass these types are also supported.
         In the semantic stage:
            Variable or IndexedElement
@@ -1008,9 +1008,9 @@ class While(Basic):
 
     Examples
     --------
-    >>> from pyccel.ast.internals import Symbol
+    >>> from pyccel.ast.internals import PyccelSymbol
     >>> from pyccel.ast.core import Assign, While
-    >>> n = Symbol('n')
+    >>> n = PyccelSymbol('n')
     >>> While((n>1), [Assign(n,n-1)])
     While(n > 1, (n := n - 1,))
     """
@@ -1671,7 +1671,7 @@ class ValuedArgument(Basic):
 
         if isinstance(expr, Argument):
             self._name = expr.name
-        elif isinstance(expr, (str, Symbol)):
+        elif isinstance(expr, (str, PyccelSymbol)):
             self._name = expr
         else:
             raise TypeError('Expecting an argument')
@@ -1969,19 +1969,19 @@ class FunctionDef(Basic):
         doc_string=None):
 
         if isinstance(name, str):
-            name = Symbol(name)
+            name = PyccelSymbol(name)
         elif isinstance(name, (tuple, list)):
             name_ = []
             for i in name:
                 if isinstance(i, str):
-                    name = name + Symbol(i)
-                elif not isinstance(i, Symbol):
-                    raise TypeError('Function name must be Symbol or string'
+                    name = name + PyccelSymbol(i)
+                elif not isinstance(i, PyccelSymbol):
+                    raise TypeError('Function name must be PyccelSymbol or string'
                                     )
             name = tuple(name_)
-        elif not isinstance(name, Symbol):
+        elif not isinstance(name, PyccelSymbol):
 
-            raise TypeError('Function name must be Symbol or string')
+            raise TypeError('Function name must be PyccelSymbol or string')
 
         # arguments
 
@@ -2597,9 +2597,9 @@ class ClassDef(Basic):
         # name
 
         if isinstance(name, str):
-            name = Symbol(name)
-        elif not isinstance(name, Symbol):
-            raise TypeError('Function name must be Symbol or string')
+            name = PyccelSymbol(name)
+        elif not isinstance(name, PyccelSymbol):
+            raise TypeError('Function name must be PyccelSymbol or string')
 
         # attributes
 
@@ -2826,7 +2826,7 @@ class Import(Basic):
         self._source = source
         self._target = []
         self._ignore_at_print = ignore_at_print
-        if isinstance(target, (str, Symbol, DottedName, AsName)):
+        if isinstance(target, (str, PyccelSymbol, DottedName, AsName)):
             self._target = [Import._format(target)]
         elif iterable(target):
             for i in target:
@@ -2839,13 +2839,13 @@ class Import(Basic):
             if '.' in i:
                 return DottedName(*i.split('.'))
             else:
-                return Symbol(i)
+                return PyccelSymbol(i)
         if isinstance(i, (DottedName, AsName)):
             return i
-        elif isinstance(i, Symbol):
+        elif isinstance(i, PyccelSymbol):
             return i
         else:
-            raise TypeError('Expecting a string, Symbol DottedName, given {}'.format(type(i)))
+            raise TypeError('Expecting a string, PyccelSymbol DottedName, given {}'.format(type(i)))
 
     @property
     def target(self):
@@ -3365,9 +3365,9 @@ class IfSection(Basic):
 
     Examples
     --------
-    >>> from pyccel.ast.internals import Symbol
+    >>> from pyccel.ast.internals import PyccelSymbol
     >>> from pyccel.ast.core import Assign, IfSection, CodeBlock
-    >>> n = Symbol('n')
+    >>> n = PyccelSymbol('n')
     >>> IfSection((n>1), CodeBlock([Assign(n,n-1)]))
     IfSection((n>1), CodeBlock([Assign(n,n-1)]))
     """
@@ -3411,9 +3411,9 @@ class If(Basic):
 
     Examples
     --------
-    >>> from pyccel.ast.internals import Symbol
+    >>> from pyccel.ast.internals import PyccelSymbol
     >>> from pyccel.ast.core import Assign, If
-    >>> n = Symbol('n')
+    >>> n = PyccelSymbol('n')
     >>> i1 = IfSection((n>1), [Assign(n,n-1)])
     >>> i2 = IfSection(True, [Assign(n,n+1)])
     >>> If(i1, i2)
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -1871,12 +1871,11 @@ class FunctionDef(Basic):
             for i in name:
                 if isinstance(i, str):
                     name_.append(PyccelSymbol(i))
-                else
+                else:
                     raise TypeError('Function name must be PyccelSymbol or string'
                                     )
             name = tuple(name_)
-        elif not isinstance(name, PyccelSymbol):
-
+        else:
             raise TypeError('Function name must be PyccelSymbol or string')
 
         # arguments
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -1792,9 +1792,6 @@ class FunctionCall(PyccelAstNode):
     def interface_name(self):
         return self._interface_name
 
-    def __str__(self):
-        return '{}({})'.format(self.func_name, ', '.join(str(a) for a in self.args))
-
     def __repr__(self):
         return '{}({})'.format(self.func_name, ', '.join(str(a) for a in self.args))
 
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -1569,7 +1569,7 @@ class ValuedArgument(Basic):
 
         if isinstance(value, (bool, int, float, complex, str)):
             value = convert_to_literal(value)
-        elif not isinstance(value, (Basic, Symbol)):
+        elif not isinstance(value, (Basic, PyccelSymbol)):
             raise TypeError("Expecting a pyccel object not {}".format(type(value)))
 
         if not isinstance(kwonly, bool):
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -9,11 +9,8 @@ from collections     import OrderedDict
 
 from sympy import preorder_traversal
 
-from sympy.simplify.radsimp   import fraction
 from sympy.core.compatibility import with_metaclass
-from sympy.core.singleton     import Singleton, S
-from sympy.core.function      import Derivative
-from sympy.core.function      import _coeff_isneg
+from sympy.core.singleton     import Singleton
 from sympy.core.expr          import Expr
 from sympy.logic.boolalg      import And as sp_And
 
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -2721,7 +2721,7 @@ class Import(Basic):
         self._source = source
         self._target = []
         self._ignore_at_print = ignore_at_print
-        if isinstance(target, (str, PyccelSymbol, DottedName, AsName)):
+        if isinstance(target, (str, DottedName, AsName)):
             self._target = [Import._format(target)]
         elif iterable(target):
             for i in target:
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -1559,7 +1559,7 @@ class ValuedArgument(Basic):
 
         if isinstance(expr, Argument):
             self._name = expr.name
-        elif isinstance(expr, (str, PyccelSymbol)):
+        elif isinstance(expr, str):
             self._name = expr
         else:
             raise TypeError('Expecting an argument')
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -7,12 +7,6 @@
 
 from collections     import OrderedDict
 
-from sympy import sympify
-from sympy import Symbol as sp_Symbol
-from sympy import Add as sp_Add, Mul as sp_Mul, Pow as sp_Pow
-from sympy import Eq as sp_Eq, Ne as sp_Ne, Lt as sp_Lt, Le as sp_Le, Gt as sp_Gt, Ge as sp_Ge
-from sympy import Integer as sp_Integer
-from sympy import Float as sp_Float, Rational as sp_Rational
 from sympy import preorder_traversal
 
 from sympy.simplify.radsimp   import fraction
@@ -20,9 +14,8 @@ from sympy.core.compatibility import with_metaclass
 from sympy.core.singleton     import Singleton, S
 from sympy.core.function      import Derivative
 from sympy.core.function      import _coeff_isneg
-from sympy.core.expr          import Expr, AtomicExpr
-from sympy.logic.boolalg      import And as sp_And, Or as sp_Or
-from sympy.logic.boolalg      import Boolean as sp_Boolean
+from sympy.core.expr          import Expr
+from sympy.logic.boolalg      import And as sp_And
 
 from sympy.utilities.iterables          import iterable
 
@@ -31,20 +24,18 @@ from pyccel.errors.errors import Errors
 from pyccel.errors.messages import RECURSIVE_RESULTS_REQUIRED
 
 from .basic     import Basic, PyccelAstNode
-from .builtins  import (PythonEnumerate, PythonLen, PythonList, PythonMap,
-                        PythonRange, PythonZip, PythonTuple, PythonBool, Lambda)
+from .builtins  import (PythonEnumerate, PythonLen, PythonMap,
+                        PythonRange, PythonZip, PythonBool, Lambda)
 from .datatypes import (datatype, DataType, NativeSymbol,
                         NativeBool, NativeRange,
                         NativeTuple, is_iterable_datatype, str_dtype)
 from .internals      import Slice, PyccelSymbol
 
-from .literals       import LiteralTrue, LiteralFalse, LiteralInteger, Nil
-from .literals       import LiteralImaginaryUnit, LiteralString
+from .literals       import LiteralInteger, Nil, convert_to_literal
 from .itertoolsext   import Product
-from .functionalexpr import GeneratorComprehension as GC
 from .functionalexpr import FunctionalFor
 
-from .operators import PyccelMul, IfTernaryOperator, Relational
+from .operators import PyccelMul, Relational
 
 from .variable import DottedName, DottedVariable, IndexedElement
 from .variable import ValuedVariable, Variable
@@ -108,15 +99,10 @@ __all__ = (
     '_atomic',
     'create_variable',
     'create_incremented_string',
-    'extract_subexpressions',
-#    'float2int',
     'get_assigned_symbols',
     'get_initial_value',
     'get_iterable_ranges',
     'inline',
-    'int2float',
-#    'is_simple_assign',
-    'local_sympify',
 #    'operator',
 #    'op_registry',
     'process_shape',
@@ -128,13 +114,6 @@ __all__ = (
 )
 
 #==============================================================================
-local_sympify = {
-    'N'    : sp_Symbol('N'),
-    'S'    : sp_Symbol('S'),
-    'zeros': sp_Symbol('zeros'),
-    'ones' : sp_Symbol('ones'),
-    'Point': sp_Symbol('Point')
-}
 
 # TODO - add EmptyStmt => empty lines
 #      - update code examples
@@ -149,7 +128,7 @@ def apply(func, args, kwargs):return func(*args, **kwargs)
 #==============================================================================
 def subs(expr, new_elements):
     """
-    Substitutes old for new in an expression after sympifying args.
+    Substitutes old for new in an expression.
 
     Parameters
     ----------
@@ -225,85 +204,6 @@ def _atomic(e, cls=None,ignore=()):
     return atoms_
 
 
-
-def extract_subexpressions(expr):
-    """this function takes an expression and returns a list
-      of statements if this expression contains sub expressions that need
-      to be evaluated outside of the expression
-
-
-      Parameters
-      ----------
-      expr : Add, Mul, Pow, FunctionCall
-
-    """
-
-    stmts = []
-    cls   = (sp_Add, sp_Mul, sp_Pow, sp_And,
-             sp_Or, sp_Eq, sp_Ne, sp_Lt, sp_Gt,
-             sp_Le, sp_Ge)
-
-    id_cls = (PyccelSymbol, IndexedElement,
-              DottedVariable, sp_Float, sp_Integer,
-              sp_Rational, LiteralImaginaryUnit,sp_Boolean,
-              LiteralTrue, LiteralFalse, LiteralString,
-              ValuedArgument, Nil, PythonList, PythonTuple,
-              StarredArguments)
-
-    func_names = ('diag', 'empty', 'zip', 'enumerate')
-    #TODO put only imported functions
-    def substitute(expr):
-        if isinstance(expr, id_cls):
-            return expr
-        if isinstance(expr, cls):
-            args = expr.args
-            args = [substitute(arg) for arg in args]
-            return expr.func(*args, evaluate=False)
-        elif isinstance(expr, FunctionCall):
-            args = substitute(expr.args)
-
-            if expr.func.name in func_names:
-                var = create_variable(expr)
-                expr = expr.func(*args, evaluate=False)
-                expr = Assign(var, expr)
-                stmts.append(expr)
-
-                return var
-            else:
-                expr = expr.func(*args, evaluate=False)
-                return expr
-        elif isinstance(expr, GC):
-            stmts.append(expr)
-            return expr.lhs
-        elif isinstance(expr,IfTernaryOperator):
-            var = create_variable(expr)
-            new = Assign(var, expr)
-            new.set_fst(expr.fst)
-            stmts.append(new)
-            return var
-        elif isinstance(expr, PythonList):
-            args = []
-            for i in expr:
-                args.append(substitute(i))
-
-            return PythonList(*args, sympify=False)
-
-        elif isinstance(expr, (tuple, list)):
-            args = []
-
-            for i in expr:
-                args.append(substitute(i))
-            return args
-
-        else:
-            raise TypeError('statement {} not supported yet'.format(type(expr)))
-
-
-    new_expr  = substitute(expr)
-    return stmts, new_expr
-
-
-
 #def collect_vars(ast):
 #    """ collect variables in order to be declared"""
 #    #TODO use the namespace to get the declared variables
@@ -340,13 +240,6 @@ def inline(func, args):
     body = func.body
     body = subs(body, zip(func.arguments, args))
     return Block(str(func.name), local_vars, body)
-
-
-def int2float(expr):
-    return expr
-
-def float2int(expr):
-    return expr
 
 def create_incremented_string(forbidden_exprs, prefix = 'Dummy', counter = 1):
     """This function takes a prefix and a counter and uses them to construct
@@ -1016,14 +909,13 @@ class While(Basic):
     """
 
     def __init__(self, test, body, local_vars=()):
-        test = sympify(test, locals=local_sympify)
 
         if PyccelAstNode.stage == 'semantic':
             if test.dtype is not NativeBool():
                 test = PythonBool(test)
 
         if iterable(body):
-            body = CodeBlock((sympify(i, locals=local_sympify) for i in body))
+            body = CodeBlock(body)
         elif not isinstance(body,CodeBlock):
             raise TypeError('body must be an iterable or a CodeBlock')
 
@@ -1072,10 +964,9 @@ class With(Basic):
         test,
         body,
         ):
-        test = sympify(test, locals=local_sympify)
 
         if iterable(body):
-            body = CodeBlock((sympify(i, locals=local_sympify) for i in body))
+            body = CodeBlock(body)
         elif not isinstance(body,CodeBlock):
             raise TypeError('body must be an iterable')
 
@@ -1675,6 +1566,14 @@ class ValuedArgument(Basic):
             self._name = expr
         else:
             raise TypeError('Expecting an argument')
+
+        if isinstance(value, (bool, int, float, complex, str)):
+            value = convert_to_literal(value)
+        elif not isinstance(value, (Basic, Symbol)):
+            raise TypeError("Expecting a pyccel object not {}".format(type(value)))
+
+        if not isinstance(kwonly, bool):
+            raise TypeError("kwonly must be a bool")
 
         self._expr   = expr
         self._value  = value
@@ -3449,19 +3348,6 @@ class StarredArguments(Basic):
         return self._starred_obj
 
 
-def is_simple_assign(expr):
-    if not isinstance(expr, Assign):
-        return False
-
-    assignable = [Variable, IndexedElement]
-    assignable += [sp_Integer, sp_Float]
-    assignable = tuple(assignable)
-    if isinstance(expr.rhs, assignable):
-        return True
-    else:
-        return False
-
-
 # ...
 
 # ...
@@ -3925,11 +3811,9 @@ def process_shape(shape):
     for s in shape:
         if isinstance(s,(LiteralInteger, Variable, Slice, PyccelAstNode, FunctionCall)):
             new_shape.append(s)
-        elif isinstance(s, sp_Integer):
-            new_shape.append(LiteralInteger(s.p))
         elif isinstance(s, int):
             new_shape.append(LiteralInteger(s))
         else:
-            raise TypeError('shape elements cannot be '+str(type(s))+'. They must be one of the following types: Integer(pyccel), Variable, Slice, PyccelAstNode, Integer(sympy), int, FunctionCall')
+            raise TypeError('shape elements cannot be '+str(type(s))+'. They must be one of the following types: LiteralInteger, Variable, Slice, PyccelAstNode, int, FunctionCall')
     return tuple(new_shape)
 
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -1791,6 +1791,9 @@ class FunctionCall(PyccelAstNode):
     def interface_name(self):
         return self._interface_name
 
+    def __str__(self):
+        return '{}({})'.format(self.func_name, ', '.join(str(a) for a in self.args))
+
     def __repr__(self):
         return '{}({})'.format(self.func_name, ', '.join(str(a) for a in self.args))
 
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -1870,8 +1870,8 @@ class FunctionDef(Basic):
             name_ = []
             for i in name:
                 if isinstance(i, str):
-                    name = name + PyccelSymbol(i)
-                elif not isinstance(i, PyccelSymbol):
+                    name_.append(PyccelSymbol(i))
+                else
                     raise TypeError('Function name must be PyccelSymbol or string'
                                     )
             name = tuple(name_)
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -1567,7 +1567,7 @@ class ValuedArgument(Basic):
         else:
             raise TypeError('Expecting an argument')
 
-        if isinstance(value, (bool, int, float, complex, str)):
+        if isinstance(value, (bool, int, float, complex, str)) and not isinstance(value, PyccelSymbol):
             value = convert_to_literal(value)
         elif not isinstance(value, (Basic, PyccelSymbol)):
             raise TypeError("Expecting a pyccel object not {}".format(type(value)))
--- a/pyccel/ast/headers.py
+++ b/pyccel/ast/headers.py
@@ -720,7 +720,7 @@ class MacroFunction(Header):
                 elif arg.argument in result_keys:
                     new = d_results[arg.argument]
                 else:
-                    raise ValueError('Unkonwn variable name')
+                    raise ValueError('Unknown variable name')
 
                 new = construct_macro(arg.name, new)
                 if isinstance(arg, MacroShape):
--- a/pyccel/ast/headers.py
+++ b/pyccel/ast/headers.py
@@ -5,14 +5,13 @@
 #------------------------------------------------------------------------------------------#
 
 from sympy.utilities.iterables import iterable
-from sympy import sympify
 
 from ..errors.errors import Errors
 from ..errors.messages import TEMPLATE_IN_UNIONTYPE
 from .core import Basic
 from .core import ValuedArgument
 from .core import FunctionDef, Interface, FunctionAddress
-from .core import local_sympify, create_incremented_string
+from .core import create_incremented_string
 from .datatypes import datatype, DataTypeFactory, UnionType
 from .macros import Macro, MacroShape, construct_macro
 from .variable import DottedName, DottedVariable
@@ -594,9 +593,6 @@ class MacroFunction(Header):
         # master can be a string or FunctionDef
         if not isinstance(master, (str, FunctionDef, Interface)):
             raise ValueError('Expecting a master name of FunctionDef')
-
-        if not(results is None):
-            results = [sympify(a, locals=local_sympify) for a in results]
 
         self._name             = name
         self._arguments        = args
--- a/pyccel/ast/headers.py
+++ b/pyccel/ast/headers.py
@@ -681,7 +681,7 @@ class MacroFunction(Header):
             for i, arg in d_arguments.items():
                 if isinstance(arg, Macro):
                     d_arguments[i] = construct_macro(arg.name,
-                                      d_arguments[arg.argument.name])
+                                      d_arguments[arg.argument])
                     if isinstance(arg, MacroShape):
                         d_arguments[i]._index = arg.index
 
--- a/pyccel/ast/headers.py
+++ b/pyccel/ast/headers.py
@@ -734,7 +734,7 @@ class MacroVariable(Header):
     """."""
 
     def __init__(self, name,  master):
-        if not isinstance(name, (str, PyccelSymbol, DottedName)):
+        if not isinstance(name, (str, DottedName)):
             raise TypeError('name must be of type str or DottedName')
 
 
--- a/pyccel/ast/headers.py
+++ b/pyccel/ast/headers.py
@@ -344,12 +344,12 @@ class FunctionHeader(Header):
                     results = []
                     _count = 0
                     for dc in d['decs']:
-                        _name, _count = create_incremented_string(used_names, 'in_', _count)
+                        _name, _count = create_incremented_string(used_names, 'in', _count)
                         var = build_argument(_name, dc)
                         decs.append(var)
                     _count = 0
                     for dc in d['results']:
-                        _name, _count = create_incremented_string(used_names, 'out_', _count)
+                        _name, _count = create_incremented_string(used_names, 'out', _count)
                         var = build_argument(_name, dc)
                         results.append(var)
                     arg_name = 'arg_{0}'.format(str(i))
--- a/pyccel/ast/headers.py
+++ b/pyccel/ast/headers.py
@@ -690,7 +690,7 @@ class MacroFunction(Header):
         if not(results is None) and not(self.results is None):
             for (r_macro, r) in zip(self.results, results):
                 # TODO improve name for other Nodes
-                d_results[r_macro.name] = r
+                d_results[r_macro] = r
         # ...
 
         # ... initialize new args with None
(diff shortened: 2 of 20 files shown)


pyccel/ast/internals.py
```diff
@@ -163,3 +162,46 @@ def __str__(self):
             stop = str(self.stop)
         return '{0} : {1}'.format(start, stop)
 
+class Symbol(Basic):
+    """
+        #TODO
+    """
+    is_Symbol = True
```
EmilyBourne, line 169: What is this for?

pyccel/ast/internals.py
```diff
@@ -163,3 +162,46 @@ def __str__(self):
...
+        return self._name
+
+    def __eq__(self, other):
+        if self is other:
+            return True
+        if type(self) is type(other):
+            try :
+                return self._name == other.name
+            except:
+                return False
```
EmilyBourne, line 188: Didn't you want to be able to compare with strings?
Can't you just do:
`return str(self)==s`
?

pyccel/ast/variable.py
```diff
@@ -119,7 +119,7 @@ def __init__(
         is_kwonly=False,
         allows_negative_indexes=False
         ):
-        super().__init__()
+        super().__init__(name)
```
EmilyBourne, line 122: Presumably you need to remove the `_name` attribute defined here?

pyccel/parser/semantic.py
```diff
@@ -2133,7 +2133,8 @@ def _visit_FunctionalFor(self, expr, **settings):
         indices = expr.indices
         dims    = []
         body    = expr.loops[1]
-
+        #TODO shouldn't be a better way than this ?
+        indices = [sp_Symbol(i.name) for i in indices]
```
EmilyBourne, line 2137: Is this needed for the atoms statement later?

EmilyBourne (commented):
This looks pretty clean, just a simple swap of the imports :)

pyccel/ast/internals.py
MenoIy: sometime we is   X._is_Symbol in the code

pyccel/ast/internals.py
EmilyBourne: Or:
`return self._name == s`

pyccel/parser/semantic.py
MenoIy: yes the one at line 2237

pyccel/ast/variable.py
MenoIy: yes if this is finished by keeping symbol inheritance in Variable we can remove it

pyccel/ast/internals.py
MenoIy: I was just afraid that some were in the code we use __eq__ between two objects of different class

pyccel/ast/internals.py
EmilyBourne: I could only find `is_Symbol` [here](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/core.py#L263) and [here](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/core.py#L269) in the function `allocatable_like`, but I couldn't find any uses of that function. Did you find it anywhere else?
If not then it looks like most of that function is broken, and the docstring isn't clear enough for me to understand what it was even supposed to do in the first place. I would just delete the function

pyccel/ast/internals.py
EmilyBourne: If we do then they can always write their own `__eq__` method which handles strings. If it's a different class then the definition of `==` might be more complicated than just having the same name anyway

pyccel/parser/semantic.py
EmilyBourne: Ok, then I suggest you leave a clear `TODO` referencing issue #694 and I'll fix it when I add the equivalent of `atoms` to `Basic`

pyccel/decorators.py
```diff
@@ -26,8 +26,8 @@
 def lambdify(f):
 
     args = f.__code__.co_varnames
-    from sympy import symbols
-    args = symbols(args)
+    from pyccel.ast.internals import Symbol
+    args = Symbol(args)
     expr = f(*args)
     def wrapper(*vals):
         return  expr.subs(zip(args,vals)).doit()
```
EmilyBourne, line 33: @saidctb Can you provide more information (or a docstring) about lambdify?
Does it handle sympy objects or pyccel objects?

pyccel/ast/internals.py
MenoIy: I'm lost can you explain more what do you expect, if I only compare the property name it can be another object with the same name and has the property, and then it will be equal.
that's why I check the type too, this is also helping with the class that inherits from Symbol like (Variable)

pyccel/decorators.py
saidctb: here it must stay with sympy

saidctb: here we are trying to mimic the lambdify of sympy

EmilyBourne:
For PR #718 I have had to change `FunctionCall.funcdef` so that it is a `Symbol` at the syntactic stage. With the code in its current state this has meant that I had to add `str` casts. I know you are looking at this currently. I don't know if you want to make this change in this PR? When you have an `__eq__` function which accepts strings, you should only need to modify 1 or 2 lines I think

pyccel/decorators.py
```diff
@@ -26,8 +26,8 @@
 def lambdify(f):
 
     args = f.__code__.co_varnames
-    from sympy import symbols
-    args = symbols(args)
+    from pyccel.ast.internals import Symbol
```
saidctb: we should keep the sympy symbol here

pyccel/ast/core.py
```diff
@@ -201,78 +199,6 @@ def subs(expr, new_elements):
     else:
         return expr
 
-
-def allocatable_like(expr, verbose=False):
```
EmilyBourne, line 205: Could you add a line about the removal of this function in the PR description please

pyccel/ast/internals.py
```diff
@@ -163,3 +162,74 @@ def __str__(self):
...
+    def __eq__(self, other):
+        if isinstance(other, str):
+            return self._name == other
+        elif type(self) is type(other):
+            return self._name == other.name
+        else:
+            return False
+
+    def __hash__(self):
+        return hash(self._name)
```
EmilyBourne, line 201: If 2 objects with the same name but different classes are not equal then shouldn't they have different hashes?

pyccel/parser/semantic.py
```diff
@@ -2139,7 +2139,8 @@ def _visit_FunctionalFor(self, expr, **settings):
         indices = expr.indices
         dims    = []
         body    = expr.loops[1]
-
+        #TODO remove the line below after adding equivalent of atoms to Basic issue #694
+        indices = [sp_Symbol(i.name) for i in indices]
```
EmilyBourne, line 2143: I think you should name this `sp_indices`. Currently you are creating the final `FunctionalFor` block with sympy objects

EmilyBourne, line 2143: This is used in the big for loop later. `atoms` is one of the reasons that sympy is needed here, but it is not the only one. There is also `simplify` and `Summation`. As these two operations are purely sympy calculations it makes sense to keep the operations in this loop as sympy operations. That way there are fewer conversions between pyccel and sympy. So I think this TODO can be removed

EmilyBourne (changes requested):
Really good work on this! Well done for sorting through all the `str` functions !

pyccel/ast/internals.py
MenoIy: seem logic but i wanted to make a string hash the same as the symbol one

pyccel/ast/core.py
```diff
@@ -601,7 +527,8 @@ class Assign(Basic):
 
     Examples
     --------
-    >>> from sympy import symbols, MatrixSymbol, Matrix
+    >>> from sympy import MatrixSymbol, Matrix
```
yguclu, line 530: Do we still use `sympy.MatrixSymbol` in Pyccel?

MenoIy: No, I'm gonna locate and update all docstrings.

pyccel/ast/internals.py
```diff
@@ -163,3 +162,74 @@ def __str__(self):
             stop = str(self.stop)
         return '{0} : {1}'.format(start, stop)
 
+class Symbol(Basic):
+    """
+    Represent variable with undefined type
```
yguclu, line 167: This docstring needs a bit more explanation.Is the following accurate?
```
Symbolic placeholder for a Python variable, which has a name but no type yet.
This is very generic, and it can also represent a function or a module.
```

yguclu (commented):
Do we still need `Variable` to subclass `Symbol`?

MenoIy:
> Do we still need `Variable` to subclass `Symbol`?

yes, either we keep the inheritance or create magic methods for Variable class (__eq__, and __hash__).

pyccel/ast/internals.py
MenoIy: done in 2b0aea0

yguclu:
> > Do we still need `Variable` to subclass `Symbol`?
> 
> yes, either we keep the inheritance or create magic methods for Variable class (**eq**, and **hash**).

If we change `Variable` so that it does not subclass `Symbol`, we will obtain this very neat situation:
1. `Symbol` instances will only be created at the syntactic stage;
2. `Variable` instances will only be created at the semantic stage.

As a consequence, we can find any undefined symbols by searching for `Symbol` instances.

@saidctb @EmilyBourne What do you think?

pyccel/ast/core.py
MenoIy: done 71c33ad , 02c4dc9

EmilyBourne: The matrix classes aren't used, but I think they have `_print_X` methods in all 4 printers

pyccel/ast/internals.py
EmilyBourne: Any reason for that?

Normally if hashes are the same, the objects are equal. I looked into it briefly and found [this link](https://hynek.me/articles/hashes-and-equality/) which says:

> Hashes can be less picky than equality checks. Since key lookups are always followed by an equality check, your hashes don’t have to be unique. That means that you can compute your hash over an immutable subset of attributes that may or may not be a unique “primary key” for the instance.
> 
> You can take this property to town by returning a constant hash and make the set or dictionary work purely on equality checks. However that would regress them into a list and lead to terrible performance.

So apparently it's ok but we are likely to lose performance which isn't great.

If the pyccel `Symbol` and the sympy `Symbol` need to have the same hash for some reason, then I am inclined to agree with Yaman that `Variable` shouldn't inherit from `Symbol` so different `Variable` classes can have different hashes

EmilyBourne:
> If we change `Variable` so that it does not subclass `Symbol`, we will obtain this very neat situation:
> 
> 1. `Symbol` instances will only be created at the syntactic stage;
> 2. `Variable` instances will only be created at the semantic stage.
> 
> As a consequence, we can find any undefined symbols by searching for `Symbol` instances.
> 
> @saidctb @EmilyBourne What do you think?

This seems clean, but then so does reusing code through inheritance. I do not have a preference but I'm not sure when we would need to find any undefined symbols. This is equivalent to finding objects which have escaped the semantic stage. I'm not sure that's possible so I'm not sure if we gain anything by breaking them up

yguclu:
In my opinion code reuse is not enough to justify double inheritance, unless we reuse a lot of code. And the `Symbol` class is quite small.

Further, here it makes sense that the `__hash__` function for `Symbol` only checks the instance `name`, but this is definitely not OK for a `Variable`, so that will need to be overridden anyway.

EmilyBourne:
> In my opinion code reuse is not enough to justify double inheritance, unless we reuse a lot of code. And the `Symbol` class is quite small.
> 
> Further, here it makes sense that the `__hash__` function for `Symbol` only checks the instance `name`, but this is definitely not OK for a `Variable`, so that will need to be overridden anyway.

In that case I agree that we can split them

pyccel/ast/core.py
yguclu: Thanks!

yguclu (changes requested):
Please separate class `Variable` from your new class `Symbol`, that is, `Variable` should not inherit from `Symbol` any more. This way `Symbol` objects can be immutable, while `Variable` objects are not.

yguclu:
@MenoIy Please resolve the conflicts with master before proceeding.

MenoIy:
I removed symbol inheritance from the Variable class: 
- now I'm using variable property name in some case : (dictionary key)
what do you think should i keep using property name or add magic method __eq__ __hash__ for Variable  @EmilyBourne?

pyccel/ast/variable.py
```diff
@@ -199,7 +198,8 @@ def __init__(
         self._order          = order
         self._is_argument    = is_argument
         self._is_kwonly      = is_kwonly
-        super().__init__(name)
+
+        PyccelAstNode.__init__(self)
```
EmilyBourne, line 202: This can be changed to `super().__init__()` now that there is no ambiguity between the superclasses

EmilyBourne:
> I removed symbol inheritance from the Variable class:
> 
> * now I'm using variable property name in some case : (dictionary key)
>   what do you think should i keep using property name or add magic method **eq** **hash** for Variable  @EmilyBourne?

Personally I would have added `__eq__` and `__hash__` to `Variable` as I think it is more natural to handle the `Variable` directly and not have to think about when to use `Variable` and when to use the `name` property. But as you've already made the changes I'll leave it up to you to decide if you want to go back through them

pyccel/ast/variable.py
```diff
@@ -32,7 +30,7 @@
     'VariableAddress'
 )
 
-class Variable(Symbol, PyccelAstNode):
+class Variable(PyccelAstNode):
```
EmilyBourne, line 33: With the double inheritance fixed, the `__new__` function should no longer be necessary

pyccel/ast/variable.py
```diff
@@ -357,6 +351,14 @@ def is_ndarray(self):
     def __str__(self):
         return str(self.name)
 
+    def __eq__(self, other):
+        if isinstance(other, Variable):
+            return self._name == other.name
+        return False
+
+    def __hash__(self):
+        return hash(type(self).__name__+ self._name)
```
EmilyBourne, line 360: The `__eq__` and `__hash__` don't match. A `ValuedVariable` is a `Variable` but it will not have the same hash as a `Variable` with the same name

MenoIy: I totally forgot about valuedVariable

EmilyBourne: `TupleVariable` and `DottedVariable` too. Especially `DottedVariable`, `cls.a`=/=`a`

MenoIy: do you think a ```DottedVariable``` should have its own``` __eq__ ```?

EmilyBourne: Actually that's a really good idea. Currently we just test the name. I'm not sure if the name of `a.b` is `'b'` or `'a.b'` if it's the first then we have a problem, if it's the second then it's a strange test for things such as `f().b`

EmilyBourne (approved):

pyccel/ast/variable.py
```diff
@@ -831,3 +831,6 @@ def __eq__(self, other):
             return self.name == other.name and self.lhs == other.lhs
 
         return False
+
+    def __hash__(self):
+        return hash(type(self).__name__+ self.name)
```
EmilyBourne, line 836: Shouldn't the `lhs` be in the hash?

MenoIy: good point

MenoIy: adding the lhs to the hash we should be 100% that whatever  the type of lhs  it should have an ```__str__``` method, or some that can identify it as its own hash method

pyccel/ast/variable.py
```diff
@@ -823,3 +825,12 @@ def lhs(self):
...
         return self._lhs
+
+    def __eq__(self, other):
+        if isinstance(other, DottedVariable):
+            return self.name == other.name and self.lhs == other.lhs
+
+        return False
+
+    def __hash__(self):
+        return hash(type(self).__name__+ self.name + str(self.lhs))
```
EmilyBourne, line 836: I don't like this. `hash` should include all the information necessary to differentiate 2 objects. `str` is a lot less strict and is just there to give a legible representation of a function. Imagine the case where the `lhs` is a function call, it may not be possible to get enough nuance just from the string representation. I would prefer:
```
return hash(type(self).__name__+ self.name) + hash(self.lhs)
```
(hash returns an int so they can be added)

saidctb: or you just can do , `hash((type(self).__name__, self.name, self.lhs))`

EmilyBourne (approved):

pyccel/ast/internals.py
```diff
@@ -169,3 +168,69 @@ def __str__(self):
...
+        super().__init__()
+
+    @property
+    def name(self):
+        """Name of the symbol
+        """
+        return self._name
+
+    def __eq__(self, other):
+        return self._name == other
```
yguclu, line 199: Shouldn't you check the type before? E.g.
```python
def __eq__(self, other):

    if type(self) is type(other):
        return self._name == other._name

    return False
```
Hashes don't have to be picky, but equality checks must be!

pyccel/ast/variable.py
```diff
@@ -357,6 +351,14 @@ def is_ndarray(self):
     def __str__(self):
         return str(self.name)
 
+    def __eq__(self, other):
+        if type(self) is type(other):
+            return self._name == other.name
+        return False
+
+    def __hash__(self):
+        return hash((type(self).__name__, self._name))
```
yguclu, line 360: Just curious, why do you use `type(self).__name__` instead of just `type(self)`?

pyccel/ast/variable.py
```diff
@@ -823,3 +825,12 @@ def lhs(self):
         The lhs is a
         """
         return self._lhs
+
+    def __eq__(self, other):
+        if isinstance(other, DottedVariable):
+            return self.name == other.name and self.lhs == other.lhs
```
yguclu, line 831: If you subclass this class and forget to override `__eq__`, this function will always return `False`. You should check if `self` and `other` have the same type instead, as you did before in the same file.

pyccel/decorators.py
```diff
@@ -26,8 +26,8 @@
 def lambdify(f):
 
     args = f.__code__.co_varnames
-    from sympy import symbols
-    args = symbols(args)
+    from sympy import Symbol
+    args = Symbol(args)
```
yguclu, line 30: Why did you change these lines? Are you sure that `sympy.Symbol` can return multiple objects like `sympy.symbols`?

yguclu (changes requested):

pyccel/ast/variable.py
EmilyBourne: You have this the wrong way round. `isinstance(other, DottedVariable)` returns `True` if `other` is a `DottedVariable` or a sub-class of `DottedVariable`, whereas `type(self) is type(other)` will only return `True` if both types are identical

pyccel/ast/internals.py
EmilyBourne: The `__eq__` operator of `Symbol` is designed such that:
```
Symbol('s')==str('s')
```
returns `True` to avoid the hundreds of casts done all over pyccel. It is just a wrapper around the string

pyccel/decorators.py
MenoIy: ah true its not the same i will correct it

pyccel/ast/variable.py
MenoIy: I just prefered to hash string  because I don't know how 'class type' hash behave

pyccel/decorators.py
MenoIy: done in 5a525fe

pyccel/ast/variable.py
MenoIy: if class X inherits from  DottedVariable and we want that X should never be equal to DottedVariable
we should use ```type(self) is type(other)``` otherwise we can keep it like this

pyccel/ast/internals.py
yguclu: Not sure I understand why mimicking Sympy is necessary. Can you give me an example?

pyccel/ast/variable.py
yguclu: Sorry, you're right. Nevertheless, the code is still wrong :wink:

An equality comparison must return `False` if the two types are not identical.

pyccel/ast/internals.py
MenoIy:
```
a = 'name'
sym = Symbol(''name)
```
to avoid writing ```str(sym) == a``` every time were not sure if it's a symbol or string
we can do ```a == sym```  just as Emily said Symbol is just a wrapper around the string

pyccel/decorators.py
yguclu: Thanks!

yguclu:
While discussing hashes, a doubt has just arisen in my mind: do we ever change the `_name` attribute of a `Variable` after its creation? I hope not, because the hash of an object should never change during its lifetime...

MenoIy:
> While discussing hashes, a doubt has just arisen in my mind: do we ever change the `_name` attribute of a `Variable` after its creation? I hope not, because the hash of an object should never change during its lifetime...

No, we don't

pyccel/ast/variable.py
yguclu: My last reply was for @EmilyBourne, but I need to manually refresh this webpage in order to see your replies, and so I am always late... @MenoIy Please use `type(self) is type(other)` as you do in other places.

EmilyBourne:
> While discussing hashes, a doubt has just arisen in my mind: do we ever change the `_name` attribute of a `Variable` after its creation? I hope not, because the hash of an object should never change during its lifetime...

I also noticed this and was a bit nervous. `Variable` has a `rename` function so it is technically possible, but I'm not sure if this function is used

pyccel/ast/internals.py
yguclu: I see. But then why do we even need a Symbol class? We could just use strings and be happy, no?

yguclu: If we care about performance, we should be sure of the types that we use at all times. Otherwise we should call `str` and be satisfied with the performance loss introduced by the `__str__` method...

pyccel/ast/variable.py
yguclu: I see, thanks!

yguclu:
> > While discussing hashes, a doubt has just arisen in my mind: do we ever change the `_name` attribute of a `Variable` after its creation? I hope not, because the hash of an object should never change during its lifetime...
> 
> I also noticed this and was a bit nervous. `Variable` has a `rename` function so it is technically possible, but I'm not sure if this function is used

Mmh, this must be checked before merging this PR!

pyccel/ast/variable.py
MenoIy: done a592199

pyccel/ast/internals.py
EmilyBourne: > I see. But then why do we even need a Symbol class? We could just use strings and be happy, no?

As I understand it, the wrapper is used to choose the correct `_visit_X` method in the `SemanticParser`. A `Symbol` is something that represents an object (variable/function/datatype/etc), while a string represents a string.

EmilyBourne: That said, I have just checked [Codacy](https://app.codacy.com/gh/pyccel/pyccel/file/52912287204/coverage?bid=15415567&fileBranchId=15415567) and it seems the `_visit_str` function is not used. Presumably it has been replaced with `_visit_LiteralString` so  perhaps we don't need a symbol class anymore

pyccel/ast/variable.py
yguclu: Thanks

pyccel/ast/internals.py
yguclu: > > I see. But then why do we even need a Symbol class? We could just use strings and be happy, no?
> 
> As I understand it, the wrapper is used to choose the correct `_visit_X` method in the `SemanticParser`. A `Symbol` is something that represents an object (variable/function/datatype/etc), while a string represents a string.

I was being ironical, but you see.. we could still achieve what you describe with a dummy class that subclasses `str`. Python strings are immutable, and hence hashable by definition...

EmilyBourne: Hmm, that sounds like a good idea to me, although we would have to track down all the uses of `.name`

saidctb: a Symbol class is like a `Name` class, it just holds the name, it should be kept though

EmilyBourne:
> > > While discussing hashes, a doubt has just arisen in my mind: do we ever change the `_name` attribute of a `Variable` after its creation? I hope not, because the hash of an object should never change during its lifetime...
> > 
> > 
> > I also noticed this and was a bit nervous. `Variable` has a `rename` function so it is technically possible, but I'm not sure if this function is used
> 
> Mmh, this must be checked before merging this PR!

`rename` is used once [here](https://github.com/pyccel/pyccel/blob/master/pyccel/codegen/printing/fcode.py#L1549-L1554). This is done to handle function types such as:
```
#$ header function high_valuedarg_1(int, (int)(int))
```
In this case the function and its argument are anonymous, however a name is required when we want to print the object in fortran

We are discussing the best solution for this.
Options:
1. Use `clone` instead of `rename`
2. Give the arguments a name from the start and find a way to not print them in C
3. Give the arguments a name from the start and leave them printed in c
4. ?

pyccel/ast/internals.py
yguclu: > a Symbol class is like a `Name` class, it just holds the name, it should be kept though,
> the Python AST uses the class `Name`

@saidctb Emily and I were discussing about using this extremely simple class:
```python
class Symbol(str):
    pass

x = Symbol('hello')
```
Here the comparison `x == 'hello'` will return `True`.

saidctb: ah okay , yes maybe

EmilyBourne: I think the class should inherit from Basic though so it can have an `fst` which lets us say at which line/column there is a problem

yguclu: OK, let's leave things as they are for now. :smiley:

yguclu:
With regard to the `rename` issue, I suppose that we create the temporary variables with random names and later we rename them in the Fortran printer where the name is actually printed?

If that is correct, these variables already had a random name, and so we should be in case 2...

EmilyBourne:
> With regard to the `rename` issue, I suppose that we create the temporary variables with random names and later we rename them in the Fortran printer where the name is actually printed?
> 
> If that is correct, these variables already had a random name, and so we should be in case 2...

No, we create them with the name that will be used in C `''` ([here](https://github.com/pyccel/pyccel/blob/f9757528b981e55423aeafc115e4ae5eb6a37add/pyccel/ast/headers.py#L346-L349) using [this](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/headers.py#L278-L279) function)

Although, given the caching system I am surprised that we are able to give them different properties (type etc) using this method. In any case these variables will all have the same hash, so presumably the only reason we don't have problems is because they are only used when calling the function

pyccel/ast/internals.py
yguclu: > I think the class should inherit from Basic though so it can have an `fst` which lets us say at which line/column there is a problem

I am not so sure about this anymore. Currently we use sympy's `Symbol`, which does not have the `fst` attribute. I imagine that in 99.9 % of cases (if not all of them) the problem is generated at another node that uses one or more `Symbols` objects, so we would still have access to the `fst` through that node.

yguclu:
> > With regard to the `rename` issue, I suppose that we create the temporary variables with random names and later we rename them in the Fortran printer where the name is actually printed?
> > If that is correct, these variables already had a random name, and so we should be in case 2...
> 
> No, we create them with the name that will be used in C `''` ([here](https://github.com/pyccel/pyccel/blob/f9757528b981e55423aeafc115e4ae5eb6a37add/pyccel/ast/headers.py#L346-L349) using [this](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/headers.py#L278-L279) function)

I still don't see why we should have a problem if we give them the Fortran names. We only need to print the function signature declaration, which we already do in C headers. There we do not print the arguments' names, so I can only assume that we already have Option 2 in place somewhere in our code base.

pyccel/ast/internals.py
EmilyBourne: This is one of the things that is improved in PR #718. As almost everything that is returned from `_visit` will then inherit from Basic, the fst can be set more consistently. In this case it is then easier to extract the fst from the `symbol` argument passed to `errors.report` rather than passing a `bounding_box`. Although the problem is not generated by the `Symbol` node, the `Symbol` node is used as the `symbol` passed to `errors.report` I think

EmilyBourne:
> > > With regard to the `rename` issue, I suppose that we create the temporary variables with random names and later we rename them in the Fortran printer where the name is actually printed?
> > > If that is correct, these variables already had a random name, and so we should be in case 2...
> > 
> > 
> > No, we create them with the name that will be used in C `''` ([here](https://github.com/pyccel/pyccel/blob/f9757528b981e55423aeafc115e4ae5eb6a37add/pyccel/ast/headers.py#L346-L349) using [this](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/headers.py#L278-L279) function)
> 
> I still don't see why we should have a problem if we give them the Fortran names.

I don't think we would have a problem, just slightly different C code

> We only need to print the function signature declaration, which we already do in C headers. There we do not print the arguments' names, so I can only assume that we already have Option 2 in place somewhere in our code base.

This is not true, in C we do print the arguments' names. But the arguments' names are all empty strings which is why it appears as though we do not. Take the following example:
```
#$ header function high_int_int_1((int)(int), (int)(int), int)
def high_int_int_1(function1, function2, a):
    x = function1(a)
    y = function2(a)
    return x + y
```
The header file contains:
```
#ifndef TEST_H
#define TEST_H

#include <stdint.h>
#include <stdlib.h>

int64_t high_int_int_1(int64_t (*function1)(int64_t ), int64_t (*function2)(int64_t ), int64_t a);

#endif // TEST_H
```
Note the space after the type `(int64_t )`. This is there because the argument name is printed after. But as it is an empty string we see nothing

yguclu:
Thanks for the explanation, I think I have the complete picture now. I was specifically thinking of standard arguments (`int64_t a` in your example) and I wrongly assumed that we did not print `a` in the C header.

To summarize:

1. We create the all `Variable` arguments of an "anonymous function signature" with an empty name `''`;
2. In the Fortran printer we rename these arguments as `in_01`, `in_02`, `out_01`, etc.;
3. In both printers we print all variables' names.

yguclu:
I imagine that those `Variable` objects that are renamed in the Fortran printer will never be used as elements of a `set` or `frozenset`, or as keys of a `dict`. So the fact that heir hash is broken should not play a role here.

Nevertheless, I prefer if we remove the `rename` method and simply create new `Variable` with an updated name. This should be possible using a `subs` method after PR #718, correct? So please just create a new issue about the need to remove the `rename` method, to which we should give the highest priority.

pyccel/ast/headers.py
```diff
@@ -342,11 +342,15 @@ def find_templates(signature, templates):
                 if (d['is_func']):
                     decs = []
                     results = []
+                    _count = 0
                     for dc in d['decs']:
-                        var = build_argument('', dc)
+                        _name, _count = create_incremented_string(set(), 'in_', _count)
```
EmilyBourne, line 347: There is no point using this function if you always pass it an empty set ^^

pyccel/codegen/printing/ccode.py
```diff
@@ -787,7 +787,22 @@ def _print_NativeComplex(self, expr):
     def _print_NativeString(self, expr):
         return 'string'
 
-    def function_signature(self, expr):
+    def function_signature(self, expr, print_arg_names = True):
```
EmilyBourne, line 790: I think the default should be False, as in most places we don't need the argument names (.h file, function as argument arguments)

pyccel/ast/variable.py
```diff
@@ -411,8 +416,10 @@ def clone(self, name, new_class = None, **kwargs):
         return cls(**new_kwargs)
 
     def rename(self, newname):
-        """Change variable name."""
-
+        """ Forbidden method for renaming the variable
+        """
+        # The name is part of the hash so it must never change
+        raise RuntimeError('Cannot modify hash definition')
         self._name = newname
```
EmilyBourne, line 423: This line is unreachable so it can be removed

pyccel/codegen/printing/ccode.py
MenoIy: i was thinking the special case here is not printing arg name 
given that we call function signature in a lot of different places

EmilyBourne: Ah ok. I thought it was only called from `_visit_FunctionDef` (True), `_visit_ModuleHeader` (currently True, but it wouldn't be unusual if it was False here, see @yguclu 's assumption above) and itself (False)

EmilyBourne: Actually the [consensus](https://stackoverflow.com/questions/7891526/should-function-declarations-include-parameter-names) seems to be that it is better for documentation/readability purposes to include the argument name in the .h file so I agree that the default should be `True`

pyccel/ast/core.py
```diff
@@ -201,78 +199,6 @@ def subs(expr, new_elements):
...
-            if not a.is_Symbol and not isinstance(a, (IndexedElement,
-                    FunctionCall)):
-                args.extend(a.args)
-
-            if isinstance(a, (Variable, IndexedElement)):
-                return a
-            elif a.is_Symbol:
-                raise TypeError('Found an unknown symbol {0}'.format(str(a)))
-    else:
-        raise TypeError('Unexpected type {0}'.format(type(expr)))
```
EmilyBourne, line 272: This function uses lots of sympy objects which don't seem to be used elsewhere. Codacy often misses new `unused-import` errors which appear due to code removal. Could you run pylint locally to try and clean this up? The command is:
```
pylint ast/core.py --disable=all --enable=unused-variable
```

pyccel/parser/semantic.py
```diff
@@ -991,10 +991,14 @@ def _visit_IndexedElement(self, expr, **settings):
 
         return self._extract_indexed_from_var(var, args, name)
 
+    def _visit_Symbol(self, expr, **settings):
+        # While and With class are using sympy.sympify
```
EmilyBourne: This should be fixed when you merge the master branch. I just removed this in PR #732

EmilyBourne: Beware of the new `from sympy import Symbol` in `pyccel/ast/macros.py`

MenoIy: Yes I just saw your PR .

EmilyBourne: There are several uses of `Symbol` in that PR which need updating:
https://github.com/pyccel/pyccel/pull/732/files#diff-707dd305f5ecd064576e2cfe995707cbe5f02e1b7cf7a9d5baa6418df3a82573R1645
https://github.com/pyccel/pyccel/pull/732/files#diff-68eb50666798e9569a48a77a7d59e728348769aeea74f48daf6a5b955586170aR10
https://github.com/pyccel/pyccel/pull/732/files#diff-68eb50666798e9569a48a77a7d59e728348769aeea74f48daf6a5b955586170aR31
https://github.com/pyccel/pyccel/pull/732/files#diff-2e950ccaf32b877cb52a5097e9f89778194c0fc725489295c13058f341373016R412
https://github.com/pyccel/pyccel/pull/732/files#diff-2e950ccaf32b877cb52a5097e9f89778194c0fc725489295c13058f341373016R452

MenoIy: yes I'm looking thanks

pyccel/ast/headers.py
```diff
@@ -340,11 +342,15 @@ def find_templates(signature, templates):
                 if (d['is_func']):
                     decs = []
                     results = []
+                    _count = 0
                     for dc in d['decs']:
-                        var = build_argument('', dc)
+                        _name, _count = create_incremented_string(used_names, 'in_', _count)
```
EmilyBourne, line 347: `create_incremented_string` adds the `_` itself, so this is not needed here

pyccel/ast/headers.py
```diff
@@ -340,11 +342,15 @@ def find_templates(signature, templates):
...
+                        var = build_argument(_name, dc)
                         decs.append(var)
+                    _count = 0
                     for dc in d['results']:
-                        var = build_argument('', dc)
+                        _name, _count = create_incremented_string(used_names, 'out_', _count)
```
EmilyBourne, line 352: See above

pyccel/ast/headers.py
```diff
@@ -264,11 +265,12 @@ def create_definition(self, templates = ()):
         # TODO factorize what can be factorized
         from itertools import product
 
-        name = str(self.name)
+        name = self.name
 
         body      = []
         cls_name  = None
         is_static = self.is_static
+        used_names = set(name)
```
EmilyBourne, line 273: The placement of this line means that we generate the signature as follows:
```
function high_int_int_1(function1, function2, a) result(Out_0009)

  implicit none

  integer(C_INT64_T) :: Out_0009
  integer(C_INT64_T), value :: a
  integer(C_INT64_T) :: x
  integer(C_INT64_T) :: y

  interface
    function function1(in__0000) result(out__0000)
      integer(C_INT64_T) :: out__0000
      integer(C_INT64_T), value :: in__0000
    end function function1

    function function2(in__0001) result(out__0001)
      integer(C_INT64_T) :: out__0001
      integer(C_INT64_T), value :: in__0001
    end function function2
  end interface
```
whereas previously we had:
```
function high_int_int_1(function1, function2, a) result(Out_0009)

  implicit none

  integer(C_INT64_T) :: Out_0009
  integer(C_INT64_T), value :: a
  integer(C_INT64_T) :: x
  integer(C_INT64_T) :: y

  interface
    function function1(in_0) result(out_0)
      integer(C_INT64_T) :: out_0
      integer(C_INT64_T), value :: in_0
    end function function1

    function function2(in_0) result(out_0)
      integer(C_INT64_T) :: out_0
      integer(C_INT64_T), value :: in_0
    end function function2
  end interface
```
(Note the arguments, previously in_0 and in_0, but now in_0 and in_1). I think both make sense, but I was just wondering whether this is a conscious choice?

pyccel/ast/headers.py
```diff
@@ -694,25 +700,25 @@ def apply(self, args, results=None):
...
-                elif arg.argument.name in result_keys:
-                    new = d_results[arg.argument.name]
+                if arg.argument in argument_keys:
+                    new = d_arguments[arg.argument]
+                    if isinstance(new, PyccelSymbol) and new in result_keys:
+                        new = d_results[new]
+                elif arg.argument in result_keys:
+                    new = d_results[arg.argument]
                 else:
                     raise ValueError('Unkonwn variable name')
```
EmilyBourne, line 723: Would you mind fixing this typo `Unkonwn`->`Unknown`

pyccel/ast/variable.py
```diff
@@ -165,6 +159,9 @@ def __init__(
             else:
                 name = DottedName(*name)
 
+        if name == '':
```
EmilyBourne, line 162: [Previously](https://github.com/pyccel/pyccel/pull/716/commits/8057e1f13e008a0e57e447ef8641b40b803d2872) you were using `.strip()` here, now that `PyccelSymbol` inherits from `str` that should be possible. Is there any reason you didn't put it back?

pyccel/codegen/printing/ccode.py
```diff
@@ -800,9 +815,9 @@ def function_signature(self, expr):
         if not args:
             arg_code = 'void'
         else:
-            arg_code = ', '.join('{}'.format(self.function_signature(i))
-                        if isinstance(i, FunctionAddress) else '{0}{1}'.format(self.get_declare_type(i), i)
-                        for i in args)
+            arg_code = ', '.join('{}'.format(self.function_signature(i, False))
+                        if isinstance(i, FunctionAddress) else '{0}{1}'.format(self.get_declare_type(i), i.name)
+                        if print_arg_names else '{}'.format(self.get_declare_type(i)) for i in args)
```
EmilyBourne, line 820: 2 comments here:
1. Do you know where the space is added between the dtype and the name?
2. I wonder if it is more readable to do: `i.name if print_arg_names else ''` rather than having 2 format strings to reduce the size of this line?

pyccel/codegen/printing/ccode.py
```diff
@@ -1424,6 +1439,9 @@ def _print_Comment(self, expr):
 
         return '/*' + comments + '*/'
 
+    def _print_PyccelSymbol(self, expr):
+        return expr
```
EmilyBourne, line 1443: Why is this added, but `_print_Symbol` not removed? Is this a new requirement? Do you know where it comes from?

pyccel/codegen/printing/fcode.py
```diff
@@ -277,6 +276,8 @@ def print_kind(self, expr):
         return iso_c_binding[self._print(expr.dtype)][expr.precision]
 
     # ============ Elements ============ #
+    def _print_PyccelSymbol(self, expr):
+        return expr
```
EmilyBourne, line 280: See above

pyccel/codegen/printing/pycode.py
```diff
@@ -368,6 +368,9 @@ def _print_PyccelNot(self, expr):
         a = self._print(expr.args[0])
         return 'not {}'.format(a)
 
+    def _print_PyccelSymbol(self, expr):
+        return expr
```
EmilyBourne, line 372: See above

pyccel/parser/semantic.py
```diff
@@ -2049,8 +2050,7 @@ def _visit_For(self, expr, **settings):
         elif isinstance(iterator, list):
             target = []
             for i in iterator:
-                name = str(i.name)
-                var  = Variable('int', name, rank=0)
+                var  = Variable('int', i, rank=0)
```
EmilyBourne, line 2053: Maybe we should rename `i`->`name` to keep the purpose clear?

pyccel/parser/semantic.py
```diff
@@ -2561,8 +2561,8 @@ def _visit_FunctionDef(self, expr, **settings):
             # to the body of the function
             body = self.garbage_collector(body)
 
-            args    = [self.get_variable(a.name) if isinstance(a, Variable) else self.get_function(str(a.name)) for a in args]
-            results = list(OrderedDict((a.name,self.get_variable(a.name)) for a in results).values())
+            args    = list(args)
```
EmilyBourne, line 2564: Isn't `args` already a list? Is this line necessary?

pyccel/ast/core.py
EmilyBourne: For example `Derivative`

EmilyBourne (changes requested):

pyccel/parser/semantic.py
MenoIy: I think its a tuple let me check

EmilyBourne: https://github.com/pyccel/pyccel/blob/master/pyccel/parser/semantic.py#L2462

pyccel/ast/variable.py
MenoIy: i remove it not because the symbol didn't have a strip method it because the name can be a dottedName

pyccel/ast/headers.py
MenoIy: my purpose was just to avoid a name clash between args and func_name

pyccel/codegen/printing/ccode.py
MenoIy: functions names are PyccelSymbol

pyccel/parser/semantic.py
MenoIy: ah true sorry

pyccel/parser/semantic.py
EmilyBourne: This can be done directly on the for loop rather than creating a temporary variable

pyccel/ast/headers.py
EmilyBourne: In that case should `d.name` be added to this set? (I.e. `function1`, `function2`, currently it just contains `high_int_int_1` I think)

pyccel/ast/core.py
EmilyBourne: (You can install pylint using pip)

pyccel/codegen/printing/ccode.py
MenoIy: 1 - space is add at get_declare_type

EmilyBourne: :( I guess we'll just have to live with the space then

pyccel/ast/core.py
MenoIy: I'm getting 
```
Your code has been rated at 10.00/10
```

EmilyBourne: Sorry, wrong command:
```
pylint ast/core.py --disable=all --enable=unused-import
```

pyccel/ast/headers.py
MenoIy: by ```d.name``` you mean each function name?

EmilyBourne: Yes, I meand [this `d`](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/headers.py#L340)
As I said currently this set contains `high_int_int_1` but it would make sense for it to contain at least `function1`, `function2`

MenoIy: but I think we only have the signature of function at this stage? we don't have arguments names that why  we create tmp name at 
```                   
arg_name = 'arg_{0}'.format(str(i))
```

EmilyBourne: Ah you are right :+1:

EmilyBourne (approved):

pyccel/codegen/printing/ccode.py
```diff
@@ -814,8 +814,9 @@ def function_signature(self, expr, print_arg_names = True):
             arg_code = 'void'
         else:
             arg_code = ', '.join('{}'.format(self.function_signature(i, False))
-                        if isinstance(i, FunctionAddress) else '{0}{1}'.format(self.get_declare_type(i), i.name)
-                        if print_arg_names else '{}'.format(self.get_declare_type(i)) for i in args)
+                        if isinstance(i, FunctionAddress)
+                        else '{0}'.format(self.get_declare_type(i)) + i.name if print_arg_names else ''
```
EmilyBourne, line 818: This line could use some parentheses

EmilyBourne: Is it:
```
('{0}'.format(self.get_declare_type(i)) + i.name) if print_arg_names else ''
```
or
```
'{0}'.format(self.get_declare_type(i)) + (i.name if print_arg_names else '')
```
(This may be obvious but I don't know the operator precedence by heart

EmilyBourne (approved):

pyccel/parser/syntactic.py
```diff
@@ -823,12 +821,12 @@ def fill_types(ls):
...
+            if not all(i[0]==j for j in i) or not isinstance(i[0], PyccelSymbol):
                 result_name, result_counter = create_variable(self._used_names,
                                                               prefix = 'Out',
                                                               counter = result_counter)
                 results.append(result_name)
-            elif isinstance(i[0], Symbol) and any(i[0].name==x.name for x in arguments):
+            elif isinstance(i[0], PyccelSymbol) and any(i[0]==x.name for x in arguments):
                 result_name, result_counter = create_variable(self._used_names,
                                                               prefix = 'Out',
                                                               counter = result_counter)
```
yguclu, line 832: We should avoid calling `i.__getitem__(0)` multiple times in this `if` block. This can be done by simply storing `i0 = i[0]` at the beginning. Further, there is a lot of duplicated code here, and it can be cleaned up:

```python
for r in zip(*returns):
    r0 = r[0]

    pyccel_symbol  = isinstance(r0, PyccelSymbol)
    same_results   = all(r0 == ri for ri in r)
    name_available = all(r0 != a.name for a in arguments)

    if pyccel_symbol and same_results and name_available:
        result_name = r0
    else:
        result_name, result_counter = create_variable(self._used_names, \
                                prefix = 'Out', counter = result_counter)

    results.append(result_name)
```

pyccel/parser/syntax/headers.py
```diff
@@ -385,7 +384,7 @@ def expr(self):
         arg_ = self.arg
         if isinstance(arg_, MacroList):
             return tuple(arg_.expr)
-        arg = Symbol(str(arg_))
+        arg = PyccelSymbol(str(arg_))
```
yguclu, line 387: Given that `PyccelSymbol` subclasses `str`, do we really need to convert its argument to a string?

pyccel/parser/syntax/headers.py
```diff
@@ -409,7 +408,7 @@ def __init__(self, **kwargs):
     @property
     def expr(self):
         name = str(self.macro)
-        arg  = Symbol(str(self.arg))
+        arg  = PyccelSymbol(str(self.arg))
```
yguclu, line 411: Same comment as above

pyccel/parser/syntax/headers.py
```diff
@@ -482,7 +481,7 @@ def expr(self):
             if isinstance(i, MacroStmt):
                 master_args.append(i.expr)
             else:
-                master_args.append(Symbol(str(i)))
+                master_args.append(PyccelSymbol(str(i)))
```
yguclu, line 484: Same comment as above

pyccel/symbolic/lambdify.py
```diff
@@ -66,7 +66,7 @@ def cse(expr):
             if not all(size):
                 raise ValueError('Unable to find range of index')
             name = str(vars_new[i].base)
-            var = Symbol(name)
+            var = PyccelSymbol(name)
```
yguclu, line 69: Cannot we merge these two lines into a single one, where we do not call `str()`?

```python
var = PyccelSymbol(vars_new[i].base)
```

pyccel/ast/core.py
```diff
@@ -427,7 +350,7 @@ def _sympystr(self, printer):
         return '{0} as {1}'.format(sstr(self.name), sstr(self.target))
 
     def __eq__(self, string):
-        if isinstance(string, (str, Symbol)):
+        if isinstance(string, (str, PyccelSymbol)):
```
yguclu, line 353: `PyccelSymbol` is now a subclass of `str`

pyccel/ast/core.py
```diff
@@ -1632,17 +1555,18 @@ class ValuedArgument(Basic):
...
     def __init__(self, expr, value, *, kwonly = False):
-        if isinstance(expr, str):
-            expr = Symbol(expr)
-
         # TODO should we turn back to Argument
 
-        if not isinstance(expr, (Symbol, Argument)):
+        if isinstance(expr, Argument):
+            self._name = expr.name
+        elif isinstance(expr, (str, PyccelSymbol)):
```
yguclu, line 1562: `PyccelSymbol` is now a subclass of `str`

pyccel/ast/core.py
```diff
@@ -1942,19 +1865,19 @@ def __init__(
...
         elif isinstance(name, (tuple, list)):
             name_ = []
             for i in name:
                 if isinstance(i, str):
-                    name = name + Symbol(i)
-                elif not isinstance(i, Symbol):
-                    raise TypeError('Function name must be Symbol or string'
+                    name = name + PyccelSymbol(i)
+                elif not isinstance(i, PyccelSymbol):
+                    raise TypeError('Function name must be PyccelSymbol or string'
```
yguclu, line 1875: This case is equivalent to an `else`, because `PyccelSymbol` is a subclass of `str` now.

pyccel/ast/core.py
```diff
@@ -1942,19 +1865,19 @@ def __init__(
...
                                     )
             name = tuple(name_)
-        elif not isinstance(name, Symbol):
+        elif not isinstance(name, PyccelSymbol):
 
-            raise TypeError('Function name must be Symbol or string')
+            raise TypeError('Function name must be PyccelSymbol or string')
```
yguclu, line 1880: This case is equivalent to an `else`, because `PyccelSymbol` is a subclass of `str` now.

pyccel/ast/core.py
```diff
@@ -2570,9 +2493,9 @@ def __init__(
         # name
 
         if isinstance(name, str):
-            name = Symbol(name)
-        elif not isinstance(name, Symbol):
-            raise TypeError('Function name must be Symbol or string')
+            name = PyccelSymbol(name)
+        elif not isinstance(name, PyccelSymbol):
+            raise TypeError('Function name must be PyccelSymbol or string')
```
yguclu, line 2498: This case is equivalent to an `else`, because `PyccelSymbol` is a subclass of `str` now.

pyccel/ast/core.py
```diff
@@ -2799,7 +2722,7 @@ def __init__(self, source, target = None, ignore_at_print = False):
         self._source = source
         self._target = []
         self._ignore_at_print = ignore_at_print
-        if isinstance(target, (str, Symbol, DottedName, AsName)):
+        if isinstance(target, (str, PyccelSymbol, DottedName, AsName)):
```
yguclu, line 2725: No need to add `PyccelSymbol` among the classes

yguclu (changes requested):
Good job @MenoIy, this change is cleaning up a lot of code! See comments

pyccel/ast/core.py
```diff
@@ -1942,19 +1865,19 @@ def __init__(
...
-            name = Symbol(name)
+            name = PyccelSymbol(name)
         elif isinstance(name, (tuple, list)):
             name_ = []
             for i in name:
                 if isinstance(i, str):
-                    name = name + Symbol(i)
-                elif not isinstance(i, Symbol):
-                    raise TypeError('Function name must be Symbol or string'
+                    name = name + PyccelSymbol(i)
```
EmilyBourne, line 1873: This code is also strange. Shouldn't it be `name_.append(PyccelSymbol(i))`?

pyccel/ast/headers.py
```diff
@@ -581,8 +587,8 @@ class MacroFunction(Header):
     """."""
 
     def __init__(self, name, args, master, master_args, results=None):
-        if not isinstance(name, (str, Symbol)):
-            raise TypeError('name must be of type str or Symbol')
+        if not isinstance(name, (str, PyccelSymbol)):
+            raise TypeError('name must be of type str or PyccelSymbol')
```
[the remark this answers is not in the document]
MenoIy: a814af0

pyccel/ast/headers.py
```diff
@@ -728,7 +734,7 @@ class MacroVariable(Header):
     """."""
 
     def __init__(self, name,  master):
-        if not isinstance(name, (str, Symbol, DottedName)):
+        if not isinstance(name, (str, PyccelSymbol, DottedName)):
```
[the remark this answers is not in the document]
MenoIy: 1341226

pyccel/ast/macros.py
```diff
@@ -28,7 +29,7 @@ class Macro(AtomicExpr, PyccelAstNode):
     _name = '__UNDEFINED__'
 
     def __init__(self, argument):
-        if not isinstance(argument, Symbol):
+        if not isinstance(argument, (PyccelSymbol, Variable)):
             raise TypeError("Argument must be a symbol not {}".format(type(argument)))
```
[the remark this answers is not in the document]
MenoIy: f1de584

pyccel/codegen/printing/ccode.py
```diff
@@ -787,7 +785,22 @@ def _print_NativeComplex(self, expr):
...
+        """Extract from function definition all the information
+        (name, input, output) needed to create the signature
+
+        parameters
+        ----------
+            expr      : FunctionDef
+                the function defintion
+
+            print_arg_names : Bool
+
```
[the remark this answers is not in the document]
MenoIy: 6aa7150

pyccel/codegen/printing/ccode.py
```diff
@@ -787,7 +785,22 @@ def _print_NativeComplex(self, expr):
     def _print_NativeString(self, expr):
         return 'string'
 
-    def function_signature(self, expr):
+    def function_signature(self, expr, print_arg_names = True):
+        """Extract from function definition all the information
+        (name, input, output) needed to create the signature
+
+        parameters
```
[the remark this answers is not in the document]
MenoIy: 5f7de03

pyccel/codegen/printing/fcode.py
```diff
@@ -410,14 +411,14 @@ def _print_Import(self, expr):
...
             elif isinstance(i, str):
                 line = '{prefix} {target}'.format(prefix=prefix,
-                                                  target=str(i))
+                                                  target=i)
 
-            elif isinstance(i, Symbol):
+            elif isinstance(i, PyccelSymbol):
                 line = '{prefix} {target}'.format(prefix=prefix,
-                                                  target=str(i.name))
+                                                  target=i)
```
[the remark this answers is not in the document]
MenoIy: 63c699c

pyccel/parser/syntactic.py
MenoIy: 049c142

pyccel/parser/syntax/headers.py
MenoIy: 25403cf

pyccel/parser/syntax/headers.py
MenoIy: 2fc1b77

pyccel/parser/syntax/headers.py
MenoIy: fa7055c

pyccel/symbolic/lambdify.py
MenoIy: 5f5d919

pyccel/ast/core.py
MenoIy: ec38596

pyccel/ast/core.py
MenoIy: 08d2e4f

MenoIy: 549cda6

pyccel/ast/core.py
MenoIy: 549cda6

pyccel/ast/core.py
MenoIy: 8092411

pyccel/ast/core.py
MenoIy: 7bbafd8

pyccel/ast/core.py
MenoIy: 08d2e4f

pyccel/ast/core.py
MenoIy: 0a67d23

pyccel/ast/utilities.py
```diff
@@ -53,7 +53,7 @@ def builtin_function(expr, args=None):
         return dic[name](*args)
 
     if name == 'map':
-        func = str(expr.args[0].name)
+        func = expr.args[0]
         args = [func]+list(args[1:])
         return PythonMap(*args)
```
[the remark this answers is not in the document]
EmilyBourne: `args` is an argument to the function. It is not `expr.args`

pyccel/codegen/printing/ccode.py
```diff
@@ -787,7 +785,22 @@ def _print_NativeComplex(self, expr):
...
+        ----------
+            expr      : FunctionDef
+                the function defintion
+
+            print_arg_names : Bool
+
+        Return
+        ------
+            String
+                Signature of the function
```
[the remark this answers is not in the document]
MenoIy: 71d44cf

pyccel/parser/base.py
```diff
@@ -522,7 +522,7 @@ def insert_import(self, expr):
             source = str(expr.source)
             if source not in pyccel_builtin_import_registery:
                 for t in expr.target:
-                    name = [str(t)]
+                    name = [t]
                     if not source in container.keys():
                         container[source] = []
                     container[source] += name
```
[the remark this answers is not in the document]
MenoIy: euh ,
1 - container[source] = [] was set to empty  list  so we can append elements to it later 
2 - the cycle purpose is to collect all targets into a list (the one set as empty [])
3 - improvement set default of ```container[source]`` outside loop and append target directly into the list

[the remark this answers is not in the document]
MenoIy:
```
            source = str(expr.source)
            if source not in pyccel_builtin_import_registery:
                if not source in container.keys():
                    container[source] = []
                container[source] += expr.target
```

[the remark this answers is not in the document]
EmilyBourne: In this case we can even remove the for loop entirely right?
```
if not source in container.keys():
    container[source] = []
container[source].extend(expr.target)
```

pyccel/codegen/printing/fcode.py
```diff
@@ -53,7 +52,7 @@
 from pyccel.ast.datatypes import NativeRange, NativeTuple
 from pyccel.ast.datatypes import CustomDataType
 
-from pyccel.ast.internals import Slice
+from pyccel.ast.internals import Slice, PyccelSymbol
```
EmilyBourne, line 55: Codacy was rasing an `unused-import` error for this. I'm not sure where it's gone but it still seems to be true

MenoIy: its not gone

EmilyBourne: It was gone in commit 71d44cf but it's back ^^

pyccel/parser/base.py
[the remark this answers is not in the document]
MenoIy: ccb0b31

EmilyBourne (approved):

pyccel/ast/utilities.py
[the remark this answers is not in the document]
yguclu: You're right, I read `expr.args[1:]` but there is no `expr` there

pyccel/parser/base.py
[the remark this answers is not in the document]
yguclu: Thanks!

pyccel/ast/utilities.py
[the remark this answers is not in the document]
yguclu: Nevertheless, can't we write this with just one (clear!) line?

```python
if name == 'map':
    return PythonMap(expr.args[0], *args[1:])
```

By the way, I don't see why `args[0]` is not used here. If `args[0]` coincides with `expr.args[0]`, then my original comment was correct... Could you check please?

yguclu (changes requested):
Thanks for the fixes. I only have one remaining comment, see https://github.com/pyccel/pyccel/pull/716#discussion_r573282280.

pyccel/ast/utilities.py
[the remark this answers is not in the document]
EmilyBourne: `args[0]` contains a `FunctionDef` while `expr.args[0]` contains a `Symbol`. This is because it is the `FunctionDef` that is collected when we visit a `Symbol` (when we visit the argument). This should make sense, however the map is then usually used in a `For` loop. Personally I think this section is poorly written.

In a `For` block there are a few steps:
- Visit the iterable to get information about the type of iterable (map/zip/variable/etc)
- Construct a syntactic node representing the collection of the iterable at each pass of the loop from the visited iterable
    E.g. For a Variable the created node is
    ```
    assign = Assign(iterator, IndexedElement(iterable, indx))
    ```
    For me this step has a logical flaw as Syntactic objects are created from a Semantic object, despite the fact that the semantic stage does not work if you pass it a semantically treated object.
    For the map this is seen [here](https://github.com/pyccel/pyccel/blob/master/pyccel/parser/semantic.py#L2003-L2010) as we construct a `FunctionCall` from the arguments of the map. The line in `ast/headers.py` ensures that the first argument is a syntactic object and not a semantic object. This object will always cause a problem if it is a `FunctionDef` instead of a `Symbol` but the other arguments could too if they were more complicated than `Variable`
- Visit the newly constructed body

We can make the changes above by also changing [this line](https://github.com/pyccel/pyccel/blob/master/pyccel/parser/semantic.py#L2005) to `func   = iterable.args[0].name`

Ideally we would also ensure that all the arguments passed are syntactic objects, but I think this deserves a PR of its own to clean up `_visit_For`

saidctb (approved):

yguclu (approved):

Issue #660, opened 2021-01-06
Use mutable objects in classes instead of indexing _args


EmilyBourne:
Most classes use the `_args` variable (or `.args` property) provided by the sympy inheritance to store their arguments. This means that the arguments are immutable and several of our properties are hard to maintain, e.g:
```
@property
def thing(self):
    return self._args[10]
```

E.g.
```
class While(Basic):
    def __new__(cls, test, body, local_vars=[]):
        test = sympify(test, locals=local_sympify)

        if PyccelAstNode.stage == 'semantic':
            if test.dtype is not NativeBool():
                test = PythonBool(test)

        if iterable(body):
            body = CodeBlock((sympify(i, locals=local_sympify) for i in body))
        elif not isinstance(body,CodeBlock):
            raise TypeError('body must be an iterable or a CodeBlock')
        return Basic.__new__(cls, test, body, local_vars)

    @property
    def test(self):
        return self._args[0]

    @property
    def body(self):
        return self._args[1]

    @property
    def local_vars(self):
        return self._args[2]
```
Should instead be:
```
class While(Basic):
    def __init__(self, test, body, local_vars=[]):
        test = sympify(test, locals=local_sympify)

        if PyccelAstNode.stage == 'semantic':
            if test.dtype is not NativeBool():
                test = PythonBool(test)

        if iterable(body):
            body = CodeBlock((sympify(i, locals=local_sympify) for i in body))
        elif not isinstance(body,CodeBlock):
            raise TypeError('body must be an iterable or a CodeBlock')
            
        self._test = test
        self._body = body
        self._local_vars = local_vars
        Basic.__init__(self)

    @property
    def test(self):
        return self._test

    @property
    def body(self):
        return self._body

    @property
    def local_vars(self):
        return self._local_vars
```

This needs fixing for all files in the folder pyccel/ast:
- [x] basic
- [x] bind_c
- [x] bitwise_operators.py
- [ ] builtins.py
- [x] core.py
- [x] cwrapper.py
- [ ] datatypes.py
- [x] functionalexpr.py
- [x] headers.py
- [x] internals.py
- [x] itertoolsext.py
- [ ] literals.py
- [ ] macros.py
- [x] mathext.py
- [ ] numpyext.py
- [x] operators.py
- [x] sympy_helper.py
- [x] utilities.py
- [ ] variable.py

EmilyBourne:
[...]

PR #717, opened 2021-01-29
Make more AST classes mutable

EmilyBourne:
Related to #660. This PR makes the classes in the following files mutable:
- `functionalexpr.py`
- `headers.py`
- `itertoolsext.py`

It also makes some small improvements to `headers.py`:
- Remove TODOs which are no longer relevant
- Remove properties from `MethodHeader` which are inherited from `FunctionHeader`
- Fix bug in function `__reduce_ex__` of `MethodHeader`

### DIFF
--- a/pyccel/ast/headers.py
+++ b/pyccel/ast/headers.py
@@ -219,7 +219,7 @@ class FunctionHeader(Header):
     Examples
     --------
 
-    >>> from pyccel.ast.core import FunctionHeader
+    >>> from pyccel.ast.headers import FunctionHeader
     >>> FunctionHeader('f', ['double'])
     FunctionHeader(f, [(NativeDouble(), [])])
     """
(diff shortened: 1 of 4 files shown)


MenoIy (approved):

saidctb (approved):

pyccel/ast/headers.py
```diff
@@ -510,7 +497,7 @@ def __reduce_ex__(self, i):
            and its arguments
            """
 
-        args = (self.name,
+        args = (self.name.split('.'),
```
yguclu, line 500: Was there a bug here?

EmilyBourne: I hadn't realised it but yes. `__reduce_ex__` should return a tuple which contains valid arguments for the class. `MethodHeader.__init__` requires the argument `name` to be a tuple or a list not a string

yguclu: Well done! Should we add a note about it in the PR's description?

yguclu (approved):

PR #718, opened 2021-01-30
Improve `pyccel.ast.basic.Basic` to contain context information and helper functions

### LINKED ISSUE #663: Find and remove remaining sympy objects with pyccel arguments

EmilyBourne:
Add `attribute_nodes` and `user_nodes` to `pyccel.ast.basic.Basic` and provide helper functions which replace previously used sympy functionality. Fixes issue #694.

Every class in the Pyccel AST now follows the following rules:
1. It inherits from `pyccel.ast.basic.Basic`
2. It has the `_attribute_nodes` attribute detailing which of its attributes can contain other ast nodes
3. It calls `super().__init__()` after initialising its attributes

As nodes are now aware of their context, it is important to call `invalidate_node` if a node is created but not used

**New concepts**
`pyccel.ast.basic.Basic` has 2 new attributes:
- `_attribute_nodes` : This is a tuple containing the names of all the attributes which can be `Basic` objects
    E.g. For an `Assign` : 
    ```
    _attribute_nodes = ('_lhs','_rhs')
    ```
- `_user_nodes` : This is a list of all the objects which use this class (i.e. for which this class is an object referenced by the first object's `_attribute_nodes`)
    E.g. For an `Assign` : 
    ```
    _user_nodes = [ CodeBlock(...) ]
    ```
    For a `CodeBlock` : 
    ```
    _user_nodes = [ FunctionDef(...) ]
    ```
    For a different `CodeBlock` : 
    ```
    _user_nodes = [ For(...) ]
    ```
    For a `Variable` :
    ```
    _user_nodes = [Allocate(..), Assign(..), PyccelAdd(..), Deallocate(..)]
    ```
    For most classes `_user_nodes` will only contain one object. A notable exception to this rule is `Variable`

**Commit Summary**
- Add functions to `pyccel.ast.basic.Basic`
     - get_user_nodes
     - get_attribute_nodes
     - substitute
     - is_atomic
     - get_direct_user_nodes
     - remove_user_node
     - invalidate_node
- Add `_attribute_nodes` to every class
- Remove `_atomic` (superseded by get_attribute_nodes)
- Remove unnecessary `Basic` inheritance
    - NativeOp
- Remove `Dict` (fixes #663)
- Add new method `add_local_var` to `FunctionDef` to modify tuple
- Add `__eq__` to classes:
    - `PyccelArraySize`
    - `Basic` (sympy.Basic uses `_args` for this purpose)
    - `Nil`
- Prefer `super().__init__` for uniform syntax and to avoid needing to update extra lines if class names change
- Do not create pseudo-sympy nodes in `pyccel_to_sympy`. Instead handle only objects which have an equivalent in sympy
- Remove unused printing functions `_print_MatrixElement`
- In complexity, use sympy only where necessary
- Ensure fst is set when it is available. This is done in `self._visit` instead of in every `_visit`
- Rename `ClassDef.parent`->`ClassDef.superclass`
- Use `user_node` information to accurately set `Allocate.status`
- Replace calls to sympy functions (e.g. `Tuple(ob).atoms` with the `Basic` equivalent)
- Remove `Dict`. This is a sympy class with no pyccel equivalent. It was handled up to the semantic stage. Tests are moved to the errors folder
- Add calls to `invalidate_node`


EmilyBourne:
Failing due to master merge of `Variable` `__init__` calls. This will be fixed when the master is merged in after PR #716 has been merged

pyccel/ast/variable.py
```diff
@@ -521,6 +526,10 @@ class ValuedVariable(Variable):
     >>> n
     n := 4
     """
+    #TODO: Can this have children?
+    # We have a problem when we do `dtype=float` as `float`
+    # becomes a class, not a Pyccel object
+    #_attribute_nodes = ('_value',)
```
EmilyBourne, line 532: This should be discussed before merging

EmilyBourne: A `ValuedVariable`  is used for default arguments of a function. E.g.:
`def f(a = 3):`
the `_value`  is `3`  which should be in the `_attribute_nodes` , but when we do:
`x = empty(dtype=float)`
`dtype=float`  is a `ValuedVariable`  and `float`  is a type not a `Basic`  so it shouldn't be in `_attribute_nodes`

pyccel/ast/basic.py
```diff
@@ -23,13 +27,217 @@ def __new__(cls, *args, **kwargs):
...
+                      The object which will be inserted instead
+        excluded_nodes : tuple of types
+                      Types for which substitute should not be called
+        """
+        if isinstance(original, tuple):
+            assert(isinstance(replacement, tuple))
+            assert(len(original) == len(replacement))
+        else:
+            original = (original,)
+            substitute = (substitute,)
```
MenoIy, line 146: substitute where did you initialize this one ?

EmilyBourne: You are right, this is the wrong name. It should be `substitute`->`replacement`
Fixed in 96906ae

pyccel/ast/basic.py
```diff
@@ -23,13 +27,217 @@ def __new__(cls, *args, **kwargs):
...
+            if v in original:
+                idx = original.index(v)
+                v.remove_user_node(self)
+                setattr(self, n, replacement[idx])
+                replacement[idx].user_nodes = self
+            elif isinstance(v, tuple):
+                new_v = []
+                for vi in v:
+                    if vi in original:
+                        idx = original.index(v)
```
MenoIy, line 159: did you mean ? ```idx = original.index(vi)```

EmilyBourne: Fixed in 8b977fc2

pyccel/ast/basic.py
```diff
@@ -23,13 +27,217 @@ def __new__(cls, *args, **kwargs):
...
+        ----------
+        search_type : ClassType or tuple of ClassTypes
+                      The types which we are looking for
+        excluded_nodes : tuple of types
+                      Types for which get_user_nodes should not be called
+
+        Results
+        -------
+        Boolean : True if one of the user nodes is an instance of
+                  the class in the argument
```
MenoIy, line 78: your are returning list of nodes , maybe you need to update doc

pyccel/ast/basic.py
```diff
@@ -23,13 +27,217 @@ def __new__(cls, *args, **kwargs):
...
+        """ Indicates whether the object has any attribute nodes
+        """
+        return bool(self._attribute_nodes)
+
     def set_fst(self, fst):
         """Sets the python.ast fst."""
-        self._fst = fst
+        if not isinstance(fst, ast.AST):
+            raise TypeError("Fst must be an AST object, not {}".format(type(fst)))
+        assert(isinstance(fst, ast.AST))
```
MenoIy, line 180: useless assert ?

pyccel/ast/basic.py
```diff
@@ -9,11 +9,15 @@
...
 - PyccelAstNode which describes each PyccelAstNode
 """
-
+import ast
 from sympy.core.basic import Basic as sp_Basic
 
 __all__ = ('Basic', 'PyccelAstNode')
 
+dict_keys   = {}.keys().__class__
+dict_values = {}.values().__class__
```
yguclu, line 18: Is this any different than writing

```python
dict_keys   = type({}.keys())
dict_values = type({}.values())
```
?

pyccel/ast/basic.py
```diff
@@ -23,13 +27,217 @@ def __new__(cls, *args, **kwargs):
...
+        """ Indicates whether the object has any attribute nodes
+        """
+        return bool(self._attribute_nodes)
+
     def set_fst(self, fst):
         """Sets the python.ast fst."""
-        self._fst = fst
+        if not isinstance(fst, ast.AST):
+            raise TypeError("Fst must be an AST object, not {}".format(type(fst)))
+        assert(isinstance(fst, ast.AST))
```
yguclu, line 180: Why do you need this assert?

pyccel/ast/basic.py
```diff
@@ -9,11 +9,15 @@
...
 from sympy.core.basic import Basic as sp_Basic
 
 __all__ = ('Basic', 'PyccelAstNode')
 
+dict_keys   = {}.keys().__class__
+dict_values = {}.values().__class__
+iterable_types = (list, tuple, dict_keys, dict_values)
+
 #==============================================================================
 class Basic(sp_Basic):
```
yguclu, line 22: I think that you should provide a few dedicated unit tests where you check that the methods of this `Basic` class work as intended. Namely:

- `get_user_nodes`
- `get_attribute_nodes`
- `get_direct_user_nodes`
- `user_nodes`
- `remove_user_node`

pyccel/ast/basic.py
```diff
@@ -23,13 +27,217 @@ def __new__(cls, *args, **kwargs):
...
+        return [p for p in self._user_nodes if condition(p)]
+
+    @property
+    def user_nodes(self):
+        """ Get the current user_nodes of the object
+        """
+        if self._user_nodes:
+            return self._user_nodes[-1]
+        else:
+            return None
```
yguclu, line 216: What is the purpose of this property exactly? Shouldn't this be a method called `get_current_user_node` (singular)?

pyccel/ast/basic.py
```diff
@@ -23,13 +27,217 @@ def __new__(cls, *args, **kwargs):
...
+
+    @user_nodes.setter
+    def user_nodes(self, user_nodes):
+        self._user_nodes.append(user_nodes)
```
yguclu, line 220: A find this setter strange: you will write `obj.user_nodes = node` instead of the more natural `obj.user_nodes += node`... So I recommend using a dedicated method called `set_current_user_node` (see comment above).

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -915,7 +915,7 @@ def _print_Module(self, expr):
             static_funcs = expr.funcs
         function_signatures = '\n'.join('{};'.format(self.function_signature(f)) for f in static_funcs)
 
-        interfaces = expr.interfaces
+        interfaces = list(expr.interfaces)
         interface_funcs = [f.name for i in interfaces for f in i.functions]
         funcs = interfaces + [f for f in expr.funcs if f.name not in interface_funcs]
```
yguclu, line 920: I don't see why we should create a new list from `expr.interfaces`, which is iterable already:

```python
interfaces = expr.interfaces  # not needed, renaming is just for clarity
interface_funcs = [f.name for i in interfaces for f in i.functions]
funcs = [*interfaces, *(f for f in expr.funcs if f.name not in interface_funcs)]
```

MenoIy (changes requested):

pyccel/parser/semantic.py
```diff
@@ -1547,17 +1547,22 @@ def _assign_lhs_variable(self, lhs, d_var, rhs, new_expressions, is_augassign, *
                                     self._current_fst_node.col_offset))
 
                         else:
-                            # TODO [YG, 04.11.2020] If we could be sure that the
-                            # array was not created in an if-then-else block, we
-                            # would use status='allocated' instead.
+                            previous_allocations = var.get_direct_user_nodes(lambda p: isinstance(p, Allocate))
```
yguclu, line 1550: Are we sure that we never get an empty list here?

pyccel/parser/semantic.py
```diff
@@ -2602,7 +2602,7 @@ def _visit_FunctionDef(self, expr, **settings):
             assigned     = [a.name for a in all_assigned if a.rank > 0]
             all_assigned = [i.name for i in all_assigned]
 
-            apps = list(Tuple(*body.body).atoms(FunctionCall))
+            apps = list(body.get_attribute_nodes(FunctionCall))
```
yguclu, line 2605: Why do we need to convert to list?

pyccel/ast/core.py
```diff
@@ -2132,6 +2143,16 @@ def rename(self, newname):
...
+    def add_local_var(self, var):
+        """
+        Add a new variable to the local variables tuple
+
+        Parameters
+        ----------
+        var : Variable
+              The new local variable
+        """
+        self._local_vars = self._local_vars + (var,)
```
yguclu, line 2155: Is it not easier to use a list to store the local variables?

yguclu (changes requested):
Excellent job @EmilyBourne!

I have a couple of comments.

pyccel/ast/basic.py
EmilyBourne: No, it is the same. I can use this syntax if you think it is clearer

pyccel/ast/basic.py
```diff
@@ -23,13 +27,217 @@ def __new__(cls, *args, **kwargs):
...
+                for vi in v:
+                    if vi in original:
+                        idx = original.index(vi)
+                        vi.remove_user_node(self)
+                        new_v.append(replacement[idx])
+                        replacement[idx].user_nodes = self
+                    else:
+                        new_v.append(vi)
+                        vi.substitute(original, replacement, excluded_nodes)
+                setattr(self, n, v)
```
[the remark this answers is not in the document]
EmilyBourne: Fixed in 9a1a43ee

pyccel/ast/basic.py
```diff
@@ -23,13 +27,217 @@ def __new__(cls, *args, **kwargs):
...
+                new_v = []
+                for vi in v:
+                    if vi in original:
+                        idx = original.index(vi)
+                        vi.remove_user_node(self)
+                        new_v.append(replacement[idx])
+                        replacement[idx].user_nodes = self
+                    else:
+                        new_v.append(vi)
+                        vi.substitute(original, replacement, excluded_nodes)
```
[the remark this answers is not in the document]
EmilyBourne: Fixed in 9a1a43ee

pyccel/ast/basic.py
EmilyBourne: Fixed in 9a1a43ee

pyccel/ast/basic.py
EmilyBourne: Fixed in 9a1a43ee

pyccel/ast/basic.py
EmilyBourne: Fixed in 9a1a43ee

pyccel/ast/basic.py
EmilyBourne: Fixed in 9a1a43ee

pyccel/ast/basic.py
EmilyBourne: You are right, I will do this

pyccel/ast/basic.py
EmilyBourne: I have renamed this `current_user_node` (9a1a43ee)

pyccel/ast/basic.py
EmilyBourne: I have renamed this `current_user_node` (9a1a43ee)

EmilyBourne: `obj.user_nodes += node` is more natural when you remember that `user_nodes` is a list, but in almost all cases `user_nodes` only contains 1 object so I think it is more natural to interact with it this way (I think it should be all classes except, `Variable`, `FunctionDef` and `ClassDef`).
In practice when we want to interact with more than the current node we should not be interacting with the `_user_nodes` directly anyway

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: You are right, I hadn't considered this syntax

pyccel/parser/semantic.py
EmilyBourne: If we reach here then the variable already exists, which means it must have been allocated. If we get an empty list then it is because we are missing an `Allocate`. This may change however when functions can return arrays

pyccel/ast/core.py
EmilyBourne: All `_attribute_nodes` are tuples. I could increase all the tests to include any iterable if you prefer? I don't think this is costly though as we only use it when creating additional variables in the printing stage which doesn't happen very often

pyccel/parser/semantic.py
EmilyBourne: We don't. Fixed in ad837cff

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: Fixed in ad837cff

EmilyBourne: 80339504

pyccel/ast/basic.py
EmilyBourne: I have removed `user_nodes`

I am not sure how to test `remove_user_node`. It should be tested within `substitute` though. I originally named this function `_remove_user_node` as it shouldn't be called from outside `Basic` but codacy didn't let me keep that name and call the function for something other than `self`

pyccel/ast/core.py
yguclu: Don't worry, it's not important.

pyccel/ast/basic.py
yguclu: OK, thanks!

pyccel/ast/basic.py
yguclu: All right

pyccel/ast/basic.py
yguclu: Yes please, I think that using `type()` is more natural.

pyccel/parser/semantic.py
yguclu: I see. Then I think we should raise a nice meaningful error if for some reason `previous_allocations` is empty. This might well save us an headache later.

pyccel/ast/basic.py
EmilyBourne: Fixed in 9a1a43e

pyccel/parser/semantic.py
EmilyBourne: Done here: e2d6b82

pyccel/ast/basic.py
EmilyBourne: The tests are in the folder `tests/ast/`

MenoIy (approved):

pyccel/ast/basic.py
yguclu: I cannot find the changes..

pyccel/ast/basic.py
yguclu: Perfect, thanks!

pyccel/ast/basic.py
EmilyBourne: Hmm, I don't know where they went. Fixed now : 590c687

yguclu: Thanks

pyccel/ast/variable.py
EmilyBourne: @yguclu 

As I had to add exceptions in my code for `PyccelSymbol` which is immutable and therefore cannot inherit from `Basic` I have added the type `type` to that list of exceptions. This fixes the issue described above.
This is done in these 2 commits: 5b32617  1e1ec11

yguclu (approved):

pyccel/ast/basic.py
```diff
@@ -69,6 +69,8 @@ def __init__(self):
                 c.set_current_user_node(self)
 
     def ignore(self, c):
+        """ Indicates if a node should be ignored when recungg
```
yguclu, line 72: "when recungg"?

yguclu (approved):

pyccel/ast/basic.py
EmilyBourne: Fixed ab26ad9

PR #719, opened 2021-02-01
Add tests for numpy functions

mostafaMamoni:
- Add unit tests for the Numpy types:
   -  int
   -  float
   -  double
   -  float32
   -  float64
   -  int32
   -  int64
   -  complex128
   -  complex64
- Add unit tests for the Numpy functions
   -  mod
   -  matmul
   -  prod (also covers `numpy.product` according to [this](https://stackoverflow.com/questions/49863633/numpy-product-vs-numpy-prod-vs-ndarray-prod))
   -  linspace
   -  where
- Remove tests for `diag` and `cross`, which are not properly working
- Add missing function printers in Python code for `linspace` and `matmul`


pyccel/ast/numpyext.py
```diff
@@ -9,7 +9,7 @@
 
 from .basic          import PyccelAstNode
 from .builtins       import (PythonInt, PythonBool, PythonFloat, PythonTuple,
-                             PythonComplex, PythonReal, PythonImag, PythonList)
+                             PythonComplex, PythonReal, PythonAbs, PythonImag, PythonList)
```
EmilyBourne, line 12: Is this a bug fix?

mostafaMamoni: Ah sorry, i missed this from #790.

mostafaMamoni: This PR should modify only `test_numpy_functions.py`, except the little new line in ccode.py

pyccel/codegen/printing/ccode.py
```diff
@@ -1201,7 +1201,7 @@ def _print_Return(self, expr):
             # make sure that stmt contains one assign node.
             assert(len(last_assign)==1)
             variables = last_assign[0].rhs.get_attribute_nodes(Variable, excluded_nodes=(FunctionDef,))
-            unneeded_var = not any(b not in vars_in_deallocate_nodes for b in variables)
+            unneeded_var = not any(b not in vars_in_deallocate_nodes and not b.is_argument for b in variables)
```
EmilyBourne, line 1204: Why do you need this? Are we deallocating arguments?!

mostafaMamoni: i'm sorry, the problem was in the first attempt to fix, it should  `unneeded_var = not any(b in vars_in_deallocate_nodes for b in variables)`

pyccel/codegen/printing/ccode.py
```diff
@@ -1191,7 +1191,8 @@ def _print_Return(self, expr):
         if expr.stmt:
             # get Assign nodes form the CodeBlock object expr.stmt.
             last_assign = expr.stmt.get_attribute_nodes(Assign)
-            vars_in_deallocate_nodes = expr.stmt.get_attribute_nodes(Deallocate, excluded_nodes=(Assign,))
+            deallocate_nodes = expr.stmt.get_attribute_nodes(Deallocate)
```
EmilyBourne, line 1194: Why did you remove `excluded_nodes=(Assign,)`? We can't assign a `Deallocate` so keeping this can make the code run faster as we don't need to descend into an assign tree

EmilyBourne:
There are collisions with the master branch here. Please resolve them

EmilyBourne:
Does this PR fix other issues? E.g. #792 ? If so, please update the description to reflect this

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5799,3 +5808,805 @@ def get_prod(arr):
...
+    @types('float')
+    @types('float32')
+    @types('float64')
+    @types('complex64')
+    @types('complex128')
+    def get_linspace(start):
+        from numpy import linspace, shape
+        stop = start + 7
+        NumberOfSamplesToGenerate = 7
+        b = linspace(start, stop, NumberOfSamplesToGenerate)
```
EmilyBourne, line 6397: Normally variables have names starting with a lower case name. As you can see from the highlighting giving your variable a name starting with an upper case letter makes github think this is a class name

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5799,3 +5808,805 @@ def get_prod(arr):
...
+    assert (f_complex128(cmplx128) == get_matmul(cmplx128))
+
+
+@pytest.mark.parametrize( 'language', (
+        pytest.param("fortran", marks = [pytest.mark.fortran,
+            pytest.mark.skip(reason="Still under maintenance, See #770")]),
+        pytest.param("c", marks = [
+            pytest.mark.skip(reason="Needs a C printer see https://github.com/pyccel/pyccel/issues/791"),
+            pytest.mark.c]
+        )
```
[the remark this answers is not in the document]
EmilyBourne: Same below

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5799,3 +5808,805 @@ def get_prod(arr):
...
+    def get_chosen_elements(arr):
+        from numpy import where, shape, array
+        arr = array(arr)
+        a = where(arr > 5, arr, arr*2)
+        s = shape(a)
+        return len(s), s[0], a[1], a[0]
+
+    bl = randint(0, 1, size=(5), dtype= bool)
+
+    integer8 = randint(min_int8, max_int8, size=(5), dtype=np.int8)
```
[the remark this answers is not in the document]
EmilyBourne: There are several occasions of this in the file, it is better to either use `5` or `(5,)`

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5799,3 +5808,805 @@ def get_prod(arr):
...
+    fl = uniform(min_float / 2, max_float / 2, size=(5))
+    fl32 = uniform(min_float32, max_float32, size=(5))
+    fl32 = np.float32(fl32)
+    fl64 = uniform(min_float64 / 2, max_float64 / 2, size=(5))
+
+    cmplx128_from_float32 = uniform(low=min_float32 / 2, high=max_float32 / 2, size=5) + uniform(low=min_float32 / 2, high=max_float32 / 2, size=5) * 1j
+    # the result of the last operation is a Python complex type which has 8 bytes in the alignment,
+    # that's why we need to convert it to a numpy.complex64 the needed type.
+    cmplx64 = np.complex64(cmplx128_from_float32)
+    cmplx128 = uniform(low=min_float64 / 2, high=max_float64 / 2, size=5) + uniform(low=min_float64 / 2, high=max_float64 / 2, size=5) * 1j
```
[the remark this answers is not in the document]
EmilyBourne: There are a few other tests like this where only the floats and complexes are divided by 2

EmilyBourne (changes requested):

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5799,3 +5808,805 @@ def get_prod(arr):
 
     assert (f_complex64(cmplx64) == get_prod(cmplx64))
     assert (f_complex128(cmplx128) == get_prod(cmplx128))
+
+@pytest.mark.parametrize( 'language', (
+        pytest.param("fortran", marks = [pytest.mark.fortran,
+            pytest.mark.skip(reason="Pyccel raises 'ValueError: Incompatible rank in variable allocation'")]),
```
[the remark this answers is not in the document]
mostafaMamoni: Pyccel seems does not to accept array 1d for numpy.matmul even with 2d it raises `TypeError: cannot determine the type of <pyccel.ast.numpyext.NumpyMatmul object at 0x1218b9d60>`

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5799,3 +5808,805 @@ def get_prod(arr):
...
+    @types('int32[:]')
+    @types('int64[:]')
+    @types('float[:]')
+    @types('float32[:]')
+    @types('float64[:]')
+    @types('complex64[:]')
+    @types('complex128[:]')
+    def get_chosen_elements(arr):
+        from numpy import where, shape, array
+        arr = array(arr)
```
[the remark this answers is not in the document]
mostafaMamoni: sorry, i missed that while cleaning.

tests/epyccel/recognised_functions/test_numpy_funcs.py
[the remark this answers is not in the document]
mostafaMamoni: i will fix that as the previous PRs.

tests/epyccel/recognised_functions/test_numpy_funcs.py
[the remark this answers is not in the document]
EmilyBourne: > Pyccel seems does not to accept array 1d

Yes, this is issue #767 

> even with 2d it raises `TypeError: cannot determine the type of <pyccel.ast.numpyext.NumpyMatmul object at 0x1218b9d60>`

Can you give an example. I am dubious that all matmul cases fail as [all these test pass](https://github.com/pyccel/pyccel/blob/master/tests/epyccel/test_arrays.py#L1609-L1699) using [all these functions](https://github.com/pyccel/pyccel/blob/master/tests/epyccel/modules/arrays.py#L613-L636)

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -4914,6 +4920,9 @@ def get_float32(a):
     assert f_bl_true_output == test_bool_true_output
     assert f_bl_false_output == test_bool_false_output
 
+    assert (type(f_bl_true_output) == type(test_bool_true_output.item()))
+    assert (type(f_bl_false_output) == type(test_bool_false_output.item()))
```
EmilyBourne, line 4924: Please use the new `matching_types` function whenever you need `.item` otherwise you will break python tests

tests/epyccel/recognised_functions/test_numpy_funcs.py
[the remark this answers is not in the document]
mostafaMamoni: i think it is just about types.

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -6065,6 +6069,10 @@ def get_matmul(arr):
         pytest.param("c", marks = [
             pytest.mark.skip(reason="Needs a C printer see https://github.com/pyccel/pyccel/issues/791"),
             pytest.mark.c]
+        ),
+        pytest.param("python", marks = [
+            pytest.mark.python,
+            pytest.mark.skip(reason="Missing Python printer")]
```
EmilyBourne, line 6075: Would you mind adding the python printers? For most classes adding the python printer is a 2 line fix so it's worth doing straight away

mostafaMamoni: sure

pyccel/codegen/printing/pycode.py
```diff
@@ -418,6 +418,18 @@ def _print_NumpyAutoFill(self, expr):
                 dtype = dtype,
                 order = expr.order)
 
+    def _print_NumpyLinspace(self, expr):
+        return expr.__str__()
```
EmilyBourne, line 422: You shouldn't do this for 2 reasons:
1. `expr.__str__()` should be written as `str(expr)`
2. `str(expr)` does not always give the same result as `_print(expr)`. Although the contents for `_print_NumpyLinspace` may be the same, that does not mean that they are also the same for `_print(expr.start)` for example

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -6094,7 +6099,7 @@ def get_matmul(arr):
         from numpy import matmul, shape
         a = matmul(arr, arr)
         s = shape(a)
-        return len(s), s[0], s[1], a[0,1], a[1,0]
+        return len(s) + s[0] + s[1] + a[0,1] + a[1,0]
```
EmilyBourne: Can you explain this change please

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5953,3 +5981,858 @@ def get_prod(arr):
...
+    @types('complex64[:,:]')
+    @types('complex128[:,:]')
+    def get_linspace(arr):
+        from numpy import linspace, shape, ones
+        numberOfSamplesToGenerate = 7
+        start = ones((2,5))
+        stop = arr
+        a = linspace(start, stop, numberOfSamplesToGenerate)
+        s = shape(a)
+        return len(s), s[0], s[1], s[2], a[0, 0, 0], a[0, 1, 0], a[1, 0, 0], a[1, 1, 0]
```
EmilyBourne, line 6786: Do you think this output is sufficient to test that linspace does everything it should? Would it be possible to return the min and max to help double check? It would also be good to check that `a` is ordered correctly if possible

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5823,10 +5851,10 @@ def get_prod(arr):
...
 
-    fl = uniform(min_float / 2, max_float / 2, size = size)
-    fl32 = uniform(min_float32, max_float32, size = size)
+    fl = uniform(min_int, max_int, size = size)
+    fl32 = uniform(min_int, max_int, size = size)
     fl32 = np.float32(fl32)
-    fl64 = uniform(min_float64 / 2, max_float64 / 2, size = size)
+    fl64 = uniform(min_int64, max_int64, size=size)
 
     cmplx128_from_float32 = uniform(low=-((-min_float32) ** (1/5)), high=(max_float32 ** (1/5)), size = size) + uniform(low=-((-min_float32) ** (1/5)), high=(max_float32 ** (1/5)), size = size) * 1j
```
EmilyBourne, line 5859: Why do you only have the calculation of the largest possible value for complexes?

tests/epyccel/recognised_functions/test_numpy_funcs.py
mostafaMamoni: ah i missed this, it was for testing with np.isclose without .any() to .all().

tests/epyccel/recognised_functions/test_numpy_funcs.py
mostafaMamoni: it seems dump, but the results of floats and ints never pass the bounds.

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5953,3 +5981,859 @@ def get_prod(arr):
...
+    @types('int32')
+    @types('int64')
+    @types('float')
+    @types('float32')
+    @types('float64')
+    @types('complex64')
+    @types('complex128')
+    def get_where(a):
+        from numpy import where, shape
+        b = where(a)
```
EmilyBourne, line 6189: I think this is the only call to `where` that is currently acceptable. Is that correct?

But I wonder if it should be disallowed in pyccel. It is a rather complicated function which returns a tuple of arrays:
```
In [1]: import numpy as np

In [2]: x = np.ones((3,4,5))

In [3]: np.where(x)
Out[3]: 
(array([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1,
        1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 2, 2,
        2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2]),
 array([0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 3, 3, 3, 3, 3, 0, 0,
        0, 0, 0, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 3, 3, 3, 3, 3, 0, 0, 0, 0,
        0, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 3, 3, 3, 3, 3]),
 array([0, 1, 2, 3, 4, 0, 1, 2, 3, 4, 0, 1, 2, 3, 4, 0, 1, 2, 3, 4, 0, 1,
        2, 3, 4, 0, 1, 2, 3, 4, 0, 1, 2, 3, 4, 0, 1, 2, 3, 4, 0, 1, 2, 3,
        4, 0, 1, 2, 3, 4, 0, 1, 2, 3, 4, 0, 1, 2, 3, 4]))
```

I suspect that currently we don't handle it like this at all.

[Numpy's documentation](https://numpy.org/doc/stable/reference/generated/numpy.where.html) specifically advises against using it like this:
>When only condition is provided, this function is a shorthand for np.asarray(condition).nonzero(). Using nonzero directly should be preferred, as it behaves correctly for subclasses. The rest of this documentation covers only the case where all three arguments are provided.

@yguclu do you have an opinion about this?

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5953,3 +5981,859 @@ def get_prod(arr):
...
+        s = shape(b)
+        return len(s), s[0], a[0]
```
EmilyBourne, line 6191: Shouldn't this return elements of `b` not elements of `a` ?

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5953,3 +5981,859 @@ def get_prod(arr):
...
+    @types('float[:]')
+    @types('float32[:]')
+    @types('float64[:]')
+    @types('complex64[:]')
+    @types('complex128[:]')
+    def get_chosen_elements(arr):
+        from numpy import where, shape
+        a = where(arr)
+        s = shape(a)
+        return len(s), s[0], a[1], a[0]
```
EmilyBourne, line 6357: Assuming `where` is correctly implemented. you are returning an array here. E.g.:
```
In [6]: x = np.ones(4)

In [7]: a=np.where(x)

In [8]: a[0]
Out[8]: array([0, 1, 2, 3])
```

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5953,3 +5981,859 @@ def get_prod(arr):
...
+    @types('float[:,:]')
+    @types('float32[:,:]')
+    @types('float64[:,:]')
+    @types('complex64[:,:]')
+    @types('complex128[:,:]')
+    def get_chosen_elements(arr):
+        from numpy import where, shape
+        a = where(arr)
+        s = shape(a)
+        return len(s), s[0], a[0,0], a[0,1], a[1,0], a[1,1]
```
EmilyBourne, line 6528: This code is not valid. It raises the following error:
```
In [11]: get_chosen_elements(x)
<ipython-input-11-dfd3b423d811> in <module>
----> 1 get_chosen_elements(x)

<ipython-input-9-3dbdda16911b> in get_chosen_elements(arr)
      3         a = where(arr)
      4         s = shape(a)
----> 5         return len(s), s[0], a[0,0], a[0,1], a[1,0], a[1,1]
      6 

TypeError: tuple indices must be integers or slices, not tuple
```
when run in python without pyccel

EmilyBourne (changes requested):

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5953,3 +5981,859 @@ def get_prod(arr):
...
+        f_integer64 = epyccel(get_matmul, language=language)
+        assert (f_integer64(integer64) == get_matmul(integer64))
+
+    f_fl = epyccel(get_matmul, language=language)
+    f_fl32 = epyccel(get_matmul, language=language)
+    f_fl64 = epyccel(get_matmul, language=language)
+
+    assert np.isclose(f_fl(fl), get_matmul(fl), rtol=RTOL, atol=ATOL).any()
+    assert np.isclose(f_fl32(fl32), get_matmul(fl32), rtol=RTOL32, atol=ATOL32).any()
+    assert np.isclose(f_fl64(fl64), get_matmul(fl64), rtol=RTOL, atol=ATOL).any()
```
EmilyBourne, line 6150: Please use `all` not `any`

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5953,3 +5981,859 @@ def get_prod(arr):
...
+    cmplx128 = uniform(low=-((abs(min_int) / size[0] * 2)**(1/2)), high=(max_int / size[0] * 2)**(1/2), size=size) + uniform(low=-((abs(min_int) / size[0] * 2)**(1/2)), high=(max_int / size[0] * 2)**(1/2), size=size) * 1j
+
+    f_bl = epyccel(get_matmul, language=language)
+
+    assert (f_bl(bl) == get_matmul(bl))
+
+    f_integer8 = epyccel(get_matmul, language=language)
+    f_integer16 = epyccel(get_matmul, language=language)
+    f_integer = epyccel(get_matmul, language=language)
+    f_integer32 = epyccel(get_matmul, language=language)
```
EmilyBourne, line 6132: Interfaces only need translating once!

tests/epyccel/recognised_functions/test_numpy_funcs.py
EmilyBourne: Which bounds is that true for? The old bounds (e.g for f164: `min_float64 / 2`) or the new bounds ( e.g. for f164 `min_int64`)?
For the new bounds that seems likely as floats can contain much larger numbers than integers. You are therefore a long way from testing the full range of floats at the moment (this may not be a problem, you are still testing a large range of floats)

pyccel/codegen/printing/pycode.py
```diff
@@ -418,6 +418,24 @@ def _print_NumpyAutoFill(self, expr):
...
+        func_name = type_name[5:].lower()
+        return "{func_name}({start}, {stop}, {size})".format(
+                func_name = func_name,
+                start = self._print(expr.start),
+                stop = self._print(expr.stop),
+                size = self._print(expr.size))
+
+    def _print_NumpyMatmul(self, expr):
+        type_name = type(expr).__name__
+        func_name = type_name[5:].lower()
```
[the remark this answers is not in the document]
EmilyBourne: If the answer is no then this point is also valid for a few other functions

EmilyBourne:
Do you know why the tests are failing?

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5910,46 +5658,522 @@ def get_prod(arr):
...
+    @types('float[:,:]')
+    @types('float32[:,:]')
+    @types('float64[:,:]')
+    @types('complex64[:,:]')
+    @types('complex128[:,:]')
+    def get_matmul(arr):
+        from numpy import matmul, shape
+        a = matmul(arr, arr)
+        s = shape(a)
+        return len(s) , s[0] , s[1] , a[0,1] , a[1,0]
```
EmilyBourne, line 5788: `a[0,1]` and `a[1,0]` are int64 if `arr` is `int64[:]` which means they are affected by issue #735

pyccel/codegen/printing/pycode.py
```diff
@@ -418,6 +418,22 @@ def _print_NumpyAutoFill(self, expr):
                 dtype = dtype,
                 order = expr.order)
 
+    def _print_NumpyLinspace(self, expr):
+        type_name = type(expr).__name__
+        func_name = type_name[5:].lower()
```
EmilyBourne, line 423: Why not `'linspace'`

pyccel/codegen/printing/pycode.py
```diff
@@ -418,6 +418,22 @@ def _print_NumpyAutoFill(self, expr):
...
+        return "{func_name}({start}, {stop}, {size})".format(
+                func_name = func_name,
+                start = self._print(expr.start),
+                stop = self._print(expr.stop),
+                size = self._print(expr.size))
+
+    def _print_NumpyMatmul(self, expr):
+        return "{func_name}({x1}, {x2})".format(
+                func_name = 'matmul',
```
EmilyBourne, line 432: Does this need to be a parameter? `"matmul({x1}, {x2})".format(..)` is faster for python

EmilyBourne (approved):
Please update the PR description

tests/epyccel/recognised_functions/test_numpy_types.py
```diff
@@ -0,0 +1,2173 @@
...
+    assert epyccel_func(integer8) == get_complex128(integer8)
+    assert epyccel_func(integer16) == get_complex128(integer16)
+    assert epyccel_func(integer) == get_complex128(integer)
+    assert epyccel_func(integer32) == get_complex128(integer32)
+    # the if block should be removed after resolving (https://github.com/pyccel/pyccel/issues/735).
+    if sys.platform != 'win32':
+        assert epyccel_func(integer64) == get_complex128(integer64)
+    assert epyccel_func(fl) == get_complex128(fl)
+    assert epyccel_func(fl32) == get_complex128(fl32)
+    assert epyccel_func(fl64) == get_complex128(fl64)
```
EmilyBourne, line 2173: I like the idea of a separate file for test_numpy_types :+1:
Could we put the functions in a separate file? That way you can reduce the number of tests by grouping tests of similar types. E.g:
```
        pytest.param("fortran", marks = [pytest.mark.fortran]),
        pytest.param("c", marks = [
            pytest.mark.skip(reason="Arrays not handled yet."),
            pytest.mark.c]
        ),
        pytest.param("python", marks = [
            pytest.mark.python]
        )
    )
)
@pytest.mark.parametrize( 'get_complex', [mod.get_complex128, mod.get_complex64])
def test_numpy_complex128_array_like_2d(language, get_complex):
```

tests/epyccel/recognised_functions/test_numpy_types.py
```diff
@@ -0,0 +1,1234 @@
...
+        fl = uniform(min_int32, max_int32)
+        fl32 = uniform(min_int32, max_int32)
+        fl32 = np.float32(fl32)
+        fl64 = uniform(min_int32, max_int32)
+
+    if get_int == get_int64: # pylint: disable=comparison-with-callable
+        fl = uniform(min_int64, max_int64)
+        fl32 = uniform(min_int64, max_int64)
+        fl32 = np.float32(fl64)
+        fl64 = uniform(min_int64, max_int64)
```
EmilyBourne, line 213: Maybe the max int should also be a function argument?
```
@pytest.mark.parametrize( 'int_choice', [(get_int, max_int), (get_int64, max_int64),
                                                                 (get_int32, max_in32), (get_int16, max_int16),
                                                                 (get_int8, max_int8) ])
def test_numpy_int_scalar(language, int_choice):
     get_int = int_choice[0]
     max_int = int_choice[1]
```

tests/epyccel/recognised_functions/test_numpy_types.py
```diff
@@ -0,0 +1,1234 @@
...
+    assert epyccel_func(integer8) == get_int(integer8)
+    assert epyccel_func(integer16) == get_int(integer16)
+    assert epyccel_func(integer) == get_int(integer)
+    assert epyccel_func(integer32) == get_int(integer32)
+    # the if block should be removed after resolving (https://github.com/pyccel/pyccel/issues/735).
+    if sys.platform != 'win32':
+        assert epyccel_func(integer64) == get_int(integer64)
+        assert epyccel_func(fl) == get_int(fl)
+        assert epyccel_func(fl64) == get_int(fl64)
+        # Python returns always -32768 wich is less than min_int8, epyccel function returns 0.
```
EmilyBourne, line 487: Is this due to the float boundaries? Is this also the case if `fl32=2` for example?

tests/epyccel/recognised_functions/test_numpy_types.py
```diff
@@ -0,0 +1,1234 @@
...
+@types('int16')
+@types('int32')
+@types('int64')
+@types('float')
+@types('float32')
+@types('float64')
+def get_float64(a):
+    from numpy import float64
+    b = float64(a)
+    return b
```
EmilyBourne, line 517: I really want to use a template for these. I don't remember if that means that the whole module would need to be translated at once though or if we could define the template once and still translate the functions individually. Do you know?

tests/epyccel/recognised_functions/test_numpy_types.py
```diff
@@ -0,0 +1,1234 @@
...
+    integer8 = randint(min_int8, max_int8, size=size, dtype=np.int8)
+    integer16 = randint(min_int16, max_int16, size=size, dtype=np.int16)
+    integer = randint(min_int, max_int, size=size, dtype=np.int)
+    integer32 = randint(min_int32, max_int32, size=size, dtype=np.int32)
+    integer64 = randint(min_int64, max_int64, size=size, dtype=np.int64)
+
+    fl = uniform(min_float / 2, max_float / 2, size = size)
+    fl32 = uniform(min_float32 / 2, max_float32 / 2, size = size)
+    fl32 = np.float32(fl32)
+    fl64 = uniform(min_float64 / 2, max_float64 / 2, size = size)
```
EmilyBourne, line 369: Why do we need to divide by 2?

mostafaMamoni: i used to use ints boundaries for this type, i think i just missed it because it has no arrays tests, for your question, Python raises `overflow` error when using min/max_float:
```python
>>> import numpy as np
>>> fl = np.random.uniform(np.finfo('float').min, np.finfo('float').max, size = 5)
  File "<stdin>", line 1, in <module>
  File "mtrand.pyx", line 1106, in numpy.random.mtrand.RandomState.uniform
OverflowError: Range exceeds valid bounds
>>> fl32 = np.random.uniform(np.finfo('float32').min, np.finfo('float32').max, size = 5)
>>> fl = np.random.uniform(np.finfo('float').min / 2, np.finfo('float').max / 2, size = 5)
```

EmilyBourne: ok :+1:

tests/epyccel/recognised_functions/test_numpy_types.py
```diff
@@ -174,45 +174,26 @@ def get_int8(a):
     b = int8(a)
     return b
 
-@pytest.mark.parametrize( 'get_int', [get_int, get_int64, get_int32, get_int16, get_int8])
-def test_numpy_int_scalar(language, get_int):
+@pytest.mark.parametrize( 'function_boundaries', [(get_int, min_int, max_int), (get_int64, min_int64, max_int64), (get_int32, min_int32, max_int32),  (get_int16, min_int16, max_int16), (get_int8, min_int8, max_int8)])
```
EmilyBourne, line 177: Please break up this huge line

EmilyBourne:
To mark 1 parameter as xfail you should be able to do something like:
```
@pytest.mark.parametrize( 'function_boundaries', [
    (get_int, min_int, max_int), 
    pytest.param((get_int64, min_int64, max_int64), marks = [pytest.mark.xfail(reason="See issue #735 ")]),
    (get_int32, min_int32, max_int32), 
    (get_int16, min_int16, max_int16),
    (get_int8, min_int8, max_int8)])
```

mostafaMamoni:
> To mark 1 parameter as xfail you should be able to do something like:
> 
> ```
> @pytest.mark.parametrize( 'function_boundaries', [
>     (get_int, min_int, max_int), 
>     pytest.param((get_int64, min_int64, max_int64), marks = [pytest.mark.xfail(reason="See issue #735 ")]),
>     (get_int32, min_int32, max_int32), 
>     (get_int16, min_int16, max_int16),
>     (get_int8, min_int8, max_int8)])
> ```

thanks, but i think will just make it inside this if block:
```python
    # the if block should be removed after resolving (https://github.com/pyccel/pyccel/issues/735).
    if sys.platform != 'win32':
        f_integer64_output = epyccel_func(integer64)
        test_int64_output = get_int(integer64)

        assert f_integer64_output == test_int64_output
        assert matching_types(f_integer64_output, test_int64_output)

        f_fl_output = epyccel_func(fl)
        test_float_output = get_int(fl)

        assert f_fl_output == test_float_output
        assert matching_types(f_fl_output, test_float_output)

        f_fl64_output = epyccel_func(fl64)
        test_float64_output = get_int(fl64)

        assert matching_types(f_fl64_output, test_float64_output)

        f_fl32_output = epyccel_func(fl32)
        test_float32_output = get_int(fl32)

        assert f_fl32_output == test_float32_output
        assert matching_types(f_fl32_output, test_float32_output)
```

tests/epyccel/recognised_functions/test_numpy_types.py
mostafaMamoni: i dont think it's due to the float boundaries since it works fine for 1d array tests

tests/epyccel/recognised_functions/test_numpy_types.py
mostafaMamoni: i dont have much experience with templates to tell you.

EmilyBourne: ok. Let's assume we can't for now then :(

tests/epyccel/recognised_functions/test_numpy_types.py
EmilyBourne: It works for 1d but not for 2d? I don't see how this could be a boundary problem? Are you sure it works for 1d?
Can you try setting the boundaries to something that would also give an output in a valid range?

mostafaMamoni: it works with max/min_int16 and less than that.
Python gives `(2, 2, 5, -32768, -32768)` at overflowing, Fortran gives `(2, 2, 5, 0,  0)` or random values with max/min_int32.
i suggest to fix it by:
```python
    fl32 = uniform(min_float32 / 2, min_float32 / 2, size = size)
    if language == 'fortran' and get_int == get_int16_arr_2d: # pylint: disable=comparison-with-callable
        fl32 = uniform(min_int16, max_int16, size = size)
```

EmilyBourne: I'm not sure it makes sense to test `np.int16(a)` with `a` larger than max/min_int16. It's not super obvious what the result is supposed to be. That said we usually follow what python does so I don't know.
Is `min_int16 == 32768`?

Is it working for c?

It's strange that only `np.int16` gives this problem and not `np.int32` (on machines other than windows)

mostafaMamoni: we dont test it in c yet.

EmilyBourne: Is min_int16 == 32768?

EmilyBourne: > Fortran gives (2, 2, 5, 0, 0) or random values with max/min_int32.

int32 not int16?

EmilyBourne: My guess would be that fortran is casting while python is taking the closest int16. Does that look right?

mostafaMamoni: yes it does, but how casting can lead to 0 ?

EmilyBourne: Casting often involves reading the data as though it already had the correct type. E.g.:
An unsigned int8 uses 8 bits:
16 in binary gives: `0001 0000`
If we want to cast to an unsigned int4 then we will take 4 of those 8 bits (I'm not sure if it's the first 4 or the last 4) so you will have:
`uint4(16) : 0000 = 0`

EmilyBourne (approved):

yguclu (approved):
Good job @mostafaMamoni!

Can you please add a new file to the tutorial folder, where you list the Numpy functions that we can print (with the arguments that we support)? This can be done in a separate PR.

PR #721, opened 2021-02-01
Support shape overriding on full/empty/ones/zeros_like functions

### LINKED ISSUE #720: Pyccel doesn't support shape overriding in Numpy functions
According to Numpy docs, shape overriding (pass `shape` argument to the following functions that override the shape of the argument `array_like`) has been added to `full_like`, `empty_like`, `zeros_like`, and `ones_like` in Numpy 1.17, Pyccel still has the old implementation (Pyccel doesn't accept the `shape` keyword when you pass it to those functions).
here is the old implementation:
https://het.as.utexas.edu/HET/Software/Numpy/reference/generated/numpy.full_like.html
the new one:
https://numpy.org/doc/stable/reference/generated/numpy.full_like.html#numpy.full_like

mostafaMamoni:
Numpy v1.17 added shape overriding (i.e. the option to override the shape of the argument) to the functions `full_like`, `empty_like`, `zeros_like`, and `ones_like`. With this PR we add Pyccel the ability to accept the new `shape` argument when passed to these functions. Fixes #720 

Old signature, as of Numpy 1.16: https://numpy.org/doc/1.16/reference/generated/numpy.full_like.html
New signature, as of Numpy 1.17: https://numpy.org/doc/1.17/reference/generated/numpy.full_like.html

Commit summary:
- Add `shape` argument to the following classes in numpyext: `NumpyFullLike`, `NumpyZerosLike`, `NumpyOnesLike`, `NumpyEmptyLike`;
- Add tests for full/empty/ones/zeros_like functions.

### DIFF
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -629,7 +629,7 @@ class NumpyFullLike:
 class NumpyEmptyLike:
     """ Represents a call to numpy.empty_like for code generation.
     """
-    def __new__(cls, a, dtype=None, order='K', subok=True,shape=None):
+    def __new__(cls, a, dtype=None, order='K', subok=True, shape=None):
 
         # NOTE: we ignore 'subok' argument
         dtype = a.dtype if (dtype is None) or isinstance(dtype, Nil) else dtype
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -617,49 +617,52 @@ class NumpyOnes(NumpyAutoFill):
 class NumpyFullLike:
     """ Represents a call to numpy.full_like for code generation.
     """
-    def __new__(cls, a, fill_value, dtype=None, order='K', subok=True):
+    def __new__(cls, a, fill_value, dtype=None, order='K', subok=True, shape=None):
 
         # NOTE: we ignore 'subok' argument
         dtype = a.dtype if (dtype is None) or isinstance(dtype, Nil) else dtype
         order = a.order if str(order).strip('\'"') in ('K', 'A') else order
-
-        return NumpyFull(Shape(a), fill_value, dtype, order)
+        shape = Shape(a) if shape is None else shape
+        return NumpyFull(shape, fill_value, dtype, order)
 
 #=======================================================================================
 class NumpyEmptyLike:
     """ Represents a call to numpy.empty_like for code generation.
     """
-    def __new__(cls, a, dtype=None, order='K', subok=True):
+    def __new__(cls, a, dtype=None, order='K', subok=True,shape=None):
 
         # NOTE: we ignore 'subok' argument
         dtype = a.dtype if (dtype is None) or isinstance(dtype, Nil) else dtype
         order = a.order if str(order).strip('\'"') in ('K', 'A') else order
+        shape = Shape(a) if shape is None else shape
 
-        return NumpyEmpty(Shape(a), dtype, order)
+        return NumpyEmpty(shape, dtype, order)
 
 #=======================================================================================
 class NumpyOnesLike:
     """ Represents a call to numpy.ones_like for code generation.
     """
-    def __new__(cls, a, dtype=None, order='K', subok=True):
+    def __new__(cls, a, dtype=None, order='K', subok=True, shape=None):
 
         # NOTE: we ignore 'subok' argument
         dtype = a.dtype if (dtype is None) or isinstance(dtype, Nil) else dtype
         order = a.order if str(order).strip('\'"') in ('K', 'A') else order
+        shape = Shape(a) if shape is None else shape
 
-        return NumpyOnes(Shape(a), dtype, order)
+        return NumpyOnes(shape, dtype, order)
 
 #=======================================================================================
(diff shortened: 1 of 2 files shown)


tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -2492,3 +2492,900 @@ def max_call(x):
     f1 = epyccel(max_call, language = language)
     x = randint(99,size=10)
     assert(f1(x) == max_call(x))
+
+@pytest.mark.parametrize( 'language', (
+        pytest.param("fortran", marks = pytest.mark.fortran),
+        pytest.param("c", marks = [
+            pytest.mark.skip(reason="arrays not implemented"),
```
EmilyBourne, line 2499: This is no longer a valid reason. Could you be more specific

pyccel/ast/numpyext.py
```diff
@@ -617,49 +617,52 @@ def fill_value(self):
...
-        return NumpyFull(Shape(a), fill_value, dtype, order)
+        shape = Shape(a) if shape is None else shape
+        return NumpyFull(shape, fill_value, dtype, order)
 
 #=======================================================================================
 class NumpyEmptyLike:
     """ Represents a call to numpy.empty_like for code generation.
     """
-    def __new__(cls, a, dtype=None, order='K', subok=True):
+    def __new__(cls, a, dtype=None, order='K', subok=True,shape=None):
```
EmilyBourne, line 632: Please add a space after the comma

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -2492,3 +2492,900 @@ def max_call(x):
...
+    f_shape_C  = epyccel(create_full_like_shape_C, language = language)
+    assert(f_shape_C(size) == create_full_like_shape_C(size))
+
+    f_shape_F  = epyccel(create_full_like_shape_F, language = language)
+    assert(f_shape_F(size) == create_full_like_shape_F(size))
+
+@pytest.mark.parametrize( 'language', (
+        pytest.param("fortran", marks = pytest.mark.fortran),
+        pytest.param("c", marks = [
+            pytest.mark.skip(reason="casting to complex in not handled correctly"),
```
EmilyBourne, line 2676: Is there an issue for this?

EmilyBourne (changes requested):
As the `X_like` functions are designed to create an array which is similar to another array, I think it is important to test them with array input arguments, not just lists.

Could you remove the newly tested functions from the list of untested objects at the satrt of this file please

tests/epyccel/recognised_functions/test_numpy_funcs.py
mostafaMamoni: i didn't find any issue related to this,

EmilyBourne: Then please can you create one

tests/epyccel/recognised_functions/test_numpy_funcs.py
mostafaMamoni: i have discussed with Moad about that, and we found two options, since we haven’t implemented tuples in C, the `shape` function won't work, so we have either make the reason more specific i suggest "tuples not implemented yet" or do not work we `shape` function and do not test the length of the array, because we calculate it through the return of the `shape` function

EmilyBourne: `tuples not implemented yet` seems fine to me

tests/epyccel/recognised_functions/test_numpy_funcs.py
mostafaMamoni: sure.

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -1284,7 +1284,7 @@ def create_full_arg_names(val):
 @pytest.mark.parametrize( 'language', (
         pytest.param("fortran", marks = pytest.mark.fortran),
         pytest.param("c", marks = [
-            pytest.mark.skip(reason="tuples not implemented"),
+            pytest.mark.skip(reason="Tuples not implemented"),
```
saniQQ, line 1287: the skip reason is different from the others here.

saniQQ (approved):

tests/epyccel/recognised_functions/test_numpy_funcs.py
EmilyBourne: Could reference the issue in the reason please. E.g:
```
pytest.mark.skip(reason="Casting to complex is not handled correctly. See issue #723 ")
```

EmilyBourne (changes requested):
As in my previous review:
> Could you remove the newly tested functions from the list of untested objects at the start of this file please

[here](https://github.com/pyccel/pyccel/blob/master/tests/epyccel/recognised_functions/test_numpy_funcs.py#L13)

EmilyBourne (approved):

yguclu (approved):

PR #724, opened 2021-02-02
Handle type casting in array_fill

mostafaMamoni:
`array_fill` is a Pyccel builtin function, written in C, which is used to fill an `array`(second argument) with a given `value` (first argument) and return a `value-filled array`. The type of `value` should be the same type of `array` to prevent wrong alignments and even crashes (in case of the data type of `value` greater than the data type of `array`). By fixing we can reactivate the x-failing C tests for `NumpyFull` and `NumpyFullLike` (soon), and fix all kinds of casting in `array_fill`.

Commit summary:
- Handle the casting (including complex casting) in `NumpyFull` class by using more flexible dictionaries and function to handle them;
- Cast dynamically in `_print_PythonComplex`, `_print_PythonFloat`, `_print_PythonInt`;
- Activate the test `test_full_dtype` in test_numpy_functions.py`.

### DIFF
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -89,6 +89,74 @@ __all__ = (
     'NumpyArange'
 )
 
+#=======================================================================================
+class NumpyComplex(PythonComplex):
+    """ Represents a call to numpy.complex() function.
+    """
+
+class NumpyComplex64(NumpyComplex):
+    _precision = dtype_registry['complex64'][1]
+
+class NumpyComplex128(NumpyComplex):
+    _precision = dtype_registry['complex128'][1]
+
+#=======================================================================================
+class NumpyFloat(PythonFloat):
+    """ Represents a call to numpy.float() function.
+    """
+
+class NumpyFloat32(NumpyFloat):
+    """ Represents a call to numpy.float32() function.
+    """
+    _precision = dtype_registry['float32'][1]
+
+class NumpyFloat64(NumpyFloat):
+    """ Represents a call to numpy.float64() function.
+    """
+    _precision = dtype_registry['float64'][1]
+
+#=======================================================================================
+# TODO [YG, 13.03.2020]: handle case where base != 10
+class NumpyInt(PythonInt):
+    """ Represents a call to numpy.int() function.
+    """
+    def __new__(cls, arg=None, base=10):
+        return PythonInt.__new__(cls, arg)
+
+class NumpyInt32(NumpyInt):
+    """ Represents a call to numpy.int32() function.
+    """
+    _precision = dtype_registry['int32'][1]
+
+class NumpyInt64(NumpyInt):
+    """ Represents a call to numpy.int64() function.
+    """
+    _precision = dtype_registry['int64'][1]
+
+#==============================================================================
+class NumpyReal(PythonReal):
+    """Represents a call to  numpy.real for code generation.
+
+    > a = 1+2j
+    > np.real(a)
+    1.0
+    """
+
+PrecisionToDtype_int = {
+    4 : NumpyInt32,
+    8 : PythonInt,
+}
+
+PrecisionToDtype_float = {
+    4 : NumpyFloat32,
+    8 : PythonFloat,
+}
+
+PrecisionToDtype_complex = {
+    4 : NumpyComplex64,
+    8 : PythonComplex,
+    16 : NumpyComplex128,
+}
 #==============================================================================
 numpy_constants = {
     'pi': Constant('real', 'pi', value=numpy.pi),
@@ -346,15 +414,6 @@ def Shape(arg):
         return PythonTuple(*arg.shape)
 
 #==============================================================================
-class NumpyReal(PythonReal):
-    """Represents a call to  numpy.real for code generation.
-
-    > a = 1+2j
-    > np.real(a)
-    1.0
-    """
-
-#==============================================================================
 class NumpyImag(PythonImag):
     """Represents a call to  numpy.real for code generation.
 
@@ -495,6 +554,16 @@ class NumpyRandint(PyccelInternalFunction):
         return self._low
 
 #==============================================================================
+def process_stype(stype, precision):
+    if stype == 'integer':
+        return PrecisionToDtype_int[precision]
+    if stype == 'real':
+        return PrecisionToDtype_float[precision]
+    if stype == 'complex':
+        return PrecisionToDtype_complex[precision]
+    if stype == 'bool':
+        return python_builtin_datatypes['bool']
+
 class NumpyFull(PyccelInternalFunction, NumpyNewArray):
     """
     Represents a call to numpy.full for code generation.
@@ -534,13 +603,12 @@ class NumpyFull(PyccelInternalFunction, NumpyNewArray):
         order = NumpyNewArray._process_order(order)
 
         # Cast fill_value to correct type
-        # TODO [YG, 09.11.2020]: treat difficult case of LiteralComplex
         from pyccel.ast.datatypes import str_dtype
         stype = str_dtype(dtype)
-        if fill_value and stype != 'complex':
-            from pyccel.codegen.printing.fcode import python_builtin_datatypes
-            cast_func  = python_builtin_datatypes[stype]
-            fill_value = cast_func(fill_value)
+        from pyccel.codegen.printing.fcode import python_builtin_datatypes
+        cast_func  = python_builtin_datatypes[stype]
+        cast_func = process_stype(stype, precision)
+        fill_value = cast_func(fill_value)
         self._shape = shape
         self._rank  = len(self._shape)
         self._dtype = dtype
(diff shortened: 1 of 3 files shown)


EmilyBourne:
This PR is marked draft and "needs_initial_review". It cannot be both as a draft is something which is not "ready for review"

saniQQ (approved):

saniQQ (approved):

pyccel/ast/numpyext.py
```diff
@@ -534,12 +614,12 @@ def __init__(self, shape, fill_value, dtype=None, order='C'):
...
 
         # Cast fill_value to correct type
-        # TODO [YG, 09.11.2020]: treat difficult case of LiteralComplex
         from pyccel.ast.datatypes import str_dtype
         stype = str_dtype(dtype)
-        if fill_value and stype != 'complex':
+        if fill_value:
             from pyccel.codegen.printing.fcode import python_builtin_datatypes
             cast_func  = python_builtin_datatypes[stype]
+            cast_func = process_stype(stype, precision)
```
EmilyBourne, line 622: Why do we need to use `stype`? Can't we use the `dtype` directly?

pyccel/ast/numpyext.py
```diff
@@ -142,21 +142,41 @@ class NumpyReal(PythonReal):
...
-}
+def process_stype(stype, precision):
+    """
+    returns a cast function from
+    PrecisionToDtype_int, PrecisionToDtype_float, PrecisionToDtype_complex, python_builtin_datatypes
+    by ginving it:
+    stype:
+        supported data type name.
+    precision:
+        describs the precision of stype, to make the returned function more specific.
```
yguclu, line 167: This docstrings needs better formatting, as usual:
```python
"""
[Function description]

Parameters
----------
stype : str
     [parameter description]

precision : int
     [parameter description]

Results
-------
<base class of all return objects> or None
    [result description]
"""
```
See https://numpydoc.readthedocs.io/en/latest/format.html#sections.

Further, there are some typos to be fixed:
- ginving -> giving
- describs -> describes

pyccel/codegen/printing/ccode.py
```diff
@@ -461,10 +461,14 @@ def _print_LiteralComplex(self, expr):
...
-            return self._print(expr.internal_var)
+            value = self._print(expr.internal_var)
+            type_name = self.find_in_dtype_registry('complex', expr.precision)
+            return '({0})({1})'.format(type_name, value)
         else:
-            return self._print(PyccelAssociativeParenthesis(PyccelAdd(expr.real,
+            value = self._print(PyccelAssociativeParenthesis(PyccelAdd(expr.real,
                             PyccelMul(expr.imag, LiteralImaginaryUnit()))))
+            type_name = self.find_in_dtype_registry('complex', expr.precision)
+            return '({0})({1})'.format(type_name, value)
```
EmilyBourne, line 471: These lines seem to be repeated could you not use the if block to determine `value` then have these 2 lines in common?

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -1360,7 +1360,6 @@ def create_full_shape_F(n,m):
 @pytest.mark.parametrize( 'language', (
         pytest.param("fortran", marks = pytest.mark.fortran),
         pytest.param("c", marks = [
-            pytest.mark.skip(reason="casting to complex in not handled correctly, see https://github.com/pyccel/pyccel/issues/723"),
```
EmilyBourne, line 1363: Can this not also be removed for `full_like`?

EmilyBourne (changes requested):

yguclu (changes requested):

pyccel/ast/numpyext.py
mostafaMamoni: i think `stype` is enough because we need just a dictionary that is more flexible than `python_builtin_datatypes`.

EmilyBourne: Can the Native types not be dictionary keys?

pyccel/ast/numpyext.py
```diff
@@ -89,6 +89,94 @@
...
+    'complex' : {
+        4 : NumpyComplex64,
+        8 : PythonComplex,
+        16 : NumpyComplex128,}
+}
+
+def process_stype(stype, precision):
+    """
+    returns a cast function from
+    PrecisionToDtype_int, PrecisionToDtype_float, PrecisionToDtype_complex, python_builtin_datatypes
```
rakati, line 162: what is `PrecisionToDtype_int, PrecisionToDtype_float, PrecisionToDtype_complex, python_builtin_datatypes`?
I think you don't need to mention all that, just the `PrecisionToDtype` dictionary is enough.
example: find the suitable cast function for a data type and its precision using the `PrecisionToDtype` dictionary.

pyccel/ast/numpyext.py
mostafaMamoni: we can do that, by `_name` attribute in type objects.

pyccel/ast/numpyext.py
```diff
@@ -89,6 +89,94 @@
...
+    'int' : {
+        4 : NumpyInt32,
+        8 : PythonInt},
+    'float' : {
+        4 : NumpyFloat32,
+        8 : PythonFloat},
+    'complex' : {
+        4 : NumpyComplex64,
+        8 : PythonComplex,
+        16 : NumpyComplex128,}
```
rakati, line 156: do we have a numpy.bool_ data type to use it here or not?

pyccel/ast/numpyext.py
```diff
@@ -89,6 +89,94 @@
...
+    if stype == 'integer':
+        return PrecisionToDtype['int'][precision]
+    if stype == 'real':
+        return PrecisionToDtype['float'][precision]
+    if stype == 'complex':
+        return PrecisionToDtype['complex'][precision]
+    if stype == 'bool':
+        from pyccel.codegen.printing.fcode import python_builtin_datatypes
+        return python_builtin_datatypes['bool']
+    return None
```
rakati, line 178: this is the last condition, I think either we should raise an error `Unsupported type` or returning the bool cast function without using the `if` statement.

pyccel/ast/numpyext.py
```diff
@@ -89,6 +89,94 @@
...
+    > a = 1+2j
+    > np.real(a)
+    1.0
+    """
+
+#==============================================================================
+PrecisionToDtype = {
+    'int' : {
+        4 : NumpyInt32,
+        8 : PythonInt},
```
rakati, line 149: can't you use `numpyInt64` instead of `pythonInt`?

pyccel/ast/numpyext.py
```diff
@@ -89,6 +89,94 @@
...
+    'float' : {
+        4 : NumpyFloat32,
+        8 : PythonFloat},
```
rakati, line 152: same here

pyccel/ast/numpyext.py
```diff
@@ -89,6 +89,94 @@
...
+    'complex' : {
+        4 : NumpyComplex64,
+        8 : PythonComplex,
```
rakati, line 155: same above

pyccel/ast/numpyext.py
EmilyBourne: I agree, `PythonInt` has a different precision on different machines (8 on linux/mac, 4 on windows) so it is important that you use `NumpyInt64`

pyccel/ast/numpyext.py
EmilyBourne: What I meant was:
```
PrecisionToDtype = {
    NativeInteger() : {
        4 : NumpyInt32,
...
```
But using the `_name` is good too :+1:

pyccel/ast/numpyext.py
```diff
@@ -89,6 +89,75 @@
...
+class NumpyReal(PythonReal):
+    """Represents a call to  numpy.real for code generation.
+
+    > a = 1+2j
+    > np.real(a)
+    1.0
+    """
+
+#==============================================================================
+PrecisionToDtype = {
```
EmilyBourne, line 146: I don't think this name describes what this dictionary does. `PrecisionToDtype` implies that the key is a `precision` and the result is the `dtype`, but here the result is a casting function

mostafaMamoni: i agree.

EmilyBourne (approved):

pyccel/ast/numpyext.py
```diff
@@ -532,12 +593,8 @@ def __init__(self, shape, fill_value, dtype=None, order='C'):
         order = NumpyNewArray._process_order(order)
 
         # Cast fill_value to correct type
-        # TODO [YG, 09.11.2020]: treat difficult case of LiteralComplex
-        from pyccel.ast.datatypes import str_dtype
-        stype = str_dtype(dtype)
-        if fill_value and stype != 'complex':
-            from pyccel.codegen.printing.fcode import python_builtin_datatypes
-            cast_func  = python_builtin_datatypes[stype]
+        if fill_value and not isinstance(fill_value, Nil):
```
EmilyBourne, line 596: We seem to have a lot of places in this file where we check `if X is not None and not isinstance(X, Nil)`, it may be beyond the scope of this PR but I think it would be nice to add a magic method `__bool__` to `Nil` which returns `False` then all these tests can be replaced with `if X`. What do you think?

mostafaMamoni: I can do that in another PR.

EmilyBourne: ok :+1:

yguclu: Good idea!

rakati (approved):

yguclu (approved):

PR #725, opened 2021-02-02
Add IfSection class

EmilyBourne:
An `Allocate` node has a status which is one of `unknown`, `allocated` or `unallocated`. Currently the status is set to `unknown` unless the `Allocate` is called at the same time as the variable is created. In order to improve this we need to be able to handle situations such as the following:
1.
```
def f(b : bool):
    if b:
        a = np.ones(4)
        ...
    a = np.zeros(7)
```
2.
```
def f(b : bool):
    if b:
        a = np.ones(4)
    else:
        a = np.zeros(7)
```
Situation 1. can be handled easily after issue #694 is fixed by checking if `a` has any parents which are `Allocatable`s with a parent that is an `If`.

The second situation is more complicated however. In this case the `If` is only created after both allocations have been created. So `a` does not have any parents which are `Allocatable`s with a parent that is an `If`.
In order to handle this it is therefore necessary to add an intermediary class `IfSection`. This way each section of the `If` block is handled separately and as a result the class `IfSection` appears in the parents of the `Allocatable` and can be used to correctly determine the status of the second `Allocatable`

### DIFF
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -3479,10 +3479,8 @@ class If(Basic):
 
     Parameters
     ----------
-    args :
-        every argument is a tuple and
-        is defined as (cond, expr) where expr is a valid ast element
-        and cond is a boolean test.
+    args : IfSection
+           All arguments are sections of the complete If block
 
     Examples
     --------
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -3424,6 +3424,54 @@ class ErrorExit(Exit):
 
     """Exit with error."""
 
+class IfSection(Basic):
+    """Represents a condition and associated code block
+    in an if statement in the code.
+
+    Parameters
+    ----------
+    cond : PyccelAstNode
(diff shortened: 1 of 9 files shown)


EmilyBourne:
This PR is a precursor to PR #718 which I have tried to split in order to help the review process

pyccel/ast/core.py
```diff
@@ -3441,28 +3489,20 @@ class If(Basic):
     >>> from sympy import Symbol
     >>> from pyccel.ast.core import Assign, If
     >>> n = Symbol('n')
-    >>> If(((n>1), [Assign(n,n-1)]), (True, [Assign(n,n+1)]))
```
MenoIy, line 3444: class If docstring needs an update

MenoIy (approved):

yguclu (approved):
Well done @EmilyBourne! May I ask you to fix the formatting in one plac

saidctb (approved):

PR #726, opened 2021-02-02
Use mutable objects in classes instead of indexing _args

### LINKED ISSUE #660: Use mutable objects in classes instead of indexing _args

EmilyBourne:
Fixes #660.
This PR makes the classes in the following files mutable:

- builtins.py
- datatypes.py
- literals.py
- macros.py
- numpyext.py
- variable.py

In addition the following improvements are made to the above files:
- Use `super.__init__()` where possible. This means that the call does not necessarily need updating if the inheritance changes
- Simplify PythonList by using PythonTuple inheritance
-  Convert `int` indexes to `LiteralInteger`s in the constructor of `IndexedElement`
- Use PyccelNot in place of SympyNot.
- Remove unnecessary __new__ functions
- Correct a bug in `PythonBool` where `PythonBool.__init__` was not called
- Remove unnecessary inheritance
- Correct typos
- Remove unnecessary sympy objects

### DIFF
--- a/pyccel/ast/builtins.py
+++ b/pyccel/ast/builtins.py
@@ -476,19 +476,20 @@ class PythonRange(Basic):
 
     def __init__(self, *args):
         # Define default values
+        n = len(args)
 
-        if len(args) == 1:
+        if n == 1:
             self._start = LiteralInteger(0)
-            self._stop = args[0]
+            self._stop  = args[0]
             self._step  = LiteralInteger(1)
-        elif len(args) == 2:
+        elif n == 2:
             self._start = args[0]
-            self._stop = args[1]
+            self._stop  = args[1]
             self._step  = LiteralInteger(1)
-        elif len(args) == 3:
+        elif n == 3:
             self._start = args[0]
-            self._stop = args[1]
-            self._step = args[2]
+            self._stop  = args[1]
+            self._step  = args[2]
         else:
             raise ValueError('Range has at most 3 arguments')
 
--- a/pyccel/ast/builtins.py
+++ b/pyccel/ast/builtins.py
@@ -475,25 +475,22 @@ class PythonRange(Basic):
     """
 
     def __init__(self, *args):
-        start = LiteralInteger(0)
-        stop = None
-        step = LiteralInteger(1)
+        # Define default values
+        self._start = LiteralInteger(0)
+        self._step  = LiteralInteger(1)
 
         if len(args) == 1:
-            stop = args[0]
+            self._stop = args[0]
         elif len(args) == 2:
-            start = args[0]
-            stop = args[1]
+            self._start = args[0]
+            self._stop = args[1]
         elif len(args) == 3:
-            start = args[0]
-            stop = args[1]
-            step = args[2]
+            self._start = args[0]
+            self._stop = args[1]
+            self._step = args[2]
         else:
             raise ValueError('Range has at most 3 arguments')
 
-        self._start = start
-        self._stop  = stop
-        self._step  = step
         super().__init__()
 
     @property
--- a/pyccel/ast/builtins.py
+++ b/pyccel/ast/builtins.py
@@ -12,8 +12,6 @@ In this module we implement some of them in alphabetical order.
 """
 
 from sympy import Symbol
-from sympy import Expr, Not
-from sympy import sympify
 
 from .basic     import Basic, PyccelAstNode
 from .datatypes import (NativeInteger, NativeBool, NativeReal,
@@ -23,7 +21,7 @@ from .internals import PyccelInternalFunction
 from .literals  import LiteralInteger, LiteralFloat, LiteralComplex, Nil
 from .literals  import Literal, LiteralImaginaryUnit, get_default_literal_value
 from .operators import PyccelAdd, PyccelAnd, PyccelMul, PyccelIsNot
-from .operators import PyccelMinus, PyccelUnarySub
+from .operators import PyccelMinus, PyccelUnarySub, PyccelNot
 
 __all__ = (
     'PythonReal',
@@ -72,7 +70,7 @@ class PythonComplexProperty(PyccelInternalFunction):
 
     def __init__(self, arg):
         self._precision = arg.precision
-        PyccelInternalFunction.__init__(self, arg)
+        super().__init__(arg)
 
     @property
     def internal_var(self):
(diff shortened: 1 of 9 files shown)


MenoIy (approved):

pyccel/ast/builtins.py
```diff
@@ -408,55 +422,18 @@ class PythonList(PythonTuple):
     _is_homogeneous = True
     """ Represent lists in the code with dynamic memory management."""
     def __init__(self, *args, **kwargs):
```
yguclu, line 424: Do we still need to accept `**kwargs` here?

pyccel/ast/builtins.py
```diff
@@ -501,7 +477,7 @@ class PythonRange(Basic):
     Range(0, n, 1)
     """
 
-    def __new__(cls, *args):
+    def __init__(self, *args):
```
yguclu, line 480: This may be beyond the scope of this PR, but nonetheless... Passed in this way to the constructor, `args` is always a tuple. So why do we check its type below? I think we have a bug.

pyccel/ast/literals.py
```diff
@@ -164,25 +157,22 @@ def python_value(self):
 class LiteralImaginaryUnit(LiteralComplex):
     """Represents the python value j"""
     def __new__(cls):
-        return LiteralComplex.__new__(cls, 0, 1)
```
yguclu, line 167: Do we still need to override the `__new__` constructor, given that we have an `__init__` function?

EmilyBourne: Yes. `LiteralComplex` has an explicit `__new__` function which converts `LiteralComplex(0,1)` into a `LiteralImaginaryUnit` if it is not one already.
Therefore this `__new__` needs overriding

pyccel/ast/builtins.py
EmilyBourne: You are correct, I will fix it. I don't think it's a bug, just unnecessary code

pyccel/ast/builtins.py
EmilyBourne: No I don't think so. The `kwargs` object is not used in the old code either

pyccel/ast/literals.py
yguclu: So the problem is that `LiteralComplex.__new__` requires the arguments `real` and `imag`?

EmilyBourne: Exactly. If I do:
```
a = LiteralImaginaryUnit()
```
Then it will inherit `__new__` from `LiteralComplex` and complain about missing arguments. The fact that there is an `__init__` has no bearing on the `__new__`

pyccel/ast/builtins.py
```diff
@@ -482,22 +482,17 @@ def __init__(self, *args):
         stop = None
```
yguclu, line 482: Since we are at it, I think this line is never used, because `stop` is always redefined.

I think it would be cleaner if we moved the other two variable definitions (for `start` and `step`) inside the `if/elif/else` block.

EmilyBourne: I'm not sure I agree. These are default values. It is better to reduce code duplication where possible. I can remove `stop` if you want but in the context of default values it makes sense to provide all 3

pyccel/ast/builtins.py
```diff
@@ -407,56 +421,19 @@ class PythonList(PythonTuple):
...
-            self._rank = max(a.rank for a in args) + 1
-            if all(sh is not None for sh in shapes):
-                self._shape = (LiteralInteger(len(args)), ) + shapes[0]
-                self._rank  = len(self._shape)
-
-    @property
-    def is_homogeneous(self):
-        return self._is_homogeneous
+    def __init__(self, *args):
+        super().__init__(*args)
```
yguclu, line 425: I suppose that Codacy is right here. We do not need an `__init__` for this class.

pyccel/ast/builtins.py
yguclu: Since this code is not compiled but interpreted, my suggestion is faster because it creates the minimum number of Python objects in memory. But it is true that one (!) additional line was duplicated:
```python
n = len(args):

if n == 1:
    start = LiteralConstant(0)
    stop  = args[0]
    step  = LiteralConstant(1)
elif n == 2:
    start = args[0]
    stop  = args[1]
    step  = LiteralConstant(1)
elif n == 3:
    start = args[0]
    stop  = args[1]
    step  = args[2]
else:
    raise ...
```

EmilyBourne: Ok. I have changed the code to match your suggestion

yguclu: Thanks!

yguclu (approved):

PR #728, opened 2021-02-04
Add magic method __bool__ to Nil class

mostafaMamoni:
We seem to have a lot of places in the code where we check `if X is not None and not isinstance(X, Nil)`. All these tests can be shortened to `if X` after adding a magic method `__bool__` to the `Nil` class, which returns False.

Commit summary:
- Create `__bool__` method in Nil class;
- Replace `isinstance(X, Nil)` with `not X`;
- Replace `not isinstance(X, Nil)` with `X`;
- For conciseness, use `or` operator instead of ternary operator where possible.

### DIFF
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -3556,7 +3556,7 @@ def get_initial_value(expr, var):
     def is_None(expr):
         """Returns True if expr is None or Nil()."""
 
-        return not expr or expr is None
+        return isinstance(expr, Nil) or expr is None
 
     # ...
 
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -3556,7 +3556,7 @@ def get_initial_value(expr, var):
     def is_None(expr):
         """Returns True if expr is None or Nil()."""
 
-        return isinstance(expr, Nil) or expr is None
+        return not expr or expr is None
 
     # ...
 
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -675,7 +675,7 @@ class NumpyFullLike:
     def __new__(cls, a, fill_value, dtype=None, order='K', subok=True, shape=None):
 
         # NOTE: we ignore 'subok' argument
-        dtype = a.dtype if not dtype else dtype
+        dtype = dtype or a.dtype
         order = a.order if str(order).strip('\'"') in ('K', 'A') else order
         shape = Shape(a) if shape is None else shape
         return NumpyFull(shape, fill_value, dtype, order)
@@ -687,7 +687,7 @@ class NumpyEmptyLike:
     def __new__(cls, a, dtype=None, order='K', subok=True, shape=None):
 
         # NOTE: we ignore 'subok' argument
-        dtype = a.dtype if not dtype else dtype
+        dtype = dtype or a.dtype
         order = a.order if str(order).strip('\'"') in ('K', 'A') else order
         shape = Shape(a) if shape is None else shape
 
@@ -700,7 +700,7 @@ class NumpyOnesLike:
     def __new__(cls, a, dtype=None, order='K', subok=True, shape=None):
 
         # NOTE: we ignore 'subok' argument
-        dtype = a.dtype if not dtype else dtype
+        dtype = dtype or a.dtype
         order = a.order if str(order).strip('\'"') in ('K', 'A') else order
         shape = Shape(a) if shape is None else shape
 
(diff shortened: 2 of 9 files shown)


pyccel/ast/numpyext.py
```diff
@@ -115,7 +115,7 @@ class NumpyNewArray(PyccelAstNode):
     @staticmethod
     def _process_order(order):
 
-        if (order is None) or isinstance(order, Nil):
+        if (order is None) or not order:
```
EmilyBourne, line 118: This can be simplified to `if not order`

pyccel/parser/semantic.py
```diff
@@ -2497,7 +2497,7 @@ def _visit_FunctionDef(self, expr, **settings):
                         if isinstance(a, ValuedArgument):
 
                             # optional argument only if the value is None
-                            if isinstance(a.value, Nil):
+                            if not a.value:
```
EmilyBourne, line 2500: This needs to be the right type. (e.g. The argument is not optional if the value is `False`)

EmilyBourne: This is true in several of the places that you have changed.
I think to work out what you can change you can follow the following guidelines:
- If `Nil` is tested but not `None` then the type is important, don't change the test
- If the test is `if X or if isinstance(X, Nil)` then you can definitely change the test
- If the test is `if (X is None) or if isinstance(X, Nil)` then the type may be important. You need to think about whether `X` can have a value other than `None` or `Nil` which could be evaluated as `False`

EmilyBourne: It may be that you can't change many of them, but I still think it is useful for `Nil` to have a `__bool__` method anyway

pyccel/ast/core.py
```diff
@@ -3556,7 +3556,7 @@ def get_initial_value(expr, var):
     def is_None(expr):
         """Returns True if expr is None or Nil()."""
 
-        return isinstance(expr, Nil) or expr is None
+        return not expr or expr is None
```
EmilyBourne, line 3559: I think this one needs changing back

EmilyBourne (changes requested):

EmilyBourne (approved):

yguclu:
I cannot understand the PR description. Care to improve?

pyccel/ast/numpyext.py
```diff
@@ -675,7 +675,7 @@ class NumpyFullLike:
     def __new__(cls, a, fill_value, dtype=None, order='K', subok=True, shape=None):
 
         # NOTE: we ignore 'subok' argument
-        dtype = a.dtype if (dtype is None) or isinstance(dtype, Nil) else dtype
+        dtype = a.dtype if not dtype else dtype
```
yguclu, line 678: This kind of statements can be written in a completely equivalent for,
```
dtype = dtype or a.dtype
```
because the Python `or` operator returns the first argument that is not False.

See [here](https://stackoverflow.com/questions/4978738/is-there-a-python-equivalent-of-the-c-sharp-null-coalescing-operator) for example.

pyccel/ast/numpyext.py
```diff
@@ -687,7 +687,7 @@ class NumpyEmptyLike:
     def __new__(cls, a, dtype=None, order='K', subok=True, shape=None):
 
         # NOTE: we ignore 'subok' argument
-        dtype = a.dtype if (dtype is None) or isinstance(dtype, Nil) else dtype
+        dtype = a.dtype if not dtype else dtype
```
yguclu, line 690: Same comment

pyccel/ast/numpyext.py
```diff
@@ -700,7 +700,7 @@ class NumpyOnesLike:
...
```
yguclu, line 703: Same comment

pyccel/ast/numpyext.py
```diff
@@ -713,7 +713,7 @@ class NumpyZerosLike:
...
```
yguclu, line 716: Same comment

yguclu (changes requested):
Since this PR is about readability and conciseness, I have a suggestion for improvement.

pyccel/ast/numpyext.py
EmilyBourne: I didn't realise we could do this! Does this mean we need to open an issue about pyccel's `or` operator? Our translations always give the result of `or` as a boolean

yguclu: I'm not sure that we can mimic Python's behavior with Pyccel... but we can try :wink:

yguclu (approved):

pyccel/ast/numpyext.py
EmilyBourne: I think it should be possible, we just need to translate:
```
a or b
```
to:
```
a if bool(a) else b
```
if `a` and `b` are not booleans. This can easily be done in the `PyccelOr.__new__`

saidctb: why did you change that , imagine if the user writes, somthing like `a = np.array([1,2,3],dtype=None)`

EmilyBourne: @saidctb The behaviour is the same as this PR ensures that `bool(Nil())` returns `False`

saidctb: ah okay

PR #730, opened 2021-02-06
Fix passing array_like argument to numpy real function

mostafaMamoni:
Numpy's real function was failing when passing an `array_like` argument containing a complex type. This PR fixes that.

**Commit summary**

- Set `rank`, `shape`, and `precision` of `NumpyReal` object by creating a constructor in its class;
- Use `internal_var` property instead of `arg` in the `NumpyReal` printer in fcode.py;
- Set shape and rank in all casting functions in module `pyccel.ast.builtins`, to make sure all of their objects have the correct information;
- Add unit tests.


tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3465,3 +3465,371 @@ def create_zeros_like_3_val():
     assert(     f3_shape() ==      create_zeros_like_3_shape()      )
     assert(isclose(     f3_val()  ,      create_zeros_like_3_val()        , rtol=RTOL, atol=ATOL))
     assert(type(f3_val())  == type(create_zeros_like_3_val().item()))
+
+
+@pytest.mark.parametrize( 'language', (
```
saniQQ, line 3470: is there no way to run these test they are skiped in fortran and C

mostafaMamoni: i think we can comment int8 data types tests specifically.

saniQQ (approved):

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -1,4 +1,4 @@
-# pylint: disable=missing-function-docstring, missing-module-docstring, unidiomatic-typecheck/
+# pylint: disable=missing-function-docstring, redefined-builtin, unused-variable, missing-module-docstring, unidiomatic-typecheck/
```
EmilyBourne, line 1: We should not disable codacy errors apart from the docstring errors without good reason. Can you explain why this is necessary?
Usually `redefined-builtin` errors can be fixed by using a new name or for imports using `import X as Y`

`unused-variable` errors can be fixed by renaming the variable to `_` as long as it is not in translated code. But I am not sure which line made you decide you needed this

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3465,3 +3465,362 @@ def create_zeros_like_3_val():
...
+    def test_bool(a):
+        import numpy as np
+        b = np.real(a)
+        return b
+
+    @types('int')
+    def test_int(a):
+        import numpy as np
+        b = np.real(a)
+        return b
```
EmilyBourne, line 3481: What is the return type in these cases? Could you add type checks to ensure that we calculate the result type correctly please

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3465,3 +3465,362 @@ def create_zeros_like_3_val():
...
+    integer64 = np.int64(randint(1e6))
+    fl = np.float(randint(1e6))
+    fl32 = np.float32(randint(1e6))
+    fl64 = np.float64(randint(1e6))
+
+    # gfortran complains about boolean to numeric convertions, given that boolean type in Fortran
+    # is not a numerical type, so Real function in Fortran doesn't accept a non-numerical type in the first argument
+    if (language == 'c'):
+        f_bl = epyccel(test_bool, language=language)
+        assert (f_bl(bl) == test_bool(bl))
```
EmilyBourne, line 3553: Does this mean that we are missing a call to a casting function in Fortran? Or do you think it is normal that this doesn't work? I'm not even sure what the expected behaviour is so I'm not sure if this is normal or not

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3465,3 +3465,362 @@ def create_zeros_like_3_val():
...
+
+    f_integer = epyccel(test_int, language=language)
+    # int8 variable does not accept negative numbers, see https://github.com/pyccel/pyccel/issues/722
+    #f_integer8 = epyccel(test_int8, language=language)
```
EmilyBourne, line 3557: Could we test this partially by always using a value for `integer8` that is in the valid positive range for an `int8`? (see https://numpy.org/devdocs/user/basics.types.html for min/max values)

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3465,3 +3465,362 @@ def create_zeros_like_3_val():
...
+    f_integer16 = epyccel(test_int16, language=language)
+    f_integer32 = epyccel(test_int32, language=language)
+    f_integer64 = epyccel(test_int64, language=language)
+
+    assert (f_integer(integer) == test_int(integer))
+    # int8 variable does not accept negative numbers, see https://github.com/pyccel/pyccel/issues/722
+    #assert (f_integer8(integer8) == test_int8(integer8))
```
EmilyBourne, line 3564: See above

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3465,3 +3465,362 @@ def create_zeros_like_3_val():
...
+    # gfortran complains about boolean to numeric convertions, given that boolean type in Fortran
+    # is not a numerical type, so Real function in Fortran doesn't accept a non-numerical type in the first argument
+    if (language == 'c'):
+        f_bl = epyccel(test_bool, language=language)
+        assert (f_bl() == test_bool())
+
+    f_integer = epyccel(test_int, language=language)
+    # int8 and int16 numpy data types not recognised by Pyccel.
+    #f_integer8 = epyccel(test_int8, language=language)
+    #f_integer16 = epyccel(test_int16, language=language)
```
EmilyBourne, line 3680: Why is this not a problem in the tests above? Do we accept `int8` arguments but not allow the creation of an array of type `int8`? That is strange ^^'

EmilyBourne (changes requested):

tests/epyccel/recognised_functions/test_numpy_funcs.py
mostafaMamoni: i can get ride of this.

tests/epyccel/recognised_functions/test_numpy_funcs.py
mostafaMamoni: sure

tests/epyccel/recognised_functions/test_numpy_funcs.py
mostafaMamoni: i just discovered that we can tests them by passing a Literal boolean to the functions.

tests/epyccel/recognised_functions/test_numpy_funcs.py
mostafaMamoni: good idea.

tests/epyccel/recognised_functions/test_numpy_funcs.py
EmilyBourne: So there is only a problem when passing a boolean variable? That's odd

tests/epyccel/recognised_functions/test_numpy_funcs.py
EmilyBourne: If that is the case then could you open an issue? This is silly and they are easy to add

tests/epyccel/recognised_functions/test_numpy_funcs.py
mostafaMamoni: it seems Pyccel doesn’t like `import X as Y` it raises ` Undefined variable` error,

EmilyBourne: Hmm, to be clear, this should be `from M import X as Y`. I'm sure it works for functions but maybe there is a problem for data types? If so can you create an issue please

tests/epyccel/recognised_functions/test_numpy_funcs.py
EmilyBourne: I'm not sure I understood what the problem is/was here

tests/epyccel/recognised_functions/test_numpy_funcs.py
EmilyBourne: You re-requested a review, but this hasn't been done, has it?

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3465,3 +3465,354 @@ def create_zeros_like_3_val():
...
+        a = real(arr)
+        s = shape(a)
+        return len(s), s[0], s[1], a[0,1], a[1,0]
+
+    def test_int16():
+        from numpy import real, shape, array, int16
+        arr = array([[4,5,6,2,1],[4,5,6,2,1]], int16)
+        a = real(arr)
+        s = shape(a)
+        return len(s), s[0], s[1], a[0,1], a[1,0]
```
EmilyBourne, line 3737: You may need to comment these functions while they cannot be tested

EmilyBourne (changes requested):
Please fix codacy problems and add the type checks mentioned in the previous review

tests/epyccel/recognised_functions/test_numpy_funcs.py
mostafaMamoni: sorry, i forgot it.

EmilyBourne: Please re-read all unresolved conversations before re-requesting reviews to avoid the situation where I need to review and just point to open conversations. Thanks

tests/epyccel/recognised_functions/test_numpy_funcs.py
mostafaMamoni: sure

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3465,3 +3465,359 @@ def create_zeros_like_3_val():
...
+    f_integer = epyccel(test_int, language=language)
+    # int8 variable does not accept negative numbers, see https://github.com/pyccel/pyccel/issues/722
+    f_integer8 = epyccel(test_int8, language=language)
+    f_integer16 = epyccel(test_int16, language=language)
+    f_integer32 = epyccel(test_int32, language=language)
+    f_integer64 = epyccel(test_int64, language=language)
+
+    assert (f_integer(integer) == test_int(integer))
+    # ensuring that we calculate the result type correctly for numpy int convertions.
+    assert (type(f_integer(integer)) == type(test_int(integer)))
```
EmilyBourne, line 3565: Could you add the type check for at least one numpy type, plus the floats and complexes please

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3465,3 +3465,359 @@ def create_zeros_like_3_val():
...
+        return len(s), s[0], a[0]
+
+    def test_int():
+        from numpy import real, shape, array, int
+        arr = array([4,5,6,2,1], int)
+        a = real(arr)
+        s = shape(a)
+        return len(s), s[0], a[0]
+
+    # def test_int8():
```
EmilyBourne, line 3611: Could you add a comment referencing issue #733 here please.

EmilyBourne (changes requested):

tests/epyccel/recognised_functions/test_numpy_funcs.py
EmilyBourne: Could you explain this further? What was/is the problem?

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -1,4 +1,4 @@
-# pylint: disable=missing-function-docstring, missing-module-docstring, unidiomatic-typecheck/
+# pylint: disable=missing-function-docstring, redefined-builtin, missing-module-docstring, unidiomatic-typecheck/
```
EmilyBourne, line 1: The remaining codacy error was about `float` not `int` or `bool` so I don't see the relevance for issue #733

tests/epyccel/recognised_functions/test_numpy_funcs.py
EmilyBourne: I would like to understand this before this PR is approved/merged and we forget about it

mostafaMamoni: after casting arg in PythonReal we dont need to pass a Literal boolean to the functions for the tests, so now we can pass it by a variable.

pyccel/ast/builtins.py
```diff
@@ -299,6 +301,8 @@ def __new__(cls, arg):
 
     def __init__(self, arg):
         self._arg = arg
+        self._shape = arg.shape
+        self._rank  = len(self._shape)
```
EmilyBourne, line 305: If we add this functionality to `PythonInt` then we should also add it to the other casting functions

pyccel/ast/numpyext.py
```diff
@@ -140,6 +140,11 @@ class NumpyReal(PythonReal):
     > np.real(a)
     1.0
     """
+    def __init__(self, arg):
+        super().__init__(arg)
+        self._precision = arg.precision
+        self._shape = process_shape(self.internal_var.shape)
+        self._rank  = len(self._shape)
```
EmilyBourne, line 147: Does this not need doing for `NumpyImag` too?

mostafaMamoni: l thought , it will be done in another PR, what do you think ?

pyccel/ast/builtins.py
mostafaMamoni: i think we should add it if we need it to prevent useless lines.

pyccel/ast/numpyext.py
EmilyBourne: There is very little difference between `NumpyReal` and `NumpyImag`. They are complimentary and it is very unexpected for one to implement more things than the other. This is why `PythonReal` and `PythonImag` have a common ancestor `PythonComplexProperty`, so I am not super comfortable with the idea of leaving this for later. Is it a lot of work?

pyccel/ast/builtins.py
EmilyBourne: Avoiding useless lines is good, but having different behaviour between casting functions is an easy way to confuse developers

pyccel/ast/numpyext.py
mostafaMamoni: with tests ? yes, especially if we discovered something.

pyccel/ast/builtins.py
mostafaMamoni: but we dont where we can add it now, given that the casting fine to me now.

pyccel/ast/numpyext.py
mostafaMamoni: > so I am not super comfortable with the idea of leaving this for later.
it will be done directly after this one to keep on track the theme of these two PRs.

pyccel/ast/builtins.py
EmilyBourne: > but we dont know where we can add it now

What do you mean? We know which classes are casting functions. Why would we not be able to add this to one of them?

pyccel/ast/numpyext.py
EmilyBourne: ok

pyccel/ast/builtins.py
mostafaMamoni: by asking where i meant specifically, but yeah you are right about "having different behaviour between casting functions is an easy way to confuse developers"

EmilyBourne: We should remove the [default values](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/builtins.py#L142-L143) now that we calculate this systematically

pyccel/ast/builtins.py
```diff
@@ -122,6 +124,8 @@ def __new__(cls, arg):
...
```
EmilyBourne, line 128: The default values set in the class are no longer needed

pyccel/ast/builtins.py
```diff
@@ -268,6 +272,8 @@ def __new__(cls, arg):
...
```
EmilyBourne, line 276: See above

pyccel/ast/builtins.py
EmilyBourne: >We should remove the default values now that we calculate this systematically

EmilyBourne (changes requested):

pyccel/ast/builtins.py
```diff
@@ -176,6 +176,7 @@ def __new__(cls, arg0, arg1=LiteralFloat(0)):
         return super().__new__(cls, arg0, arg1)
 
     def __init__(self, arg0, arg1 = LiteralFloat(0)):
+        self._shape = arg0.shape if arg0.shape > arg1.shape else arg1.shape
```
EmilyBourne, line 179: As we are extending the behaviour of `PythonComplex` beyond what it does in python I don't think we need to handle strange corner cases like this
In addition what happens when `arg0.shape` is `PyccelArraySize(Variable(x),0)`?

mostafaMamoni: ah ok.

EmilyBourne (approved):

pyccel/ast/builtins.py
```diff
@@ -80,7 +80,9 @@ class PythonReal(PythonComplexProperty):
     arg : Variable, Literal
     """
     def __new__(cls, arg):
-        if arg.dtype is not NativeComplex():
+        if isinstance(arg.dtype, NativeBool):
+            return PythonInt(arg)
+        elif arg.dtype is not NativeComplex():
```
yguclu, line 85: Is there a reason why we do not write
```python
if arg.dtype is not NativeBool():
```
or
```python
if not isinstance(arg.dtype, NativeComplex):
```
?

mostafaMamoni: no, we can work with one of them, its just because i always work with `isinstance`.

saidctb: we should use `isinstance` if the types are no longer singletons, if they are we can use `is`

EmilyBourne: `DataType` objects are singletons

yguclu: @mostafaMamoni Whatever solution you choose, please use the same everywhere in the `if/elif/else` block.

yguclu (approved):

PR #744, opened 2021-02-11
[BUGFIX] Break infinite recursion

### LINKED ISSUE #743: Infinite loop when a recursive function is defined in another function
**Describe the bug**
When we create a `FunctionDef` we know its signature but nothing else. That signature needs to be inserted into the namespace so that recursive calls can find it. The signature is an empty `FunctionDef` defined [here](https://github.com/pyccel/pyccel/blob/master/pyccel/parser/semantic.py#L2560) . Once we have collected the body we then create the real `FunctionDef` here .
The original `FunctionDef` is therefore a temporary which should be forgotten about after this point.

Before PR #718 we left the first `FunctionDef` in place, the signature is all that we need for printing so it is not a problem if we don't have the body.

However when objects have users this leads to a strange situation. Any `Variable`s that are arguments of the function have both `FunctionDef` objects as users. If you then want to rename an argument using `substitute(a,b)` you will find that `a` still has one parent at the end of this operation (the signature `FunctionDef`) this is not a problem per se as the signature is not used, but it is unexpected (if the signature is used then it will be in the tree so the argument will be updated leaving `a` userless as expected).
As a result I thought it made sense to only have one `FunctionDef` instance for each function. So I replace all instances of the signature with the full `FunctionDef`. However when you then try to call one of the recursive functions on an object containing this `FunctionDef` we have an infinite loop

**To Reproduce**
```python
@types('int')
def omp_tasks(x):
    @types('int', results='int')
    def fib(n):
        if n < 2:
            return n
        i = fib(n-1)
        j = fib(n-2)
        return i + j
    m = fib(x)
    return m
print(omp_tasks(10))
```

Causes an infinite loop in the `substitute` function

cc: @nhamidn

EmilyBourne:
Keep function signature in recursive function call instead of replacing it with the complete function definition. Fixes #743

In addition an attribute `_recursion_in_progress` is added to `pyccel.ast.basic.Basic` to break any future accidental recursion

### DIFF
--- a/pyccel/ast/basic.py
+++ b/pyccel/ast/basic.py
@@ -106,7 +106,6 @@ class Basic(sp_Basic):
                requested type which contain self
         """
         if self._recursion_in_progress:
-            self._recursion_in_progress = False
             return []
 
         if len(self._user_nodes) == 0:
@@ -140,7 +139,6 @@ class Basic(sp_Basic):
                requested type which exist in self
         """
         if self._recursion_in_progress:
-            self._recursion_in_progress = False
             return
         self._recursion_in_progress = True
 
@@ -186,7 +184,6 @@ class Basic(sp_Basic):
                       Types for which substitute should not be called
         """
         if self._recursion_in_progress:
-            self._recursion_in_progress = False
             return
         self._recursion_in_progress = True
 
--- a/pyccel/ast/basic.py
+++ b/pyccel/ast/basic.py
@@ -105,10 +105,7 @@ class Basic(sp_Basic):
         list : List containing all objects of the
                requested type which contain self
         """
-        if self._recursion_in_progress:
-            return []
-
-        if len(self._user_nodes) == 0:
+        if self._recursion_in_progress or len(self._user_nodes) == 0:
             return []
         else:
             self._recursion_in_progress = True
@@ -139,7 +136,7 @@ class Basic(sp_Basic):
                requested type which exist in self
         """
         if self._recursion_in_progress:
-            return
+            return []
         self._recursion_in_progress = True
 
         results = []
--- a/pyccel/ast/basic.py
+++ b/pyccel/ast/basic.py
@@ -199,11 +199,12 @@ class Basic(sp_Basic):
 
         def prepare_sub(found_node):
             idx = original.index(found_node)
+            rep = replacement[idx]
             if not self.ignore(found_node):
                 found_node.remove_user_node(self)
-            if not self.ignore(replacement[idx]):
-                replacement[idx].set_current_user_node(self)
-            return replacement[idx]
+            if not self.ignore(rep):
+                rep.set_current_user_node(self)
+            return rep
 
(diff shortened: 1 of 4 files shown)


pyccel/ast/basic.py
```diff
@@ -183,8 +199,10 @@ def substitute(self, original, replacement, excluded_nodes = ()):
 
         def prepare_sub(found_node):
             idx = original.index(found_node)
-            found_node.remove_user_node(self)
-            replacement[idx].set_current_user_node(self)
+            if not self.ignore(found_node):
+                found_node.remove_user_node(self)
+            if not self.ignore(replacement[idx]):
+                replacement[idx].set_current_user_node(self)
             return replacement[idx]
```
yguclu, line 206: I wonder if the definition of this function inside the `substitute` method could slow down the Python interpreter, especially when `substitute` is called recursively. This deserves a bit of profiling on a complex example.

If it does, you could define the method (where I reduced the number of calls to `__getitem__`from 3 to just 1)
```python
def prepare_substitution(self, original, replacement, node):
    idx = original.index(node)
    rep = replacement[idx]
    if not self.ignore(node):
        node.remove_user_node(self)
    if not self.ignore(rep):
        rep.set_current_user_node(self)
    return rep
```
and then use a lambda function for clarity:
```python
prepare_sub = lambda found_node: self.prepare_substitution(original, replacement, found_node)
```

nhamidn (approved):

pyccel/ast/basic.py
EmilyBourne: I tried to do some simple profiling using the `test_substitution` method. As a rule of thumb it seems that defining a `prepare_substitution(self, original, replacement, node)` with a lambda function causes a slowdown (3rd significant digit), I suspect that creating a lambda function is not much cheaper than creating a function.
Using the function directly instead of via a lambda function seems to cause a slight speedup (4th significant digit).

But these results should be taken with a grain of salt as my computer is not very reliable. The results were found by running the test 1000 times using timeit. Each set of 1000 took around 10s (but closing internet made me gain a second, which shows how trustworthy this is :/ )

pyccel/ast/basic.py
```diff
@@ -132,6 +138,9 @@ def get_attribute_nodes(self, search_type, excluded_nodes = ()):
         list : List containing all objects of the
                requested type which exist in self
         """
+        if self._recursion_in_progress:
+            return
```
MenoIy, line 142: shouldn't this return empty list ?

EmilyBourne: Good catch. Fixed in 1d0f330

MenoIy (approved):

pyccel/ast/basic.py
EmilyBourne: For now I have just fixed the `__getitem__` calls (also 4th significant difference). See 4e00d21

yguclu (approved):

PR #750, opened 2021-02-13
Avoid printing unused variables in C

mostafaMamoni:
Currently, Pyccel prints unnecessary temporary variables in C, except when using an indexed element on the right-hand side of the assignment, So this pull request will solve this issue by:
- Adding `is_temp` attribute to `Variable` class, and use it in `print_functionDef` in ccode.py to undeclare unused variables,
- re-implement `print_return` in ccode.py, so we do not print any unnecessary temporary variable anymore, considering the deallocations nodes, and return directly the assigned value to that temporary variable.

### DIFF
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -21,7 +21,7 @@ from pyccel.errors.errors import Errors
 from pyccel.errors.messages import RECURSIVE_RESULTS_REQUIRED
 
 from .basic     import Basic, PyccelAstNode
-from .builtins  import (PythonEnumerate, PythonLen, PythonMap,
+from .builtins  import (PythonEnumerate, PythonLen, PythonMap, PythonTuple,
                         PythonRange, PythonZip, PythonBool, Lambda)
 from .datatypes import (datatype, DataType, NativeSymbol,
                         NativeBool, NativeRange,
@@ -419,6 +419,10 @@ class Assign(Basic):
         status=None,
         like=None,
         ):
+        if isinstance(lhs, (tuple, list)):
+            lhs = PythonTuple(*lhs)
+        if isinstance(rhs, (tuple, list)):
+            rhs = PythonTuple(*rhs)
         self._lhs = lhs
         self._rhs = rhs
         self._status = status
@@ -643,7 +647,7 @@ class CodeBlock(Basic):
         for i in body:
             if isinstance(i, CodeBlock):
                 ls += i.body
-            else:
+            elif i is not None and not isinstance(i, EmptyNode):
                 ls.append(i)
         self._body = tuple(ls)
         super().__init__()
@@ -658,6 +662,9 @@ class CodeBlock(Basic):
 
     def insert2body(self, obj):
         self._body = tuple(self.body + (obj,))
+
+    def __str__(self):
+        return 'CodeBlock({})'.format(self.body)
 
 class AliasAssign(Basic):
 
@@ -2149,17 +2156,17 @@ class FunctionDef(Basic):
 
         self._name = newname
 
-    def add_local_var(self, var):
+    def add_local_vars(self, *variables):
         """
-        Add a new variable to the local variables tuple
+        Add (a) new variable(s) to the local variables tuple
 
         Parameters
         ----------
         var : Variable
               The new local variable
         """
-        var.set_current_user_node(self)
-        self._local_vars = self._local_vars + (var,)
+        _ = [v.set_current_user_node(self) for v in variables]
+        self._local_vars += variables
 
     def __getnewargs__(self):
         """
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -248,9 +248,7 @@ def create_variable(forbidden_names, prefix = None, counter = 1):
 
     name, counter = create_incremented_string(forbidden_names, prefix, counter = counter)
 
-    ps = PyccelSymbol(name)
-    ps.is_temp = True
-    return ps, counter
+    return PyccelSymbol(name, is_temp=True), counter
 
 
 class AsName(Basic):
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -1749,8 +1749,8 @@ class Return(Basic):
 
     def __init__(self, expr, stmt=None):
 
-        if stmt and not isinstance(stmt, (Assign, CodeBlock)):
-            raise TypeError('stmt should only be of type Assign')
+        if stmt and not isinstance(stmt, CodeBlock):
+            raise TypeError('stmt should only be of type CodeBlock')
 
         self._expr = expr
         self._stmt = stmt
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -7,20 +7,15 @@
 
 from collections     import OrderedDict
 
-from sympy import preorder_traversal
-
-from sympy.core.compatibility import with_metaclass
-from sympy.core.singleton     import Singleton
-from sympy.core.expr          import Expr
 from sympy.logic.boolalg      import And as sp_And
-
-from sympy.utilities.iterables          import iterable
 
 
 from pyccel.errors.errors import Errors
 from pyccel.errors.messages import RECURSIVE_RESULTS_REQUIRED
 
-from .basic     import Basic, PyccelAstNode
+from pyccel.utilities.metaclasses import Singleton
+
+from .basic     import Basic, PyccelAstNode, iterable
 from .builtins  import (PythonEnumerate, PythonLen, PythonMap, PythonTuple,
                         PythonRange, PythonZip, PythonBool, Lambda)
 from .datatypes import (datatype, DataType, NativeSymbol,
@@ -169,44 +164,11 @@ def subs(expr, new_elements):
         new_expr = expr.subs(new_elements)
         new_expr.set_fst(expr.fst)
         return new_expr
-    elif isinstance(expr, Expr):
+    elif isinstance(expr, PyccelAstNode):
         return expr.subs(new_elements)
 
     else:
         return expr
-
-
-
-#def collect_vars(ast):
-#    """ collect variables in order to be declared"""
-#    #TODO use the namespace to get the declared variables
-#    variables = {}
-#    def collect(stmt):
-#
-#        if isinstance(stmt, Variable):
-#            if not isinstance(stmt.name, DottedName):
-#                variables[stmt.name] = stmt
-#        elif isinstance(stmt, (tuple, list)):
-#            for i in stmt:
-#                collect(i)
-#        if isinstance(stmt, For):
-#            collect(stmt.target)
-#            collect(stmt.body)
-#        elif isinstance(stmt, FunctionalFor):
-#            collect(stmt.lhs)
-#            collect(stmt.loops)
-#        elif isinstance(stmt, If):
-#            collect(stmt.bodies)
-#        elif isinstance(stmt, (While, CodeBlock)):
-#            collect(stmt.body)
-#        elif isinstance(stmt, (Assign, AliasAssign, AugAssign)):
-#            collect(stmt.lhs)
-#            if isinstance(stmt.rhs, (Linspace, Diag, Where)):
-#                collect(stmt.rhs.index)
-#
-#
-#    collect(ast)
-#    return variables.values()
 
 def inline(func, args):
     local_vars = func.local_vars
@@ -322,9 +284,8 @@ class AsName(Basic):
     def target(self):
         return self._target
 
-    def _sympystr(self, printer):
-        sstr = printer.doprint
-        return '{0} as {1}'.format(sstr(self.name), sstr(self.target))
+    def __str__(self):
+        return '{0} as {1}'.format(str(self.name), str(self.target))
 
     def __eq__(self, string):
         if isinstance(string, str):
@@ -342,8 +303,8 @@ class Dlist(PyccelAstNode):
 
     Parameters
     ----------
-    value : Expr
-           a sympy expression which represents the initilized value of the list
+    value : PyccelAstNode
+           an expression which represents the initilized value of the list
 
     shape : the shape of the array
     """
@@ -372,7 +333,7 @@ class Assign(Basic):
 
     Parameters
     ----------
-    lhs : Expr
+    lhs : PyccelAstNode
         In the syntactic stage:
            Object representing the lhs of the expression. These should be
            singular objects, such as one would use in writing code. Notable types
@@ -381,7 +342,7 @@ class Assign(Basic):
         In the semantic stage:
            Variable or IndexedElement
 
-    rhs : Expr
+    rhs : PyccelAstNode
         In the syntactic stage:
           Object representing the rhs of the expression
         In the semantic stage :
@@ -429,9 +390,8 @@ class Assign(Basic):
         self._like = like
         super().__init__()
 
-    def _sympystr(self, printer):
-        sstr = printer.doprint
-        return '{0} := {1}'.format(sstr(self.lhs), sstr(self.rhs))
+    def __str__(self):
+        return '{0} := {1}'.format(str(self.lhs), str(self.rhs))
 
     @property
     def lhs(self):
@@ -567,10 +527,9 @@ class Allocate(Basic):
     def status(self):
         return self._status
 
-    def _sympystr(self, printer):
-        sstr = printer.doprint
+    def __str__(self):
         return 'Allocate({}, shape={}, order={}, status={})'.format(
-                sstr(self.variable), sstr(self.shape), sstr(self.order), sstr(self.status))
+                str(self.variable), str(self.shape), str(self.order), str(self.status))
 
     def __eq__(self, other):
         if isinstance(other, Allocate):
@@ -666,6 +625,27 @@ class CodeBlock(Basic):
     def __str__(self):
         return 'CodeBlock({})'.format(self.body)
 
+    def __reduce_ex__(self, i):
+        """ Used by pickle to create an object of this class.
+
+          Parameters
+          ----------
+
+          i : int
+           protocol
+
+          Results
+          -------
+
+          out : tuple
+           A tuple of two elements
+           a callable function that can be called
+           to create the initial version of the object
+           and its arguments.
+        """
+        kwargs = dict(body = self.body)
+        return (apply, (self.__class__, (), kwargs))
+
 class AliasAssign(Basic):
 
     """Represents aliasing for code generation. An alias is any statement of the
@@ -706,9 +686,8 @@ class AliasAssign(Basic):
         self._rhs = rhs
         super().__init__()
 
-    def _sympystr(self, printer):
-        sstr = printer.doprint
-        return '{0} := {1}'.format(sstr(self.lhs), sstr(self.rhs))
+    def __str__(self):
+        return '{0} := {1}'.format(str(self.lhs), str(self.rhs))
 
     @property
     def lhs(self):
@@ -747,9 +726,8 @@ class SymbolicAssign(Basic):
         self._rhs = rhs
         super().__init__()
 
-    def _sympystr(self, printer):
-        sstr = printer.doprint
-        return '{0} := {1}'.format(sstr(self.lhs), sstr(self.rhs))
+    def __str__(self):
+        return '{0} := {1}'.format(str(self.lhs), str(self.rhs))
 
     @property
     def lhs(self):
@@ -760,11 +738,11 @@ class SymbolicAssign(Basic):
         return self._rhs
 
 
-# The following are defined to be sympy approved nodes. If there is something
+# The following were defined to be sympy approved nodes. If there is something
 # smaller that could be used, that would be preferable. We only use them as
 # tokens.
 
-class NativeOp(with_metaclass(Singleton)):
+class NativeOp(metaclass=Singleton):
 
     """Base type for native operands."""
 
@@ -814,7 +792,7 @@ class AugAssign(Assign):
 
     Parameters
     ----------
-    lhs : Expr
+    lhs : PyccelAstNode
         In the syntactic stage:
            Object representing the lhs of the expression. These should be
            singular objects, such as one would use in writing code. Notable types
@@ -826,7 +804,7 @@ class AugAssign(Assign):
     op : str
         Operator (+, -, /, \*, %).
 
-    rhs : Expr
+    rhs : PyccelAstNode
         In the syntactic stage:
           Object representing the rhs of the expression
         In the semantic stage :
@@ -867,10 +845,9 @@ class AugAssign(Assign):
 
         super().__init__(lhs, rhs, status, like)
 
-    def _sympystr(self, printer):
-        sstr = printer.doprint
-        return '{0} {1}= {2}'.format(sstr(self.lhs), self.op._symbol,
-                sstr(self.rhs))
+    def __str__(self):
+        return '{0} {1}= {2}'.format(str(self.lhs), self.op._symbol,
+                str(self.rhs))
 
     @property
     def op(self):
@@ -887,9 +864,9 @@ class While(Basic):
 
     Parameters
     ----------
-    test : expression
-        test condition given as a sympy expression
-    body : sympy expr
+    test : PyccelAstNode
+        test condition given as an expression
+    body : list of Pyccel objects
         list of statements representing the body of the While statement.
 
     Examples
@@ -941,9 +918,9 @@ class With(Basic):
 
     Parameters
     ----------
-    test : expression
-        test condition given as a sympy expression
-    body : sympy expr
+    test : PyccelAstNode
+        test condition given as an expression
+    body : list of Pyccel objects
         list of statements representing the body of the With statement.
 
     Examples
@@ -1327,11 +1304,11 @@ class For(Basic):
 
     Parameters
     ----------
-    target : symbol
+    target : symbol / Variable
         symbol representing the iterator
     iter : iterable
         iterable object. for the moment only Range is used
-    body : sympy expr
+    body : list of pyccel objects
         list of statements representing the body of the For statement.
 
     Examples
@@ -1488,12 +1465,11 @@ class ConstructorCall(Basic):
         self._arguments = arguments
         super().__init__()
 
-    def _sympystr(self, printer):
-        sstr = printer.doprint
-        name = sstr(self.name)
+    def __str__(self):
+        name = str(self.name)
         args = ''
         if not self.arguments is None:
-            args = ', '.join(sstr(i) for i in self.arguments)
+            args = ', '.join(str(i) for i in self.arguments)
         return '{0}({1})'.format(name, args)
 
     @property
@@ -1601,11 +1577,9 @@ class ValuedArgument(Basic):
     def is_kwonly(self):
         return self._kwonly
 
-    def _sympystr(self, printer):
-        sstr = printer.doprint
-
-        argument = sstr(self.argument)
-        value = sstr(self.value)
+    def __str__(self):
+        argument = str(self.argument)
+        value = str(self.value)
         return '{0}={1}'.format(argument, value)
 
 class FunctionCall(PyccelAstNode):
@@ -1747,7 +1721,7 @@ class Return(Basic):
 
     Parameters
     ----------
-    expr : sympy expr
+    expr : PyccelAstNode
         The expression to return.
 
     stmts :represent assign stmts in the case of expression return
@@ -1858,11 +1832,6 @@ class FunctionDef(Basic):
                  '_functions',
                  '_interfaces'
                  )
-
-    def __new__(cls, *args, **kwargs):
-        kwargs.pop('decorators', None)
-        kwargs.pop('templates', None)
-        return super().__new__(cls, *args, **kwargs)
 
     def __init__(
         self,
@@ -2424,10 +2393,6 @@ class ValuedFunctionAddress(FunctionAddress):
     """
     _attribute_nodes = (*FunctionAddress._attribute_nodes, '_value')
 
-    def __new__(cls, *args, **kwargs):
-        kwargs.pop('value', Nil())
-        return super().__new__(cls, *args, **kwargs)
-
     def __init__(self, *args, **kwargs):
         self._value = kwargs.pop('value', Nil())
         super().__init__(*args, **kwargs)
@@ -2463,8 +2428,6 @@ class BindCFunctionDef(FunctionDef):
         The function from which the c-compatible version was created
     """
     _attribute_nodes = (*FunctionDef._attribute_nodes, '_original_function')
-    def __new__(cls, *args, original_function, **kwargs):
-        return super().__new__(cls, *args, **kwargs)
 
     def __init__(self, *args, original_function, **kwargs):
         self._original_function = original_function
@@ -2729,9 +2692,6 @@ class ClassDef(Basic):
         else:
             return self.is_iterable or self.is_with_construct
 
-    def _eval_subs(self, old , new):
-        return self
-
     @property
     def is_unused(self):
         return False
@@ -2809,13 +2769,12 @@ class Import(Basic):
             raise TypeError('to_ignore must be a boolean.')
         self._ignore_at_print = to_ignore
 
-    def _sympystr(self, printer):
-        sstr = printer.doprint
-        source = sstr(self.source)
+    def __str__(self):
+        source = str(self.source)
         if len(self.target) == 0:
             return 'import {source}'.format(source=source)
         else:
-            target = ', '.join([sstr(i) for i in self.target])
+            target = ', '.join([str(i) for i in self.target])
             return 'from {source} import {target}'.format(source=source,
                     target=target)
 
@@ -2843,7 +2802,7 @@ class FuncAddressDeclare(Basic):
         An instance of FunctionAddress.
     intent: None, str
         one among {'in', 'out', 'inout'}
-    value: Expr
+    value: PyccelAstNode
         variable value
     static: bool
         True for a static declaration of an array.
@@ -2922,7 +2881,7 @@ class Declare(Basic):
         Variables must be of the same type.
     intent: None, str
         one among {'in', 'out', 'inout'}
-    value: Expr
+    value: PyccelAstNode
         variable value
     static: bool
         True for a static declaration of an array.
@@ -3026,8 +2985,8 @@ class SymbolicPrint(Basic):
 
     Parameters
     ----------
-    expr : sympy expr
-        The expression to return.
+    expr : PyccelAstNode
+        The expression to print
 
     Examples
     --------
@@ -3111,7 +3070,7 @@ class EmptyNode(Basic):
     """
     _attribute_nodes = ()
 
-    def _sympystr(self, printer):
+    def __str__(self):
         return ''
 
 
@@ -3140,9 +3099,29 @@ class Comment(Basic):
     def text(self):
         return self._text
 
-    def _sympystr(self, printer):
-        sstr = printer.doprint
-        return '# {0}'.format(sstr(self.text))
+    def __str__(self):
+        return '# {0}'.format(str(self.text))
+
+    def __reduce_ex__(self, i):
+        """ Used by pickle to create an object of this class.
+
+          Parameters
+          ----------
+
+          i : int
+           protocol
+
+          Results
+          -------
+
+          out : tuple
+           A tuple of two elements
+           a callable function that can be called
+           to create the initial version of the object
+           and its arguments.
+        """
+        kwargs = dict(text = self.text)
+        return (apply, (self.__class__, (), kwargs))
 
 
 class SeparatorComment(Comment):
@@ -3265,7 +3244,7 @@ class Assert(Basic):
 
     Parameters
     ----------
-    test: Expr
+    test: PyccelAstNode
         boolean expression to check
 
     Examples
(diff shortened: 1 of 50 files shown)


pyccel/parser/semantic.py
```diff
@@ -2362,9 +2362,12 @@ def _visit_Return(self, expr, **settings):
             if not (isinstance(r, PyccelSymbol) and r == v):
                 a = Assign(v, r)
                 a.set_fst(expr.fst)
+                a = self._visit_Assign(a)
+                ind = a.rhs.get_attribute_nodes(IndexedElement)
+                if len(ind) == 0 and not isinstance(a.rhs, IndexedElement):
+                    a.lhs.is_temp = True
```
MenoIy, line 2368: Im not sure about this . if the rhs of the Assign is not an indexed element and it don't contain an indexed element
you take it as temporary variable .
and in the printing stage you search for deallocate node 
I think this will not work for arguments example 
```PYTHON
@types('int[:]')
def func(c):
    return c[0]
```

pyccel/parser/semantic.py
```diff
@@ -2364,7 +2364,7 @@ def _visit_Return(self, expr, **settings):
                 a.set_fst(expr.fst)
                 a = self._visit_Assign(a)
                 ind = a.rhs.get_attribute_nodes(IndexedElement)
-                if len(ind) == 0 and not isinstance(a.rhs, IndexedElement):
+                if len(ind) == 0 and not isinstance(a.rhs, IndexedElement) or isinstance(a.rhs, IndexedElement) and a.rhs.base.is_argument:
```
MenoIy, line 2367: hmm, are you sure about this ? 
did you try to test something like this ? when the rhs is not and indexed Element but it contain indexed Element of an argument example PyccelAdd 
``` PYTHON
@types('int[:]', 'int[:]')
def test(a, b):
        return a[0] + b[0]
```
result : 
```C
#include "avoid.h"
#include <stdint.h>
#include <stdlib.h>
#include "ndarrays.h"

int64_t test(t_ndarray a, t_ndarray b)
{
    int64_t Out_0001;
    return a.nd_int64[get_index(a, 0)] + b.nd_int64[get_index(b, 0)];
}
```

MenoIy (changes requested):

pyccel/ast/variable.py
```diff
@@ -286,6 +290,20 @@ def is_pointer(self, is_pointer):
             raise TypeError('is_pointer must be a boolean.')
         self._is_pointer = is_pointer
 
+    @property
+    def is_temp(self):
+        """
+        Indicates if object assigned to an object that has non-allocated objects
+        or an indexedElements [Default value: False]
```
nhamidn, line 297: im not sure i understand this

pyccel/parser/semantic.py
EmilyBourne: I am also unsure about this. What about a case without an `IndexedElement`
```
def sum(a : 'int[:]'):
    total = 0
    for i in range(a.shape[0]):
        total += a[i]
    return total

def sum_to_n(n : 'int'):
    import numpy as np
    c = np.arange(1,b+1)
    return sum(c)
```

pyccel/parser/semantic.py
```diff
@@ -2362,9 +2362,20 @@ def _visit_Return(self, expr, **settings):
...
+                if len(ind) == 0 and not isinstance(a.rhs, IndexedElement):
+                    a.lhs.is_temp = True
+                elif isinstance(a.rhs, IndexedElement) and a.rhs.base.is_argument:
+                    a.lhs.is_temp = True
+                elif len(ind) > 0:
+                    a.lhs.is_temp = True
+                    for b in ind:
+                        if not b.base.is_argument:
+                            a.lhs.is_temp = False
+                            break
```
EmilyBourne, line 2376: This section sets the `is_temp` property. Can't we add this property to `PyccelSymbol` and then it will be known from creation?

EmilyBourne:
I'm not sure that your definition of `is_temp` is sufficiently general for it to be a useful attribute of `Variable`. Does this implementation match what was discussed with @saidctb ?

pyccel/parser/semantic.py
mostafaMamoni: i think adding this property to `PyccelSymbol` will cause modifications in many places, so it better to keep it in `Variable`, since it already has this kind of properties unlike `PyccelSymbol`

mostafaMamoni:
> I'm not sure that your definition of `is_temp` is sufficiently general for it to be a useful attribute of `Variable`. Does this implementation match what was discussed with @saidctb ?

the definition was "every variable created by Pyccel", and that's what Said mentioned, and i think the definition can be changed depend on the usage, for now, we are using it just to store the information from an `IndexedElement` and then return it.

tests/epyccel/test_return.py
```diff
@@ -0,0 +1,140 @@
...
+    assert (epyc_create_multi_arrs(7) == create_multi_arrs(7))
+
+    epyc_expr_arrs_elements = epyccel(expr_arrs_elements, language=language, fflags=flags)
+    assert (epyc_expr_arrs_elements(7) == expr_arrs_elements(7))
+
+    epyc_complex_expr = epyccel(complex_expr, language=language, fflags=flags)
+    assert (epyc_complex_expr(7) == complex_expr(7))
+
+    epyc_multi_allocs = epyccel(multi_allocs, language=language, fflags=flags)
+    assert (epyc_multi_allocs(7) == multi_allocs(7))
```
EmilyBourne, line 140: Please split this into multiple functions. This helps for a couple of reasons:
1. It is easier to see which method raised the error
2. We can get information about all failures at once instead of just the first failure. If we only wanted the first failure we would have run pytest with the `-x` option

EmilyBourne (changes requested):
The suggestion of @saidctb to have an `is_temp` property for "every variable created by Pyccel" was proposed as this property could be used in multiple places. There are already enough mechanisms in place which would allow you to declare the required variables that I don't think this property should be added if we are changing the definition.

The current definition is:
> Indicates if object assigned to an object that has non-allocated objects
        or an indexedElements [Default value: False]

That means that `b` in the following code is a temp:
```
import numpy as np
a = np.ones(3)
b = a[0]
```
However your implementation does not match this. If the definition is sufficiently complicated that it is hard to write a clear docstring then it may be worth asking the question of whether the property is really useful.

Can you explain why you need to handle the return in the semantic stage and the codegen stage?

Why is the semantic stage so complicated? You are handling every possible option with ifs. This makes it very likely that you have missed something. E.g:
```
def my_sum(a : 'int[:]'):
    total = 0
    for i in range(a.shape[0]):
        total += a[i]
    return total

def sum_to_n(n : 'int'):
    import numpy as np
    c = np.arange(1,n+1)
    return my_sum(c)*2
```
You are interested in the variables in the rhs of the assignment. Why can you not just use `get_attribute_nodes` to help you?

mostafaMamoni:
> The suggestion of @saidctb to have an `is_temp` property for "every variable created by Pyccel" was proposed as this property could be used in multiple places. There are already enough mechanisms in place which would allow you to declare the required variables that I don't think this property should be added if we are changing the definition.
> 
> The current definition is:
> 
> > Indicates if object assigned to an object that has non-allocated objects
> > or an indexedElements [Default value: False]
> 
> That means that `b` in the following code is a temp:
> 
> ```
> import numpy as np
> a = np.ones(3)
> b = a[0]
> ```
> 
> However your implementation does not match this. If the definition is sufficiently complicated that it is hard to write a clear docstring then it may be worth asking the question of whether the property is really useful.
> 
> Can you explain why you need to handle the return in the semantic stage and the codegen stage?
> 
> Why is the semantic stage so complicated? You are handling every possible option with ifs. This makes it very likely that you have missed something. E.g:
> 
> ```
> def my_sum(a : 'int[:]'):
>     total = 0
>     for i in range(a.shape[0]):
>         total += a[i]
>     return total
> 
> def sum_to_n(n : 'int'):
>     import numpy as np
>     c = np.arange(1,n+1)
>     return my_sum(c)*2
> ```
> 
> You are interested in the variables in the rhs of the assignment. Why can you not just use `get_attribute_nodes` to help you?

i tried `get_attribute_nodes` for FunctionCall object, it return `Variable` objects in `args` + `Variable` object in the definition of the function, for example this case:
```python
def sum_(a : 'int[:]'):
    total = 0
    for i in range(a.shape[0]):
        total += a[i]
    return total

def sum_to_n(n : 'int'):
    import numpy as np
    c = np.arange(1,n+1)
    return sum_(c)
```
will return `[c, a, total, total, i, a, total, a, i, i, total, i]` which is not really helpful.
About `is_temp` property, we can re-define it with "Indicates if the variable is created by Pyccel".

EmilyBourne:
> i tried get_attribute_nodes for FunctionCall object, it return Variable objects in args + Variable objects in the definition of the function, for example this case:
> ```
> def sum_(a : 'int[:]'):
>     total = 0
>     for i in range(a.shape[0]):
>         total += a[i]
>     return total
> 
> def sum_to_n(n : 'int'):
>     import numpy as np
>     c = np.arange(1,n+1)
>     return sum_(c)
> ```
> will return [c, a, total, total, i, a, total, a, i, i, total, i] which is not really helpful.

Try excluding the contents of a `FunctionDef`:
`expr.rhs.get_attribute_nodes(Variable, excluded_nodes=(FunctionDef,))`

pyccel/parser/semantic.py
```diff
@@ -2362,9 +2362,15 @@ def _visit_Return(self, expr, **settings):
             if not (isinstance(r, PyccelSymbol) and r == v):
                 a = Assign(v, r)
                 a.set_fst(expr.fst)
+                a = self._visit_Assign(a)
+                variables = a.rhs.get_attribute_nodes(Variable, excluded_nodes=(FunctionDef,))
+                a.lhs.is_temp = True
+                for b in variables:
+                    if not b.is_argument and b.allocatable:
+                        a.lhs.is_temp = False
+                        break
```
EmilyBourne, line 2371: This can be simplified
```
a.lhs.is_temp = not any(b.allocatable and not b.is_argument for b in variables)
```

pyccel/codegen/printing/ccode.py
```diff
@@ -1202,13 +1206,37 @@ def _print_Constant(self, expr):
...
-            code += 'return 0;'
-        return code
+            # get Assign nodes form CodeBlock object expr.stmt
+            assigns = expr.stmt.get_attribute_nodes(Assign)
+
+            # check the Assign objects list in case of
+            # the user assigns a variable to an object contains IndexedElement object
+            if not assigns:
+                return 'return {0};'.format(self._print(args[0]))
+            last_assign = assigns[-1]
```
EmilyBourne, line 1223: Can there be more than one assign? If so how do we know that we can discard the others?

pyccel/codegen/printing/ccode.py
```diff
@@ -1202,13 +1206,37 @@ def _print_Constant(self, expr):
...
+            deallocations = expr.stmt.get_attribute_nodes(Deallocate)
+
+            # print deallocation nodes
+            if deallocations:
+                code += '\n' + self._print(last_assign)
+                for a in deallocations:
+                    code += '\n' + self._print(a)
```
EmilyBourne, line 1230: Can `Return.stmt` contain anything other than deallocates and an extra assign? Currently there is nothing stopping us including anything else so I'm not sure we can definitely throw away anything else. Could we not simply do:
```
if last_assign.lhs.is_temp:
    code = '\n'.join(l for l in return.stmt if l is not last_assign)
    code += '\nreturn {}'.format(last_assign.rhs)
else:
    code = '\n'.join(l for l in return.stmt)
    code += '\nreturn {}'.format(return.val)
```
?
(Sorry my code is not 100% correct or neat, I'm on my phone. Hopefully it's enough to get the idea)

pyccel/codegen/printing/ccode.py
```diff
@@ -1202,13 +1206,37 @@ def _print_Constant(self, expr):
...
+
+            # return the rhs(what the temporary variable is holding) if the temporary variable is not holding a value from a freed variable,
+            # else return lhs(the temporary variable).
+            freed_vars = [d.variable for d in deallocations]
+            if all(v not in freed_vars for v in last_assign.rhs.get_attribute_nodes(Variable)):
```
EmilyBourne, line 1235: Hasn't this already been checked in the semantic parser?

EmilyBourne (changes requested):

pyccel/codegen/printing/ccode.py
mostafaMamoni: According to the `Return` docstring `stmt` "represent assign stmts in the case of expression return", so i think it will be one assign unless you have a test case, i was thinking about multi-values return with expressions this should be done in this statement.
```python
        if len(args) > 1:
            if expr.stmt:
                return self._print(expr.stmt)+'\n'+'return 0;'
            return 'return 0;'
```

EmilyBourne: No, I think you are right. I was just wondering whether it made more sense to do:
```
assert(len(assigns)==1)
last_assign = assigns[0]
```
rather than `assigns[-1]` as `assigns[-1]` gives the impression that there could be a few

mostafaMamoni: `assert(len(assigns)==1)` is a good idea, we can use it to make sure that `stmt` has one assign

pyccel/codegen/printing/ccode.py
mostafaMamoni: sorry, i forgot to use `is_temp` attribute, it is really useful in this case

MenoIy (approved):

EmilyBourne:
Could you use your solution to fix the same problem in the python printer?

EmilyBourne:
@saidctb are you happy with the way `is_temp` is used in this PR?

pyccel/parser/semantic.py
```diff
@@ -2345,12 +2345,14 @@ def _visit_Return(self, expr, **settings):
         return_vars = self.get_function(f_name).results
         assigns     = []
         for v,r in zip(return_vars, results):
-            if not (isinstance(r, PyccelSymbol) and r == v):
+            if not (isinstance(r, PyccelSymbol) and r == (v.name if isinstance(v, Variable) else v)):
```
saidctb, line 2348: why is the type of v  `Variable`, the nodes are not annotated yet no?

EmilyBourne: `v` is usually a `PyccelSymbol` but if the results are specified (e.g. in the case of a recursive function) then `v` is a `Variable` created [here](https://github.com/pyccel/pyccel/blob/master/pyccel/parser/semantic.py#L2531-L2541)

saidctb: ah okay

pyccel/parser/semantic.py
saidctb: why is it going to cause modifications in many places?

mostafaMamoni: I meant if we add it now we wont use it, since we only count on `Variable` object, and as i said its better to keep `PyccelSymbol` clean, and add this kind of attributes to `Variable` since it had a lot of them, unless you have some critical enhancement that requires `is_temp` in `PyccelSymbol`.

mostafaMamoni:
@EmilyBourne sorry if i can't work with `_additional_declares` because it cause a lot of checks, and when i check it in `FunctionDef` here`if isinstance(i, Variable) and i not in self._additional_declares:` user's variables will not get declared, here is `_additional_declares` version :
```python
            if not last_assign:
                if isinstance(args[0], Variable) and not args[0].is_argument:
                    self._additional_declare.append(args[0])
                return 'return {0};'.format(self._print(args[0]))

            assert(len(last_assign)==1)
            variables = last_assign[0].rhs.get_attribute_nodes(Variable, excluded_nodes=(FunctionDef,))
            should_print = not any(b.allocatable and not b.is_argument for b in variables)
            if should_print:
                code = '\n'.join(self._print(a) for a in expr.stmt.body if a is not last_assign[0])
                if isinstance(last_assign[0].rhs, Variable) and not last_assign[0].rhs.is_argument:
                    self._additional_declare.append(last_assign[0].rhs)
                return code + '\nreturn {};'.format(self._print(last_assign[0].rhs))
            else:
                code = '\n'+self._print(expr.stmt)
                self._additional_declare.append(last_assign[0].lhs)
                return code + '\nreturn {};'.format(self._print(last_assign[0].lhs))
        if isinstance(args[0], Variable) and not args[0].is_argument:
            self._additional_declare.append(args[0])
        return 'return {0};'.format(self._print(args[0]))
```
So i worked with `_shouldnt_declare`🙂. unless you have something better.

EmilyBourne:
@mostafaMamoni 
My suggestion:
In `_visit_FunctionDef`
```
       if len(expr.results) <= 1 :
            for r in expr.results:
                if isinstance(r, Variable) and not r.is_temp:
                    decs += [Declare(r.dtype, r)]
                elif not isinstance(r, Variable):
                    decs += [FuncAddressDeclare(r)]
        decs  = '\n'.join(self._print(i) for i in decs)
        self._additional_declare.clear()
```
in `_visit_Return`
```
if expr.stmt:
    # get Assign nodes form the CodeBlock object expr.stmt.
    assigns = expr.stmt.get_attribute_nodes(Assign)

    # make sure that stmt contains at least one assign node.
    assert(len(assigns)>=1)
    last_assign = assigns[-1]
    variables = last_assign.rhs.get_attribute_nodes(Variable, excluded_nodes=(FunctionDef,))
    unneeded_var = not any(v.allocatable and not v.is_argument for v in variables)
    if unneeded_var:
        code = '\n'.join(self._print(a) for a in expr.stmt.body if a is not last_assign[0])
        return code + '\nreturn {};'.format(self._print(last_assign[0].rhs))
    else:
        self._additional_declare.append(expr.expr)
        code = '\n'+self._print(expr.stmt)

return 'return {0};'.format(self._print(args[0]))
```

mostafaMamoni:
> @mostafaMamoni
> My suggestion:
> In `_visit_FunctionDef`
> 
> ```
>        if len(expr.results) <= 1 :
>             for r in expr.results:
>                 if isinstance(r, Variable) and not r.is_temp:
>                     decs += [Declare(r.dtype, r)]
>                 elif not isinstance(r, Variable):
>                     decs += [FuncAddressDeclare(r)]
>         decs += [Declare(i.dtype, i) for i in self._additional_declare]
>         decs  = '\n'.join(self._print(i) for i in decs)
>         self._additional_declare.clear()
> ```
> 
> in `_visit_Return`
> 
> ```
> if expr.stmt:
>     # get Assign nodes form the CodeBlock object expr.stmt.
>     assigns = expr.stmt.get_attribute_nodes(Assign)
> 
>     # make sure that stmt contains at least one assign node.
>     assert(len(assigns)>=1)
>     last_assign = assigns[-1]
>     variables = last_assign.rhs.get_attribute_nodes(Variable, excluded_nodes=(FunctionDef,))
>     unneeded_var = not any(v.allocatable and not v.is_argument for v in variables)
>     if unneeded_var:
>         code = '\n'.join(self._print(a) for a in expr.stmt.body if a is not last_assign[0])
>         return code + '\nreturn {};'.format(self._print(last_assign[0].rhs))
>     else:
>         self._additional_declare.append(expr.expr)
>         code = '\n'+self._print(expr.stmt)
> 
> return 'return {0};'.format(self._print(args[0]))
> ```

if you will use `is_temp` in FunctionDef, why you didn’t use it in `print_return`, in addition, and i think what Said said is every `Variable is created by Pyccel` should have `is_temp=True`, So `is_temp` is no longer helpful in the printer since we will have the dummy variables have `is_temp=True`.

mostafaMamoni:
> > @mostafaMamoni
> > My suggestion:
> > In `_visit_FunctionDef`
> > ```
> >        if len(expr.results) <= 1 :
> >             for r in expr.results:
> >                 if isinstance(r, Variable) and not r.is_temp:
> >                     decs += [Declare(r.dtype, r)]
> >                 elif not isinstance(r, Variable):
> >                     decs += [FuncAddressDeclare(r)]
> >         decs += [Declare(i.dtype, i) for i in self._additional_declare]
> >         decs  = '\n'.join(self._print(i) for i in decs)
> >         self._additional_declare.clear()
> > ```
> > 
> > 
> > in `_visit_Return`
> > ```
> > if expr.stmt:
> >     # get Assign nodes form the CodeBlock object expr.stmt.
> >     assigns = expr.stmt.get_attribute_nodes(Assign)
> > 
> >     # make sure that stmt contains at least one assign node.
> >     assert(len(assigns)>=1)
> >     last_assign = assigns[-1]
> >     variables = last_assign.rhs.get_attribute_nodes(Variable, excluded_nodes=(FunctionDef,))
> >     unneeded_var = not any(v.allocatable and not v.is_argument for v in variables)
> >     if unneeded_var:
> >         code = '\n'.join(self._print(a) for a in expr.stmt.body if a is not last_assign[0])
> >         return code + '\nreturn {};'.format(self._print(last_assign[0].rhs))
> >     else:
> >         self._additional_declare.append(expr.expr)
> >         code = '\n'+self._print(expr.stmt)
> > 
> > return 'return {0};'.format(self._print(args[0]))
> > ```
> 
> if you will use `is_temp` in `print_FunctionDef`, why you didn’t use it in `print_return`, in addition, and i think what Said said is every `Variable is created by Pyccel` should have `is_temp=True`, So `is_temp` is no longer helpful in the printer since we will have the dummy variables have `is_temp=True`.

EmilyBourne:
@mostafaMamoni 
> 
> if you will use `is_temp` in `print_FunctionDef`, why you didn’t use it in `print_return`, in addition, and i think what Said said is every `Variable is created by Pyccel` should have `is_temp=True`, So `is_temp` is no longer helpful in the printer since we will have the dummy variables have `is_temp=True`.

Beware in the code above I am using Said's definition of `is_temp`
I never declare the Variable in Return.expr if there is a Return.stmt. If the Variable needs declaring then this is done by _additional_declarations

EmilyBourne:
@mostafaMamoni
Without using `is_temp` at all my suggestion would be:
In _visit_FunctionDef
```
        decs  = [Declare(i.dtype, i) if isinstance(i, Variable) else FuncAddressDeclare(i) for i in expr.local_vars]
```
in _visit_Return
```
if expr.stmt:
    # get Assign nodes form the CodeBlock object expr.stmt.
    assigns = expr.stmt.get_attribute_nodes(Assign)

    # make sure that stmt contains at least one assign node.
    assert(len(assigns)>=1)
    last_assign = assigns[-1]
    variables = last_assign.rhs.get_attribute_nodes(Variable, excluded_nodes=(FunctionDef,))
    unneeded_var = not any(v.allocatable and not v.is_argument for v in variables)
    if unneeded_var:
        code = '\n'.join(self._print(a) for a in expr.stmt.body if a is not last_assign[0])
        return code + '\nreturn {};'.format(self._print(last_assign[0].rhs))
    else:
        code = '\n'+self._print(expr.stmt)

self._additional_declare.append(args[0])
return 'return {0};'.format(self._print(args[0]))
```

mostafaMamoni:
> @mostafaMamoni
> Without using `is_temp` at all my suggestion would be:
> In _visit_FunctionDef
> 
> ```
>         decs  = [Declare(i.dtype, i) if isinstance(i, Variable) else FuncAddressDeclare(i) for i in expr.local_vars]
>         decs += [Declare(i.dtype, i) for i in self._additional_declare]
> ```
> 
> in _visit_Return
> 
> ```
> if expr.stmt:
>     # get Assign nodes form the CodeBlock object expr.stmt.
>     assigns = expr.stmt.get_attribute_nodes(Assign)
> 
>     # make sure that stmt contains at least one assign node.
>     assert(len(assigns)>=1)
>     last_assign = assigns[-1]
>     variables = last_assign.rhs.get_attribute_nodes(Variable, excluded_nodes=(FunctionDef,))
>     unneeded_var = not any(v.allocatable and not v.is_argument for v in variables)
>     if unneeded_var:
>         code = '\n'.join(self._print(a) for a in expr.stmt.body if a is not last_assign[0])
>         return code + '\nreturn {};'.format(self._print(last_assign[0].rhs))
>     else:
>         code = '\n'+self._print(expr.stmt)
> 
> self._additional_declare.append(args[0])
> return 'return {0};'.format(self._print(args[0]))
> ```

i think you missed `self._additional_declare.append(last_assign[0].lhs)` here:
```python
    if unneeded_var:
        code = '\n'.join(self._print(a) for a in expr.stmt.body if a is not last_assign[0])
        return code + '\nreturn {};'.format(self._print(last_assign[0].rhs))
```
and `args[0]` could come as a `FuncAddressDeclare` object, So it needs `FuncAddressDeclare()` `print_FunctionDef`, i don't know where exactly i need to set `is_temp` to `True` in semantic.py, because it comes with wrong value in user's variables, by the way you meant `_print_` not `_visit_`.

EmilyBourne:
@mostafaMamoni
> i think you missed `self._additional_declare.append(last_assign[0].lhs)` here:
> 
> ```python
>     if unneeded_var:
>         code = '\n'.join(self._print(a) for a in expr.stmt.body if a is not last_assign[0])
>         return code + '\nreturn {};'.format(self._print(last_assign[0].rhs))
> ```

No. `self._additional_declare` is for things that need to be declared. Here we print `return rhs` so we don't need to declare the `lhs`

> and `args[0]` could come as a `FuncAddressDeclare` object, So it needs `FuncAddressDeclare()` `print_FunctionDef`

Ah you are right.

> i don't know where exactly i need to set `is_temp` to `True` in semantic.py, because it comes with wrong value in user's variables,

The symbols for Variables such as `Out_0001` are created in the syntactic stage by a call to `create_variable`. `create_variable` creates temporary symbols (it needs renaming one day) so the easiest thing would be to set the `is_temp` property [here](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/core.py#L251)

> by the way you meant `_print_` not `_visit_`.

Oops. You are right

pyccel/codegen/printing/pycode.py
```diff
@@ -124,13 +124,20 @@ def _print_FunctionDef(self, expr):
...
-        if expr.expr:
-            ret = ','.join([self._print(i) for i in expr.expr])
-            code += 'return {}'.format(ret)
-        return code
+            for i in expr.stmt.body:
+                rhs_list.append(i.rhs)
+                lhs_list.append(i.lhs)
+        for a in expr.expr:
+            if lhs_list and a not in lhs_list:
+                expr_return_vars.append(a)
```
EmilyBourne, line 136: Can you give an example of where this block is needed?

EmilyBourne (changes requested):

pyccel/ast/core.py
```diff
@@ -248,7 +248,9 @@ def create_variable(forbidden_names, prefix = None, counter = 1):
 
     name, counter = create_incremented_string(forbidden_names, prefix, counter = counter)
 
-    return PyccelSymbol(name), counter
+    ps = PyccelSymbol(name)
+    ps.is_temp = True
```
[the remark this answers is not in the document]
mostafaMamoni: i tried many things, but str class blocked me, here is what i tried:
```python
def __init__(self, is_temp=False)
      self._is_temp = is_temp
      super().__init__(self)
```

```python
def __init__(self, is_temp=False)
      self._is_temp = is_temp
     super().__init__(self)
```

```python
def __init__(self, is_temp=False)
      self._is_temp = is_temp
     super().__init__(self)
```

```python
def __init__(self, name, is_temp=False)
      self._is_temp = is_temp
     super().__init__(name)
```
i don't clearly remember the message error, but it was something like `is_temp is a bad argument for str`

pyccel/ast/variable.py
```diff
@@ -286,6 +290,19 @@ def is_pointer(self, is_pointer):
...
+        """
+        Indicates if the variable is created by Pyccel and it is not necessary to print [Default value: False]
+        """
+        return self._is_temp
+
+    @is_temp.setter
+    def is_temp(self, is_temp):
+        if not isinstance(is_temp, bool):
+            raise TypeError('is_temp must be a boolean.')
+        self._is_temp = is_temp
```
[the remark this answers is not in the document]
mostafaMamoni: it is just a setter so isn’t necessary right now, but i think it is good to keep it, to make it easier for manipulation.

pyccel/codegen/printing/pycode.py
mostafaMamoni: we don't want to print Dummy variables from `expr.expr`, in all cases `expr.expr` will contain either dummy variables or user's variables, that's why i collect unneeded dummy variables from `expr.stmt` if they exist and avoid them is this `if` block, the other variables that come from `rhs`s 
in`expr.stmt`  I'm printing them in the return.

pyccel/ast/core.py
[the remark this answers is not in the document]
EmilyBourne: That sounds like you are passing the wrong arguments to `__new__` it is inherited from `str` so if you didn't redefine it too then the arguments won't match

pyccel/ast/variable.py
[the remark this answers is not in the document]
EmilyBourne: Ideally this property should be set at creation and should never change so if we can manage without a setter I think we should remove it. We can always put it back if we find we need it in another PR

pyccel/codegen/printing/pycode.py
EmilyBourne: But can we ever have variables in `expr.expr` which don't exist in `expr.stmt` if there is an `expr.stmt`?

pyccel/codegen/printing/ccode.py
```diff
@@ -1201,13 +1205,32 @@ def _print_Constant(self, expr):
...
+            # Check the Assign objects list in case of
+            # the user assigns a variable to an object contains IndexedElement object.
+            if not last_assign:
+                return 'return {0};'.format(self._print(args[0]))
+
+            # make sure that stmt contains one assign node.
+            assert(len(last_assign)==1)
+            variables = last_assign[0].rhs.get_attribute_nodes(Variable, excluded_nodes=(FunctionDef,))
+            should_print = not any(b.allocatable and not b.is_argument for b in variables)
+            if should_print:
```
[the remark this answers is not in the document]
mostafaMamoni: you are right.

pyccel/ast/core.py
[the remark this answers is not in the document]
EmilyBourne: Try:
```
def __new__(cls, name, is_temp=False)
     super().__new__(cls, name)

def __init__(self, name, is_temp=False)
      self._is_temp = is_temp
     super().__init__()
```

pyccel/codegen/printing/pycode.py
mostafaMamoni: yes, if you have them as arguments

EmilyBourne: Ah ok :+1:

pyccel/codegen/printing/pycode.py
```diff
@@ -124,13 +124,22 @@ def _print_FunctionDef(self, expr):
         return code
 
     def _print_Return(self, expr):
-        code = ''
+        rhs_list = []
+        lhs_list = []
+        expr_return_vars = []
+        print("stmt == ", expr.stmt)
+        print("expr == ", expr.expr)
```
EmilyBourne, line 131: This shouldn't have been committed ;)

mostafaMamoni: 😂

pyccel/ast/internals.py
```diff
@@ -182,6 +182,25 @@ class PyccelSymbol(str, Immutable):
...
+    @property
+    def is_temp(self):
+        """
+        Indicates if the variable created by Pyccel [default value: False]
+        """
+        return self._is_temp
+
+    @is_temp.setter
+    def is_temp(self, is_temp):
+        self._is_temp = is_temp
```
EmilyBourne, line 203: As with `Variable` can we avoid the setter? This is more important for `PyccelSymbol` as it should be immutable

pyccel/codegen/printing/pycode.py
```diff
@@ -124,13 +124,20 @@ def _print_FunctionDef(self, expr):
...
-            code += 'return {}'.format(ret)
-        return code
+            for i in expr.stmt.body:
+                rhs_list.append(i.rhs)
+                lhs_list.append(i.lhs)
+        for a in expr.expr:
+            if lhs_list and a not in lhs_list:
+                expr_return_vars.append(a)
+            elif not lhs_list:
+                expr_return_vars.append(a)
```
EmilyBourne, line 138: Can this be reduced to a list comprehension (list comprehensions run a lot faster than for loops):
```
expr_return_vars = [a if not lhs_list or a not in lhs_list for a in expr.expr]
```
?

EmilyBourne (approved):

saidctb (approved):

pyccel/ast/internals.py
```diff
@@ -182,6 +182,21 @@ class PyccelSymbol(str, Immutable):
...
     x
     """
+    _is_temp = False
+
+    def __new__(cls, name, is_temp=False):
+        return super().__new__(cls, name)
+
+    def __init__(self, name, is_temp=False):
+        self._is_temp = is_temp
+        super().__init__()
```
yguclu, line 192: Why do we need a class attribute `_is_temp` if we assign the instance attribute `_is_temp`  (which effectively overwrites the class attribute) in the `__init__`?

pyccel/ast/internals.py
```diff
@@ -182,6 +182,21 @@ class PyccelSymbol(str, Immutable):
...
+
+    @property
+    def is_temp(self):
+        """
+        Indicates if the variable created by Pyccel [default value: False]
+        """
+        return self._is_temp
```
yguclu, line 199: Please modify this docstring:

- Add `was` after "variable"
- Change "variable" with `symbol`
- Reconsider the meaning of this whole sentence: every symbol is created by Pyccel!
What did you really mean here? Maybe this?
```
Indicates if this symbol represents a temporary variable created by Pyccel,
and was not present in the original Python code.
```

pyccel/ast/variable.py
```diff
@@ -85,6 +85,8 @@ class base if variable is an object or an object member [Default value: None]
 
     is_const: bool
         if object is a const argument of a function [Default value: False]
+    is_temp: bool
+        Indicates if the variable is created by Pyccel and it is not necessary to print [Default value: False]
```
yguclu, line 89: See my previous comment.

Further, please add empty line before `is_temp: bool` to match formatting.

pyccel/ast/variable.py
```diff
@@ -286,6 +290,13 @@ def is_pointer(self, is_pointer):
             raise TypeError('is_pointer must be a boolean.')
         self._is_pointer = is_pointer
 
+    @property
+    def is_temp(self):
+        """
+        Indicates if the variable is created by Pyccel and it is not necessary to print [Default value: False]
+        """
```
yguclu, line 297: See my previous comment.

pyccel/codegen/printing/pycode.py
```diff
@@ -124,13 +124,15 @@ def _print_FunctionDef(self, expr):
...
-            code += self._print(expr.stmt)+'\n'
-        if expr.expr:
-            ret = ','.join([self._print(i) for i in expr.expr])
-            code += 'return {}'.format(ret)
-        return code
+            rhs_list = [i.rhs for i in expr.stmt.body]
+            lhs_list = [i.lhs for i in expr.stmt.body]
+        expr_return_vars = [a for a in expr.expr if not lhs_list or a not in lhs_list]
+
+        return 'return ' + ','.join(self._print(i) for i in expr_return_vars + rhs_list)
```
yguclu, line 135: I find the structure of these few lines of code surprisingly confusing. For example, you set `expr_return_vars = []`, and then you always overwrite it. I suggest the following:

```python

    rhs_list = [i.rhs for i in expr.stmt.body] if expr.stmt else []
    lhs_list = [i.lhs for i in expr.stmt.body] if expr.stmt else []
    expr_return_vars = [a for a in expr.expr if a not in lhs_list]
    
    return 'return ' + ','.join(self._print(i) for i in expr_return_vars + rhs_list)

yguclu (changes requested):

pyccel/ast/internals.py
mostafaMamoni: because PyccelSymbol inherits from `str`  which has no argument called '_is_temp`,` __new__() got an unexpected keyword argument 'is_temp'`  will be raised, that's why we need to redefine `__new__` in PyccelSymbol, so that we can use `is_temp` in PyccelSymbol.

mostafaMamoni: because PyccelSymbol inherits from str which has no argument called '_is_temp, `new() got an unexpected keyword argument 'is_temp'` will be raised, that's why we need to redefine `__new__` in PyccelSymbol, so that we can use `is_temp` in PyccelSymbol.

pyccel/ast/internals.py
mostafaMamoni: yes, exactly.

pyccel/ast/internals.py
EmilyBourne: A class attribute is an attribute which can be called like: `PyccelSymbol._is_temp` without an instance of the class. This is what you have defined at line 185

EmilyBourne: An instance attribute is an attribute which is attached to `self`, i.e. the one defined at line 191

yguclu: @mostafaMamoni I am specifically referring to lines 185 and 191, as explained in @EmilyBourne's comments.

I think that you can just remove line 185.

yguclu (approved):

Issue #735, opened 2021-02-09
Pyccel can only return native python types for scalars


EmilyBourne:
**Describe the improvement**
Currently pyccel always returns scalar values by constructing them using `Py_BuildValue` and the native python flag. This means that the result is a native python type. We should be able to return types with a given precision as we are able to receive them as arguments. This is especially important for the type `int64` on a windows machine as this will be truncated to an `int32` if it is returned in a native `int`. Numpy types can be constructed using the function [PyArray_Scalar](https://numpy.org/doc/stable/reference/c-api/array.html?highlight=pyarray_scalar#c.PyArray_Scalar)

**Test Code**
Provide code which does not currently work but which should do when this issue is fixed:
```python
    @types('real')
    def fabs_call_r(x):
        from numpy import fabs
        return fabs(x)

    f1 = epyccel(fabs_call_r, language = language)
    x = uniform(high=1e6)
    assert(type(f1(x)) == type(fabs_call_r(x)))

```

**Language**
If appropriate, please specify which language the python code is translated to (Fortran by default)

**Additional Context**
This is the equivalent of issue #734 for return values

EmilyBourne:
As a first step this issue should be fixed for all types whose precision differs from the default precision. To return the correct type when the precision is equal to the default precision (e.g. `float64` instead of `float`) additional changes would need to be made in pyccel to allow the differentiation of these 2 cases

PR #755, opened 2021-02-14
Fix casting boolean in numpy float/complex

mostafaMamoni:
Passing a boolean to numpy.float/complex caused a compilation error of the generated Fortran code. This was fixed by casting `arg` arguments using the casting function `PythonInt` in both functions in case of `arg.dtype` is a `NativeBool` object.

**Commit summary**

- Cast `arg0` or `arg1` in `PythonComplex` constructor if they are boolean;
- Cast `arg` in `PythonFloat` if it is a boolean;
- Add tests.


.travis.yml
```diff
@@ -33,7 +33,7 @@ before_script:  # Coverage instructions - setup
   - export COVERAGE_PROCESS_START=$(pwd)/.coveragerc
 
 script:
-  - cd tests/epyccel/recognised_functions
+  - cd tests
```
EmilyBourne: travis is no longer used for testing. To see your problem faster I suggest modifying [this line](https://github.com/pyccel/pyccel/blob/master/.github/workflows/Github_pytest.yml#L59)
to:
```
python -m pytest -rx -m "not parallel and c" epyccel/recognised_functions -k test_numpy_complex -x
```

EmilyBourne:
PR #759 has just been merged. There is a chance that this fixes the problems you are having here with 64 bit integers. Can you fix the merge conflicts please

mostafaMamoni:
> PR #759 has just been merged. There is a chance that this fixes the problems you are having here with 64 bit integers. Can you fix the merge conflicts please

thank you.

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3877,3 +3871,858 @@ def test_complex128():
...
+    import numpy as np
+
+    if sys.platform == 'win32':
+        integer = randint(-2147483648, 2147483647)
+    else:
+        integer = randint(-9223372036854775808, 9223372036854775807)
+    integer8 = np.int8(randint(-128, 127))
+    integer16 = np.int16(randint(-32768, 32767))
+    integer32 = np.int32(randint(-2147483648, 2147483647))
+    integer64 = np.int64(randint(-9223372036854775808, 9223372036854775807))
```
EmilyBourne, line 4034: I think you need to pass the dtype to `randint` to ensure that it creates integers of the correct type. By casting you are doing:
```
tmp = randint(-9223372036854775808, 9223372036854775807, dtype=int) #int32 on windows so overflow/underflow error
integer64 = np.int64(tmp)
```

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3877,3 +3871,858 @@ def test_complex128():
...
+    fl = np.float(randint(-2147483648, 2147483647))
+    fl32 = np.float32(randint(-2147483648, 2147483647))
+    fl64 = np.float64(randint(-9223372036854775808, 9223372036854775807))
```
EmilyBourne, line 4037: Why are you using integers to test float? Why not use `rand` instead of `randint`?

tests/epyccel/recognised_functions/test_numpy_funcs.py
mostafaMamoni: that's will be changed as in #763 tests

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3877,3 +3887,849 @@ def test_complex128():
...
+        b = np.int(a)
+        return b
+
+    import numpy as np
+
+    integer8 = randint(np.iinfo(np.int8(1)).min, np.iinfo(np.int8(1)).max, size=(5), dtype=np.int8)
+    integer16 = randint(np.iinfo(np.int16(1)).min, np.iinfo(np.int16(1)).max, size=(5), dtype=np.int16)
+    integer = randint(min_int, max_int, size=(5), dtype=np.int)
+    integer32 = randint(min_int32, max_int32, size=(5), dtype=np.int32)
+    integer64 = randint(min_int64, max_int64, size=(5), dtype=np.int64)
```
EmilyBourne, line 3952: Why are you creating arrays here?

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3877,3 +3887,849 @@ def test_complex128():
...
+        b = np.int32(a)
+        return b
+
+    import numpy as np
+
+    integer8 = randint(np.iinfo(np.int8(1)).min, np.iinfo(np.int8(1)).max, size=(5), dtype=np.int8)
+    integer16 = randint(np.iinfo(np.int16(1)).min, np.iinfo(np.int16(1)).max, size=(5), dtype=np.int16)
+    integer = randint(min_int, max_int, size=(5), dtype=np.int)
+    integer32 = randint(min_int32, max_int32, size=(5), dtype=np.int32)
+    integer64 = randint(min_int64, max_int64, size=(5), dtype=np.int64)
```
EmilyBourne, line 4046: Why are you creating arrays here?

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3877,3 +3887,849 @@ def test_complex128():
...
+
+    fl = uniform(min_float / 2, max_float / 2, size=(5))
+    fl32 = uniform(min_float32, max_float32, size=(5))
```
EmilyBourne, line 4049: Why are you creating arrays here?

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3877,3 +3887,849 @@ def test_complex128():
...
+    fl32 = np.float32(fl32)
+    fl64 = uniform(min_float64 / 2, max_float64 / 2, size=(5))
```
EmilyBourne, line 4051: Why are you creating arrays here?

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3877,3 +3887,849 @@ def test_complex128():
...
```
EmilyBourne, line 4145: Why are you creating arrays here?

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3877,3 +3887,849 @@ def test_complex128():
...
```
EmilyBourne, line 4240: Why are you creating arrays here?

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3877,3 +3887,849 @@ def test_complex128():
...
```
EmilyBourne, line 4334: Why are you creating arrays here?

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3877,3 +3887,849 @@ def test_complex128():
...
```
EmilyBourne, line 4428: Why are you creating arrays here?

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3877,3 +3887,849 @@ def test_complex128():
...
```
EmilyBourne, line 4522: Why are you creating arrays here?

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3877,3 +3887,849 @@ def test_complex128():
...
```
EmilyBourne, line 4616: Why are you creating arrays here?

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3877,3 +3887,849 @@ def test_complex128():
...
```
EmilyBourne, line 4710: Why are you creating arrays here?

tests/epyccel/recognised_functions/test_numpy_funcs.py
mostafaMamoni: because we need to pass an array filled with random values with the right data type to imag function, if i did it inside the functions `iinfo` `finfo` and `uniform` will not be recognized by epyccel.

EmilyBourne: Wrong PR? This PR is not about the `imag` function

These arguments are passed to functions such as:
```
@types('int8')
def test_int8_int(a):
    import numpy as np
    b = np.int(a)
    return b
```
which take scalar arguments

pyccel/codegen/printing/ccode.py
```diff
@@ -1239,6 +1239,8 @@ def _print_Nil(self, expr):
         return 'NULL'
 
     def _print_PyccelAdd(self, expr):
+        if isinstance(expr.args[0].dtype, NativeInteger):
+            self._additional_imports.add('stdint')
```
EmilyBourne, line 1243: Why do you need `stdint` to do addition? I would have expected you to need it for casting

mostafaMamoni: yes, i mentioned it in a previous meeting.

EmilyBourne:
The remaining errors seem to be being caused by issue #735. The result is being stored in the wrong type which doesn't have enough precision. Please do one of the following:
1. Mark the test as xfail for windows
2. Comment the int64 test and leave a comment referencing issue #735 
3. Reduce the size of the arguments passed to the test so that the result is small enough to fit in an int32

I would prefer option 1 or 2

mostafaMamoni:
> The remaining errors seem to be being caused by issue #735. The result is being stored in the wrong type which doesn't have enough precision. Please do one of the following:
> 
> 1. Mark the test as xfail for windows
> 2. Comment the int64 test and leave a comment referencing issue #735
> 3. Reduce the size of the arguments passed to the test so that the result is small enough to fit in an int32
> 
> I would prefer option 1 or 2

i would like to comment the tests and referencing issue #735.

EmilyBourne (approved):

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3877,3 +3887,853 @@ def test_complex128():
...
+    @types('float64')
+    def test_float64_complex128(a):
+        import numpy as np
+        b = np.complex128(a)
+        return b
+
+    import numpy as np
+
+    integer8 = randint(np.iinfo(np.int8(1)).min, np.iinfo(np.int8(1)).max, dtype=np.int8)
+    integer16 = randint(np.iinfo(np.int16(1)).min, np.iinfo(np.int16(1)).max, dtype=np.int16)
```
nhamidn, line 4706: integer8 = randint(np.iinfo('int8').min, np.iinfo('int8').max, dtype=np.int8)
integer16 = randint(np.iinfo('int16').min, np.iinfo('int16').max, dtype=np.int16)

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3877,3 +3887,853 @@ def test_complex128():
...
+    @types('float64')
+    def test_float64_complex64(a):
+        import numpy as np
+        b = np.complex64(a)
+        return b
+
+    import numpy as np
+
+    integer8 = randint(np.iinfo(np.int8(1)).min, np.iinfo(np.int8(1)).max, dtype=np.int8)
+    integer16 = randint(np.iinfo(np.int16(1)).min, np.iinfo(np.int16(1)).max, dtype=np.int16)
```
nhamidn, line 4612: same here

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3877,3 +3887,853 @@ def test_complex128():
...
+    @types('float64')
+    def test_float64_double(a):
+        import numpy as np
+        b = np.double(a)
+        return b
+
+    import numpy as np
+
+    integer8 = randint(np.iinfo(np.int8(1)).min, np.iinfo(np.int8(1)).max, dtype=np.int8)
+    integer16 = randint(np.iinfo(np.int16(1)).min, np.iinfo(np.int16(1)).max, dtype=np.int16)
```
nhamidn, line 4518: same here

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3877,3 +3887,853 @@ def test_complex128():
...
+    @types('float64')
+    def test_float64_float64(a):
+        import numpy as np
+        b = np.float64(a)
+        return b
+
+    import numpy as np
+
+    integer8 = randint(np.iinfo(np.int8(1)).min, np.iinfo(np.int8(1)).max, dtype=np.int8)
+    integer16 = randint(np.iinfo(np.int16(1)).min, np.iinfo(np.int16(1)).max, dtype=np.int16)
```
nhamidn, line 4424: same here

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3877,3 +3887,853 @@ def test_complex128():
...
+    @types('float64')
+    def test_float64_float32(a):
+        import numpy as np
+        b = np.float32(a)
+        return b
+
+    import numpy as np
+
+    integer8 = randint(np.iinfo(np.int8(1)).min, np.iinfo(np.int8(1)).max, dtype=np.int8)
+    integer16 = randint(np.iinfo(np.int16(1)).min, np.iinfo(np.int16(1)).max, dtype=np.int16)
```
nhamidn, line 4330: same here

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3877,3 +3887,853 @@ def test_complex128():
...
+    @types('float64')
+    def test_float64_float(a):
+        import numpy as np
+        b = np.float(a)
+        return b
+
+    import numpy as np
+
+    integer8 = randint(np.iinfo(np.int8(1)).min, np.iinfo(np.int8(1)).max, dtype=np.int8)
+    integer16 = randint(np.iinfo(np.int16(1)).min, np.iinfo(np.int16(1)).max, dtype=np.int16)
```
nhamidn, line 4236: same here and all the others

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3877,3 +3887,853 @@ def test_complex128():
...
+    @types('float64')
+    def test_float64_int(a):
+        import numpy as np
+        b = np.int(a)
+        return b
+
+    import numpy as np
+
+    integer8 = randint(np.iinfo(np.int8(1)).min, np.iinfo(np.int8(1)).max, dtype=np.int8)
+    integer16 = randint(np.iinfo(np.int16(1)).min, np.iinfo(np.int16(1)).max, dtype=np.int16)
```
nhamidn, line 3949: same here

nhamidn (changes requested):

nhamidn (approved):

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3877,3 +3894,835 @@ def test_complex128():
...
+    def test_float32_int(a):
+        from numpy import int as NumpyInt
+        b = NumpyInt(a)
+        return b
+
+    @types('float64')
+    def test_float64_int(a):
+        from numpy import int as NumpyInt
+        b = NumpyInt(a)
+        return b
```
EmilyBourne, line 3951: As discussed, could you simplify all these functions into 1 interface

mostafaMamoni: yes im working on it.

nhamidn (approved):

EmilyBourne (approved):

pyccel/ast/builtins.py
```diff
@@ -260,7 +260,9 @@ class PythonFloat(PyccelAstNode):
     _attribute_nodes = ('_arg',)
 
     def __new__(cls, arg):
-        if isinstance(arg, LiteralFloat):
+        if isinstance(arg.dtype, NativeBool):
+            return PythonInt(arg)
+        elif isinstance(arg, LiteralFloat):
             return LiteralFloat(arg, precision = cls._precision)
```
yguclu, line 266: This may not be totally relevant to this PR, but it's quite close. Given that a `LiteralFloat` object is essentially immutable, can't we simply return the argument in this case? That is
```python
elif isinstance(arg, LiteralFloat):
    return arg
```

EmilyBourne: You need be sure that you have the right precision. So this would be:
```
elif isinstance(arg, LiteralFloat) and arg.precision == cls._precision:
    return arg
elif isinstance(arg, LiteralFloat):
    return LiteralFloat(arg, precision = cls._precision)
```

yguclu: But the precision is in the class and it is fixed, or am I missing something?

EmilyBourne: The precision of the result is fixed, but that doesn't mean that it matches the precision of the argument.
Imagine if you do:
```
x = np.float32(3)
y = float(x)
```
With your code y will still be a `float32` instead of a `float64`

yguclu: I thought that `x` would be a `Variable` in your example, not a `LiteralFloat`.

EmilyBourne: Ah you are right, sorry the example would have to be:
```
y = float(np.float32(3))
```

yguclu: OK, I see the problem now. @mostafaMamoni you can write this:
```python
elif isinstance(arg, LiteralFloat):
    prec = cls._precision
    return arg if arg.precision == prec else LiteralFloat(arg, precision = prec)
```

yguclu: By the way, why do we pass `arg` to `LiteralFloat` instead of `arg.p`?

EmilyBourne: `arg.p` is a sympy relic that only ever worked for integers and is removed in PR #764.
We could use `arg.python_value` though as this is what `LiteralFloat` will do in the `__init__` method anyway

yguclu: >`arg.p` is a sympy relic that only ever worked for integers and is removed in PR #764.

Nice!

I don't want to cause confusion. Whatever seems more appropriate for this PR.

mostafaMamoni: do you mean `arg.python_value` ?. you can find `p` attribute in `LiteralInteger` only.

yguclu: OK, use `python_value` then:
```python
elif isinstance(arg, LiteralFloat):
    ap = arg.precision
    cp = cls._precision
    return arg if ap == cp else LiteralFloat(arg.python_value, precision = cp)
```

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -2,12 +2,35 @@
...
+min_int = iinfo('int').min
+max_int = iinfo('int').max
+
+min_int32 = iinfo('int32').min
+max_int32 = iinfo('int32').max
+
+min_int64 = iinfo('int64').min
+max_int64 = iinfo('int64').max
+
+min_float = sys.float_info.min
```
yguclu, line 26: This does not work?
```python
min_float = finfo('float').min
```

pyccel/ast/builtins.py
```diff
@@ -260,7 +260,9 @@ class PythonFloat(PyccelAstNode):
...
         elif isinstance(arg, LiteralInteger):
             return LiteralFloat(arg.p, precision = cls._precision)
```
yguclu, line 268: Can we use `arg.python_value` here, or it does not exist for an integer?

tests/epyccel/recognised_functions/test_numpy_funcs.py
mostafaMamoni: it fails in windows because `finfo` is Numpy's function it returns the minimum value of a float64 which unavoidable in windows, that's why we need that value from the system, not from a Library.

yguclu: Why does `max_float = finfo('float').max` work instead?

pyccel/ast/builtins.py
yguclu: Sorry, I see that this will be changed in PR #764. Never mind.

EmilyBourne: We can but I think it is beyond the scope of this PR as it is fixed in #764 anyway

pyccel/ast/builtins.py
yguclu: @EmilyBourne Should we leave this to PR #764 instead?

EmilyBourne: Yes i think it makes more sense to have this discussion there

tests/epyccel/recognised_functions/test_numpy_funcs.py
mostafaMamoni: exactly that was strange for me too, i will give `finfo('float').min` another try and see what's exactly the problem.

EmilyBourne: On my computer I get:
- `finfo('float').min` : -1.7976931348623157e+308
- `sys.float_info.min` : 2.2250738585072014e-308

So it seems that `finfo('float').min` is the smallest value, while `sys.float_info.min` is the smallest absolute value.

Currently tests are failing on all systems (so this doesn't seem to be a windows problem) for tests which test `np.log`. This makes sense as `np.log` requires a positive argument

mostafaMamoni: sorry i was wrong about `finfo('float').min`.  in general `numpu.log` should be tested only with positive scalars other than that it will give an `undefined behaviour` like `<stdin>:1: RuntimeWarning: invalid value encountered in log` and `1` or `nan` as a results.

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -547,7 +564,7 @@ def log_call_i(x):
         return log(x)
 
     f1 = epyccel(log_call_i, language = language)
-    x = randint(low=min_float, high=1e6)
+    x = randint(low=0, high=1e6)
     assert(isclose(f1(x), log_call_i(x), rtol=RTOL, atol=ATOL))
```
yguclu, line 568: For the logarithm `0` is not an acceptable argument, so you should start from `1` in this case.

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -558,7 +575,7 @@ def log_call_r(x):
         return log(x)
 
     f1 = epyccel(log_call_r, language = language)
-    x = uniform(low=min_float, high=1e6)
+    x = uniform(low=0, high=max_float)
```
yguclu, line 578: For the logarithm 0 is not an acceptable argument, so you should choose a very small positive lower bound. For example abs(min_float) should work.

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -570,8 +587,8 @@ def log_phrase(x,y):
         return a
 
     f2 = epyccel(log_phrase, language = language)
-    x = uniform(low=min_float, high=1e6)
-    y = uniform(low=min_float, high=1e6)
+    x = uniform(low=0, high=1e6)
+    y = uniform(low=0, high=1e6)
```
yguclu, line 591: Same comment as above

yguclu (changes requested):
Thanks @mostafaMamoni, the tests are very extensive! I have only a couple of comments.

tests/epyccel/recognised_functions/test_numpy_funcs.py
yguclu: Thanks for looking into this.

tests/epyccel/recognised_functions/test_numpy_funcs.py
EmilyBourne: Or `sys.float_info.min` ;)

yguclu (approved):

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -570,8 +587,8 @@ def log_phrase(x,y):
         return a
 
     f2 = epyccel(log_phrase, language = language)
-    x = uniform(low=min_float, high=1e6)
-    y = uniform(low=min_float, high=1e6)
+    x = uniform(low=abs(min_float), high=1e6)
+    y = uniform(low=abs(min_float), high=1e6)
```
EmilyBourne, line 591: abs(min_float) is not a good value for this:
```
In [6]: abs(finfo('float').min) == finfo('float').max
Out[6]: True
```

yguclu: Wow :sweat_smile:

mostafaMamoni: ah yes.

yguclu: OK then, we have to use `finfo(float).tiny`

yguclu: @mostafaMamoni Sorry for the confusion.

EmilyBourne (approved):

Issue #734, opened 2021-02-09
64 bit integers are not valid arguments on windows


EmilyBourne:
**Describe the bug**
When numpy arguments are collected in the C-API wrapper they are not collected as PyObjects. Instead we use the basic Python C-API tools to collect them by converting them to the native python type, then back to the correct precision type. This is bad for 3 reasons:

1. There is unnecessary casting between types
2. We cannot collect a signed `int8` (see issue #722)
3. On windows a native `int` is 32 bits so when we try to pass an int which uses more than 32 bits, an error is raised as python is unable to convert it to a native `int`

**To Reproduce**
Provide code to reproduce the behavior (I do not have access to a windows machine to test this, but it should produce the problem):
```python
from pyccel.epyccel import epyccel
import numpy as np

def f(a : 'int64'):
    print(a)
    
f2 = epyccel(f)
f2(np.int64(3,000,000,000))
```

**Expected behavior**
Any object with specified precision should be collected into a `PyObject` and retrieved using `PyArray_ScalarAsCtype` (as is done for Interfaces)

cc: @rakati

Issue #722, opened 2021-02-01
int8 variable does not accept negative numbers


rakati:
**Describe the bug**
the `int8 `variable is treated in the wrapper as unsigned char.

**To Reproduce**
Provide code to reproduce the behavior:
```python
>> from pyccel.decorators import types
>> from pyccel.epyccel import epyccel
>> @type('int8')
      def neg(x):
           return -x
>> c_neg = epyccel(neg, language='c')
>> c_neg(-1)
OverflowError                             Traceback (most recent call last)
<ipython-input-11-cdbc6a3aa4c1> in <module>
----> 1 f_neg(-1)

OverflowError: unsigned byte integer is less than minimum
>> neg(-1)
1
>> c_neg(5)
-5
```
the wrapper function of `neg` function:
```c
PyObject *neg_wrapper(PyObject *self, PyObject *args, PyObject *kwargs)
{
    int8_t x;
    int8_t Out_0001;
    PyObject *result;
    static char *kwlist[] = {
        "x",
        NULL
    };

    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "b", kwlist, &x))
    {
        return NULL;
    }
    Out_0001 = neg(x);
    result = Py_BuildValue("b", Out_0001);
    return result;
}
```
From the wrapper function above we can see that the character `'b'` has been used for parsing and building the `int8 `variable, but returning to the [python C API](https://docs.python.org/3/c-api/arg.html) documentation, the `'b'` is used for parsing  `unsigned char` and for building `char`.

**Language**
Fortran, C

PR #759, opened 2021-02-16
Correctly handle scalar Numpy arguments in C wrapper

MenoIy:
Solve #734 and solve #722:

- Collect function arguments as Python objects;
- Clean code by removing some old functions;
- Add new tests;
- Modify some old tests' arguments after new changes.

### DIFF
--- a/pyccel/codegen/printing/cwrappercode.py
+++ b/pyccel/codegen/printing/cwrappercode.py
@@ -187,7 +187,18 @@ class CWrapperCodePrinter(CCodePrinter):
     # --------------------------------------------------------------------
     # Functions that take care of creating cast or convert type function call :
     # --------------------------------------------------------------------
-    def _get_types_check(self, variable, collect_var, error_check = True):
+    def _get_array_type_check(self, variable, collect_var):
+       "TODO"
+
+        numpy_dtype = self.find_in_numpy_dtype_registry(variable)
+        arg_dtype   = self.find_in_dtype_registry(self._print(variable.dtype), variable.precision)
+
+        check = PyccelNe(FunctionCall(numpy_get_type, [collect_var]), numpy_dtype)
+        error = PyErr_SetString('PyExc_TypeError', '"{} must be {}"'.format(variable, arg_dtype))
+
+        return check, error
+    
+    def _get_scalar_type_check(self, variable, collect_var, error_check = True):
         "TODO"
         
         if not error_check :
@@ -210,12 +221,12 @@ class CWrapperCodePrinter(CCodePrinter):
 
         valued_var_check = PyccelEq(VariableAddress(collect_var), VariableAddress(Py_None))
         optional_var_collect = []
-        
+
         if variable.is_optional:
             optional_var_collect = [Assign(VariableAddress(variable), VariableAddress(tmp_variable))]
             section = IfSection(valued_var_check, [Assign(VariableAddress(variable), Nil())])
 
-        else :
+        else:
             section = IfSection(valued_var_check, [Assign(variable, variable.value)])
 
         return section, optional_var_collect
@@ -248,7 +259,7 @@ class CWrapperCodePrinter(CCodePrinter):
         var = tmp_variable if tmp_variable else variable
         sections = []
 
-        numpy_check, python_check, error = self._get_types_check(variable, collect_var, error_check)
+        numpy_check, python_check, error = self._get_scalar_type_check(variable, collect_var, error_check)
         
         python_collect  = [Assign(var, self.get_collect_function_call(variable, collect_var))]
         numpy_collect   = [FunctionCall(PyArray_ScalarAsCtype, [collect_var, var])]
@@ -369,11 +380,7 @@ class CWrapperCodePrinter(CCodePrinter):
         error = PyErr_SetString('PyExc_TypeError', '"{} must have rank {}"'.format(collect_var, str(collect_var.rank)))
         body  += [IfSection(check, [error, Return([Nil()])])]
         if check_type : #Type check
-            numpy_dtype = self.find_in_numpy_dtype_registry(variable)
-            arg_dtype   = self.find_in_dtype_registry(self._print(variable.dtype), variable.precision)
-            check = PyccelNe(FunctionCall(numpy_get_type, [collect_var]), numpy_dtype)
-            info_dump = PythonPrint([FunctionCall(numpy_get_type, [collect_var]), numpy_dtype])
-            error = PyErr_SetString('PyExc_TypeError', '"{} must be {}"'.format(variable, arg_dtype))
+            check, error = self._get_array_type_check(variable, collect_var)
             body += [IfSection(check, [info_dump, error, Return([Nil()])])]
 
         if collect_var.rank > 1 and self._target_language == 'fortran' :#Order check
@@ -401,11 +408,12 @@ class CWrapperCodePrinter(CCodePrinter):
         if variable.rank > 0:
             body = self._body_array(variable, collect_var, check_type)
 
-        if variable.is_optional:
-            tmp_variable = Variable(dtype=variable.dtype, precision = variable.precision,
+        else:
+            if variable.is_optional:
+                tmp_variable = Variable(dtype=variable.dtype, precision = variable.precision,
                                     name = self.get_new_name(used_names, variable.name+"_tmp"))
 
-        body = self._create_collecting_value_body(variable, collect_var, check_type, tmp_variable)
+            body = self._create_collecting_value_body(variable, collect_var, check_type, tmp_variable)
 
         return body, tmp_variable
 
@@ -443,18 +451,7 @@ class CWrapperCodePrinter(CCodePrinter):
             collect_type = PyccelPyObject()
             collect_var = Variable(dtype=collect_type, is_pointer=True,
                 name = self.get_new_name(used_names, variable.name+"_tmp"))
-
-        '''elif isinstance(variable, ValuedVariable):
-            collect_type = PyccelPyObject()
-            collect_var = Variable(dtype=collect_type, is_pointer=True,
-                name = self.get_new_name(used_names, variable.name+"_tmp"))
-
-
-        elif variable.dtype is NativeComplex():
-            collect_type = PyccelPyObject()
-            collect_var = Variable(dtype=collect_type, is_pointer=True,
-                name = self.get_new_name(used_names, variable.name+"_tmp"))'''
-
+                
         return collect_var
 
     def get_PyBuildValue(self, used_names, variable):
--- a/pyccel/codegen/printing/cwrappercode.py
+++ b/pyccel/codegen/printing/cwrappercode.py
@@ -188,7 +188,10 @@ class CWrapperCodePrinter(CCodePrinter):
     # Functions that take care of creating cast or convert type function call :
     # --------------------------------------------------------------------
     def _get_array_type_check(self, variable, collect_var):
-       "TODO"
+        """
+        Responsible for collecting array type check and creating the
+        corresponding error code
+        """
 
         numpy_dtype = self.find_in_numpy_dtype_registry(variable)
         arg_dtype   = self.find_in_dtype_registry(self._print(variable.dtype), variable.precision)
@@ -199,7 +202,10 @@ class CWrapperCodePrinter(CCodePrinter):
         return check, error
     
     def _get_scalar_type_check(self, variable, collect_var, error_check = True):
-        "TODO"
+        """
+        Responsible for collecting numpy and python type check and creating the
+        corresponding error code
+        """
         
         if not error_check :
             scalar_check =  FunctionCall(PyArray_CheckScalar, [collect_var])
@@ -223,11 +229,12 @@ class CWrapperCodePrinter(CCodePrinter):
         if the valuedVariable is optional create body to collect the new value
 
         Parameters:
+        ----------
         tmp_variable : Variable
             The temporary variable  to hold result
         variable     : Variable
             The optional variable
-        collect_var  : variable
+        collect_var  : Variable
             the pyobject type variable  holder of value
     
         Returns
@@ -274,7 +281,7 @@ class CWrapperCodePrinter(CCodePrinter):
 
         Returns
         -------
-        body : [If block]
+        body : If block
         """
 
         var      = tmp_variable if tmp_variable else variable
@@ -297,9 +304,7 @@ class CWrapperCodePrinter(CCodePrinter):
         if error_check:
             sections.append(IfSection(LiteralTrue(), [error, Return([Nil()])]))
 
-        body = [If(*sections)]
-
-        return body
+        return If(*sections)
 
     def get_collect_function_call(self, variable, collect_var):
         """
@@ -434,7 +439,7 @@ class CWrapperCodePrinter(CCodePrinter):
                 tmp_variable = Variable(dtype=variable.dtype, precision = variable.precision,
                                         name = self.get_new_name(used_names, variable.name+"_tmp"))
 
-            body = self._body_scalar(variable, collect_var, check_type, tmp_variable)
+            body = [self._body_scalar(variable, collect_var, check_type, tmp_variable)]
 
         return body, tmp_variable
 
--- a/pyccel/codegen/printing/cwrappercode.py
+++ b/pyccel/codegen/printing/cwrappercode.py
@@ -334,9 +334,10 @@ class CWrapperCodePrinter(CCodePrinter):
 
         Returns:
         -------
-        python_type_check : FunctionCall
+        python_type_check : FunctionCall | LiteralTrue | None
             functionCall responsible for checking python data type
-            LiteralTrue() when error_check is False
+            LiteralTrue when error_check is False
+            None when default system precision is different than the variable precision
 
         numpy_type_check : FunctionCall or None
             functionCall responsible for checking numpy data type
--- a/pyccel/codegen/printing/cwrappercode.py
+++ b/pyccel/codegen/printing/cwrappercode.py
@@ -192,6 +192,16 @@ class CWrapperCodePrinter(CCodePrinter):
     def _get_wrapper_name(self, used_names, func):
         """
         create wrapper function name
+
+        Parameters:
+        -----------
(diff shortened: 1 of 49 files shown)


MenoIy (commented):
I was forced to change some old tests arguments so they can look like decorators args

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -184,41 +189,21 @@ def _get_check_type_statement(self, variable, collect_var):
...
-            the variable needed to collect
-        collect_var : variable
-            the pyobject variable
-        tmp_variable : variable
-            temporary variable to hold value default None
-
-        Returns
-        -------
-        body : If block
+        create wrapper function name
```
EmilyBourne, line 194: Could you add `Parameters` to this docstring please

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -273,88 +258,146 @@ def get_cast_function_call(self, cast_type, arg):
...
+    # --------------------------------------------------------------------
+    # Functions that take care of collecting (data type, rank) checks and creating the error code
+    # --------------------------------------------------------------------
+    def _generate_TypeError_message(self, variable):
+        """
+        Generate TypeError message from the variable informations (datatype, precision)
+        """
+        dtype     = variable.dtype
+        precision = '' if isinstance(dtype, NativeBool) else '{} bit '.format(variable.precision * 2 * 8)\
+                       if isinstance(dtype, NativeComplex) else '{} bit '.format(variable.precision * 8)
```
EmilyBourne, line 270: I don't find this very readable in one line. Is there a reason you didn't want to create a block?

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -273,88 +258,146 @@ def get_cast_function_call(self, cast_type, arg):
...
+        message = '"{var_name} must be {precision}{dtype}"'.format(
+                        var_name  = variable,
+                        precision = precision,
+                        dtype     = self._print(variable.dtype))
+        return message
+
+    def _get_array_type_check(self, variable, collect_var):
+        """
+        Responsible for collecting array type check and creating the
+        corresponding error code
```
EmilyBourne, line 281: Could you add `Parameters` and `Results` please

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -273,88 +258,146 @@ def get_cast_function_call(self, cast_type, arg):
...
+
+        check = PyccelNe(FunctionCall(numpy_get_type, [collect_var]), numpy_dtype)
+        error = PyErr_SetString('PyExc_TypeError', '"{} must be {}"'.format(variable, arg_dtype))
+
+        return check, error
+
+    def _get_scalar_type_check(self, variable, collect_var, error_check = True):
+        """
+        Responsible for collecting numpy and python type check and creating the
+        corresponding error code
```
EmilyBourne, line 295: Could you add `Parameters` and `Results` please

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -273,88 +258,146 @@ def get_cast_function_call(self, cast_type, arg):
...
-        body += [IfSection(LiteralTrue(), [self._create_collecting_value_body(variable, collect_var, tmp_variable),
-                    Assign(VariableAddress(variable), VariableAddress(tmp_variable))])]
-        body = [If(*body)]
 
-        return body
+        valued_var_check  = PyccelEq(VariableAddress(collect_var), VariableAddress(Py_None))
+        collect_body      = []
+
+        if variable.is_optional:
+            collect_body  = [Assign(VariableAddress(variable), VariableAddress(tmp_variable))]
```
EmilyBourne, line 344: Should this be an `AliasAssign`?

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -273,88 +258,146 @@ def get_cast_function_call(self, cast_type, arg):
...
+            section       = IfSection(valued_var_check, [Assign(VariableAddress(variable), Nil())])
```
EmilyBourne, line 345: See above. Shouldn't this be an `AliasAssign`?

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -425,23 +464,17 @@ def _body_management(self, used_names, variable, collect_var, cast_function, che
         Responsible for calling functions that take care of body creation
         """
```
EmilyBourne, line 465: Could you add `Parameters` to this docstring please.

Are all the arguments still used?

tests/epyccel/recognised_functions/test_math_funcs.py
```diff
@@ -400,7 +400,7 @@ def floor_return_type_real(x):
     fflags = "-Werror -Wconversion"
     f1 = epyccel(floor_return_type_real, language = language, fflags=fflags)
 
-    x = randint(100)
+    x = uniform(100)
```
EmilyBourne, line 403: Can you explain this change?

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -2577,7 +2577,7 @@ def create_full_like_arg_names(val):
         a = full_like(arr, val, float, 'F', shape = (2,3))
         return a[0,0],a[0,1],a[0,2],a[1,0],a[1,1],a[1,2]
 
-    size = randint(10)
+    size = uniform(10)
```
EmilyBourne, line 2580: Can you explain this change?

tests/epyccel/test_kind.py
```diff
@@ -32,7 +32,7 @@ def real_greater_bool(x0, x1):
     assert(real_greater_bool(1.5,1.2)==epyc_real_greater_bool(1.5,1.2))
 
 def test_input_output_matching_types(language):
-    @types('float32', 'float32')
+    @types('real', 'real')
```
EmilyBourne, line 35: Can you explain this change?

EmilyBourne (changes requested):

tests/epyccel/test_kind.py
MenoIy: - the argument feed to this function were python float.
- the precision of a python float is 64 bit on all system.

- now that we are collecting value as Pyobject I tried to avoid calling numpy and python check if the default precision is different as the variable precision.


- as example if the default precision is 64 bit and the variable is 64 bit I call both (python check and numpy check) because it can be anyone of them
- in the other hand the is no need to call python check

MenoIy: so i had 2 choice either change argument to np.float32 or change decorator to real

tests/epyccel/recognised_functions/test_numpy_funcs.py
MenoIy: same as above the type set in decorators is real but randint generate int

pyccel/codegen/printing/cwrappercode.py
MenoIy: Actually, no reason, just bad habit i will change it

tests/epyccel/test_kind.py
EmilyBourne: Ah ok no problem then :+1:
Although would you mind using `'float'` instead of `'real'`? `real` is the name of floating point types in fortran, but not in any other language. Everywhere else `real` is the real part of a complex number. `PythonReal` and `NumpyReal` have caused confusion a few times because we insist on using `real` everywhere in pyccel for `float`

pyccel/codegen/printing/cwrappercode.py
MenoIy: variable is not a pointer

pyccel/codegen/printing/cwrappercode.py
MenoIy: true I have set the castfunction to None and forgot to remove it

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: Then how are you assigning to its address?

Is it not a pointer as it is an optional function argument in the c function? But here it is a local variable so I think it probably should be. This probably means that we did this poorly last time ^^'. Maybe we should clone the arguments at the start and change optional variables to pointers? Optional variables which are not arguments in their context don't make sense

pyccel/codegen/printing/cwrappercode.py
MenoIy: done 637c2bd

pyccel/codegen/printing/cwrappercode.py
MenoIy: done 42ce0db

pyccel/codegen/printing/cwrappercode.py
MenoIy: done 537d915

pyccel/codegen/printing/cwrappercode.py
MenoIy: done cfb3455

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -273,88 +268,187 @@ def get_cast_function_call(self, cast_type, arg):
 
         return FunctionCall(cast_function, [arg])
 
+    # --------------------------------------------------------------------
+    # Functions that take care of collecting (data type, rank) checks and creating the error code
+    # --------------------------------------------------------------------
+    def _generate_TypeError_message(self, variable):
+        """
+        Generate TypeError message from the variable informations (datatype, precision)
```
EmilyBourne, line 276: Grammar : plural of `information` is `information`

pyccel/codegen/printing/cwrappercode.py
MenoIy: done 15bdea2

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -273,88 +268,187 @@ def get_cast_function_call(self, cast_type, arg):
...
+
+        message = '"{var_name} must be {precision}{dtype}"'.format(
+                        var_name  = variable,
+                        precision = precision,
+                        dtype     = self._print(variable.dtype))
+        return message
+
+    def _get_array_type_check(self, variable, collect_var):
+        """
+        Responsible for collecting array type check and creating the
```
EmilyBourne, line 295: Grammar : `Responsible of` -> `Responsible for`

This error appears a few times

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -273,88 +268,187 @@ def get_cast_function_call(self, cast_type, arg):
...
+        corresponding error code
+
+        Parameters:
+        ----------
+        variable     : Variable
+            The optional variable
+        collect_var  : Variable
+            the pyobject type variable  holder of value
```
EmilyBourne, line 303: I don't understand this. Do you mean:
`variable of type pyobject which holds the value` ?
What value?

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -273,88 +268,187 @@ def get_cast_function_call(self, cast_type, arg):
...
+        """
+        Responsible for collecting numpy and python type check and creating the
+        corresponding error code
+
+        Parameters:
+        ----------
+        variable     : Variable
+            The optional variable
+        collect_var  : Variable
+            the pyobject type variable  holder of value
```
EmilyBourne, line 331: See above

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -273,88 +268,187 @@ def get_cast_function_call(self, cast_type, arg):
...
-            True if the type is needed
 
         Returns
         -------
-        body : list
-            A list of statements
+        section      :
+            IfSection
+        collect_body :
+            body that collect new optional variable value
```
EmilyBourne, line 388: Grammar : `body that collect` -> `body which collects`

But maybe better:
`list containing the lines necessary to collect the new optional variable value`

pyccel/codegen/printing/cwrappercode.py
MenoIy: hmm will it be ok to put it here and update the comment ?
https://github.com/pyccel/pyccel/blob/c36ca7071eaed1c6775d87308acf618092fd3197/pyccel/codegen/printing/cwrappercode.py#L796

EmilyBourne (changes requested):

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: Yes, that is exactly the right place for it :+1:

pyccel/codegen/printing/cwrappercode.py
MenoIy: forgive my poor english

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: I don't know if you can do it in that list without making it unreadable though, you might need to just add a second list

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: no problem

pyccel/codegen/printing/cwrappercode.py
MenoIy: the value collected with PyAarg_Parsetuple

pyccel/codegen/printing/cwrappercode.py
MenoIy: done 0da2e5b

pyccel/codegen/printing/cwrappercode.py
MenoIy: done 0da2e5b

pyccel/codegen/printing/cwrappercode.py
MenoIy: done 0da2e5b

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: I wondered why you corrected this in the wrong direction. Somehow I managed to put it on the wrong line. "Responsible for" is correct english

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: `variable which holds the value collected with PyArg_Parsetuple` ?

tests/epyccel/test_kind.py
MenoIy: 'real' is used in a lot of place do you want me to change it everywhere ?

pyccel/codegen/printing/cwrappercode.py
MenoIy: -_-

tests/epyccel/test_kind.py
EmilyBourne: No. It's fine. I just don't want to add to the problem. If you could just change it in the 3/4 places you have added it that would be great

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: sorry

pyccel/codegen/printing/cwrappercode.py
MenoIy: yes i did it with a second list

pyccel/codegen/printing/cwrappercode.py
MenoIy: np xD
3260790

pyccel/codegen/printing/cwrappercode.py
MenoIy: done in
3665e15   333a1a6

pyccel/codegen/printing/cwrappercode.py
MenoIy: done in 333a1a6

tests/epyccel/test_kind.py
MenoIy: done in ceef690

pyccel/codegen/printing/cwrappercode.py
MenoIy: change in 
74cd27e

pyccel/codegen/printing/cwrappercode.py
MenoIy: done 
333a1a6

pyccel/codegen/printing/cwrappercode.py
MenoIy: @rakati  is there any reason why did you use  is_pointer and not rank > 0 ?
```PYTHON
wrapper_body += [Deallocate(i) for i in local_arg_vars if i.is_pointer]
```

EmilyBourne: :+1: that looks clearer. It just looks like you need to handle `Deallocate` for fortran (do nothing)

EmilyBourne (approved):

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -273,88 +268,187 @@ def get_cast_function_call(self, cast_type, arg):
...
+        python_type_check : FunctionCall
+            functionCall responsible for checking python data type
+            None when error_check is False
+
+        numpy_type_check : FunctionCall or None
+            functionCall responsible for checking numpy data type
+
+        error : FunctionCall or None
+            function call that raise TypeError
+            None when error_check is False
```
EmilyBourne, line 346: These are not in the same order as in the returns below

`python_type_check` says it is "None when error_check is False" but it is `LiteralTrue()`

EmilyBourne (changes requested):

pyccel/codegen/printing/cwrappercode.py
MenoIy: done 1039567

MenoIy: 2b1f9bf

EmilyBourne (approved):

rakati (approved):

yguclu (approved):

PR #762, opened 2021-02-16
Remove Sympy inheritance from AST nodes

### LINKED ISSUE #661: Remove non-standard sympy inheritance (e.g. sp_Float for LiteralFloat)

### LINKED ISSUE #662: Remove sympy inheritance (Basic, Expr, etc)
The inheritance from sympy is occasionally the cause of various bugs. In addition, changes to pyccel mean that it no longer needs this inheritance. It should therefore be removed. Instead all nodes in the pyccel ast should inherit from `pyccel.ast.basic.Basic`

### LINKED ISSUE #666: Remove _sympystr methods
If a class only has a `_sympystr` method then the method should be converted to a `__str__` magic method instead. If the `__str__` (or `__repr__`) already exists then the `_sympystr` method can just be removed.

EmilyBourne:
Remove sympy inheritance from pyccel AST nodes. Fixes #661. Fixes #662.
Remove unnecessary references to sympy. Fixes #666 

**Commit Summary**
- remove sympy inheritance
- replace references to `Expr` with `PyccelAstNode`
- Change `PyCodePrinter` inheritance to match `FCodePrinter` and `CCodePrinter`
- Use `kwargs` for dictionaries to avoid dangerous default value
- Add `__reduce_ex__` functions to `CodeBlock` and `Comment`
- Add `Singleton` class and `ArgumentSingleton` to replace sympy's singleton
- Remove handling of assignment where rhs is `Block` (i.e `With` etc)
- Add `length` property to `PythonZip`
- Clean functionexpr inheritance
- Remove `_sympystr` methods

### DIFF
--- a/pyccel/ast/basic.py
+++ b/pyccel/ast/basic.py
@@ -292,14 +292,6 @@ class Basic:
         """
         return len(self._user_nodes)==0
 
-    def __eq__(self, other):
-        #TODO: Remove with sympy inheritance
-        return id(self) == id(other)
-
-    def __hash__(self):
-        #TODO: Remove with sympy inheritance
-        return id(self)
-
 class PyccelAstNode(Basic):
     """Class from which all nodes containing objects inherit
     """
--- a/pyccel/ast/basic.py
+++ b/pyccel/ast/basic.py
@@ -29,9 +29,6 @@ class Basic:
     _fst = None
     _ignored_types = (Immutable, type)
 
-    def __new__(cls, *args, **kwargs):
-        hashable_args  = [a if not isinstance(a, list) else tuple(a) for a in args]
-
     def __init__(self):
         self._user_nodes = []
         self._fst = []
--- a/pyccel/ast/basic.py
+++ b/pyccel/ast/basic.py
@@ -46,8 +46,8 @@ class Basic:
                 c = convert_to_literal(c)
                 setattr(self, c_name, c)
 
-            elif isinstance(c, iterable_types):
-                if any(isinstance(ci, iterable_types) for ci in c):
+            elif iterable(c):
+                if any(iterable(ci) for ci in c):
                     raise TypeError("Basic child cannot be a tuple of tuples")
                 c = tuple(ci if (not isinstance(ci, (int, float, complex, str, bool)) \
                                  or self.ignore(ci)) \
@@ -180,8 +180,8 @@ class Basic:
             return
         self._recursion_in_progress = True
 
-        if isinstance(original, iterable_types):
-            assert(isinstance(replacement, iterable_types))
+        if iterable(original):
+            assert(iterable(replacement))
             assert(len(original) == len(replacement))
         else:
             original = (original,)
@@ -192,7 +192,7 @@ class Basic:
             rep = replacement[idx]
             if not self.ignore(found_node):
                 found_node.remove_user_node(self)
-            if isinstance(rep, iterable_types):
+            if iterable(rep):
                 for r in rep:
                     if not self.ignore(r):
                         r.set_current_user_node(self)
@@ -219,7 +219,7 @@ class Basic:
                             new_vi = prepare_sub(vi)
                         elif not self.ignore(vi):
                             vi.substitute(original, replacement, excluded_nodes)
-                    if isinstance(new_vi, iterable_types):
+                    if iterable(new_vi):
                         new_v.extend(new_vi)
                     else:
                         new_v.append(new_vi)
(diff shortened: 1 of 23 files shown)


pyccel/ast/basic.py
```diff
@@ -10,29 +10,25 @@
...
 """
 import ast
-from sympy.core.basic import Basic as sp_Basic
 
 __all__ = ('Basic', 'PyccelAstNode')
 
 dict_keys   = type({}.keys())
 dict_values = type({}.values())
 iterable_types = (list, tuple, dict_keys, dict_values)
+iterable = lambda x : isinstance(x, iterable_types)
```
MenoIy, line 19: where do you use this ?

EmilyBourne: It replaces [this](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/core.py#L17)

EmilyBourne: I am now using it more systematically. f943dbd

MenoIy (approved):

pyccel/ast/basic.py
yguclu: In order to know if an object is iterable, shouldn't we check it with
```python
hasattr(x, '__iter__')
```
?

pyccel/ast/functionalexpr.py
```diff
@@ -59,24 +57,27 @@ def index(self):
...
-    _attribute_nodes = FunctionalFor._attribute_nodes
+class FunctionalMin(GeneratorComprehension):
+    """ Represents a call to min for a list argument
+    >>> min([i in range(5)])
+    """
     name = 'min'
-
-#==============================================================================
-class FunctionalMap(GeneratorComprehension, FunctionalFor):
-    _attribute_nodes = FunctionalFor._attribute_nodes
```
yguclu, line 82: Did you remove this class intentionally?

pyccel/ast/itertoolsext.py
```diff
@@ -17,15 +17,15 @@ class Product(Basic):
 
     arg : list, tuple
     """
-    _attribute_nodes = ('_args',)
+    _attribute_nodes = ('_elements',)
```
yguclu, line 20: Is this a bugfix?

pyccel/utilities/metaclasses.py
```diff
@@ -0,0 +1,32 @@
...
+            cls._instances[cls] = super().__call__(*args, **kwargs)
+        return cls._instances[cls]
+
+class ArgumentSingleton(type):
+    """ Indicates that there is only one instance of the class
+    for any given set of arguments
+    """
+    _instances = {}
+    def __call__(cls, *args, **kwargs):
+        index = (cls, *args, *kwargs.values())
```
yguclu, line 29: I am worried that the order in which the keyword arguments are passed may play a role here. Further, I think that we should compare the keys as well in order to be safe. What about this?
```python
index = (cls, *args, *sorted(kwargs.items()))
```

pyccel/ast/basic.py
EmilyBourne: I don't think so. With that definition we would also include things like `TupleVariable`. In `Basic` this is definitely not what is wanted. I don't think it is what is wanted in core.py either as it is usually used to confirm that the object can be stored in a CodeBlock

EmilyBourne: Maybe there is a better name for this? tuple_like?

pyccel/ast/functionalexpr.py
EmilyBourne: It is not used and we have the class `PythonMap`. If there is a difference between the two classes then we should keep this one but I realised I couldn't think of one when I tried to add the docstring

pyccel/ast/itertoolsext.py
EmilyBourne: Yes

pyccel/utilities/metaclasses.py
EmilyBourne: I can do this. I didn't think this class would be used for cases which had multiple arguments, only for simple things such as `LiteralTrue`

EmilyBourne: Fixed in c4c7b0e

pyccel/ast/basic.py
yguclu: It's fine, thanks

pyccel/ast/itertoolsext.py
yguclu: Thanks for explaining!

pyccel/utilities/metaclasses.py
yguclu: Thanks

yguclu (approved):

PR #763, opened 2021-02-22
Fix bugs in numpy imag

mostafaMamoni:
Numpy's imag function is failing when passing an array argument containing.

Commit summary:
- Set rank, shape, and precision of `NumpyImag` object by creating a constructor in its class.
- Create `__new__` function in `NumpyImag`.
- Add tests.


pyccel/ast/numpyext.py
```diff
@@ -425,7 +425,7 @@ class NumpyImag(PythonImag):
     """
     def __new__(cls, arg):
         if not isinstance(arg.dtype, NativeComplex):
-            return arg
+            return NumpyZeros(arg.shape, dtype='int')
```
EmilyBourne, line 428: Right class :) . Are you sure about the dtype?

EmilyBourne: (e.g. `NativeReal`)

mostafaMamoni: ah sorry 😐 , that was just for some tests.

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3877,3 +3875,425 @@ def test_complex128():
     assert (f_complex64() == test_complex64())
     assert (f_complex128() == test_complex128())
 
+@pytest.mark.parametrize( 'language', (
+        pytest.param("fortran", marks = pytest.mark.fortran),
+        pytest.param("c", marks = [
+            pytest.mark.skip(reason="Cimage is not supporting arrays yet"),
```
EmilyBourne, line 3881: To support `imag` for arrays you need to set `is_elemental` to true.

See `NumpyExp` for an example by following the inheritance tree
`NumpyExp`->`NumpyUfuncUnary`->`NumpyUfuncBase`->`PyccelInternalFunction`

pyccel/ast/internals.py
```diff
@@ -37,7 +37,7 @@ def is_elemental(self):
         """ Indicates whether the function should be
         called elementwise for an array argument
         """
-        return False
+        return True
```
EmilyBourne, line 40: Please do not change the default behaviour. The property should be overriden in `NumpyImag` not changed here

pyccel/ast/numpyext.py
```diff
@@ -416,13 +415,34 @@ def Shape(arg):
...
+        self._precision = arg.precision
+        self._shape = process_shape(self.internal_var.shape)
+        self._rank  = len(self._shape)
+
+    @property
+    def is_elemental(self):
+        """ Indicates whether the function should be
+        called elementwise for an array argument
+        """
+        return True
```
EmilyBourne, line 445: Could you add this to `NumpyReal` too please we didn't do it when you were fixing `NumpyReal` as the loop unravelling wasn't implemented yet

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3877,3 +3875,416 @@ def test_complex128():
...
+    integer32 = np.int32(randint(1e6))
+    integer64 = np.int64(randint(1e6))
+    fl = np.float(randint(1e6))
+    fl32 = np.float32(randint(1e6))
+    fl64 = np.float64(randint(1e6))
+    cmplx64 = np.complex64(1+5j)
+    cmplx128 = np.complex128(1+5j)
+
+    f_bl = epyccel(test_bool, language=language)
+    f_bl = epyccel(test_bool, language=language)
```
rakati, line 3961: what is the difference btw these two lines?

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3877,3 +3875,416 @@ def test_complex128():
...
+        return b
+
+    import numpy as np
+
+    integer = randint(1e6)
+    bl = np.bool(randint(1e5))
+    integer8 = np.int8(randint(0, 127))
+    integer16 = np.int16(randint(1e6))
+    integer32 = np.int32(randint(1e6))
+    integer64 = np.int64(randint(1e6))
```
rakati, line 3953: can you use a higher range here, more than the int32 max?

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3877,3 +3875,416 @@ def test_complex128():
...
+    @types('complex128')
+    def test_complex128(a):
+        import numpy as np
+        b = np.imag(a)
+        return b
+
+    import numpy as np
+
+    integer = randint(1e6)
+    bl = np.bool(randint(1e5))
```
rakati, line 3949: you don't need a random number here, just test the two cases true and false.

tests/epyccel/recognised_functions/test_numpy_funcs.py
rakati: just for int64

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3877,3 +3875,416 @@ def test_complex128():
...
+        a = imag(arr)
+        s = shape(a)
+        return len(s), s[0], s[1], a[0,1], a[1,0]
+
+    def test_int():
+        from numpy import imag, shape, array, int as NumpyInt
+        arr = array([[4,5,6,2,1],[4,5,6,2,1]], NumpyInt)
+        a = imag(arr)
+        s = shape(a)
+        return len(s), s[0], s[1], a[0,1], a[1,0]
```
rakati, line 4195: for `test_numpy_imag_array_like_2d` can you use random values with the suggested types, and I think that it will be better to separate tests with functions, each function handle a case, so when the test is failed it will be easy to recognized and deal with it.

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -547,7 +561,7 @@ def log_call_i(x):
         return log(x)
 
     f1 = epyccel(log_call_i, language = language)
-    x = randint(low=min_float, high=1e6)
+    x = randint(low=min_int, high=max_int)
```
EmilyBourne, line 564: `log` is a function which is applied to positive numbers. It shouldn't be used for negative numbers. If it is used for negative numbers then it returns a `RuntimeWarning` and `nan`, but `nan`s cannot be compared:
```
In [2]: a=np.log(-5)
/home/emily/.local/bin/ipython3:1: RuntimeWarning: invalid value encountered in log
  #!/usr/bin/python3

In [3]: a
Out[3]: nan

In [4]: b=np.log(-4)
/home/emily/.local/bin/ipython3:1: RuntimeWarning: invalid value encountered in log
  #!/usr/bin/python3

In [5]: b
Out[5]: nan

In [6]: a==b
Out[6]: False

In [7]: a-b
Out[7]: nan

In [8]: np.isclose(a,b)
Out[8]: False
```

You should use `x = randint(low=0, high=max_int)`

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3877,3 +3891,449 @@ def test_complex128():
...
+        integer = randint(-9223372036854775808, 9223372036854775807)
+    integer8 = np.int8(randint(-128, 127))
+    integer16 = np.int16(randint(-32768, 32767))
+    integer32 = np.int32(randint(-2147483648, 2147483647))
+    integer64 = np.int64(randint(-9223372036854775808, 9223372036854775807))
+    fl = np.float(randint(-2147483648, 2147483647))
+    fl32 = np.float32(randint(-2147483648, 2147483647))
+    fl64 = np.float64(randint(-9223372036854775808, 9223372036854775807))
+    cmplx64 = np.complex64(1+5j)
+    cmplx128 = np.complex128(1+5j)
```
rakati, line 3976: can you use a random number here?
you can generate random complex using `cmplx = randint(some_range) + randint(some_range) * 1j`
you may use the `numpy.random.uniform` for a random float number in place of randint.

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3877,3 +3891,449 @@ def test_complex128():
...
+        integer = randint(-2147483648, 2147483647)
+    else:
+        integer = randint(-9223372036854775808, 9223372036854775807)
+    integer8 = np.int8(randint(-128, 127))
+    integer16 = np.int16(randint(-32768, 32767))
+    integer32 = np.int32(randint(-2147483648, 2147483647))
+    integer64 = np.int64(randint(-9223372036854775808, 9223372036854775807))
+    fl = np.float(randint(-2147483648, 2147483647))
+    fl32 = np.float32(randint(-2147483648, 2147483647))
+    fl64 = np.float64(randint(-9223372036854775808, 9223372036854775807))
```
rakati, line 3974: can you use your declared variable of number limits instead of literal values?
aka max_float ...

EmilyBourne:
The order of `NumpyImag` and `NumpyReal` needs to be set

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3536,25 +3550,33 @@ def test_complex128(a):
...
-    bl = np.bool(randint(1e5))
-    integer8 = np.int8(randint(0, 127))
-    integer16 = np.int16(randint(1e6))
-    integer32 = np.int32(randint(1e6))
-    integer64 = np.int64(randint(1e6))
-    fl = np.float(randint(1e6))
-    fl32 = np.float32(randint(1e6))
-    fl64 = np.float64(randint(1e6))
+    integer8 = randint(np.iinfo(np.int8(1)).min, np.iinfo(np.int8(1)).max, dtype=np.int8)
+    integer16 = randint(np.iinfo(np.int16(1)).min, np.iinfo(np.int16(1)).max, dtype=np.int16)
```
rakati, line 3554: instead of `np.iinfo(np.int8(1)).min` simply use `iinfo('int8').min`.
the same for the others, I don't see any meaning for `int8(1)` and i`nt16(1)`.

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3640,118 +3662,134 @@ def test_complex128(a):
...
+    integer64 = randint(min_int64, max_int64, size=(5), dtype=np.int64)
+
+    fl = uniform(min_float / 2, max_float / 2, size=(5))
+    fl32 = uniform(min_float32, max_float32, size=(5))
+    fl32 = np.float32(fl32)
+    fl64 = uniform(min_float64 / 2, max_float64 / 2, size=(5))
+
+    cmplx64 = uniform(low=min_float32 / 2, high=max_float32 / 2, size=5) + uniform(low=max_float32 / 2, high=max_float32 / 2, size=5) * 1j
+    cmplx64 = np.complex64(cmplx64)
+    cmplx128 = uniform(low=min_float64 / 2, high=max_float64 / 2, size=5) + uniform(low=min_float64 / 2, high=max_float64 / 2, size=5) * 1j
```
rakati, line 3762: try to choose meaningful names here, you may use this notation:
`arr_1d_int = randint(min_int, max_int, size=(5), dtype=np.int)`
`arr_1d_int32 = randint(min_int32, max_int32, size=(5), dtype=np.int32)`

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3640,118 +3662,134 @@ def test_complex128(a):
...
-        from numpy import real, shape, array, complex128
-        arr = array([4,5,6,2,1], complex128)
+    @types('complex128[:]')
+    def test_complex128(arr):
+        from numpy import real, shape
         a = real(arr)
         s = shape(a)
         return len(s), s[0], a[0]
 
+    import numpy as np
```
rakati, line 3745: the import of `numpy as np` is already done in this [line ](https://github.com/pyccel/pyccel/pull/763/files#diff-666e38e0289c81a3a42526c953aabe890f5b6e8d0d40fe03e9260900df37d6c0R3551)

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3764,116 +3802,577 @@ def test_complex128():
...
+        b = np.imag(a)
+        return b
+
+    @types('complex128')
+    def test_complex128(a):
+        import numpy as np
+        b = np.imag(a)
+        return b
+
+    import numpy as np
```
rakati, line 4002: same here, double import.

rakati (changes requested):
this is a huge file and hard to track, I suggested in a previous comment to reorder the lines, and try to make for each function the assert and generating of the random values near to it, or create a separate module for that, because this file is very long, I don't know in the future if someone has failed the test with how he can handle that. so please consider that and try to make this file smaller or just reorder things to be easy to follow.

mostafaMamoni:
> this is a huge file and hard to track, I suggested in a previous comment to reorder the lines, and try to make for each function the assert and generating of the random values near to it, or create a separate module for that, because this file is very long, I don't know in the future if someone has failed the test with how he can handle that. so please consider that and try to make this file smaller or just reorder things to be easy to follow.

the tests should be in the same file as wroten in the issue, for making the asserts near of the there functions i don't see any difference it will just re-forming the test, the error detection should be easy between your suggestion and the current implementation.

tests/epyccel/recognised_functions/test_numpy_funcs.py
mostafaMamoni: `arr` keyword is mentioned in the name of the test, so these names are describing just the data type that will be tested.

tests/epyccel/recognised_functions/test_numpy_funcs.py
mostafaMamoni: sorry but i can't see the line.

EmilyBourne: Imports should go at the top of the file unless you have a special reason for them to be elsewhere (e.g. in a function which will be epyccelized to avoid loosing the import)

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3764,116 +3813,580 @@ def test_complex128():
...
+    integer32 = randint(min_int32, max_int32, size=(2, 5), dtype=np.int32)
+    integer64 = randint(min_int64, max_int64, size=(2, 5), dtype=np.int64)
+
+    fl = uniform(min_float / 2, max_float / 2, size=(2, 5))
+    fl32 = uniform(min_float32, max_float32, size=(2, 5))
+    fl32 = np.float32(fl32)
+    fl64 = uniform(min_float64 / 2, max_float64 / 2, size=(2, 5))
+
+    cmplx64 = uniform(low=min_float32 / 2, high=max_float32 / 2, size=(2, 5)) + uniform(low=max_float32 / 2, high=max_float32 / 2, size=(2, 5)) * 1j
+    cmplx64 = np.complex64(cmplx64)
```
rakati, line 3907: I don't see the purpose of the second line.

mostafaMamoni: without the tests was failing with `cmplx64 should be 64 bits complex`

mostafaMamoni: by the way, can you please resolve what is already done.

tests/epyccel/recognised_functions/test_numpy_funcs.py
rakati: take a look above you will find these variables use for single values, not arrays.

tests/epyccel/recognised_functions/test_numpy_funcs.py
EmilyBourne: The first line is a 128 bit `complex` while the second line converts it to a 64 bit `complex64`. Maybe the first line should create a variable with a different name to make this clear?

tests/epyccel/recognised_functions/test_numpy_funcs.py
mostafaMamoni: same thing answer, those names are just to identify the type.

pyccel/ast/numpyext.py
EmilyBourne: This comment seems to have become lost. Please add this property to `NumpyReal`

EmilyBourne (changes requested):

pyccel/ast/numpyext.py
```diff
@@ -416,13 +423,35 @@ def Shape(arg):
...
+    """Represents a call to  numpy.imag for code generation.
 
     > a = 1+2j
     > np.imag(a)
     2.0
     """
+    def __new__(cls, arg):
+        if not isinstance(arg.dtype, NativeComplex):
+            dtype=NativeInteger() if isinstance(arg.dtype, NativeBool) else arg.dtype
+            if isinstance(arg, Literal) or (isinstance(arg, Variable) and not arg.allocatable):
```
EmilyBourne, line 437: Shouldn't this condition be `if isinstance(arg, Literal) and arg.rank == 0` ? This case doesn't handle the  case where the argument is a stack array

pyccel/ast/numpyext.py
```diff
@@ -416,13 +423,35 @@ def Shape(arg):
...
+                return convert_to_literal(0, dtype, arg.precision)
+            return NumpyZeros(arg.shape, dtype=dtype)
+        return super().__new__(cls, arg)
+
+    def __init__(self, arg):
+        super().__init__(arg)
+        self._precision = arg.precision
+        self._order = arg.order
+        self._shape = process_shape(self.internal_var.shape)
```
EmilyBourne, line 446: Do we need to process the shape? Won't it already be processed?

pyccel/ast/numpyext.py
```diff
@@ -416,13 +423,35 @@ def Shape(arg):
...
+        self._rank  = len(self._shape)
```
EmilyBourne, line 447: I think you have forgotten to set the dtype

EmilyBourne (changes requested):

rakati (approved):

pyccel/ast/numpyext.py
```diff
@@ -416,13 +423,36 @@ def Shape(arg):
...
     > a = 1+2j
     > np.imag(a)
     2.0
     """
+    def __new__(cls, arg):
+        if not isinstance(arg.dtype, NativeComplex):
+            dtype=NativeInteger() if isinstance(arg.dtype, NativeBool) else arg.dtype
+            if isinstance(arg, (Literal, Variable)) and arg.rank == 0:
+                return convert_to_literal(0, dtype, arg.precision)
+            return NumpyZeros(arg.shape, dtype=dtype)
```
EmilyBourne, line 439: Why do you only return a literal if the argument is of type: `Literal` or `Variable` ?

pyccel/ast/numpyext.py
```diff
@@ -416,13 +423,36 @@ def Shape(arg):
...
+        return super().__new__(cls, arg)
+
+    def __init__(self, arg):
+        super().__init__(arg)
+        self._precision = arg.precision
+        self._order = arg.order
+        self._dtype = arg.dtype if not isinstance(arg.dtype, NativeComplex) else NativeReal()
```
EmilyBourne, line 446: Is this if ternary useful? If the dtype is not `NativeComplex` then `__new__` should return a different type so this `__init__` will never be called. Or have I missed something?

EmilyBourne (changes requested):

pyccel/ast/numpyext.py
mostafaMamoni: what if dtype is `NativeComplex` then return will be a complex type

EmilyBourne: To be clear, my suggested fix is:
```
self._dtype = NativeReal()
```

The question is when is the dtype not `NativeComplex`?

pyccel/ast/numpyext.py
mostafaMamoni: because the imag part of every non-complex scalar is 0, so i took care about just the `dtype`, i suspect `Variable` objects can surprise me but since the rank is set correctly, i think it is fine to return just 0 literal with the correct `dtype` for know.

EmilyBourne: So:
```
np.imag(1+2)
```
Should return:
```
[0]
```
?

EmilyBourne: Returning a literal for every non-complex scalar is exactly right. But that is not what you are doing

pyccel/ast/numpyext.py
EmilyBourne: When `__new__` returns `super().__new__(...)` then the `__init__` of the class is called. When `__new__` returns anything else then it isn't. So I think `NumpyImag.__init__` should only be called if the dtype of the argument is `NativeComplex()`?

mostafaMamoni: yes, you are right.

EmilyBourne (approved):
Please merge the master branch and resolve the conflicts in test_numpy_functions.py

EmilyBourne (approved):

EmilyBourne (approved):

yguclu (approved):

PR #764, opened 2021-02-23
Store LiteralInteger value in a private variable

EmilyBourne:
**Commit Summary**
- Store LiteralInteger value in a private variable
- Remove dead code in parser/semantic.py
- Use operations rather than calling sympy operations directly

### DIFF
--- a/pyccel/ast/sympy_helper.py
+++ b/pyccel/ast/sympy_helper.py
@@ -129,7 +129,7 @@ def pyccel_to_sympy(expr, symbol_map, used_names):
     if isinstance(expr, LiteralInteger):
         return sp.Integer(expr.python_value)
 
-    elif isinstance(expr, LiteralReal):
+    elif isinstance(expr, LiteralFloat):
         return sp.Float(expr.python_value)
 
     elif isinstance(expr, LiteralComplex):
--- a/pyccel/ast/sympy_helper.py
+++ b/pyccel/ast/sympy_helper.py
@@ -19,7 +19,7 @@ from .builtins  import PythonRange, PythonTuple
 
 from .mathext   import MathCeil
 
-from .literals  import LiteralInteger, LiteralFloat
+from .literals  import LiteralInteger, LiteralFloat, Literal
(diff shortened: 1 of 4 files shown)


MenoIy (approved):

pyccel/ast/sympy_helper.py
```diff
@@ -139,7 +142,7 @@ def pyccel_to_sympy(expr, symbol_map, used_names):
 
     elif isinstance(expr, PyccelMul):
         args = [pyccel_to_sympy(e, symbol_map, used_names) for e in expr.args]
-        return sp.Mul(*args)
+        return args[0] * args[1]
```
yguclu, line 145: Is our `PyccelMul` a binary operation, or can it have more than two arguments? Sorry but I don't remember

pyccel/ast/sympy_helper.py
```diff
@@ -151,11 +154,11 @@ def pyccel_to_sympy(expr, symbol_map, used_names):
 
     elif isinstance(expr, PyccelAdd):
         args = [pyccel_to_sympy(e, symbol_map, used_names) for e in expr.args]
-        return sp.Add(*args)
+        return args[0] + args[1]
```
yguclu, line 157: Same question as for `PyccelMul`

pyccel/ast/sympy_helper.py
EmilyBourne: Yes, `PyccelMul` inherits from `PyccelBinaryOperator`. This mimicks the way python's ast handles this (arithmetic operators are all instances of `ast.BinOp`)

pyccel/ast/sympy_helper.py
EmilyBourne: Yes, see above

pyccel/ast/sympy_helper.py
yguclu: Thanks

yguclu (approved):

Issue #758, opened 2021-02-15
Wrapper generates potentially dangerous warnings on macos


EmilyBourne:
**Describe the bug**
Numpy provides shapes with the type `npy_intp` which varies for different platforms. On mac it is equivalent to `long`. However we always collect it as a pointer to `int64_t`. As the contents of pointers are not cast but merely interpreted as the expected type this could lead to errors.

**To Reproduce**
The following code:
```python
def test_bool():
        from numpy import real, shape, array
        arr = array([[4,5,6,2,1],[4,5,6,2,1]], bool)
        a = real(arr)
        s = shape(a)
        return len(s), s[0], s[1], a[0,1], a[1,0]
```

generates the following warning on macos
```bash
  /Users/runner/hostedtoolcache/Python/3.9.1/x64/lib/python3.9/site-packages/pyccel/codegen/python_wrapper.py:143: UserWarning: mod_86106oe7_wrapper.c:24:38: warning: incompatible pointer types passing 'npy_intp *' (aka 'long *') to parameter of type 'int64_t *' (aka 'long long *') [-Wincompatible-pointer-types]
      c.shape = numpy_to_ndarray_shape(PyArray_SHAPE(o), c.nd);
                                       ^~~~~~~~~~~~~~~~
  /Users/runner/work/pyccel/pyccel/tests/__epyccel__/__pyccel__/ndarrays/ndarrays.h:118:46: note: passing argument to parameter 'np_shape' here
  int64_t     *numpy_to_ndarray_shape(int64_t *np_shape, int nd);
                                               ^
  mod_86106oe7_wrapper.c:25:42: warning: incompatible pointer types passing 'npy_intp *' (aka 'long *') to parameter of type 'int64_t *' (aka 'long long *') [-Wincompatible-pointer-types]
      c.strides = numpy_to_ndarray_strides(PyArray_STRIDES(o), c.type_size, c.nd);
                                           ^~~~~~~~~~~~~~~~~~
  /Users/runner/work/pyccel/pyccel/tests/__epyccel__/__pyccel__/ndarrays/ndarrays.h:117:48: note: passing argument to parameter 'np_strides' here
  int64_t     *numpy_to_ndarray_strides(int64_t *np_strides, int type_size, int nd);
                                                 ^
```

**Expected behavior**
We should use `npy_intp` directly to avoid all potential errors

PR #766, opened 2021-02-24
[WIP] wrapper

MenoIy:
# TODO

- [ ] finish docs
- [ ] clean code
- [ ] either remove flags from test or cast things in ndarrays lib


pyccel/stdlib/cwrapper/cwrapper.c
```diff
@@ -0,0 +1,216 @@
...
+	return true;
+}
+
+bool	PyFloat_to_Float(Pyobject *o, float *f)
+{
+	double	out;
+
+
+	out = PyFloat_AsDouble(o);
+	if (out  == -1.0 && PyErr_Occured())
```
EmilyBourne, line 112: Why do you need both these conditions? Is `PyErr_Occured()` not sufficient?

pyccel/stdlib/cwrapper/cwrapper.c
```diff
@@ -0,0 +1,216 @@
...
+** A Cast function that convert numpy array variable into ndarray variable,
+** by copying its information and data to a new variable of type ndarray struct
+** and return this variable to be used inside c code.
+*/
+
+t_ndarray		PyArray_to_ndarray(PyObject *o)
+{
+	t_ndarray	array;
+
+	c.nd          = PyArray_NDIM(o);
```
EmilyBourne, line 143: c = array ?

MenoIy: yeah I finish up by keeping array declared inside function, I still don't know if i need to pass array by adress as argument

EmilyBourne: It depends on how you plan to use the function. I think it's up to you. I like it as a return

pyccel/ast/numpy_wrapper.py
```diff
@@ -0,0 +1,289 @@
+#------------------------------------------------------------------------------------------#
+# This file is part of Pyccel which is released under MIT License. See the LICENSE file or #
+# go to https://github.com/pyccel/pyccel/blob/master/LICENSE for full license details.     #
+#------------------------------------------------------------------------------------------#
+# pylint: disable=missing-function-docstring
+
+"""
+Handling the transitions between python code and C code.
```
mostafaMamoni, line 8: can you be more specific in this docstring ?

pyccel/ast/numpy_wrapper.py
```diff
@@ -0,0 +1,289 @@
...
+numpy_long_type         = Variable(dtype=NativeInteger(),  name = 'NPY_LONG', precision = 4)
+numpy_ulong_type        = Variable(dtype=NativeInteger(),  name = 'NPY_ULONG', precision = 4)
+numpy_longlong_type     = Variable(dtype=NativeInteger(),  name = 'NPY_INT64', precision = 4)
+numpy_ulonglong_type    = Variable(dtype=NativeInteger(),  name = 'NPY_ULONGLONG', precision = 4)
+numpy_float_type        = Variable(dtype=NativeInteger(),  name = 'NPY_FLOAT', precision = 4)
+numpy_double_type       = Variable(dtype=NativeInteger(),  name = 'NPY_DOUBLE', precision = 4)
+numpy_longdouble_type   = Variable(dtype=NativeInteger(),  name = 'NPY_LONGDOUBLE', precision = 4)
+numpy_cfloat_type       = Variable(dtype=NativeInteger(),  name = 'NPY_CFLOAT', precision = 4)
+numpy_cdouble_type      = Variable(dtype=NativeInteger(),  name = 'NPY_CDOUBLE', precision = 4)
+numpy_clongdouble_type  = Variable(dtype=NativeInteger(),  name = 'NPY_CLONGDOUBLE', precision = 4)
```
mostafaMamoni, line 146: why all those objects have the same precision ?, and it is good for compatibility ?.

EmilyBourne: These objects all have the same type. They do not represent a bool object or a float object etc, but are just flags that numpy uses to tell us which datatype is saved in the array. (This code is not new, it is just moved from ast/cwrapper.py to split the python and the numpy things)

The compatibility question is a good one though. Are these objects stored in an `int`? According to [this](https://numpy.org/doc/1.17/reference/c-api.dtype.html) they are enumerate values, so `int` makes sense to me

pyccel/ast/numpy_wrapper.py
```diff
@@ -0,0 +1,289 @@
...
+    'numpy_dtype_registry',
+    'PyArray_CheckScalar',
+    'PyArray_ScalarAsCtype',
+)
+
+#-------------------------------------------------------------------
+#                      Numpy functions
+#-------------------------------------------------------------------
+# All the numpy function list are part of  numpy/c api
+# https://numpy.org/doc/stable/reference/c-api/array.html
```
EmilyBourne, line 48: Are these functions still used? Or is everything handled in cwrapper.h/cwrapper.c?

pyccel/ast/numpy_wrapper.py
```diff
@@ -0,0 +1,289 @@
...
+            flag = numpy_flag_c_contig
+    else:
+        flag = LiteralInteger(0)
+
+    func = FunctionDef(name = 'Check_Array',
+                       arguments = [Variable(dtype = PyccelPyObject(), name = 'a', is_pointer = True),
+                                    Variable(dtype = NativeInteger(),  name = 'rank'),
+                                    Variable(dtype = NativeInteger(),  name = 'flag')],
+                       body      = [],
+                       results   = [Variable(dtype = PyccelPyArrayObject(), name = 'array', is_pointer = True)])
```
EmilyBourne, line 253: Does the `FunctionDef` need to be recreated every time we call `Check_array`? I don't see anything that depends on the arguments

pyccel/codegen/cwrapper.py
```diff
@@ -81,7 +83,16 @@ def create_c_setup(mod_name,
     linker_flags_str   = ('extra_link_args = {0}'.format(print_list(linker_flags))
                    if flags else None)
 
-    args = [mod, wrapper_file, files, include_str, libs_str, libdirs_str, flags_str, linker_flags_str]
+    numpy_max_acceptable_version = [1, 19]
```
EmilyBourne, line 86: Maybe it would make more sense to mark the last tested version of numpy in ast/numpywrapper.py?

pyccel/stdlib/cwrapper/cwrapper.c
```diff
@@ -0,0 +1,273 @@
...
+ *
+ *  Returns    :
+ * 		reference to PyArray Object
+ *      returns NULL on error
+ * reference of the used c/python api function
+ * -------------------------------------------
+ * https://numpy.org/doc/stable/reference/c-api/array.html
+ */
+
+PyArrayObject	*Check_Array(PyObject *a, int rank, int flags)
```
EmilyBourne, line 238: Shouldn't `Check_Array` also check the dtype?

EmilyBourne: Is this done separately to avoid having lots of if blocks for the correct error message?
I'm not sure whether to suggest passing the string as an argument, or if that will lead to a big ugly function call

pyccel/ast/numpy_wrapper.py
MenoIy: not all of them
```
numpy_get_type
PyArray_ScalarAsCtype
```
are still used

pyccel/ast/numpy_wrapper.py
MenoIy: yeah it can be global one and used in functionCall with the corresponding arguments

pyccel/stdlib/cwrapper/cwrapper.c
MenoIy: using interface and to avoid multiple type check the check data is only written when needed in the converter function

EmilyBourne: ok 👍

MenoIy: ==> I'm not sure whether to suggest passing the string as an argument, or if that will lead to a big ugly function call ?
i was thinking about something like  global static array so when check type is needed i can pass the type num as arguments or 0 if not needed   , and use array for error code

pyccel/stdlib/cwrapper/cwrapper.h
```diff
@@ -0,0 +1,49 @@
...
+float			PyFloat_to_Float(PyObject *o);
+double			PyDouble_to_Double(PyObject *o);
+
+bool			PyBool_to_Bool(PyObject *o);
+
+/* numpy array to ndarray */
+//t_ndarray	PyArray_to_ndarray(PyArrayObject *o);
+
+
+/* casting c type to python object */
```
EmilyBourne, line 39: Do you plan to also add the reverse casts?

pyccel/stdlib/cwrapper/cwrapper.c
```diff
@@ -0,0 +1,400 @@
...
+ * https://docs.python.org/3/c-api/complex.html#c.PyComplex_RealAsDouble
+ * https://docs.python.org/3/c-api/complex.html#c.PyComplex_ImagAsDouble
+ * https://docs.python.org/3/c-api/long.html#c.PyLong_AsLong
+ * https://docs.python.org/3/c-api/long.html#c.PyLong_AsLongLong
+ */
+
+float complex PyComplex_to_Complex64(PyObject *o)
+{
+	float complex	c;
+	float			real_part;
```
EmilyBourne, line 37: Why so much whitespace?

pyccel/stdlib/cwrapper/cwrapper.h
```diff
@@ -0,0 +1,49 @@
...
+PyObject	*Complex_to_PyComplex(double complex c);
+
+PyObject	*Bool_to_PyBool(bool b);
+
+PyObject	*Int_to_PyLong(int64_t i);
+
+PyObject	*Double_to_PyDouble(double d);
```
EmilyBourne, line 46: Could we add the numpy types here too?

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -56,1026 +50,697 @@ class CWrapperCodePrinter(CCodePrinter):
...
-        self._to_free_PyObject_list = []
-        self._function_wrapper_names = dict()
-        self._global_names = set()
-        self._module_name = None
-
+        self._target_language             = target_language
+        self._function_wrapper_names      = dict()
+        self._global_names                = set()
+        self._module_name                 = None
+        self.converter_functions_dict     = {}
```
EmilyBourne, line 57: There are 2 different ways to initialise dicts here. You should pick 1

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -56,1026 +50,697 @@ class CWrapperCodePrinter(CCodePrinter):
...
+                precision = precision,
+                dtype     = variable.dtype)
+        return PyErr_SetString('PyExc_TypeError', message)
 
+    #--------------------------------------------------------------------
+    #                   Convert functions
+    #--------------------------------------------------------------------
 
-    def _body_scalar(self, variable, collect_var, error_check = False, tmp_variable = None):
+    def generate_scalar_converter_function(self, used_names, variable, check_is_needed = True):
```
EmilyBourne, line 260: Does this need to be in the codegen? Or can it be elsewhere e.g. in the ast or maybe better in a new utilities folder. It is not AST but it is not printing either, it is logic

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -56,1026 +50,697 @@ class CWrapperCodePrinter(CCodePrinter):
...
+        funcDef = FunctionDef(name     = func_name,
+                            arguments  = [py_variable, c_variable],
+                            results    = [Variable(name = 'r', dtype = NativeInteger(), is_temp = True)],
+                            body       = body)
 
-        return If(*sections)
+        return funcDef
 
-    def _body_array(self, variable, collect_var, check_type = False) :
+    def generate_array_converter_function(self, used_names, variable, check_is_needed = True):
```
EmilyBourne, line 313: See above

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -56,1026 +50,697 @@ class CWrapperCodePrinter(CCodePrinter):
...
 
-            body = [self._body_scalar(variable, collect_var, check_type, tmp_variable)]
+    def generate_tuple_converter_function(self, used_names, variable):
+        pass
+        #TODO
 
-        return body, tmp_variable
+    def generate_list_converter_function(self, used_names, variable):
+        pass
+        #TODO
```
EmilyBourne, line 359: Will this convert to or from list? We don't handle lists so you don't need to convert to a list. We just want a permissive version of tuple collection

pyccel/parser/syntactic.py
```diff
@@ -912,7 +912,7 @@ def _visit_Call(self, stmt):
         return func
 
     def _visit_keyword(self, stmt):
-
+        print(['HERE'], stmt)
```
EmilyBourne, line 915: Don't forget to remove this

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: Then again, maybe that's true for most things in this file

pyccel/stdlib/cwrapper/cwrapper.c
```diff
@@ -0,0 +1,400 @@
...
+	array->type_size   = PyArray_ITEMSIZE(pyarray);
+	array->type        = PyArray_TYPE(pyarray);
+	array->length      = PyArray_SIZE(pyarray);
+	array->buffer_size = PyArray_NBYTES(pyarray);
+
+	array->shape       = (int64_t*)malloc(sizeof(int64_t) * array->nd);
+	memcpy(array->shape, PyArray_SHAPE(pyarray), sizeof(int64_t) * array->nd);
+
+	array->strides     = (int64_t*)malloc(sizeof(int64_t) * array->nd);
+	memcpy(array->strides, PyArray_STRIDES(pyarray), sizeof(int64_t) * array->nd);
```
EmilyBourne, line 395: The shape and stride cannot be handled with memcpy. See [ndarrays.c](https://github.com/pyccel/pyccel/blob/master/pyccel/stdlib/ndarrays/ndarrays.c#L326-L351) and issue #758

pyccel/stdlib/cwrapper/cwrapper.h
MenoIy: normally yes, but we should know when, I support we can do it for all low precision.

pyccel/codegen/printing/cwrappercode.py
MenoIy: yes I'm going to move them all to the ast/x_wrapper once i finish

pyccel/codegen/printing/cwrappercode.py
MenoIy: from pylist  to c array

pyccel/stdlib/cwrapper/cwrapper.c
MenoIy: ah, I forget about it thx

pyccel/codegen/printing/cwrappercode.py
MenoIy: anyway  im gonna remove it later. it will be handled same as tuple

pyccel/stdlib/cwrapper/cwrapper.h
EmilyBourne: :/ Yes we really need to figure out a way to differentiate between `np.float64` and `float` (in a new issue and PR). I wonder if we could save the precision as -1 when it is the default 🤔

EmilyBourne: > I suppose we can do it for all low precision.

Or better all precision which doesn't match the default, then it will at least fix the problem where we truncate int64 output

pyccel/stdlib/cwrapper/cwrapper.c
```diff
@@ -0,0 +1,381 @@
...
+
+    return ndarray_strides;
+}
+
+/*
+** copy numpy shape to nd_array shape, and return it in a new array, to
+** avoid the problem of variation of system architecture because numpy shape
+** is not saved in fixed length type.
+*/
+static int64_t     *_numpy_to_ndarray_shape(int64_t *np_shape, int nd)
```
EmilyBourne, line 337: Can we use `npy_intp` for the numpy shape type? As we are already importing python things here it should be ok and it will fix issue #758

pyccel/stdlib/cwrapper/cwrapper.c
```diff
@@ -0,0 +1,381 @@
...
+ * reference of the used c/numpy api function
+ * -------------------------------------------
+ * https://numpy.org/doc/stable/reference/c-api/array.html
+ */
+
+/*
+** convert numpy strides to nd_array strides, and return it in a new array, to
+** avoid the problem of different implementations of strides in numpy and ndarray.
+*/
+static int64_t	*_numpy_to_ndarray_strides(int64_t *np_strides, int type_size, int nd)
```
EmilyBourne, line 321: See above. Can we use `npy_intp`?

pyccel/stdlib/cwrapper/cwrapper.c
MenoIy: yes

EmilyBourne:
I have a couple of questions/comments about this:
> i give up working with O& flag when parsing arguments for 3 reason :
>   - non compatibilite with interfaces
>   - impossible to use ndarrays with fortran
>   - memory leak : when managing error we return early befor freeing allocated object

1. Interfaces - We already discussed that this solution is impossible for interfaces. Do you think it is a problem if interfaces are handled differently? I don't see how they could be handled identically to the rest. The best I can see is:
    ```
    PyObject *gen_2_wrapper(PyObject *self, PyObject *args, PyObject *kwargs)
    {
        int64_t check;
        PyObject *y_tmp;
        PyObject *x_tmp;
        PyObject *result;
        static char *kwlist[] = {
            "y",
            "x",
            NULL
        };
    
        if (!PyArg_ParseTupleAndKeywords(args, kwargs, "OO", kwlist, &y_tmp, &x_tmp))
        {
            return NULL;
        }
        check = type_check(y_tmp, x_tmp);
        if (!check)
        {
            return NULL;
        }
        else if (check == 82)
        {
            result = gen_2_00_wrapper(self, args, kwargs);
        }
        else if (check == 37)
        {
            result = gen_2_01_wrapper(self, args, kwargs);
        }
        else
        {
            PyErr_SetString(PyExc_TypeError, "\"Arguments combinations don\'t exist\"");
            return NULL;
        }
        return result;
    }
    PyObject *gen_2_00_wrapper(PyObject *self, PyObject *args, PyObject *kwargs)
    {
        double y;
        int64_t x;
        double Out_0001;
        PyObject *result;
        static char *kwlist[] = {
            "y",
            "x",
            NULL
        };
    
        if (!PyArg_ParseTupleAndKeywords(args, kwargs, "O&O", kwlist, &y, PyDouble_to_Double, &x, PyInt64_to_Int64))
        {   
            return NULL;
        }
        Out_0001 = gen_2_00(y, x);
        result = Py_BuildValue("d", Out_0001);
        return result;
    }
    ```
    where at least the functions are printed in the same way as usual

2. Can you explain why it is impossible to use `ndarrays` with fortran? Is this due to the signature problem?

3. The memory leak problem has a solution which is mentioned in the [python docs](https://docs.python.org/3/c-api/arg.html#other-objects)
    > If the converter returns Py_CLEANUP_SUPPORTED, it may get called a second time if the argument parsing eventually fails, giving the converter a chance to release any memory that it had already allocated. In this second call, the object parameter will be NULL; address will have the same value as in the original call.

pyccel/stdlib/cwrapper/cwrapper.c
```diff
@@ -0,0 +1,648 @@
...
+	return true;
+}
+
+
+
+/*
+** convert numpy strides to nd_array strides, and return it in a new array, to
+** avoid the problem of different implementations of strides in numpy and ndarray.
+*/
+static int64_t	*_numpy_to_ndarray_strides(npy_intp  *np_strides, int type_size, int nd)
```
EmilyBourne, line 490: Fixes #758

pyccel/stdlib/cwrapper/cwrapper.c
```diff
@@ -0,0 +1,648 @@
...
+static bool _check_pyarray_order(PyArrayObject *a, int flag)
+{
+	char	order;
+
+	if (flag == NO_ORDER_CHECK)
+		return true;
+
+	if (!PyArray_CHKFLAGS(a, flag))
+	{
+		order = flag == NPY_ARRAY_C_CONTIGUOUS ? 'C' : 'F';
```
EmilyBourne, line 446: Fixes #806

EmilyBourne:
The only unused concept from this PR is having cast functions so that we could use the `O&` format for [PyArg_ParseTupleAndKeywords](https://docs.python.org/3/c-api/arg.html#other-objects) . This would make the function printing much more compact, but interfaces would still need to be handled differently. It is not a critical change.

PR #775, opened 2021-02-26
Print abs function in C

### LINKED ISSUE #774: Builtin abs function not working in c language
**Describe the bug**
Builtin abs function not implemented in c language

**To Reproduce**
```python
@types('int')
def f(x):
    return abs(x)
```

**Error : **
pyccel:
 |fatal [codegen]: first.py [3,11]| Pyccel has encountered syntax that has not been implemented yet. (<pyccel.ast.builtins.PythonAbs object at 0x7fa62ca792b0>)

**Language**
c

abkhaoula:
Add a printer for Python's native `abs` function to the `CCodePrinter` class. Fixes #774


pyccel/codegen/printing/ccode.py
```diff
@@ -431,6 +431,12 @@ def fill_NumpyArange(self, expr, lhs):
 
     # ============ Elements ============ #
 
+    def _print_PythonAbs(self, expr):
+        """ print the python builtin function abs
+        args : variable
+        """
+        return "abs({})".format(self._print(expr.arg))
```
MenoIy, line 438: Not like python C ```abs``` work only with integer types ```int abs(int x)```
so if you try  : 
```PYTHON
a = abs(2.2)
print(a)
```
you get ```2.000000```
try to use the corresponding absolute value for each type : (abs for int) and (fabs for float) etc ...
you may need to add import  like <math.h>  or  <complex.h> you can do that by using 
same as here https://github.com/pyccel/pyccel/blob/master/pyccel/codegen/printing/ccode.py#L459

MenoIy (changes requested):
Good work, just a few comments : 
- there are some useless changes in the repo you can check by going to File changed above
when you try to push change use ```git status``` to verify maded changes and ```git push <file name> ``` to commit change .

pyccel/codegen/printing/ccode.py
MenoIy: you can get the type of the arguments in ```expr.arguments.dtype``` see  ``` pyccel/ast/datatypes``` for classes definitions

EmilyBourne (changes requested):
There are 2 new files in your PR which are not needed:
- doc/latex_macros.sty
- tests/run_tests_py3.sh

Please add tests for the builtin `abs` function. I think the most logical place for these new tests would be in a new file, e.g:
`tests/epyccel/test_builtins.py`

See [test_kind.py](https://github.com/pyccel/pyccel/blob/master/tests/epyccel/test_kind.py) for an example of a test file. A few rules for test files:
- All test files should start with [this line](https://github.com/pyccel/pyccel/blob/master/tests/epyccel/test_kind.py#L1) (so codacy doesn't complain about missing docstrings in tests)
- The names of all test functions should start with `test_` (so that they are found by [pytest](https://docs.pytest.org/en/stable/getting-started.html))

It is important to add tests to ensure that you have not missed any use cases and also so that anybody making modifications later does not accidentally break something which worked previously

MenoIy (changes requested):
Looks good you just need to add some tests ```pyccel/tests/epyccel/```
And for the empty file changes can you use the command ```git config core.filemode```
if it show true that mean that Git picks up (chmod : permission) changes. so you need to set it as false , you can try https://stackoverflow.com/questions/1580596/how-do-i-make-git-ignore-file-mode-chmod-changes

EmilyBourne:
>if it show true that mean that Git picks up (chmod : permission) changes

Ah! I had not realised that these files were present in the repository already and that this is what the change was. I think it makes sense for `run_tests_py3.sh` to be executable so that you can do:
`./tests/run_tests_py3.sh`
but `doc/latex_macros.sty` shouldn't be executable anyway. So if you want to keep the change to this file then I am happy with that, but it should be mentioned in the PR description

MenoIy (approved):

EmilyBourne (approved):

tests/epyccel/test_builtins.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def test_abs_c(language):
+    @types('complex')
+    def f1(x):
+        return abs(x)
+
+    f2 = epyccel(f1, language=language)
+
+    assert f1(3j+2) == f2(3j+2)
```
yguclu, line 32: And here you can check a few other complex numbers, for example:
```
3 + 4j
3 - 4j
5 + 0j
0 + 5j
0 + 0j
```
The absolute value is 5 for the first four of these, and 0 for the last one.

pyccel/codegen/printing/ccode.py
```diff
@@ -431,6 +431,16 @@ def fill_NumpyArange(self, expr, lhs):
...
 
+    def _print_PythonAbs(self, expr):
+        func = "abs"
+        if expr.arg.dtype is NativeReal():
+            self._additional_imports.add("math")
+            func = "fabs"
+        elif expr.arg.dtype is NativeComplex():
+            self._additional_imports.add("complex")
+            func = "cabs"
+        return "{}({})".format(func, self._print(expr.arg))
```
yguclu, line 442: As a matter of style and readability, it is good habit not to redefine a variable  unless strictly necessary.

Here you are changing the value of `func` if the dtype matches a certain condition, but it is clearer and more natural to use an `else` condition instead:

```python
        if expr.arg.dtype is NativeReal():
            self._additional_imports.add("math")
            func = "fabs"
        elif expr.arg.dtype is NativeComplex():
            func = "cabs"
        else:
            func = "abs"
```

yguclu (changes requested):
Thanks for this PR, @abkhaoula! I have a couple of comments

tests/epyccel/test_builtins.py
```diff
@@ -0,0 +1,32 @@
...
+from pyccel.decorators import types
+
+def test_abs_i(language):
+    @types('int')
+    def f1(x):
+        return abs(x)
+
+    f2 = epyccel(f1, language=language)
+
+    assert f1(11) == f2(11)
```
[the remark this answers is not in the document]
abkhaoula: just fixed it in [1c4b8e6](https://github.com/pyccel/pyccel/pull/775/commits/1c4b8e666ea0c0a126891181115fc620634f86dd)

yguclu (approved):

PR #777, opened 2021-03-01
Fix bugs in numpy.prod

mostafaMamoni:
Currently `Numpy.prod` does not accept scalar arguments or arguments of `bool` type, and it gives wrong results when passing arrays with int8, int16, int32 types. This pull request should fix this:

- Fix scalar arguments bug by converting it to list, and cast the argument if it is a bool;
- Cast `arg` to `int64` in NumpyProd if it is `int` and its precision is less than 8;
- Add tests.

### DIFF
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -343,8 +343,7 @@ class NumpyProduct(PyccelInternalFunction):
             raise TypeError('Unknown type of  %s.' % type(arg))
         super().__init__(arg)
         self._arg = PythonList(arg) if arg.rank == 0 else self._args[0]
-        self._arg = PythonInt(self._arg) if (isinstance(arg.dtype, NativeBool) or \
-                    isinstance(arg.dtype, NativeInteger)) else self._arg
+        self._arg = PythonInt(self._arg) if isinstance(arg.dtype, (NativeInteger, NativeBool)) else self._arg
         self._dtype = self._arg.dtype
         self._rank  = 0
         self._shape = ()
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -343,7 +343,7 @@ class NumpyProduct(PyccelInternalFunction):
             raise TypeError('Unknown type of  %s.' % type(arg))
         super().__init__(arg)
         self._arg = PythonList(arg) if arg.rank == 0 else self._args[0]
-        self._arg = PythonInt(self._arg) if isinstance(arg.dtype, NativeBool) else self._arg
+        self._arg = PythonInt(self._arg) if (isinstance(arg.dtype, NativeBool) or (isinstance(arg.dtype, NativeInteger) and self._arg.precision < 8)) else self._arg
         self._dtype = self._arg.dtype
         self._rank  = 0
         self._shape = ()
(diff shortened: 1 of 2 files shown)


pyccel/ast/numpyext.py
```diff
@@ -342,14 +342,16 @@ def __init__(self, arg):
         if not isinstance(arg, PyccelAstNode):
             raise TypeError('Unknown type of  %s.' % type(arg))
         super().__init__(arg)
-        self._dtype = arg.dtype
+        self._arg = PythonList(arg) if arg.rank == 0 else self._args[0]
+        self._arg = PythonInt(self._arg) if (isinstance(arg.dtype, NativeBool) or (isinstance(arg.dtype, NativeInteger) and self._arg.precision < 8)) else self._arg
```
EmilyBourne, line 346: Could you break this line up please, it barely fits on the page.

EmilyBourne, line 346: > self._arg.precision < 8

Are you sure this condition is correct on windows?

EmilyBourne (changes requested):

pyccel/ast/numpyext.py
mostafaMamoni: i think it is fine since we check the bigger precision in all types, if PythonInt function will create an int32 object, to be honest, i can't think of falling cases, because precision in windows will not pass 8.

EmilyBourne: It is not a failing case that I am worried about but an unnecessary cast. I don't think this causes a bug, I just think it adds an unnecessary node which will slow the code down at some point

EmilyBourne: You seem to have done the opposite of what I am suggesting and are now casting all the time. To be clear my suggested improvement is:
`self._arg.precision < default_precision['int']`

Presumably windows doesn't truncate int64_t objects to int32_t?

EmilyBourne (approved):

nhamidn (approved):

pyccel/ast/numpyext.py
```diff
@@ -342,14 +342,18 @@ def __init__(self, arg):
         if not isinstance(arg, PyccelAstNode):
             raise TypeError('Unknown type of  %s.' % type(arg))
         super().__init__(arg)
-        self._dtype = arg.dtype
+        self._arg = PythonList(arg) if arg.rank == 0 else self._args[0]
+        self._arg = PythonInt(self._arg) if (isinstance(arg.dtype, NativeBool) or \
+                    (isinstance(arg.dtype, NativeInteger) and self._arg.precision < default_precision['int']))\
+                    else self._arg
```
yguclu, line 348: I don't understand what is going on here in the case where `arg` is a scalar`bool`: first you convert it to a `list` and then you call `int()` on the list?

mostafaMamoni: it is not always 'then', if it is a scalar in general we should convert it to list because in Fortran `product` does not accept scalars, so the first block is just about scalar to list, the second one is about casting boolean to int and increase the precision of integers to the default_precision for the sake of a correct result.

EmilyBourne: The confusion here, may be due to a modification made to `PythonInt` and the other casting functions.
When called on an array in python these functions will raise an error, but `PythonInt` will just cast the array to a different type

yguclu: I see. I did not remember that `PythonInt` could be applied to a `PythonList`.

Maybe we should update the docstrings of the casting functions...

yguclu (approved):

pyccel/ast/numpyext.py
mostafaMamoni: Actually, we dont mention which data structure we support in the docstring, but it needs to be more detailed.

PR #779, opened 2021-03-02
Handle bool arguments in numpy.mod 

mostafaMamoni:
Commit summary:
- Create NumpyMod constructor, and cast NumpyMod arguments to int type whenever they are booleans.
- Add tests

### DIFF
--- a/tests/epyccel/recognised_functions/test_numpy_funcs.py
+++ b/tests/epyccel/recognised_functions/test_numpy_funcs.py
@@ -5035,11 +5035,13 @@ def test_numpy_mod_scalar(language):
     assert f_integer32_output == test_int32_output
     assert type(f_integer32_output) == type(test_int32_output.item())
 
-    f_integer64_output = f_integer64(integer64)
-    test_int64_output = get_mod(integer64)
+    # the if block shoud be removed after resolving (https://github.com/pyccel/pyccel/issues/735).
+    if sys.platform != 'win32':
+        f_integer64_output = f_integer64(integer64)
+        test_int64_output = get_mod(integer64)
 
-    assert f_integer64_output == test_int64_output
-    assert type(f_integer64_output) == type(test_int64_output.item())
+        assert f_integer64_output == test_int64_output
+        assert type(f_integer64_output) == type(test_int64_output.item())
 
     f_fl = epyccel(get_mod, language=language)
     f_fl32 = epyccel(get_mod, language=language)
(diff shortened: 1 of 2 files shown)


EmilyBourne (approved):

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -4954,3 +4954,245 @@ def get_complex128(a):
 
     assert f_fl64_output == test_float64_output
     assert type(f_fl64_output) == type(test_float64_output.item())
+
+@pytest.mark.parametrize( 'language', (
+        pytest.param("fortran", marks = [pytest.mark.fortran]),
+        pytest.param("c", marks = [
+            pytest.mark.skip(reason="Mod function not supported in C"),
+            pytest.mark.c]
```
MenoIy, line 4962: is there any problem with C ? I understand that for array it seem more complicated but for scalar type i think its need only a printer. is there any raison why you skip C tests ?

mostafaMamoni: the reason: array it seems more complicated, a printer for scalars is a good idea but is it worth implementing it without arrays?

EmilyBourne: If it is trivial for scalars, and arrays are implemented element-wise then it is also trivial for arrays (see your solution for real/imag)

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -4954,3 +4954,245 @@ def get_complex128(a):
...
+        from numpy import mod, shape
+        a = mod(arr, arr)
+        s = shape(a)
+        return len(s), s[0], a[0]
+
+    bl = randint(0, 1, size=(5), dtype= bool)
+
+    integer8 = randint(min_int8, max_int8, size=(5), dtype=np.int8)
+    integer16 = randint(min_int16, max_int16, size=(5), dtype=np.int16)
+    integer = randint(min_int, max_int, size=(5), dtype=np.int)
```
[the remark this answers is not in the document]
EmilyBourne: Maybe we should open an issue to clean up all our uses of `np.int` in our tests?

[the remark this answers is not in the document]
EmilyBourne: I am happy to leave it in this PR though as there are lots of tests with very similar code and this all needs cleaning up anyway,
e.g. to a function:
```
@pytest.mark.parametrize('func',[get_float, get_int, ...])
def test_all_types_1_arg(self, func, language):
     py_func = epyccel(func, language=language)
     integer8 = randint(min_int8, max_int8, size=(5), dtype=np.int8)
     assert py_func(integer8 ) == func(integer8 )£
     etc
```
Bonus points if we can make dtype an argument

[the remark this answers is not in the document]
mostafaMamoni: yes, it is used a lot

tests/epyccel/recognised_functions/test_numpy_funcs.py
mostafaMamoni: you are right. about the tests i cannot do:
```python
s = shape(a)
return len(s), s[0], a[0]
```
because tuples are not implemented in C yet, So i think it is better to add missing C printers and activate their tests in another PR, since this issue/PR are just about Fortran

EmilyBourne: Ok. Could you open an issue for `NumpyMod` in C please

MenoIy (approved):

yguclu (approved):

PR #780, opened 2021-03-02
Allow negative step in for loops

### LINKED ISSUE #709: For Loop for loop descending order are not working in C  and Fortran
**Describe the bug**
For Loop for loop descending order are not working in C and have some bugs in Fortran

**To Reproduce**
```python
for i in range(10, 1, -1):
    print(i)
```
```Python
>> 10 9 8 7 6 5 4 3 2
```

```C
#include <stdlib.h>
#include <stdio.h>
#include <stdint.h>
int main()
{
    int64_t i;

    for (i = 10; i < 1; i += -1)
    {
        printf("%ld\n", i);
    }
    return 0;
}
```
here the for loop condition should be changed. 
```Fortran
program prog_test

implicit none

integer(C_INT64_T) :: i

do i = 10_C_INT64_T, 1_C_INT64_T-1_C_INT64_T, -1_C_INT64_T
  print *, i
end do

end program prog_test
```
```python
>> 10 9 8 7 6 5 4 3 2 1 0
```
here the end of the for loop is wrong

abkhaoula:
Allow for loops of descending order (with negative step). This fixes #709:

- Fix method `_print_PythonRange` of Fortran printer;
- Fix method `_print_For` of C printer;
- Add tests.

### DIFF
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -1349,9 +1349,14 @@ class CCodePrinter(CodePrinter):
             step  = self._print(expr.iterable.step )
         else:
             raise NotImplementedError("Only iterable currently supported is Range")
-        return ('for ({target} = {start}; {target} < {stop}; {target} += '
+        if (step[0] == "-"):
+            return ('for ({target} = {start}; {target} > {stop}; {target} += '
                 '{step})\n{{\n{body}\n}}').format(target=target, start=start,
                 stop=stop, step=step, body=body)
+        else:
+            return ('for ({target} = {start}; {target} < {stop}; {target} += '
+                    '{step})\n{{\n{body}\n}}').format(target=target, start=start,
+                                                      stop=stop, step=step, body=body)
 
     def _print_CodeBlock(self, expr):
         body_exprs, new_vars = expand_to_loops(expr, self._parser.get_new_variable, language_has_vectors = False)
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -528,6 +528,7 @@ class CCodePrinter(CodePrinter):
         cond = self._print(expr.cond)
         value_true = self._print(expr.value_true)
         value_false = self._print(expr.value_false)
+        print('here is fine')
         return '{cond} ? {true} : {false}'.format(cond = cond, true =value_true, false = value_false)
 
     def _print_LiteralTrue(self, expr):
@@ -1349,9 +1350,26 @@ class CCodePrinter(CodePrinter):
             step  = self._print(expr.iterable.step )
         else:
             raise NotImplementedError("Only iterable currently supported is Range")
-        return ('for ({target} = {start}; (({step} > 0 ? 1 : -1)*({target})) < (({step} > 0 ? 1 : -1)*({stop})); {target} += '
+
+        test_step = expr.iterable.step
+        if isinstance(test_step, PyccelUnarySub):
+            test_step = expr.iterable.step.args[0]
+
+        # testing if the step is a value or an expression
+        if isinstance(test_step, Literal):
+            if isinstance(expr.iterable.step, PyccelUnarySub):
+                return ('for ({target} = {start}; {target} > {stop}; {target} += '
+                        '{step})\n{{\n{body}\n}}').format(target=target, start=start,
+                                                          stop=stop, step=step, body=body)
+            else:
+                return ('for ({target} = {start}; {target} < {stop}; {target} += '
+                        '{step})\n{{\n{body}\n}}').format(target=target, start=start,
+                                                          stop=stop, step=step, body=body)
+        else:
+            return (
+                'for ({target} = {start}; (({step} > 0 ? 1 : -1)*({target})) < (({step} > 0 ? 1 : -1)*({stop})); {target} += '
                 '{step})\n{{\n{body}\n}}').format(target=target, start=start,
-                stop=stop, step=step, body=body)
+                                                  stop=stop, step=step, body=body)
 
     def _print_CodeBlock(self, expr):
         body_exprs, new_vars = expand_to_loops(expr, self._parser.get_new_variable, language_has_vectors = False)
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -1350,6 +1350,7 @@ class CCodePrinter(CodePrinter):
         else:
             raise NotImplementedError("Only iterable currently supported is Range")
 
+        test_step = None
         if isinstance(test_step, PyccelUnarySub):
             test_step = expr.iterable.step.args[0]
         else:
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -1317,8 +1317,7 @@ class CCodePrinter(CodePrinter):
         return '{} = {};'.format(lhs, rhs)
 
     def _print_For(self, expr):
-        loop_counter = self._print(expr.target)
-        loop_counter = self._print(expr.target)
+        counter = self._print(expr.target)
         body  = self._print(expr.body)
         if isinstance(expr.iterable, PythonRange):
             start = self._print(expr.iterable.start)
@@ -1334,13 +1333,13 @@ class CCodePrinter(CodePrinter):
         # testing if the step is a value or an expression
         if isinstance(test_step, Literal):
             op = '>' if isinstance(expr.iterable.step, PyccelUnarySub) else '<'
-            return ('for ({loop_counter} = {start}; {loop_counter} {op} {stop}; {loop_counter} += '
-                        '{step})\n{{\n{body}\n}}').format(loop_counter=loop_counter, start=start, op=op,
+            return ('for ({counter} = {start}; {counter} {op} {stop}; {counter} += '
+                        '{step})\n{{\n{body}\n}}').format(counter=counter, start=start, op=op,
                                                           stop=stop, step=step, body=body)
         else:
             return (
-                'for ({loop_counter} = {start}; ({step} > 0) ? ({loop_counter} < {stop}) : ({loop_counter} > {stop}); {loop_counter} += '
-                '{step})\n{{\n{body}\n}}').format(loop_counter=loop_counter, start=start,
+                'for ({counter} = {start}; ({step} > 0) ? ({counter} < {stop}) : ({counter} > {stop}); {counter} += '
+                '{step})\n{{\n{body}\n}}').format(counter=counter, start=start,
                                                   stop=stop, step=step, body=body)
 
     def _print_CodeBlock(self, expr):
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -1317,7 +1317,8 @@ class CCodePrinter(CodePrinter):
         return '{} = {};'.format(lhs, rhs)
 
     def _print_For(self, expr):
-        target = self._print(expr.target)
+        loop_counter = self._print(expr.target)
+        loop_counter = self._print(expr.target)
         body  = self._print(expr.body)
         if isinstance(expr.iterable, PythonRange):
             start = self._print(expr.iterable.start)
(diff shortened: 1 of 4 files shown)


pyccel/codegen/printing/ccode.py
```diff
@@ -1349,9 +1349,14 @@ def _print_For(self, expr):
             step  = self._print(expr.iterable.step )
         else:
             raise NotImplementedError("Only iterable currently supported is Range")
-        return ('for ({target} = {start}; {target} < {stop}; {target} += '
+        if (step[0] == "-"):
```
EmilyBourne, line 1352: A better way to test this would be `if isinstance(step, PyccelUnarySub)` that way if the printer is modified for example to add spaces the index isn't invalidated

EmilyBourne (changes requested):
Please add tests 😉

pyccel/codegen/printing/ccode.py
```diff
@@ -1349,7 +1349,7 @@ def _print_For(self, expr):
             step  = self._print(expr.iterable.step )
         else:
             raise NotImplementedError("Only iterable currently supported is Range")
-        return ('for ({target} = {start}; {target} < {stop}; {target} += '
+        return ('for ({target} = {start}; (({step} > 0 ? 1 : -1)*({target})) < (({step} > 0 ? 1 : -1)*({stop})); {target} += '
```
MenoIy, line 1352: you can break this statement into 2 case:
- case 1 : target is known (Literal)
     in this case we can  know it target is positive or negative, so additional check are not needed in the generated code
     like you were doing in a4258ad
- case 2 : target is unknown (Variable)
     in this case you can keep what you did here
     you can improve printed code by using IfTernaryOperator  see ast/operators 
     ```
     {step} > 0 ? 1 : -1
     ```
     is equivalent to ```IfTernaryOperator(PyccelGt(step, 0), 1, -1)```
     or more better ```IfTernaryOperator(PyccelGt(step, 0), target, PyccelUnarySub(target))```
     otherwise you can keep what you did.

pyccel/codegen/printing/fcode.py
```diff
@@ -1787,8 +1787,8 @@ def _print_AugAssign(self, expr):
 
     def _print_PythonRange(self, expr):
         start = self._print(expr.start)
-        stop  = self._print(expr.stop) + '-' + self._print(LiteralInteger(1))
         step  = self._print(expr.step)
+        stop = self._print(expr.stop) + '- ( (merge(1_C_INT64_T, -1_C_INT64_T, ' + self._print(expr.step) + ' > 0_C_INT64_T)) *' + self._print(LiteralInteger(1)) + ')'
```
MenoIy, line 1791: you can use``` self._print(LiteralInteger(1)``` instead of ```1_C_INT64_T```
                   ``` self._print(LiteralInteger(-1)``` instead of ```-1_C_INT64_T```
and let the printed precision  depend on the system used .

like in my previous comment you should try to use PyccelNods , they make code look clearer and readable.

you can use formatted code to clean the generated code example ```hello {1}'.format('world')```

tests/epyccel/modules/loops.py
```diff
@@ -215,3 +215,10 @@ def while_not_0( n ):
     while n:
         n -= 1
     return n
+
+@types( int, int, int )
+def for_loop(start, stop, step):
+    x = 0
+    for i in range(start, stop, step):
+        x += i
```
MenoIy, line 223: can you add a single test where loop step is known ?

MenoIy (changes requested):

pyccel/codegen/printing/fcode.py
```diff
@@ -1787,8 +1787,20 @@ def _print_AugAssign(self, expr):
...
+        if isinstance(test_step, PyccelUnarySub):
+            test_step = expr.step.args[0]
+
+        # testing if the step is a value or an expression
+        if isinstance(test_step, Literal):
+            stop = self._print(expr.stop) + '-' + self._print(LiteralInteger(1))
+            if isinstance(expr.step, PyccelUnarySub):
+                stop = self._print(expr.stop) + '+' + self._print(LiteralInteger(1))
+        else:
+            stop = self._print(expr.stop) + '- ( (merge('+self._print(LiteralInteger(1))+', '+self._print(LiteralInteger(-1))+', ' + self._print(expr.step) + ' > '+ self._print(LiteralInteger(0)) +')) *' + self._print(LiteralInteger(1)) + ')'
```
MenoIy, line 1802: this can be approved to look more readable  using PyccelNodes
```PYTHON
stop = self._print(expr.stop) + '-' + self._print(LiteralInteger(1))
```
can be 
```PYTHON
stop = PyccelAdd(expr.stop, LiteralInteger(1))
```
the same for the next condition
```PYTHON
stop = PyccelMinus(expr.stop, LiteralInteger(1))
```
and then you do 
```PYTHON
stop = self._print(stop)
```

MenoIy: for Unknown  expression : 
```PTHON
            stop = IfTernaryOperator(PyccelGt(expr.step, LiteralInteger(0)), 
                PyccelMinus(expr.stop, LiteralInteger(1)),
                PyccelAdd(expr.stop, LiteralInteger(1)))
            stop = self._print(stop)

``` 
you can check operators/IfternaryOperator  docstring I let you do the same for C

MenoIy (changes requested):

pyccel/codegen/printing/ccode.py
```diff
@@ -1349,9 +1349,22 @@ def _print_For(self, expr):
             step  = self._print(expr.iterable.step )
         else:
             raise NotImplementedError("Only iterable currently supported is Range")
-        return ('for ({target} = {start}; {target} < {stop}; {target} += '
+
+        test_step = expr.iterable.step
+        if isinstance(test_step, PyccelUnarySub):
+            test_step = expr.iterable.step.args[0]
```
EmilyBourne, line 1355: Why not use an if/else block rather than setting `test_step` twice?

pyccel/codegen/printing/fcode.py
```diff
@@ -1787,8 +1787,23 @@ def _print_AugAssign(self, expr):
...
+
+        test_step = expr.step
+        if isinstance(test_step, PyccelUnarySub):
+            test_step = expr.step.args[0]
+
+        # testing if the step is a value or an expression
+        if isinstance(test_step, Literal):
+            stop = PyccelMinus(expr.stop, LiteralInteger(1))
+            if isinstance(expr.step, PyccelUnarySub):
+                stop = PyccelAdd(expr.stop, LiteralInteger(1))
```
EmilyBourne, line 1800: See above, why not use an if/else rather than setting stop twice?

pyccel/codegen/printing/ccode.py
abkhaoula: just fixed it in [cec0ecc](https://github.com/pyccel/pyccel/pull/780/commits/cec0ecc4ab3d036b396f014715bbe158e3f12a44)

pyccel/codegen/printing/fcode.py
abkhaoula: Here too is fixed in [cec0ecc](https://github.com/pyccel/pyccel/pull/780/commits/cec0ecc4ab3d036b396f014715bbe158e3f12a44)

pyccel/codegen/printing/ccode.py
EmilyBourne: Ah sorry I didn't notice that `test_step` was used in the if condition

yguclu: @abkhaoula What Emily meant with her last message is that your original code was correct. Even after b4350b7 the behavior is wrong because you set `test_step` to `None` and then check its type, which does not make sense. So you should change the code back to what it was after commit 29128dc.

yguclu: This is one of those cases where it is OK to set a variable twice! :wink:

abkhaoula: right make sense. thank you 😅

yguclu: You are welcome!

MenoIy (approved):

EmilyBourne (approved):

pyccel/codegen/printing/ccode.py
```diff
@@ -1325,9 +1325,22 @@ def _print_For(self, expr):
...
+            op = '>' if isinstance(expr.iterable.step, PyccelUnarySub) else '<'
+            return ('for ({target} = {start}; {target} {op} {stop}; {target} += '
+                        '{step})\n{{\n{body}\n}}').format(target=target, start=start, op=op,
+                                                          stop=stop, step=step, body=body)
+        else:
+            return (
+                'for ({target} = {start}; (({step} > 0 ? 1 : -1)*({target})) < (({step} > 0 ? 1 : -1)*({stop})); {target} += '
                 '{step})\n{{\n{body}\n}}').format(target=target, start=start,
-                stop=stop, step=step, body=body)
+                                                  stop=stop, step=step, body=body)
```
yguclu, line 1343: Is it really needed to print this complicated exit condition? I think that this will work in C:
```c
for (i = start; step > 0 ? (i < stop) : (i > stop); i += step)
```

yguclu: By the way, can we use the name `index` instead of `target`?

yguclu (changes requested):

pyccel/codegen/printing/ccode.py
abkhaoula: yes, I ll do that and I ll edit the for loop for C

abkhaoula: About the target variable I saw that [here](https://github.com/pyccel/pyccel/blob/2db510aef750bcfd59c61e62f400edbdd1e616bf/pyccel/ast/core.py#L1297-L1308) we defined the index iterator as `target` that is why I saw it was better to stay aligned with the same vocabulary. Do I still change the name from `target`  to `index`

yguclu: Actually I keep asking for using `index` instead of `target` every time I can.

I am not 100% sure about the `For` class is `ast.core`, but here `target` is an index and so the name choice is obvious to me.

yguclu: To be totally rigorous, we may want to call this "loop counter". But definitely not "target".

EmilyBourne: In general `target` is not necessarily an `index`. E.g:
```
for v in arr:
```
The target is `v`. I'm not sure `target` is the clearest name, but I'm not sure `index` is more accurate generally. Here it is accurate as we are handling a `PythonRange`

yguclu: To summarize, I think the best name here is `counter`. That should make everybody happy! :wink:

abkhaoula: Understood

abkhaoula: done in [251cfc2](https://github.com/pyccel/pyccel/pull/780/commits/251cfc2b9b43bb5df1b23f1d6b02d6f007a6f26f)

tests/epyccel/modules/loops.py
```diff
@@ -215,3 +215,16 @@ def while_not_0( n ):
...
+    return x
+
+def for_loop2():
+    x = 0
+    for i in range(1, 10, -1):
+        x += i
+    return x
```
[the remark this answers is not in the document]
abkhaoula: done in [251cfc2](https://github.com/pyccel/pyccel/pull/780/commits/251cfc2b9b43bb5df1b23f1d6b02d6f007a6f26f)

EmilyBourne:
@nhamidn this PR and PR #796 are both failing due to the same problem in Open mp. Do you have any idea what's causing the problem? I'm not sure why it only started appearing now

pyccel/codegen/printing/ccode.py
yguclu: Thanks @abkhaoula. But have you changed the loop condition as I suggested at the beginning of this discussion? It does not appear so. As a reminder:
```python
for (counter = start; step > 0 ? (counter < stop) : (counter > stop); counter += step)
```

yguclu (changes requested):
Please address the comment I made 5 days ago. I posted a reminder two days ago...

abkhaoula:
> Please address the comment I made 5 days ago. I posted a reminder two days ago...

I am Deeply sorry, I didn't notice it

yguclu (approved):

PR #783, opened 2021-03-02
Complete python printer to enable epyccel tests

EmilyBourne:
Add print methods to the python printer to allow the tests in `tests/epyccel` to be enabled. Enable all printers except:
- Any that require ValuedVariable (default values or optional arguments)
- Openmp tests
- test_epyccel_decorators.py (Expected to raise NotImplementedError, but this problem is not present in python)

**Commit Summary**
- Add set to iterable types
- [BUG_FIX] Correct is_atomic definition
- [BUG_FIX] Correct `FunctionCall.interface_name`
- Save import targets in a set to avoid importing the same object multiple times
- Correct python output filename
- Remove unnecessary check
- Collect numpy object functions for imports
- Add `python` to epyccel language tests
- Complete python printer so (most) epyccel tests run
- Create a `matching_types` function for checking result types in `test_numpy_funcs.py`
- Mark tests as skip or xfail as appropriate
- Add type check showing issue #784


EmilyBourne:
The speed of the python tests shows that the bottlenecks in pyccel come from one of the following places:
- cwrapper
- compiling

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5247,25 +5365,31 @@ def get_complex128(a):
...
-    assert type(f_fl32_output) == type(test_float32_output.item())
+    assert matching_types(f_fl32_output, test_float32_output)
 
     f_fl64_output = f_fl64(fl64)
     test_float64_output = get_complex128(fl64)
 
     assert f_fl64_output == test_float64_output
-    assert type(f_fl64_output) == type(test_float64_output.item())
+    assert matching_types(f_fl64_output, test_float64_output)
+    assert matching_types(f_fl64_output, test_float64_output)
```
MenoIy, line 5381: same test

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5560,7 +5699,7 @@ def get_prod(a):
     assert f_bl_true_output == test_bool_true_output
     assert f_bl_false_output == test_bool_false_output
 
-    assert (type(f_bl_true_output) == type(test_bool_false_output.item()))
+    assert matching_types(f_bl_true_output, test_bool_false_output)
```
MenoIy, line 5702: this test should compare the same function type

MenoIy (commented):
Good job !

pyccel/ast/basic.py
```diff
@@ -232,7 +232,7 @@ def prepare_sub(found_node):
     def is_atomic(self):
         """ Indicates whether the object has any attribute nodes
         """
-        return bool(self._attribute_nodes)
+        return not bool(self._attribute_nodes)
```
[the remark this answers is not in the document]
EmilyBourne: `is_atomic` means that the object has no attribute nodes. It is an "atom". So it means that `len(self._attribute_nodes)==0` which is the opposite of what it did before

tests/epyccel/recognised_functions/test_numpy_funcs.py
EmilyBourne: Fixed 4a03f72

tests/epyccel/recognised_functions/test_numpy_funcs.py
EmilyBourne: Good catch .
Fixed 1ca0f0d

pyccel/ast/basic.py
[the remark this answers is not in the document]
EmilyBourne: I have clarified the docstring 0877bb8

MenoIy (approved):

.github/workflows/Github_pytest.yml
```diff
@@ -60,7 +60,11 @@ jobs:
         working-directory: ./tests
       - name: Serial fortran tests and non-tagged tests
         run: |
-          python -m pytest -rx -m "not parallel and not c" --ignore=printers --ignore=symbolic --ignore=ndarrays
+          python -m pytest -rx -m "not parallel and not c and not python" --ignore=printers --ignore=symbolic --ignore=ndarrays
```
yguclu, line 63: I've just checked, one can write `not (c or python)`

.github/workflows/master.yml
```diff
@@ -62,7 +62,11 @@ jobs:
         working-directory: ./tests
       - name: Serial fortran tests and non-tagged tests
         run: |
-          python -m pytest -rx -m "not parallel and not c" --ignore=printers --ignore=symbolic --ignore=ndarrays
+          python -m pytest -rx -m "not parallel and not c and not python" --ignore=printers --ignore=symbolic --ignore=ndarrays
```
yguclu, line 65: Same here: one can write `not (c or python)`

pyccel/ast/numpyext.py
```diff
@@ -652,8 +656,9 @@ def __init__(self, shape, fill_value, dtype=None, order='C'):
 
         # Cast fill_value to correct type
         if fill_value and not isinstance(fill_value, Nil):
```
yguclu, line 658: Could we provide `Nil` with a `__bool__` magic method? If so we would just write
```python
if fill_value:
```

.github/workflows/Github_pytest.yml
EmilyBourne: Done in 4d42e8d

.github/workflows/master.yml
EmilyBourne: Done in 4d42e8d

pyccel/ast/numpyext.py
EmilyBourne: It already has one. This ought to have been changed when it was added I think. Fixed in 47a2297

tests/epyccel/test_epyccel_division.py
```diff
@@ -144,7 +144,7 @@ def fdiv_i_i(x, y):
     y = randint(low=1, high= 100, dtype='int8')
 
     assert (f(x, y) == fdiv_i_i(x, y))
-    assert isinstance(f(x, y), type(fdiv_i_i(x, y).item()))
+    assert isinstance(f(x, y), (type(fdiv_i_i(x, y).item()), type(fdiv_i_i(x, y))))
```
yguclu, line 147: Why do we need to use the `item` method here?

EmilyBourne: Pyccel cannot return numpy types (see #735 ) so we cannot check if the result type matches exactly because it won't. The best we can do is check that the underlying type (returned by `.item()`) matches.
We also need to check against the correct type as python returns the correct type and `isinstance(np.int8(2), int)` returns `False`
Here `type(fdiv_i_i(x, y).item())` returns `int` and `type(fdiv_i_i(x, y))` returns `np.int8`

yguclu: Thanks! I forgot that #735 is still open...

pyccel/codegen/printing/pycode.py
```diff
@@ -32,6 +35,12 @@ def _construct_header(func_name, args):
     return pattern.format(name=func_name, args=args)
 
 #==============================================================================
+
+import_target_swap = {
+        'numpy' : {'double' : 'float64', 'product' : 'prod'},
+        'numpy.random' : {'random' : 'rand'}
+        }
```
yguclu, line 42: Could you explain how you use this dictionary? A line of comment may be useful here.

EmilyBourne: It maps objects to equivalent objects. For example:
```
In [2]: from numpy import double

In [3]: a = double(4)

In [4]: print(type(a))
<class 'numpy.float64'>
```
`numpy.double` is just an alias for `numpy.float64`, internally we will use `NumpyFloat64`, so we would print:
```
from numpy import double
a = float64(4)
```
The import is collected from the user's code, but the function call is printed from our class. The dictionary therefore maps imported objects to the alias which is used internally in pyccel.
I will try to add a clear comment.

EmilyBourne: Fixed in b33d098

yguclu (approved):

pyccel/codegen/printing/pycode.py
yguclu: Thanks!

PR #787, opened 2021-03-04
Bugfix: Correct return type of `float` with boolean argument

### LINKED ISSUE #784: Booleans cast to floats give integers
**Describe the bug**
When casting a boolean to a float the result has the wrong type

**To Reproduce**
```python
def f(a : bool):
    return float(a)
```

```bash/fortran/c
module tmp2

implicit none

contains

function f(a) result(Out_0001)

  implicit none

  integer(C_INT64_T) :: Out_0001
  logical(C_BOOL), value :: a

  Out_0001 = MERGE(1_C_INT64_T, 0_C_INT64_T, a)
  return

end function f

end module tmp2
```

cc: @mostafaMamoni

EmilyBourne:
Fixes #784

### DIFF
--- a/pyccel/ast/builtins.py
+++ b/pyccel/ast/builtins.py
@@ -260,8 +260,6 @@ class PythonFloat(PyccelAstNode):
     _attribute_nodes = ('_arg',)
 
     def __new__(cls, arg):
-        if isinstance(arg.dtype, NativeBool):
-            return PythonInt(arg)
         if isinstance(arg, LiteralFloat) and arg.precision == cls._precision:
             return arg
         if isinstance(arg, (LiteralInteger, LiteralFloat)):
--- a/pyccel/codegen/printing/fcode.py
+++ b/pyccel/codegen/printing/fcode.py
@@ -802,9 +802,6 @@ class FCodePrinter(CodePrinter):
         prec_code = self._print(prec)
         return 'floor({}, kind={})'.format(arg_code, prec_code)
 
-    def _print_Real(self, expr):
-        return expr.fprint(self._print)
-
     def _print_PythonComplex(self, expr):
         if expr.is_cast:
             code = 'cmplx({0}, kind={1})'.format(expr.internal_var,
--- a/pyccel/codegen/printing/fcode.py
+++ b/pyccel/codegen/printing/fcode.py
@@ -646,8 +646,6 @@ class FCodePrinter(CodePrinter):
     def _print_PythonReal(self, expr):
         value = self._print(expr.internal_var)
         return 'real({0})'.format(value)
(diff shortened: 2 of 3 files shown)


pyccel/ast/builtins.py
```diff
@@ -260,8 +260,6 @@ class PythonFloat(PyccelAstNode):
     _attribute_nodes = ('_arg',)
 
     def __new__(cls, arg):
-        if isinstance(arg.dtype, NativeBool):
-            return PythonInt(arg)
```
mostafaMamoni, line 264: i was thinking about changing the dtype of the class here

EmilyBourne: So that a `PythonInt` returns a float? That sounds like a documentation nightmare to me

mostafaMamoni: I mean changing the dtype instead of casting, but i think we should not modify the object, so it is fase to use it in the printer.

EmilyBourne: Ah, so:
```
def __new__(cls, arg):
    arg.dtype=float
    return arg
```
2 counter examples:
Example 1:
```
a = 2
b = float(a)
```
You don't want to change the dtype of the `Variable` a

Example 2:
```
c = float(a+b)
```
if `a+b` is a complex and you just change the type of `a+b` then you will print:
```
float c = a + b
```
and you will have warnings about truncation

mostafaMamoni: it will look weird but here is what i meant:
```python
            cls._dtype = NativeFloat()
```

mostafaMamoni (approved):

pyccel/codegen/printing/fcode.py
```diff
@@ -616,8 +616,6 @@ def _print_PythonSum(self, expr):
     def _print_PythonReal(self, expr):
         value = self._print(expr.internal_var)
         return 'real({0})'.format(value)
-    def _print_PythonFloat(self, expr):
-        return expr.fprint(self._print)
```
yguclu, line 620: We had two methods with the same name?? 😱

yguclu: By the way, I've just noticed that we have a `_print_Real` method... Do you know what it's used for?

EmilyBourne: Yes, apparently!

Hmm, you are right. I suspect it used to print `NumpyReal` (which prints the real part of a complex number) before it was renamed. Codacy confirms that it is unused so I have deleted it. See 7177ad1

yguclu: Thanks!

yguclu (approved):

PR #788, opened 2021-03-04
Handle OpenMP differences between C and Fortran

### LINKED ISSUE #589: NOWAIT Clause not handled for Fortran
**Describe the bug**
We allow the User to specify the `nowait` at the beginning of the `Loop Construct`, which will be a problem for `Fortran` and not for `C`, given that we don't use the `#$ omp end for` pragma anymore, and because the `nowait` clause must be printed at the end of the `Loop Construct`.
We must add the `#$ omp end for` pragma for `Fortran` when the `nowait` clause is specified.

For example this python code : 

```python
result = 0
#$ omp parallel num_threads(4) shared(result)
#$ omp for nowait
for i in range(0, 1337):
  #$ omp critical(test)
  result += i
  #$ omp end critical(test)
#$ omp end parallel
```

Must generate this Fortran code
```fortran
result = 0_C_INT64_T
!$omp parallel num_threads(4) shared(result)
!$omp do
do i = 0_C_INT64_T, 1337_C_INT64_T-1_C_INT64_T, 1_C_INT64_T
  !$omp critical
  result = result + i
  !$omp end critical
end do
!$omp do nowait
!$omp end parallel
```

**Language**
Fortran

nhamidn:
Handle some special cases in OpenMP that differ from C to Fortran. Fix #589:

- Add `nowait` clause after the for loop in Fortran in the case of `for`, `single` or `sections` construct;
- Add `simd` construct as a combination to the `for` construct;
- Add tests for the `nowait` clause;
- Raise error when there is no for loop after a worksharing-loop construct;
- Add error tests.


pyccel/ast/core.py
```diff
@@ -1346,8 +1346,16 @@ def __init__(
...
+        self._nowait_expr = None
         super().__init__()
 
+    @property
+    def nowait_expr(self):
+        return self._nowait_expr
+
+    def set_nowait_expr(self, expr):
+        self._nowait_expr = expr
+
```
MenoIy, line 1358: can't you use setter decorator here ?
```Python
@nowait_expr.setter
```

pyccel/parser/semantic.py
```diff
@@ -853,8 +859,45 @@ def _visit_CommentBlock(self, expr, **settings):
         return expr
     def _visit_AnnotatedComment(self, expr, **settings):
         return expr
+
     def _visit_OmpAnnotatedComment(self, expr, **settings):
+        code = expr._user_nodes
+        code = code[len(code) - 1]
```
MenoIy, line 865: if you wanna last element you can do ```code[-1]```

pyccel/parser/semantic.py
```diff
@@ -853,8 +859,45 @@ def _visit_CommentBlock(self, expr, **settings):
...
+        if isinstance(expr, (OMP_For_Loop, OMP_Simd_Construct,
+                            OMP_Distribute_Construct, OMP_TaskLoop_Construct)) or combined_loop:
+            msg = "Statement after {} must be a for loop.".format(type(expr).__name__)
+            if index == (len(code.body) - 1):
+                errors.report(msg, symbol=type(expr).__name__,
+                severity='fatal', blocker=self.blocking)
+            for i, node in enumerate(code.body):
+                if i == index + 1:
+                    if isinstance(node, For):
+                        print(expr.txt.startswith(' simd'))
```
MenoIy, line 886: trailing print

nhamidn: oups

pyccel/ast/core.py
nhamidn: when i use the setter decorator i get this error : `TypeError: 'NoneType' object is not callable`

nhamidn: fixed

MenoIy (approved):

pyccel/parser/semantic.py
```diff
@@ -853,8 +859,44 @@ def _visit_CommentBlock(self, expr, **settings):
...
     def _visit_OmpAnnotatedComment(self, expr, **settings):
+        code = expr._user_nodes
+        code = code[-1]
+        index = code.body.index(expr)
+        combined_loop = expr.combined and ('for' in expr.combined or 'distribute' in expr.combined or 'taskloop' in expr.combined)
+
+        if isinstance(expr, (OMP_Sections_Construct, OMP_Single_Construct)) \
+           and expr.has_nowait:
+            for i, node in enumerate(code.body):
+                if i > index:
```
EmilyBourne, line 872: Couldn't this be simplified:
```
for i, node in enumerate(code.body[index+1:], index + 1):
```
?

pyccel/parser/semantic.py
```diff
@@ -853,8 +859,44 @@ def _visit_CommentBlock(self, expr, **settings):
...
+                            node.has_nowait = True
+
+        if isinstance(expr, (OMP_For_Loop, OMP_Simd_Construct,
+                            OMP_Distribute_Construct, OMP_TaskLoop_Construct)) or combined_loop:
+            msg = "Statement after {} must be a for loop.".format(type(expr).__name__)
+            if index == (len(code.body) - 1):
+                errors.report(msg, symbol=type(expr).__name__,
+                severity='fatal', blocker=self.blocking)
+            for i, node in enumerate(code.body):
+                if i == index + 1:
```
EmilyBourne, line 884: Why is the for loop necessary here? Can't you just do:
```
if len(code.body) > index + 1:
    i = index + 1
```
?

EmilyBourne: Ah sorry, I missed:
```
elif isinstance(node, (Comment, CommentBlock, Pass)):
                        index += 1
```

EmilyBourne: But I still think this could be simplified with a `while` block or at least by only starting the for loop at index + 1

pyccel/parser/semantic.py
```diff
@@ -853,8 +859,46 @@ def _visit_CommentBlock(self, expr, **settings):
...
+                if isinstance(node, Omp_End_Clause):
+                    if node.txt.startswith(expr.name, 4):
+                        node.has_nowait = True
+
+        if isinstance(expr, (OMP_For_Loop, OMP_Simd_Construct,
+                            OMP_Distribute_Construct, OMP_TaskLoop_Construct)) or combined_loop:
+            msg = "Statement after {} must be a for loop.".format(type(expr).__name__)
+            if index == (len(code.body) - 1):
+                errors.report(msg, symbol=type(expr).__name__,
+                severity='fatal', blocker=self.blocking)
```
EmilyBourne, line 881: Should this condition be `if index >= (len(code.body)-1)`?

nhamidn: im checking if the OpenMP expr is  at the end of the parent node, so I don't think there a case where `index > len(code.body)`

pyccel/parser/semantic.py
```diff
@@ -853,8 +859,46 @@ def _visit_CommentBlock(self, expr, **settings):
...
+
+            index += 1
+            while index < len(code.body):
```
EmilyBourne, line 884: Some people consider it bad style to use `break`s. I think this is pretty clear but I wonder if the following would be clearer:
```
index += 1
while isinstance(code.body[index], (Comment, CommentBlock, Pass)) and index < len(code.body):
    index += 1

if index < len(code.body) and isinstance(code.body[index], For):
    if expr.has_nowait:
        nowait_expr = '!$omp end do'
        if expr.txt.startswith(' simd'):
            nowait_expr += ' simd'
        nowait_expr += ' nowait\n'
        code.body[index].nowait_expr = nowait_expr
else:
    errors.report(msg, symbol=type(code.body[index]).__name__,
        severity='fatal', blocker=self.blocking)
```
what do you think?

pyccel/parser/semantic.py
```diff
@@ -853,8 +859,46 @@ def _visit_CommentBlock(self, expr, **settings):
...
+                if isinstance(code.body[index], For):
+                    if expr.has_nowait:
+                        nowait_expr = '!$omp end do'
+                        if expr.txt.startswith(' simd'):
+                            nowait_expr += ' simd'
+                        nowait_expr += ' nowait\n'
+                        code.body[index].nowait_expr = nowait_expr
+                    break
+                elif isinstance(code.body[index], (Comment, CommentBlock, Pass)):
+                    index += 1
```
EmilyBourne, line 894: Why is `pass` an acceptable statement after an omp pragma? Can you give me an example?

pyccel/parser/semantic.py
```diff
@@ -2054,7 +2098,9 @@ def _visit_For(self, expr, **settings):
         if isinstance(iterable, Variable):
             return ForIterator(target, iterable, body)
 
-        return For(target, iterable, body, local_vars=local_vars)
+        for_expr = For(target, iterable, body, local_vars=local_vars)
+        for_expr.nowait_expr = expr.nowait_expr
```
EmilyBourne, line 2102: Why not make this a constructor argument then you don't need the setter?

pyccel/parser/syntax/openmp.py
```diff
@@ -83,15 +90,22 @@ def check_get_clauses(name, valid_clauses, clauses, combined = None):
     """
     Function to check if the clauses are correct for a given construct.
```
EmilyBourne, line 91: Does this docstring need updating?

pyccel/parser/semantic.py
nhamidn: because in pass stmt, we print a comment in the generated code, and in python this code:
```
def test():
    pass
    print("hello world")

test()
```
give this : 
`hello world`

so i think we don't need to raise error when there is a `pass` statement as we convert it to a comment.

EmilyBourne: You're right this behaviour that is easy to replicate even if a user who writes:
```
#$ omp for nowait schedule(runtime)
pass
for i in range(0, n):
    for j in range(0, m):
        b[i,j] = i-j
```
is definitely doing something strange. That doesn't mean that we have to make it our problem 😄

pyccel/parser/semantic.py
nhamidn: yes, this is good

pyccel/parser/semantic.py
nhamidn: i use the setter also in the line 893

pyccel/parser/semantic.py
```diff
@@ -853,8 +859,44 @@ def _visit_CommentBlock(self, expr, **settings):
         return expr
     def _visit_AnnotatedComment(self, expr, **settings):
         return expr
+
     def _visit_OmpAnnotatedComment(self, expr, **settings):
+        code = expr._user_nodes
```
EmilyBourne, line 864: I thought codacy usually complained about this kind of statement. Is there no public method letting you get this?

EmilyBourne (approved):

yguclu (commented):
Good job @nhamidn! Is the documentation already up to date?

pyccel/parser/semantic.py
nhamidn: i don't think there a method for this

nhamidn:
> Good job @nhamidn! Is the documentation already up to date?

the documentation didn't mention the clause, but i think we need to mention those special cases, I'm gonna add them to the documentation.

yguclu (approved):

Issue #604, opened 2020-12-18
Create option to generate Fortran code with default integers


Beliavsky:
Python integers are translated to integer(C_LONG_LONG), which are integers with a large range. However, the Fortran code is cluttered with _C_LONG_LONG after all integer constants, which makes it less readable. Could you
(1) add an option where Python integers are mapped to default Fortran integers?
(2) and/or rename C_LONG_LONG with 

`use ISO_C_BINDING, only: i64 => C_LONG_LONG`

and change the generated Fortran code accordingly?

saidctb:
agree, we should probably rename  the iso_c_binding types to make them more readable.
 what do you think @yguclu  and @EmilyBourne ?

EmilyBourne:
You are right.  This was discussed [here](https://github.com/pyccel/pyccel/issues/487#issuecomment-717852036) but we never created an issue for it so it was forgotten. The main difficulty comes from ensuring that the names are not used by the user's variables. This is linked to issue #227

saidctb:
> You are right. This was discussed [here](https://github.com/pyccel/pyccel/issues/487#issuecomment-717852036) but we never created an issue for it so it was forgotten. The main difficulty comes from ensuring that the names are not used by the user's variables. This is linked to issue #227

I think we can use the C types as names, and then we will treat the name claching  the same way in fortran or C

yguclu:
As mentioned above, we have discussed about this before. But since this is an "aesthetic" problem it has never reached a high priority... for obvious reasons we try to fix bugs first! 😅 

When we decide to tackle this issue, I think that using the `=>` syntax as suggested by @Beliavsky is the cleanest solution. But first we need to make sure that our treatment of name clashes in the generated code is bullet-proof.

yguclu:
As for mapping Python integers to default Fortran integers, in the best-case scenario this adds complications at the Python-Fortran interface. In the worst-case scenario the generated Fortran code will not reproduce the behavior of the original Python code. This is why we are happy with using long integers, a solution which has proven to be solid across multiple platforms.

EmilyBourne:
Duplicate of #856

PR #790, opened 2021-03-04
Fix numpy.norm bugs

mostafaMamoni:
Fix bugs in `NumpyNorm` of module `psydac.ast.numpyext`:
- `AttributeError: 'NumpyNorm' object has no attribute '_args'`;
- Some casting problems on the Fortran side.

Commit summary:
- Create `dim` property by moving up `super().__init__(arg, dim)`;
- Implement a casting process to make the Fortran code compilable;
- Import `PythonAbs`;
- Add tests for `NumpyNorm`.


EmilyBourne:
Is this PR ready for review?

pyccel/ast/numpyext.py
```diff
@@ -786,22 +786,22 @@ def __new__(cls, a, dtype=None, order='K', subok=True, shape=None):
...
     def __init__(self, arg, dim=None):
-        if isinstance(dim, ValuedArgument):
-            dim = dim.value
-        if self.dim is not None:
-            sh = list(sh)
-            del sh[self.dim]
-            self._shape = tuple(sh)
-        else:
-            self._shape = ()
-        self._rank = len(self._shape)
```
EmilyBourne, line 799: Why did you remove this? The shape is now constant. Does this mean we have lost the implementation of the optional `axis` argument? If there is no code associated with the `dim` parameter then I don't think it's a problem I was just wondering if you thought about this?

pyccel/ast/numpyext.py
```diff
@@ -786,22 +786,22 @@ def __new__(cls, a, dtype=None, order='K', subok=True, shape=None):
...
         super().__init__(arg, dim)
+        if isinstance(arg.dtype, NativeBool):
+            arg = PythonFloat(PythonInt(arg))
```
EmilyBourne, line 795: Why do you need to call 2 casting functions? Will `PythonFloat(arg)` not achieve the same result?

pyccel/ast/numpyext.py
```diff
@@ -786,22 +786,22 @@ def __new__(cls, a, dtype=None, order='K', subok=True, shape=None):
...
+        elif isinstance(arg.dtype, NativeComplex):
+            arg = PythonAbs(arg)
```
EmilyBourne, line 797: Is this needed? Does fortran's `norm` not do this?

pyccel/ast/numpyext.py
```diff
@@ -786,22 +786,22 @@ def __new__(cls, a, dtype=None, order='K', subok=True, shape=None):
...
+        elif not isinstance(arg.dtype, NativeReal):
+            arg = PythonFloat(arg)
```
EmilyBourne, line 799: Presumably this is done to change the precision? Why is this necessary? Does `np.norm` not match the precision of its results to the precision of its arguments?

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5953,3 +5953,301 @@ def get_prod(arr):
...
+    integer = randint(min_int, max_int, dtype=np.int)
+    integer32 = randint(min_int32, max_int32, dtype=np.int32)
+    integer64 = randint(min_int64, max_int64, dtype=np.int64)
+
+    fl = uniform(low=-(abs(min_float)**(1/2)), high=abs(max_float)**(1/2))
+    fl32 = uniform(low=-(abs(min_float32)**(1/2)), high=abs(max_float32)**(1/2))
+    fl32 = np.float32(fl32)
+    fl64 = uniform(low=-(abs(min_float64)**(1/2)), high=abs(max_float64)**(1/2))
+
+    cmplx128_from_float32 = uniform(low=-((abs(min_float32) / 2)**(1/2)), high=((abs(max_float32) / 2)**(1/2))) + uniform(low=-((abs(max_float32) / 2)**(1/2)), high=((abs(max_float32) / 2)**(1/2))) * 1j
```
EmilyBourne, line 5998: This line doesn't fit on my screen. Please break it up

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5953,3 +5953,301 @@ def get_prod(arr):
...
+    @types('complex64[:]')
+    @types('complex128[:]')
+    def get_norm(arr):
+        from numpy.linalg import norm
+        a = norm(arr)
+        return a
+
+    size = 5
+    rtol=1e-6
+    atol=1e-7
```
EmilyBourne, line 6121: This tolerance is quite high compared to the tolerance in the rest of this file. Do you know why it needs to be so large?

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5953,3 +5953,301 @@ def get_prod(arr):
...
+    fl = uniform(low=-((abs(min_float) / size)**(1/2)), high=(abs(max_float) / size)**(1/2), size=size)
+    fl32 = uniform(low=-((abs(min_float32) / size)**(1/2)), high=(abs(max_float32) / size)**(1/2), size=size)
+    fl32 = np.float32(fl32)
+    fl64 = uniform(low=-((abs(min_float64) / size)**(1/2)), high=(abs(max_float64) / size)**(1/2), size=size)
+
+    cmplx128_from_float32 = uniform(low=-((abs(min_float32) / (size * 2))**(1/2)), high=(abs(max_float32) / (size * 2))**(1/2), size=size) + uniform(low=-((abs(min_float32) / (size * 2))**(1/2)), high=(abs(max_float32) / (size * 2))**(1/2), size=size) * 1j
+    # the result of the last operation is a Python complex type which has 8 bytes in the alignment,
+    # that's why we need to convert it to a numpy.complex64 the needed type.
+    cmplx64 = np.complex64(cmplx128_from_float32)
+    cmplx128= uniform(low=-((abs(min_float64) / (size * 2))**(1/2)), high=(abs(max_float64) / (size * 2))**(1/2), size=size) + uniform(low=-((abs(min_float64) / (size * 2))**(1/2)), high=(abs(max_float64) / (size * 2))**(1/2), size=size) * 1j
```
EmilyBourne, line 6140: These lines are too long

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5953,3 +5953,301 @@ def get_prod(arr):
...
+    fl32 = uniform(low=-((abs(min_float32) / (size[0] * size[1]))**(1/2)), high=(abs(max_float32) / (size[0] * size[1]))**(1/2), size=size)
+    fl32 = np.float32(fl32)
+    fl64 = uniform(low=-((abs(min_float64) / (size[0] * size[1]))**(1/2)), high=(abs(max_float64) / (size[0] * size[1]))**(1/2), size=size)
+
+    cmplx128_from_float32 = uniform(low=-((abs(min_float32) / (size[0] * size[1] * 2))**(1/2)), high=(abs(max_float32) / (size[0] * size[1] * 2))**(1/2), size=size) + uniform(low=-((abs(min_float32) / (size[0] * size[1] * 2))**(1/2)), high=(abs(max_float32) / (size[0] * size[1] * 2))**(1/2), size=size) * 1j
+    # the result of the last operation is a Python complex type which has 8 bytes in the alignment,
+    # that's why we need to convert it to a numpy.complex64 the needed type.
+    cmplx64 = np.complex64(cmplx128_from_float32)
+    cmplx128= uniform(low=-((abs(min_float64) / (size[0] * size[1] * 2))**(1/2)), high=(abs(max_float64) / (size[0] * size[1] * 2))**(1/2), size=size) + uniform(low=-((abs(min_float64) / (size[0] * size[1] * 2))**(1/2)), high=(abs(max_float64) / (size[0] * size[1] * 2))**(1/2), size=size) * 1j
```
EmilyBourne, line 6222: Lines too long

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5953,3 +5953,301 @@ def get_prod(arr):
...
+
+    assert np.isclose(f_fl(fl), get_norm(fl), rtol=rtol, atol=atol)
+    assert np.isclose(f_fl32(fl32), get_norm(fl32), rtol=rtol, atol=atol)
+    assert np.isclose(f_fl64(fl64), get_norm(fl64), rtol=rtol, atol=atol)
+
+    f_complex64 = epyccel(get_norm, language=language)
+    f_complex128 = epyccel(get_norm, language=language)
+
+    assert np.isclose(f_complex64(cmplx64), get_norm(cmplx64), rtol=1e-6, atol=1e-7)
+    assert np.isclose(f_complex128(cmplx128), get_norm(cmplx128), rtol=rtol, atol=atol)
```
EmilyBourne, line 6253: Why do you specify `rtol` and `atol` on line 6252 but use the variables on line 6253?

EmilyBourne (changes requested):

tests/epyccel/recognised_functions/test_numpy_funcs.py
mostafaMamoni: sorry, i missed that.

tests/epyccel/recognised_functions/test_numpy_funcs.py
EmilyBourne: Do you know which cases? Maybe there is a problem in the generated code

pyccel/ast/numpyext.py
mostafaMamoni: Fortran `norm` doesn't do this.

pyccel/ast/numpyext.py
EmilyBourne: I'm not sure of the relevance `PythonFloat(arg)` returns a `REAL` type

pyccel/ast/numpyext.py
EmilyBourne: ok. I wonder if this is the cause of the imprecision? It is not great to do a square root and a square:
```
n = sqrt(sqrt(a**2 + b**2)**2 + ...)
```
Is there any way to translate:
```
norm(a)
```
to:
```
norm([*a.real, *a.imag])
```
? 
Ideally only in the fortran printer?

tests/epyccel/recognised_functions/test_numpy_funcs.py
mostafaMamoni: There’s float32 and complex64.

pyccel/ast/numpyext.py
mostafaMamoni: by `REAL` type i mean in Fortran, in Pyccel there's NativeFloat

EmilyBourne: Yes I understand, my question is why do you need `PythonInt`?

tests/epyccel/recognised_functions/test_numpy_funcs.py
EmilyBourne: Could the difference be related to the precision then? What is the output type for `norm(a)` when a is a `float32[:]`?

pyccel/ast/numpyext.py
```diff
@@ -786,22 +786,22 @@ def __new__(cls, a, dtype=None, order='K', subok=True, shape=None):
 class NumpyNorm(PyccelInternalFunction):
     """ Represents call to numpy.norm"""
     _dtype = NativeReal()
+    _shape = ()
+    _rank  = 0
```
EmilyBourne, line 790: Where is the precision set? I can't find it but I'm not sure if that's just because it's hard to read between the open conversations

pyccel/ast/numpyext.py
mostafaMamoni: i think there is three reasons for this:
1- PythonFloat doesn’t handle bool type
2- if we do `arg = PythonFloat(arg)` instead, it will be translated to something like 'real(TRUE)' which not acceptable in Fortran.
3- if have to handle bool type in PythonFloat i think we will do the same thing, cast `arg` to int type and then to float.

EmilyBourne: It does now. I think it was fixed in PR #783

pyccel/ast/numpyext.py
EmilyBourne: Never mind, there is a `not`

pyccel/ast/numpyext.py
mostafaMamoni: and in case of arrays ?

EmilyBourne: Hence the unpacking `*a.real`, I'm not sure how you can write this in pyccel though, I'll leave this up to you to investigate

pyccel/ast/numpyext.py
EmilyBourne: See [here](https://github.com/pyccel/pyccel/blob/master/pyccel/codegen/printing/fcode.py#L786-L787)

pyccel/codegen/printing/pycode.py
```diff
@@ -458,6 +458,15 @@ def _print_NumpyRandint(self, expr):
             args += ", size = {}".format(size)
         return "randint({})".format(args)
 
+    def _print_NumpyNorm(self, expr):
+        type_name = type(expr).__name__
+        name = type_name[5:].lower()
+        print(name)
```
EmilyBourne, line 464: This print can be removed

EmilyBourne: `print(name)` this print, not the `_print_NumpyNorm` function ^^

pyccel/ast/numpyext.py
```diff
@@ -786,22 +786,32 @@ def __new__(cls, a, dtype=None, order='K', subok=True, shape=None):
...
     def __init__(self, arg, dim=None):
-        if isinstance(dim, ValuedArgument):
-            dim = dim.value
+        super().__init__(arg, dim)
+        cmplx_precision = None
+        if isinstance(arg.dtype, NativeBool):
+            arg = PythonFloat(arg)
+        elif isinstance(arg.dtype, NativeComplex):
+            cmplx_precision = arg.precision
+            arg = PythonAbs(arg)
```
EmilyBourne, line 799: I think the addition of the `PythonAbs` should be done in `fcode.py` as it is only a problem in fortran

pyccel/codegen/printing/pycode.py
```diff
@@ -458,6 +458,13 @@ def _print_NumpyRandint(self, expr):
             args += ", size = {}".format(size)
         return "randint({})".format(args)
 
+    def _print_NumpyNorm(self, expr):
+        type_name = type(expr).__name__
+        name = type_name[5:].lower()
+        return "{name}({arg})".format(
+                name = name,
+                arg  = self._print(expr.python_arg))
```
EmilyBourne, line 466: Can we not print `expr.dim`?

mostafaMamoni: because we test only the basic case of numpy.norm `dim` will come `None` which has no printer so `Error at code generation` will be raised. numpy.norm has new additional arguments in the [docs](https://numpy.org/doc/stable/reference/generated/numpy.linalg.norm.html) , and think `keepdim` in this doc can refer to `dim` in Pyccel, So `dim` should not have `None` as a default value.

EmilyBourne: I think `dim` corresponds to `axis` given the effect it has on the shape. It seems to be implemented in fortran so maybe we can add this to the tests? Of course it should only be printed in the cases where `expr.dim is not None`

pyccel/ast/numpyext.py
```diff
@@ -786,25 +786,44 @@ def __new__(cls, a, dtype=None, order='K', subok=True, shape=None):
...
     _dtype = NativeReal()
-
-    def __init__(self, arg, dim=None):
-        if isinstance(dim, ValuedArgument):
-            dim = dim.value
-        if self.dim is not None:
-            sh = list(sh)
-            del sh[self.dim]
+    _shape = ()
+    _rank  = 0
```
EmilyBourne, line 790: Why do you need these class properties? Class properties should be used for the shape when the shape is always the same. This is not the case here as you set the shape below

pyccel/ast/numpyext.py
```diff
@@ -786,25 +786,44 @@ def __new__(cls, a, dtype=None, order='K', subok=True, shape=None):
...
+
+    def __init__(self, arg, axis=None):
+        super().__init__(arg, axis)
+        cmplx_precision = None
+        if isinstance(arg.dtype, NativeBool):
+            arg = PythonFloat(arg)
+        elif isinstance(arg.dtype, NativeComplex):
+            cmplx_precision = arg.precision
+            arg = PythonAbs(arg)
```
EmilyBourne, line 799: As already discussed I think the call to  `PythonAbs` should be added in fcode.py as other languages may implement a norm for complexes. This should be used if possible as it removes unnecessary sqrt calls.

E.g. In a language with cnorm, the complex norm is calculated as:
```
sqrt( a.real**2 + a.imag**2 + b.real**2 + b.imag**2 + ...)
```
but in a language without a complex norm it must be calculated as:
```
sqrt( sqrt(a.real**2 + a.imag**2)**2 + sqrt(b.real**2 + b.imag**2)**2 + ...)
```

EmilyBourne, line 799: I see that you store the precision as `PythonAbs` is a builtin function which therefore always returns an object with the default precision. Could you not get round this problem by using `NumpyAbs`?

pyccel/ast/numpyext.py
```diff
@@ -786,25 +786,44 @@ def __new__(cls, a, dtype=None, order='K', subok=True, shape=None):
...
+            arg = PythonFloat(arg)
+        self._arg = PythonList(arg) if arg.rank == 0 else arg
+        self._precision = cmplx_precision if cmplx_precision else arg.precision
+        if self.axis is not None:
+            sh = list(arg.shape)
+            del sh[self.axis]
             self._shape = tuple(sh)
         else:
             self._shape = ()
         self._rank = len(self._shape)
```
EmilyBourne, line 810: If shape and rank are set then order should also be set

pyccel/ast/numpyext.py
```diff
@@ -786,25 +786,44 @@ def __new__(cls, a, dtype=None, order='K', subok=True, shape=None):
...
-        super().__init__(arg, dim)
 
     @property
     def arg(self):
+        return self._arg
+
+
+    @property
+    def python_arg(self):
+        """numpy.norm arguement without casting.
```
EmilyBourne, line 819: Typo `arguement`->`argument`

pyccel/codegen/printing/fcode.py
```diff
@@ -658,8 +658,9 @@ def _print_NumpyEmpty(self, expr):
     def _print_NumpyNorm(self, expr):
         """Fortran print."""
 
-        if expr.dim:
-            rhs = 'Norm2({},{})'.format(self._print(expr.arg),self._print(expr.dim))
+        if expr.axis:
+            axis = '1' if expr.axis == 0 else self._print(expr.axis)
```
EmilyBourne, line 662: The axis indexing begins at 0 in python and 1 in fortran. To get the same result you cannot just fix the first case. This will ensure that the code compiles and runs but it will not give the correct answer

pyccel/codegen/printing/pycode.py
```diff
@@ -360,6 +360,9 @@ def _print_Break(self, expr):
     def _print_Continue(self, expr):
         return 'continue'
 
+    def _print_NoneType(self, expr):
+        return 'None'
```
EmilyBourne, line 364: We don't want to print `None`. If `None` was passed to the function then it will be a `Nil` object by this stage, if nothing was passed then we will have `None` at this stage, this just means that we don't want to print the argument. Otherwise we will have unreadable code. E.g.:
```
x = full((3,4), 4.5, dtype=None, order=None)
```
This situation will get worse and worse for more complex functions with lots of arguments

pyccel/codegen/printing/pycode.py
```diff
@@ -458,6 +461,14 @@ def _print_NumpyRandint(self, expr):
             args += ", size = {}".format(size)
         return "randint({})".format(args)
 
+    def _print_NumpyNorm(self, expr):
+        type_name = type(expr).__name__
+        name = type_name[5:].lower()
+        return  "{name}({arg},axis={axis})".format(
+                name = name,
+                arg  = self._print(expr.python_arg),
+                axis=self._print(expr.axis))
```
EmilyBourne, line 470: If `expr.axis=None` then we don't want to print the argument (see fcode.py for an example)

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5953,3 +5956,308 @@ def get_prod(arr):
...
+        from numpy.linalg import norm
+        b = norm(a)
+        return b
+
+    @types('complex64')
+    @types('complex128')
+    def get_norm_complex(a):
+        from numpy.linalg import norm
+        b = norm(a)
+        return b
```
EmilyBourne, line 5991: You have split the tests in the scalar case, but not in the array cases and you have also added logic to the code so you don't need to split the cases. Please handle this the same way in all the tests

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5953,3 +5956,308 @@ def get_prod(arr):
...
+
+    size = (2, 5)
+
+    bl = randint(0, 1, size=size, dtype= bool)
+
+    integer8 = randint(min_int8, max_int8, size=size, dtype=np.int8)
+    integer16 = randint(min_int16, max_int16, size=size, dtype=np.int16)
+    integer = randint(min_int, max_int, size=size, dtype=np.int)
+    integer32 = randint(min_int32, max_int32, size=size, dtype=np.int32)
+    integer64 = randint(min_int64, max_int64, size=size, dtype=np.int64)
```
EmilyBourne, line 6219: Are you sure there are no overflow errors for the higher precision integers?

EmilyBourne (changes requested):

pyccel/ast/numpyext.py
mostafaMamoni: i though we should keep it in the ast since numpy is doing it internally.

EmilyBourne: The question is whether numpy is the outlier here, or fortran. C obviously does neither as it doesn't handle arrays. I wonder what julia does.

That said I suspect that when we implement it in c we will create a `norm` and a `cnorm` function call as this will avoid the unnecessary `sqrt**2` and will therefore run faster. If that is the case then the `abs` will cause problems

tests/epyccel/recognised_functions/test_numpy_funcs.py
mostafaMamoni: yes, i tested it.

pyccel/ast/numpyext.py
mostafaMamoni: `NumpyAbs` return the same as `PythonAbs`.

tests/epyccel/recognised_functions/test_numpy_funcs.py
EmilyBourne: Prod promotes its arguments to floats so this is not a problem

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5953,3 +5956,308 @@ def get_prod(arr):
...
+              uniform(low=-((abs(min_float64) / (size[0] * size[1] * 2))**(1/2)), high=(abs(max_float64) / (size[0] * size[1] * 2))**(1/2), size=size) * 1j
+
+    f_bl = epyccel(get_norm, language=language)
+
+    assert np.isclose(f_bl(bl), get_norm(bl), rtol=RTOL, atol=ATOL).any()
+
+    f_integer8 = epyccel(get_norm, language=language)
+    f_integer16 = epyccel(get_norm, language=language)
+    f_integer = epyccel(get_norm, language=language)
+    f_integer32 = epyccel(get_norm, language=language)
```
EmilyBourne, line 6241: Why are you translating the function multiple times?! You have an interface so it only needs translating once

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5953,3 +5956,308 @@ def get_prod(arr):
...
+
+    assert np.isclose(f_integer8(integer8), get_norm(integer8), rtol=RTOL, atol=ATOL).any()
+    assert np.isclose(f_integer16(integer16), get_norm(integer16), rtol=RTOL, atol=ATOL).any()
+    assert np.isclose(f_integer(integer), get_norm(integer), rtol=RTOL, atol=ATOL).any()
+    assert np.isclose(f_integer32(integer32), get_norm(integer32), rtol=RTOL, atol=ATOL).any()
```
EmilyBourne, line 6246: You should be using `.all` not `.any`. You want all the results to match not one of them!

mostafaMamoni: you are right.

pyccel/ast/numpyext.py
EmilyBourne: Ah sorry, you are right. But they also both preserve the precision:
```
In [2]: type(abs(np.float32(3)))
Out[2]: numpy.float32

In [3]: type(np.abs(np.float32(3)))
Out[3]: numpy.float32

In [4]: type(np.abs(np.complex64(3)))
Out[4]: numpy.float32

In [5]: type(abs(np.complex64(3)))
Out[5]: numpy.float32
```

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5953,3 +5956,256 @@ def get_prod(arr):
...
+    @types('int64[:,:]')
+    @types('float[:,:]')
+    @types('float32[:,:]')
+    @types('float64[:,:]')
+    @types('complex64[:,:]')
+    @types('complex128[:,:]')
+    def get_norm(arr):
+        from numpy.linalg import norm
+        from numpy import shape
+        a = norm(arr, axis=1)
```
EmilyBourne, line 6172: Currently there are problems with this call but our tests are not extensive enough to spot it. Perhaps you could make axis an argument to `get_norm` then different cases can be tested. Ie.:
```
def get_norm(arr, ax):
        from numpy.linalg import norm
        from numpy import shape
        a = norm(arr, axis=ax)
        ...
```
Then you can test with `ax=0` and `ax=1`

mostafaMamoni: what values can `axis` take ?

EmilyBourne: Any value between `0` and `rank` so `0` for 1D arrays, `0` or `1` for 2D arrays, `0`,`1`,`2` for 3D arrays, etc

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5953,3 +5956,329 @@ def get_prod(arr):
...
+    @types('int8[:,:,:]')
+    @types('int16[:,:,:]')
+    @types('int32[:,:,:]')
+    @types('int64[:,:,:]')
+    @types('float[:,:,:]')
+    @types('float32[:,:,:]')
+    @types('float64[:,:,:]')
+    @types('complex64[:,:,:]')
+    @types('complex128[:,:,:]')
+    def get_norm(arr):
```
EmilyBourne, line 6238: As discussed, please add a test with Fortran ordering

pyccel/codegen/printing/fcode.py
```diff
@@ -657,13 +657,14 @@ def _print_NumpyEmpty(self, expr):
 
     def _print_NumpyNorm(self, expr):
         """Fortran print."""
-
-        if expr.dim:
-            rhs = 'Norm2({},{})'.format(self._print(expr.arg),self._print(expr.dim))
+        arg = PythonAbs(expr.arg) if isinstance(expr.arg.dtype, NativeComplex) else expr.arg
+        if expr.axis:
+            axis = str(abs(expr.axis.python_value - arg.rank)) + '_C_DOUBLE'
```
EmilyBourne, line 662: Please use pyccel nodes rather than python values, this way we will be sure that the kind tag is correct and it can be easily modified everywhere if we need to change it (e.g. when we fix #604 )

pyccel/codegen/printing/pycode.py
```diff
@@ -458,6 +458,14 @@ def _print_NumpyRandint(self, expr):
             args += ", size = {}".format(size)
         return "randint({})".format(args)
 
+    def _print_NumpyNorm(self, expr):
+        type_name = type(expr).__name__
+        name = type_name[5:].lower()
```
EmilyBourne, line 463: Is there any reason you don't just use `'norm'`?

EmilyBourne (changes requested):

pyccel/codegen/printing/fcode.py
```diff
@@ -659,8 +659,12 @@ def _print_NumpyNorm(self, expr):
         """Fortran print."""
         arg = PythonAbs(expr.arg) if isinstance(expr.arg.dtype, NativeComplex) else expr.arg
         if expr.axis:
-            axis = str(abs(expr.axis.python_value - arg.rank)) + '_C_DOUBLE'
-            code = 'Norm2({},{})'.format(self._print(arg), axis)
+            axis = expr.axis
+            if arg.order != 'F':
+                axis = PythonAbs(PyccelMinus(expr.axis, LiteralInteger(arg.rank)))
```
EmilyBourne, line 664: Isn't `axis < rank`? Why not just do `PyccelMinus(LiteralInteger(arg.rank), expr.axis)` to avoid the abs?

pyccel/codegen/printing/fcode.py
```diff
@@ -659,8 +659,12 @@ def _print_NumpyNorm(self, expr):
...
+            else:
+                axis = LiteralInteger(expr.axis.python_value + 1)
```
EmilyBourne, line 666: Why do you need a `+1` here, but not above?

mostafaMamoni: i'm trying to understand why does not give the same result as Fortran at Fortran ordered array with many ways and this is one of them, after some testing i don't think this can help anymore.

mostafaMamoni: given that Python gives the same result with both `C` and `F` order.

EmilyBourne:
Please add the necessary slots to fix the failing tests

pyccel/codegen/printing/fcode.py
EmilyBourne: @EmilyBourne 
> Why do you need a `+1` here, but not above?

`arg.rank` is the size not the ubound so the `+1` is already contained in this variable

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5953,3 +5956,534 @@ def get_prod(arr):
...
+    assert np.isclose(epyccel_func(integer8), get_norm(integer8), rtol=RTOL, atol=ATOL).all()
+    assert np.isclose(epyccel_func(integer16), get_norm(integer16), rtol=RTOL, atol=ATOL).all()
+    assert np.isclose(epyccel_func(integer), get_norm(integer), rtol=RTOL, atol=ATOL).all()
+    assert np.isclose(epyccel_func(integer32), get_norm(integer32), rtol=RTOL, atol=ATOL).all()
+    assert np.isclose(epyccel_func(integer64), get_norm(integer64), rtol=RTOL, atol=ATOL).all()
+    assert np.isclose(epyccel_func(fl), get_norm(fl), rtol=RTOL, atol=ATOL).all()
+    assert np.isclose(epyccel_func(fl32), get_norm(fl32), rtol=RTOL32, atol=ATOL32).all()
+    assert np.isclose(epyccel_func(fl64), get_norm(fl64), rtol=RTOL, atol=ATOL).all()
+    assert np.isclose(epyccel_func(cmplx64), get_norm(cmplx64), rtol=RTOL32, atol=ATOL32).all()
+    assert np.isclose(epyccel_func(cmplx128), get_norm(cmplx128), rtol=RTOL, atol=ATOL).all()
```
EmilyBourne, line 6347: As was recently pointed out to me, numpy has an `allclose` function which is faster than using `isclose(..).all()`

EmilyBourne (changes requested):
Could you add a multidimensional test without the axis parameter please

mostafaMamoni:
> Could you add a multidimensional test without the axis parameter please

you mean 3d array, right ?

mostafaMamoni:
> Could you add a multidimensional test without the axis parameter please

we already have a test for scalars, and 6 tests, 1/2/3d array tests with(out) axis argument. did i miss something

EmilyBourne:
> > Could you add a multidimensional test without the axis parameter please
> 
> we already have a test for scalars, and 6 tests, 1/2/3d array tests with(out) axis argument. did i miss something

Maybe I missed something? I see
- test for scalars
- 2 tests with 1D arrays one with C ordering, one with F ordering (1D arrays don't have an ordering one of these tests is not needed) both of which have `axis=0`
- 2 tests with 2D arrays, both of which have `axis=0` and `axis=1`
- 2 tests with 3D arrays, both of which have `axis=0`, `axis=1` and `axis=2`

I don't see a test like:
```
...
@types('complex128[:,:]')
def get_norm(arr):
        from numpy.linalg import norm
        return norm(arr)
```

EmilyBourne (approved):

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5953,3 +5956,489 @@ def get_prod(arr):
...
+    @types('float32[:,:]')
+    @types('float64[:,:]')
+    @types('complex64[:,:]')
+    @types('complex128[:,:]')
+    def get_norm(arr):
+        from numpy.linalg import norm
+        from numpy import shape
+        a = norm(arr)
+        s = shape(a)
+        return len(s), s[0], a[0]
```
EmilyBourne, line 6174: When no axis is specified `norm` sums over all the axes, so it returns a scalar

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5953,3 +5956,489 @@ def get_prod(arr):
...
+    @types('float32[:,:,:]')
+    @types('float64[:,:,:]')
+    @types('complex64[:,:,:]')
+    @types('complex128[:,:,:]')
+    def get_norm(arr):
+        from numpy.linalg import norm
+        from numpy import shape
+        a = norm(arr)
+        s = shape(a)
+        return len(s), s[0], a[0][0]
```
EmilyBourne, line 6323: When no axis is specified `norm` sums over all the axes, so it returns a scalar

EmilyBourne (approved):

pyccel/ast/numpyext.py
```diff
@@ -819,30 +819,47 @@ def __new__(cls, a, dtype=None, order='K', subok=True, shape=None):
...
-            del sh[self.dim]
+            del sh[self.axis]
             self._shape = tuple(sh)
             self._order = arg.order
         else:
             self._shape = ()
             self._order = None
         self._rank = len(self._shape)
-        super().__init__(arg, dim)
+        self._order = arg.order
```
yguclu, line 843: I can't understand what is happening with the variable `cmplx_precision`.

It is set to `None` on line 827 and then its value is checked on line 833... but we know that the value is `None`, of course!

Can you explain?

yguclu (changes requested):
Nice job! I have one comment

pyccel/ast/numpyext.py
mostafaMamoni: i was using `PythonAbs` in case of `arg` is a complex type, and because we to save the precision of `arg` and `PythonAbs` has a default precision which is 8, when passing complex arg with 4 precision, `PythonAbs` will change it to 8 precision which we didn't want to have, so i had to store the original precision of before passing `arg` to `PythonAbs`. but now we are handling complex types in the Fortran printer since Fortran `Norm2` has a problem with complex type and Python not, so this variable is useless.

pyccel/ast/numpyext.py
```diff
@@ -824,13 +824,12 @@ class NumpyNorm(PyccelInternalFunction):
 
     def __init__(self, arg, axis=None):
         super().__init__(arg, axis)
-        cmplx_precision = None
         if isinstance(arg.dtype, NativeBool):
             arg = PythonFloat(arg)
         elif not isinstance(arg.dtype, (NativeComplex, NativeReal)):
             arg = PythonFloat(arg)
```
yguclu: I am also confused by this if block: if `arg.dtype` is a `NativeBool` the second condition evaluates to `True`, so why do you need the first condition? In other words, you could just write
```python
if not isinstance(arg.dtype, (NativeComplex, NativeReal)):
    arg = PythonFloat(arg)
```
and obtain exactly the same behavior as in the code.

yguclu (changes requested):

pyccel/ast/numpyext.py
yguclu: Thanks for removing it.

pyccel/ast/numpyext.py
mostafaMamoni: you are right, for this i was believing that PythonFloat doesn’t handle bool type so this block was something like:
```python
            arg = PythonFloat(PythonInt(arg))
```
after Emily's fix, bool type is handled in PythonFloat Fortran pinter. so then that block becomes:
```python
            arg = PythonFloat(arg)
```
so, it's a useless block too, thanks.

yguclu (approved):

PR #796, opened 2021-03-06
Add __slots__ to classes in AST

### LINKED ISSUE #515: Add __slots__ to classes in ast
Using __slots__ can accelerate code. The following modules need this improvement
- [x] ast.basic
- [x] ast.core
- [x] ast.datatypes
- [x] ast.headers
- [x] ast.builtins
- [x] ast.numpyext
- [x] ast.macros
- [x] ast.cwrapper
- [x] ast.functionalexpr
- [x] ast.mathext
- [x] ast.operators
- [x] ast.bitwise_operators
- [x] ast.internals
- [x] ast.itertoolsext
- [x] ast.literals
- [x] ast.variable

EmilyBourne:
Add `__slots__` to all classes in the AST. This fixes #515:

- Do not allow AST class instances to have an internal `__dict__`; this saves memory and accelerates attribute access;
- As before, if all instances of a given class share a common value for an attribute, store this value in a class variable;
- If an attribute's value varies among different instances of the same class, use a slot variable for it;
- Separate out Python printer tests to allow more useful analysis of timings.


yguclu:
@EmilyBourne Some conflicts need to be resolved.

pyccel/ast/datatypes.py
```diff
@@ -221,8 +220,8 @@ def codomain(self):
                   'str': String}
 
 
-class UnionType(Basic):
-    _attribute_nodes = ()
+class UnionType:
```
EmilyBourne, line 223: @saidctb I have a Codacy error here:
> Missing class docstring

I'm not clear sure what this class is for so I'm not sure how to write the docstring. Can you provide any insight?

EmilyBourne: I had a go myself

pyccel/ast/basic.py
```diff
@@ -307,11 +308,20 @@ class PyccelAstNode(Basic):
...
+        if not hasattr(self, '_shape'):
+            self._shape = None
+        if not hasattr(self, '_rank'):
+            self._rank = None
+        if not hasattr(self, '_dtype'):
+            self._dtype = None
+        if not hasattr(self, '_precision'):
+            self._precision = None
+        if not hasattr(self, '_order'):
+            self._order = None
```
yguclu, line 324: I wonder if we really want every `PyccelAstNode` object to have all these slots. Couldn't we achieve the same behavior using the properties directly? For example

```python
class PyccelAstNode(Basic):
    stage = None
    __slots__ = ()

    # No __init__() function!

    @property
    def shape(self):
        return None

    @property
    def dtype(self):
        return None

    @property
    def precision(self):
        return None

    @property
    def order(self):
        return None
```

By the way, I don't recall the reason for having two separate classes `Basic` and `PyccelAstNode`. Can you remind me please?

yguclu: It is also possible to declare class-wide default values as before and let the subclasses declare only the slots that are needed:

```python
class PyccelAstNode(Basic):

    stage = None

    _shape     = None
    _rank      = None
    _dtype     = None
    _precision = None
    _order     = None

    __slots__ = ()  # fundamental!


class PythonInt(PyccelAstNode):
    _dtype = 'int'

    __slots__ = ('_precision',)
```

This would also allow us to remove properties in the future, if we need to further optimize our code.

EmilyBourne: The names are poorly chosen. I had a discussion about this with @saidctb recently. `Basic` ought to be named `PyccelAstNode` and `PyccelAstNode` ought to be named something else, although I'm not sure what.
In the current state:
`Basic` - All classes in the pyccel AST inherit from this class. It takes care of the fst and the tree
`PyccelAstNode` - This represents an object that can be used in an `Assign` expression. Every object that inherits from this class should set:
- dtype
- precision
- shape
- rank
In addition if the rank > 1 then it should also set:
- order

I think there may be 1 or 2 classes which don't do this, hence the need for `if hasattr` but probably these classes either have bugs or ought to inherit from `Basic` instead. Maybe we should create an issue to clean up these classes

EmilyBourne: > I wonder if we really want every `PyccelAstNode` object to have all these slots

If it inherits from `PyccelAstNode` and it doesn't need these slots then it probably shouldn't inherit from `PyccelAstNode`

EmilyBourne: > Couldn't we achieve the same behavior using the properties directly?

Yes, but then we would need to override the property in lots of places. If we do this then there is no point in this class

EmilyBourne: > It is also possible to declare class-wide default values as before and let the subclasses declare only the slots that are needed

This is possible, but again it will lead to a lot of duplication

yguclu: > > I wonder if we really want every `PyccelAstNode` object to have all these slots
> 
> If it inherits from `PyccelAstNode` and it doesn't need these slots then it probably shouldn't inherit from `PyccelAstNode`

I don't necessarily agree on this. A slot is something that changes from one object to another of the same class. So many subclasses do not need those slots, but just a class-wide value for those properties.

EmilyBourne: Yes, ok. It doesn't need the slots, it just needs the attributes. I sent the first message without having seen your second message so I thought you were thinking of objects with dtype = None

EmilyBourne: > It is also possible to declare class-wide default values as before and let the subclasses declare only the slots that are needed

This should be possible I think.

yguclu: My basic point is this: since `__slots__` are an optimization, we should be careful to use them the best we can, otherwise we are not going to improve anything and all our efforts will be in vain.

EmilyBourne:
```
class PyccelAstNode(Basic):

    stage = None

    __slots__ = ()  # fundamental!

    @property
    def shape(self):
        return self._shape #pylint : disable=no-member
```

all subclasses must either define a class property or a slot for:
- dtype
- precision
- shape
- rank
- order

pyccel/ast/basic.py
```diff
@@ -308,49 +308,35 @@ class PyccelAstNode(Basic):
...
-        if not hasattr(self, '_shape'):
-            self._shape = None
-        if not hasattr(self, '_rank'):
-            self._rank = None
-        if not hasattr(self, '_dtype'):
-            self._dtype = None
-        if not hasattr(self, '_precision'):
-            self._precision = None
-        if not hasattr(self, '_order'):
-            self._order = None
```
yguclu: I think that you still need to set `__slots__ = ()` in this class, in order to avoid the creation of empty dictionaries `__dict__ = {}` in the instances of the subclasses.

EmilyBourne: You are right. Fixed in 57df337

EmilyBourne:
This branch is now passing. The remaining codacy errors are related to setting dtype, precision etc outside the `__init__` function (this is not strictly the case as they are set in functions called by the `__init__` function, but pylint doesn't have this nuance). This will be fixed in PR#797 so I am going to leave it as it is for now and mark it as blocked

yguclu (approved):

yguclu (approved):

saidctb (approved):

yguclu (approved):

yguclu (changes requested):
I have just realized that some classes in our AST do not have instance attributes and do not use `__slots__`. See for example `PythonReal` and `PythonImag` in **pyccel/ast/builtins.py**. 

As a consequence, instances of those classes will have a useless empty `__dict__`. Unless there is a good reason not to do so, I think that we should hunt them down and set
```python
__slots__ = ()
```

EmilyBourne:
> I have just realized that some classes in our AST do not have instance attributes and do not use `__slots__`. See for example `PythonReal` and `PythonImag` in **pyccel/ast/builtins.py**.
> 
> As a consequence, instances of those classes will have a useless empty `__dict__`. Unless there is a good reason not to do so, I think that we should hunt them down and set
> 
> ```python
> __slots__ = ()
> ```

You are right. For some reason I was convinced that one of the conclusions of our conversation was that `__slots__ = ()` was useless but looking back at the conversation I think I misunderstood. I will fix it

EmilyBourne:
@yguclu The tests seem to be slower after the addition of the empty slots. I'm not sure what would cause that, but I wonder if this is what I remembered discussing previously?

yguclu:
> @yguclu The tests seem to be slower after the addition of the empty slots. I'm not sure what would cause that, but I wonder if this is what I remembered discussing previously?

They were only slower on Linux, so I think it's likely that the reason is different. We may want to look at the timings of the Python tests separately, as they are much faster and therefore more effected by these changes.

EmilyBourne:
@yguclu
> They were only slower on Linux, so I think it's likely that the reason is different. We may want to look at the timings of the Python tests separately, as they are much faster and therefore more affected by these changes.

I have separated out the python timings, but as we don't have this separation on any other branch I'm not sure if it's faster or slower. But with the speed at which the python tests are running I struggle to see how it could be responsible for the tests being 5mins slower

yguclu:
> @yguclu
> 
> > They were only slower on Linux, so I think it's likely that the reason is different. We may want to look at the timings of the Python tests separately, as they are much faster and therefore more affected by these changes.
> 
> I have separated out the python timings, but as we don't have this separation on any other branch I'm not sure if it's faster or slower. But with the speed at which the python tests are running I struggle to see how it could be responsible for the tests being 5mins slower

I agree, and I don't see any reason to worry about this any longer. I think this PR can be merged now.

yguclu (approved):
Excellent job @EmilyBourne! Thank you!

PR #797, opened 2021-03-08
Allow simplification of pyccel-created nodes

abkhaoula:
Allowing simplification of pyccel-created nodes : 

- [x] In `ast.operators` modify the functions `_set_dtype` and `_set_shape_rank` so they are staticmethods (this allows them to be called without an instance). They should return the required dtype/precision, shape/rank instead of setting it directly. The names should be changed accordingly (e.g.` _calculate_dtype`,` _calculate_shape_rank`
- [x] Add an optional boolean keyword argument to the `PyccelOperator` class to indicate whether or not the instance is being created as part of a translation or not
- [x] Add `__new__` functions as necessary to fix what is described above. As a minimum this should include `PyccelAdd` and `PyccelMinus`. For example `PyccelAdd` should contain:
 ```
def __new__(cls, arg1, arg2, simplify = False):
        if simplify:
            if isinstance(arg2, PyccelUnarySub):
                return PyccelMinus(arg1, arg2.args[0])
            elif isinstance(arg1, Literal) and isinstance(arg2, Literal):
                dtype, prec = cls._calculate_dtype(arg1, arg2)
                return get_literal(arg1.value + arg2.value, dtype, prec)
        return super().__new__(cls, arg1, arg2, simplify = simplify)
```
- [x] Add` simplify = True` to all nodes where it could help for clarity. This will probably include most nodes created in the codegen stage

### CODE REVIEW

File: .idea/inspectionProfiles/Project_Default.xml [hunk 1]
```diff
@@ -0,0 +1,59 @@
+<component name="InspectionProjectProfileManager">
```
EmilyBourne, line 1: This file seems to have been committed accidentally

.idea/inspectionProfiles/profiles_settings.xml
```diff
@@ -0,0 +1,6 @@
+<component name="InspectionProjectProfileManager">
```
EmilyBourne, line 1: This file too

.idea/modules.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
EmilyBourne, line 1: And this one

.idea/pyccel.iml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
EmilyBourne, line 1: Also this

.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
EmilyBourne, line 1: Also this

pyccel/ast/operators.py
```diff
@@ -761,14 +813,12 @@ class PyccelBooleanOperator(PyccelOperator):
...
     """
-    _dtype = NativeBool()
-    _rank  = 0
-    _shape = ()
-    _precision = default_precision['bool']
 
-    def _set_dtype(self):
+    @staticmethod
+    def _calculate_dtype():
         pass
```
EmilyBourne, line 819: Previously we had:
```
def _set_dtype(self):
    pass
```
which means that the dtype and precision were not set. This is done because `self._dtype` already returns something `NativeBool()`. This works because the `_dtype` is a class property. If it is no longer a class property then the dtype needs to be set. So here we should have:
```
@staticmethod
    def _calculate_dtype():
         _dtype = NativeBool()
        return _dtype, _precision
```

pyccel/ast/operators.py
```diff
@@ -761,14 +813,12 @@ class PyccelBooleanOperator(PyccelOperator):
...
         The second argument passed to the operator
     """
-    _dtype = NativeBool()
-    _rank  = 0
-    _shape = ()
-    _precision = default_precision['bool']
 
-    def _set_dtype(self):
+    @staticmethod
+    def _calculate_dtype():
```
EmilyBourne, line 818: This function is called from `PyccelOperator.__init__`, therefore all the different versions of this function need to have the same prototype. Otherwise there is no way to know what arguments need to be passed

pyccel/ast/operators.py
```diff
@@ -197,27 +197,25 @@ class PyccelUnaryOperator(PyccelOperator):
     def __init__(self, arg):
         super().__init__(arg)
 
-    def _set_dtype(self):
+    @staticmethod
+    def _calculate_dtype(_args):
```
EmilyBourne, line 201: Have you thought about using [tuple unpacking](https://www.geeksforgeeks.org/packing-and-unpacking-arguments-in-python/) (i.e. `*_args`) here instead of passing `_args`?

.idea/inspectionProfiles/Project_Default.xml
abkhaoula: yes sorry about that

pyccel/ast/operators.py
abkhaoula: I ll try that

pyccel/ast/operators.py
abkhaoula: ow right. Thank you Very much

pyccel/ast/operators.py
abkhaoula: Ok. This was one of the things I didn't understand. Thank you so much.

pyccel/ast/operators.py
```diff
@@ -198,21 +198,21 @@ def __init__(self, arg):
         super().__init__(arg)
 
     @staticmethod
-    def _calculate_dtype(_args):
+    def _calculate_dtype(*_args):
         """ Sets the dtype and precision
         They are chosen to match the argument
         """
-        a = _args[0]
+        a = _args[0][0]
```
EmilyBourne, line 205: Why `[0][0]`? Is this because you are calling this as:
```
self._calculate_dtype(self.args)
```
?
You should do:
```
self._calculate_dtype(*self.args)
```
This way `_args` and `self.args` will contain the same thing, and you can even do:
```
def _calculate_dtype(arg1, arg2):
```
in the classes where this makes sense

pyccel/ast/operators.py
```diff
@@ -321,89 +335,68 @@ class PyccelBinaryOperator(PyccelOperator):
...
+        strs      = [a for a in _args[0] if a.dtype is NativeString()]
 
         if strs:
-            self._handle_str_type(strs)
             assert len(integers + reals + complexes) == 0
+            raise TypeError("unsupported operand type(s) for /: 'str' and 'str'")
         elif complexes:
-            self._handle_complex_type(complexes)
+            _dtype = NativeComplex()
+            _precision = max(a.precision for a in complexes)
```
EmilyBourne, line 357: You have removed the `_handle_X_type` functions. Some of these functions are overridden in subclasses. E.g. [here](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/operators.py#L565-L567) and [here](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/operators.py#L484-L485)

EmilyBourne: Ah, I see you have seen this :+1:

pyccel/ast/operators.py
```diff
@@ -481,8 +474,38 @@ def __new__(cls, arg1, arg2):
...
+        If one argument is a string then all arguments must be strings
+        If the arguments are numeric then the dtype and precision
+        match the broadest type and the largest precision
+        e.g.
+            1 + 2j -> PyccelAdd(LiteralInteger, LiteralComplex) -> complex
+        """
+        integers  = [a for a in _args[1] if a.dtype in (NativeInteger(),NativeBool())]
+        reals     = [a for a in _args[1] if a.dtype is NativeReal()]
+        complexes = [a for a in _args[1] if a.dtype is NativeComplex()]
+        strs      = [a for a in _args[1] if a.dtype is NativeString()]
```
EmilyBourne, line 488: There is a lot of code duplication here. I think these should be class methods rather than static methods. See [this](https://www.geeksforgeeks.org/class-method-vs-static-method-python/) for a description of static vs class methods. I think with a class method, you should be able to do `cls._handle_integers()` to call the correct version of the method

EmilyBourne:
I'm not sure if you noticed but there are also operators in `bitwise_operators.py` (it's split into 2 files to avoid circular dependencies)

pyccel/ast/operators.py
abkhaoula: I edited it in [6efbe03](https://github.com/pyccel/pyccel/pull/797/commits/6efbe034730311dd0100d5dd5d9f6444449e94f1)

pyccel/ast/operators.py
abkhaoula: done in [6efbe03](https://github.com/pyccel/pyccel/pull/797/commits/6efbe034730311dd0100d5dd5d9f6444449e94f1)

pyccel/ast/operators.py
```diff
@@ -978,8 +955,8 @@ def _calculate_dtype(*_args):
         """
         Sets the dtype and precision for IfTernaryOperator
         """
-        value_true = _args[0][1]
```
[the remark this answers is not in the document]
abkhaoula: done in [6c29ed6](https://github.com/pyccel/pyccel/pull/797/commits/6c29ed6f42951d429b1057c48337497045a1ea3d)

EmilyBourne:
This looks better. I notice you didn't use `classmethod`s in `bitwise_operators.py`, but I think this should be ok as I can't think of a situation where we would internally generate a bitwise operator.

You should make sure you fix the codacy errors too

abkhaoula:
> This looks better. I notice you didn't use `classmethod`s in `bitwise_operators.py`, but I think this should be ok as I can't think of a situation where we would internally generate a bitwise operator.
> 
> You should make sure you fix the codacy errors too

In the `_calculate_dtype` the method not only returns the `_dtype` and `_precision` but it also modifies the `_args` too so I figured I do need to call self that is why I didn't use a `classmethod`.

EmilyBourne:
> > This looks better. I notice you didn't use `classmethod`s in `bitwise_operators.py`, but I think this should be ok as I can't think of a situation where we would internally generate a bitwise operator.
> > You should make sure you fix the codacy errors too
> 
> In the `_calculate_dtype` the method not only returns the `_dtype` and `_precision` but it also modifies the `_args` too so I figured I do need to call self that is why I didn't use a `classmethod`.

Yes I saw. The reason that we are making these functions `classmethod`s or `staticmethod`s is so we can call them from `__new__` (where we won't have access to `self`), but this functionality will only be needed for the classes that will be simplified. If we don't simplify bitwise operators then what you did should be fine :+1:

pyccel/ast/bitwise_operators.py
```diff
@@ -78,25 +79,30 @@ def _set_dtype(self):
...
 
     def _handle_integer_type(self, integers):
-        self._dtype     = NativeInteger()
-        self._precision = max(a.precision for a in integers)
+        _dtype     = NativeInteger()
+        _precision = max(a.precision for a in integers)
         self._args = [PythonInt(a) if a.dtype is NativeBool() else a for a in integers]
+        return _dtype, _precision
 
 #==============================================================================
```
EmilyBourne, line 107: I think you missed `PyccelBitComparisonOperator`

abkhaoula: right, I did it on [9742d51](https://github.com/pyccel/pyccel/pull/797/commits/9742d511bef41e1cb61d7f57ba7340b159527d1d)

pyccel/ast/bitwise_operators.py
```diff
@@ -68,7 +69,7 @@ class PyccelBitOperator(PyccelOperator):
     _rank = 0
     _shape = ()
```
EmilyBourne, line 70: I think these class variables need to be set in `_calculate_shape_rank`

EmilyBourne, line 70: These class objects can be removed now

EmilyBourne:
Don't forget to merge the master branch in. Try and run any tests you think are relevant locally to see if everything is working. E.g. here:
```
pytest test_arrays.py
pytest test_bitwise.py
```

Here are a few useful `pytest` flags:
- `-k test_name` runs any tests with `test_name` in their name
- `-x` pytest exits if any tests fail
- `-v` run pytest in verbose mode. This will print the name of each test function
- `-s` outputs any prints to the terminal rather than hiding them unless there is an error
- `--lf` Run the "last failed" tests first
- `-m language` Run all tests which are marked as language

pyccel/ast/bitwise_operators.py
```diff
@@ -40,14 +40,15 @@ class PyccelInvert(PyccelUnaryOperator):
...
         if a.dtype not in (NativeInteger(), NativeBool()):
-            raise TypeError('unsupported operand type(s): {}'.format(self))
+            raise TypeError('unsupported operand type(s): {}'.format(_args))
 
         self._args      = (PythonInt(a) if a.dtype is NativeBool() else a,)
-
-        self._precision = a.precision
+        _precision = a.precision
+        _dtype = a.dtype
+        return _dtype, _precision
```
EmilyBourne, line 51: You have the wrong dtype here

EmilyBourne: Ah sorry, you are right. The dtype is set after the casting so it is correct. The class property is no longer necessary though

EmilyBourne (changes requested):
This looks nearly right for the first point. There are just a couple of discussions above which have not been addressed and you need to merge in the master branch

pyccel/ast/bitwise_operators.py
abkhaoula: I added the class property because I saw that the `_args` may get updated `if a.dtype is NativeBool()` . I don't know if I understood it wrong ...

EmilyBourne: The class property is the property defined at line 41

EmilyBourne: The `dtype` is always `NativeInteger` here. If the argument is an integer then it remains the same, if it is a boolean then it is cast to an integer. So the args are updated but the final result is always an integer.

What you have done is fine. But you need to remove line 41.

EmilyBourne: > done in cab4c95

I think this is the wrong commit message, I just see a removed print

EmilyBourne: Ah no sorry there is a problem with the dtype, you are collecting the dtype from the original argument, not the corrected argument

EmilyBourne (approved):

pyccel/parser/syntactic.py
```diff
@@ -499,13 +499,13 @@ def _visit_BinOp(self, stmt):
         second = self._visit(stmt.right)
 
         if isinstance(stmt.op, ast.Add):
-            return PyccelAdd(first, second)
+            return PyccelAdd(first, second, simplify = True)
```
EmilyBourne, line 502: Only things in the codegen stage, and maybe the semantic stage should be simplified. The aim is to preserve the user's code where possible. So if the user writes:
```
a = x+(-5)
```
then maybe he has a reason for this strange syntax, so we want to print:
```
a = x+(-5)
```

but if the user writes:
```
a = randint(-x, y)
```
there isn't an equivalent for this function in fortran. Currently this means that we print:
```
a = floor((y - (-x)) * Dummy_0001 + (-x), kind=C_INT64_T)
```
but in this case the strange syntax is our fault. So it should be simplified to:
```
a = floor((y + x) * Dummy_0001 - x, kind=C_INT64_T)
```

pyccel/ast/operators.py
```diff
@@ -490,14 +490,17 @@ class PyccelAdd(PyccelArithmeticOperator):
     """
     _precedence = 12
 
-    def __new__(cls, arg1, arg2):
+    def __new__(cls, arg1, arg2, simplify = False):
+        if simplify:
+            if isinstance(arg2, PyccelUnarySub):
+                return PyccelMinus(arg1, arg2.args[0], simplify = True)
```
EmilyBourne, line 496: It would also be nice to simplify if both arguments are literals. I.e.:
```
if isinstance(arg1, Literal) and isinstance(arg2, Literal):
    dtype, precision = _calculate_dtype(arg1, arg2)
    if dtype is NativeReal():
        return LiteralFloat(arg1.python_value+arg2.python_value, precision=precision)
    elif ...:
        etc
```

EmilyBourne: That way we can simplify `a = np.empty(10)` from the current print `allocate(a(0:10_C_LONG_LONG - 1_C_LONG_LONG))` to `allocate(a(0:9_C_LONG_LONG))`

pyccel/ast/operators.py
```diff
@@ -559,6 +567,14 @@ def __new__(cls, arg1, arg2, simplify = False):
...
             if isinstance(arg2, PyccelUnarySub):
                 return PyccelAdd(arg1, arg2.args[0], simplify = True)
+            elif isinstance(arg1, Literal) and isinstance(arg2, Literal):
+                dtype, precision = cls._calculate_dtype(arg1, arg2)
+                if dtype is NativeInteger():
+                    return LiteralInteger(arg1.python_value - arg2.python_value, precision=precision)
+                if dtype is NativeReal():
+                    return LiteralFloat(arg1.python_value - arg2.python_value, precision=precision)
+                if dtype is NativeComplex():
+                    return LiteralComplex(arg1.python_value - arg2.python_value, precision=precision)
```
EmilyBourne, line 577: Codacy is correct that `LiteralComplex` does not have a constructor which takes just a complex. It is up to you whether you add this constructor, or you modify the call here to:
```
result = arg1.python_value - arg2.python_value
return LiteralComplex(result.real, result.imag, precision=precision)
```

pyccel/codegen/printing/ccode.py
```diff
@@ -430,14 +430,14 @@ def _print_LiteralComplex(self, expr):
             return self._print(PyccelAssociativeParenthesis(PyccelMul(expr.imag, LiteralImaginaryUnit())))
         else:
             return self._print(PyccelAssociativeParenthesis(PyccelAdd(expr.real,
-                            PyccelMul(expr.imag, LiteralImaginaryUnit()))))
+                            PyccelMul(expr.imag, LiteralImaginaryUnit()), simplify = True)))
```
EmilyBourne, line 433: If you do not simplify `PyccelMul` then this `simplify` will have no effect. However if you added a `simplify` to `PyccelMul` to generate the appropriate `LiteralComplex` then the expression would correctly simplify to a `LiteralComplex`. But (big but), this is **dangerous**. If you simplify here, you could cause an infinite loop. We are trying to print a `LiteralComplex`. In order to do that in c we need to print:
```
real + imag * _Complex_I
```
If you simplify here then you will get a `LiteralComplex` which is printed by `_print_LiteralComplex` which needs to print at this line which simplifies to a `LiteralComplex` which is printed by `_print_LiteralComplex` etc

pyccel/codegen/printing/ccode.py
```diff
@@ -430,14 +430,14 @@ def _print_LiteralComplex(self, expr):
...
 
     def _print_PythonComplex(self, expr):
         if expr.is_cast:
             value = self._print(expr.internal_var)
         else:
             value = self._print(PyccelAssociativeParenthesis(PyccelAdd(expr.real,
-                            PyccelMul(expr.imag, LiteralImaginaryUnit()))))
+                            PyccelMul(expr.imag, LiteralImaginaryUnit()), simplify = True)))
```
EmilyBourne, line 440: This is similar to above. In order for anything to happen you also need to simplify `PyccelMul`. Here there is no risk of an infinite loop as it will go from `_print_PythonComplex` to `_print_LiteralComplex`, but we also don't gain anything by doing this

pyccel/ast/operators.py
```diff
@@ -494,6 +494,14 @@ def __new__(cls, arg1, arg2, simplify = False):
         if simplify:
             if isinstance(arg2, PyccelUnarySub):
                 return PyccelMinus(arg1, arg2.args[0], simplify = True)
+            elif isinstance(arg1, Literal) and isinstance(arg2, Literal):
+                dtype, precision = cls._calculate_dtype(arg1, arg2)
+                if dtype is NativeInteger():
+                    return LiteralInteger(arg1.python_value + arg2.python_value, precision=precision)
```
EmilyBourne, line 500: We should be careful here. Currently to tell if an integer is negative we check if `isinstance(a, PyccelUnarySub)`, so a `LiteralInteger` should always be positive (we need to add a check for this to enforce it). I think we should add a test here so:
```
result = arg1.python_value + arg2.python_value
if result >= 0:
    return LiteralInteger(result, precision=precision)
else:
    return PyccelUnarySub(LiteralInteger(-result, precision=precision))
```

pyccel/ast/operators.py
```diff
@@ -559,6 +567,14 @@ def __new__(cls, arg1, arg2, simplify = False):
         if simplify:
             if isinstance(arg2, PyccelUnarySub):
                 return PyccelAdd(arg1, arg2.args[0], simplify = True)
+            elif isinstance(arg1, Literal) and isinstance(arg2, Literal):
+                dtype, precision = cls._calculate_dtype(arg1, arg2)
+                if dtype is NativeInteger():
+                    return LiteralInteger(arg1.python_value - arg2.python_value, precision=precision)
```
EmilyBourne, line 573: See above. `PyccelUnarySub` may be needed

pyccel/codegen/printing/fcode.py
```diff
@@ -980,9 +980,9 @@ def _print_MacroCount(self, expr):
                             shape.append(i.stop)
                     elif i.stop is None:
                         if (isinstance(i.start, (int, LiteralInteger)) and i.start<s-1) or not(isinstance(i.start, (int, LiteralInteger))):
-                            shape.append(PyccelMinus(s, i.start))
+                            shape.append(PyccelMinus(s, i.start), simplify = True)
```
EmilyBourne, line 983: I think this simplify is in the wrong set of parentheses

pyccel/codegen/printing/fcode.py
```diff
@@ -1740,11 +1740,11 @@ def _print_AugAssign(self, expr):
         like   = expr.like
 
         if isinstance(op, AddOp):
-            rhs = PyccelAdd(lhs, rhs)
+            rhs = PyccelAdd(lhs, rhs, simplify = True)
         elif isinstance(op, MulOp):
             rhs = PyccelMul(lhs, rhs)
         elif isinstance(op, SubOp):
-            rhs = PyccelMinus(lhs, rhs)
+            rhs = PyccelMinus(lhs, rhs, simplify = True)
```
EmilyBourne, line 1747: These functions handle the `AugAssign` operator. The `AugAssign` operator handles the operators `+=`, `-=`, `*=`, etc.
As this is an assignment the `lhs` cannot be a `Literal`, so I don't think it is worth simplifying these expressions

pyccel/codegen/printing/fcode.py
```diff
@@ -1765,13 +1765,13 @@ def _print_PythonRange(self, expr):
         # testing if the step is a value or an expression
         if isinstance(test_step, Literal):
             if isinstance(expr.step, PyccelUnarySub):
-                stop = PyccelAdd(expr.stop, LiteralInteger(1))
+                stop = PyccelAdd(expr.stop, LiteralInteger(1), simplify = True)
             else:
-                stop = PyccelMinus(expr.stop, LiteralInteger(1))
+                stop = PyccelMinus(expr.stop, LiteralInteger(1)), simplify = True
```
EmilyBourne, line 1770: The simplify is the wrong side of the parentheses here

EmilyBourne (changes requested):
Hey, the code that is generated now looks lots cleaner this is a big improvement. Thanks :)

There are a few errors in your code which need fixing. You also need to merge the master into your code  (this is why the tests are not running). When you merge:
- pyccel/ast/bitwise_operators.py
- pyccel/ast/operators.py

there are a couple of pylint remarks at the top of the files which mention this PR. Your fixes should fix the problems that they're suppressing so please remove these comments. These files should be reasonably easy to merge, but if the conflicts are confusing send me a message and I'll help you work out which version needs to be kept :+1:

I also found an extra place where the expression should be simplified. It is [here](https://github.com/pyccel/pyccel/blob/master/pyccel/codegen/printing/fcode.py#L1129-L1142). This code predates the introduction of the `Literal` class so we have forgotten to correct the code and handle the operation with `PyccelMinus` and `LiteralInteger`. `PyccelMinus` with a `simplify` needs adding here.

Here is an example illustrating the problem:
```
@stack_array('x')
def f():
    import numpy as np
    x = np.zeros(3)
```

pyccel/ast/operators.py
abkhaoula: done in [d289729](https://github.com/pyccel/pyccel/pull/797/commits/d289729fd8de96a2f0200dea18fbb6e770bbf690)

pyccel/ast/operators.py
abkhaoula: done in [d289729](https://github.com/pyccel/pyccel/pull/797/commits/d289729fd8de96a2f0200dea18fbb6e770bbf690)

pyccel/ast/operators.py
```diff
@@ -568,12 +576,20 @@ def __new__(cls, arg1, arg2, simplify = False):
         if simplify:
             if isinstance(arg2, PyccelUnarySub):
                 return PyccelAdd(arg1, arg2.args[0], simplify = True)
-            elif isinstance(arg1, Literal) and isinstance(arg2, Literal):
+            else:
```
EmilyBourne: Why did you remove this check?

abkhaoula: If I understand it right, this check insures that arg1 and arg2 are positive. I see there is no use to it as long as we are checking later if the result is positive or negative.

abkhaoula: I understand I should have it back (the tests don't work unless I have this check), I ll try to understand why

EmilyBourne: > If I understand it right, this check insures that arg1 and arg2 are positive

No, this checks if arg1 and arg2 are literal values. That means it is a string, a (positive) float, a (positive) int, a complex or a boolean. But if you don't check then you also have to handle types such as `Variable`, `PyccelMod`, `NumpyRand`, etc. Basically anything that can be added. These other classes don't have a `python_value` property as we don't know their value and that value may not even be constant

pyccel/ast/operators.py
```diff
@@ -2,8 +2,6 @@
 # This file is part of Pyccel which is released under MIT License. See the LICENSE file or #
 # go to https://github.com/pyccel/pyccel/blob/master/LICENSE for full license details.     #
 #------------------------------------------------------------------------------------------#
-# TODO [EB 12.03.21]: Remove pylint command with PR #797
-# pylint: disable=W0201
```
EmilyBourne, line 6: This command also needs removing in bitwise_operators.py

pyccel/ast/operators.py
```diff
@@ -206,7 +204,8 @@ def __init__(self, arg):
     @staticmethod
     def _calculate_dtype(*_args):
         """ Sets the dtype and precision
-        They are chosen to match the argument
+        They are chosen to match the argument unless the class has
+        a _dtype or _precision member
```
EmilyBourne: Is this new or is it undoing corrections that you already did? I'm not sure this description is still accurate

pyccel/ast/bitwise_operators.py
```diff
@@ -41,16 +41,18 @@ class PyccelInvert(PyccelUnaryOperator):
     """
     __slots__ = ()
     _precedence = 14
+    _dtype     = NativeInteger()
 
-    def _set_dtype(self):
-        self._dtype     = NativeInteger()
-        a = self._args[0]
+    def _calculate_dtype(self, *_args):
+        _dtype = NativeInteger()
```
EmilyBourne, line 47: You should either have a class variable which you return:
```
class PyccelInvert(..):
    ...
    _dtype = NativeInteger()
    
    def _calculate_dtype(self, *_args):
         dtype = self._dtype
         ....
         return dtype, _precision
```
Or you need to remove the class variable

pyccel/ast/operators.py
```diff
@@ -115,8 +113,8 @@ def __init__(self, *args):
         if self.stage == 'syntactic':
             super().__init__()
             return
-        self._set_dtype()
-        self._set_shape_rank()
+        self._dtype, self._precision = self._calculate_dtype(*self._args)
```
EmilyBourne, line 116: I think this pylint error can be deactivated with a comment. `PyccelOperator` doesn't have this function as it is an abstract class. The only other way of doing this would be to define the functions but have them do nothing but raise an error

EmilyBourne (changes requested):
As mentioned before:
> I also found an extra place where the expression should be simplified. It is here. This code predates the introduction of the Literal class so we have forgotten to correct the code and handle the operation with PyccelMinus and LiteralInteger. PyccelMinus with a simplify needs adding here.
> 
> Here is an example illustrating the problem:
>
> @stack_array('x')
> ```
> def f():
>     import numpy as np
>     x = np.zeros(3)
> ```

pyccel/ast/operators.py
```diff
@@ -2,6 +2,8 @@
 # This file is part of Pyccel which is released under MIT License. See the LICENSE file or #
 # go to https://github.com/pyccel/pyccel/blob/master/LICENSE for full license details.     #
 #------------------------------------------------------------------------------------------#
+# TODO [EB 12.03.21]: Remove pylint command with PR #797
+# pylint: disable=W0201
```
EmilyBourne: This command needs removing

pyccel/ast/operators.py
abkhaoula: fixed in [f09813c](https://github.com/pyccel/pyccel/pull/797/commits/f09813cc5a2ea6be686a649e8feeb2ec1abab23a)

pyccel/ast/operators.py
abkhaoula: I don't see this command in bitwise_operators.py

pyccel/ast/bitwise_operators.py
abkhaoula: done [f29d118](https://github.com/pyccel/pyccel/pull/797/commits/f29d1184964b4a001e57e7ad97cebec75c31d550)

pyccel/ast/operators.py
EmilyBourne: The error should only be deactivated at this line, not for the whole file

pyccel/ast/bitwise_operators.py
EmilyBourne: You still need to specify `self` to access this variable. See the pylint error

pyccel/ast/operators.py
```diff
@@ -106,7 +106,7 @@ class PyccelOperator(PyccelAstNode):
     args: tuple
         The arguments passed to the operator
     """
-    __slots__ = ('_args','_dtype','_precision','_shape','_rank','_order')
+    __slots__ = ('_args','_order', )
```
EmilyBourne, line 109: `_order` can probably be treated the same way as `_dtype`, `_precision`, `_shape` and `_rank`. It tells us if an array is saved in Fortran ordering (column-major) or C ordering (row-major) (see [here](https://en.wikipedia.org/wiki/Row-_and_column-major_order) for more details). It should be set to `None` if `_rank`<2 so anywhere the `_rank` is a class property probably `_order` can be too

pyccel/ast/operators.py
```diff
@@ -124,6 +124,12 @@ def __init__(self, *args):
             self._order = None
         super().__init__()
 
+    def _set_dtype(self):
+        self._dtype, self._precision = self._calculate_dtype(*self._args)
```
EmilyBourne, line 128: This codacy error can be disabled locally by writing:
```
self._dtype, self._precision = self._calculate_dtype(*self._args) # pylint: disable=no-member
```

pyccel/ast/operators.py
```diff
@@ -124,6 +124,12 @@ def __init__(self, *args):
...
+
+    def _set_shape_rank(self):
+        self._shape, self._rank = self._calculate_shape_rank(*self._args)
```
EmilyBourne, line 131: As above this codacy error can be disabled locally

pyccel/ast/operators.py
EmilyBourne: I am happy to leave this improvement for another PR though as it may require significantly more code?

pyccel/ast/operators.py
```diff
@@ -488,24 +511,46 @@ class PyccelAdd(PyccelArithmeticOperator):
     arg2: PyccelAstNode
         The second argument passed to the operator
     """
-    __slots__ = ()
+    __slots__ = ('_dtype','_precision',)
```
EmilyBourne, line 514: Classes inherit slots from the classes that they inherit from so `PyccelAdd` already has a `_dtype` and `_precision` as they are inherited from `PyccelArithmeticOperator` which inherits them from `PyccelBinaryOperator`

pyccel/ast/operators.py
```diff
@@ -585,12 +650,17 @@ class PyccelDiv(PyccelArithmeticOperator):
     arg2: PyccelAstNode
         The second argument passed to the operator
     """
-    __slots__ = ()
+    __slots__ = ('_dtype','_precision')
```
EmilyBourne, line 653: See above. These slots are inherited

pyccel/ast/operators.py
```diff
@@ -654,12 +724,13 @@ class PyccelComparisonOperator(PyccelBinaryOperator):
     arg2: PyccelAstNode
         The second argument passed to the operator
     """
-    __slots__ = ()
+    __slots__ = ('_dtype','_precision',)
```
EmilyBourne, line 727: See above. These slots are inherited

pyccel/ast/operators.py
```diff
@@ -796,15 +867,18 @@ class PyccelBooleanOperator(PyccelOperator):
     arg2: PyccelAstNode
         The second argument passed to the operator
     """
-    __slots__ = ()
-    _default_precision = default_precision['bool']
+    __slots__ = ('_dtype','_precision','_shape','_rank',)
```
EmilyBourne, line 870: The `_dtype`, `_precision`, `_shape`, and `_rank` don't need to be slots. They are always equal to `NativeBool()`, `default_precision['bool']`, `()`, and `0` so these should be class properties. The setters should then be overridden with:
```
def _set_dtype(self):
    pass
def _set_shape_rank(self):
    pass
```

pyccel/ast/bitwise_operators.py
```diff
@@ -68,9 +66,8 @@ class PyccelBitOperator(PyccelOperator):
     arg2: PyccelAstNode
         The second argument passed to the operator
     """
-    __slots__ = ()
```
EmilyBourne, line 71: All classes should define `__slots__` even if it contains nothing

EmilyBourne: In this class the `__slots__` should be `_precision`. The `_dtype`, `_shape` and `_rank` are all the same for all instances of the class so they should be class properties

EmilyBourne (changes requested):
As mentioned before:
> I also found an extra place where the expression should be simplified. It is here. This code predates the introduction of the Literal class so we have forgotten to correct the code and handle the operation with PyccelMinus and LiteralInteger. PyccelMinus with a simplify needs adding here.
> 
> Here is an example illustrating the problem:
>
> @stack_array('x')
> ```
> def f():
>     import numpy as np
>     x = np.zeros(3)
> ```

pyccel/ast/operators.py
abkhaoula: I tried doing this in [6d59098](https://github.com/pyccel/pyccel/pull/797/commits/6d59098c7b9b632c977a4c8d7d5321e3fe7fc58e) but it generated some problems so i retrieved it in [9cd1fc1](https://github.com/pyccel/pyccel/pull/797/commits/9cd1fc168b74f0f27d9bfc0be5dc22f93bb61f0e)

EmilyBourne: The reason it didn't work in 6d59098 is because you didn't set the class property. You need to make sure that `self._dtype` returns something for each class. Either it will return the class property (same object for every instance) or it will return the attribute set in `_set_dtype`. It looks like in 6d59098 you removed it from `_set_dtype` but hadn't yet added it as a class property. Hence the error (I imagine it failed in one of the methods of `PyccelAstNode` which tries to return this property)

pyccel/ast/bitwise_operators.py
```diff
@@ -68,9 +66,10 @@ class PyccelBitOperator(PyccelOperator):
...
-
-    def _set_dtype(self):
+    _shape = ()
+    _rank = 0
+    __slots__ = ('_dtype','_precision','_order')
```
EmilyBourne, line 71: If `_shape=()` then `_order = None`

pyccel/ast/operators.py
```diff
@@ -796,15 +867,18 @@ class PyccelBooleanOperator(PyccelOperator):
...
         The second argument passed to the operator
     """
-    __slots__ = ()
-    _default_precision = default_precision['bool']
+    _dtype = NativeBool()
+    _precision = default_precision['bool']
+    _rank = 0
+    _shape = ()
+
+    __slots__ = ('_order')
```
EmilyBourne, line 875: As above `_order = None` so this doesn't need to be a slot

EmilyBourne (changes requested):
In addition to the place mentioned in my last few reviews, there are also a few places that need simplify that were introduced in PR #833. They are marked with a TODO

pyccel/ast/variable.py
```diff
@@ -817,13 +817,12 @@ def __getitem__(self, *args):
...
             if isinstance(i, Slice) and j<len(args):
-                # TODO: Replace with simplify = True in PR 797
                 if j == 0:
                     i = args[j]
                 elif i.step == 1:
-                    i = PyccelAdd(i.start, args[j])
+                    i = PyccelAdd(i.start, args[j], simplify = True)
                 else:
-                    i = PyccelAdd(i.start, PyccelMul(i.step, args[j]))
+                    i = PyccelAdd(i.start, PyccelMul(i.step, args[j]), simplify = True)
```
EmilyBourne, line 825: Do we also have a `simplify = True` for `PyccelMul` to handle `1*x`?

EmilyBourne (changes requested):
There is still one more place where `simplify = True` should be added. I already mentioned it twice above.

> I also found an extra place where the expression should be simplified. It is [here](https://github.com/pyccel/pyccel/blob/master/pyccel/codegen/printing/fcode.py#L1129-L1142). This code predates the introduction of the `Literal` class so we have forgotten to correct the code and handle the operation with `PyccelMinus` and `LiteralInteger`. `PyccelMinus` with a `simplify` needs adding here.
>
> Here is an example illustrating the problem:
> ```
>     @stack_array('x')
>     def f():
>         import numpy as np
>         x = np.zeros(3)
>     ```

EmilyBourne:
Could you also simplify the expressions [here](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/variable.py#L767-L779) please. Without this the shape of an indexed object is quite complicated which can cause problems

pyccel/ast/variable.py
abkhaoula: Should I do the same for `PyccelDiv`

EmilyBourne: Why not. I don't think it's used as much but it can't do any harm :+1:

pyccel/ast/operators.py
```diff
@@ -575,6 +575,22 @@ class PyccelMul(PyccelArithmeticOperator):
     __slots__ = ()
     _precedence = 13
 
+    def __new__(cls, arg1, arg2, simplify = False):
+        print("######################################")
+        print(arg1)
+        print(arg2)
+        print("######################################")
```
EmilyBourne, line 582: Forgotten prints

pyccel/codegen/printing/ccode.py
```diff
@@ -470,17 +470,17 @@ def _print_Literal(self, expr):
 
     def _print_LiteralComplex(self, expr):
         if expr.real == LiteralFloat(0):
-            return self._print(PyccelAssociativeParenthesis(PyccelMul(expr.imag, LiteralImaginaryUnit())))
+            return self._print(PyccelAssociativeParenthesis(PyccelMul(expr.imag, LiteralImaginaryUnit(), simplify = True)))
         else:
             return self._print(PyccelAssociativeParenthesis(PyccelAdd(expr.real,
-                            PyccelMul(expr.imag, LiteralImaginaryUnit()))))
+                            PyccelMul(expr.imag, LiteralImaginaryUnit(), simplify = True))))
```
EmilyBourne, line 476: This shouldn't be simplified. You cannot get simpler than a LiteralComplex so you end up with an infinite loop.
E.g: consider `LiteralComplex(1,2)`
Here `expr=LiteralComplex(1,2)`
We want to use `PyccelAssociativeParenthesis`, `PyccelMul`, and `LiteralImaginaryUnit` to print it as `(1+2j)` but this expression is then simpilified to `LiteralComplex(1,2)` which we try to print using this function and so on

pyccel/ast/variable.py
```diff
@@ -765,15 +765,15 @@ def __init__(
...
-                        stop = PyccelAdd(s, stop)
+                        stop = PyccelAdd(s, stop) # cannot simplify
 
-                    _shape = stop if start is None else PyccelMinus(stop, start, simplify=True)
+                    _shape = stop if start is None else PyccelMinus(stop, start) # cannot simplify
                     if step is not None:
                         if isinstance(step, PyccelUnarySub):
                             start = s if a.start is None else start
-                            _shape = start if a.stop is None else PyccelMinus(start, stop, simplify=True)
+                            _shape = start if a.stop is None else PyccelMinus(start, stop) # cannot simplify
```
EmilyBourne: Can you explain why these can't be simplified? I can't see anything obvious

pyccel/codegen/printing/ccode.py
```diff
@@ -480,7 +480,7 @@ def _print_PythonComplex(self, expr):
             value = self._print(expr.internal_var)
         else:
             value = self._print(PyccelAssociativeParenthesis(PyccelAdd(expr.real,
-                            PyccelMul(expr.imag, LiteralImaginaryUnit()))))
+                            PyccelMul(expr.imag, LiteralImaginaryUnit(), simplify = True))))
```
EmilyBourne, line 483: I'm not sure this adds anything

EmilyBourne (approved):

pyccel/ast/variable.py
EmilyBourne: Did you get rid of this due to the errors [here](https://github.com/pyccel/pyccel/runs/2326040155) ? I've just had a look at what is generated. It's not a problem with your code. It's a problem with how we're printing the shape and how we're handling the shape when it's reassigned. It would be better to mark the test as xfail I think. I would open an issue for it, but the error is hidden without this improvement

EmilyBourne: To be clear the test is translating:
```
def reassigned_pointer():
    from numpy import array, shape
    a = array([1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
    c = a[1:]
    c[1] = 0
    c = a
    return c[0], c[1], c[2], shape(c)[0]
```
In this code the shape of `c` changes. First it is `shape(a)-1=10-1=9`, then it is `shape(a)=10`. Before the improvement above we had `shape(c) = (10-1)` so we decided it was too complicated to print the shape, and we printed `c.shape` instead (which means the code works even if we are using the wrong shape internally). After your fix, the shape is `9` which is simple enough that we print it as is and so the problem becomes apparent

abkhaoula: It was this that generated the last test fails in my last commit.
Which was the case in `test_arrays_view` function in `tests/pyccel/test_pyccel.py`
while trying to view the array in this function in fortran

```python
def array_view_3():
    a = np.array([1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
    b = a[1:-1]
    c = b[1:-1]
    c = a
    for i in range(np.shape(c)[0]):
        print(c[i])
```

without simplification the loop looks like this 
```fortran
do i = 0_C_INT64_T, size(c, kind=C_INT64_T) - 1_C_INT64_T, 1_C_INT64_T
  print *, c(i)
end do
```
but with simplification it looks like this
```fortran
do i = 0_C_INT64_T, 5_C_INT64_T, 1_C_INT64_T
  print *, c(i)
end do
```
Which is weird. It looks like it always take the size of c when it was first declared.

EmilyBourne: I found another way to reproduce the bug. See #847

EmilyBourne: > Which is weird. It looks like it always take the size of c when it was first declared.

That's exactly what's happening. Normally you can't change the shape of a `Variable` so there isn't a problem but when you reallocate a pointer you can change the shape

EmilyBourne: I think the best thing to do would be to keep this code and add an `xfail` mark to the failing test referencing issue #847. That way we already have a way to see if #847 is fixed

abkhaoula: ok

pyccel/codegen/printing/ccode.py
abkhaoula: I ll remove it

pyccel/ast/variable.py
EmilyBourne: This problem should be fixed with PR #851 . You can create a temporary branch combining this branch and branch devel-issue847 to check

abkhaoula: done in [simplified_nodes_devel_issue847](https://github.com/pyccel/pyccel/tree/simplified_nodes_devel_issue847)

EmilyBourne: Great if those tests pass and you're happy with the solution then you can approve PR #851

EmilyBourne (approved):
Great work! Thanks for this. This will be a big improvement to pyccel!

mostafaMamoni (approved):

tests/pyccel/test_pyccel.py
```diff
@@ -547,6 +547,7 @@ def test_c_arrays(language):
     pyccel_test("scripts/c_arrays.py", language=language, output_dtype=types)
 
 #------------------------------------------------------------------------------
+@pytest.mark.xfail(reason = 'issue #847: Shape of pointer is not updated when pointer is reassigned')
```
EmilyBourne, line 550: #847 has now been fixed in the master branch so this line can be removed

pyccel/ast/bitwise_operators.py
```diff
@@ -68,9 +66,14 @@ class PyccelBitOperator(PyccelOperator):
...
-
-    def _set_dtype(self):
+    _shape = ()
+    _rank = 0
+    _order = None
+    __slots__ = ('_dtype','_precision')
+
+    def _set_order(self):
+        pass
+    def _calculate_dtype(self, *_args):
```
yguclu, line 76: Is there a reason for calling this variable `_args` instead of simply `args` (without underscore)?

pyccel/ast/bitwise_operators.py
```diff
@@ -68,9 +66,14 @@ class PyccelBitOperator(PyccelOperator):
...
-    __slots__ = ()
-
-    def _set_dtype(self):
+    _shape = ()
+    _rank = 0
+    _order = None
+    __slots__ = ('_dtype','_precision')
+
+    def _set_order(self):
+        pass
```
yguclu, line 75: Please add an empty line after the end of this function

pyccel/ast/bitwise_operators.py
```diff
@@ -80,26 +83,27 @@ def _set_dtype(self):
...
-            raise TypeError('cannot determine the type of {}'.format(self))
+            raise TypeError('cannot determine the type of {}'.format(_args))
 
     def _set_shape_rank(self):
-        self._rank = 0
-        self._shape = ()
+        pass
+
 
     def _handle_integer_type(self, integers):
```
yguclu, line 102: Here you have 2 empty lines, but one is enough

pyccel/ast/bitwise_operators.py
```diff
@@ -80,26 +83,27 @@ def _set_dtype(self):
...
-        self._dtype     = NativeInteger()
-        self._precision = max(a.precision for a in integers)
+        _dtype     = NativeInteger()
+        _precision = max(a.precision for a in integers)
         self._args = [PythonInt(a) if a.dtype is NativeBool() else a for a in integers]
+        return _dtype, _precision
```
yguclu, line 106: Can we call these variables just `dtype` and `precision` (without underscore)?

pyccel/ast/bitwise_operators.py
```diff
@@ -163,13 +167,14 @@ class PyccelBitComparisonOperator(PyccelBitOperator):
...
         elif all(a.dtype is NativeBool() for a in integers):
-            self._dtype = NativeBool()
+            _dtype = NativeBool()
         else:
-            self._dtype = NativeInteger()
+            _dtype = NativeInteger()
             self._args = [PythonInt(a) if a.dtype is NativeBool() else a for a in integers]
-        self._precision = max(a.precision for a in integers)
+        _precision = max(a.precision for a in integers)
+        return _dtype, _precision
```
yguclu, line 177: Same comment about the variable names: `dtype` and `precision` do not need an underscore.

pyccel/ast/bitwise_operators.py
```diff
@@ -42,15 +40,15 @@ class PyccelInvert(PyccelUnaryOperator):
...
+        a = _args[0]
         if a.dtype not in (NativeInteger(), NativeBool()):
-            raise TypeError('unsupported operand type(s): {}'.format(self))
+            raise TypeError('unsupported operand type(s): {}'.format(_args))
 
         self._args      = (PythonInt(a) if a.dtype is NativeBool() else a,)
-
-        self._precision = a.precision
+        _precision = a.precision
+        return _dtype, _precision
```
yguclu, line 51: Can we just use the names `args`, `dtype` and `precision` without underscores?

yguclu (changes requested):
Great job @abkhaoula!

I've left a few comments around. Moreover:
- File `tests/pyccel/scripts/runtest_headers.py` is empty, what is it needed for?
- Please update the patch number in variable `__version__` in `pyccel/version.py`

EmilyBourne:
`tests/pyccel/scripts/runtest_headers.py` is generated automatically when we run `tests/pyccel/test_pyccel.py` maybe we should add it to the `.gitignore`?

yguclu:
> `tests/pyccel/scripts/runtest_headers.py` is generated automatically when we run `tests/pyccel/test_pyccel.py` maybe we should add it to the `.gitignore`?

Good suggestion. For @abkhaoula:
- Remove file `tests/pyccel/scripts/runtest_headers.py` from the repository
- Add the file's path to the `.gitignore` file

pyccel/ast/bitwise_operators.py
abkhaoula: I didn't know the underscore was reserved for private variables 👍🏻

EmilyBourne (approved):

abkhaoula:
> > `tests/pyccel/scripts/runtest_headers.py` is generated automatically when we run `tests/pyccel/test_pyccel.py` maybe we should add it to the `.gitignore`?
> 
> Good suggestion. For @abkhaoula:
> 
> * Remove file `tests/pyccel/scripts/runtest_headers.py` from the repository
> * Add the file's path to the `.gitignore` file

done in [d8f459f](https://github.com/pyccel/pyccel/pull/797/commits/d8f459f07ce63ca641f1123e2ba5121c782d1014)

Issue #705, opened 2021-01-25
Improve the bind_c files


saidctb:
The way we generate the `bind c` files, makes the compilation slow by the fortran compiler.
For this example
```python
def f(x:int, y:float):
    return x + y 
```
We generate the following `bind c` file
```fortran
function f(x, y) bind(c) result(Out_0001)

  use test, only: mod_f => f
  use, intrinsic :: ISO_C_BINDING
  implicit none

  integer(C_INT64_T), value :: x
  real(C_DOUBLE), value :: y
  real(C_DOUBLE) :: Out_0001
  
  Out_0001 = mod_f(x, y)
  
end function f
```
We would like to generate it in the following way
```fortran
module bind_c_test

  use, intrinsic :: ISO_C_BINDING
  use test

  implicit none

contains

  function bind_c_f(x, y) bind(c) result(Out_0001) 
    integer(C_INT64_T), value :: x
    real(C_DOUBLE), value :: y
    real(C_DOUBLE) :: Out_0001
  
    Out_0001 = f(x, y)
  
  end function bind_c_f

end module bind_c_test
```
The `c wrapper` file will have the following signature:
`double bind_c_f(int x, double y);`
instead of:
`double f(int x, double y);`

yguclu:
As a first step we can keep the names that we have, and just make the minimum number of changes:

1. Put all functions inside an enclosing module called `bind_c_[...]`;
2. Collect all the `use` statements outside of the functions.

PR #800, opened 2021-03-09
Put all imports at the top of the file instead of one in each function

EmilyBourne:
This is the first step to fixing #705

### DIFF
--- a/pyccel/ast/bind_c.py
+++ b/pyccel/ast/bind_c.py
@@ -4,7 +4,7 @@
 # go to https://github.com/pyccel/pyccel/blob/master/LICENSE for full license details.     #
 #------------------------------------------------------------------------------------------#
 
-from pyccel.ast.core import FunctionCall
+from pyccel.ast.core import FunctionCall, Module
 from pyccel.ast.core import FunctionAddress
 from pyccel.ast.core import FunctionDef, BindCFunctionDef
 from pyccel.ast.core import Assign
@@ -13,9 +13,10 @@ from pyccel.ast.core import AsName
 from pyccel.ast.variable import Variable
 
 __all__ = (
-   'as_static_function',
-   'as_static_function_call',
-   'sanitize_arguments',
+    'as_static_module',
+    'as_static_function',
+    'as_static_function_call',
+    'sanitize_arguments',
 )
 
 #=======================================================================================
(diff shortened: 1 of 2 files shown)


pyccel/ast/bind_c.py
```diff
@@ -114,21 +115,61 @@ def as_static_function(func, name=None):
...
+    ==========
+    func     : FunctionDef
+               The function to be translated
+    mod_name : str
+               The name of the module which contains func
+    name     : str
+               The new name of the function
+    imports  : list
+               An optional parameter into which any required imports
+               can be collected
```
yguclu, line 156: If `imports` is not passed this function will behave (almost) as before?

pyccel/codegen/printing/fcode.py
```diff
@@ -1494,7 +1494,6 @@ def _print_BindCFunctionDef(self, expr):
         interfaces = '\n'.join(self._print(i) for i in expr.interfaces)
         arg_code  = ', '.join(self._print(i) for i in chain( arguments, results ))
         imports   = ''.join(self._print(i) for i in expr.imports)
-        imports += 'use, intrinsic :: ISO_C_BINDING'
```
yguclu, line 1497: I thought that we needed this in order to get the integer kind type parameters. What has changed?

pyccel/ast/bind_c.py
yguclu: I've made a quick search, and this function is called without the `imports` argument in `cwrappercode.py`. Can you have a look at it?

EmilyBourne: It's just used to get the function signature to be able to print it at the top of the `X_wrapper.c` file. It's definitely overkill to create a full `FunctionDef` just for that though, we should be calling `as_static_function` or better `as_static_function_signature`

EmilyBourne: Actually the cwrapper really ought to just operate on the created module rather than calling these functions again

EmilyBourne: I'm not sure this is worth correcting now thought as @MenoIy is rewriting lots of the cwrapper. I suspect it will be easier to do this afterwards

EmilyBourne: The implementation is not trivial in the case of static functions if we want to avoid repetition. I think I will add a `as_static_function_signature` for now and we can always improve it later

EmilyBourne: See issue #801

yguclu: Thanks!

PR #807, opened 2021-03-16
Speed up unit tests by running them in parallel with pytest-xdist

### LINKED ISSUE #748: Speed up unit tests using many processes
According to [this page](https://docs.github.com/en/actions/reference/specifications-for-github-hosted-runners#supported-runners-and-hardware-resources) in the documentation of GitHub Actions, the virtual machines have 2 or 3 cores available. We should take advantage of that to speed up our unit tests.

It is possible to run `pytest` with multiple sub-processes by using the [pytest-xdist](https://pypi.org/project/pytest-xdist) plugin. This is currently our preferred strategy, but we are open to other suggestions.

When running our tests in parallel the `teardown` function can lead to the `__pyccel__` folder being deleted by one thread while another is still using it. To fix this we should:
- Increase (a lot) the size of the random string used by epyccel (see #188 )
- And maybe improve the random seed
- Do not clean up the temporary directories when pytest is being run with xdist (the rest of the time it should teardown to avoid polluting developers' computers)

In addition a `pyccel-clean` script must be added in order to clean up all `__epyccel__` and `__pyccel__` folders

### LINKED ISSUE #811: Sometimes epyccel/test_parallel_epyccel.py failed in Linux platform.
you can see the error trace from github actions [here](https://github.com/pyccel/pyccel/pull/719/checks?check_run_id=2135996682#step:8:375).

EmilyBourne:
Install pytest-xdist and use it to run unit tests with as many processes as possible. Fixes #748 

**Commit Summary**
- Install pytest-xdist
- Run tests with option `-n auto`
- Add `pyccel_clean.py` file
- Add `pyccel-clean` script
- Only copy and recompile internal libraries if they are not already present
- Ensure that all generated files contain a tag (this was not the case when translating modules)
- Increase the size of the random string (seed is not changed as it has no effect for `random.SystemRandom()`)
- Remove duplicate function `random_string` from `parser/utilities.py` where it was unused
- Only teardown tests when running in serial
- Run parallel tests in a separate section
- invalidate caches before importing newly created file (this was done for the serial case, and is added for mpi cases). Fixes #811 
- Add `xdist_incompatible` pytest marker. This is necessary as some pyccel tests reuse files. pyccel does not mangle names in order to allow importing so this causes name clashes when running in parallel
- Remove tests/printers folder. This contains tests for printing sympy objects with the pyccel printers which no longer makes any sense. Occasionally pytest tries to parse these files despite the `--ignore` flag and fails


tests/pyccel/test_pyccel.py
```diff
@@ -235,36 +225,10 @@ def pyccel_test(test_file, dependencies = None, compile_with_pyccel = True,
...
-        elif not os.path.isfile(file_name):
-            teardown(file_name)
-        elif not f.endswith(".py") and not f.endswith(".pyh") and not f.endswith(".pyccel"):
-            os.remove(file_name)
-
 #==============================================================================
 # UNIT TESTS
 #==============================================================================
+@pytest.mark.xdist_incompatible
 def test_relative_imports_in_project(language):
```
yguclu, line 232: Why is this test incompatible with `pytest-xdist`? It creates its own temporary directories, doesn't it?

tests/pyccel/test_pyccel.py
```diff
@@ -280,6 +244,7 @@ def test_relative_imports_in_project(language):
     compare_pyth_fort_output(pyth_out, fort_out)
 
 #------------------------------------------------------------------------------
+@pytest.mark.xdist_incompatible
 def test_absolute_imports_in_project(language):
```
yguclu, line 248: I have the same question about this test

pyccel/codegen/python_wrapper.py
```diff
@@ -119,7 +119,7 @@ def create_shared_library(codegen,
 
         setup_code = create_c_setup(sharedlib_modname, wrapper_filename,
                 dep_mods, compiler, includes, libs + extra_libs, libdirs + extra_libdirs, c_flags)
-        setup_filename = "setup_{}.py".format(module_name)
```
yguclu, line 122: Why changing the name of the setup file, if the name of the bind_c and wrapper files are not changed?

pyccel/codegen/pipeline.py
```diff
@@ -325,20 +330,34 @@ def handle_error(stage):
...
+                    internal_libs_name.add(lib)
+                    # add source file without extension to internal_libs_files
+                    internal_libs_files.extend(os.path.splitext(f)[0] for f in source_files)
+                    # add library path to internal_libs_path
+                    internal_libs_path.append(lib_dest_path)
+
+                    # stop if object files already exist
+                    object_files = [s.replace(ext, obj_ext) for s in source_files]
+                    if all(os.path.exists(o) for o in object_files):
+                        continue
```
yguclu, line 360: Can you please explain these changes?

pyccel/codegen/pipeline.py
```diff
@@ -36,6 +36,11 @@
     "c" : ".c",
     "fortran": ".f90",
 }
+# map language to its object file extension
+lang_obj_ext_dict = {
+    "c" : ".o",
+    "fortran": ".mod",
+}
```
yguclu, line 43: Maybe the names are wrong here. A Fortran `.mod` file contains the kind of information that one finds in a C header file, and is definitely not an object file. Fortran object files normally end with `.o` just like in C.

pyccel/codegen/pipeline.py
```diff
@@ -325,20 +330,34 @@ def handle_error(stage):
...
                     # new one from pyccel stdlib
                     lib_dest_path = os.path.join(pyccel_dirpath, lib_name)
-                    if os.path.exists(lib_dest_path):
-                        shutil.rmtree(lib_dest_path)
-                    shutil.copytree(lib_path, lib_dest_path)
+                    if not os.path.exists(lib_dest_path):
+                        try:
+                            shutil.copytree(lib_path, lib_dest_path)
+                        except FileExistsError:
+                            pass
```
yguclu, line 337: Are you trying to copy only the minimum number of files here? If yes:

- do you see a significant speedup in doing so?
- the two lines of comment just above should be updated.

.github/workflows/master.yml
```diff
@@ -60,23 +62,34 @@ jobs:
         uses: ./.github/actions/pip_installation
       - name: Serial C tests
         run: |
-          python -m pytest -rx -m "not parallel and c" --ignore=printers --ignore=symbolic --ignore=ndarrays
+          python -m pytest -n auto -rx -m "not (parallel or xdist_incompatible) and c" --ignore=printers --ignore=symbolic --ignore=ndarrays
+          pyccel-clean
+          python -m pytest -rx -m "xdist_incompatible and not parallel and c" --ignore=printers --ignore=symbolic --ignore=ndarrays
+          pyccel-clean
```
yguclu, line 68: Do we need to explicitly call `pyccel-clean` if we don't use `pytest-xdist`? I've understood that we don't need to.

.github/workflows/master.yml
```diff
@@ -60,23 +62,34 @@ jobs:
...
       - name: ND-Array tests
         run: |
-          python -m pytest ndarrays/ -rx
+          python -m pytest -n auto ndarrays/ -rx
+          pyccel-clean
         working-directory: ./tests
       - name: Parallel tests
         run: |
           mpiexec -n 4 python -m pytest epyccel/test_parallel_epyccel.py -v -m parallel -rx
+          pyccel-clean
```
yguclu, line 92: Same question as above

yguclu (changes requested):
Very nice! This PR brings a great speedup to our testing suite!

I have a few comments/questions

tests/pyccel/conftest.py
```diff
@@ -0,0 +1,33 @@
...
+    for root, _, files in os.walk(path_dir):
+        for name in files:
+            if name.startswith(".coverage"):
+                shutil.copyfile(os.path.join(root,name),os.path.join(os.getcwd(),name))
+
+    config = session.config
+    xdist_plugin = config.pluginmanager.getplugin("xdist")
+    if xdist_plugin is None or "PYTEST_XDIST_WORKER_COUNT" not in os.environ \
+            or os.getenv('PYTEST_XDIST_WORKER_COUNT') == 1:
+        pyccel_clean(path_dir)
```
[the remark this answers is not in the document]
EmilyBourne: In the epyccel directory I clean up once at the start of the session and then after every individual test.
There is no particular reason to clean before starting only in one file. I can add this to pyccel or remove it from epyccel if you would like.
The reason for testing after each individual test in epyccel is so that I can examine the marks on the test and not clean up parallel tests.
In pyccel there are no parallel tests so it is faster to clean up everything at the end

pyccel/codegen/pipeline.py
EmilyBourne: The reason for avoiding the copy is not to reduce the number of files that are copied around but is to avoid one thread modifying a file (by overwriting it) while another thread is using it (compiling). I'm not sure if the copy itself is a problem, but previously we were deleting before copying and this was definitely creating a race condition

EmilyBourne: Fixed in c012c7fc

.github/workflows/master.yml
EmilyBourne: When running in parallel with mpi4py, we have the same problems as with pytest-xdist so we don't clean up after parallel tests

.github/workflows/master.yml
EmilyBourne: Fixed in fd8c81af

tests/pyccel/test_pyccel.py
EmilyBourne: See above

tests/pyccel/test_pyccel.py
EmilyBourne: Sorry I saw the same file name but missed that `test_relative_imports_in_project` and `test_absolute_imports_in_project` are operating in different folders

EmilyBourne: Fixed in 74065a32

tests/pyccel/test_pyccel.py
EmilyBourne: Fixed in 74065a32

pyccel/codegen/pipeline.py
EmilyBourne: Hmm, yes I was hoping to identify if the file had been compiled with the correct language, in case there were fortran and c versions of a file with the sam name, but you are correct this is not possible

pyccel/codegen/pipeline.py
EmilyBourne: Identifying that the files are at the correct location is not enough to know if they have been compiled as they may have been copied in `convert_only` mode here I try to identify if the code has already been compiled, and exit if this is the case. This is done to avoid problems between threads. E.g. if thread 1 tries to link using these files while thread 2 is trying to recompile them

pyccel/commands/pyccel_clean.py
```diff
@@ -0,0 +1,39 @@
...
+
+    files = os.listdir(path_dir)
+    for f in files:
+        file_name = os.path.join(path_dir,f)
+        if f in  ("__pyccel__", "__epyccel__"):
+            shutil.rmtree( file_name, ignore_errors=True)
+        elif not os.path.isfile(file_name) and recursive:
+            pyccel_clean(file_name)
+        elif f.endswith(ext_suffix):
+            os.remove(file_name)
```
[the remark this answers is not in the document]
EmilyBourne: Fixed in 3fee981

tests/pyccel/conftest.py
```diff
@@ -0,0 +1,33 @@
...
+from pyccel.commands.pyccel_clean import pyccel_clean
+
+@pytest.fixture( params=[
+        pytest.param("fortran", marks = pytest.mark.fortran),
+        pytest.param("c", marks = pytest.mark.c)
+    ],
+    scope='module'
+)
+def language(request):
+    return request.param
```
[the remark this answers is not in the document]
EmilyBourne: No. PR #783 only enabled epyccel tests. For pyccel tests we also need to correct all the compilation commands in this file. It is doable but it is a separate problem

pyccel/codegen/python_wrapper.py
EmilyBourne: I can't remember why I changed this, I have changed it back. See fee9773

pyccel/codegen/pipeline.py
yguclu: I see, thanks

.github/workflows/master.yml
yguclu: I think that there is a significant difference between running our "parallel-pyccel" tests and running the serial tests with `pytest-xdist`. In the first case only process #0 is actually compiling the files, while the other processes just load the shared library after it's been created. This means that we can easily add an MPI barrier to our tear-down functions and let process #0 take care of it. What do you think?

tests/pyccel/conftest.py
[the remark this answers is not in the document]
EmilyBourne: See 2bcd07b I am now using the same conftest for all tests

.github/workflows/master.yml
EmilyBourne: Fixed in ab6cfe9

pyccel/codegen/pipeline.py
yguclu: Thanks, I think I've understood now. Is this correct?
- if all object files for one internal library already exist, you move to the next library
- if one object file for that library is missing, you (re)compile all source files.

I cannot convince myself that this approach is completely water-tight, though. Here we have multiple processes that are compiling libraries, potentially at the same time, without talking to each other. So we have no barriers, and no synchronization whatsoever. I hope that the operating system is able to handle multiple read/write operations properly, but I suspect that we need to use a file-locking mechanism.

tests/conftest.py
```diff
@@ -30,6 +31,11 @@ def pytest_runtest_teardown(item, nextitem):
         marks = [m.name for m in item.own_markers ]
         if 'parallel' not in marks:
             pyccel_clean(path_dir, remove_shared_libs = True)
+        else:
+            comm = MPI.COMM_WORLD
+            comm.Barrier()
+            if comm.rank == root:
+                pyccel_clean(path_dir, remove_shared_libs = True)
```
yguclu, line 39: You may need to add another barrier after cleaning

pyccel/codegen/pipeline.py
EmilyBourne: Yes that's correct. I agree I have seen it occasionally cause problems but I'm not sure how to do better

yguclu: Should we try this out? https://pypi.org/project/filelock

tests/conftest.py
EmilyBourne: Fixed in ec0cc27

tests/pyccel/conftest.py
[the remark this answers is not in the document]
yguclu: Great! That's much cleaner

tests/pyccel/conftest.py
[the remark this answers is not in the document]
yguclu: I see, thanks

pyccel/codegen/pipeline.py
EmilyBourne: See 57be261

pyccel/codegen/pipeline.py
```diff
@@ -11,6 +11,7 @@
 import sys
 import shutil
 from collections import OrderedDict
+rom filelock import FileLock
```
yguclu, line 14: Typo!!!!

EmilyBourne: a9cb525

pyccel/codegen/pipeline.py
EmilyBourne: I have removed this

pyccel/codegen/pipeline.py
EmilyBourne: Done this is now working

yguclu: Nice!

.github/actions/pytest_run_python/action.yml
```diff
@@ -4,7 +4,9 @@ runs:
   using: "composite"
   steps:
     - name: Python tests with pytest
-      run:
-        python -m pytest -rx -m "not parallel and python" --ignore=printers --ignore=symbolic --ignore=ndarrays
+      run: |
+        python -m pytest -n auto -rx -m "not (parallel or xdist_incompatible) and python" --ignore=printers --ignore=symbolic --ignore=ndarrays
+        #python -m pytest -rx -m "xdist_incompatible and not parallel and python" --ignore=printers --ignore=symbolic --ignore=ndarrays
```
yguclu, line 9: What's the meaning of this commented line?

.github/workflows/Github_pytest.yml
```diff
@@ -56,22 +58,43 @@ jobs:
...
         run: |
-          python -m pytest -rx -m "not parallel and not (c or python)" --ignore=printers --ignore=symbolic --ignore=ndarrays
+          python -m pytest -n auto -rx -m "not (parallel or xdist_incompatible) and python" --ignore=symbolic --ignore=ndarrays
         working-directory: ./tests
-      - name: Serial Python tests
+      - name: Serial Python tests cleanup
         run: |
-          python -m pytest -rx -m "not parallel and python" --ignore=printers --ignore=symbolic --ignore=ndarrays
+          pyccel-clean
+          #python -m pytest -rx -m "xdist_incompatible and not parallel and python" --ignore=symbolic --ignore=ndarrays
```
yguclu, line 88: Should we remove this commented line?

.github/actions/pytest_run_python/action.yml
EmilyBourne: If we have any python tests which are incompatible with `pytest.xdist` then this line needs uncommenting. However it cannot be uncommented now as pytest raises error code `5` if no tests are found

.github/workflows/Github_pytest.yml
EmilyBourne: See above

.github/workflows/master.yml
```diff
@@ -58,22 +60,42 @@ jobs:
...
+          python -m pytest -rx -m "xdist_incompatible and not parallel and not (c or python)" --ignore=symbolic --ignore=ndarrays
+        working-directory: ./tests
+      - name: Serial Python tests run in parallel
+        run: |
+          python -m pytest -n auto -rx -m "not (parallel or xdist_incompatible) and python" --ignore=symbolic --ignore=ndarrays
+        working-directory: ./tests
+      - name: Serial Python tests cleanup
+        run: |
+          pyccel-clean
+          #python -m pytest -rx -m "xdist_incompatible and not parallel and python" --ignore=symbolic --ignore=ndarrays
```
yguclu, line 89: Should we remove this commented line?

EmilyBourne: See above

.github/workflows/Github_pytest.yml
yguclu: I see, but surely this line does not belong to step "Serial Python tests cleanup"?

.github/workflows/master.yml
yguclu: See above :wink:

.github/workflows/Github_pytest.yml
EmilyBourne: No, but I'm not sure that I can create an action which does nothing. I.e.
```
- name: Serial Python tests run in parallel
   run : |
       #python -m pytest -rx -m "xdist_incompatible and not parallel and python" --ignore=symbolic --ignore=ndarrays
   working-director : tests
```

pyccel/commands/pyccel_clean.py
```diff
@@ -0,0 +1,39 @@
...
+    Parameters
+    ----------
+    path_dir  : str
+                The path to the folder which should be cleaned
+                Default : current working directory
+    recursive : bool
+                Indicates whether the function should recurse
+                into sub-folders
+                Default : True
+    """
```
yguclu, line 27: Please update the docstring with the parameter `remove_shared_libs`.

pyccel/commands/pyccel_clean.py
```diff
@@ -0,0 +1,39 @@
...
+#------------------------------------------------------------------------------------------#
+""" Module containing scripts to remove pyccel generated objects
+"""
+import os
+import shutil
+import sysconfig
+
+ext_suffix = sysconfig.get_config_var('EXT_SUFFIX')
+
+def pyccel_clean(path_dir = None, recursive = True, remove_shared_libs = False):
```
yguclu, line 14: If used from the command line, I think that we need to provide a proper parser to this small function, or we might even want to use it as a subcommand of `pyccel`. What do you think? Maybe in another PR?

.github/workflows/Github_pytest.yml
yguclu: I think it's OK (and clearer) if you comment out multiple lines.

.github/actions/pip_installation/action.yml
```diff
@@ -31,4 +31,8 @@ runs:
       shell: bash
       run:
         python -m pip install "pytest>=4.5"
+    - name: Install parallel pytest
+      shell: bash
+      run:
+        python -m pip install "pytest-xdist"
```
yguclu, line 37: Can you add a line to our README file in [this section](https://github.com/pyccel/pyccel#additional-packages)?

tests/epyccel/test_arrays.py
```diff
@@ -13,7 +13,7 @@
 def test_array_int32_1d_scalar_add(language):
 
     f1 = arrays.array_int32_1d_scalar_add
-    f2 = epyccel( f1 , language = language)
+    f2 = epyccel( f1 , language = language, verbose=True)
```
yguclu, line 16: Is the `verbose` flag useful in your opinion, or is it just a remnant of your debugging efforts?

.github/actions/pytest_run/action.yml
```diff
@@ -5,10 +5,13 @@ runs:
...
-        mpiexec -n 4 ${MPI_OPTS} python -m pytest epyccel/test_parallel_epyccel.py -v -m parallel -rx
-        #mpiexec -n 4 ${MPI_OPTS} python -m pytest epyccel -v -m parallel -rx
+        python -m pytest -n auto -rx -m "not (parallel or xdist_incompatible) and c" --ignore=printers --ignore=symbolic --ignore=ndarrays -x
+        python -m pytest -rx -m "xdist_incompatible and not parallel and c" --ignore=printers --ignore=symbolic --ignore=ndarrays
+        pyccel-clean
+        python -m pytest -n auto -rx -m "not (parallel or xdist_incompatible) and not (c or python)" --ignore=printers --ignore=symbolic --ignore=ndarrays
+        python -m pytest -rx -m "xdist_incompatible and not parallel and not (c or python)" --ignore=printers --ignore=symbolic --ignore=ndarrays
+        pyccel-clean
+        python -m pytest -n auto ndarrays/ -rx
+        pyccel-clean
```
yguclu, line 15: The `printers` directory was removed, correct?

The flag `--ignore=printers` should be removed from all of our commands.

yguclu (changes requested):

tests/epyccel/test_arrays.py
EmilyBourne: Your merge caused a RuntimeError. I put the verbose in to try to find it but it doesn't seem to be reproducible :(

.github/workflows/Github_pytest.yml
EmilyBourne: Fixed in 54cb3c8

.github/workflows/master.yml
EmilyBourne: Fixed in 54cb3c8

pyccel/commands/pyccel_clean.py
EmilyBourne: Fixed 832526a

pyccel/commands/pyccel_clean.py
EmilyBourne: Would this prevent me from calling the function from python?

.github/actions/pytest_run/action.yml
EmilyBourne: Fixed in cff9757

tests/epyccel/test_arrays.py
EmilyBourne: Removed in afaafad

.github/actions/pip_installation/action.yml
EmilyBourne: Fixed 54de294

pyccel/commands/pyccel_clean.py
EmilyBourne: I am reusing this file for the teardown. I'm not sure what the effect of calling `parser.parse_args()` repeatedly inside pytest might be

EmilyBourne:
It seems there is still a collision, see https://github.com/pyccel/pyccel/runs/2140653596. But I am not sure what is causing this

pyccel/commands/pyccel_clean.py
yguclu: I am suggesting to have a separate function that has a parser, and calls this one.

EmilyBourne: Fixed 0e14e63

README.rst
```diff
@@ -218,6 +218,10 @@ In order to run the unit tests and to get a coverage report, four additional Pyt
   pip3 install --user pytest
   pip3 install --user coverage
 
+Most of the unit tests can also be run in parallel. This can be done by installing one additional package::
+
+  pip3 install --user pytest-xdist
+
```
yguclu, line 224: Could you please fix line 213 of this file in this PR?

It says "four additional Python packages should be installed", but they are not four.

EmilyBourne: Fixed in 7c944fb

yguclu: Thanks!

pyccel/codegen/pipeline.py
```diff
@@ -325,41 +326,45 @@ def handle_error(stage):
                     # remove library folder to avoid missing files and copy
                     # new one from pyccel stdlib
                     lib_dest_path = os.path.join(pyccel_dirpath, lib_name)
-                    if os.path.exists(lib_dest_path):
-                        shutil.rmtree(lib_dest_path)
-                    shutil.copytree(lib_path, lib_dest_path)
+                    lock_path = lib_dest_path + '.lock'
+                    lock = FileLock(lock_path)
+                    with lock:
```
yguclu: If the variables `lock` and `lock_path` are not used anywhere else in this file, I would recommend that we condense these three lines into one, as you do on line 353:
```python
with FileLock(lib_dest_path + '.lock'):
```

EmilyBourne: Beat you to it. See 96ae409 ;)

pyccel/codegen/python_wrapper.py
```diff
@@ -129,8 +130,16 @@ def create_shared_library(codegen,
 
         if verbose:
             print(' '.join(cmd))
-        p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True)
-        out, err = p.communicate()
+        lock_names = [d+'.lock' for d in dep_mods]
+        locks = [FileLock(n) for n in lock_names]
```
yguclu, line 134: Could you combine these two lines as well?

pyccel/codegen/pipeline.py
yguclu: :laughing:

pyccel/codegen/python_wrapper.py
EmilyBourne: Fixed in 2a4be0b

yguclu (approved):

PR #817, opened 2021-03-24
Implement PythonLen in C

### LINKED ISSUE #785: PythonLen is not implemented in C
**Describe the bug**
The `len` function is very useful and is used in several tests but it is not implemented in C. This is therefore blocking some tests

**Test Code**
```python
def f(a : 'int[:,:]'):
    return len(a)
```

**Language**
C

EmilyBourne:
Implement PythonLen in C. Fixes #785

### DIFF
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -445,6 +445,13 @@ class CCodePrinter(CodePrinter):
         self._additional_imports.add("complex")
         return '_Complex_I'
 
+    def _print_PythonLen(self, expr):
+        var = expr.arg
+        if var.rank > 0:
+            return self._print(PyccelArraySize(var, 0))
+        else:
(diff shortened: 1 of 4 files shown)


rakati (approved):

pyccel/codegen/printing/ccode.py
```diff
@@ -445,6 +445,13 @@ def _print_LiteralImaginaryUnit(self, expr):
         self._additional_imports.add("complex")
         return '_Complex_I'
 
+    def _print_PythonLen(self, expr):
+        var = expr.arg
+        if var.rank > 0:
+            return self._print(var.shape[0])
+        else:
+            raise NotImplementedError("PythonLen not implemented for type {}".format(type(expr.arg)))
```
yguclu, line 453: Could it be useful to also print the expression `expr` in the error message?

EmilyBourne: Fixed in 1f005cf

yguclu (approved):

Issue #802, opened 2021-03-11
Python interfaces


EmilyBourne:
**Describe the bug**
Interfaces in python assume that the code is the same in all languages and therefore print the body of the first `FunctionDef` with all the decorators. However the code generated by pyccel is not always the same (e.g. if the input can be either a boolean or a numeric type)

**To Reproduce**
Provide code to reproduce the behavior:
```python
@types('float')
@types('complex')
def get_imag(a):
    import numpy as np
    return np.imag(a)
```

Provide the generated code, or the error message:
```python
from pyccel.decorators import types

@types('float')
@types('complex')
def get_imag(a):


    from numpy import imag
    return 0.0
```

**Expected behavior**
There are multiple possible solutions:
- We could raise an error, complaining about incompatible types
- We could ensure all code is preserved so it is identical at printing (I don't like this solution as the value of `np.imag(a)` for all types other than `complex` would have to be corrected manually for every language
- We could print all function def bodies, and collect versions giving the same code:
    ```
    from pyccel.decorators import types

    @types('float')
    @types('complex')
    def get_imag(a):
        if isinstance(a, float):
            from numpy import imag
            return 0.0
        else:
            from numpy import imag
            return (a).imag
    ```
    but this code is not understood by pyccel
- We could generate multiple translatable functions and a wrapper:
    ```
    from pyccel.decorators import types

    @types('float')
    def get_imag_00(a):
        from numpy import imag
        return 0.0

    @types('complex')
    def get_imag_01(a):
        from numpy import imag
        return (a).imag
        
    def get_imag(a):
        if isinstance(a, float):
             return get_imag_00(a)
        elif isinstance(a, complex):
             return get_imag_01(a)
        else:
             raise TypeError("Unexpected type")
    ```

**Language**
python

**Additional context**
This is causing some of our tests to be incomplete, but the tests could also be fixed by avoiding interfaces. In an ideal world, the dtype should be a pytest parameter

ratnania:
What if we have a python decorator  that will concatenate the function name with the associated type, in order to avoid the last defined function get_imag?

EmilyBourne:
>What if we have a python decorator that will concatenate the function name with the associated type, in order to avoid the last defined function get_imag?

I'm not sure that we can avoid the last function no matter the name of the functions as in pyccel we always aim to have a result that can be called in the same way as the original code

EmilyBourne:
Duplicate of #885

PR #819, opened 2021-03-24
Use numpy cast functions to cast arrays

### LINKED ISSUE #809: Numpy.complex64 and Numpy.complex128 doesn’t work with arrays
**Describe the bug**
Pyccel can't generate Fortran/Python code when using Numpy.complex64 and Numpy.complex128 with arrays.

**To Reproduce**
```python
@types('int[:,:]')
def get_complex64_arr_2d(arr):
    from numpy import complex64, shape
    a = complex64(arr)
    s = shape(a)
    return len(s), s[0], s[1], a[0,0], a[0,1]
```

The error message:
`AttributeError: _dtype`

**Expected behavior**

**Language**
Python/Fortran

**Additional context**
tests are already in #719.

EmilyBourne:
Python's builtin functions cannot be called with arrays as arguments. Internally pyccel's AST allows this in order to facilitate casting. However this sometimes leads to confusion. In python numpy also has casting functions which can be called on arrays and have the expected behaviour. This PR therefore restores the classes `PythonInt`, `PythonFloat`, `PythonComplex` and `PythonBool` to classes which only handle scalars, ensures that all Numpy casting functions handle arrays, and calls the numpy casting functions instead of the pure python casting functions. This fixes #809 

**Commit Summary**
- [BUGFIX] Correctly set dtype etc for empty tuples
- Ensure that all numpy casting functions accept arrays as arguments (fixes #809)
- Use numpy casting instead of python casting when casting internally

### DIFF
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -190,6 +190,12 @@ class NumpyReal(PythonReal):
     1.0
     """
     __slots__ = ('_rank','_shape','_order')
+    def __new__(cls, arg):
+        if isinstance(arg.dtype, NativeBool):
+            return NumpyInt(arg)
+        else:
+            return super().__new__(cls, arg)
+
     def __init__(self, arg):
         super().__init__(arg)
         self._precision = arg.precision
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -89,31 +89,6 @@ __all__ = (
 )
 
 #=======================================================================================
-class NumpyComplex(PythonComplex):
-    """ Represents a call to numpy.complex() function.
-    """
-    __slots__ = ('_rank','_shape','_order')
-    def __init__(self, arg0, arg1 = None):
-        if arg1 is not None:
-            raise NotImplementedError("Use builtin complex function not deprecated np.complex")
-        self._shape = arg0.shape
-        self._rank  = arg0.rank
-        self._order = arg0.order
-        super().__init__(arg0)
-
-class NumpyComplex64(NumpyComplex):
-    """ Represents a call to numpy.complex64() function.
-    """
-    __slots__ = ()
-    _precision = dtype_registry['complex64'][1]
-
-class NumpyComplex128(NumpyComplex):
-    """ Represents a call to numpy.complex128() function.
-    """
-    __slots__ = ()
-    _precision = dtype_registry['complex128'][1]
-
-#=======================================================================================
 class NumpyFloat(PythonFloat):
     """ Represents a call to numpy.float() function.
     """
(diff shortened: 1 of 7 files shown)


mostafaMamoni (approved):
i didn't notice that in my journey with numpy functions, we definitely shouldn’t change the behaviour of Python builtins even in Pyccel's internals.

pyccel/ast/numpyext.py
```diff
@@ -171,6 +190,12 @@ class NumpyReal(PythonReal):
     1.0
     """
     __slots__ = ('_rank','_shape','_order')
+    def __new__(cls, arg):
+        if isinstance(arg.dtype, NativeBool):
+            return NumpyInt(arg)
```
yguclu, line 195: Shouldn't we also handle the case where `isinstance(arg.dtype, NativeInt)`?

EmilyBourne: This is handled in `PythonReal.__new__` which contains:
```
def __new__(cls, arg):
            return PythonInt(arg)
        elif not isinstance(arg.dtype, NativeComplex):
            return arg
        else:
            return super().__new__(cls)
```
Here it is necessary to handle booleans as in numpy function calls the arguments can be arrays. For other types there is no casting involved, we just return the argument so the contents of `PythonReal.__new__` are sufficient. As a result I didn't think it was necessary to duplicate the condition.

yguclu: I see, thanks.

yguclu (changes requested):
Could you add a couple of tests checking that #809 is fixed?

EmilyBourne:
> Could you add a couple of tests checking that #809 is fixed?

I knew I'd forgotten something ! I have activated (and fixed) the tests which were created in PR #719 . I ran into some unexpected python problems which required reordering classes in ast/numpyext.py. It is now working, but issue #802 is preventing me from adding the tests to our unit tests

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -54,11 +54,20 @@
...
+        else:
+            return isinstance(pyccel_result, type(python_result)) or \
+                    isinstance(pyccel_result.item(), type(python_result))
     else:
-        return isinstance(pyccel_result, type(python_result))
+        if isinstance(python_result, np.generic):
+            #TODO: Remove when #735 is fixed
+            return isinstance(pyccel_result, (type(python_result.item()), type(python_result)))
+        else:
+            return isinstance(pyccel_result, type(python_result))
```
yguclu, line 71: It seems to me that when `isinstance(pyccel_result, type(python_result)) == True` this function always returns `True`. If that is correct, we can greatly simplify this code with an early exit condition:

```python
if type(pyccel_result) is type(python_result):
    return True
```

And then treat the other corner cases.

EmilyBourne: You are right. Done in ac9de23

yguclu (approved):
Great job! Thank you!

PR #823, opened 2021-03-25
Speed up visit of PyccelAdd and PyccelMul nodes in semantic parser

### LINKED ISSUE #815: Semantic parser is extremely slow at processing long expressions
Bug description
--------------------
Pyccel is extremely slow when it processes long expressions.

To Reproduce
--------------------
The following function is extremely slow to parse, and it effectively hangs on my computer:

```python
def determinant_4x4(M: 'float[:,:](order=C)') -> 'float':
    return (
          M[0,0] * M[1,1] * M[2,2] * M[3,3]
        + M[0,0] * M[1,2] * M[2,3] * M[3,1]
        + M[0,0] * M[1,3] * M[2,1] * M[3,2]
        + M[0,1] * M[1,0] * M[2,3] * M[3,2]
        + M[0,1] * M[1,2] * M[2,0] * M[3,3]
        + M[0,1] * M[1,3] * M[2,2] * M[3,0]
        + M[0,2] * M[1,0] * M[2,1] * M[3,3]
        + M[0,2] * M[1,1] * M[2,3] * M[3,0]
        + M[0,2] * M[1,3] * M[2,0] * M[3,1]
        + M[0,3] * M[1,0] * M[2,2] * M[3,1]
        + M[0,3] * M[1,1] * M[2,0] * M[3,2]
        + M[0,3] * M[1,2] * M[2,1] * M[3,0]
        - M[0,0] * M[1,1] * M[2,3] * M[3,2]
        - M[0,0] * M[1,2] * M[2,1] * M[3,3]
        - M[0,0] * M[1,3] * M[2,2] * M[3,1]
        - M[0,1] * M[1,0] * M[2,2] * M[3,3]
        - M[0,1] * M[1,2] * M[2,3] * M[3,0]
        - M[0,1] * M[1,3] * M[2,0] * M[3,2]
        - M[0,2] * M[1,0] * M[2,3] * M[3,1]
        - M[0,2] * M[1,1] * M[2,0] * M[3,3]
        - M[0,2] * M[1,3] * M[2,1] * M[3,0]
        - M[0,3] * M[1,0] * M[2,1] * M[3,2]
        - M[0,3] * M[1,1] * M[2,2] * M[3,0]
        - M[0,3] * M[1,2] * M[2,0] * M[3,1]
    )
```

This can be shown by calling pyccel with the `-e` flag, which stops execution after the semantic stage:

```bash
$ pyccel -e test.py
```

By commenting out most of the expression, and progressively increasing the number of lines it contains, one can see that the problem arises when the number of lines approaches 10.

Expected behavior
-----------------------
Pyccel should be able to process long expressions without any problems.

Language
-------------
This is a problem in the semantic parser. As such, it is independent of the language.

EmilyBourne:
Avoid creating temporary arguments which need to be invalidated when visiting `PyccelAdd` or `PyccelMul`. Fixes #815

**Commit Summary**

- Create _create_PyccelOperator function to avoid code duplication and unnecessary node invalidation
- Rename _visit_Dlist -> _create_Dlist
- Group utility functions in pyccel/parser/semantic.py and add some small docstrings
- Increase patch number: Pyccel's version is now 1.2.2


yguclu (changes requested):
Could you separate the new `_create` utility functions from the `_visit` methods? 

Actually it would be really useful to collect all utility functions in a separate section of the file:

- `_assign_lhs_variable`
- `_create_variable`
- `_create_PyccelOperator`
- `_create_Dlist`
- `_ensure_target`
- `_extract_indexed_from_var`
- `_handle_function_args`
- `_handle_function`

Some of these functions do not even need to be methods of the class `SemanticParser` actually. But this is a different discussion to have...

yguclu (changes requested):
Please update the patch number in `pyccel.__version__`.
We are going to create a new release after this bugfix.

yguclu (approved):

PR #828, opened 2021-03-30
Raise error when function call has wrong number of arguments

EmilyBourne:
Pyccel is not guaranteed to work when the underlying python code will not run, however the error raised when too few arguments are passed to the function is particularly unclear:
```
pyccel:
 |fatal [codegen]: tmp.py| _print_NoneType is not yet implemented for language : Fortran
Pyccel has encountered syntax that has not been implemented yet. Please create an issue at https://github.com/pyccel/pyccel/issues and provide a small example of your problem. Do not forget to specify your target language
```
The `NoneType` appears as pyccel uses `None` internally as a placeholder for arguments which have no default value. It is trivial to test that the number of arguments is correct and to raise a more human readable error with an identifiable line of code

**Commit Summary**
- Stop obfuscation of fst when one pyccel object is represented by multiple python.ast nodes (e.g. `pyccel.ast.Assign` is represented by `ast.assign` and `ast.Expr`)
- Raise an error if too many arguments are passed to a function
- Raise an error if too few arguments are passed to a function

### DIFF
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -902,7 +902,7 @@ class SemanticParser(BasicParser):
                 args.append(a)
         return args
 
-    def _handle_function(self, func, args, **settings):
+    def _handle_function(self, expr, func, args, **settings):
         if not isinstance(func, (FunctionDef, Interface)):
             args, kwargs = split_positional_keyword_arguments(*args)
             for a in args:
@@ -915,8 +915,16 @@ class SemanticParser(BasicParser):
 
             return expr
         else:
-            expr = FunctionCall(func, args, self._current_function)
-            return expr
+            if len(args) > len(func.arguments):
+                errors.report("Too many arguments passed in function call",
+                        symbol = expr,
+                        severity='fatal')
+            new_expr = FunctionCall(func, args, self._current_function)
+            if None in new_expr.args:
+                errors.report("Too few arguments passed in function call",
+                        symbol = expr,
+                        severity='fatal')
+            return new_expr
 
     def _create_variable(self, name, dtype, rhs, d_lhs):
         """
@@ -1471,7 +1479,7 @@ class SemanticParser(BasicParser):
                     if new_name != rhs_name:
                         if hasattr(func, 'clone'):
                             func  = func.clone(new_name)
-                    return self._handle_function(func, args, **settings)
+                    return self._handle_function(expr, func, args, **settings)
                 elif isinstance(rhs, Constant):
                     var = first[rhs_name]
                     if new_name != rhs_name:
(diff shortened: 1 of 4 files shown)


pyccel/parser/semantic.py
```diff
@@ -902,7 +902,7 @@ def _handle_function_args(self, arguments, **settings):
                 args.append(a)
         return args
 
-    def _handle_function(self, func, args, **settings):
+    def _handle_function(self, expr, func, args, **settings):
         if not isinstance(func, (FunctionDef, Interface)):
             args, kwargs = split_positional_keyword_arguments(*args)
             for a in args:
```
yguclu, line 908: Could you please add a docstring to this function? It has become really obscure.

pyccel/parser/semantic.py
```diff
@@ -915,8 +915,16 @@ def _handle_function(self, func, args, **settings):
 
             return expr
```
yguclu, line 916: Shouldn't we call this variable `new_expr` as done below?

yguclu (changes requested):

pyccel/parser/semantic.py
EmilyBourne: Fixed in 008e068

pyccel/parser/semantic.py
EmilyBourne: Fixed in 008e068

pyccel/parser/semantic.py
```diff
@@ -903,6 +903,19 @@ def _handle_function_args(self, arguments, **settings):
         return args
 
     def _handle_function(self, expr, func, args, **settings):
+        """
+        Create a FunctionCall or an instance of a PyccelInternalFunction
+        from the function informations and arguments
```
yguclu, line 908: Isn't "information" an uncountable noun in English?

pyccel/parser/semantic.py
```diff
@@ -902,7 +902,20 @@ def _handle_function_args(self, arguments, **settings):
...
+
+        Parameters
+        ==========
+        expr : PyccelAstNode
+               The expression where this call is found (used for error output)
+        func : FunctionDef, Interface or PyccelInternalFunction type
+               The function being called
+        args : tuple
+               The arguments passed to the function
+        """
```
yguclu, line 918: Thanks! But I think the "Returns" section is missing...

pyccel/parser/semantic.py
EmilyBourne: Fixed in 47de259

pyccel/parser/semantic.py
EmilyBourne: Fixed in 47de259

yguclu (approved):

PR #831, opened 2021-03-30
Remove unnecessary NativeOp class

EmilyBourne:
The code contains the following message:
> The following were defined to be sympy approved nodes. If there is something
> smaller that could be used, that would be preferable. We only use them as
> tokens.

Now that sympy has been removed this is no longer necessary and a simple string can be used.

### DIFF
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -36,7 +36,6 @@ errors = Errors()
 
 # TODO [YG, 12.03.2020]: Move non-Python constructs to other modules
 # TODO [YG, 12.03.2020]: Rename classes to avoid name clashes in pyccel/ast
-# NOTE: commented-out symbols are never used in Pyccel
 __all__ = (
     'AliasAssign',
     'Allocate',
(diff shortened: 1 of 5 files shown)


pyccel/ast/core.py
```diff
@@ -38,7 +38,6 @@
 # TODO [YG, 12.03.2020]: Rename classes to avoid name clashes in pyccel/ast
 # NOTE: commented-out symbols are never used in Pyccel
```
yguclu, line 39: I think that this "NOTE" is no longer relevant. Please remove it

pyccel/ast/core.py
```diff
@@ -859,23 +803,34 @@ def __init__(
...
             raise TypeError('Unrecognized Operator')
 
         self._op = op
 
         super().__init__(lhs, rhs, status, like)
 
     def __str__(self):
-        return '{0} {1}= {2}'.format(str(self.lhs), self.op._symbol,
+        return '{0} {1}= {2}'.format(str(self.lhs), self.op,
                 str(self.rhs))
```
yguclu, line 815: Could this fit in a single line now?

yguclu (changes requested):
This looks very good to me. I have just a couple of minor comments

pyccel/ast/core.py
EmilyBourne: Fixed in 908cae3

pyccel/ast/core.py
EmilyBourne: Fixed in 908cae3

yguclu (approved):

Issue #736, opened 2021-02-10
`_visit_For` in parser/semantic.py mixes syntactic and semantic stages dangerously


EmilyBourne:
## Problem
The function `_visit_For` in parser/semantic.py mixes syntactic and semantic stages. This is dangerous as the semantic stage can fail if it visits an object which has already been visited.
This problem is due to the structure of a `For` block and the translation of the iterable object

## `For` Class : Terminology

A `For` class contains the following objects:
- target
- iterable
- body

The roles of these different objects can be seen in the following code snippet
```
for target in iterable:
    body
```

## `_visit_For` current implementation

Currently `_visit_For` does the following (the parts of the class are indicated in _italics_ for syntactic and **bold** for semantic):
1. Create a new loop scope (_target_, _iterable_ and _body_ begin in syntactic form)
2. Visit _iterable_ to create the semantic **iterable**
3. Change the current stage `PyccelAstNode.stage = 'syntactic'` (this is necessary as we now create syntactic objects, so the type checks in the AST must be turned off)
4. Only `PythonRange` **iterable**s are acceptable in fortran, so for all other **iterable**s _additional nodes_ are added to the body.
    For example:
    i. `for v in variable` the **iterable** is a `Variable` so the _additional node_ is :
        Assign(_target_, IndexedElement(**iterable**, _indx_)) # Problem as we are mixing syntactic and semantic
    ii. `for v in zip(a, b)` the **iterable** is a `PythonZip` so the _additional nodes_ are :
        Assign(_target_[0], IndexedElement(**a**, _indx_)) # Problem as we are mixing syntactic and semantic
        Assign(_target_[1], IndexedElement(**b**, _indx_)) # Problem as we are mixing syntactic and semantic
5. Visit the _target_ and create one or more `Variable`s. This does not use the standard method `_create_variable` as there is no one `rhs`. It is always a scalar integer
6. Change the current stage back to `PyccelAstNode.stage = 'semantic'`
7. Visit _body_ (which contains the _additional nodes_ created in step 4.
8. Create and return the `For` object

**Problems**
- Step 4 creates syntactic objects which may have semantic arguments
- Steps 5 and 6 are in the wrong order
- Changing all iterables to `PythonRange` stops us printing the simplified form in languages which support them (e.g. python, C++)

## Proposed Solution
Create a class `Iterable` containing:
- iterable : The object over which we iterate (Variable, Map, Zip, Range, ...)
[...]

PR #832, opened 2021-03-30
Add FunctionalFor, Enumerate, Zip, and Map to C printer

EmilyBourne:
Add FunctionalFor, Enumerate, Zip, and Map to C printer and activate tests. FunctionalFor is also added to the python printer but the iterators cannot be added easily due to issue #736

### DIFF
--- a/tests/epyccel/test_compare_expressions.py
+++ b/tests/epyccel/test_compare_expressions.py
@@ -6,9 +6,8 @@ from pyccel.decorators import types
 #==============================================================================
 class epyccel_test:
     """
-    Class to pyccelize module then compare different results
-    This avoids the need to pyccelize the file multiple times
-    or write the arguments multiple times
+    Class to pyccelize module or function then compare different results
+    This avoids the need to pyccelize the object multiple times
     """
     def __init__(self, f, lang='fortran'):
         self._f  = f
--- a/tests/epyccel/test_compare_expressions.py
+++ b/tests/epyccel/test_compare_expressions.py
@@ -1,22 +1,6 @@
 # pylint: disable=missing-function-docstring, missing-module-docstring/
-import numpy as np
-from pyccel.epyccel import epyccel
 from pyccel.decorators import types
-
-#==============================================================================
-class epyccel_test:
-    """
-    Class to pyccelize module or function then compare different results
-    This avoids the need to pyccelize the object multiple times
-    """
-    def __init__(self, f, lang='fortran'):
-        self._f  = f
-        self._f2 = epyccel(f, language=lang)
-
-    def compare_epyccel(self, *args):
-        out1 = self._f(*args)
-        out2 = self._f2(*args)
-        assert np.equal(out1, out2 )
+from .utilities import epyccel_test
 
 #==============================================================================
 @types('int, int, int')
--- a/tests/epyccel/test_compare_expressions.py
+++ b/tests/epyccel/test_compare_expressions.py
@@ -1,6 +1,6 @@
 # pylint: disable=missing-function-docstring, missing-module-docstring/
 from pyccel.decorators import types
-from .utilities import epyccel_test
+from utilities import epyccel_test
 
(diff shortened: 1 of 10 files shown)


tests/epyccel/test_functionals.py
```diff
@@ -1,43 +1,99 @@
...
-    out1 = f(*args)
-    out2 = f2(*args)
-    assert equal(out1, out2).all()
+#==============================================================================
+class epyccel_test:
+    """
+    Class to pyccelize module then compare different results
+    This avoids the need to pyccelize the file multiple times
+    or write the arguments multiple times
+    """
```
yguclu, line 15: Nice idea!

This should save time when the same function is called multiple times with different arguments. But I don't see that happening in this file, maybe you wanted to add more tests?

By the way, I think that in the docstring you meant to write "Class to pyccelize function", not "module". Is that right?

tests/epyccel/test_functionals.py
```diff
@@ -1,43 +1,99 @@
...
 
-def test_functional_for_2d_dependant_range():
-    compare_epyccel(functionals.functional_for_2d_dependant_range_1)
-    compare_epyccel(functionals.functional_for_2d_dependant_range_2)
-    compare_epyccel(functionals.functional_for_2d_dependant_range_3)
+def test_functional_for_2d_dependant_range(language):
+    test = epyccel_test(functionals.functional_for_2d_dependant_range_1, lang=language)
+    test.compare_epyccel()
+    test.compare_epyccel()
+    test.compare_epyccel()
```
yguclu, line 82: Why are we not testing the functions `functional_for_2d_dependant_range_2` and `functional_for_2d_dependant_range_3` anymore?

And what is the point of comparing the results multiple times if we do not change the input arguments?

yguclu (changes requested):

tests/epyccel/test_functionals.py
EmilyBourne: Ah sorry, I misread this and thought that it was the same test. That's why I added `epyccel_test` to this file. I have removed it. See d5f741d

tests/epyccel/test_compare_expressions.py
```diff
@@ -6,9 +6,8 @@
 #==============================================================================
 class epyccel_test:
     """
-    Class to pyccelize module then compare different results
-    This avoids the need to pyccelize the file multiple times
-    or write the arguments multiple times
+    Class to pyccelize module or function then compare different results
+    This avoids the need to pyccelize the object multiple times
     """
     def __init__(self, f, lang='fortran'):
```
yguclu, line 12: I see that multiple files contain this class. Maybe we can move it to a separate "helper" file?

And by the way, this class cannot work for modules at the moment, only functions. If you want to use it for pyccelizing modules (and then compare functions therein) you will need to make it more complex.

yguclu (changes requested):

tests/epyccel/test_compare_expressions.py
EmilyBourne: I think that would be helpful but I'm not sure where it makes sense to put that file. Should it be in the tests folder? Or in the pyccel folder?

> And by the way, this class cannot work for modules at the moment, only functions.

Ah you are right sorry. I don't think it should be used for modules as this may cause confusion about what is causing compiler failures and means that 1 failure in a module would make all tests reliant on that module fail

yguclu: > I'm not sure where it makes sense to put that file

I would put this class in the module `tests/epyccel/utilities.py`.



> I don't think it should be used for modules as this may cause confusion about what is causing compiler failures

I agree that it's better to only use this class for comparing functions. So please update its docstring accordingly.

EmilyBourne: Done in 02e374d

yguclu (approved):

yguclu (approved):

PR #836, opened 2021-04-03
Set order to None for non array variable

rakati:


nhamidn:
is there any test ?

rakati:
> is there any test ?

I will add a simple test

nhamidn (approved):

mostafaMamoni (approved):

EmilyBourne (approved):

EmilyBourne:
Any idea why this led to inhomogeneous tuples rather than an AttributeError when you collect the order?

pyccel/ast/variable.py
```diff
@@ -142,6 +142,7 @@ def __init__(
 
         if rank == 0:
             shape = ()
+            order=None
```
yguclu, line 145: Shouldn't we change the arguments' default values instead, that is:
```python
def __init__(
    ...
    rank  = 0,
    shape = (),    # was None
    order = None,  # was 'C'
```
If the default values are not compatible with each other, then maybe we should not provide them at all...

Please check if my suggestion is possible (it may require some further changes in the class).

yguclu (changes requested):

pyccel/ast/variable.py
rakati: ok, I will see about that, I think I will make this PR as drafted, I am not confident in this fix.

EmilyBourne: I don't agree that we should change the default arguments. While it seems more coherent to have the default arguments match (rank = 0, so shape =(), and order = None) I think the current behaviour makes the default more obvious. Let me summarise the current situation:
- **rank** : Should _always_ be provided but defaults to 0 as it is common to insert a scalar in the code (e.g. an index which is required in fortran but not in python)
- **shape** : Should be provided when known. If unknown it defaults to `shape = None`. However in the special case of `rank = 0` we can deduce that `shape = ()` (this therefore overrides any provided shape
- **order** : Only required for objects with `rank > 1`. When it is needed it defaults to `'C'`. When it is not needed `order = None`

With the current implementation, if `rank = 0` then the `shape` and `order` properties are ignored so the arguments are not technically incompatible.

If you want the default values to be compatible with one another then there are a few choices:
1. `rank = 0, shape = (), order = None`
    - Default is a scalar
    - When creating an argument we will need to remember that the default shape for an unknown shape is `None`
    - We need to check that the provided shape makes sense with the provided rank as the default shape may be incompatible
    - We need to check that the order is not `None` for an array with 2 or more dimensions and/or we need to default to `'C'` in the body of the `__init__` method
2. `rank, shape = None, order = 'C'`
    - Always provide the rank
    - This requires adding code everywhere we create a Variable on the fly
    - The order may still be "incompatible" with the shape

EmilyBourne, line 145: The order should also be set to `None` for rank 1 arrays

EmilyBourne (changes requested):

pyccel/ast/variable.py
yguclu: > - **rank** : Should *always* be provided but defaults to 0 as it is common to insert a scalar in the code (e.g. an index which is required in fortran but not in python)

I see a contradiction in this reasoning. If `rank` should always be given, then we should not provide a default value for it.

> With the current implementation, if `rank = 0` then the `shape` and `order` properties are ignored so the arguments are not technically incompatible.

What you are saying is not totally correct. `order` needs to be set to `None` for scalar quantities, and this is what this PR is about.

yguclu: > 1. `rank = 0, shape = (), order = None`
    - Default is a scalar
    - When creating an argument we will need to remember that the default shape for an unknown shape is `None`
    - We need to check that the provided shape makes sense with the provided rank as the default shape may be incompatible
    - We need to check that the order is not `None` for an array with 2 or more dimensions and/or we need to default to `'C'` in the body of the `__init__` method

Yes, this is what I was suggesting. Thanks for describing the details!

For what is worth, we can also set `shape = None` as this attribute carries absolute no meaning when `rank = 0`.

EmilyBourne: > I see a contradiction in this reasoning. If `rank` should always be given, then we should not provide a default value for it.

I agree, this is the second solution

> What you are saying is not totally correct. `order` needs to be set to `None` for scalar quantities, and this is what this PR is about.

Sorry, I meant the argument is ignored, not the property

yguclu: @rakati We just had a discussion on this topic during the daily scrum meeting, and we agreed on a solution that should require relatively few modifications to the library:

1. Use the defaults `rank = 0`, `shape = None`, `order = None` in order to keep scalar variables' creation as simple as possible;
2. Add some consistency checks when the parameters are passed to the constructor:

   - make sure that `isinstance(rank, int)` and `rank >= 0`
   - if `rank <= 1` check that both `order` and `shape` are `None`
   - if `rank >= 2` check that `order in ('C', 'F')`
   - if `shape is not None` check that `len(shape) == rank`

   Since these checks are internal to Pyccel, it is appropriate to use `assert` statements without worrying about raising a neat exception.

EmilyBourne: Correction:
> if `rank <= 1` check that both `order` and `shape` are `None`

Should be:
if `rank <= 1` check that both `order` and `shape` are `None`

The `shape` can be provided for 1D arrays (and presumably it is not a problem if it is provided for scalars?)

yguclu: You're right for 1D arrays of course, but there something really wrong is going on if `shape` is provided for scalars. :wink:

Let me fix the comment above.

EmilyBourne:
@nhamidn This PR is now more than a year old. Do you plan to return to it and complete it at some point? Or would you like me to pick it up?

EmilyBourne (approved):

EmilyBourne:
@yguclu This PR seems to have been forgotten. Can it be merged now?

yguclu:
> @yguclu This PR seems to have been forgotten. Can it be merged now?

Yes, I think so.

yguclu (approved):

PR #837, opened 2021-04-07
Do not require that there is a comment before an else block

### LINKED ISSUE #829: Comments and OpenMP pragmas are printed in the wrong place
Comments and omp pragmas are printer in the wrong place

```python
def func(x : int):
    if x == 0:
        pass
    else:
        # x is not 0
        pass
```

```fortran
subroutine func(x)

  implicit none

  integer(C_INT64_T), value :: x

  if (x == 0_C_INT64_T) then
    ! pass
    !x is not 0
  else
    ! pass
  end if

end subroutine func
```
The same behavior is reproduced when using omp pragmas

EmilyBourne:
Fixes #829 by correcting the following:
- It is not necessary that there is a comment before an else block
- The indentation must mean the same thing throughout. The indentation of an if/else block should be the indentation of its body, not the `else` command
- Use [astunparse](https://pypi.org/project/astunparse/) to add tests to ensure that the parsed code with comments matches the input code

**Additional Commits**
- Allow mac to choose its own version of gfortran (installation fails at linking step with previous install step)
- Correct test_syntax.py to check that there are no syntax errors raised


saidctb (approved):

yguclu (commented):
Can we add a unit test that uses OpenMP and would fail without these changes?

EmilyBourne:
> Can we add a unit test that uses OpenMP and would fail without these changes?

@nhamidn do you have an example to hand?

nhamidn:
> > Can we add a unit test that uses OpenMP and would fail without these changes?
> 
> @nhamidn do you have an example to hand?

for example this code should work fine but because of the issue it won't work because we check if there is a for loop after the pragma:

```
def func(x : int):
    result = 0
    if x < 30:
        return x
    else:
        #$ omp for reduction(+:result)
        for i in range(x):
            result = result + i

    return result
```

EmilyBourne:
> > > Can we add a unit test that uses OpenMP and would fail without these changes?
> > 
> > 
> > @nhamidn do you have an example to hand?
> 
> for example this code should work fine but because of the issue it won't work because we check if there is a for loop after the pragma:
> 
> ```
> def func(x : int):
>     result = 0
>     if x < 30:
>         return x
>     else:
>         #$ omp for reduction(+:result)
>         for i in range(x):
>             result = result + i
> 
>     return result
> ```

This code runs correctly in python though right? So we won't get the wrong result if the fix didn't work?

nhamidn:
> > > > Can we add a unit test that uses OpenMP and would fail without these changes?
> > > 
> > > 
> > > @nhamidn do you have an example to hand?
> > 
> > 
> > for example this code should work fine but because of the issue it won't work because we check if there is a for loop after the pragma:
> > ```
> > from pyccel.decorators import types
> > def func(x : int):
> >     result = 0
> >     if x < 30:
> >         return x
> >     else:
> >         #$ omp for reduction(+:result)
> >         for i in range(x):
> >             result = result + i
> > 
> >     return result
> > ```
> 
> This code runs correctly in python though right? So we won't get the wrong result if the fix didn't work?

yes but without the fix we can't generate a C or Fortran code due to the check of the for loof after the pragma.

pyccel/parser/extend_tree.py
```diff
@@ -101,15 +101,12 @@ def get_last_lineno(ast):
...
+def insert_comments(ast, comment_lines_no, comments, else_no, attr='body'):
     if len(comments) == 0:
         return
     body        = getattr(ast, attr)
 
-    if attr == 'orelse':
-        assert col_offset is not None
-    else:
-        col_offset = body[0].col_offset
+    col_offset = body[0].col_offset
```
saidctb, line 109: the col_offset is not set properly sometimes by the Python parser and different Python versions give different results,  you think that we don't need that?

EmilyBourne: > the col_offset is not set properly sometimes by the Python parser

Interesting, I did not know that. I know that the `col_offset` is not set for ast objects such as `Module`, but then they are not needed here.

> different Python versions give different results

As long as the numbers are consistent within a given version, I don't see this as a problem

> you think that we don't need that?

The reason I changed this is because the following code:
```
def f():
    # 1
    if True:
        # 2
        pass
    else:
        # 3
        for i in range(4):
            # 4
            pass
        # 5
    # 6
```
visits `insert_comments` and sets `col_offset` to the following values:
- `insert_comments(Module, ..., attr='body',...)` sets `col_offset=0` (ok)
- `insert_comments(FunctionDef, ..., attr='body',...)` sets `col_offset=4` (ok)
- `insert_comments(If, ..., attr='body',...)` sets `col_offset=8` (ok)
- `insert_comments(If, ..., attr='orelse',...)` sets `col_offset=4` (wrong)
- `insert_comments(For, ..., attr='body',...)` sets `col_offset=12` (ok)

so the only time we set `col_offset` to something other than `body[0].col_offset` we are setting it to the wrong value

EmilyBourne: The alternative fix would be to modify [this line](https://github.com/pyccel/pyccel/blob/master/pyccel/parser/extend_tree.py#L192) and [this line](https://github.com/pyccel/pyccel/blob/master/pyccel/parser/extend_tree.py#L248) to pass the correct `col_offset`. But I don't know how to calculate that without checking the `col_offset` of items in the body. In the example above the additional each line increases the offset by 4, but that is due to my preferred indenting not anything inherent to python

saidctb: yes,  maybe it's better to add the [unparseast](https://github.com/simonpercivall/astunparse) module to generate the string back to be able to compare it with the original.
because I tried something like
```python
if True:   
    pass  
elif False: 
    pass  
else:  
    # 3  
    pass 
    # 4  
```
and it didn't work

EmilyBourne: Interesting, this works for me with python 3.8

saidctb: I have Python 3.6.9, it puts the last comment after the else block instead of the end of the else block.

saidctb:
```python

Comment (saidctb) [reply pyccel/parser/extend_tree.py:109]: we can do this
```python
class Unparser(astunparse.Unparser): 
    """Unparser the AST""" 
 
    def _CommentLine(self, node): 
        self.write('\n' + ' '*node.col_offset + node.s) 

```
and check that `Uparser(extend_tree(code)) == code`

EmilyBourne:
The problem is caused by the fact that the following code is parsed differently by different versions of python:
```
if a:
    b = (a + 2)
elif (a + 1):
    b = (a + 1)
else:
    b = a
```
python3.6:
```
if a:
    b = (a + 2)
else:
    if (a + 1):
        b = (a + 1)
    else:
        b = a
```
python3.9:
```
if a:
    b = (a + 2)
elif (a + 1):
    b = (a + 1)
else:
    b = a
```

pyccel/parser/extend_tree.py
EmilyBourne: done

README.rst
```diff
@@ -212,6 +212,7 @@ In order to run the unit tests and to get a coverage report, five additional Pyt
   pip3 install --user mpi4py
   pip3 install --user tblib
   pip3 install --user pytest
+  pip3 install --user astunparse
```
yguclu, line 215: We now have six additional packages, not five anymore. Hence in the sentence just above this list I suggest that we change the word "five" with "some" or "a few", so that we will not need to update it in the future.

pyccel/parser/extend_tree.py
```diff
@@ -106,8 +106,9 @@ def insert_comments(ast, comment_lines_no, comments, else_no, attr='body', col_o
         return
     body        = getattr(ast, attr)
 
-    if attr == 'orelse':
-        assert col_offset is not None
+    # Fix necessary for python <= 3.6
+    if attr=='orelse' and isinstance(body[0], IfNode):
+        assert(col_offset is not None)
```
yguclu, line 111: Python's `assert` statement is not a function, so the added parentheses are unnecessary. I personally find them a bit confusing.

tests/parser/test_comments.py
```diff
@@ -0,0 +1,122 @@
...
+            o = o[2:]
+            # Change indentation to match
+            indent = copy_indents[ci]
+            diff = (indent-copy_indents[ci-1])
+            i = ci
+            while i<len(copy_indents) and copy_indents[i]>=indent:
+                copy_indents[i]-=diff
+                i+=1
+        # Check lines
+        assert(o==c)
```
yguclu, line 105: Can we please remove these unnecessary parentheses?

tests/syntax/test_syntax.py
```diff
@@ -15,13 +15,15 @@
...
+    errors.reset()
 
     pyccel = Parser(f)
     pyccel.parse()
 
-    # reset Errors singleton
-    errors = Errors()
-    errors.reset()
+    # Assert syntactic success
+    assert(not errors.has_errors())
```
yguclu, line 26: Can we please remove these unnecessary parentheses?

yguclu (changes requested):
This is a very thorough PR @EmilyBourne, well done!

I have a few minor comments

pyccel/parser/extend_tree.py
EmilyBourne: I find it confusing without them ^^ It looks too much like python 2 syntax. I didn't realise that it wasn't a function though, I thought it was like `print` I will change them

EmilyBourne: Fixed in 3a24360

README.rst
EmilyBourne: Fixed in 92deae4

tests/parser/test_comments.py
EmilyBourne: Fixed in 3a24360

tests/syntax/test_syntax.py
EmilyBourne: Fixed in 3a24360

pyccel/parser/extend_tree.py
yguclu: Thanks 😄

yguclu (approved):

EmilyBourne:
Waiting for [github actions problem](https://www.githubstatus.com/) to be fixed

EmilyBourne:
Fixed

PR #840, opened 2021-04-07
Fixing Arithmetic / Memory Complexity

abkhaoula:
Fixing the computation of Arithmetic , Memory and Intensity Complexity


pyccel/complexity/basic.py
```diff
@@ -25,16 +38,230 @@ def __init__(self, filename_or_text):
...
+#        print('>>> ', expr, type(expr))
+
+        classes = type(expr).__mro__
+        for cls in classes:
+            method = '_cost_' + cls.__name__
+            if hasattr(self, method):
+                obj = getattr(self, method)(expr, **settings)
+                return obj
+            else:
+                raise NotImplementedError('{} not available for {}'.format(method, type(expr)))
```
EmilyBourne, line 155: Shouldn't this `raise` be after the for block?

pyccel/complexity/arithmetic.py
```diff
@@ -4,87 +4,270 @@
...
+            return ops + 1
+        else:
+            return ops + Symbol('ARCTANH')
+
+    def _cost_PyccelAdd(self, expr, **settings):
+        ops = sum(self._cost(i, **settings) for i in expr.args)
+        if self.mode:
+            return ops + 1
+        else:
+            return ops + ADD
```
EmilyBourne, line 199: I see that you are creating `Symbol` objects for the Numpy classes, but you are using the same `Symbol` object for basic operators. I assume this is done to reduce memory consumption for functions that you don't think will be used often and reduce calculation time for operators which are likely to be used a lot?

pyccel/complexity/memory.py
```diff
@@ -10,228 +10,187 @@
...
+        return sum(self._cost(i, **settings) for i in expr.args)
+
+    def _cost_NumpyCosh(self, expr, **settings):
+        return sum(self._cost(i, **settings) for i in expr.args)
+
+    def _cost_NumpyTanh(self, expr, **settings):
+        return sum(self._cost(i, **settings) for i in expr.args)
+
+    def _cost_NumpyArcsinh(self, expr, **settings):
+        return sum(self._cost(i, **settings) for i in expr.args)
```
EmilyBourne, line 186: If all these Numpy functions do the same thing, you should be able to simplify it to something like:
```
def _cost_NumpyUfuncBase(self, expr, **settings):
        return sum(self._cost(i, **settings) for i in expr.args
```
(even if 1 or 2 do something different you can just specialise these cases)

pyccel/complexity/memory.py
```diff
@@ -10,228 +10,187 @@
...
+    def _cost_AugAssign(self, expr, **settings):
+        ntimes = self._compute_size_lhs(expr)
+        # Right? Because x += a should also READ x which is in lrs
+        return  ntimes * (self._cost(expr.rhs, **settings) + WRITE + READ)
 
-    elif isinstance(expr, For):
-        s = pyccel_to_sympy(expr.iterable, symbol_map, used_names).size
-        ops = sum(count_access(i, visual) for i in expr.body.body)
-        return ops*s
+    def _cost_PyccelAdd(self, expr, **settings):
```
EmilyBourne, line 65: As below, can't you use `_cost_PyccelOperator` to reduce code duplication here?

pyccel/complexity/basic.py
```diff
@@ -25,16 +38,230 @@ def __init__(self, filename_or_text):
...
+        return self._costs
+
+    @property
+    def shapes(self):
+        """Returns shapes of allocated arrays."""
+        return self._shapes
+
+    @property
+    def mode(self):
+        return self._mode
```
EmilyBourne, line 81: What does this do? Can you add a docstring please

pyccel/complexity/basic.py
```diff
@@ -25,16 +38,230 @@ def __init__(self, filename_or_text):
...
+
+    @property
+    def visual(self):
+        return self._visual
```
EmilyBourne, line 85: What does this do? Can you add a docstring please

pyccel/complexity/basic.py
```diff
@@ -25,16 +38,230 @@ def __init__(self, filename_or_text):
...
+        simplify: bool
+            if true, calls sympy simplify function
+
+        bigo: list
+            a list containing variable names (str) for which we want to take the
+            highest degree (expansion serie)
+        """
+        # ...
+        self._visual = visual
+        self._mode = mode
```
EmilyBourne, line 106: Don't you need to check that the user provided a valid mode?

pyccel/complexity/arithmetic.py
```diff
@@ -4,87 +4,270 @@
...
+from collections import OrderedDict
+
 from sympy import count_ops as sympy_count_ops
-from sympy import Tuple
+from sympy import Symbol
 
-from pyccel.ast.core     import For, Assign, CodeBlock, Comment
-from pyccel.ast.numpyext import NumpyZeros, NumpyOnes
+from pyccel.ast.core     import Comment, EmptyNode
+from pyccel.ast.core     import AddOp, SubOp, MulOp, DivOp
```
EmilyBourne, line 26: These classes were recently removed (see PR#831)

pyccel/complexity/basic.py
EmilyBourne: I think you have misunderstood something. With the behaviour that you have implemented the for loop is useless. I assume the behaviour is supposed to match what is done in the semantic and syntactic parser and the code printers. The way it works there is as follows:
Suppose we want to print `NumpyTanh` first we collect the inheritance tree of `NumpyTanh` which gives us:
`('NumpyTanh', 'NumpyUfuncUnary', 'NumpyUfuncBase', 'PyccelInternalFunction', 'PyccelAstNode', 'Basic')`
This means that the cost functions which are acceptable for visiting this object are:
- `_cost_NumpyTanh`
- `_cost_NumpyUfuncUnary`
- `_cost_NumpyUfuncBase`
- `_cost_PyccelInternalFunction`
- `_cost_PyccelAstNode`
- `'_cost_Basic`

We run through these possible functions choosing the one which is the most specialised and only if none of these methods exist do we need to raise an error

pyccel/complexity/arithmetic.py
EmilyBourne: Yes of course, I just wondered whether there was a reason to create a symbol as in line 192 for some cases but have one symbol for the whole file (e.g `ADD` used in line 199). Presumably this is a question that only Ahmed can answer. Normally I would prefer to do the same thing throughout the file rather than using different strategies in different places.

But if creating `Symbol`s is slow then I can see my explanation above being a possible excuse for having 2 strategies

pyccel/complexity/memory.py
EmilyBourne: See above explanation of `_cost` method

pyccel/complexity/arithmetic.py
abkhaoula: Oh right, I understand now

pyccel/complexity/basic.py
abkhaoula: I couldn't figure out what this is for, it seems like there is no use for it yet

EmilyBourne: Maybe we can remove it? @ratnania ?

pyccel/complexity/basic.py
abkhaoula: Thank you

tests/complexity/scripts/ex1.py
```diff
@@ -0,0 +1,355 @@
+# ...
```
EmilyBourne, line 1: Usually we don't insist on documentation in test files. So all new files created in the `tests/` folder should contain the following in the first line of the file:
```
```

pyccel/complexity/intensity.py
```diff
@@ -0,0 +1,152 @@
...
+    expr = sympify(str(expr))
+    if expr.atoms(sympy.Function):
+        for shape_expr in expr.atoms(sympy.Function):
+            expr = expr.subs(shape_expr, "shape_" + str(shape_expr.args[0]) + '_' + str(shape_expr.args[1]))
+        P = Poly(expr, *args)
+        lt = LT(P)
+        for var in lt.free_symbols:
+            var = str(var)
+            if "shape" in var:
+                lt = lt.subs(var, "shape(" + var.split("_")[1] + "," + var.split("_")[2] + ")")
```
EmilyBourne, line 75: I think this needs updating to match your new definition of `SHAPE`

pyccel/complexity/intensity.py
```diff
@@ -0,0 +1,154 @@
...
+    args = list(args)
+
+    # TODO _leading_term does not retrieve the right output
+    lt_f = _leading_term(f, *args)
+    lt_m = _leading_term(m, *args)
+
+    return lt_f/lt_m
+
+# ==============================================================================
+class ComputationalIntensity(object):
```
EmilyBourne, line 119: This class needs a docstring

tests/complexity/scripts/ex_assembly.py
```diff
@@ -50,11 +50,11 @@ def assemble_matrix_ex01(ne1: 'int', ne2: 'int',
                             v = 0.0
                             for g1 in range(0, k1):
                                 for g2 in range(0, k2):
-                                    bi_0 = basis_1[ie1, il_1, 0, g1] * basis_2[ie2, il_2, 0, g2]
+                                    # bi_0 = basis_1[ie1, il_1, 0, g1] * basis_2[ie2, il_2, 0, g2]
```
EmilyBourne, line 53: I think this has changed your computational intensity results. Do you not need `bi_0`/`bj_0`? Do you think it makes sense to change the expected intensity or are these here for a reason? We could just return them to avoid them being seen as useless?

abkhaoula: I think as long as it is just a function we are using for testing purpose it would be better to delete them and edit the complexity value

EmilyBourne: Ok. Then should we delete them rather than just commenting?

tests/complexity/test_arithmetic.py
```diff
@@ -0,0 +1,220 @@
...
+    mode = None
+
+    complexity = OpComplexity(f)
+    complexity.cost(mode=mode, simplify=True, bigo=None)
+    # complexity.cost(mode=mode, simplify=True, bigo=['n'])
+
+    print('----------------------')
+    i = 0
+    comp = [n**3*(ADD + MUL),
+            n**3*(ADD + MUL)
```
EmilyBourne, line 193: Could you add a comment mentioning the functions as you did above please

tests/complexity/test_arithmetic.py
```diff
@@ -0,0 +1,220 @@
...
+
+# ==============================================================================
+def test_complexity_mxm():
+
+    f = path_dir + '/mxm.py'
+    mode = None
+
+    complexity = OpComplexity(f)
+    complexity.cost(mode=mode, simplify=True, bigo=None)
+    # complexity.cost(mode=mode, simplify=True, bigo=['n'])
```
EmilyBourne, line 188: Is this in comment due to the problem with `_leading_terms` mentioned above?

tests/complexity/test_arithmetic.py
```diff
@@ -0,0 +1,220 @@
...
+    f = path_dir + '/qr.py'
+    mode = None
+
+    complexity = OpComplexity(f)
+    complexity.cost(mode=mode, simplify=True, bigo=None)
+    # complexity.cost(mode=mode, simplify=True, bigo=['n'])
+
+    print('----------------------')
+    i = 0
+    comp = [n*(ADD*n**2 - ADD + 3*DIV*n - 3*DIV + MUL*n**2 - MUL + 6*SQRT + 3*SUB*n + 3*SUB)/6
```
EmilyBourne, line 215: See above, please add the function name for clarity

tests/complexity/test_intensity.py
```diff
@@ -0,0 +1,226 @@
...
+    f = path_dir + '/ex_assembly.py'
+    mode = None
+
+    complexity = ComputationalIntensity(f)
+    complexity.cost(mode=mode)
+#    complexity.cost(mode=mode, simplify=True, bigo=['n'])
+
+    print('----------------------')
+    i = 0
+    comp = [(10 * MUL + 3 * ADD)/(18*READ + 7*WRITE)
```
EmilyBourne, line 174: See above, function name please

tests/complexity/test_intensity.py
```diff
@@ -0,0 +1,226 @@
...
+    mode = None
+
+    complexity = ComputationalIntensity(f)
+    complexity.cost(mode=mode)
+#    complexity.cost(mode=mode, simplify=True, bigo=['n'])
+
+    print('----------------------')
+    i = 0
+    comp = [(ADD + MUL)/(3*READ + WRITE),
+            b*(ADD+MUL) /(2 * READ + 2 * WRITE + b * (3 * READ + WRITE))
```
EmilyBourne, line 198: See above, function name please

tests/complexity/test_intensity.py
```diff
@@ -0,0 +1,226 @@
...
+    f = path_dir + '/qr.py'
+    mode = None
+
+    complexity = ComputationalIntensity(f)
+    complexity.cost(mode=mode)
+#    complexity.cost(mode=mode, simplify=True, bigo=['n'])
+
+    print('----------------------')
+    i = 0
+    comp = [((ADD + MUL)/3)/(READ + WRITE/3),
```
EmilyBourne, line 220: See above, function name please

tests/complexity/test_memory.py
```diff
@@ -0,0 +1,214 @@
...
+    mode = None
+
+    complexity = MemComplexity(f)
+    complexity.cost(mode=mode, simplify=True, bigo=None)
+#    complexity.cost(mode=mode, simplify=True, bigo=['n'])
+
+    print('----------------------')
+    i = 0
+    comp = [READ + WRITE + n**3*(3*READ + WRITE),
+            (b*(READ + 3*WRITE*b**2 + WRITE) + n**2*(2*b*(READ + WRITE) + n*(2*READ + 2*WRITE + b*(3*READ + WRITE))))/b
```
EmilyBourne, line 186: See above, function name please

tests/complexity/test_memory.py
```diff
@@ -0,0 +1,214 @@
...
+    f = path_dir + '/qr.py'
+    mode = None
+
+    complexity = MemComplexity(f)
+    complexity.cost(mode=mode, simplify=True, bigo=None)
+#    complexity.cost(mode=mode, simplify=True, bigo=['n'])
+
+    print('----------------------')
+    i = 0
+    comp = [READ*n**3/2 + 3*READ*n**2/2 + READ + WRITE*n**3/6 + 3*WRITE*n**2/2 + WRITE*n/3 + WRITE,
```
EmilyBourne, line 208: See above, function name please

tests/complexity/test_memory.py
```diff
@@ -0,0 +1,214 @@
...
+    f = path_dir + '/ex_assembly.py'
+    mode = None
+
+    complexity = MemComplexity(f)
+    complexity.cost(mode=mode, simplify=True, bigo=None)
+#    complexity.cost(mode=mode, simplify=True, bigo=['n'])
+
+    print('----------------------')
+    i = 0
+    comp = [2*READ + WRITE*k1*k2 + WRITE*(p1 + 1)*(p2 + 1) + 2*WRITE + ne1*(READ + WRITE + ne2*(READ + WRITE*k1*k2 + WRITE + (p1 + 1)**2*(p2 + 1)**2*(14*READ + 6*WRITE + k1*k2*(18*READ + 7*WRITE)) + (p1 + 1)*(p2 + 1)*(READ*(p1 + 1)*(p2 + 1) + WRITE) + (p1 + 1)*(p2 + 1)*(READ + WRITE + k1*(READ + WRITE + k2*(5*READ + 2*WRITE)))))]
```
EmilyBourne, line 163: See above, function name please. Also could you please break this line up so it fits on a screen please

tests/complexity/test_memory.py
```diff
@@ -0,0 +1,214 @@
...
+p1 = Symbol('p1')
+p2 = Symbol('p2')
+ne1 = Symbol('ne1')
+ne2 = Symbol('ne2')
+
+base_dir = os.path.dirname(os.path.realpath(__file__))
+path_dir = os.path.join(base_dir, 'scripts')
+
+files = sorted(os.listdir(path_dir))
+files = [os.path.join(path_dir,f) for f in files if (f.endswith(".py"))]
```
EmilyBourne, line 30: Are these used?

tests/complexity/test_intensity.py
```diff
@@ -0,0 +1,226 @@
...
+p1 = Symbol('p1')
+p2 = Symbol('p2')
+ne1 = Symbol('ne1')
+ne2 = Symbol('ne2')
+
+base_dir = os.path.dirname(os.path.realpath(__file__))
+path_dir = os.path.join(base_dir, 'scripts')
+
+files = sorted(os.listdir(path_dir))
+files = [os.path.join(path_dir,f) for f in files if (f.endswith(".py"))]
```
EmilyBourne, line 47: Are these used?

tests/complexity/test_arithmetic.py
```diff
@@ -0,0 +1,220 @@
...
+p1 = Symbol('p1')
+p2 = Symbol('p2')
+k1 = Symbol('k1')
+k2 = Symbol('k2')
+
+base_dir = os.path.dirname(os.path.realpath(__file__))
+path_dir = os.path.join(base_dir, 'scripts')
+
+files = sorted(os.listdir(path_dir))
+files = [os.path.join(path_dir,f) for f in files if (f.endswith(".py"))]
```
EmilyBourne, line 45: Are these used?

pyccel/complexity/basic.py
```diff
@@ -25,16 +39,242 @@ def __init__(self, filename_or_text):
...
+        e = iter_range.stop
+        e = pyccel_to_sympy(e, self._symbol_map, self._used_names)
+
+        step = iter_range.step
+        step = pyccel_to_sympy(step, self._symbol_map, self._used_names)
+        # ...
+
+        end = ((e-b) / step)-1
+        # translate so that we always start from 0
+        return summation(ops, (i, 0, end))
```
EmilyBourne, line 213: Translating so that the index always starts from 0 leads to wrong results if loops inside loops depend on the index. E.g:
```
def test(degree : int):
    a = 0
    for j in range(1, degree+1):
        for r in range(j):
            a += 1
return a
```

PR #841, opened 2021-04-08
Fix stack arrays in C

EmilyBourne:
Fixes #760 :
- Ensure that the shape of the stack array is not a call to its shape property at allocation
- Create a local variable containing the data of the stack array if the size is not a literal

PR #845, opened 2021-04-12
Fix memory allocation for nd stack array (n>1)

### LINKED ISSUE #844: Maths is not correctly implemented for shape of nd stack array, n>1
**Describe the bug**
The size of the data allocated for a 2d array is calculated using `'*'.join` instead of `PyccelMul` in cases where the shape is already an equation this can lead to missing parentheses

**To Reproduce**
```python
from numpy import empty

@stack_array('theCoeffs')
def f(deg1 : 'int', deg2 : 'int'):
    theCoeffs = empty([deg1+1,deg2+1])
```

```bash/fortran/c
#include "tmp.h"
#include <stdlib.h>
#include <stdint.h>
#include "ndarrays.h"

void f(int64_t deg1, int64_t deg2)
{
    t_ndarray theCoeffs;
    double theCoeffs_data_0001[deg1 + 1*deg2 + 1];
    theCoeffs = (t_ndarray){
        .nd_double=theCoeffs_data_0001,
        .shape=(int64_t[]){deg1 + 1, deg2 + 1},
        .strides=(int64_t[2]){0},
        .nd=2,
        .type=nd_double,
        .is_view=false
    };
    stack_array_init(&theCoeffs);
}
```

**Expected behavior**
```
#include "tmp.h"
#include <stdlib.h>
#include <stdint.h>
#include "ndarrays.h"

void f(int64_t deg1, int64_t deg2)
{
    t_ndarray theCoeffs;
    double theCoeffs_data_0001[(deg1 + 1) * (deg2 + 1)];
    theCoeffs = (t_ndarray){
        .nd_double=theCoeffs_data_0001,
        .shape=(int64_t[]){deg1 + 1, deg2 + 1},
        .strides=(int64_t[2]){0},
        .nd=2,
        .type=nd_double,
        .is_view=false
    };
    stack_array_init(&theCoeffs);
}
```

**Language**
C

EmilyBourne:
The size of the data allocated for a nd array was calculated using '*'.join instead of `PyccelMul`. In cases where the shape is already an equation this can lead to missing parentheses. This PR fixes this (fixes #844). A test is also added

[BUGFIX] `np.mod(False, False)` causes division by 0 so these tests are removed

### DIFF
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -323,7 +323,9 @@ class CCodePrinter(CodePrinter):
             sympy_shapes = [pyccel_to_sympy(s, symbol_map, used_names_tmp) for s in lhs.alloc_shape]
 
             length = functools.reduce(operator.mul, sympy_shapes)
-            length_code = '*'.join(self._print(i) for i in lhs.alloc_shape)
+            printable_length = functools.reduce(PyccelMul, lhs.alloc_shape)
+
+            length_code = self._print(printable_length)
 
             if length.is_constant():
                 buffer_array = "({declare_dtype}[{length}]){{}}".format(
--- a/tests/epyccel/recognised_functions/test_numpy_funcs.py
+++ b/tests/epyccel/recognised_functions/test_numpy_funcs.py
@@ -4116,14 +4116,9 @@ def test_numpy_mod_scalar(language):
     f_bl_true_output = epyccel_func(True)
     test_bool_true_output = get_mod(True)
 
-    f_bl_false_output = epyccel_func(False)
-    test_bool_false_output = get_mod(False)
-
     assert f_bl_true_output == test_bool_true_output
-    assert f_bl_false_output == test_bool_false_output
 
     assert matching_types(f_bl_true_output, test_bool_true_output)
-    assert matching_types(f_bl_false_output, test_bool_false_output)
 
     f_integer_output = epyccel_func(integer)
     test_int_output  = get_mod(integer)
@@ -4207,7 +4202,7 @@ def test_numpy_mod_array_like_1d(language):
 
     size = 5
 
-    bl = randint(0, 1, size=size, dtype= bool)
+    bl = np.full(size, True, dtype= bool)
 
     integer8 = randint(min_int8, max_int8, size=size, dtype=np.int8)
     integer16 = randint(min_int16, max_int16, size=size, dtype=np.int16)
@@ -4266,7 +4261,7 @@ def test_numpy_mod_array_like_2d(language):
 
     size = (2, 5)
 
-    bl = randint(0, 1, size=size, dtype= bool)
+    bl = np.full(size, True, dtype= bool)
 
     integer8 = randint(min_int8, max_int8, size=size, dtype=np.int8)
     integer16 = randint(min_int16, max_int16, size=size, dtype=np.int16)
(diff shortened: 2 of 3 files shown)


saniQQ (approved):

pyccel/codegen/printing/ccode.py
```diff
@@ -323,7 +323,9 @@ def arrayFill(self, expr):
             sympy_shapes = [pyccel_to_sympy(s, symbol_map, used_names_tmp) for s in lhs.alloc_shape]
 
             length = functools.reduce(operator.mul, sympy_shapes)
-            length_code = '*'.join(self._print(i) for i in lhs.alloc_shape)
+            printable_length = functools.reduce(PyccelMul, lhs.alloc_shape)
+
+            length_code = self._print(printable_length)
```
yguclu, line 328: Do we still need the `sympy_shapes` variable?

yguclu (changes requested):

pyccel/codegen/printing/ccode.py
EmilyBourne: Yes `length` and `length_code` are 2 different objects. `length` is a sympy object containing the length so we can tell if the length is known at compile time or not. `length_code` is the code that will be printed for the array size. If the `length` is constant then the code is generated differently. E.g:
1.
```
@stack_array('a')
def f():
    a = np.empty(3)
```
generates:
```
void f(void)
{
    t_ndarray a;
    a = (t_ndarray){
        .nd_double=(double[3]){},
        .shape=(int64_t[]){3},
        .strides=(int64_t[1]){0},
        .nd=1,
        .type=nd_double,
        .is_view=false
    };
    stack_array_init(&a);
}
```
which is compact and requires no additional variables.
2. For the following example:
```
@stack_array('a')
def f(d : int):
    a = np.empty(d)
```
the same cannot be done as a literal cannot have unknown size. We must therefore do:
```
void f(int64_t d)
{
    t_ndarray a;
    double a_data_0001[d];
    a = (t_ndarray){
        .nd_double=a_data_0001,
        .shape=(int64_t[]){d},
        .strides=(int64_t[1]){0},
        .nd=1,
        .type=nd_double,
        .is_view=false
    };
    stack_array_init(&a);
}
```

This was the subject of PR #841

EmilyBourne: This may be more complicated than it needs to be though? Perhaps it is sufficient to do an atomic search for an instance of a `Variable`?

yguclu: I was thinking that maybe, once we simplify `lhs.alloc_shape`, it should reduce to a `Literal` object if it is a constant. Wishful thinking on my side?

EmilyBourne: I think so. While something like `2+2` should reduce to a `Literal` something more complicated like: `(1+d)*(1-d)+d**2` will not. That said, I don't have any examples of where the shape would be this complicated. Hence why it is perhaps sufficient to just check if the expression contains a `Variable`

yguclu: I see. The current implementation is the safest option for now, we'll think about something faster the day when this becomes a computational bottleneck.

yguclu (approved):
Could you please expand the PR's description a bit?

EmilyBourne:
> Could you please expand the PR's description a bit?

Better?

PR #846, opened 2021-04-12
Fix flattening the nd array when dim > 2

### LINKED ISSUE #751: np.array in C does not handle nd arrays, only 1 or 2d
**Describe the bug**
When passing lists of lists of ... to `np.array` an error is raised in C if the dimension is greater than 2

**To Reproduce**
```python
def array_real_3d_C_array_initialization_2():
    from numpy import array
    x = array([[[0., 1., 2., 3.], [4., 5., 6., 7.], [8., 9., 10., 11.]],
              [[12., 13., 14., 15.], [16., 17., 18., 19.], [20., 21., 22., 23.]]], order='C')
```

Provide the error message:
```bash/fortran/c
pyccel:
 |fatal [codegen]: array_view.py [4,16]| Pyccel has encountered syntax that has not been implemented yet. Please create an issue at https://github.com/pyccel/pyccel/issues and provide a small example of your problem. Do not forget to specify your target language (PythonList(0.0, 1.0, 2.0, 3.0))
```

**Language**
C

saniQQ:
Fixes #751 where the `reduce` function does not flatten the array enough. This is achieved by adding the recursive method `_flatten_list` to the C code printer.

### DIFF
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -258,6 +258,13 @@ class CCodePrinter(CodePrinter):
         rows, cols = mat.shape
         return ((i, j) for i in range(rows) for j in range(cols))
 
+    def _flatten_list(self, irregular_list):
+        if isinstance(irregular_list, (PythonList, PythonTuple)):
+            f_list = [element for item in irregular_list for element in self._flatten_list(item)]
+            return f_list
+        else:
+            return [irregular_list]
+
     #========================== Numpy Elements ===============================#
     def copy_NumpyArray_Data(self, expr):
         """ print the assignment of a NdArray
@@ -275,7 +282,6 @@ class CCodePrinter(CodePrinter):
         """
         rhs = expr.rhs
         lhs = expr.lhs
-        print(lhs, rhs)
         if rhs.rank == 0:
             raise NotImplementedError(str(expr))
         dummy_array_name, _ = create_incremented_string(self._parser.used_names, prefix = 'array_dummy')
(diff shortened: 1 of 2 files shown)


pyccel/codegen/printing/ccode.py
```diff
@@ -274,6 +275,7 @@ def copy_NumpyArray_Data(self, expr):
         """
         rhs = expr.rhs
         lhs = expr.lhs
+        print(lhs, rhs)
```
EmilyBourne, line 278: Forgotten print

pyccel/codegen/printing/ccode.py
```diff
@@ -282,7 +284,9 @@ def copy_NumpyArray_Data(self, expr):
         arg = rhs.arg
         if rhs.rank > 1:
             # flattening the args to use them in C initialization.
-            arg = functools.reduce(operator.concat, arg)
+            flatten_list = lambda irregular_list:[element for item in irregular_list for element in flatten_list(item)] if isinstance(irregular_list, (PythonList, PythonTuple)) else [irregular_list]
```
EmilyBourne, line 287: Is there a reason for this to be a lambda rather than a local function? In any case it needs breaking up into several lines as it does not fit on my screen

EmilyBourne (changes requested):
Please add a test that would have failed before this PR. E.g. 4d

EmilyBourne (approved):

yguclu (approved):
Nice job @saniQQ!

PR #851, opened 2021-04-13
Set shape to PyccelArraySize if Variable is a pointer

### LINKED ISSUE #847: Shape of pointer is not updated when pointer is reassigned
**Describe the bug**
A pointer can point at objects with different shapes as long as they have the same rank. However as we set the shape in the constructor and never modify it, if a pointer is reallocated then it has the wrong shape

**To Reproduce**
```python
from numpy import array, shape
a = array([1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
b = array([1, 2, 3, 4, 5, 6, 7, 8, 9])
c = a
c = a[1:]
d = c*b
print(d)
```

raises the following error when translated with pyccel:
```bash
pyccel:
 |fatal [semantic]: tmp.py [6,4]| operands could not be broadcast together with shapes (<pyccel.ast.literals.LiteralInteger object at 0x7f72c92c7b80>,) (<pyccel.ast.literals.LiteralInteger object at 0x7f72c8f4b950>,) (c * b)
```
but runs correctly in python

**Expected behavior**
The shape should be set to a tuple of `PyccelArraySize`s when the `Variable` is a pointer

EmilyBourne:
Set shape to PyccelArraySize if Variable is a pointer as it can point to objects of different shapes during its lifetime. Fixes #847

**Commit Summary**
- Add `__repr__` printer for Literal
- Don't print tuple for error message (calls `__repr__`)
- Reorder argument collection in `Variable` so `is_pointer` is known before calculating shape
- Set shape to PyccelArraySize if Variable is a pointer as it can point to objects of different shapes during its lifetime (fixes #847)
- Add test

### DIFF
--- a/pyccel/ast/literals.py
+++ b/pyccel/ast/literals.py
@@ -49,6 +49,9 @@ class Literal(PyccelAstNode):
     def python_value(self):
         """ Get python literal represented by this instance """
 
+    def __repr__(self):
+        return "Literal({})".format(repr(self.python_value))
+
     def __str__(self):
         return str(self.python_value)
 
--- a/pyccel/ast/operators.py
+++ b/pyccel/ast/operators.py
@@ -88,8 +88,10 @@ def broadcast(shape_1, shape_2):
                 and not (sy_e1 - sy_e2).is_constant():
             new_shape.append(e1)
         else:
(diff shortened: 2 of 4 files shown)


yguclu (approved):
Nice fix!

PR #852, opened 2021-04-13
Remove useless statements and raise a warning

### LINKED ISSUE #425: missing semicolon when performing standalones arithmetic or bitwise operations
**Performing a standalone arithmetic or bitwise operation leads to a missing semicolon in c**
performing an arithmetic operation without it being assigned to a variable nor being inside an if or while statement... leads to a missing semicolon in the generated code.

**To Reproduce**

```
@types('int', 'int')
def sum_two_numbers(x, y):
    x + y
```

**The Generated Code**
```
void sum_two_numbers(long x, long y)
     x + y
```

**Error Message**
```
/.../__pyccel__/ntest.c: In function ‘sum_tow_numbers’:
/.../__pyccel__/ntest.c:4:7: warning: statement with no effect [-Wunused-value]
     x + y
     ~~^~~
/.../__pyccel__/ntest.c:5:1: error: expected ‘;’ before ‘}’ token
 }
 ^
```

**Language**
C

EmilyBourne:
Remove any statements which have no effect (e.g. standalone arithmetic) and raise a warning to warn the user. Fixes #425

Ensure modulo tests do not test 0 (causes floating point exception)

### DIFF
--- a/tests/epyccel/recognised_functions/test_numpy_funcs.py
+++ b/tests/epyccel/recognised_functions/test_numpy_funcs.py
@@ -4108,9 +4108,7 @@ def test_numpy_mod_scalar(language):
     assert matching_types(f_bl_true_output, test_bool_true_output)
 
     def test_int(min_int, max_int, dtype):
-        integer = randint(min_int, max_int, dtype=dtype)
-        if 0 == integer:
-            integer += 1
+        integer = randint(min_int, max_int, dtype=dtype) or 1
 
         f_integer_output = epyccel_func(integer)
         test_int_output  = get_mod(integer)
@@ -4188,8 +4186,7 @@ def test_numpy_mod_array_like_1d(language):
 
     def test_int(min_int, max_int, dtype):
         integer = randint(min_int, max_int-1, size=size, dtype=dtype)
-        if 0 in integer:
-            integer += 1
+        integer = np.where(integer==0, integer, 1)
         assert epyccel_func(integer) == get_mod(integer)
 
     test_int(min_int8 , max_int8 , np.int8)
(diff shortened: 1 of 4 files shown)


mostafaMamoni (approved):

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -4099,58 +4099,37 @@ def get_mod(a):
...
     assert f_bl_true_output == test_bool_true_output
-
     assert matching_types(f_bl_true_output, test_bool_true_output)
 
-    f_integer_output = epyccel_func(integer)
-    test_int_output  = get_mod(integer)
+    def test_int(min_int, max_int, dtype):
+        integer = randint(min_int, max_int, dtype=dtype)
+        if 0 == integer:
+            integer += 1
```
yguclu, line 4113: Can't you just write this?
```python
integer = randint(min_int, max_int, dtype=dtype) or 1
```

EmilyBourne: You are right, I am not used to this syntax. Fixed in 5bd7230

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -4248,8 +4245,7 @@ def get_mod(arr):
 
     def test_int(min_int, max_int, dtype):
         integer = randint(min_int, max_int-1, size=size, dtype=dtype)
-        if 0 in integer:
-            integer += 1
+        integer = np.where(integer==0, integer, 1)
```
yguclu: I had not realized `integer` was an array!!

EmilyBourne: It wasn't  in the line you commented on. But I realised that this was wrong too ^^'

yguclu: But shouldn't this be the opposite?
```python
integer = np.where(integer == 0, 1, integer)
```
According to the [documentation](https://numpy.org/doc/stable/reference/generated/numpy.where.html):

> `numpy.where(condition[, x, y])`
> Return elements chosen from x or y depending on condition.
> Where True, yield x, otherwise yield y.

EmilyBourne: Oops. Fixed in 3ff7a13

PR #862, opened 2021-04-15
Match results of integer modulo operations between languages.

### LINKED ISSUE #854: Modulo does not behave in the same way in C as in Fortran/Python
**Describe the bug**
Negative numbers are not handled in the same way by the modulo operator in different languages

**To Reproduce**
```python
a = (-1) % 10
print(a)
```
outputs `9`

The fortran code is:
```bash/fortran/c
integer(C_INT64_T) :: a

a = MODULO((-1_C_INT64_T),10_C_INT64_T)
print *, a
```
which also outputs `9`

The C code is:
```
int64_t a;
a = (-1) % 10;
printf("%ld\n", a);
```
which outputs `-1`

**Expected behavior**
C should give the same results as python

**Language**
C

**Additional context**
Causing problems in pygyro

MostafaMamouni:
Fix #854 using an expression that matches the result of modulo operations between C with Python.

Commit summary:
- Use an arithmetic expression instead of modulo operator/function in C printer;
- Use C macros to improve readability;
- Test negative numbers.

### DIFF
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -626,13 +626,13 @@ class CCodePrinter(CodePrinter):
         second = self._print(expr.args[1])
 
         if expr.dtype is NativeInteger():
-            return "MOD_PYC({n}, {base})".format(n=first, base=second)
+            return "MOD_PYC1({n}, {base})".format(n=first, base=second)
 
         if expr.args[0].dtype is NativeInteger():
             first = self._print(NumpyFloat(expr.args[0]))
         if expr.args[1].dtype is NativeInteger():
             second = self._print(NumpyFloat(expr.args[1]))
-        return "FMOD_PYC({n}, {base})".format(n=first, base=second)
+        return "FMOD_PYC1({n}, {base})".format(n=first, base=second)
 
     def _print_PyccelPow(self, expr):
         b = expr.args[0]
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -620,18 +620,19 @@ class CCodePrinter(CodePrinter):
 
     def _print_PyccelMod(self, expr):
         self._additional_imports.add("math")
+        self._additional_imports.add("pyc_math")
 
         first = self._print(expr.args[0])
         second = self._print(expr.args[1])
 
         if expr.dtype is NativeInteger():
-            return "((({n} % {base}) + {base}) % {base})".format(n=first, base=second)
+            return "MOD_PYC({n}, {base})".format(n=first, base=second)
 
         if expr.args[0].dtype is NativeInteger():
             first = self._print(NumpyFloat(expr.args[0]))
         if expr.args[1].dtype is NativeInteger():
             second = self._print(NumpyFloat(expr.args[1]))
-        return "(fmod(((fmod({n}, {base})) + {base}), {base}))".format(n=first, base=second)
+        return "FMOD_PYC({n}, {base})".format(n=first, base=second)
 
     def _print_PyccelPow(self, expr):
         b = expr.args[0]
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -760,7 +760,7 @@ class CCodePrinter(CodePrinter):
                     args = []
                 for_index = Variable(NativeInteger(), name = self._parser.get_new_name('i'))
                 self._additional_declare.append(for_index)
-                max_index = PyccelMinus(PythonLen(orig_args[i]), LiteralInteger(1), simplify = True)
+                max_index = PyccelMinus(orig_args[i].shape[0], LiteralInteger(1), simplify = True)
                 for_range = PythonRange(max_index)
                 print_body = [ orig_args[i][for_index] ]
                 if orig_args[i].rank == 1:
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -626,13 +626,13 @@ class CCodePrinter(CodePrinter):
         second = self._print(expr.args[1])
 
         if expr.dtype is NativeInteger():
-            return "MOD_PYC3({n}, {base})".format(n=first, base=second)
+            return "MOD_PYC1({n}, {base})".format(n=first, base=second)
 
         if expr.args[0].dtype is NativeInteger():
             first = self._print(NumpyFloat(expr.args[0]))
         if expr.args[1].dtype is NativeInteger():
             second = self._print(NumpyFloat(expr.args[1]))
-        return "FMOD_PYC3({n}, {base})".format(n=first, base=second)
+        return "FMOD_PYC1({n}, {base})".format(n=first, base=second)
 
     def _print_PyccelPow(self, expr):
         b = expr.args[0]
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -626,13 +626,13 @@ class CCodePrinter(CodePrinter):
         second = self._print(expr.args[1])
 
         if expr.dtype is NativeInteger():
-            return "MOD_PYC1({n}, {base})".format(n=first, base=second)
+            return "MOD_PYC3({n}, {base})".format(n=first, base=second)
 
         if expr.args[0].dtype is NativeInteger():
             first = self._print(NumpyFloat(expr.args[0]))
         if expr.args[1].dtype is NativeInteger():
             second = self._print(NumpyFloat(expr.args[1]))
-        return "FMOD_PYC1({n}, {base})".format(n=first, base=second)
+        return "FMOD_PYC3({n}, {base})".format(n=first, base=second)
 
     def _print_PyccelPow(self, expr):
         b = expr.args[0]
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -631,7 +631,7 @@ class CCodePrinter(CodePrinter):
             first = self._print(NumpyFloat(expr.args[0]))
         if expr.args[1].dtype is NativeInteger():
             second = self._print(NumpyFloat(expr.args[1]))
-        return "fmod({}, {})".format(first, second)
+        return "(fmod(((fmod({n}, {base})) + {base}), {base}))".format(n=first, base=second)
 
     def _print_PyccelPow(self, expr):
         b = expr.args[0]
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -41,6 +41,7 @@ from pyccel.ast.utilities import expand_to_loops
 from pyccel.ast.variable import ValuedVariable
 from pyccel.ast.variable import PyccelArraySize, Variable, VariableAddress
 from pyccel.ast.variable import DottedName
+from pyccel.ast.variable import InhomogeneousTupleVariable
 
 from pyccel.ast.sympy_helper import pyccel_to_sympy
 
@@ -836,6 +837,9 @@ class CCodePrinter(CodePrinter):
         return '{}(*{})({});\n'.format(ret_type, name, arg_code)
 
     def _print_Declare(self, expr):
+        if isinstance(expr.variable, InhomogeneousTupleVariable):
+            return ''.join(self._print_Declare(Declare(v.dtype,v,intent=expr.intent, static=expr.static)) for v in expr.variable)
+
         declaration_type = self.get_declare_type(expr.variable)
         variable = self._print(expr.variable.name)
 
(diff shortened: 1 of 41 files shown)


nhamidn (approved):

tests/epyccel/test_epyccel_mod.py
```diff
@@ -11,7 +11,7 @@ def modulo_i_i(x, y):
         return x % y
 
     f = epyccel(modulo_i_i, language=language)
-    x = randint(1e6)
+    x = randint(-1e6, 1e6)
     y = randint(low=1, high=100) # low=1 for avoid zero-division error
```
EmilyBourne, line 15: Could we also add a test with a negative `y` please. Beware, `y` must never be `0`!

tests/epyccel/test_epyccel_mod.py
```diff
@@ -23,7 +23,7 @@ def modulo_r_r(x, y):
         return x % y
 
     f = epyccel(modulo_r_r, language=language)
-    x = uniform(high=1e6)
+    x = uniform(low=-1e6, high=1e6)
     y = uniform(low=1, high=1e2) # low=1 for avoid zero-division error
```
EmilyBourne, line 27: See above, test negative `y` too

tests/epyccel/test_epyccel_mod.py
```diff
@@ -35,7 +35,7 @@ def modulo_r_i(x, y):
         return x % y
 
     f = epyccel(modulo_r_i, language=language)
-    x = uniform(high=1e6)
+    x = uniform(low=-1e6, high=1e6)
     y = randint(low=1, high=100) # low=1 for avoid zero-division error
```
EmilyBourne, line 39: See above, test negative `y` too

tests/epyccel/test_epyccel_mod.py
```diff
@@ -47,7 +47,7 @@ def modulo_i_r(x, y):
         return x % y
 
     f = epyccel(modulo_i_r, language=language)
-    x = randint(1e6)
+    x = randint(-1e6, 1e6)
     y = uniform(low=1, high=1e2) # low=1 for avoid zero-division error
```
EmilyBourne, line 51: See above, test negative `y` too

tests/epyccel/test_epyccel_mod.py
```diff
@@ -23,7 +23,7 @@ def modulo_r_r(x, y):
         return x % y
 
     f = epyccel(modulo_r_r, language=language)
-    x = uniform(high=1e6)
+    x = uniform(low=-1e6, high=1e6)
```
EmilyBourne, line 26: I think it would be better to have 2 tests here. One with a positive `x` and one with a negative `x`. If we had this test before your fix then the test would fail some of the time, but not all the time which makes it much harder to identify the problem

tests/epyccel/test_epyccel_mod.py
```diff
@@ -35,7 +35,7 @@ def modulo_r_i(x, y):
         return x % y
 
     f = epyccel(modulo_r_i, language=language)
-    x = uniform(high=1e6)
+    x = uniform(low=-1e6, high=1e6)
```
EmilyBourne, line 38: See above, positive and negative test (and maybe a test with x==0) please

tests/epyccel/test_epyccel_mod.py
```diff
@@ -47,7 +47,7 @@ def modulo_i_r(x, y):
         return x % y
 
     f = epyccel(modulo_i_r, language=language)
-    x = randint(1e6)
+    x = randint(-1e6, 1e6)
```
EmilyBourne, line 50: See above

EmilyBourne (changes requested):
Please add a description to your PR

pyccel/codegen/printing/ccode.py
```diff
@@ -625,13 +625,13 @@ def _print_PyccelMod(self, expr):
...
         if expr.dtype is NativeInteger():
-            return "{} % {}".format(first, second)
+            return "((({n} % {base}) + {base}) % {base})".format(n=first, base=second)
 
         if expr.args[0].dtype is NativeInteger():
             first = self._print(NumpyFloat(expr.args[0]))
         if expr.args[1].dtype is NativeInteger():
             second = self._print(NumpyFloat(expr.args[1]))
-        return "fmod({}, {})".format(first, second)
+        return "(fmod(((fmod({n}, {base})) + {base}), {base}))".format(n=first, base=second)
```
EmilyBourne, line 634: Could we use the basic expression (e.g. `a % b`) if the second argument is a `Literal`?
e.g.:
```
a % b
a % 2
a % -2
```
would give:
```
(((a % b) + b) % b)
a % 2
(a % 2) + 2
```

EmilyBourne, line 634: Why do you need brackets around the outermost expression?

tests/epyccel/test_epyccel_mod.py
```diff
@@ -11,38 +11,68 @@ def modulo_i_i(x, y):
...
+    nx = uniform(low=-1e6, high=-1)
     y = randint(low=-100, high=100)
     y = 1 if y == 0 else y
 
-    assert(isclose(f(x, y), modulo_r_i(x, y), rtol=1e-15, atol=1e-15))
-    assert isinstance(f(x, y), type(modulo_r_i(x, y)))
+
+    f_output = f(x, y)
+    modulo_r_i_output = modulo_r_i(x, y)
+    assert(isclose(f_output, modulo_r_i_output, rtol=1e-15, atol=1e-15))
```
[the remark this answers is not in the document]
EmilyBourne: This doesn't seem to have fixed the problem as we can see on linux. The error message is:
```
>       assert(isclose(f(x, y, z), modulo_multiple(x, y, z), rtol=RTOL, atol=ATOL))
E       assert False
E        +  where False = isclose(8.964345582700844, 8.964345582702663, rtol=1e-14, atol=1e-15)
E        +    where 8.964345582700844 = <built-in function modulo_multiple>(309160, 9399.40736420054, 93)
E        +    and   8.964345582702663 = <function test_modulo_multiple.<locals>.modulo_multiple at 0x7fcc64c07c80>(309160, 9399.40736420054, 93
```
A quick test shows that for these values to be considered "close" the tolerances would need to be:
- rtol=1e-12
- atol=1e-13

This is quite a high tolerance, especially for a basic operator so I don't really want to use this tolerance if it can be avoided.

Maybe we could try an IfTernaryOperator?
E.g. `x % y` could be translated to `x < 0 ? x % y + y : x % y`

But I am not sure to what extent this would slow down the code or impact it's readability.

I already think that the translation of:
```
return x % y % z
```
to:
```
return (fmod(((fmod((fmod(((fmod((double)(x), y)) + y), y)), (double)(z))) + (double)(z)), (double)(z)));
```
gives code which is much harder to understand. For 1 modulo operator it seems ok, but as soon as there is more happening in the line it quickly becomes hard to read.

Imagine if we make things worse by making `y` an indexed element, e.g.:
```
return x % y[0] % z
```
now we generate
```
return (fmod(((fmod((fmod(((fmod((double)(x), y.nd_double[get_index(y, 0)])) + y.nd_double[get_index(y, 0)]), y.nd_double[get_index(y, 0)])), (double)(z))) + (double)(z)), (double)(z)));
```
personally I cannot understand this line.
I suspect we ought to make the modulo operator a function in our internal library in order to avoid this kind of obfuscation. Maybe we can discuss this in the scrum today. I would like to get a second opinion

pyccel/codegen/printing/ccode.py
EmilyBourne: I.e. why:
```
(fmod(((fmod(x, y)) + y), y))
```
and not:
```
fmod(fmod(x, y) + y, y)
```

tests/epyccel/test_epyccel_mod.py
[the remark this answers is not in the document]
EmilyBourne: > Maybe we could try an `IfTernaryOperator`?
> E.g. `x % y` could be translated to `x < 0 ? x % y + y : x % y`
> 
> But I am not sure to what extent this would slow down the code or impact it's readability.

As per [here](https://stackoverflow.com/a/28889216/5957195), this should not slow down the code too much, however the condition for the `IfTernaryOperator` is more complicated than above:
`((x < 0 and y > 0) or (x > 0 and y < 0)) ? x % y + y : x % y`

EmilyBourne:
We discussed this PR during the scrum today. The result of the discussion was:
- The current implementation is not very readable for expressions such as `x % y[0] % z`. A function (or a macro) should therefore be created in the `math` internal library
- The error is quite large but it is not clear which result is more accurate, the one provided by python or the one provided by the c implementation.
    We should check this by using the mpmath library to carry out the calculation with a higher precision. We can start with the test case which is failing on linux :
        `309160 % 9399.40736420054 % 93`
    If the current implementation is more accurate than the python implementation then we should keep it and increase the tolerance in the tests.
    If the python implementation is more accurate then we should try the implementation with the `IfTernaryOperator` to see if this is more accurate
- It is better to avoid function calls where the call is more costly than the contents of the function. As operators are expected to be fast this may be the case here. A test should be carried out to see if there is a noticeable difference between using a function and using a macro. If there is then a macro should be used. It can be defined in `stdlib/math/pyc_math.h`

EmilyBourne:
A quick test with `mpmath` seems to indicate that python is more accurate:
```
In [1]: from mpmath import mp                                                                                                       

In [2]: mp.dps                                                                                                                      
Out[2]: 15

In [3]: mp.dps = 100                                                                                                                

In [4]: x = mp.mpmathify(309160)                                                                                                    

In [5]: y = mp.mpmathify(9399.40736420054)                                                                                          

In [6]: z = mp.mpmathify(93)                                                                                                        

In [7]: x % y % z                                                                                                                   
Out[7]: mpf('8.9643455827026627957820892333984375')
```
this should also be tested in cases with different signs

tests/epyccel/test_epyccel_mod.py
```diff
@@ -1,57 +1,111 @@
...
+    modulo_i_i_output = modulo_i_i(x, y)
+    assert modulo_i_i(x, y)== f(x, y)
+    assert modulo_i_i_output == f_output
+    assert isinstance(f_output, type(modulo_i_i_output))
+
+    #test negative x
+    f_output = f(nx, y)
+    modulo_i_i_output = modulo_i_i(nx, y)
+    assert modulo_i_i_output == f_output
+    assert isinstance(f_output, type(modulo_i_i_output))
```
EmilyBourne, line 40: To be certain that this works I think it would be best to have 4 tests:
```
modulo_i_i(x,y)
modulo_i_i(nx,y)
modulo_i_i(x,ny)
modulo_i_i(nx,ny)
```

tests/epyccel/test_epyccel_mod.py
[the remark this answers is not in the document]
EmilyBourne: Correction:
XOR can be more succinctly written as:
`((x > 0) != (y > 0)) ? x % y + y : x % y`

pyccel/stdlib/math/pyc_math.h
```diff
@@ -8,6 +8,14 @@
 
 #include <stdint.h>
 
+/*
+** (((N % M) + M) % M) and fmod(fmod(N, M) + M, M)) are used to handle the negative
+** operands of modulo operator.
+*/
+
+#define MOD_PYC(N, M) ((N < 0 || M < 0) ? (((N % M) + M) % M) : (N % M))
+#define FMOD_PYC(N, M) ((N < 0 || M < 0) ? fmod(fmod(N, M) + M, M) : fmod(N, M))
```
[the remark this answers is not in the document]
MostafaMamouni: > Shouldn't this be an XOR?

it can be both.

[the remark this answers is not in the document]
MostafaMamouni: > Are you sure the second modulo is needed? Or is it just there for the case N<0, M<0 ?

if you meant the second modulo in the whole macro, i think it's necessary, because the expression for negative operands is missing the precision for positive ones, and just a simple modulo in C will solve this problem.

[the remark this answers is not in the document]
EmilyBourne: I meant this modulo:
```
fmod(fmod(N, M) + M, M)
  ^
```
I think it is only needed to handle the case `N<0`, `M<0` when you don't use an XOR. As a call to `fmod` can be costly I think it would be better to use the XOR and only have one `fmod` call in each branch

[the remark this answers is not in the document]
MostafaMamouni: sorry, but i couldn’t follow in the XOR point, is the XOR will work if `N < 0` and `M < 0` ?.

MostafaMamouni:
@EmilyBourne thank you for the optimization.

EmilyBourne (approved):

pyccel/stdlib/math/pyc_math.h
```diff
@@ -9,7 +9,7 @@
 #include <stdint.h>
 
 /*
-** (((N % M) + M) % M) and fmod(fmod(N, M) + M, M)) are used to handle the negative
+** ((N % M) + M) and fmod(N, M) + M, M) are used to handle the negative
```
EmilyBourne, line 12: Extra `, M)`

yguclu (approved):
Good job @MostafaMamouni!

PR #867, opened 2021-04-22
Task

MenoIy:
**Approach**
Create new decorator _task_ that will allow _pyccel_  to recognize that the current function will be used in task parallelism and automatically calculate its dependencies.
_task_ decorator have two optional arguments  :
- **type**  : type of the task if its  'master'  that mean that this function is responsible for running other task functions, default 'child'
- **threads_num** : number of threads reserved for this task region

A task function can only run in parallel if the parent function is a task of type 'master'.
```Python
@task
def A():
    # do something

@task('master')
def main():
    a = A()
```
Block of task is identifies by the first occurrence of a task (currently : in function call / Assign) until the first non task element or the end of the program.
```Python
# Function  A and B are tasks

@task('master')
def main():
    x, y = # do something

    # this will run in parallel
    # ------
    a = A(x) # first task A
    b = B(y) # second task B
    # ------

    c = a + b # first non task block
```
Dependencies between class is calculated by collecting all the information of task function into a namespace list
that help to update dependencies(In, out) shared elements ( for openmp)  and precoders function (for other library)
as we go on semantic stage.


pyccel/ast/core.py
```diff
@@ -3821,6 +3821,64 @@ def __reduce_ex__(self, i):
         mod_name  = self.mod_name)
         return (apply, (self.__class__, (), kwargs))
 
+
+class TaskFunctionCall(FunctionCall):
+    pass
+
+class Task(Basic):
```
EmilyBourne, line 3828: Shouldn't these be in `ast/omp.py` ?

tutorial/task_presentation.md
```diff
@@ -0,0 +1,131 @@
+# task-parallelism with Pyccel
+
+## A task
+
+Parallel program is viewed as a collection of tasks that communicate by sending
```
EmilyBourne, line 5: Missing article `A parallel program`

tutorial/task_presentation.md
```diff
@@ -0,0 +1,131 @@
...
+messages to each other through dependencies.
+A task consists of an executable unit or a unit of computation (think of it as a program),
+together with its local memory and a collection of I/O ports. that can/should execute in
+parallel with other tasks.
+
+## Context
+
+A problem can be borken into multiple parts each part can be represented as a task, each
```
EmilyBourne, line 13: Typo: `borken`->`broken`

EmilyBourne, line 13: This sentence is missing some punctuation. I suggest:
```
A problem can be broken into multiple parts. Each part can be represented as a task. Each
```

tutorial/task_presentation.md
```diff
@@ -0,0 +1,131 @@
...
+task is a separate unit of work that can be independant or may take some dependencies from other
+tasks. task with dependencies may only start when all antecedents have completed.
```
EmilyBourne, line 15: Sentences must start with an uppercase letter: `tasks. Tasks with dependencies`

tutorial/task_presentation.md
```diff
@@ -0,0 +1,131 @@
...
+together with its local memory and a collection of I/O ports. that can/should execute in
+parallel with other tasks.
+
+## Context
+
+A problem can be borken into multiple parts each part can be represented as a task, each
+task is a separate unit of work that can be independant or may take some dependencies from other
+tasks. task with dependencies may only start when all antecedents have completed. 
+So in general the dependencies between tasks can be represented as a directed acyclic graph
+where tasks form the vertices edge are the dependencies between task.
```
EmilyBourne, line 17: Should this be:
```
So in general the dependencies between tasks can be represented as a directed acyclic graph
where the vertices are tasks and the edges are the dependencies between task.
```
?

tutorial/task_presentation.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Usage:
+Using the decorator task allows pyccel to recognize that the current function will
```
EmilyBourne, line 20: I think you should highlight the decorator name in some way (code snippet/italics/bold/etc)

tutorial/task_presentation.md
```diff
@@ -0,0 +1,131 @@
...
+@task('master')
+def recursive() -> int:
+    a = recursive()
+    # do something
+```
+
+
+A block of tasks that need to run in parallel is by the first occurrence of a task 
+function in (FunctionCall or Assign) until the first non task block of the end of the program.
+So avoid writing a codeblock that depend on a task result between tasks calls.
```
EmilyBourne, line 46: Maybe clearer:
```
The parallel block generated by pyccel will only contain function calls and assignments which save the result of a function call. As a result it is advisable to avoid writing a code block which depends on a task result between task calls; otherwise the parallel block will be split into multiple blocks
```

tutorial/task_presentation.md
```diff
@@ -0,0 +1,131 @@
...
+    c = C(a + b)
+    d = D(a)
+```
+
+dependencies graph:
+A---- ------D
+     |------C
+B----
+
+in this scenario A, B will run in parallel, C will wait for then to complets, and D will only wait for A to complete.
```
EmilyBourne, line 105: Typo `complets`->`complete`

tutorial/task_presentation.md
```diff
@@ -0,0 +1,131 @@
...
+    d = D(a)
+    e = E(1)
+```
+
+dependencies graph:
+A -----------> D
+E
+
+
+in this scenario D will wait for A to complet, but E is independant so it can run in parallel with A or D (depend on available threads)
```
EmilyBourne, line 121: Typo `complet`->`complete`
Typo `independant`->`independent`
Grammar `depend`->`depending`

tutorial/task_presentation.md
```diff
@@ -0,0 +1,131 @@
...
+
+
+Currently complex manipulations of task are not completly supported (in automatic generation) yet like :
```
EmilyBourne, line 124: Typo `completly`->`completely`
English improvement:
`Currently complex manipulations of tasks are not completely supported (in automatic generation) yet. For example the following examples are not supported :`

tutorial/task_presentation.md
```diff
@@ -0,0 +1,131 @@
...
+- task as if condition                       ```if(x > A(2))```
+- task as a for loop condition               ```for i in range(A(2))```
+- task as function or another task argument  ```A(B(1))```
+- task as index                              ```arr[A(1)]```
+- task in operations                         ```a = A(1) + 2```
+
+so the best practice will be to avoid any undefined behavior by storing the task result in a variable and us it freely.
```
EmilyBourne, line 131: Typo `us`->`use`
English improvement:
`The best practice in order to avoid any undefined behavior is to store the task result in a variable and then use it freely.`

PR #872, opened 2021-04-23
Fix Python printer and test it

### LINKED ISSUE #643: Improve Python Printer (pycode)

### LINKED ISSUE #868: Wrong generated code in Python when using print function
**Describe the bug**
The Python printer generate a code with an invalid syntax, when using the print function

**To Reproduce**
```python
def func():
    print("test")

    a = 0
```

Generated to this:
```python

def func():
    print('test')a = 0
```
And

This code
```python
def func():
    for i in range(7):
        print("test")

    a = 0
```

Generated to this:
```python

def func():
    for i in range(0, 7, 1):
        print('test'
    a = 0
```

Per my testing it does not appear to use the `_print_Print` method in `pycode.py`

**Language**
Python

nhamidn:
Finish improving PythonPrinter (fixes #643) so it is at a similar level to the fortran and c printers. Fixing `PythonPrint` (fixes #868) allows the pyccel tests to be activated to prove this

**Commit Summary**
- Only strip `\n` from code (avoids bugs like #868 occurring in the future)
- Add missing `\n` to `_print_PythonPrint` (fixes #868)
- Add OMP printers to python printer
- Add reason in OMP tests
- Activate "pyccel" tests for python (fixes #643 )


tests/pyccel/test_pyccel.py
```diff
@@ -27,17 +28,36 @@ def get_abs_path(relative_path):
...
-        exefile = exefile + ".exe"
-    if not os.path.isfile(exefile):
-        dirname = os.path.dirname(filename)
-        basename = "prog_"+os.path.basename(filename)
-        exefile = os.path.join(dirname, os.path.splitext(basename)[0])
+
+    if language == 'python':
+        result = ''
+        if sys.platform == "win32":
+            result = exefile.split('\\')
```
EmilyBourne, line 37: You can use `os.path.basename(filename)` to do this in a system independent way (see original code)

tests/pyccel/test_pyccel.py
```diff
@@ -27,17 +28,36 @@ def get_abs_path(relative_path):
...
+                result[-1] = "prog_"+result[-1]
+            exefile = "\\".join(result)
+        else:
+            result = exefile.split('/')
+            result.insert(-1, "py")
+            if prog is True:
+                result[-1] = "prog_"+result[-1]
+            exefile = "/".join(result)
+        exefile += '.py'
+        exefile = os.path.normpath(exefile)
```
EmilyBourne, line 49: This if/else is not necessary if you use functions from the `os.path` module as is done in the non-python case

tests/pyccel/test_pyccel.py
```diff
@@ -229,10 +253,13 @@ def pyccel_test(test_file, dependencies = None, compile_with_pyccel = True,
             dependencies = []
         if language=='fortran':
             compile_fortran(cwd, test_file, dependencies)
-        else:
+        elif language == 'c':
             compile_c(cwd, test_file, dependencies)
 
-    lang_out = get_lang_output(get_exe(test_file))
+    if language == "python":
+        lang_out = get_python_output(get_exe(test_file, language, prog=prog))
```
EmilyBourne, line 260: I'm not sure this line makes sense if `prog=False`? If it is not a program then nothing will execute so there will be no output to collect, or am I missing something?

tests/pyccel/test_pyccel.py
```diff
@@ -297,6 +324,13 @@ def test_imports_compile(language):
 #------------------------------------------------------------------------------
 @pytest.mark.xdist_incompatible
 def test_imports_in_folder(language):
+    if language == 'python':
+        current_folder = os.path.abspath(__file__)
+        result = current_folder.split('/')
+        result = result[:-1]
+        current_folder = "/".join(result)
+        if os.path.exists(current_folder+"/scripts/py/folder1") is False:
+            shutil.copytree(current_folder+"/scripts/folder1", current_folder+"/scripts/py/folder1")
```
EmilyBourne, line 333: I am not sure this will work on windows. Please use functions from the `os.path` module

EmilyBourne, line 333: Why do you need to copy the files?

tests/pyccel/test_pyccel.py
```diff
@@ -375,8 +409,12 @@ def test_expressions(language):
...
             [float]*3 + [int] + [float]*2 + [int] + [float]*3 + [int] + \
             [float]*3 + [int]*2 + [float]*2 + [int]*5 + [complex] + [bool]*9
-    pyccel_test("scripts/expressions.py", language=language,
-                output_dtype = types)
+    if language == "python":
+        pyccel_test("scripts/expressions.py", language=language,
+            output_dtype = str)
+    else:
+        pyccel_test("scripts/expressions.py", language=language,
+            output_dtype = types)
```
EmilyBourne, line 417: Can you explain why these have different output_dtypes?

tests/pyccel/test_pyccel.py
```diff
@@ -550,15 +594,21 @@ def test_c_arrays(language):
...
-    pyccel_test("scripts/arrays_view.py", language=language, output_dtype=types)
+    pyccel_test("scripts/arrays_view.py", language=language, output_dtype=types, prog=True)
 
 #------------------------------------------------------------------------------
 def test_headers(language):
+    if language == "python":
+        base_dir = os.path.dirname(os.path.realpath(__file__))
+        full_path = os.path.join(base_dir, os.path.normpath("scripts/py"))
+        if os.path.exists(full_path) is False:
+            os.mkdir(full_path)
```
EmilyBourne, line 605: I think this directory will be created by the `pyccel` command

tests/pyccel/test_pyccel.py
```diff
@@ -585,9 +635,11 @@ def test_headers(language):
 
     compile_pyccel(cwd, test_file, pyccel_commands)
 
-
-    lang_out = get_lang_output(get_exe(test_file))
-    assert int(lang_out)
+    if language == "python":
+        lang_out = get_python_output(get_exe(test_file, language, prog=True))
+    else:
+        lang_out = get_lang_output(get_exe(test_file, language))
```
EmilyBourne, line 641: We seem to have this output a lot. I wonder if we should combine these functions into a `get_output` function which receives the language as an argument. What do you think?

tests/pyccel/test_pyccel.py
nhamidn: i set the value of prog when its a program like in the line 639 which is a program

tests/pyccel/test_pyccel.py
nhamidn: because the generated python code is in the folder `/scripts/py` and the import are in the `scripts` folder I copy them to make the code work

tests/pyccel/test_pyccel.py
EmilyBourne: This explains why you need `prog=True`, but not why you need `prog=False` if `prog` is always true (this seems to be the assumption made for fortran/c) then this argument adds nothing

tests/pyccel/test_pyccel.py
EmilyBourne: Ah is this a special case for examples with multiple folders? Have you run a test without pytest to be sure that you are testing the generated code and not the copied code?

tests/pyccel/test_pyccel.py
```diff
@@ -12,8 +12,8 @@
 #==============================================================================
 
 @pytest.fixture( params=[
-        #pytest.param("fortran", marks = pytest.mark.fortran),
-        #pytest.param("c", marks = pytest.mark.c),
+        pytest.param("fortran", marks = pytest.mark.fortran),
+        pytest.param("c", marks = pytest.mark.c),
```
EmilyBourne, line 16: Just guessing, but if you want to test just the python code on the runner it is better to comment this:
- https://github.com/pyccel/pyccel/blob/master/.github/workflows/Github_pytest.yml#L24-L25
- https://github.com/pyccel/pyccel/blob/master/.github/workflows/Github_pytest.yml#L63-L80
- https://github.com/pyccel/pyccel/blob/master/.github/workflows/Github_pytest.yml#L124-L125

tests/pyccel/test_pyccel.py
nhamidn: i think this test not working well, I'm gonna try to fix it

tests/pyccel/test_pyccel.py
nhamidn: i see, no need for the prog argument, I removed it.

tests/pyccel/test_pyccel.py
nhamidn: this one is fixed

EmilyBourne (approved):

yguclu (approved):

PR #877, opened 2021-04-26
Raise an error if an argument with the wrong dtype/precision/rank is passed to a function

### LINKED ISSUE #869: A function called with arguments of incorrect types does not report an error
**Describe the bug**
A function called with arguments of incorrect types does not report an error. When translating to fortran this leads to a compiler error. When translating to C the value will be cast to the correct type which may lead to unexpected results

**To Reproduce**
```python
def f(x : int):
    y = x + 1
    return y

z = f(2.3)
print(z)
```

In fortran this code results in:
```bash/fortran/c
  File "/home/emily/.local/bin/pyccel", line 33, in <module>
  File "/home/emily/.local/lib/python3.8/site-packages/pyccel/commands/console.py", line 240, in pyccel
    execute_pyccel(filename,
  File "/home/emily/.local/lib/python3.8/site-packages/pyccel/codegen/pipeline.py", line 423, in execute_pyccel
    compile_files(fname, f90exec, flags,
  File "/home/emily/.local/lib/python3.8/site-packages/pyccel/codegen/utilities.py", line 150, in compile_files
    output = subprocess.check_output(cmd, stderr=subprocess.STDOUT, shell=True)
  File "/usr/lib/python3.8/subprocess.py", line 411, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib/python3.8/subprocess.py", line 512, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command 'gfortran  -O3 -fPIC  -I"/home/emily/Code/test/tmp/__pyccel__" -I"/home/emily/.local/lib/python3.8/site-packages/numpy/core/include" -I"/home/emily/Code/test/tmp/__pyccel__" /home/emily/Code/test/tmp/__pyccel__/tmp.o "/home/emily/Code/test/tmp/__pyccel__/prog_tmp.f90" -o "prog_tmp"  ' returned non-zero exit status 1.
```
The compiler error is:
```
/home/emily/Code/test/tmp/__pyccel__/prog_tmp.f90:10:4:

   10 | z = f(2.3_C_DOUBLE)
      |    1
Error: Type mismatch in argument ‘x’ at (1); passed REAL(8) to INTEGER(8)
```

In C the program compiles and the output is:
```
3
```

**Expected behavior**
An error (`severity='error'`) should be raised in the semantic stage to force the user to either use interfaces or cast explicitly

EmilyBourne:
Raise an error if the argument is incompatible due to a mismatch in dtype/precision/rank (fixes #869)

**Commit Summary**
- Raise an error if the argument is incompatible due to a mismatch in dtype/precision/rank (fixes #869)
- Add a test which should fail and raise the error
- [BUGFIX] Correct shape/rank of `MacroShape`
- [BUGFIX] Ensure stage is set to 'syntactic' when parsing headers (important for tests/parser/test_headers.py)
- [BUGFIX] Pass rhs of expected rank to LaPack routines

### DIFF
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -3230,9 +3230,7 @@ class SemanticParser(BasicParser):
         return macro
 
     def _visit_MacroShape(self, expr, **settings):
-        arg = self._visit(expr.argument)
-        index = self._visit(expr.index)
-        return MacroShape(arg, index)
+        return expr
 
     def _visit_MacroVariable(self, expr, **settings):
 
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -925,6 +925,65 @@ class SemanticParser(BasicParser):
                 args.append(a)
         return args
 
+    def get_type_description(self, var, include_rank = True):
+        """
+        Provides a text description of the type of a variable
+        (useful for error messages)
+        Parameters
+        ----------
+        var          : Variable
+                       The variable to describe
+        include_rank : bool
+                       Indicates whether rank information should be included
+                       Default : True
+        """
+        descr = '{dtype}(kind={precision})'.format(
+                        dtype     = var.dtype,
+                        precision = var.precision)
+        if include_rank and var.rank>0:
+            descr += '[{}]'.format(','.join(':'*var.rank))
+        return descr
+
+    def _check_argument_compatibility(self, input_args, func_args, expr, elemental):
+        """
+        Check that the provided arguments match the expected types
+
+        Parameters
+        ----------
+        input_args : list
+                     The arguments provided to the function
+        func_args  : list
+                     The arguments expected by the function
+        expr       : PyccelAstNode
+                     The expression where this call is found (used for error output)
+        elemental  : bool
+                     Indicates if the function is elemental
+        """
+        if elemental:
+            incompatible = lambda i_arg, f_arg: \
+                        (i_arg.dtype is not f_arg.dtype or \
+                        i_arg.precision != f_arg.precision)
+        else:
+            incompatible = lambda i_arg, f_arg: \
+                        (i_arg.dtype is not f_arg.dtype or \
+                        i_arg.precision != f_arg.precision or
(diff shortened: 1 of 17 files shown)


pyccel/parser/semantic.py
```diff
@@ -964,6 +964,19 @@ def _handle_function(self, expr, func, args, **settings):
...
+                         input_arg.precision == func_arg.precision and \
+                        (input_arg.rank      == func_arg.rank     or func.is_elemental) \
+                        for input_arg, func_arg in zip(new_expr.args, func.arguments) \
+                            if input_arg is not Nil() and \
+                                not isinstance(func_arg, FunctionAddress) and \
+                                func_arg.dtype is not NativeGeneric())
+                if not matching_types:
+                    errors.report(INCOMPATIBLE_ARGUMENT,
+                            symbol = expr,
+                            severity='error')
```
MenoIy, line 979: Can't this be in a small function or precise witch argument is incompatible?

EmilyBourne: I agree that a function makes sense. Do you think it should be a member of `FunctionDef` or `SemanticParser`?

> or precise witch argument is incompatible?

The reason I did not do this was because I was worried that the code might be slow as this is checked at every function call. But I did not benchmark it. As long as there are not too many arguments/the function call does not appear too often, it should not really cause problems. I will improve this :+1:

MenoIy: did you try to use __handle_function_args_ loop?

EmilyBourne: The function `_handle_function_args` has no knowledge of the `FunctionDef` and it is also used for other objects such as `PyccelInternalFunction`. If I checked it in the for loop then I would need to add an `if isinstance(func, FunctionDef)` inside the for loop. This is something that should be avoided at all cost from a performance point of view

EmilyBourne: Fixed in a799243

MenoIy (approved):

yguclu (approved):

PR #879, opened 2021-04-26
Allow ndarrays to be optional in C

### LINKED ISSUE #757: Arrays cannot be optional in C
**Describe the bug**
Arrays are never printed as pointers in C. Therefore they cannot be optional

**To Reproduce**
```python
@types( 'int[:]', 'int[:]')
def f12(x, y = None):
    if y is None:
        for i in range(x.shape[0]):
            x[i] *= 2
    else :
        for i in range(x.shape[0]):
            x[i] = x[i] // y[i]
```

```c
void f12(t_ndarray x, t_ndarray y)
{
    int64_t i;
    if (y == NULL)
    {
        for (i = 0; i < x.shape[0]; i += 1)
        {
            x.nd_int64[get_index(x, i)] *= 2;
        }
    }
    else
    {
        for (i = 0; i < x.shape[0]; i += 1)
        {
            x.nd_int64[get_index(x, i)] = (int64_t)floor((double)(x.nd_int64[get_index(x, i)]) / (double)(y.nd_int64[get_index(y, i)]));
        }
    }
}
```

The generated code raises the error:
```
/home/emily/Code/test/tmp/__pyccel__/test.c: In function ‘f12’:
/home/emily/Code/test/tmp/__pyccel__/test.c:11:11: error: invalid operands to binary == (have ‘t_ndarray {aka struct s_ndarray}’ and ‘void *’)
     if (y == NULL)
           ^~
```

**Language**
C

EmilyBourne:
Allow ndarrays to be optional in C by correctly implementing them as pointers in this case. Fixes #757

**Commit Summary**
- Correct `_print_IndexedElement` to access pointer elements correctly
- Correct `get_declare_type` to add pointer flag if necessary to ndarrays.
- Activate tests (require wrapper to pass)

_Blocked by #766_


EmilyBourne:
Fixed in #957

PR #880, opened 2021-04-26
Do not use shell=True in subprocess calls

### LINKED ISSUE #516: Replace check_output with Popen

### LINKED ISSUE #517: Fix codacy subprocess call with shell=True errors

EmilyBourne:
Stop using `shell=True` in subprocess calls (fixes #517). This allows us to use `Popen` to collect compiler errors (fixes #516) and to remove duplication on the GitHub runners to handle Windows' shell preferences. As a result compiler errors are now displayed in place of a useless traceback.

**Commit Summary**

- Save flags in a list of strings
- Make `cmd` a list of strings
- Remove `shell=True` from subprocess calls (fixes #517)
- Use `Popen` instead of `check_output` and collect and display any error messages (fixes #516)
- Use GitHub Actions on the Windows runner. As a terminal is no longer spawned it does not matter how Python is launched


MenoIy (approved):

pyccel/codegen/utilities.py
```diff
@@ -118,40 +116,42 @@ def compile_files(filename, compiler, flags,
...
+        libs_flags = ['-l{}'.format(i) for i in libs]
 
     if sys.platform == "win32" and compiler == "mpif90":
         compiler = "gfortran"
-        filename += ' "{}"'.format(os.path.join(os.environ["MSMPI_LIB64"], 'libmsmpi.a'))
+        m_code.append(os.path.join(os.environ["MSMPI_LIB64"], 'libmsmpi.a'))
 
-    cmd = '{0} {1} {2} {3} {4} {5} {6} {7}'.format( \
-        compiler, flags, m_code, filename, o_code, binary, libs_flags, j_code)
+    cmd = [compiler] + flags + m_code + [filename, o_code, binary] + libs_flags + j_code
```
yguclu, line 139: I suspect that Python creates many intermediate `list` objects when processing this line. Shouldn't the following be preferable instead?
```python
cmd = [compiler, *flags, *m_code, filename, o_code, binary, *libs_flags, *j_code]
```

EmilyBourne: Fixed in 4c28592

yguclu (approved):

PR #883, opened 2021-04-29
Replace get_index with inline strides

### LINKED ISSUE #882: The `get_index` method slows down the arrays in C
The following function runs really slow in Pyccel with C compared to Pyccel with Fortran
```python
@types("real[:,:,:,:]", "real[:,:]", "real[:,:]")
def dot(mat, x, out):
    for i1 in range(0, 128, 1):
        for i2 in range(0, 128, 1):
            v = 0.0
            for k1 in range(0, 9, 1):
                for k2 in range(0, 9, 1):
                    v += mat[4 + i1,4 + i2,k1,k2]*x[i1 + k1,i2 + k2]
            out[4 + i1,4 + i2] = v
```

this behavior is due to the use of the method `get_index` in the C code
```c
void lo_dot(t_ndarray mat, t_ndarray x, t_ndarray out, int64_t n1, int64_t n2)
{
    int64_t i1;
    int64_t i2;
    double v;
    int64_t k1;
    int64_t k2;
    for (i1 = 0; i1 < 128; i1 += 1)
    {
        for (i2 = 0; i2 < 128; i2 += 1)
        {
            v = 0.0;
            for (k1 = 0; k1 < 9; k1 += 1)
            {
                for (k2 = 0; k2 < 9; k2 += 1)
                {
                    v += mat.nd_double[get_index(mat, 4 + i1, 4 + i2, k1, k2)] * x.nd_double[get_index(x, i1 + k1, i2 + k2)];
                }
            }
            out.nd_double[get_index(out, 4 + i1, 4 + i2)] = v;
        }
    }
}
```

The function `get_index` should therefore be replaced with an explicit calculation of the index. However this will make the code hard to read. A macro should therefore be used if possible

saniQQ:
Replace `get_index` C function with inline strides arithmetic to increase performance (fixes #882); use a macro to enhance readability.


EmilyBourne:
This PR seems to have slowed the tests by a minute or two (presumably due to compilation time), but it fixes issue #882 on my computer:
```
pyccel execution time with fortran : 0.0029075145721435547
pyccel execution time with c : 0.0025358200073242188
```

pyccel/stdlib/ndarrays/ndarrays.h
```diff
@@ -32,6 +32,34 @@ typedef struct  s_slice
...
+#define GET_INDEX_EXP10(t, arr,ndim, a, b, c, d, e, f, g, h, i, j) GET_INDEX_EXP9(t, arr,ndim - 1, a, b, c, d, e, f, g, h, i) + t(arr,ndim, j)
+#define GET_INDEX_EXP11(t, arr,ndim, a, b, c, d, e, f, g, h, i, j, k) GET_INDEX_EXP10(t, arr,ndim - 1, a, b, c, d, e, f, g, h, i, j) + t(arr,ndim, k)
+#define GET_INDEX_EXP12(t, arr,ndim, a, b, c, d, e, f, g, h, i, j, k, l) GET_INDEX_EXP11(t, arr,ndim - 1, a, b, c, d, e, f, g, h, i, j, k) + t(arr,ndim, l)
+#define GET_INDEX_EXP13(t, arr,ndim, a, b, c, d, e, f, g, h, i, j, k, l, m) GET_INDEX_EXP12(t, arr,ndim - 1, a, b, c, d, e, f, g, h, i, j, k, l) + t(arr,ndim, m)
+#define GET_INDEX_EXP14(t, arr,ndim, a, b, c, d, e, f, g, h, i, j, k, l, m, n) GET_INDEX_EXP13(t, arr,ndim - 1, a, b, c, d, e, f, g, h, i, j, k, l, m) + t(arr,ndim,n)
+#define GET_INDEX_EXP15(t, arr,ndim, a, b, c, d, e, f, g, h, i, j, k, l, m, n, o) GET_INDEX_EXP14(t, arr,ndim - 1, a, b, c, d, e, f, g, h, i, j, k, l, m, n) + t(arr,ndim,o)
+#define GET_INDEX_EXP16(t, arr,ndim, a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p) GET_INDEX_EXP15(t, arr,ndim - 1, a, b, c, d, e, f, g, h, i, j, k, l, m, n, o) + t(arr,ndim,p)
```
EmilyBourne, line 50: I wonder if this could be made clearer written in the opposite direction. Ie:
```
#define GET_INDEX_EXP1(t, arr,ndim, a) t(arr,ndim, a)
#define GET_INDEX_EXP2(t, arr,ndim, a, b) t(arr,ndim, a) + GET_INDEX_EXP1(t, arr,ndim - 1, b)
#define GET_INDEX_EXP3(t, arr,ndim, a, ...) t(arr,ndim, a) + GET_INDEX_EXP2(t, arr,ndim - 1, __VA_ARGS__)
...
#define GET_INDEX_EXP16(t, arr,ndim, a, ...) t(arr,ndim, a) + GET_INDEX_EXP15(t, arr,ndim - 1, __VA_ARGS__)
```

pyccel/codegen/printing/ccode.py
```diff
@@ -937,7 +937,8 @@ def _print_IndexedElement(self, expr):
             inds = [self._cast_to(i, NativeInteger(), 8).format(self._print(i)) for i in inds]
         else:
             raise NotImplementedError(expr)
-        return "%s.%s[get_index(%s, %s)]" % (base_name, dtype, base_name, ", ".join(inds))
+        return "GET_ELEMENT(%s, %s, %s)" % (base_name, dtype, ", ".join(inds))
```
EmilyBourne, line 940: I think we should raise an error here if `rank>15`

pyccel/stdlib/ndarrays/ndarrays.h
```diff
@@ -32,6 +32,34 @@ typedef struct  s_slice
...
+
+
+#define NUM_ARGS_H1(dummy, x16, x15, x14, x13, x12, x11, x10, x9, x8, x7, x6, x5, x4, x3, x2, x1, x0, ...) x0
+#define NUM_ARGS(...) NUM_ARGS_H1(dummy, __VA_ARGS__, 16, 15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 0)
+#define GET_INDEX_FUNC_H3(t, arr, ndim, ...) GET_INDEX_EXP##ndim(t, arr, ndim - 1, __VA_ARGS__)
+#define GET_INDEX_FUNC_H2(t, arr, ndim, ...) GET_INDEX_FUNC_H3(t, arr, ndim, __VA_ARGS__)
+#define GET_INDEX_FUNC(t, arr, ...) GET_INDEX_FUNC_H2(t, arr, NUM_ARGS(__VA_ARGS__), __VA_ARGS__)
+
+#define GET_INDEX(arr, ...) GET_INDEX_FUNC(INDEX, arr, __VA_ARGS__)
```
EmilyBourne, line 59: Couldn't this call `GET_INDEX_FUNC_H2` directly?

EmilyBourne: Or even `GET_INDEX_FUNC_H3`?

pyccel/stdlib/ndarrays/ndarrays.h
```diff
@@ -32,6 +32,34 @@ typedef struct  s_slice
...
+#define GET_INDEX_EXP13(t, arr,ndim, a, b, c, d, e, f, g, h, i, j, k, l, m) GET_INDEX_EXP12(t, arr,ndim - 1, a, b, c, d, e, f, g, h, i, j, k, l) + t(arr,ndim, m)
+#define GET_INDEX_EXP14(t, arr,ndim, a, b, c, d, e, f, g, h, i, j, k, l, m, n) GET_INDEX_EXP13(t, arr,ndim - 1, a, b, c, d, e, f, g, h, i, j, k, l, m) + t(arr,ndim,n)
+#define GET_INDEX_EXP15(t, arr,ndim, a, b, c, d, e, f, g, h, i, j, k, l, m, n, o) GET_INDEX_EXP14(t, arr,ndim - 1, a, b, c, d, e, f, g, h, i, j, k, l, m, n) + t(arr,ndim,o)
+#define GET_INDEX_EXP16(t, arr,ndim, a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p) GET_INDEX_EXP15(t, arr,ndim - 1, a, b, c, d, e, f, g, h, i, j, k, l, m, n, o) + t(arr,ndim,p)
+
+
+#define NUM_ARGS_H1(dummy, x16, x15, x14, x13, x12, x11, x10, x9, x8, x7, x6, x5, x4, x3, x2, x1, x0, ...) x0
+#define NUM_ARGS(...) NUM_ARGS_H1(dummy, __VA_ARGS__, 16, 15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 0)
+#define GET_INDEX_FUNC_H3(t, arr, ndim, ...) GET_INDEX_EXP##ndim(t, arr, ndim - 1, __VA_ARGS__)
+#define GET_INDEX_FUNC_H2(t, arr, ndim, ...) GET_INDEX_FUNC_H3(t, arr, ndim, __VA_ARGS__)
```
EmilyBourne, line 56: Does this macro (`GET_INDEX_FUNC_H2`) do anything?

pyccel/stdlib/ndarrays/ndarrays.h
saniQQ: I was comparing to the preprocessed code so it made sense to me to go from the index 0 to n.

pyccel/stdlib/ndarrays/ndarrays.h
saniQQ: yes, could compress it to that.

pyccel/stdlib/ndarrays/ndarrays.h
EmilyBourne: I think this should give the same code.
Imagine:
```
GET_INDEX_EXP3(t, arr,3, 4, 5, 6)
```
Before it gave us:
```
GET_INDEX_EXP2(t, arr, 2, 4, 5) + t(arr, ndim, 6) = GET_INDEX_EXP1(t, arr, 1, 4) + t(arr, ndim, 5) + t(arr, ndim, 6) = t(arr, ndim, 4) + t(arr, ndim, 5) + t(arr, ndim, 6)
```
With my modification it gives us:
```
t(arr, ndim, 4) + GET_INDEX_EXP2(t, arr, 2, 5, 6) = t(arr, ndim, 4) + t(arr, ndim, 5) + GET_INDEX_EXP1(t, arr, 1, 6) = t(arr, ndim, 4) + t(arr, ndim, 5) + t(arr, ndim, 6)
```

pyccel/stdlib/ndarrays/ndarrays.h
saniQQ: should be removed. :+1:

pyccel/stdlib/ndarrays/ndarrays.h
EmilyBourne: Ah! No sorry you're right the `ndim` argument will be wrong

EmilyBourne: But why do we need to pass this? Is it not known? Could we not do either:
```
#define GET_INDEX_EXP1(t, arr, a) t(arr,0, a)
#define GET_INDEX_EXP2(t, arr, a, b) GET_INDEX_EXP1(t, arr,a) + t(arr,1, b)
#define GET_INDEX_EXP3(t, arr, a, b, c) GET_INDEX_EXP2(t, arr,a, b) + t(arr,2, c)
...
#define GET_INDEX_EXP16(t, arr, a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p) GET_INDEX_EXP15(t, arr, a, b, c, d, e, f, g, h, i, j, k, l, m, n, o) + t(arr,15,p)
```
or
```
#define GET_INDEX_EXP1(t, arr,ndim, a) t(arr,ndim-1, a)
#define GET_INDEX_EXP2(t, arr,ndim, a, b) t(arr,ndim-2, a) + GET_INDEX_EXP1(t, arr,ndim, b)
#define GET_INDEX_EXP3(t, arr,ndim, a, ...) t(arr,ndim-3, a) + GET_INDEX_EXP2(t, arr,ndim, __VA_ARGS__)
...
#define GET_INDEX_EXP16(t, arr,ndim, a, ...) t(arr,ndim-16, a) + GET_INDEX_EXP15(t, arr,ndim, __VA_ARGS__)
```
?

pyccel/stdlib/ndarrays/ndarrays.h
saniQQ: i can't compress it directly to `GET_INDEX_FUNC_H3` because I need one macro call to expand the `NUM_ARGS` call

pyccel/stdlib/ndarrays/ndarrays.h
saniQQ: I think removing the `ndim` better so it is readable even after the preprocessing

pyccel/codegen/printing/ccode.py
saniQQ: would it make sense if I raised the error in `print_Allocate`?

pyccel/stdlib/ndarrays/ndarrays.h
EmilyBourne: Don't they both generate the same code?

pyccel/codegen/printing/ccode.py
EmilyBourne: That sounds like a better idea, then you're only warned once per array. But what happens if the array is an argument to a function?

pyccel/stdlib/ndarrays/ndarrays.h
EmilyBourne: I said `call` not `compress` so I just meant:
```
#define GET_INDEX_FUNC(t, arr, ndim, ...) GET_INDEX_EXP##ndim(t, arr, __VA_ARGS__)

#define GET_INDEX(arr, ...) GET_INDEX_FUNC(INDEX, arr, NUM_ARGS(__VA_ARGS__), __VA_ARGS__)
```
(more or less what you did)

pyccel/stdlib/ndarrays/ndarrays.h
saniQQ: its the same but when having the `ndim` then preprocessed code will still have [ndim - 1 - 1 - 1] but we can have the exact number.

pyccel/stdlib/ndarrays/ndarrays.h
EmilyBourne: Now it seems to be `GET_INDEX_FUNC` which doesn't do anything ^^

pyccel/stdlib/ndarrays/ndarrays.h
EmilyBourne: Ah ok, then yes your solution is better. It also seems to be compiling faster now so the tests aren't slower

pyccel/stdlib/ndarrays/ndarrays.h
saniQQ: I think it is still necessary the problem is with the `##` operator which concatenate the literal value of `ndim` if we don't have a step for only expanding the `NUM_ARGS` call then the MAcro is expanded to `GET_INDEX_EXPNUM_ARGS`.

EmilyBourne: Currently the code I see is:
```
#define GET_INDEX_FUNC_H2(t, arr, ndim, ...) GET_INDEX_EXP##ndim(t, arr, __VA_ARGS__)
#define GET_INDEX_FUNC(t, arr, ndim, ...) GET_INDEX_FUNC_H2(t, arr, ndim, __VA_ARGS__)
```
So the `##` operator (which I agree we need to keep) is in `GET_INDEX_FUNC_H2` not `GET_INDEX_FUNC`

EmilyBourne: To be clear I would replace those 2 lines with:
```
#define GET_INDEX_FUNC(t, arr, ndim, ...) GET_INDEX_EXP##ndim(t, arr, __VA_ARGS__)
```

saniQQ: if we do that then we are not trying to get the value of `NUM_ARGS` so it is not expanded and when we try to use the `##` he will concatenate the literal string which is `NUM_ARGS`

saniQQ: `#define GET_INDEX_FUNC(t, arr, ndim, ...) GET_INDEX_FUNC_H2(t, arr, ndim, __VA_ARGS__)`
in this line, we are using the LHS `ndim` which is `NUM_ARGS` in the RHS so it is expanded to a string that contains the number.
then we are using it in `#define GET_INDEX_FUNC_H2(t, arr, ndim, ...) GET_INDEX_EXP##ndim(t, arr, __VA_ARGS__)`
so now when we concatenate we have the actual number as a string and not `NUM_ARGS`

EmilyBourne: Ah ok sorry, I assumed the expansion was done in the call. I don't see the difference between passing it here and passing it earlier but a quick test shows you are right :+1:

EmilyBourne (approved):

pyccel/errors/messages.py
```diff
@@ -27,6 +27,7 @@
 INCOMPATIBLE_ARGUMENT = 'Incompatible argument {} passed to function (expected {}). Please cast the argument explicitly or overload the function (see https://github.com/pyccel/pyccel/blob/master/tutorial/headers.md for details)'
 
 UNSUPPORTED_ARRAY_RETURN_VALUE = 'Array return arguments are currently not supported'
+UNSUPPORTED_ARRAY_RANK = 'Arrays of dimensions > 15 are currently not supported'
```
yguclu, line 30: I thought your implementation worked up to 16...

EmilyBourne:
> This PR seems to have slowed the tests by a minute or two (presumably due to compilation time), but it fixes issue #882 on my computer:
> 
> ```
> pyccel execution time with fortran : 0.0029075145721435547
> pyccel execution time with c : 0.0025358200073242188
> ```

After changes to the macro the difference in compile time seems to have disappeared but the new benchmarks remain the same

pyccel/errors/messages.py
yguclu: @saniQQ It appears that you have written also a macro for 16 array dimensions:

```python
#define GET_INDEX_EXP16(t, arr, a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p) GET_INDEX_EXP15(t, arr, a, b, c, d, e, f, g, h, i, j, k, l, m, n, o) + t(arr, 15, p)
```

So why raising an error if the user wants to use 16 indices?

saniQQ: I will remove that, i did not notice it.

saniQQ: Is it better to stop at 15 or have more dimensions?

EmilyBourne: As per our discussion the other day Fortran2008 says that Fortran is limited to 15 dimensions so this seems like a sensible place to stop

pyccel/stdlib/ndarrays/ndarrays.h
```diff
@@ -32,6 +32,31 @@ typedef struct  s_slice
...
+#define GET_INDEX_EXP10(t, arr, a, b, c, d, e, f, g, h, i, j) GET_INDEX_EXP9(t, arr, a, b, c, d, e, f, g, h, i) + t(arr, 9, j)
+#define GET_INDEX_EXP11(t, arr, a, b, c, d, e, f, g, h, i, j, k) GET_INDEX_EXP10(t, arr, a, b, c, d, e, f, g, h, i, j) + t(arr, 10, k)
+#define GET_INDEX_EXP12(t, arr, a, b, c, d, e, f, g, h, i, j, k, l) GET_INDEX_EXP11(t, arr, a, b, c, d, e, f, g, h, i, j, k) + t(arr, 11, l)
+#define GET_INDEX_EXP13(t, arr, a, b, c, d, e, f, g, h, i, j, k, l, m) GET_INDEX_EXP12(t, arr, a, b, c, d, e, f, g, h, i, j, k, l) + t(arr, 12, m)
+#define GET_INDEX_EXP14(t, arr, a, b, c, d, e, f, g, h, i, j, k, l, m, n) GET_INDEX_EXP13(t, arr, a, b, c, d, e, f, g, h, i, j, k, l, m) + t(arr, 13, n)
+#define GET_INDEX_EXP15(t, arr, a, b, c, d, e, f, g, h, i, j, k, l, m, n, o) GET_INDEX_EXP14(t, arr, a, b, c, d, e, f, g, h, i, j, k, l, m, n) + t(arr, 14, o)
+
+#define NUM_ARGS_H1(dummy, x16, x15, x14, x13, x12, x11, x10, x9, x8, x7, x6, x5, x4, x3, x2, x1, x0, ...) x0
+#define NUM_ARGS(...) NUM_ARGS_H1(dummy, __VA_ARGS__, 16, 15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 0)
```
EmilyBourne, line 52: Should the 16 not also be removed from these 2 macros?

yguclu (approved):

PR #884, opened 2021-04-29
Treat extra spaces in flag options

### LINKED ISSUE #881: Extra spaces in flags option causes failure
**Describe the bug**
If there is more than 1 consecutive flag in the string passed to the `fflags` option then an error is raised because pyccel tries to pass an empty string to gfortran

**To Reproduce**
For any file with the command:
```python
pyccel file.py --flags "-O3 -march=native -mtune=native  -mavx -ffast-math"
```

The error message is produced:
```bash
  File "/home/said/.local/bin/pyccel", line 33, in <module>
  File "/home/said/phd/pyccel/pyccel/commands/console.py", line 255, in pyccel
    folder        = args.output)
  File "/home/said/phd/pyccel/pyccel/codegen/pipeline.py", line 435, in execute_pyccel
    language=language)
  File "/home/said/phd/pyccel/pyccel/codegen/utilities.py", line 152, in compile_files
    raise RuntimeError(err_msg)
gfortran: error: : No such file or directory
```

**Expected behavior**
The `split` should avoid creating empty stringse.

EmilyBourne:
Handle extra spaces in the `fflags` argument passed to `execute_pyccel`. Fixes #881. Add an unnecessary space to a function using `fflags` to act as a test.

### DIFF
--- a/pyccel/codegen/pipeline.py
+++ b/pyccel/codegen/pipeline.py
@@ -220,7 +220,7 @@ def execute_pyccel(fname, *,
                                  accelerator=accelerator,
                                  includes=())
     elif fflags is not None:
-        fflags = fflags.strip().split(' ')
+        fflags = [f for f in fflags.strip().split(' ') if f]
     else:
         fflags = [] # Used for python
 
--- a/pyccel/codegen/pipeline.py
+++ b/pyccel/codegen/pipeline.py
@@ -220,7 +220,7 @@ def execute_pyccel(fname, *,
                                  accelerator=accelerator,
                                  includes=())
     elif fflags is not None:
-        fflags = [f for f in fflags.strip().split(' ') if f]
+        fflags = fflags.split()
     else:
         fflags = [] # Used for python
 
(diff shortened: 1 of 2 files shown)


saidctb (approved):

pyccel/codegen/pipeline.py
```diff
@@ -220,7 +220,7 @@ def handle_error(stage):
                                  accelerator=accelerator,
                                  includes=())
     elif fflags is not None:
-        fflags = fflags.strip().split(' ')
+        fflags = [f for f in fflags.strip().split(' ') if f]
```
yguclu, line 223: Isn't it sufficient to use `split` without any argument? That is
```python
fflags = fflags.split()
```
According to the [documentation](https://docs.python.org/3/library/stdtypes.html#str.split)
> If sep is not specified or is None, a different splitting algorithm is applied: runs of consecutive whitespace are regarded as a single separator, and the result will contain no empty strings at the start or end if the string has leading or trailing whitespace

yguclu (changes requested):

yguclu (approved):

pyccel/codegen/pipeline.py
yguclu: Done in cd131e5

Issue #541, opened 2020-11-13
Link libraries with specified compiler


EmilyBourne:
### Description
As a USER I would like my code to be linked with the compiler that was used to compile it in order to ensure that Interprocedural Optimisations (e.g inlining) can be carried out

### Current Behaviour
Currently when translating a python file to fortran the following 4 files are created:
- X.f90 (the direct translation)
- bind_c_X.f90 (the c compatible wrapper for X.f90
- X_wrapper.c (the c-python interface)
- setup_X.py

The steps for generating the X.so file are:

1. Compile X.f90 by hand
2. Compile bind_c_X.f90 by hand
3. Call setup_X.py which asks python to compile X_wrapper.c and link it with the object files

These steps are done with different compilers:
1. Compile X.f90 (user defined compiler, e.g. ifort)
2. Compile bind_c_X.f90 (user defined compiler, e.g. ifort)
3. Compile X_wrapper.c (python default compiler e.g. gcc)
3. Link files (python default compiler e.g. gcc)

In **f2py** the steps are also done with different compilers:
1. Compile X.f90 (user defined compiler, e.g. ifort)
2. Compile bind_c_X.f90 (user defined compiler, e.g. ifort)
3. Compile X_wrapper.c (python default compiler e.g. gcc)
3. Link files (user defined compiler, e.g. ifort)

### Desired behaviour
The steps should use compilers from the same vendor, e.g:
1. Compile X.f90 (user defined compiler, e.g. ifort)
2. Compile bind_c_X.f90 (user defined compiler, e.g. ifort)
3. Compile X_wrapper.c (c compiler matching user defined fortran compiler e.g. icc)
3. Link files (user defined compiler e.g. ifort)

In order to do this we must stop creating the `setup_X.py` file and compile by hand instead. This is difficult as we need to handle:
- **Flags** : Python uses a set of default flags which carry out some optimisations. These flags are platform/compiler/version dependant
- **Includes** : We need to include the python and numpy headers. Python headers can be found using `distutils.sysconfig.get_python_inc()`. Numpy headers can be found using `numpy.get_include()`
- **Libs**: A library is sometimes needed to compile the python files. This seems to be true notably on mac. The libraries used by python can be found using the command `python3-config --libs`, but this command can be problematic if more than one version of python is installed (e.g. I have python 3.6 and python 3.8, this command always returns libraries relevant for python3.6 on my computer)
[...]

PR #886, opened 2021-05-03
Wrapper part 1 (create C library for wrapper functions)

### LINKED ISSUE #758: Wrapper generates potentially dangerous warnings on macos

### LINKED ISSUE #806: Wrapper doesn't raise an error if C ordering is passed in place of F ordering
**Describe the bug**
The C/python API raises an error if an array with Fortran ordering is passed to a function which expects C ordering, but it doesn't raise an error if an array with C ordering is passed to a function which expects F ordering

**To Reproduce**
This raises an error:
```python

def print_arr(arr : 'int[:,:](order=C)'):
    n,m = arr.shape
    for i in range(n):
        print(arr[i,:])

import numpy as np
size = (2, 5)
arr = np.random.randint(1, 30, size=size, dtype=np.int64).T # F ordering
epy_func = epyccel(print_arr, language='fortran')
epy_func(arr)
print_arr(arr)
```
But this doesn't:
```python
rom pyccel.epyccel import epyccel

def print_arr(arr : 'int[:,:](order=F)'):
    n,m = arr.shape
    for i in range(n):
        print(arr[i,:])

import numpy as np
size = (2, 5)
arr = np.random.randint(1, 30, size=size, dtype=np.int64) # C ordering
epy_func = epyccel(print_arr, language='fortran')
epy_func(arr)
print_arr(arr)
```

It prints:
```
                   26                   27                    6                   12                    8
                   29                    3                   25                   26                   13
[26 29 27  3  6]
[25 12 26  8 13]
```

**Expected behavior**
The error should be raised

cc:  @mostafaMamoni  @MenoIy

MenoIy:
Add internal libraries to facilitate the definition of the wrapper file. 
Fixes #758 
Fixes #806 
This branch sometimes leads to a small slowdown, but it is not clear why that is the case. We suspect that the slowdown only occurs when translating simple files containing 1 function, but that a speedup should be seen when translating a module containing many functions

**Commit Summary**
- [CODACY] Remove unused imports
- Split `ast.cwrapper` into 2 modules. The new module `ast.numpy_wrapper` contains all the functions which interact with the numpy library
- Make the compiler dependencies an ordered dictionary
- Ensure that `.h` files include as few other `.h` files as possible (speeds up compiling)
- Add docstrings to `CWrapperCodePrinter`
- Create `cwrapper/cwrapper.c` and `cwrapper/cwrapper_ndarrays.c` to store all functions useful for converting between python and C
- Use functions to simplify `CWrapperCodePrinter`
- Put `numpy_version` into its own file to allow it to be imported by `cwrapper.c`, be adaptable to the user's system, and be defined somewhere where `NPY_1_19_API_VERSION` is defined
- Add `cwrapper/cwrapper.c` and `cwrapper/cwrapper_ndarrays.c` to the shared library creation step


pyccel/codegen/cwrapper.py
```diff
@@ -59,7 +60,7 @@ def create_c_setup(mod_name,
     else:
         code += "\n"
 
-    wrapper_file = "[ r'{0}' ]".format(wrapper_file)
+    wrapper_file = "[ r'{0}' , r'cwrapper/cwrapper.c']".format(wrapper_file) #TODO find another way
```
EmilyBourne, line 63: If this file is compiled every time then you need to add it to the file locking system [here](https://github.com/pyccel/pyccel/blob/master/pyccel/codegen/python_wrapper.py#L135-L143) to prevent 2 threads trying to use it at the same time (this is the cause of the current failures)

EmilyBourne:
Could you not use the same method to compile `cwrapper.c` as for `ndarrays.c`? I.e. add it to the [internal_libs](https://github.com/pyccel/pyccel/blob/master/pyccel/codegen/pipeline.py#L30-L33) and add it to the libs that need compiling [here](https://github.com/pyccel/pyccel/blob/master/pyccel/codegen/pipeline.py#L323), e.g.:
`if lib in [*codegen.get_printer_imports(), 'cwrapper']:`
?

pyccel/stdlib/cwrapper/cwrapper.c
```diff
@@ -0,0 +1,600 @@
...
+bool    PyIs_Int32(PyObject *o, bool hard_check)
+{
+	#ifdef _WIN32
+		return PyLong_Check(o) || PyArray_IsScalar(o, Int32);
+	#endif
+
+	if (hard_check == true)
+		return PyArray_IsScalar(o, Int32);
+
+	return PyLong_Check(o) || PyArray_IsScalar(o, Int32);
```
EmilyBourne, line 300: You need to use `#ifdef`, `#else`, `#endif` to avoid warnings about unreachable code. Test that pass the `-Werror` flag cause these warnings to make the test fail.

EmilyBourne:
My guess as to why this is so slow compared to the previous version is that now we are compiling all the helper functions for every test. This is 600 lines of code! The simplest way to fix this would be to only recompile if the file has changed. However, this means that we need to remove the compilation from the `setup_X.py` file. I don't know if this is currently possible. It may be related to issue #541

pyccel/stdlib/cwrapper/cwrapper.c
```diff
@@ -0,0 +1,600 @@
...
+	return object;
+}
+//--------------------------------------------------------//
+PyObject	*Float_to_PyDouble(float *d)
+{
+	PyObject	*object;
+
+	object = PyFloat_FromDouble((double)*d);
+
+	return object;
```
EmilyBourne, line 251: This could be written more compactly:
```
return PyFloat_FromDouble((double)*d);
```

EmilyBourne:
This seems to be related to compile time:
# Profiling results
**Master branch**
```
         17671769 function calls (17199104 primitive calls) in 209.511 seconds

   Ordered by: internal time
   List reduced from 8706 to 30 due to restriction <30>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     1291  168.221    0.130  168.221    0.130 {method 'poll' of 'select.poll' objects}
     1396    3.780    0.003    4.855    0.003 /home/emily/.local/lib/python3.8/site-packages/mpmath/identification.py:17(pslq)
   238565    1.776    0.000    1.776    0.000 {built-in method posix.stat}
     2022    1.735    0.001    1.735    0.001 {built-in method posix.read}
   784567    1.410    0.000    2.916    0.000 /usr/lib/python3.8/ast.py:214(iter_child_nodes)
   277944    1.327    0.000    2.097    0.000 /usr/lib/python3.8/posixpath.py:71(join)
2973616/2972971    1.295    0.000    1.314    0.000 {built-in method builtins.isinstance}
      376    1.171    0.003    1.171    0.003 {built-in method _posixsubprocess.fork_exec}
278624/277668    1.054    0.000    1.506    0.000 {built-in method builtins.next}
28980/276    1.044    0.000    6.275    0.023 /home/emily/.local/lib/python3.8/site-packages/pyccel/commands/pyccel_clean.py:15(pyccel_clean)
    29651    1.036    0.000    1.036    0.000 {built-in method posix.listdir}
      682    0.940    0.001    0.940    0.001 {built-in method builtins.compile}
132534/29902    0.871    0.000    2.905    0.000 /usr/lib/python3.8/os.py:280(walk)
  1012471    0.848    0.000    1.149    0.000 /usr/lib/python3.8/ast.py:202(iter_fields)
   340332    0.623    0.000    3.763    0.000 /usr/lib/python3.8/ast.py:325(walk)
       22    0.574    0.026    4.474    0.203 /home/emily/.local/lib/python3.8/site-packages/_pytest/_code/source.py:146(get_statement_startend2)
981899/981637    0.541    0.000    0.558    0.000 {built-in method builtins.getattr}
27411/26875    0.537    0.000    0.920    0.000 /home/emily/.local/lib/python3.8/site-packages/pyccel/ast/basic.py:34(__init__)
   340184    0.483    0.000    3.035    0.000 {method 'extend' of 'collections.deque' objects}
   228711    0.445    0.000    2.218    0.000 /usr/lib/python3.8/genericpath.py:27(isfile)
```

**This branch**
```
         16876188 function calls (16441834 primitive calls) in 308.112 seconds

   Ordered by: internal time
   List reduced from 8701 to 15 due to restriction <15>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     1470  267.856    0.182  267.856    0.182 {method 'poll' of 'select.poll' objects}
     1390    3.674    0.003    4.721    0.003 /home/emily/.local/lib/python3.8/site-packages/mpmath/identification.py:17(pslq)
     2208    1.777    0.001    1.777    0.001 {built-in method posix.read}
   238858    1.718    0.000    1.718    0.000 {built-in method posix.stat}
   782395    1.414    0.000    2.981    0.000 /usr/lib/python3.8/ast.py:214(iter_child_nodes)
   278770    1.308    0.000    2.061    0.000 /usr/lib/python3.8/posixpath.py:71(join)
      376    1.193    0.003    1.193    0.003 {built-in method _posixsubprocess.fork_exec}
2645131/2644487    1.170    0.000    1.188    0.000 {built-in method builtins.isinstance}
278663/277710    1.046    0.000    1.484    0.000 {built-in method builtins.next}
    29546    0.998    0.000    0.998    0.000 {built-in method posix.listdir}
28875/275    0.992    0.000    6.067    0.022 /home/emily/.local/lib/python3.8/site-packages/pyccel/commands/pyccel_clean.py:15(pyccel_clean)
      683    0.928    0.001    0.928    0.001 {built-in method builtins.compile}
  1009357    0.916    0.000    1.218    0.000 /usr/lib/python3.8/ast.py:202(iter_fields)
133818/30046    0.841    0.000    2.863    0.000 /usr/lib/python3.8/os.py:280(walk)
   339390    0.622    0.000    3.737    0.000 /usr/lib/python3.8/ast.py:325(walk)
```
All the difference comes from `{method 'poll' of 'select.poll' objects}`

# Compile Time
Looking in detail we see the following timings for the `communicate` method (used when waiting for the result of a `Popen` call [here](https://github.com/pyccel/pyccel/blob/master/pyccel/codegen/python_wrapper.py#L139-L140) and [here](https://github.com/pyccel/pyccel/blob/master/pyccel/codegen/utilities.py#L147-L148) ):
**Master branch**
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      376    0.012    0.000  168.521    0.448 /usr/lib/python3.8/subprocess.py:980(communicate)
```
**This branch**
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      376    0.013    0.000  268.152    0.713 /usr/lib/python3.8/subprocess.py:980(communicate)
```

The entire 100s difference can be explained by the compile time

# Proposed solution

Only compile `cwrapper.c` once. Ideally we would compile this file in a similar way to `ndarrays.c` and other internal libraries, but I suspect this may be complicated as the python include folders and libraries must be known. I suggest that the folder `cwrapper/` is only copied if `stdlib/cwrapper/cwrapper.c` is older than `__pyccel__/cwrapper/cwrapper.c` (see `os.path.getmtime`). Similarly we should pass either `cwrapper.c` or `cwrapper.o` to `setup_X.py` depending on the age of `cwrapper.o` compared to `cwrapper.c`. Care should be taken to ensure that the status of `cwrapper.o` is checked in a filelock to avoid deadlocks

pyccel/stdlib/cwrapper/cwrapper.c
EmilyBourne: Fixed in 7661749

pyccel/stdlib/cwrapper/cwrapper.c
```diff
@@ -0,0 +1,600 @@
...
+	// https://numpy.org/doc/stable/reference/c-api/array.html#c.PyArray_ScalarAsCtype
+	if (PyArray_IsScalar(object, Complex64))
+		PyArray_ScalarAsCtype(object, &c);
+
+	else
+	{
+		real_part = (float)PyComplex_RealAsDouble(object);
+		imag_part = (float)PyComplex_ImagAsDouble(object);
+
+		c = CMPLXF(real_part, imag_part);
```
EmilyBourne, line 53: Are you sure about this function? Windows doesn't seem to like it. If I google it I get mostly C++ references rather than C references. When translating expressions like this we usually write:
`real_part + imag_part * _Complex_I`
I found a reference for the contents of `complex.h` [here](https://pubs.opengroup.org/onlinepubs/007904975/basedefs/complex.h.html), it doesn't mention `cmplxf`, and `cmplx`. Maybe they are extensions available with some compilers but not in the standard.
To be sure I think we should use the same syntax as normally with `_Complex_I`

EmilyBourne: According to [this](https://en.cppreference.com/w/c/numeric/complex/CMPLX) they are available from C11. Normally we stick to C99

EmilyBourne:
This branch is blocking around 15 different issues. It needs fixing ASAP. I am happy to do this myself if you do not want to / do not have time to @MenoIy

EmilyBourne:
As the compilation of this module requires information about the python compile options it may be advisable to fix #541 before attempting to fix the compilation

EmilyBourne:
Blocked by #928 (should be fixable in <5 lines after that PR)

EmilyBourne:
After PR #928 , a slow down is still seen. However it is not as significant. The resulting times are still faster than in the master branch before PR #928

EmilyBourne:
After profiling, we can see that the compilation is still the bottleneck. We have:

Average time per call | function
------------------ |----
0.239 | compile_module
0.113 | compile_shared_library
0.038 | cwrappercode
0.019 | SemanticParser::annotate
0.016 | CodePrinter::doprint
0.013 | SyntacticParser::parse

EmilyBourne:
On the master branch we have:

Average time per call	function
0.242	compile_module
0.110	compile_shared_library
0.043	cwrappercode
0.019	SemanticParser::annotate
0.018	CodePrinter::doprint
0.013	SyntacticParser::parse

pyccel/ast/numpy_wrapper.py
```diff
@@ -0,0 +1,283 @@
...
+    recent than this then deprecation warnings are shown.
+
+    The last acceptable numpy version is 1.19. If the current version is older
+    than this then the last acceptable numpy version is the current version
+    """
+    numpy_max_acceptable_version = [1, 19]
+    numpy_current_version = [int(v) for v in np.version.version.split('.')[:2]]
+    numpy_api_macro = '# define NPY_NO_DEPRECATED_API NPY_{}_{}_API_VERSION\n'.format(
+        min(numpy_max_acceptable_version[0], numpy_current_version[0]),
+	    min(numpy_max_acceptable_version[1], numpy_current_version[1]))
```
rakati, line 56: I don't think this is a good way to handle Major and Minor versions, those items should not be separated.
another thing is that I understand from numpy C API is that the `NPY_NO_DEPRECATED_API` mechanism should be used with the new version realized, but I see here the opposite.

EmilyBourne: You are right about not splitting the major and minor versions. Fixed in 754fd1d

EmilyBourne: The flag should be used with the newest version that is tested (numpy 1.19 for us), however, if you are running numpy 1.18 then the macro `NPY_1_19_API_VERSION` is not defined so an error will be raised. Instead we need to use `NPY_1_18_API_VERSION` so the macro exists. I believe there are no changes to the numpy wrapper since numpy 1.7 so it is not a problem if the macro defines a lower version.

Numpy version installed | macro used
----|----
1.10 | `NPY_1_10_API_VERSION`
1.18 | `NPY_1_18_API_VERSION`
1.19 | `NPY_1_19_API_VERSION`
1.20 | `NPY_1_19_API_VERSION`
1.24 | `NPY_1_19_API_VERSION`

saidctb:
we can try to use [ccache](https://ccache.dev/) to speed up the compilation process

EmilyBourne:
@saidctb 
> we can try to use [ccache](https://ccache.dev/) to speed up the compilation process

I'm not sure how that helps as [this](https://github.com/pyccel/pyccel/blob/wrapper_part_1/pyccel/codegen/utilities.py#L90-L96) is avoiding recompiling the same file twice anyway. Would it help for the same file in different folders?

In any case I think it is beyond the scope of this PR. Do you agree?

saidctb:
> compile_module

what does `compile_module` mean?

saidctb:
> On the master branch we have:
> Average time per call 	function
> 0.242 	compile_module
> 0.110 	compile_shared_library
> 0.043 	cwrappercode
> 0.019 	SemanticParser::annotate
> 0.018 	CodePrinter::doprint
> 0.013 	SyntacticParser::parse
> 
> So I cannot see a significant slow down on my computer

what does compile_module mean?

EmilyBourne:
@saidctb 
> what does compile_module mean?

See the new code in `codegen/compiling/compilers.py` from #928 . `compile_module` is the function responsible for compiling a .f90/.c file to a `.o` file

saidctb:
> @saidctb
> 
> > what does compile_module mean?
> 
> See the new code in `codegen/compiling/compilers.py` from #928 . `compile_module` is the function responsible for compiling a .f90/.c file to a `.o` file

ah okay so we have no chance of speeding that up

EmilyBourne:
@saidctb
> ah okay so we have no chance of speeding that up

Not that I found. See slack for more details/discussion

pyccel/ast/numpy_wrapper.py
rakati: what is the case for the opposite, if the user uses a higher version than 1.19?

pyccel/stdlib/cwrapper/cwrapper.c
```diff
@@ -0,0 +1,508 @@
+#include "cwrapper.h"
+
+
+
```
rakati, line 4: add licensing header and a small description of the file contents.

pyccel/stdlib/cwrapper/cwrapper_ndarrays.c
```diff
@@ -0,0 +1,84 @@
+#include "cwrapper_ndarrays.h"
+
+/*
```
rakati, line 3: same here.

pyccel/stdlib/cwrapper/cwrapper_ndarrays.h
```diff
@@ -0,0 +1,9 @@
+#ifndef CWRAPPER_NDARRAYS_H
+# define CWRAPPER_NDARRAYS_H
```
rakati, line 2: same here.

tests/ndarrays/test_ndarrays.c
```diff
@@ -1,5 +1,7 @@
 #include "ndarrays.h"
 #include <unistd.h>
+#include <stdio.h>
```
rakati, line 3: same here

pyccel/stdlib/cwrapper/cwrapper.c
```diff
@@ -0,0 +1,508 @@
...
+float complex PyComplex_to_Complex64(PyObject *object)
+{
+	float complex	c;
+
+	// https://numpy.org/doc/1.17/reference/c-api.array.html#c.PyArray_IsScalar
+	// https://numpy.org/doc/stable/reference/c-api/array.html#c.PyArray_ScalarAsCtype
+	if (PyArray_IsScalar(object, Complex64))
+		PyArray_ScalarAsCtype(object, &c);
+
```
rakati, line 42: delete this line

pyccel/stdlib/cwrapper/cwrapper.c
```diff
@@ -0,0 +1,508 @@
...
+int64_t	PyInt64_to_Int64(PyObject *object)
+{
+	int64_t		i;
+	long long	out;
+
+	out = PyLong_AsLongLong(object);
+
+	i = (int64_t)out;
+
+	return	i;
```
rakati, line 76: why we need a `long long` intermediate variable `out`?
can't we return directly `return *(int64_t)PyLong_AsLongLong(object);`

pyccel/stdlib/cwrapper/cwrapper.c
```diff
@@ -0,0 +1,508 @@
...
+int32_t	PyInt32_to_Int32(PyObject *object)
+{
+	int32_t	i;
+	long	out;
+
+	out = PyLong_AsLong(object);
+
+	i = (int32_t)out;
+
+	return	i;
```
rakati, line 88: same here.

pyccel/stdlib/cwrapper/cwrapper.c
```diff
@@ -0,0 +1,508 @@
...
+int16_t	PyInt16_to_Int16(PyObject *object)
+{
+	int16_t	i;
+	long	out;
+
+	out = PyLong_AsLong(object);
+
+	i = (int16_t)out;
+
+	return	i;
```
rakati, line 100: same here

pyccel/stdlib/cwrapper/cwrapper.c
```diff
@@ -0,0 +1,508 @@
...
+int8_t	PyInt8_to_Int8(PyObject *object)
+{
+	int8_t	i;
+	long	out;
+
+	out = PyLong_AsLong(object);
+
+	i = (int8_t)out;
+
+	return	i;
```
rakati, line 112: same here

pyccel/stdlib/cwrapper/cwrapper.c
rakati: those functions are very small, I suggest inline them for performance, and add them the header directly,

pyccel/stdlib/cwrapper/cwrapper.c
```diff
@@ -0,0 +1,508 @@
...
+ * https://docs.python.org/3/c-api/complex.html#c.PyComplex_Check
+ * https://docs.python.org/3/c-api/float.html#c.PyFloat_Check
+ * https://docs.python.org/3/c-api/bool.html#c.PyBool_Check
+ * https://numpy.org/doc/1.17/reference/c-api.array.html#c.PyArray_IsScalar
+ */
+
+bool    PyIs_Int8(PyObject *o, bool hard_check)
+{
+	if (hard_check == true)
+		return PyArray_IsScalar(o, Int8);
```
rakati, line 273: please can you explain what does hard check stand for?

pyccel/stdlib/cwrapper/cwrapper.c
EmilyBourne: Do you think we need the description in the `.c` file or just in the `.h` file? In c I am not sure that it is necessary for a `.c` file which does not contain a `main` to have a description. What do you think?

pyccel/stdlib/cwrapper/cwrapper.c
EmilyBourne: Why?

pyccel/stdlib/cwrapper/cwrapper.c
EmilyBourne: I agree about reducing this to one line. I knew there was something I wanted to simplify in this file but I forgot what. I think this was it :+1:

pyccel/stdlib/cwrapper/cwrapper.c
rakati: it's an empty line ^^,` else` should follow directly `if` statement, code standard quality.

pyccel/stdlib/cwrapper/cwrapper.c
rakati: yes, it's better to include it in `.h` files.

pyccel/stdlib/cwrapper/cwrapper.c
EmilyBourne: > those functions are very small, I suggest inline them for performance, and add them the header directly,

This depends on where we want to see better performance. The cwrapper shouldn't be the bottleneck of a translated code. Inlining can be a costly operation for the compiler. So I am concerned that we may see a slowdown in the runtime of our tests and very little speedup in return. I should test this

pyccel/stdlib/cwrapper/cwrapper.c
EmilyBourne: Given the code it seems that `hard_check` indicates whether it is assumed that the type is a numpy type or whether conversions are allowed

EmilyBourne: This code should not do `hard_check == true`. Comparison of a boolean to true is a code smell

EmilyBourne: https://github.com/pyccel/pyccel/blob/wrapper_part_1/pyccel/stdlib/cwrapper/cwrapper.c#L250-L257

pyccel/stdlib/cwrapper/cwrapper.c
EmilyBourne: Ah, yes you are right. I assumed there was no else, but looking at it in context, it is strange

pyccel/stdlib/cwrapper/cwrapper.c
EmilyBourne: Licensing fixed  : 682b9a3

pyccel/stdlib/cwrapper/cwrapper.h
```diff
@@ -0,0 +1,63 @@
+#ifndef CWRAPPER_H
+# define CWRAPPER_H
+# define PY_SSIZE_T_CLEAN
```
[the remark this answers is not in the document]
EmilyBourne: Licensing fixed  : 682b9a3

pyccel/stdlib/cwrapper/cwrapper_ndarrays.h
EmilyBourne: Licensing fixed  : 682b9a3

tests/ndarrays/test_ndarrays.c
EmilyBourne: Licensing fixed  : 682b9a3

pyccel/stdlib/cwrapper/cwrapper.h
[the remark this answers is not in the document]
EmilyBourne: Documentation fixed : b0dccb5

pyccel/stdlib/cwrapper/cwrapper_ndarrays.h
EmilyBourne: Documentation fixed : 8d936df

pyccel/stdlib/cwrapper/cwrapper.c
EmilyBourne: Fixed 7d3d8d3

pyccel/stdlib/cwrapper/cwrapper.c
EmilyBourne: Fixed 7d3d8d3
Inlining to test

pyccel/stdlib/cwrapper/cwrapper.c
EmilyBourne: Fixed 7d3d8d3

pyccel/stdlib/cwrapper/cwrapper.c
EmilyBourne: Fixed 7d3d8d3

pyccel/stdlib/cwrapper/cwrapper.c
EmilyBourne: Regarding inlining:
https://softwareengineering.stackexchange.com/a/56230

EmilyBourne: It seems the runner is running slowly today so I'm not certain about the performance, but there doesn't seem to be a slowdown between [this run](https://github.com/pyccel/pyccel/actions/runs/1122914121) and [this run](https://github.com/pyccel/pyccel/actions/runs/1122931679) so I am keeping the inlining

EmilyBourne: Inlining fixed : 812d1eb

pyccel/stdlib/cwrapper/cwrapper.c
EmilyBourne: I am also not clear on why `hard_check` is needed. I don't know when it is set to `False`. It also seems that this argument is always set literally and never parametrically, so I would rather break these up into 2 functions to avoid unnecessary switches. Maybe we can ask @MenoIy to tell us more about this parameter?

EmilyBourne: > I would rather break these up into 2 functions to avoid unnecessary switches

Done 47b2c39

pyccel/stdlib/cwrapper/cwrapper.c
```diff
@@ -0,0 +1,358 @@
...
+bool    PyIs_Int32(PyObject *o, bool hard_check)
+{
+	#ifdef _WIN32
+		return PyLong_Check(o) || PyArray_IsScalar(o, Int32);
+    #else
+        if (hard_check)
+            return PyArray_IsScalar(o, Int32);
+
+        return PyLong_Check(o) || PyArray_IsScalar(o, Int32);
+	#endif
```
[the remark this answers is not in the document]
EmilyBourne: This is done as pyccel does not differentiate between `int` and `int64` (`int32` on windows). Beware the macro used here does not match the macro below

pyccel/stdlib/cwrapper/cwrapper.c
EmilyBourne: I added more inlining, we will see how this affects the compile time moving forward

pyccel/stdlib/cwrapper/cwrapper.c
EmilyBourne: After discussion, this parameter is not necessary. It has been removed: b8cc4c5

pyccel/stdlib/cwrapper/cwrapper.h
```diff
@@ -0,0 +1,207 @@
...
+    return PyArray_IsScalar(o, Int16);
+}
+//--------------------------------------------------------//
+inline bool    PyIs_Int32(PyObject *o)
+{
+#ifdef _WIN32
+    return PyLong_Check(o) || PyArray_IsScalar(o, Int32);
+#else
+    return PyArray_IsScalar(o, Int32);
+#endif
```
rakati, line 169: I think you need to reverse this too.

EmilyBourne: No. The original code was correct. It's just that before the code used `#ifdef` and `#ifndef` to keep a similar order for the commands. I found this confusing as I assumed that the top section was always windows so I changed it. The aim is to only use `PyLongCheck` if the precision matches the default precision (32 on windows, 64 on linux)

rakati: Ah, okay.

pyccel/stdlib/cwrapper/cwrapper.c
rakati: great.

pyccel/stdlib/cwrapper/cwrapper.c
rakati: You did a great job @EmilyBourne. finally, we can work on many issues blocked by the wrapper.

EmilyBourne:
> You did a great job @EmilyBourne. finally, we can work on many issues blocked by the wrapper.

All credit goes to @MenoIy

saidctb (approved):

EmilyBourne (approved):

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -406,12 +311,7 @@ def _body_scalar(self, variable, collect_var, error_check = False, tmp_variable
...
         Responsible for collecting value and managing error and create the body
         of arguments in format:
-            if collect_var is numpy_type:
-                collect_value from numpy type
-            elif collect_var is python_type: #When needed
-                collect value from python type
-            else
-                raise an error
+
         Parameters:
```
yguclu, line 315: No need for colon `:`

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -431,42 +331,28 @@ def _body_scalar(self, variable, collect_var, error_check = False, tmp_variable
...
-                }else if(Type Check == False){
-                    Print TypeError Wrong type
-                    return Null
-                }else if (order check == False){ #check for order for rank > 1
-                    Print NotImplementedError Wrong Order
-                    return Null
-                }
-                collect the value from PyArrayObject
-
         Parameters:
```
yguclu, line 356: No need for colon `:`

yguclu (changes requested):
Excellent work! I have a few minor comments

pyccel/codegen/utilities.py
```diff
@@ -22,7 +22,7 @@
 language_extension = {'fortran':'f90', 'c':'c', 'python':'py'}
 
 #==============================================================================
-def copy_internal_library(lib_folder, pyccel_dirpath):
+def copy_internal_library(lib_folder, pyccel_dirpath, extra_files = None):
```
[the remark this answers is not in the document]
EmilyBourne: Fixed 92dc8b50

The file is not overwritten every time, it is only overwritten if the other files in the folder are also overwritten

pyccel/stdlib/cwrapper/cwrapper_ndarrays.h
```diff
@@ -0,0 +1,34 @@
...
+ * avoid the problem of different implementations of strides in numpy and ndarray.
+ * Parameters :
+ *     np_strides : npy_intp array
+ *     type_size  : data type enum
+ *     nd : size of the array
+ *
+ * Returns    :
+ *     ndarray_strides : a new array with new strides values
+ */
+t_ndarray	pyarray_to_c_ndarray(PyArrayObject *o);
```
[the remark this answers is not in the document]
EmilyBourne: Oops, I copied and pasted from the wrong function. Fixed 1a2db384

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -54,28 +46,53 @@
...
     def stored_in_c_pointer(self, a):
+        """
+        Return True if variable is pointer or stored in a pointer
+        Parameters:
+        -----------
+        a      : Variable
+            Variable holding information needed (is_pointer, is_optional)
+        Returns: Boolean
+        --------
+        """
```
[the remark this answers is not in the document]
EmilyBourne: Fixed f3a7aa40

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -93,6 +110,15 @@ def function_signature(self, expr):
...
     def get_declare_type(self, expr):
+        """
+        Get the declaration type of a variable
+        Parameters:
+        -----------
+        variable : Variable
+            Variable holding information needed to choose the declaration type
+        Returns: String
+        --------
+        """
```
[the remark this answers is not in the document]
EmilyBourne: Fixed f3a7aa40

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -106,27 +132,42 @@ def get_declare_type(self, expr):
...
+        Parameters:
+        -----------
+        name       : String
+            The desired name
+        used_names : Set of strings
+            Set of variable and function names to avoid name collisions
+
+        Returns: Variable
+        -------
+        """
```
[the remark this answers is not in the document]
EmilyBourne: Fixed f3a7aa40

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -106,27 +132,42 @@ def get_declare_type(self, expr):
...
+        -----------
+        dtype : String
+            expression data type
+
+        prec  : Integer
+            expression precision
+
+        Returns: String
+        --------
         """
```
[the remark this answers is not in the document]
EmilyBourne: Fixed f3a7aa40

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: Fixed f3a7aa40

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: Fixed f3a7aa40

yguclu (approved):

yguclu (approved):

PR #890, opened 2021-05-05
Add Matmul operator

### LINKED ISSUE #839: MatMul operator is not recognised
**Describe the bug**
A clear and concise description of what you would like to be implemented.

**Test Code**
```python
@types('real[:,:], real[:], real[:]')
def array_real_2d_1d_matmul(A, x, out):
    out[:] = A @ x
```

currently raises:
```
pyccel:
 |fatal [syntax]: tmp.py [3,13]| Pyccel has encountered syntax that it does not recognise (BinOp(left=Name(id='A', ctx=Load()), op=MatMult(), right=Name(id='x', ctx=Load())))
```
but should be equivalent to:
```python
@types('real[:,:], real[:], real[:]')
def array_real_2d_1d_matmul(A, x, out):
    from numpy import matmul
    out[:] = matmul(A, x)
```

EmilyBourne:
Add the matmul operator `@` to pyccel. Fixes #839 

**Commit Summary**
- Do not calculate shape etc in syntactic stage of `NumpyMatmul`
- Add matmul operator to syntactic stage
- Add matmul operator to semantic stage
- Add matmul operator test
- Activate python matmul tests
- Add matmul to imports if matmul operator used

### DIFF
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -474,11 +474,14 @@ class NumpyMatmul(PyccelInternalFunction):
     __slots__ = ('_dtype','_precision','_shape','_rank','_order')
 
     def __init__(self, a ,b):
+        super().__init__(a, b)
+        if PyccelAstNode.stage == 'syntactic':
+            return
+
         if not isinstance(a, PyccelAstNode):
             raise TypeError('Unknown type of  %s.' % type(a))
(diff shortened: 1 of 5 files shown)


nhamidn (approved):

yguclu (approved):
Very nice @EmilyBourne!

PR #893, opened 2021-05-14
Fix aliases in python printing. Add name attribute to PyccelInternalFunctions

### LINKED ISSUE #754: Add the attribute **name** to numpy functions
For each of our numpy functions, we need to add the attribute **name**.
For example: for **NumpySin** we shall have **sin** etc

### LINKED ISSUE #865: Name clash avoidance ignored in python
**Describe the bug**
Python ignores any use of `AsName` and keeps the default names

**To Reproduce**
```python
from numpy import ones as arr_create

x = arr_create(5)
print(x)
```

```python
from numpy import ones as arr_create

if __name__ == "__main__":
    x = ones((5), dtype=float, order='C')
    print(x
```

**Expected behavior**
The python printer should be able to preserve the names used. For example a dictionary could be used to collect the aliases from the imports

**Language**
python

**Additional Notes**
1. The example above shows that `_print_PythonPrint` is missing a new line ending (the new line ending is removed for indenting, here the closing bracket is the last character so it is removed instead
2. The solution to this should also take into consideration aliased functions. E.g:
    ```
    from numpy import ones as arr_create, product as elem_prod

    x = arr_create(5)
    p = elem_prod(x)
    print(p)
    ```
    which is currently translated to:
    ```
    from numpy import ones as arr_create, product as elem_prod

    if __name__ == "__main__":
        x = ones((5), dtype=float, order='C')
        p = prod(x)
        print(p
    ```

### LINKED ISSUE #873: Imports are missing in the generated code by the PythonPrinter
**Describe the bug**
The python generated code does not contain all the imports from the same library.

**To Reproduce**
This python code :
```python
 from numpy import array
 from numpy import zeros

 a = array([0,1,2,3,4,5,6,7,8,9])

 b = zeros(10)
```

is translated to this python code:
```python
from numpy import zeros

if __name__ == "__main__":
    a = array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9], dtype=int, order='C')
    b = zeros((10), dtype=float, order='C')
```

**Expected behavior**
The generated code must contain all the imports.

**Language**
Python

Once this issue is resolved one test must be activated in `tests/pyccel` folder #872

EmilyBourne:
Fix aliases in python (fixes #865). Fix missing imports in python (fixes #873). Provide names for `PyccelInternalFunction`s (fixes #754 ).

**Commit Summary**
- Provide names for `PyccelInternalFunction`s (fixes #754 )
- Use `PyccelInternalFunction` to reduce code duplication in `codegen.printing.python`
- Add `_aliases` dictionary to handle aliased names in python
- Print imports (but not `additional_imports`) before body to ensure proper collection of aliases
- Explicitly forbid 2 functions with the same name
- Activate pyccel tests with imports (tests aliases and missing imports)
- Put back gfortran link but only link if file does not exist (should avoid this problem in the future)


pyccel/codegen/printing/pycode.py
```diff
@@ -215,59 +219,47 @@ def _print_PythonList(self, expr):
...
 
     def _print_PythonInt(self, expr):
         type_name = type(expr).__name__.lower()
         is_numpy  = type_name.startswith('numpy')
         precision = str(expr.precision*8) if is_numpy else ''
-        name = 'int{}'.format(precision)
+        name = self._aliases.get(expr.name, expr.name)
         if is_numpy:
-            self.insert_new_import(Import('numpy',name))
+            self.insert_new_import('numpy',expr.name,name)
```
yguclu, line 231: I think that this function call would be much easier to understand if you passed the arguments with keywords.

pyccel/codegen/printing/pycode.py
```diff
@@ -494,35 +494,15 @@ def _print_NumpyRandint(self, expr):
...
-        return "{}({})".format(name, args)
-
-    def _print_Max(self, expr):
-        args = ', '.join(self._print(e) for e in expr.args)
-        return 'max({})'.format(args)
-
-    def _print_Min(self, expr):
-        args = ', '.join(self._print(e) for e in expr.args)
-        return 'min({})'.format(args)
-
```
yguclu, line 525: Are these not needed anymore?

yguclu (changes requested):

pyccel/codegen/printing/pycode.py
```diff
@@ -215,59 +219,47 @@ def _print_PythonList(self, expr):
...
-    def _print_PythonMin(self, expr):
-        return 'min({})'.format(self._print(expr.args[0]))
-
-    def _print_PythonMax(self, expr):
-        return 'max({})'.format(self._print(expr.args[0]))
-
-    def _print_PythonSum(self, expr):
-        return 'sum({})'.format(self._print(expr.args[0]))
-
```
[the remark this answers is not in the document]
EmilyBourne: No, thanks to the `name` attribute of `PyccelInternalFunction` they can now all be handled in the same way with the new `_print_PyccelInternalFunction` function

pyccel/codegen/printing/pycode.py
EmilyBourne: See above, handled in `_print_PyccelInternalFunction`

pyccel/codegen/printing/pycode.py
```diff
@@ -188,10 +192,10 @@ def _print_Return(self, expr):
         return 'return {}\n'.format(','.join(self._print(i) for i in expr_return_vars + rhs_list))
 
     def _print_Program(self, expr):
+        imports  = '\n'.join(self._print(i) for i in expr.imports)
         body  = self._print(expr.body)
         body = self._indent_codestring(body)
-        imports  = [*expr.imports, *self.get_additional_imports()]
-        imports  = '\n'.join(self._print(i) for i in imports)
+        imports += '\n'.join(self._print(i) for i in self.get_additional_imports())
```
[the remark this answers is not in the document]
EmilyBourne: Yes, it is important that we print the user defined imports before printing the body in order to collect any aliases provided by the user.
Consider the code:
```
from numpy import ones as true_arr

x = true_arr(3)
```
By visiting `expr.imports` first we collect `ones as true_arr` in order to be able to print `true_arr` when we reach `_print_NumpyOnes`. If we revert these changes then we would print:
```
from numpy import ones as true_arr

x = ones(3)
```
(printing the additional imports still needs to be done after the body as these are created in `_print` calls

pyccel/ast/core.py
```diff
@@ -2824,7 +2824,10 @@ def __str__(self):
                     target=target)
 
     def define_target(self, new_target):
-        self._target.add(new_target)
+        if iterable(new_target):
+            self._target.update(new_target)
+        else:
+            self._target.add(new_target)
```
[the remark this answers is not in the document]
EmilyBourne: Done: 8960e6b

pyccel/codegen/printing/pycode.py
EmilyBourne: Done 480851c

EmilyBourne: Modified e5ec188

pyccel/ast/core.py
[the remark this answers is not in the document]
yguclu: Thanks for adding a docstring to this method. But the class' docstring also needs updating... :wink:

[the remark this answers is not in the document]
EmilyBourne: Done f180322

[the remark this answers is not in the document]
EmilyBourne: I haven't described the `ignore_at_print` parameter because I'm not sure what it does. It's something to do with the headers

pyccel/ast/core.py
```diff
@@ -2747,7 +2747,9 @@ class Import(Basic):
 
     Parameters
     ----------
-    target : str, list, tuple
+    source : str, DottedName, AsName
+        the module from which we import
+    target : str, AsName, list, tuple
         targets to import
```
yguclu, line 2753: Thanks for updating this.

I am now convinced that the parameters' names are quite unfortunate, and that confuses me a lot. The Python [documentation](https://docs.python.org/3/reference/simple_stmts.html#the-import-statement) does not talk about "sources" and "targets", but rather "modules" and "identifiers". In fact the example

```
Import('foo')
```
Has a source but not a target, hence according to our docstring this means that we are importing nothing from module `foo`, which is not the case of course.

Even semantically, the terms "source" and "target" makes no sense to me here, because the target belongs to the source (the imported name belongs to a module).

For this class specifically, I think that we should change the parameters to `module` and `names`, and the method `define_target` should be called `add_imported_name`. In class `PythonCodePrinter`, the method `insert_new_import` should have parameters `(module, name, alias)`.


Do we agree with me? We can do this in another PR of course.

pyccel/ast/core.py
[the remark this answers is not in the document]
yguclu: Thanks

pyccel/codegen/printing/pycode.py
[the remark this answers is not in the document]
yguclu: I see. The problem is that the `_print` function calls have (very important!) side effects, so they must be called in the correct order. I am not very happy with this, but I suppose that we cannot avoid it unless we precede the printer with an additional semantic step that is language-dependent.

pyccel/codegen/printing/pycode.py
```diff
@@ -215,59 +220,54 @@ def _print_PythonList(self, expr):
...
         is_numpy  = type_name.startswith('numpy')
         precision = str(expr.precision*8) if is_numpy else ''
-        name = 'int{}'.format(precision)
-        if is_numpy:
-            self.insert_new_import(Import('numpy',name))
+        name = self._aliases.get(type(expr), expr.name)
+        if is_numpy and name == expr.name:
+            self.insert_new_import(
+                    source = 'numpy',
+                    target = expr.name)
```
yguclu, line 234: I don't understand why you added the condition `name == expr.name` and removed the argument `alias = name`... Can you please explain?

EmilyBourne: If `name != expr.name` that means that the object is stored in the `_aliases` dictionary. The only way for it to be there is if there is already an import for this object, so we don't need to create an additional one

pyccel/ast/core.py
EmilyBourne: > Has a source but not a target, hence according to our docstring this means that we are importing nothing from module foo, which is not the case of course.

I think this actually is the case when translating to fortran, if there's no target then we drop the import

> For this class specifically, I think that we should change the parameters to `module` and `names`

This makes sense to me but I think it's better kept for another PR

yguclu (approved):

PR #898, opened 2021-05-25
Fix complex cast bug

### LINKED ISSUE #897: Casting a complex expression fails on fortran
**Describe the bug**
Casting a complex expression fails on fortran

**To Reproduce**
```python
c = 2
d = 4

value =  complex(c + d * 1j )
```

```bash/fortran/c
/home/emily/Code/test/tmp/__pyccel__/tmp3.f90:13:8:

   13 |   value = cmplx(c + d * 1j, kind=C_DOUBLE_COMPLEX)
      |        1
Error: Syntax error in VALUE statement at (1)
```

**Expected behavior**
There is a missing call to `_print` which causes us to print `1j`

**Language**
Fortran

EmilyBourne:
Ensure `_print` call is used in `_print_PythonComplex` to fix #897. Add test

### DIFF
--- a/pyccel/codegen/printing/fcode.py
+++ b/pyccel/codegen/printing/fcode.py
@@ -862,7 +862,8 @@ class FCodePrinter(CodePrinter):
 
     def _print_PythonComplex(self, expr):
         if expr.is_cast:
-            code = 'cmplx({0}, kind={1})'.format(expr.internal_var,
+            var = self._print(expr.internal_var)
+            code = 'cmplx({0}, kind={1})'.format(var,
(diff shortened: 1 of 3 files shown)


yguclu (changes requested):
Very nice, that was a simple fix! Could you add a test with a more complicated expression (not literal) as shown in the bug description?

EmilyBourne:
> Very nice, that was a simple fix! Could you add a test with a more complicated expression (not literal) as shown in the bug description?

Done d4f31eb

yguclu (approved):

PR #901, opened 2021-05-25
Fix empty returns

### LINKED ISSUE #900: An empty return does not work in pyccel
**Describe the bug**
When an empty return is used pyccel assumes that the function returns something which causes problems

**To Reproduce**
```python
def print_not_zero(a : int):
    if a==0:
        return
    print(a)
    print(a*10)
```

The error message in developer mode is:
```bash/fortran/c
pyccel:
   File "/home/emily/.local/lib/python3.8/site-packages/pyccel/parser/semantic.py", line 1390, in _visit
  File "/home/emily/.local/lib/python3.8/site-packages/pyccel/parser/semantic.py", line 2610, in _visit_Return
    a = self._visit_Assign(a)
  File "/home/emily/.local/lib/python3.8/site-packages/pyccel/parser/semantic.py", line 2046, in _visit_Assign
    d_var  = self._infere_type(rhs, **settings)
  File "/home/emily/.local/lib/python3.8/site-packages/pyccel/parser/semantic.py", line 813, in _infere_type
    errors.report(PYCCEL_RESTRICTION_TODO+'\n'+msg, symbol=expr,
  File "/home/emily/.local/lib/python3.8/site-packages/pyccel/errors/errors.py", line 257, in report
    traceback = ''.join(tb.format_stack(limit=5))
|fatal [semantic]: tmp.py [4,8]| Pyccel has encountered syntax that has not been implemented yet. Please create an issue at https://github.com/pyccel/pyccel/issues and provide a small example of your problem. Do not forget to specify your target language
Type of Object : Nil cannot be infered
```

**Expected behaviour**
An empty return should be acceptable and should create a subroutine not a function

EmilyBourne:
**Commit Summary**
- Add `Return.__repr__`
- Add failing test
- Ensure 'return' or 'return None' is saved as 'return' (fixes #900 for fortran and python)
- Add enumerate in pycode
- Handle empty return in C (fixes #900 for C)

### DIFF
--- a/pyccel/codegen/printing/pycode.py
+++ b/pyccel/codegen/printing/pycode.py
@@ -267,7 +267,6 @@ class PythonCodePrinter(CodePrinter):
 
     def _print_PythonRange(self, expr):
         return 'range({start}, {stop}, {step})'.format(
-                name  = expr.name,
                 start = self._print(expr.start),
                 stop  = self._print(expr.stop ),
                 step  = self._print(expr.step ))
--- a/pyccel/codegen/printing/pycode.py
+++ b/pyccel/codegen/printing/pycode.py
@@ -273,6 +273,12 @@ class PythonCodePrinter(CodePrinter):
                 stop  = self._print(expr.stop ),
                 step  = self._print(expr.step ))
 
+    def _print_PythonEnumerate(self, expr):
+        name = self._aliases.get(type(expr), expr.name)
+        return '{name}({elem})'.format(
+                name = name,
+                elem = self._print(expr.element))
+
     def _print_PythonReal(self, expr):
         return '({}).real'.format(self._print(expr.internal_var))
 
--- a/pyccel/codegen/printing/pycode.py
+++ b/pyccel/codegen/printing/pycode.py
@@ -266,17 +266,14 @@ class PythonCodePrinter(CodePrinter):
             return '{}({}+{}*1j)'.format(name, self._print(expr.real), self._print(expr.imag))
 
     def _print_PythonRange(self, expr):
-        name = self._aliases.get(type(expr), expr.name)
-        return '{name}({start}, {stop}, {step})'.format(
-                name  = name,
+        return 'range({start}, {stop}, {step})'.format(
+                name  = expr.name,
                 start = self._print(expr.start),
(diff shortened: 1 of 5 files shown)


pyccel/codegen/printing/pycode.py
```diff
@@ -273,6 +273,12 @@ def _print_PythonRange(self, expr):
                 stop  = self._print(expr.stop ),
                 step  = self._print(expr.step ))
 
+    def _print_PythonEnumerate(self, expr):
+        name = self._aliases.get(type(expr), expr.name)
+        return '{name}({elem})'.format(
+                name = name,
+                elem = self._print(expr.element))
```
yguclu, line 280: Could the built-in Python function `enumerate` get aliased?

pyccel/parser/syntactic.py
```diff
@@ -589,11 +589,13 @@ def _visit_Compare(self, stmt):
...
-            results = [results]
-        expr = Return(results)
-        return expr
+        if stmt.value is None:
+            return Return([])
+        else:
+            results = self._visit(stmt.value)
+            if not isinstance(results, (list, PythonTuple, PythonList)):
+                results = [results]
+            return Return(results)
```
yguclu, line 598: I am not sure of this way to fix issue #900. In Python `return` is totally equivalent to `return None` and so we should be able to handle both cases in exactly the same way.

pyccel/codegen/printing/pycode.py
EmilyBourne: You are right, this was copied from range which also cannot be aliased. Fixed in 2ff1eda

pyccel/parser/syntactic.py
EmilyBourne: I agree that it is the same in python, but it would not be the same in Fortran/C, I imagine it would be approximately equivalent to `return NULL` with the return type becoming a pointer (but this seems hard/dangerous to implement).

If we consider that `return` and `return None` are strictly equivalent then I could change the condition to `if results is Nil()`

yguclu (approved):
Good job!

yguclu:
Please update the PR's description

EmilyBourne:
> Please update the PR's description

Done

PR #902, opened 2021-05-25
Allow stand-alone expressions

### LINKED ISSUE #896: Cannot call a function without collecting its results
**Describe the bug**
Sometimes functions return arguments which the user does not need to use (e.g. a status indicating if the function call succeeded or not). The user may decide that they do not need to use this value. In that case the code will not run

**To Reproduce**
```python
import numpy as np
def fill_a( r: float, a: 'int[:]' ):

    if ( r == 0.0 ):
        return 0

    for i in range ( len(a) ):
        a[i] = 1.0 / r

    return 1

r = 4.0
a = np.empty(10)
fill_a ( r, a )
```

In fortran:
```bash/fortran/c
/home/emily/Code/test/tmp/__pyccel__/prog_tmp.f90:13:8:

   13 |   fill_a(r, a)
      |        1
Error: ‘fill_a’ at (1) is not a variable

```
In C:
```
/home/emily/Code/test/tmp/__pyccel__/prog_tmp.c: In function ‘main’:
/home/emily/Code/test/tmp/__pyccel__/prog_tmp.c:10:17: error: expected ‘;’ before ‘free_array’
   10 |     fill_a(r, a)free_array(a);
      |                 ^~~~~~~~~~
      |                 ;
```

**Expected behavior**
A temporary variable should be created to collect the result

EmilyBourne:
Allow a stand-alone expression, which appears as a statement by itself with its return value not used or stored. Notably, this handles the case of a function call whose results are not collected (fixes #896).

This is achieved by creating a temporary variable to collect the result of any stand-alone expression. The variable is then dropped if the expression is a function call with no return. This is contrary to the method used in PR #852 where useless statements were removed, but allows us to correctly handle expressions such as `f()+g()`

**Commit Summary**
- Create a temporary variable to collect the result of any stand-alone expression (fixes #896)
- Remove warning created by PR #852 which can no longer be raised
- Ensure a `TupleVariable` can be created from the results of a function call
- Add tests
- Correct `is_temp` collection in `create_new_variable`
- Move warning test to codegen


pyccel/parser/syntactic.py
```diff
@@ -260,7 +260,11 @@ def _visit_Module(self, stmt):
         return code
 
     def _visit_Expr(self, stmt):
-        return self._visit(stmt.value)
+        val = self._visit(stmt.value)
+        if isinstance(val, FunctionCall):
+            tmp_var,_ = create_variable(self._used_names)
+            val = Assign(tmp_var, val)
+        return val
```
yguclu, line 267: Can you add a few words of comment in the code? It will be good as a future reference

yguclu: What happens if the content of `Expr` does not amount to a function call, but maybe it contains some function calls and the resulting expression is not assigned to a left-hand-side variable?

pyccel/codegen/printing/fcode.py
```diff
@@ -1301,7 +1301,8 @@ def _print_Assign(self, expr):
 
             # in the case of a function that returns a list,
             # we should append them to the procedure arguments
-            if isinstance(expr.lhs, (tuple, list, PythonTuple)):
+            if isinstance(expr.lhs, (tuple, list, PythonTuple, InhomogeneousTupleVariable)) \
+                    or (isinstance(expr.lhs, HomogeneousTupleVariable) and expr.lhs.is_stack_array):
```
yguclu, line 1305: Do we have a test for the last case that you consider here?
```python
isinstance(expr.lhs, HomogeneousTupleVariable) and expr.lhs.is_stack_array
```

pyccel/parser/syntactic.py
EmilyBourne: Hmm, in this case I think it will be removed. If we do:
```
3+4
```
then pyccel raises a warning and ignores this line. I think this is also the case if you do:
```
f() + g()
```
but I haven't checked this corner case

EmilyBourne: I've just checked, they both raise the same warning
```
pyccel:
 |warning [semantic]: tmp.py [2]| Expression with no effect has been removed (3 + 4)
```
```
pyccel:
 |warning [semantic]: tmp.py [10]| Expression with no effect has been removed (f() + g())
```

EmilyBourne: > Can you add a few words of comment in the code? It will be good as a future reference

Done 7a57d40

pyccel/codegen/printing/fcode.py
EmilyBourne: I think [this](https://github.com/pyccel/pyccel/blob/master/tests/epyccel/modules/tuples.py#L238) should test it

EmilyBourne: So it is tested whenever the output is multiple values of the same type, unless that type is str or similar that fortran can't handle in a tuple

EmilyBourne: Tested [here](https://github.com/pyccel/pyccel/blob/master/tests/pyccel/scripts/runtest_multiple_results.py#L35)

pyccel/parser/syntactic.py
yguclu: How do we know if the expression has indeed no side effect? Shouldn't we always assign to a temporary variable and keep the expression, in order to be on the safe side?

EmilyBourne: I agree, this is the solution that we came up with when considering expressions such as `3+4` (see #852), but function calls are much more likely to have side effects. I have ported this solution to all expressions. See e9da279

yguclu: Just to be sure... if the expression contained a function `f(x)` call which does not return anything (or that equivalently returns None), and we now assign to a temporary variable, do we still generate the correct Fortran call to a subroutine (not function)?

EmilyBourne: We print `call f(x)`. See [here](https://github.com/pyccel/pyccel/pull/902/files#diff-9ba0c438140113a543a41e16000b939989e841d4fc6a6a4112b5c8e5d8807524R1868)

yguclu: Nice!

yguclu (approved):
This is very good! Can you please expand the PR description given the latest changes? Especially regarding the treatment of `Expr` and the need to revert #852.

EmilyBourne:
> This is very good! Can you please expand the PR description given the latest changes? Especially regarding the treatment of `Expr` and the need to revert #852.

Done

PR #903, opened 2021-05-25
Improve error messages

### LINKED ISSUE #899: Pyccel should be aware of python builtin libraries and raise an error if something unknown is imported
**Describe the bug**
Pyccel should be aware of python builtin libraries and should raise an error if something is imported for which python does not have an implementation

**Test Code**
```python
from sys import exit
```

Currently this raises the error:
```
  File "/home/ratnania/PYCCEL/pyccel/pyccel/parser/base.py", line 101, in get_filename_from_import
    package_dir = str(package.__path__[0])
AttributeError: module 'sys' has no attribute '__path__'
```
This happens because pyccel looks for the file `sys.py`. As `sys` is a builtin library, we should never search for the file `sys.py`.

Instead we should raise a neat error, e.g:
```
| error [semantic]: tmp.py [1]| Package sys is not currently supported by pyccel.
```

When the package is partially supported an error indicating this should be raised, e.g for:
```python
from numpy import sign
```
we should raise something like:
```
| error [semantic]: tmp.py [1]| Function sign of package numpy is not currently supported by pyccel.
```

EmilyBourne:
**Commit Summary**
- Make min/max inhomogeneous arguments error message more explicit
- Stop searching for files corresponding to builtin libraries (fixes #899)
- Raise clear errors when importing an unimplemented builtin library or an unimplemented function from a known library (fixes #899)
- Make `INCOMPATIBLE_TYPES_IN_ASSIGNMENT` error more descriptive
- Fail gracefully if arguments passed to internal function in function call do not meet expectations
- Add error tests


yguclu (approved):
Excellent!

PR #906, opened 2021-06-03
Fix optional argument collection in C

### LINKED ISSUE #699: Optional argument incorrectly collected in C
**Describe the bug**
If an argument is optional then it is a null pointer. The null pointer cannot contain a value, thus if we want to modify it we must allocate it

**To Reproduce**
```python
from numpy import shape

def f(a : 'float[:]', n : int = None):
    if n is None:
        n = shape(a)[0]
    print(n)
```

```c
#include "shape.h"
#include "ndarrays.h"
#include <stdlib.h>
#include <stdint.h>
#include <stdio.h>

void f(t_ndarray a, int64_t *n)
{
    if (n == NULL)
    {
        (*n) = a.shape[0];
    }
    printf("%ld\n", n);
}
```

**Expected behavior**
Create a modifiable partner for every optional variable :
```c
#include "shape.h"
#include "ndarrays.h"
#include <stdlib.h>
#include <stdint.h>
#include <stdio.h>

void f(t_ndarray a, int64_t *n)
{
    int64_t n_val;
    if (n == NULL)
    {
        n = &n_val;
        (*n) = a.shape[0];
    }
    printf("%ld\n", n);
}
```

**Language**
C

EmilyBourne:
**Commit Summary**
- Create a dummy variable if assigning to an optional variable
- Always point the optional variable at the dummy variable's memory before assigning (fixes #699)
- Add test
- Update `gfortran` on macos to `gfortran-10` (`gfortran-8` was removed from the macos image, see https://github.com/actions/virtual-environments/issues/3378)


yguclu (commented):
As I understand it, this approach only works for intent(in) optional arguments?!

EmilyBourne:
> As I understand it, this approach only works for intent(in) optional arguments?!

Yes and no. If you have an intent(inout) argument (therefore an array) and you modify it. I.e:
```
def f(a : 'int[:]' = None):
    if a is not None:
        a[:] = 1
```
then this should still work.
It should be translated to:
```
void f(t_ndarray* a)
{
    int64_t i_0001;
    if (a != NULL)
    {
        for (i_0001 = 0; i_0001 < a->shape[0]; i_0001 += 1)
        {
            GET_ELEMENT(*a, nd_int64, i_0001) = 1;
        }
    }
}
```
(Currently this isn't the case due to issue #757 (see PR #879, blocked by PR #886) )
The additional object is only declared and used if you assign a `Variable` to something (above we assign an `IndexedElement`). Ie:
```
def f(a : 'int[:]' = None):
    if a is not None:
        a = np.ones(4)
```
but in this case `a` is not `inout` in python either.
For a class the case is similar, if we write `cls_obj.var = 3` then `cls_obj` is optional, but `cls_obj.var` isn't

pyccel/codegen/printing/ccode.py
```diff
@@ -1404,18 +1405,30 @@ def _print_AugAssign(self, expr):
...
+        rhs = expr.rhs
+        if isinstance(lhs, Variable) and lhs.is_optional:
+            if lhs in self._optional_partners:
+                tmp_var = self._optional_partners[lhs]
+            else:
+                tmp_var_name = self._parser.get_new_name()
+                tmp_var = lhs.clone(tmp_var_name, is_optional=False)
+                self._additional_declare.append(tmp_var)
+                self._optional_partners[lhs] = tmp_var
+            prefix_code = self._print(AliasAssign(lhs, tmp_var))
```
yguclu, line 1419: Could you explain with some comments in the code what `self._optional_partners` is and how to use it? This is beneficial for future developers.

yguclu (changes requested):

pyccel/codegen/printing/ccode.py
EmilyBourne: Is this sufficiently clear? 4014166

yguclu (approved):
Thanks for the additional comments!

pyccel/codegen/printing/ccode.py
yguclu: It is indeed,

PR #907, opened 2021-06-09
Ensure expression creates variable with the correct class

### LINKED ISSUE #871: ndarray created from expression cannot use class methods/properties. cls_base should be a member of PyccelAstNode
**Describe the bug**
Currently `cls_base` is a property of `Variable` but not of any other class. This prevents the class information being communicated through operators. Which in turn prevents the use of class properties and methods

**To Reproduce**
Both of the following codes raise the same error:
```python
def f(x : 'int[:]', y : 'int[:]'):
    z = x + y
    a = z.sum()
    return a
```
```
def g(x : 'int[:]'):
    z = x[1:]
    a = z.sum()
    return a
```
The error message is:
```bash
pyccel:
 |fatal [semantic]: tmp.py [4,8]| Attribute sum not found
```

This occurs despite the fact that both of the following codes are valid:
```
def h(x : 'int[:]'):
    z = x
    a = z.sum()
    return a
```
```
def p(x : 'int[:]'):
    a = x.sum()
    return a
```

**Expected behavior**
`cls_base` should be a member of `PyccelAstNode`. It should be determined in the `SemanticParser` in order to avoid circular imports (i.e. it should not be necessary to import `ast.numpyext` in `ast.operators` )

**Additional context**
Currently this problem is only visible for the `ndarray` type but fixing it would allow us to support similar syntax for complex numbers (`a.real`, `a.imag`).

This problem probably also exists for the cuda arrays which are under development (cc: @saniQQ )

EmilyBourne:
Ensure that an `ndarray` created from an expression can use class methods/properties (fixes #871). The method suggested to fix this in the issue is to make `cls_base` a member of `PyccelAstNode` however it seems it is impossible to do this without either making `cls_base` an obligatory argument of all classes inheriting from `PyccelAstNode` or adding lots of circular dependencies. Instead the determination of the `cls_base` is handled in `SemanticParser._infere_type`. This method is called in `SemanticParser._visit_DottedName` in order to ensure the correct class is considered

**Commit Summary**
- Create `pyccel.ast.class_defs` to contain definitions of classes recognised by pyccel (python types, and ndarrays)
- Add class definitions for basic types: complex, float, int, bool, string, tuple
- Find `cls_base` in `SemanticParser._infere_type`
- Modify `SemanticParser._visit_DottedName` to collect the class type from `_infere_type` (allows this to work for expressions too)
- Add tests
- Reorder imports in `pyccel.parser.semantic` to order sources alphabetically


pyccel/ast/class_defs.py
```diff
@@ -0,0 +1,169 @@
...
+#=======================================================================================
+
+def get_cls_base(dtype, rank):
+    """
+    From the dtype and rank, determine the base class of an object
+    """
+    if rank == 0:
+        return literal_classes[dtype]
+    else:
+        return NumpyArrayClass
```
yguclu, line 168: Do you remember if we support Numpy arrays of rank 0?

yguclu (approved):
Great job @EmilyBourne!

What do you think of updating the library version so that I can upload the latest code to PyPI?

EmilyBourne:
> Great job @EmilyBourne!
> 
> What do you think of updating the library version so that I can upload the latest code to PyPI?

I agree. I have updated the patch version

pyccel/ast/class_defs.py
EmilyBourne: Currently we don't. @rakati was going to change the rank so this was possible (see #836) but this got stuck. If/when it is added this line will need to be changed to `if rank is None`

Issue #885, opened 2021-04-30
Implement multiple dispatch


EmilyBourne:
**Describe the bug**
Sometimes we want different functions to run different methods depending on the type. In python this is often handled with the `isinstance` decorator, however in pyccel this would be unmanageable. We would therefore prefer to use multiple dispatch. Implementation of this would fix issue #802 

**Test Code**
Provide code which does not currently work but which should do when this issue is fixed:
```python
@dispatch('float')
def get_real(a : 'float'):
    return a

@dispatch('complex')
def get_real(a : 'complex'):
    return a.real
```
The dispatch decorator can either be written ourselves in pyccel or we could add a dependency to [multipledispatch](https://pypi.org/project/multipledispatch/) (last pypi update 2018, last github update 2020(adding python3.8 to testing matrix))

This code should be translated to an `Interface`. In C and fortran this should work out of the box, in python we need to print the individual functions with the dispatch decorator when necessary. Ideally these functions should be grouped if the code is identical. E.g:
```python
@dispatch(('int'),('float'))
@template('T',['int','float'])
def get_real(a : 'T'):
    return a

@dispatch('complex')
def get_real(a : 'complex'):
    return a.real
```

EmilyBourne:
Blocked by #886

PR #909, opened 2021-06-16
Add homogeneous tuples in C

### LINKED ISSUE #850: Homogeneous tuples in C
**Describe the bug**
Homogeneous tuples cannot be implemented in the same way as inhomogeneous tuples as they have an additional difficulty. They must be indexable and the size may be unknown. However they are not as complicated as arrays. They cannot be poiner targets and cannot be modified so there is no need to handle slices or pointers

**Test Code**
```python
a = (1,2,3)
```

**Expected Behaviour**
Mimicking, the solution in Fortran, we should use the ndarray structure and treat a tuple containing objects of different shapes as inhomogeneous

**Language**
C

EmilyBourne:
Add homogeneous tuples in C. Fixes #850 . Fix bugs in unravelling. Improve python printer

**Commit Summary**
- When substituting add new user node before removing old user node to avoid unnecessary node invalidation
- Add `clear_user_nodes` function to prevent objects holding their syntactic and semantic users when they are valid for both stages (e.g. Literals)
- Add simplification step to `PyccelAdd` avoid printing `a+0`
- Unravel expressions such as `a[1:3] = (4,5)` or `a[:] = np.array([3,4])` these expressions don't require memory allocation and have known size
- Correct `Duplicate` loop creation (fixed wrong size, collect multiple expressions to handle tuple assignment unravelling)
- Update `expand_inhomog_tuple_assignments` doc
- Ensure `Duplicate`, `Concatenate`, and `HomogeneousTupleVariable` assignments have memory allocation before unravelling (accessing elements)
- Add `__iter__` to `HomogeneousTupleVariable` (default `__iter__` calls `__getitem__` until `IndexError` is raised, but we don't always know the size)
- Allow `VariableAddress` to contain `IndexedElement`
- Allow `copy_NumpyArray_Data` to also handle homogenous `PythonTuple`s
- Make tuples of arrays inhomogeneous (neither our C framework, nor fortran can handle arrays of pointers)
- Activate tuple tests
- Add functions in functions to python printer
- Add `_print_Duplicate` and `_print_Concatenate` to python printer
- Correct tuple indexing in python printer to print `a[i][j]` instead of `a[i,j]` when indexing `((1,2),(3,4))`
- Avoid printing `from pyccel.decorators import types` multiple times
- Add `_print_Interface` (to be improved, see #885) to python printer

### DIFF
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -1461,8 +1461,8 @@ class SemanticParser(BasicParser):
         expr.clear_user_nodes()
         return expr
     def _visit_Comment(self, expr, **settings):
-        return expr
         expr.clear_user_nodes()
+        return expr
     def _visit_CommentBlock(self, expr, **settings):
         expr.clear_user_nodes()
         return expr
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -1108,7 +1108,7 @@ class SemanticParser(BasicParser):
 
             d_lhs['is_pointer'] = any(v.is_pointer for v in elem_vars)
             d_lhs['is_stack_array'] = d_lhs.get('is_stack_array', False) and not d_lhs['is_pointer']
-            if is_homogeneous:
+            if is_homogeneous and not (d_lhs['is_pointer'] and isinstance(rhs, PythonTuple)):
                 lhs = HomogeneousTupleVariable(dtype, name, **d_lhs, is_temp=is_temp)
             else:
                 lhs = InhomogeneousTupleVariable(elem_vars, dtype, name, **d_lhs, is_temp=is_temp)
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -1451,21 +1451,27 @@ class SemanticParser(BasicParser):
     def _visit_Nil(self, expr, **settings):
         expr.clear_user_nodes()
         return expr
+
     def _visit_EmptyNode(self, expr, **settings):
         expr.clear_user_nodes()
         return expr
+
     def _visit_Break(self, expr, **settings):
         expr.clear_user_nodes()
         return expr
+    
     def _visit_Continue(self, expr, **settings):
         expr.clear_user_nodes()
         return expr
+
     def _visit_Comment(self, expr, **settings):
         expr.clear_user_nodes()
         return expr
+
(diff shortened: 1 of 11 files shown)


pyccel/parser/semantic.py
```diff
@@ -1446,18 +1449,25 @@ def _visit_CodeBlock(self, expr, **settings):
...
         return expr
     def _visit_Comment(self, expr, **settings):
+        expr.clear_user_nodes()
         return expr
     def _visit_CommentBlock(self, expr, **settings):
+        expr.clear_user_nodes()
         return expr
     def _visit_AnnotatedComment(self, expr, **settings):
+        expr.clear_user_nodes()
         return expr
```
yguclu, line 1471: Could you please separate these methods from each other with an empty line?

tests/epyccel/modules/tuples.py
```diff
@@ -88,7 +88,9 @@ def inhomogenous_tuple_3():
     return ai[0], ai[1], ai[2]
 
 def inhomogenous_tuple_2_levels_1():
-    ai = ((1,2), (4,False), (3.0, 'boo'))
+    # TODO [EB 15.06.21] Put back original test when strings are supported
+    #ai = ((1,2), (4,False), (3.0, 'boo'))
+    ai = ((1,2), (4,False), (3.0, True))
```
yguclu, line 93: How was the original test working?

yguclu (changes requested):
Excellent work! I only have two comments

tests/epyccel/modules/tuples.py
EmilyBourne: Fortran supports strings. As this is an inhomogeneous tuple it causes no problems in fortran, but strings aren't supported in C at all

pyccel/parser/semantic.py
EmilyBourne: Fixed. See c62542a

yguclu: Thanks

tests/epyccel/modules/tuples.py
yguclu: Maybe we should write "when strings are supported **in C**" then...

EmilyBourne: Fixed 62ec0d7

yguclu (approved):

PR #921, opened 2021-07-10
Ensure pyccel knows when shape is not known

### LINKED ISSUE #920: Objects allocated in If blocks remember their allocation shape as a known value
**Describe the bug**
Objects allocated in If blocks remember their allocation shape as a known value. In an if block this inital allocation shape, does not necessarily correspond to the final shape so it leads to wrong code

**To Reproduce**
```python
import numpy as np

b = False

if b:
    a = np.array((1,2,3))
else:
    a = np.array((2,3,4,5))

c = a.shape()
```

```bash/fortran/c
program prog_tmp3

  implicit none

  logical(C_BOOL) :: b
  integer(C_INT64_T), allocatable :: a(:)
  integer(C_INT64_T) :: c(0:0_C_INT64_T)

  b = .False._C_BOOL
  if (b) then
    allocate(a(0:2_C_INT64_T))
    a = [1_C_INT64_T, 2_C_INT64_T, 3_C_INT64_T]
  else
    allocate(a(0:3_C_INT64_T))
    a = [2_C_INT64_T, 3_C_INT64_T, 4_C_INT64_T, 5_C_INT64_T]
  end if
  c = [3_C_INT64_T]

end program prog_tmp3
```

**Expected behavior**
The final shape should be unknown and therefore the last line should be:
```
c = [size(a, kind=C_INT64_T)]
```

EmilyBourne:
Set shape to unknown when it is not set to the same size in all branches of an if block. Fixes #920 

**Commit Summary**
- Add method to `Variable` to indicate that shape is unknown after declaration
- Add test which fails on master branch
- Set shape as changeable when it is not set to the same size in all branches of an if block


pyccel/parser/semantic.py
```diff
@@ -2648,6 +2649,19 @@ def _visit_IfSection(self, expr, **settings):
 
     def _visit_If(self, expr, **settings):
         args = [self._visit(i, **settings) for i in expr.blocks]
+        allocations = [arg.get_attribute_nodes(Allocate) for arg in args]
+        var_shapes = [dict() for _ in range(len(args))]
+        for i,allocs in enumerate(allocations):
+            for a in allocs:
+                var_shapes[i][a.variable] = a.shape
+        variables = [v for branch in var_shapes for v in branch]
+        print(variables)
```
yguclu, line 2658: A debugging statement?

pyccel/parser/semantic.py
```diff
@@ -2648,6 +2649,19 @@ def _visit_IfSection(self, expr, **settings):
...
+
+        for v in variables:
+            if not all(v in branch_shapes.keys() for branch_shapes in var_shapes) \
+                    or not all(var_shapes[0][v]==branch_shapes[v] \
+                                for branch_shapes in var_shapes[1:]):
+                v.set_changeable_shape()
```
yguclu, line 2664: I don't know if this optimization makes a noticeable difference, but `var_shapes[0][v]` could be stored in a temporary variable before the `if` statement.

pyccel/parser/semantic.py
```diff
@@ -2648,6 +2649,19 @@ def _visit_IfSection(self, expr, **settings):
 
     def _visit_If(self, expr, **settings):
         args = [self._visit(i, **settings) for i in expr.blocks]
+        allocations = [arg.get_attribute_nodes(Allocate) for arg in args]
+        var_shapes = [dict() for _ in range(len(args))]
+        for i,allocs in enumerate(allocations):
+            for a in allocs:
+                var_shapes[i][a.variable] = a.shape
```
yguclu, line 2656: Could this code be a bit faster and cleaner?
```python
var_shapes = [{a.variable : a.shape for a in allocs} for allocs in allocations]

**Review by yguclu — CHANGES_REQUESTED** [review]
What happens in the case of nested `if/elif/else` blocks?

File: pyccel/parser/semantic.py (hunk 3 above)
Comment (EmilyBourne) [reply pyccel/parser/semantic.py:2656]: I'm not sure. Do you know how this reacts if the same key appears twice?

File: pyccel/parser/semantic.py (hunk 1 above)
Comment (EmilyBourne) [reply pyccel/parser/semantic.py:2658]: oops

File: pyccel/parser/semantic.py (hunk 3 above)
Comment (EmilyBourne) [reply pyccel/parser/semantic.py]: It seems to react the same way as it does in a loop so this is much cleaner :). Fixed in ada1f0e1

File: pyccel/parser/semantic.py (hunk 1 above)
Comment (EmilyBourne) [reply pyccel/parser/semantic.py]: Fixed in ada1f0e1

File: pyccel/parser/semantic.py (hunk 2 above)
Comment (EmilyBourne) [reply pyccel/parser/semantic.py]: Fixed in ada1f0e1

**Comment by EmilyBourne** [comment]
> What happens in the case of nested `if/elif/else` blocks?

Hmm, you are right there seems to still be a bug in this case

File: pyccel/parser/semantic.py [hunk 4]
```diff
@@ -1317,14 +1317,29 @@ def _assign_lhs_variable(self, lhs, d_var, rhs, new_expressions, is_augassign, *
...
+                                alloc_ifsection = last_alloc_ifsection[-1]
+                                last_alloc_ifsection = alloc_ifsection.get_user_nodes(IfSection)
+
+                            ifsection_has_if = len(alloc_ifsection.get_direct_user_nodes(
+                                                                lambda x: isinstance(x,If))) == 1 \
+                                            if alloc_ifsection else False
+
+                            if alloc_ifsection and not ifsection_has_if:
+                                status = last_allocation.status
+                            elif last_allocation.get_user_nodes((If, For, While)):
```
rakati, line 1341: what do you think about the case of a function inside a function?

rakati (approved):

saidctb (approved):

yguclu (approved):

PR #922, opened 2021-07-10
Convert IfTernaryOperator returning array or string to If block

### LINKED ISSUE #492: Strings in  IfTernaryOperator are not working in Fortran

### LINKED ISSUE #493: Arrays in IfTernaryOperator are not working in Fortran

EmilyBourne:
Convert `IfTernaryOperator` returning array or string to If block. Fixes #493 and fixes #492 .

**Commit Summary**
- In pytest report, output information about XPASS rather than XFAIL
- Make tuples `allocatable` by default. Stack arrays can cause problems if they have different sizes in different branches of an if block
- Use `is_ndarray` to determine if an object is a target not `allocatable` (stack arrays can be targets)
- [BUGFIX] Add an `allocate` node when shapes match if previous allocate was in an `IfSection`
- Add tests
- Use unix line endings in test file
- Remove xfail marks
- Use the same method to handle if ternary operator in expression or generator comprehension in expression


pyccel/parser/semantic.py
```diff
@@ -178,6 +178,8 @@ def __init__(self, inputs, **kwargs):
         # used to store the local variables of a code block needed for garbage collecting
         self._allocs = []
 
+        # used to store code split into multiple lines to be reinserted in the CodeBlock
+        self._additional_exprs = []
         # we use it to detect the current method or function
```
yguclu, line 183: @EmilyBourne Is this comment line still meaningful to you? I cannot understand it...

EmilyBourne: This still makes sense to me. When we are visiting a line sometimes it needs to be split into multiple parts. E.g:
```
a = 55 + (c if b else d)
```
Needs to be written as:
```
Dummy = c if b else d
a = 55 + Dummy
```
However when we call `_visit_IfTernaryOperator` we need to return a `PyccelAstNode` (and not a `CodeBlock`) so that the `PyccelAdd` knows how to handle the result. So the additional expression `Dummy = c if b else d` goes in `self._additional_exprs` and is reinserted in the `CodeBlock` once the `_visit` command exits to there (see [here](https://github.com/pyccel/pyccel/blob/devel-issue493/pyccel/parser/semantic.py#L1612-L1613))

yguclu (approved):

pyccel/parser/semantic.py
yguclu: Thanks for the explanation! But I was referring to `# we use it to detect the current method or function` 😜

EmilyBourne: Ah! Hmm, no. Removed lines don't show up on git blame

EmilyBourne: It predates our versioning tags

yguclu: Shall we remove that line then?

EmilyBourne: Apparently the line below the comment was removed on this commit: 2757f9db4a5d76f468b3d766778886a17d74ef1f
Which was about creating the scoping system (24.11.2018). I will remove the comment

EmilyBourne: Fixed 854ffb68

yguclu (approved):

PR #924, opened 2021-07-11
Initialise stack array at declaration without contents

### LINKED ISSUE #923: Contents of numpy arrays are declared locally
**Describe the bug**
Contents of numpy arrays are declared locally which means that if they exist in a block they can go out of scope almost immediately

**To Reproduce**
```python
import numpy as np

@stack_array('a')
def f(b : bool):

    if b:
        a = np.array([1,2,3])
    else:
        a = np.array([4,5,6])

    print(a)
```

```bash/fortran/c
#include "tmp.h"
#include <stdlib.h>
#include "ndarrays.h"
#include <stdbool.h>
#include <stdio.h>
#include <stdint.h>

void f(bool b)
{
    t_ndarray a;
    int64_t i;
    if (b)
    {
        int64_t array_dummy_0001[] = {1, 2, 3};
        a = (t_ndarray){
            .nd_int64=array_dummy_0001,
            .shape=(int64_t[]){3},
            .strides=(int64_t[1]){0},
            .nd=1,
            .type=nd_int64,
            .is_view=false
        };
        stack_array_init(&a);
    }
    else
    {
        int64_t array_dummy_0002[] = {4, 5, 6};
        a = (t_ndarray){
            .nd_int64=array_dummy_0002,
            .shape=(int64_t[]){3},
            .strides=(int64_t[1]){0},
            .nd=1,
            .type=nd_int64,
            .is_view=false
        };
        stack_array_init(&a);
    }
    printf("%s", "[");
    for (i = 0; i < 2; i += 1)
    {
        printf("%ld ", GET_ELEMENT(a, nd_int64, i));
    }
    printf("%ld]\n", GET_ELEMENT(a, nd_int64, 2));
}

```

**Expected behavior**
Memory problems should be avoided. However I believe this means that it is impossible to specify the array with `{}` notation. Instead unravelling must be used

**Language**
C

EmilyBourne:
Initialise stack array at declaration without contents. This allows stack arrays to then be treated the same as allocatable arrays and fixes #923 

Previously the following code:
```python
import numpy as np

@stack_array('a')
def f(b : bool):

    if b:
        a = np.array([1,2,3])
    else:
        a = np.array([4,5,6])

    print(a)
```

generated:
```c
#include "tmp.h"
#include <stdlib.h>
#include "ndarrays.h"
#include <stdbool.h>
#include <stdio.h>
#include <stdint.h>

void f(bool b)
{
    t_ndarray a;
    int64_t i;
    if (b)
    {
        int64_t array_dummy_0001[] = {1, 2, 3};
        a = (t_ndarray){
            .nd_int64=array_dummy_0001,
            .shape=(int64_t[]){3},
            .strides=(int64_t[1]){0},
            .nd=1,
            .type=nd_int64,
            .is_view=false
        };
        stack_array_init(&a);
    }
    else
    {
        int64_t array_dummy_0002[] = {4, 5, 6};
        a = (t_ndarray){
            .nd_int64=array_dummy_0002,
            .shape=(int64_t[]){3},
            .strides=(int64_t[1]){0},
            .nd=1,
            .type=nd_int64,
            .is_view=false
        };
        stack_array_init(&a);
    }
    printf("%s", "[");
    for (i = 0; i < 2; i += 1)
    {
        printf("%ld ", GET_ELEMENT(a, nd_int64, i));
    }
    printf("%ld]\n", GET_ELEMENT(a, nd_int64, 2));
}

```

Now it generates:
```c
#include "tmp.h"
#include "ndarrays.h"
#include <stdint.h>
#include <stdbool.h>
#include <stdlib.h>
#include <stdio.h>

void f(bool b)
{
    int64_t array_dummy_0003[3];
    t_ndarray a = (t_ndarray){
        .nd_int64=array_dummy_0003,
        .shape=(int64_t[]){3},
        .strides=(int64_t[1]){0},
        .nd=1,
        .type=nd_int64,
        .is_view=false
    };
    stack_array_init(&a);
    int64_t i;
    if (b)
    {
        int64_t array_dummy_0001[] = {1, 2, 3};
    }
    else
    {
        int64_t array_dummy_0002[] = {4, 5, 6};
        memcpy(a.nd_int64, array_dummy_0002, a.buffer_size);
    }
    printf("%s", "[");
    for (i = 0; i < 2; i += 1)
    {
        printf("%ld ", GET_ELEMENT(a, nd_int64, i));
    }
    printf("%ld]\n", GET_ELEMENT(a, nd_int64, 2));
}
```

### DIFF
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -282,7 +282,6 @@ class CCodePrinter(CodePrinter):
             String
                 Return a str that contains the declaration of a dummy data_buffer
                        and a call to an operator which copies it to an NdArray struct
-                if the ndarray is a stack_array the str will contain the initialization
         """
         rhs = expr.rhs
         lhs = expr.lhs
@@ -298,18 +297,12 @@ class CCodePrinter(CodePrinter):
 
         if isinstance(arg, Variable):
             arg = self._print(arg)
-            if expr.lhs.is_stack_array:
-                cpy_data = self._init_stack_array(expr, rhs.arg)
-            else:
-                cpy_data = "memcpy({0}.{2}, {1}.{2}, {0}.buffer_size);\n".format(lhs, arg, dtype)
+            cpy_data = "memcpy({0}.{2}, {1}.{2}, {0}.buffer_size);\n".format(lhs, arg, dtype)
             return '%s' % (cpy_data)
         else :
             arg = ', '.join(self._print(i) for i in arg)
             dummy_array = "%s %s[] = {%s};\n" % (declare_dtype, dummy_array_name, arg)
-            if expr.lhs.is_stack_array:
-                cpy_data = self._init_stack_array(expr, dummy_array_name)
-            else:
-                cpy_data = "memcpy({0}.{2}, {1}, {0}.buffer_size);\n".format(self._print(lhs), dummy_array_name, dtype)
+            cpy_data = "memcpy({0}.{2}, {1}, {0}.buffer_size);\n".format(self._print(lhs), dummy_array_name, dtype)
             return  '%s%s' % (dummy_array, cpy_data)
 
     def arrayFill(self, expr):
@@ -323,37 +316,11 @@ class CCodePrinter(CodePrinter):
         ------
             String
                 Return a str that contains a call to the C function array_fill,
-                if the ndarray is a stack_array the str will contain the initialization
         """
         rhs = expr.rhs
         lhs = expr.lhs
         code_init = ''
         declare_dtype = self.find_in_dtype_registry(self._print(rhs.dtype), rhs.precision)
-
-        if lhs.is_stack_array:
-            symbol_map = {}
-            used_names_tmp = self._parser.used_names.copy()
-            sympy_shapes = [pyccel_to_sympy(s, symbol_map, used_names_tmp) for s in lhs.alloc_shape]
-
-            length = functools.reduce(operator.mul, sympy_shapes)
-            printable_length = functools.reduce(PyccelMul, lhs.alloc_shape)
-
-            length_code = self._print(printable_length)
-
-            if length.is_constant():
-                buffer_array = "({declare_dtype}[{length}]){{}}".format(
-                                        declare_dtype = declare_dtype,
-                                        length=length_code)
-            else:
-                dummy_array_name, _ = create_incremented_string(self._parser.used_names,
-                                                                prefix = lhs.name+'_data')
-                code_init += "{dtype} {name}[{length}];\n".format(
-                        dtype  = declare_dtype,
-                        name   = dummy_array_name,
-                        length = length_code)
-                buffer_array = dummy_array_name
-
-            code_init += self._init_stack_array(expr, buffer_array)
 
         if rhs.fill_value is not None:
             if isinstance(rhs.fill_value, Literal):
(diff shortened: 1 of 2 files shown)


pyccel/codegen/printing/ccode.py
```diff
@@ -323,76 +316,53 @@ def arrayFill(self, expr):
...
         Parameters
         ----------
             expr : PyccelAstNode
                 The Assign Node used to get the lhs and rhs
-            buffer_array : String
-                The data buffer
         Returns
         -------
             Returns a string that contains the initialization of a stack_array
         """
```
yguclu, line 342: There are now two different return values, so the docstring should be updated accordingly

yguclu (changes requested):
Could you add an example to the PR description, with the generated C code before and after this PR?

pyccel/codegen/printing/ccode.py
EmilyBourne: Fixed 16f3bab2

yguclu:
Thanks for adding the code snippet.

I wonder if the C compilers are able to understand that the variables `array_dummy_0001` and `array_dummy_0002` are temporaries, and if they can therefore "optimize out" the useless call to `memcpy`.

Otherwise, we may want to avoid a call to `memcpy` altogether. I have found this trick on the internet, by wrapping the dummy array into a struct (which can be initialized after declaration):
```C
struct array_int64_t_3 { int64_t arr[3]; };  // dummy type
struct array_int64_t_3 array_dummy_0001;     // dummy struct variable

t_ndarray a;

if (b)
{
    array_dummy_0001 = (struct array_int64_t_3) { .arr = {1, 2, 3} }; // initialize dummy struct
    a = (t_ndarray){
        .nd_int64 = array_dummy_0001.arr, // NOTE: .arr
        .shape    = (int64_t[]){3},
        .strides  = (int64_t[1]){0},
        .nd       = 1,
        .type     = nd_int64,
        .is_view  = false
    };
    stack_array_init(&a);
}
else
{
    array_dummy_0001 = (struct array_int64_t_3) { .arr = {4, 5, 6} }; // initialize dummy struct
    a = (t_ndarray){
        .nd_int64 = array_dummy_0001.arr, // NOTE: .arr
        .shape    = (int64_t[]){3},
        .strides  = (int64_t[1]){0},
        .nd       = 1,
        .type     = nd_int64,
        .is_view  = false
    };
    stack_array_init(&a);
}
```
What do you think?

EmilyBourne:
> Thanks for adding the code snippet.
> 
> I wonder if the C compilers are able to understand that the variables `array_dummy_0001` and `array_dummy_0002` are temporaries, and if they can therefore "optimize out" the useless call to `memcpy`.
> 
> Otherwise, we may want to avoid a call to `memcpy` altogether. I have found this trick on the internet, by wrapping the dummy array into a struct (which can be initialized after declaration):
>
> What do you think?

Hmm, I strongly suspect that they won't "optimize out" `memcpy`.

Do you think it is a problem if we set shape/stride in multiple places? Will this result in them being on the heap? What about:
```
struct array_int64_t_3 { int64_t arr[3]; };  // dummy type
struct array_int64_t_3 array_dummy_0001;     // dummy struct variable

t_ndarray a = (t_ndarray){
        .shape    = (int64_t[]){3},
        .strides  = (int64_t[1]){0},
        .nd       = 1,
        .type     = nd_int64,
        .is_view  = false
    };

if (b)
{
    array_dummy_0001 = (struct array_int64_t_3) { .arr = {1, 2, 3} }; // initialize dummy struct
    a.nd_int64 = array_dummy_0001.arr;
    stack_array_init(&a);
}
else
{
    array_dummy_0001 = (struct array_int64_t_3) { .arr = {4, 5, 6} }; // initialize dummy struct
    a.nd_int64 = array_dummy_0001.arr;
    stack_array_init(&a);
}
```
?

yguclu:
I don't think there is a problem if we set shape/strides in multiple places. I was just trying to write a code that is as close as possible to what gets generated on the `master` branch! 😉 

The last code you proposed looks really good to me. Any chance you can profile it and compare it to what you currently have on this branch?

EmilyBourne:
Timeit results for the following code:
```
import numpy as np

@stack_array('a')
def f(b : bool):

    if b:
        a = np.array([1,2,3])
    else:
        a = np.array([4,5,6])

    return a[0], a[1], a[2]
```
with repeats = 5, and number = 1 000 000 give:
Python: [6.601867691999814, 6.9459093150003355, 7.238584541999899, 6.99193672399997, 6.957575059999726]
Fortran: [0.703614565999942, 0.6876528880002297, 0.7600069530003566, 0.6912920069999018, 0.6921548260002055]
Current C: [0.7649617430001854, 0.8280447170000116, 0.7402553920001083, 0.7536303840001892, 0.7700385480002296]
Proposed C:[0.7391574590001255, 0.6891662809998707, 0.7866308180000487, 0.7608685060004063, 0.8502697529997931]

So there doesn't seem to be much/any gain on my computer.

Given that fact I don't know if it is worth making this change given that it brings multiple potential problems:
- The structure name must avoid potential name collisions
- The structure should probably be reused to help avoid name collisions if there are multiple arrays requiring the same sized buffer
- The size can be dependant on an argument, I'm not sure if this could cause problems for structures? E.g:
    ```
    @stack_array('tmp')
    def test_degree(degree : int):
        from numpy import empty

        tmp = empty(degree+1, dtype=float)
        for i in range(degree+1):
            tmp[i]=0.
        return 1
    ```
- In multiple dimensions the name of the structure is perhaps harder to choose `array_int64_t_3_4`? `array_int64_t_12`?

yguclu:
Thanks for investigating this!

Given that the timings do not show a clear winner, I think there is no need to modify this branch.

yguclu (approved):

PR #928, opened 2021-07-27
Clean compilation

### LINKED ISSUE #541: Link libraries with specified compiler

### LINKED ISSUE #917: Pickling and unpickling should use flock
**Describe the bug**
Writing or reading needs to be protected in a multi thread environment. This is the case for the compiler commands, but it hasn't been put in place for the pickling

**Additional context**
Very very occasionally causes failures in unit tests

EmilyBourne:
Overhaul the compilation method. A class `Compiler` is now used to handle all compilation steps. A class `CompileObj` is used to store all information required to compile a given file. Default compile configurations for basic compilers (gcc, gfortran, icc, ifort) are saved in json files in `pyccel/compilers` these are determined thanks to the `sysconfig` module. Fixes #541.

After this PR, tests run much faster. The times below show the last 3 tests run on the master and `ebourne_compiling` branches, followed by the average times
**Master branch**
|Linux (python3.6)|MacOs|Windows|
|-------------------|--------|----------|
|19m37                | 16m54   |    28m34  |
|18m46                | 21m13   |    24m28  |
|21m13                | 19m53   |    23m54  |
|                           |               |                 |
|19m52                | 19m20   |    25m39  |

**ebourne_compiling**
|Linux (python3.6)|MacOs|Windows|
|-------------------|--------|----------|
|13m13                | 17m23   |    15m17  |
|11m59                | 16m17   |    15m25  |
|15m30                | 17m52   |    18m01  |
|                           |               |                 |
|13m34                | 17m11   |    16m11  |

**Supported compilers:**
The following compilers have been tested:
- *GNU* : `gcc`/`gfortran`
- *intel* : `icc`/`ifort`   (tested by @EmilyBourne )
- *PGI* : `pgcc`/`pgfortran` (tested by @saidctb )
- *nvidia* : `nvc`/`nvfort` (tested by @rakati )

**Commit Summary**
- Create `CompileObj` to hold all information required to compile a given file
- Create `Compiler` to hold all information about a given compiler
- Remove `cwrapper.py`. The `setup_X.py` file is no longer required
- Use `Compiler` and `CompileObj` to simplify `pipeline.py` and `python_wrapper.py`
- Allow multiple `accelerators`
- Make `mpi` an accelerator instead of specifying the `mpi_compiler`
- Specify the compiler family (e.g. `GNU`) instead of the compiler name
- Add a `wrapper_flags` argument to separate the flags for the file (fortran) from the flags for the wrapper (c)
- When compiling a program, compile everything as modules then link separately. This procedure is closer to the procedure used for shared libraries
- Rename `pyc_math` to `pyc_math_c` and `pyc_math_f90`. This allows us to differentiate between the `.o` files created from the source files which means we don't need to systematically recompile the library
- Only overwrite stdlib files if they are out of date
- Only recompile stdlib files if the source files have been updated
- Add `pyccel.compilers.generate_default` to generate the default json files
- Lock file during pickling. Fixes #917
- Add a file describing how to use different compilers


EmilyBourne:
It seems either impossible or inadvisable to generate the json during the installation. See [here](https://stackoverflow.com/a/24749871/5957195)

EmilyBourne:
@saidctb Do you have access to `nvfort` / `nvc` to test default compiler parameters for nvidia?

pyccel/codegen/compiling/basic.py
```diff
@@ -0,0 +1,237 @@
...
+        self._prog_target = rel_mod_name
+        if sys.platform == "win32":
+            self._prog_target += '.exe'
+
+        self._target = self._module_target if is_module else self._prog_target
+        self._target = os.path.join(folder, self._target)
+
+        self._lock         = FileLock(self.target+'.lock')
+
+        self._flags        = list(flags or ())
```
rakati, line 77: the default value for flags is seted as `()` so it is enough to write just `self._flags        = list(flags)`

pyccel/codegen/compiling/basic.py
```diff
@@ -0,0 +1,237 @@
...
+        self._includes     = [folder, *(includes or ())]
+        self._libs         = list(libs or ())
+        self._libdirs      = list(libdirs or ())
+        self._accelerators = set(accelerators or ())
```
rakati, line 81: same here

pyccel/codegen/compiling/basic.py
EmilyBourne: The problem is that the default value [here](https://github.com/pyccel/pyccel/blob/master/pyccel/codegen/pipeline.py#L48-L64) is `None`, so that gets passed through anyway, hence the problem here. Do you think I should change the defaults here? I would prefer to change the defaults in `execute_pyccel` but I'm not sure if it will work. I will try, I might have changed enough for it to be ok now

EmilyBourne: My bad, you are right that works now. Fixed in 518629a

pyccel/codegen/compiling/basic.py
EmilyBourne: Fixed in 518629a

pyccel/codegen/python_wrapper.py
```diff
@@ -26,22 +19,16 @@
 
 #==============================================================================
 
-PY_VERSION = sys.version_info[0:2]
-
 fortran_c_flag_equivalence = {'-Wconversion-extra' : '-Wconversion' }
```
rakati, line 22: can we add those flags to the JSON file too?

EmilyBourne: We can, but I'm not sure where they should go. I have tried to find the most general solution possible so translating flags only seems logical if I provide equivalences for all the flags (e.g. intel<->GNU). The solution I came up with was to split the flags into:
- Flags for compiling the source code (which must be compatible with the target language, e.g. Fortran)
- Flags for compiling the wrapper (which must be compatible with C)
This solution means that this line is dead code. I have removed it in 3cfaa14d
Do you think that is sufficient or do you think we need some kind of flag equivalence map?

pyccel/codegen/compiling/basic.py
```diff
@@ -0,0 +1,237 @@
...
+                 folder,
+                 is_module    = True,
+                 flags        = (),
+                 includes     = (),
+                 libs         = (),
+                 libdirs      = (),
+                 dependencies = (),
+                 accelerators = ()):
+        if not all(isinstance(d, CompileObj) for d in dependencies):
+            raise TypeError("Dependencies require necessary compile information")
```
rakati, line 59: I think we don't need this check, we are doing it in the `add_dependencies` function.

EmilyBourne: I agree. Fixed in a114d1cc

pyccel/codegen/python_wrapper.py
rakati: yes, it's enough.
I was before thinking about a method to simplifying compiler flags for the user so the user doesn't need to know flags differences in compilers, and we in out part creat map for the flags between compilers and suggest a friendly name for the user, but this seems hard and loosing of time.
there are a lot of important things to do than that.

EmilyBourne: My suspicion is that either a user won't know much about flags (and therefore won't bother to specify any), or they will know the flags for the compiler they usually use and in that case they can specify that compiler.

Flags are usually useful for accelerating code (e.g. `-funroll-loops`), but these flags won't do much/anything in the wrapper anyway so, I don't think it's a huge problem if the flags are only applied to the source

pyccel/stdlib/math/pyc_math_c.c
```diff
@@ -3,7 +3,7 @@
 /* or go to https://github.com/pyccel/pyccel/blob/master/LICENSE for full license details. */
 /* --------------------------------------------------------------------------------------- */
 
-#include "pyc_math.h"
+#include "pyc_math_c.h"
```
rakati, line 6:
```c
#ifndef         PYC_MATH_H
#define         PYC_MATH_H
```
to be changed to 
```c
#ifndef         PYC_MATH_C_H
#define         PYC_MATH_C_H
```
to avoid infinite include-loop.

.gitignore
```diff
@@ -37,6 +37,8 @@ doc/source
 
 *.swp
 
+pyccel/compilers/*.json
```
rakati, line 40: just question, why you ignored `.json` file?

pyccel/stdlib/math/pyc_math_c.c
EmilyBourne: The infinite loop is already avoided (the name of the macro can be whatever), but you are right it is much better if this matches the file name. Fixed in 0762ad1c

.gitignore
EmilyBourne: They need to be generated on your computer so that the config values provided by `sysconfig` match the setup on your computer (os, python version, libraries used when python was compiled). I don't want anyone to, for example, accidentally commit a `.json` configured for linux, which would then break the setup for anyone on mac or windows

pyccel/codegen/python_wrapper.py
rakati: I agree with you. in my opinion the only the user has the right to apply some flags on the source files at his own risk.
for us, we may only apply some flags on the libraries we create and we tested already.

EmilyBourne: Flags for a specific library should also be able to be done easily now by passing the flags [here](https://github.com/pyccel/pyccel/blob/ebourne_compiling/pyccel/codegen/pipeline.py#L31-L33)

rakati (commented):
can you add `__slots__` for the new classes you create?

pyccel/codegen/python_wrapper.py
rakati: Excellent job @Emily, I always think about improving the compiling in Pyccel, but I didn't have the necessary courage ^_^

EmilyBourne:
> can you add `__slots__` for the new classes you create?

Good catch. Fixed 730e2060

pyccel/codegen/python_wrapper.py
EmilyBourne: Haha, neither did I. That's why it took a year to get done. But I think it's necessary to fix #886 cleanly. It needs to include the python files. With these improvements we just need to specify `accelerators=['python']` but without them there would have been some strange commands all over the place to find the right include files :/

rakati: we will take care of that too.

rakati (approved):

saidctb:
> @saidctb Do you have access to `nvfort` / `nvc` (successor to `pgfortran`, see [here](https://forums.developer.nvidia.com/t/nvfortran-vs-pgfortran/165768/2)) to test default compiler parameters for nvidia?

yes I have access to them in the cluster

saidctb:
did you notice any speedup in the compilation process?

EmilyBourne:
@saidctb 
> did you notice any speedup in the compilation process?

Yes, as mentioned in the PR description, tests are now running between 4 and 6 mins faster

EmilyBourne:
> > @saidctb Do you have access to `nvfort` / `nvc` (successor to `pgfortran`, see [here](https://forums.developer.nvidia.com/t/nvfortran-vs-pgfortran/165768/2)) to test default compiler parameters for nvidia?
> 
> yes I have access to them in the cluster

Do you think you would be able to complete [this file](https://github.com/pyccel/pyccel/blob/ebourne_compiling/pyccel/compilers/generate_default.py) to be able to add back support for those compilers? I had a quick look yesterday but I couldn't find the documentation for the flags so as I'm not really familiar with the compiler I got a bit stuck

saidctb:
I'm testing this branch in the cluster and I have this problem
```
gfortran -O3 -fPIC -Wno-unused-result -Wsign-compare -DNDEBUG -O2 -Wall -fPIC -O2 -isystem /u/saicctb/conda-envs/myenv/include -fPIC -O2 -isystem /u/saicctb/conda-envs/myenv/include -pthread -B /u/saicctb/conda-envs/myenv/compiler_compat libpython3.9.a -pthread -B /u/saicctb/conda-envs/myenv/compiler_compat -shared -Wl,-rpath,/u/saicctb/conda-envs/myenv/lib -Wl,-rpath-link,/u/saicctb/conda-envs/myenv/lib -Wl,-rpath,/u/saicctb/conda-envs/myenv/lib -Wl,-rpath-link,/u/saicctb/conda-envs/myenv/lib -I /cobra/u/saicctb/__pyccel__ -I /cobra/u/saicctb/__pyccel__ -I /cobra/u/saicctb/__pyccel__ -I /cobra/u/saicctb/__pyccel__ -I /cobra/u/saicctb/__pyccel__ -I /cobra/u/saicctb/__pyccel__ -I /cobra/u/saicctb/__pyccel__ -I /u/saicctb/conda-envs/myenv/include/python3.9 -I /u/saicctb/.local/lib/python3.9/site-packages/numpy/core/include -L /u/saicctb/conda-envs/myenv/lib -L /u/saicctb/conda-envs/myenv/lib -L /u/saicctb/conda-envs/myenv/lib -L /u/saicctb/conda-envs/myenv/lib/python3.9/config-3.9-x86_64-linux-gnu -lm /cobra/u/saicctb/__pyccel__/test_pyccel.o /cobra/u/saicctb/__pyccel__/bind_c_test_pyccel.o /cobra/u/saicctb/__pyccel__/test_pyccel_wrapper.o -o /cobra/u/saicctb/__pyccel__/test_pyccel.cpython-39-x86_64-linux-gnu.so

  File "/u/saicctb/.local/bin/pyccel", line 33, in <module>
  File "/cobra/u/saicctb/pyccel/pyccel/commands/console.py", line 216, in pyccel
    execute_pyccel(filename,
  File "/cobra/u/saicctb/pyccel/pyccel/codegen/pipeline.py", line 370, in execute_pyccel
    generated_filepath = create_shared_library(codegen,
  File "/cobra/u/saicctb/pyccel/pyccel/codegen/python_wrapper.py", line 89, in create_shared_library
    sharedlib_filepath = src_compiler.compile_shared_library(wrapper_compile_obj,
  File "/cobra/u/saicctb/pyccel/pyccel/codegen/compiling/compilers.py", line 371, in compile_shared_library
  File "/cobra/u/saicctb/pyccel/pyccel/codegen/compiling/compilers.py", line 400, in run_command
    raise RuntimeError(err_msg)
gfortran: error: libpython3.9.a: No such file or directory
```
did you have this problem before?

saidctb:
> > > @saidctb Do you have access to `nvfort` / `nvc` (successor to `pgfortran`, see [here](https://forums.developer.nvidia.com/t/nvfortran-vs-pgfortran/165768/2)) to test default compiler parameters for nvidia?
> > 
> > 
> > yes I have access to them in the cluster
> 
> Do you think you would be able to complete [this file](https://github.com/pyccel/pyccel/blob/ebourne_compiling/pyccel/compilers/generate_default.py) to be able to add back support for those compilers? I had a quick look yesterday but I couldn't find the documentation for the flags so as I'm not really familiar with the compiler I got a bit stuck

I can try to check what f2py does, and do the same for pyccel

saidctb:
I also have this problem, after your last changes
```
(/u/saicctb/conda-envs/myenv) saicctb@cobra02:~> pyccel test_pyccel.py --verbose
  File "/u/saicctb/.local/bin/pyccel", line 33, in <module>
  File "/cobra/u/saicctb/pyccel/pyccel/commands/console.py", line 131, in pyccel
    from pyccel.codegen.pipeline  import execute_pyccel
  File "/cobra/u/saicctb/pyccel/pyccel/codegen/pipeline.py", line 24, in <module>
    from .compiling.compilers import Compiler
  File "/cobra/u/saicctb/pyccel/pyccel/codegen/compiling/compilers.py", line 42, in <module>
    different_version(next(iter(available_compilers.values()))):
  File "/cobra/u/saicctb/pyccel/pyccel/codegen/compiling/compilers.py", line 34, in different_version
    compiler['python_version'] != python_version
KeyError: 'python_version'
```

EmilyBourne:
> I also have this problem, after your last changes
> 
> ```
> (/u/saicctb/conda-envs/myenv) saicctb@cobra02:~> pyccel test_pyccel.py --verbose
> Traceback (most recent call last):
>   File "/u/saicctb/.local/bin/pyccel", line 33, in <module>
>     sys.exit(load_entry_point('pyccel', 'console_scripts', 'pyccel')())
>   File "/cobra/u/saicctb/pyccel/pyccel/commands/console.py", line 131, in pyccel
>     from pyccel.codegen.pipeline  import execute_pyccel
>   File "/cobra/u/saicctb/pyccel/pyccel/codegen/pipeline.py", line 24, in <module>
>     from .compiling.compilers import Compiler
>   File "/cobra/u/saicctb/pyccel/pyccel/codegen/compiling/compilers.py", line 42, in <module>
>     different_version(next(iter(available_compilers.values()))):
>   File "/cobra/u/saicctb/pyccel/pyccel/codegen/compiling/compilers.py", line 34, in different_version
>     compiler['python_version'] != python_version
> KeyError: 'python_version'
> ```

This is because you need to regenerate the json files. In the future when modifying these files we will change the pyccel version number so they are updated automatically. Here I can't do that so you need to delete all the json files in the folder pyccel/compilers/ so they are regenerated the next time you run pyccel

saidctb:
> > I also have this problem, after your last changes
> > ```
> > (/u/saicctb/conda-envs/myenv) saicctb@cobra02:~> pyccel test_pyccel.py --verbose
> > Traceback (most recent call last):
> >   File "/u/saicctb/.local/bin/pyccel", line 33, in <module>
> >     sys.exit(load_entry_point('pyccel', 'console_scripts', 'pyccel')())
> >   File "/cobra/u/saicctb/pyccel/pyccel/commands/console.py", line 131, in pyccel
> >     from pyccel.codegen.pipeline  import execute_pyccel
> >   File "/cobra/u/saicctb/pyccel/pyccel/codegen/pipeline.py", line 24, in <module>
> >     from .compiling.compilers import Compiler
> >   File "/cobra/u/saicctb/pyccel/pyccel/codegen/compiling/compilers.py", line 42, in <module>
> >     different_version(next(iter(available_compilers.values()))):
> >   File "/cobra/u/saicctb/pyccel/pyccel/codegen/compiling/compilers.py", line 34, in different_version
> >     compiler['python_version'] != python_version
> > KeyError: 'python_version'
> > ```
> 
> This is because you need to regenerate the json files. In the future when modifying these files we will change the pyccel version number so they are updated automatically. Here I can't do that so you need to delete all the json files in the folder pyccel/compilers/ so they are regenerated the next time you run pyccel

the problem is fixed, but I still have the same problem as before

EmilyBourne:
@saidctb 
> I'm testing this branch in the cluster, and I have this problem
> 
> ```
> gfortran -O3 -fPIC -Wno-unused-result -Wsign-compare -DNDEBUG -O2 -Wall -fPIC -O2 -isystem /u/saicctb/conda-envs/myenv
> /include -fPIC -O2 -isystem /u/saicctb/conda-envs/myenv/include -pthread -B /u/saicctb/conda-envs/myenv/compiler_compat 
> libpython3.9.a -pthread -B /u/saicctb/conda-envs/myenv/compiler_compat -shared -Wl,-rpath,/u/saicctb/conda-envs/myenv/lib 
> -Wl,-rpath-link,/u/saicctb/conda-envs/myenv/lib -Wl,-rpath,/u/saicctb/conda-envs/myenv/lib -Wl,-rpath-link,/u/saicctb/conda-
> envs/myenv/lib -I /cobra/u/saicctb/__pyccel__ -I /cobra/u/saicctb/__pyccel__ -I /cobra/u/saicctb/__pyccel__ -I /cobra/u/saicctb
> /__pyccel__ -I /cobra/u/saicctb/__pyccel__ -I /cobra/u/saicctb/__pyccel__ -I /cobra/u/saicctb/__pyccel__ -I /u/saicctb/conda-
> envs/myenv/include/python3.9 -I /u/saicctb/.local/lib/python3.9/site-packages/numpy/core/include -L /u/saicctb/conda-
> envs/myenv/lib -L /u/saicctb/conda-envs/myenv/lib -L /u/saicctb/conda-envs/myenv/lib -L /u/saicctb/conda-envs/myenv
> /lib/python3.9/config-3.9-x86_64-linux-gnu -lm /cobra/u/saicctb/__pyccel__/test_pyccel.o /cobra/u/saicctb/__pyccel__
> /bind_c_test_pyccel.o /cobra/u/saicctb/__pyccel__/test_pyccel_wrapper.o -o /cobra/u/saicctb/__pyccel__/test_pyccel.cpython-39-x86_64-linux-gnu.so
> 
> ERROR at shared library generation stage
> Traceback (most recent call last):
>   File "/u/saicctb/.local/bin/pyccel", line 33, in <module>
>     sys.exit(load_entry_point('pyccel', 'console_scripts', 'pyccel')())
>   File "/cobra/u/saicctb/pyccel/pyccel/commands/console.py", line 216, in pyccel
>     execute_pyccel(filename,
>   File "/cobra/u/saicctb/pyccel/pyccel/codegen/pipeline.py", line 370, in execute_pyccel
>     generated_filepath = create_shared_library(codegen,
>   File "/cobra/u/saicctb/pyccel/pyccel/codegen/python_wrapper.py", line 89, in create_shared_library
>     sharedlib_filepath = src_compiler.compile_shared_library(wrapper_compile_obj,
>   File "/cobra/u/saicctb/pyccel/pyccel/codegen/compiling/compilers.py", line 371, in compile_shared_library
>     self.run_command(cmd, verbose)
>   File "/cobra/u/saicctb/pyccel/pyccel/codegen/compiling/compilers.py", line 400, in run_command
>     raise RuntimeError(err_msg)
> RuntimeError: Failed to build module
> gfortran: error: libpython3.9.a: No such file or directory
> ```
> 
> did you have this problem before?
> also the flags are repeated many times

Hmm, I haven't had this problem. Normally it uses `-lpython3.9`. Can you have a look at the output of `sysconfig.get_config_vars()` and see what is in the key `BLDLIBRARY` (on my computer it is `-lpython3.8`). For me the only place in that dict where I see `python3.8.a` is in the key `LIBRARY`

The keys are also repeated lots when using f2py, but I have reduced it somewhat by avoiding repeating includes

saidctb:
I get this
```
>>> import sysconfig
>>> sysconfig.get_config_vars()['BLDLIBRARY']
'libpython3.9.a'
```

EmilyBourne:
> I get this
> 
> ```
> >>> import sysconfig
> >>> sysconfig.get_config_vars()['BLDLIBRARY']
> 'libpython3.9.a'
> ```

Could you try this command:
```
gfortran -O3 -fPIC -Wno-unused-result -Wsign-compare -DNDEBUG -O2 -Wall -fPIC -O2 -isystem /u/saicctb/conda-envs/myenv
/include -fPIC -O2 -isystem /u/saicctb/conda-envs/myenv/include -pthread -B /u/saicctb/conda-envs/myenv/compiler_compat 
-lpython3.9 -pthread -B /u/saicctb/conda-envs/myenv/compiler_compat -shared -Wl,-rpath,/u/saicctb/conda-envs/myenv/lib 
-Wl,-rpath-link,/u/saicctb/conda-envs/myenv/lib -Wl,-rpath,/u/saicctb/conda-envs/myenv/lib -Wl,-rpath-link,/u/saicctb/conda-
envs/myenv/lib -I /cobra/u/saicctb/__pyccel__ -I /cobra/u/saicctb/__pyccel__ -I /cobra/u/saicctb/__pyccel__ -I /cobra/u/saicctb
/__pyccel__ -I /cobra/u/saicctb/__pyccel__ -I /cobra/u/saicctb/__pyccel__ -I /cobra/u/saicctb/__pyccel__ -I /u/saicctb/conda-
envs/myenv/include/python3.9 -I /u/saicctb/.local/lib/python3.9/site-packages/numpy/core/include -L /u/saicctb/conda-
envs/myenv/lib -L /u/saicctb/conda-envs/myenv/lib -L /u/saicctb/conda-envs/myenv/lib -L /u/saicctb/conda-envs/myenv
/lib/python3.9/config-3.9-x86_64-linux-gnu -lm /cobra/u/saicctb/__pyccel__/test_pyccel.o /cobra/u/saicctb/__pyccel__
/bind_c_test_pyccel.o /cobra/u/saicctb/__pyccel__/test_pyccel_wrapper.o -o /cobra/u/saicctb/__pyccel__/test_pyccel.cpython-39-x86_64-linux-gnu.so
```

saidctb:
> > I get this
> > ```
> > >>> import sysconfig
> > >>> sysconfig.get_config_vars()['BLDLIBRARY']
> > 'libpython3.9.a'
> > ```
> 
> Could you try this command:
> 
> ```
> gfortran -O3 -fPIC -Wno-unused-result -Wsign-compare -DNDEBUG -O2 -Wall -fPIC -O2 -isystem /u/saicctb/conda-envs/myenv
> /include -fPIC -O2 -isystem /u/saicctb/conda-envs/myenv/include -pthread -B /u/saicctb/conda-envs/myenv/compiler_compat 
> -lpython3.9 -pthread -B /u/saicctb/conda-envs/myenv/compiler_compat -shared -Wl,-rpath,/u/saicctb/conda-envs/myenv/lib 
> -Wl,-rpath-link,/u/saicctb/conda-envs/myenv/lib -Wl,-rpath,/u/saicctb/conda-envs/myenv/lib -Wl,-rpath-link,/u/saicctb/conda-
> envs/myenv/lib -I /cobra/u/saicctb/__pyccel__ -I /cobra/u/saicctb/__pyccel__ -I /cobra/u/saicctb/__pyccel__ -I /cobra/u/saicctb
> /__pyccel__ -I /cobra/u/saicctb/__pyccel__ -I /cobra/u/saicctb/__pyccel__ -I /cobra/u/saicctb/__pyccel__ -I /u/saicctb/conda-
> envs/myenv/include/python3.9 -I /u/saicctb/.local/lib/python3.9/site-packages/numpy/core/include -L /u/saicctb/conda-
> envs/myenv/lib -L /u/saicctb/conda-envs/myenv/lib -L /u/saicctb/conda-envs/myenv/lib -L /u/saicctb/conda-envs/myenv
> /lib/python3.9/config-3.9-x86_64-linux-gnu -lm /cobra/u/saicctb/__pyccel__/test_pyccel.o /cobra/u/saicctb/__pyccel__
> /bind_c_test_pyccel.o /cobra/u/saicctb/__pyccel__/test_pyccel_wrapper.o -o /cobra/u/saicctb/__pyccel__/test_pyccel.cpython-39-x86_64-linux-gnu.so
> ```

it works when I use `-lpython3.9`

EmilyBourne:
@saidctb Is the problem now fixed if you regenerate the json files?

saidctb:
> @saidctb Is the problem now fixed if you regenerate the json files?

it works !

saidctb (approved):

saidctb:
```
(/u/saicctb/conda-envs/myenv) saicctb@cobra02:~/__pyccel__> pgfortran -help
Reading rcfile /cobra/u/system/soft/SLE_12/packages/x86_64/pgi/20.4/linux86-64-llvm/20.4/bin/.pgfortranrc

Overall switches:
-###                Display commands that would be executed
-c                  Stop after assemble; output in object file
-dryrun             Display commands that would be executed
-drystdinc          Display standard include directories and exit
-echo[=go|stop]     Echo the command line flags and stop (default) or continue (=go). This is useful when the compiler is invoked by a script.
-help[=groups|asm|debug|language|linker|opt|other|overall|phase|prepro|suffix|switch|target|variable]
                    Show compiler switches
-Minform=inform|warn|severe|fatal|[no]file
                    Set error level of messages to display
    [no]file        Print source file names as they are compiled
-Mkeepasm           Preserve intermediate assembly language file
-Mllvm[=3.5|3.8|3.9|4.0|5.0|6.0|7.0|8.0|9.0]
                    Use the llvm code generator
    3.5             Enable llvm 3.5 as default tool
    3.8             Enable llvm 3.8 as default tool
    3.9             Enable llvm 3.9 as default tool
    4.0             Enable llvm 4.0 as default tool
    5.0             Enable llvm 5.0 as default tool
    6.0             Enable llvm 6.0 as default tool
    7.0             Enable llvm 7.0 as default tool
    8.0             Enable llvm 8.0 as default tool
    9.0             Enable llvm 9.0 as default tool
-Mnollvm            Use the nollvm code generator
-o<outputfile>      Name the output file
-S                  Stop before assembling; output in .s file
-show               Show compiler configuration
-silent             Inhibit warning messages; same as -Minform=severe
-time               Print execution time for each compiler step
-V[<ver>]           Display or change compiler version
-v                  Display each command as it is run
-w                  Inhibit warning messages; same as -Minform=severe
-Wh,<arg>           Pass argument to F90 front end
-Wi,<arg>           Pass argument to IPA Linker
-Yi,<ipadir>        Change IPA tools directory

Optimization switches:
-fast               Common optimizations; includes -O2 -Munroll=c:1 -Mlre -Mautoinline
                    == -Mvect=simd -Mflushz -Mcache_align
-fPIC               Generate position independent code
-fpic               Generate position independent code
-KPIC               Generate position independent code
-Kpic               Generate position independent code
-M[no]autoinline[=maxsize:<n>|totalsize:<n>|nostatic]
                    Enable automatic function inlining in C/C++
    maxsize:<n>     Automatically inline functions size n and less
    totalsize:<n>   Limit automatic inlining to total size of n
    nostatic        Do not inline static functions without inline keyword
-Mcache_align       Align large objects on cache-line boundaries
-Mconcur[=allcores|[no]altcode:<n>|altreduction:<n>|[no]assoc|[no]cncall|[no]innermost|levels:<n>]
                    Generate parallel loops
    allcores        Use all available cores; specify at link time
    [no]altcode:<n> Execute alternate serial code if loop count is n or less
    altreduction:<n>
                    Execute alternate serial code if loop has a reduction
                    and count is n or less
    [no]assoc       Enable parallelization of loops with associative reductions
    [no]cncall      Assume loops containing calls are safe to parallelize
    [no]innermost   Enable parallelization of innermost loops
    levels:<n>      Parallelize loops nested at most n levels deep
-M[no]depchk        Check dependence relations for vector or parallel code
-M[no]dse           Dead store elimination
-Mextract=<size>|<func>|lib:<extlib>|name:<func>|size:<n>
                    Extract subprograms for inlining
    <size>          Set maximum function size to extract
    <func>          Extract function func
    lib:<extlib>    Store information in directory extlib
    name:<func>     Extract function func
    size:<n>        Limit size of extracted functions to n
-M[no]fma           Enable (disable) generation of FMA instructions
-M[no]idiom         Loop idiom recognition
-M[no]info[=all|accel|ccff|ftn|inline|intensity|ipa|loop|lre|mp|opt|par|pcast|pfo|stat|time|vect]
                    Generate informational messages about optimizations
    all             -Minfo=accel,inline,ipa,loop,lre,mp,opt,par,vect
    accel           Enable Accelerator information
    ccff            Append information to object file
    ftn             Enable Fortran-specific information
    inline          Enable inliner information
    intensity       Enable compute intensity information
    ipa             Enable IPA information
    loop            Enable loop optimization information
    lre             Enable LRE information
    mp              Enable OpenMP information
    opt             Enable optimizer information
    par             Enable parallelizer information
    pcast           Enable PCAST information
    pfo             Enable profile feedback information
    stat            Same as -Minfo=time
    time            Display time spent in compiler phases
    vect            Enable vectorizer information
-Minline[=lib:<inlib>|<maxsize>|<func>|except:<func>|name:<func>|maxsize:<n>|totalsize:<n>|smallsize:<n>|reshape]
                    Enable function inlining
    lib:<inlib>     Use extracted functions from inlib
    <maxsize>       Set maximum function size to inline
    <func>          Inline function func
    except:<func>   Do not inline function func
    name:<func>     Inline function func
    maxsize:<n>     Inline only functions smaller than n
    totalsize:<n>   Limit inlining to total size of n
    smallsize:<n>   Always inline functions smaller than n
    reshape         Allow inlining in Fortran even when array shapes do not match
    -Minline        Inline all functions that were extracted
-Minstrument[=functions]
                    Generate additional code for instrumentation; implies -Minfo=ccff -Mframe
    functions       Enable instrumentation of functions
    -Minstrument    Enable instrumentation of functions
-M[no]ipa[=[no]align|[no]arg|[no]cg|[no]const|except:<func>|fast|[no]force|[no]globals|inline:{<n>|except:<func>|nopfo|[no]reshape}|ipofile|jobs:<n>|[no]keepobj|[no]libc|[no]libinline|[no]libopt|[no]localarg|main:<func>|[no]ptr|[no]alias|[no]reaggregation|[no]f90ptr|[no]pure|quiet|required|safe:<name>|[no]safeall|[no]shape|summary|[no]vestigial|acc]
                    Enable Interprocedural Analysis
    [no]align       Recognize when targets of pointer dummy are aligned
    [no]arg         Remove arguments replaced by -Mipa=ptr,const
    [no]cg          Generate call graph information for pgicg tool
    [no]const       Interprocedural constant propagation
    except:<func>   Used with -Mipa=inline, don't inline func
    fast            Equivalant to -Mipa=align,arg,const,f90ptr,shape,libc,globals,localarg,alias,ptr,pure
    [no]force       Force all objects to recompile regardless of whether IPA information has changed
    [no]globals     Optimize references to globals
    inline          Automatically determine which functions to inline
     <n>            Limit to n levels of inlining
     except:<func>  Don't inline func
     nopfo          Ignore PFO frequency information
     [no]reshape    Inline with mismatched array shapes
    ipofile         Save IPO information in .ipo files
    jobs:<n>        Recompile n jobs in parallel
    [no]keepobj     Keep IPA-optimized object files
    [no]libc        Optimize calls to certain libc functions
    [no]libinline   Inline (don't inline) routines even from libraries; implies -Mipa=inline
    [no]libopt      Optimize (don't optimize) routines in libraries
    [no]localarg    -Mipa=arg plus externalizes local pointer targets
    main:<func>     Specify a function to serve as a global entry point; may appear multiple times; disables linking
    [no]ptr         Pointer disambiguation
    [no]alias       Pointer alias analysis; implies -Mipa=ptr
    [no]reaggregation
                    IPA guided structure reaggregation
    [no]f90ptr      Fortran 90 Pointer disambiguation
    [no]pure        Pure function detection
    required        Return error condition if IPA is inhibited
    safe:<name>     Declare a function or all functions in a library to be side-effect free
    [no]safeall     Allow unsafe optimization by assuming unknown function references are unimportant
    [no]shape       Fortran 90 shape propagation
    summary         Only collect IPA summary information; prevents IPA optimization for these files
    [no]vestigial   Remove functions that are never called
    acc             Automatically mark OpenACC Routine Seq and generate device code
    -Mipa           Default enables constant propagation
-M[no]lre[=[no]assoc]
                    Loop-carried redundancy elimination
    [no]assoc       Allow expression reassociation
-M[no]movnt         Force generation of nontemporal moves
-Mneginfo[=all|accel|ftn|inline|ipa|loop|lre|mp|opt|par|pfo|vect]
                    Generate informational messages about optimizations
    all             -Mneginfo=accel,inline,ipa,loop,lre,mp,opt,par,vect
    accel           Enable Accelerator information
    ftn             Enable Fortran-specific information
    inline          Enable inliner information
    ipa             Enable IPA information
    loop            Enable loop optimization information
    lre             Enable LRE information
    mp              Enable OpenMP information
    opt             Enable optimizer information
    par             Enable parallelizer information
    pfo             Enable profile feedback information
    vect            Enable vectorizer information
-Mnoopenmp          Ignore OpenMP directives; use with -mp
-Mnotarget_temps    Disable a temporary when passing an array for a callee assumed-shape variable with the target attribute
-Mnovintr           Do not generate vector intrinsic calls
-[no]mp[=[no]align|bind]
                    Enable OpenMP (implies -Mrecursive in Fortran)
    [no]align       Modify default loop scheduling to prefer aligned array references
    bind            Bind threads to cores; specify at link time
-Mpfi[=[no]indirect]
                    Enable profile-feedback instrumentation
    [no]indirect    Save [don't save] indirect function call targets
-Mpfo[=[no]indirect|[no]layout|dir:<dir>]
                    Enable profile-feedback optimizations
    [no]indirect    Enable [disable] indirect function call inlining
    [no]layout      Enable [disable] dynamic code layout
    dir:<dir>       Specify directory containing pgfi.out file
-M[no]pre           Enable partial redundancy elimination
-M[no]prefetch[=distance:<d>|n:<n>|nta|plain|t0|w]
                    Enable prefetch instructions
    distance:<d>    Set prefetch-ahead distance in cache lines
    n:<n>           Set maximum number of prefetches to generate
    nta             Use prefetchnta
    plain           Use prefetch
    t0              Use prefetcht0
    w               Allow prefetchw
-Mprof[=[no]ccff|dwarf]
                    Generate additional code for profiling; implies -Minfo=ccff
    [no]ccff        Enable (disable) CCFF information
    dwarf           Add limited DWARF info sufficient for performance profilers
-M[no]propcond      Enable constant propagation from assertions derived from equality conditionals
-Msafe_lastval      Allow parallelization of loops with conditional scalar assignments
-M[no]stack_arrays  Place automatic arrays on the stack
-M[no]target_temps  Enable [disable] a temporary when passing an array for a callee assumed-shape variable with the target attribute
-M[no]unroll[=c:<n>|n:<n>|m:<n>]
                    Enable loop unrolling
    c:<n>           Completely unroll loops with loop count n or less
    n:<n>           Unroll single-block loops n times
    m:<n>           Unroll multi-block loops n times
    -Munroll        Unroll loops with a loop count of 1
-M[no]unsafe_par_align
                    Generate aligned moves in parallelized loops if first element is aligned
-M[no]vect[=[no]altcode|[no]assoc|cachesize:<c>|[no]fuse|[no]gather|[no]idiom|levels:<n>|nocond|[no]partial|prefetch|[no]short|[no]simd[:{128|256}]|[no]simdresidual|[no]sizelimit[:n]|[no]sse|[no]tile]
                    Control automatic vector pipelining
    [no]altcode     Generate appropriate alternative code for vectorized loops
    [no]assoc       Allow [disallow] reassociation
    cachesize:<c>   Optimize for cache size c
    [no]fuse        Enable [disable] loop fusion
    [no]gather      Enable [disable] vectorization of indirect array references
    [no]idiom       Enable [disable] idiom recognition
    levels:<n>      Maximum nest level of loops to optimize
    nocond          Disable vectorization of loops with conditionals
    [no]partial     Enable [disable] partial loop vectorization via inner loop distribution
    prefetch        Generate prefetch instructions
    [no]short       Enable [disable] short vector operations
    [no]simd[:{128|256}]
                    Generate [don't generate] SIMD instructions
     128            Use 128-bit SIMD instructions
     256            Use 256-bit SIMD instructions
     512            Use 512-bit SIMD instructions
    [no]simdresidual
                    Enable [disable] vectorization of the residual loop of a vectorized loop
    [no]sizelimit[:n]
                    Limit size of vectorized loops
    [no]sse         The [no]sse option is deprecated, use [no]simd instead.
    [no]tile        Enable [disable] loop tiling
-O                  Set opt level. All -O1 optimizations plus traditional global scalar optimizations performed
-O0                 Set opt level. Creates a basic block for each statement. No scheduling or global optimizations performed
-O1                 Set opt level. Some scheduling and register allocation is enabled. No global optimizations performed
-O2                 Set opt level. All -O optimizations plus SIMD code generation
                    == -Mvect=simd
-O3                 Set opt level. All -O2 optimizations plus more aggressive code hoisting and scalar replacement, that may or may not be profitable, performed
                    == -Mvect=simd
-O4                 Set opt level. All -O3 optimizations plus more aggressive hoisting of guarded expressions performed
                    == -Mvect=simd
-pg                 Enable gprof profiling; implies -Mframe

Debugging switches:
-C                  Generate code to check array bounds
-g                  Generate information for debugger
-gopt               Generate information for debugger without disabling optimizations
-M[no]bounds        Generate code to check array bounds
-Mchkfpstk          Check consistency of floating point stack at subprogram calls (32-bit only)
-Mchkptr            Check for NULL pointers
-Mchkstk            Check for sufficient stack space upon subprogram entry
-Mcoff              Generate COFF format object
-Mdwarf3            Generate DWARF3 debug information with -g
-Melf               Generate ELF format object
-Memit-dwarf-common-blocks-name
                    Enable generation of DWARF name for Fortran COMMON blocks
-Memit-dwarf-inlined
                    Enable generation of DWARF for inlined functions
-Mnodwarf           Do not add DWARF debug information
-[no]traceback      Add debug information for runtime traceback

Preprocessor switches:
-cpp                == -Mpreprocess
-D<macro>           Define a preprocessor macro
-dD                 (C only) Print macros and values from source files
-dI                 (C only) Print include file names
-dM                 (C only) Print macros and values, including predefined and command-line macros
-dN                 (C only) Print macro names from source files
-E                  Stop after preprocessor; print output on standard output
-F                  Stop after preprocessing, save output in .f file
-I<incdir>          Add directory to include file search path
-Mnostddef          Do not use standard macro definitions
-Mnostdinc          Do not use standard include directories
-Mpreprocess        Run preprocessor for assembly and Fortran files
-U<macro>           Undefine a preprocessor macro

Assembler switches:
-Wa,<arg>           Pass argument to assembler

Linker switches:
-acclibs            Append Accelerator libraries to the link line
--[no-]as-needed    Passed to linker; only set DT_NEEDED for the following shared libraries if they are used
-Bdynamic           Passed to linker; specify dynamic binding
-Bstatic            Passed to linker; specify static binding
-Bstatic_nvrtl      Use to link static compiler runtime libraries with dynamic system libraries; implies -Mnorpath
-Bsymbolic          Passed to linker; specify symbolic binding
-c++libs            Append gnu compatible C++ libraries to the link line
-cudalibs           Link with CUDA-enabled libraries
-fortranlibs        Append pgf90 libraries to the link line
-g77libs            Include g77 or gfortran library when linking
-L<libdir>          Passed to linker; Add directory to library search path
-l<lib>             Passed to linker; Add library name to library search list
-m                  Passed to linker; display link map
-Mcudalib[=cublas|cufft|curand|cusolver|cusparse|cutensor]
                    Add appropriate versions of the CUDA-optimized libraries
-M[no]eh_frame      Add link flags to preserve exception-handling frame information
-Mhugetlb           Link with the huge TLB page library
-Mlfs               Link with library directory for large file support
-Mmkl               Add Intel's Math Kernel libraries to the link line. Requires that the environment variable MKLROOT be set to the path to the local MKL installation--the part of the path that precedes 'lib/<arch>'.
-Mmpi=mpich|sgimpi|mpich1|mpich2|mvapich1
                    Use default or specified MPI communication libraries
    mpich           Use default or MPIDIR-specified MPICH v3.0 libraries
    sgimpi          Use default or MPI_ROOT-specified SGI MPI libraries
    mpich1          DEPRECATED: Use MPIDIR to specify MPICH1 libraries
    mpich2          DEPRECATED: Use MPIDIR to specify MPICH2 libraries
    mvapich1        DEPRECATED: Use MPIDIR to specify MVAPICH1 libraries
-Mnorpath           Don't add -rpath paths to link line
-Mnostartup         Do not use standard linker startup file
-Mnostdlib          Do not use standard linker libraries
-Mscalapack         Add Scalapack libraries
-R<ldarg>           Passed to linker; just link symbols from object, or add directory to run time search path
-r                  Generate relocatable object; passed to linker
-rpath <dir>        Passed to linker; add directory to runtime shared library search path
-s                  Passed to linker; strip symbol table from exectable
-shared             Used to produce shared libraries; implies -fpic
-soname <soname>    Passed to linker; sets DT_SONAME field to the specified name
-u<undef>           Passed to linker; generate undefined reference
--[no-]whole-archive
                    Passed to linker; includes all objects in subsequent archives
-Wl,<arg>           Pass argument to linker
-Xlinker <arg>      Pass argument to linker

Language-specific switches:
-byteswapio         Swap byte-order for unformatted input/output
-FSUFFIX=<suff>     Define a Fortran suffix requiring preprocessing
-fsuffix=<suff>     Define a Fortran suffix
-i2                 Treat INTEGER as INTEGER*2
-i4                 Treat INTEGER as INTEGER*4
-i8                 Treat INTEGER and LOGICAL as INTEGER*8 and LOGICAL*8
-i8storage          INTEGER and LOGICAL occupy 8 bytes
-Mallocatable=95|03 Select semantics for assignments to allocatables
    95              Select Fortran 1995 semantics
    03              Select Fortran 2003 semantics
-M[no]backslash     Treat backslash like any other character in character strings
-M[no]builtin       Use built-in functions
-Mbyteswapio        Swap byte-order for unformatted input/output
-Mcontiguous        Assume POINTERs are contiguous
-Mcray=pointer      Set Cray pointer semantics
-Mcuda[=cc30|cc35|cc50|cc60|cc62|cc70|cc72|cc75|ccall|cudaX.Y|[no]debug|[no]lineinfo|fastmath|[no]flushz|keepgpu|keepbin|keepptx|madconst|maxregcount:<n>|charstring|[no]fma|loadcache:{L1|L2}|[no]unroll|ptxinfo|[no]rdc]
                    Enable CUDA Fortran
    cc30            Compile for compute capability 3.0
    cc35            Compile for compute capability 3.5
    cc50            Compile for compute capability 5.0
    cc60            Compile for compute capability 6.0
    cc62            Compile for compute capability 6.2
    cc70            Compile for compute capability 7.0
    cc72            Compile for compute capability 7.2
    cc75            Compile for compute capability 7.5
    ccall           Compile for all supported compute capabilities
    cudaX.Y         Use CUDA X.Y Toolkit compatibility, where installed
    [no]debug       Generate GPU debug information
    [no]lineinfo    Generate GPU line information
    fastmath        Use fast math library
    [no]flushz      Enable flush-to-zero mode on the GPU
    keepgpu         Keep kernel source files
    keepbin         Keep CUDA binary files
    keepptx         Keep PTX portable assembly files
    madconst        Put Module Array Descriptors in CUDA Constant Memory
    maxregcount:<n> Set maximum number of registers to use on the GPU
    charstring      Enable limited support for character strings in GPU kernels
    [no]fma         Generate fused mul-add instructions (default at -O3)
    loadcache       Choose what hardware level cache to use for global memory loads
     L1             Use L1 cache
     L2             Use L2 cache
    [no]unroll      Enable automatic inner loop unrolling (default at -O3)
    ptxinfo         Print informational messages from PTXAS
    [no]rdc         Generate relocatable device code
-M[no]dclchk        Check that all variables are declared
-M[no]defaultunit   Treat '*' unit as standard input/output
-M[no]dlines        Compile lines beginning with the letter D
-Mdollar=<char>     Map dollar sign in identifiers to this character
-Mextend            Allow source lines up to 132 characters
-Mfixed             Assume fixed-format source
-M[no]free[form]    Assume free-format source
-M[no]i4            Treat INTEGER as INTEGER*4
-M[no]iomutex       Generate critical sections around Fortran input/output
-M[no]llalign       Require 8 byte alignment for long long
-Mnomain            Don't include Fortran main program object module
-module <dir>       Add search directory for module information files
-M[no]onetrip       Force DO loops to execute at least once
-M[no]r8            Promote REAL variables and constants to DOUBLE PRECISION
-M[no]r8intrinsics[=float]
                    Treat REAL,CMPLX intrinsics as DBLE,DCMPLX
    float           Treat FLOAT intrinsic as DBLE as well
-M[no]recursive     Generate code to allow recursive subprograms
-M[no]ref_externals Force references to any EXTERNAL subprograms
-M[no]save          Assume all variables have SAVE attribute
-M[no]signextend    Extend the sign bit when converting shorter to longer integer
-Mstandard          Check standard conformance
-M[no]unixlogical   Treat any nonzero logical value as .TRUE.
-M[no]upcase        Preserve upper and lower case in identifiers; Fortran keywords must be lower case
--no_preincludes    Ignore all preincluded files: used for compiling preprocessed files
-r4                 Treat REAL as REAL*4
-r8                 Treat REAL as REAL*8

Target-specific switches:
-acc[=[no]autopar|[no]routineseq|legacy|strict|verystrict|sync|[no]wait]
                    Enable OpenACC directives
    [no]autopar     Enable (default) or disable loop autoparallelization within acc parallel
    [no]routineseq  Compile every routine for the device
    legacy          Suppress warnings about deprecated PGI accelerator directives
    strict          Issue warnings for non-OpenACC accelerator directives
    verystrict      Fail with an error for any non-OpenACC accelerator directive
    sync            Ignore async clauses
    [no]wait        Wait for each device kernel to finish
-K[no]ieee          Use IEEE division, optionally enable traps
-Ktrap=divz|denorm|fp|inexact|inv|none|ovf|unf
                    Determine IEEE Trap conditions
-m[no-]3dnow        Enable [disable] 3DNow! instructions
-m[no-]3dnowa       Enable [disable] 3DNow! Athlon instructions
-m[no-]abm          Enable [disable] ABM instructions
-m[no-]aes          Enable [disable] AES instructions
-m[no-]avx          Enable [disable] AVX instructions
-m[no-]avx2         Enable [disable] AVX2 instructions
-m[no-]avx512bitalg Enable [disable] AVX-512 bit algorithms
-m[no-]avx512bw     Enable [disable] AVX-512 byte and word instructions
-m[no-]avx512cd     Enable [disable] AVX-512 conflict detection instructions
-m[no-]avx512dq     Enable [disable] AVX-512 doubleword and quadword instructions
-m[no-]avx512er     Enable [disable] AVX-512 exponential and reciprocal instructions
-m[no-]avx512f      Enable [disable] AVX-512 instructions
-m[no-]avx512ifma   Enable [disable] AVX-512 integer fused multiply-add instructions
-m[no-]avx512pf     Enable [disable] AVX-512 prefetch instructions
-m[no-]avx512vbmi   Enable [disable] AVX-512 vector byte manipulation instructions
-m[no-]avx512vbmi2  Enable [disable] AVX-512 further vector byte manipulation instructions
-m[no-]avx512vl     Enable [disable] AVX-512 vector length extensions
-m[no-]avx512vpopcntdq
                    Enable [disable] AVX-512 population count instructions
-m[no-]bmi          Enable [disable] BMI instructions
-m[no-]bmi2         Enable [disable] BMI2 instructions
-m[no-]clflushopt   Enable [disable] optimized cache line flush instructions
-m[no-]clzero       Enable [disable] cache line zero instructions
-mcmodel=small|medium
                    Set the memory model
    small           Small model, total memory size < 2GB
    medium          Medium model, total memory size > 2GB, object size < 2GB; implies -Mlarge_arrays
-M[no]daz           Treat denormalized numbers as zero
-m[no-]f16c         Enable [disable] 16-bit floating-point conversion instructions
-M[no]flushz        Set SSE to flush-to-zero mode
-m[no-]fma          Enable [disable] three-operand fused multiply-add
-m[no-]fma4         Enable [disable] four-operand fused multiply-add
-M[no]fpapprox[=div|sqrt|rsqrt]
                    Perform certain fp operations using low-precision approximation
    div             Approximate floating point division
    sqrt            Approximate floating point square root
    rsqrt           Approximate floating point reciprocal square root
    -Mfpapprox      Approximate div,sqrt,rsqrt
-M[no]fpmisalign    Allow use of vector arithmetic instructions for unaligned operands
-M[no]fprelaxed[=[no]div|[no]intrinsic|[no]recip|[no]sqrt|[no]rsqrt|[no]order]
                    Perform certain fp operations using relaxed precision
    [no]div         Perform divide with relaxed precision
    [no]intrinsic   Perform intrinsics with relaxed precision
    [no]recip       Perform reciprocal with relaxed precision
    [no]sqrt        Perform square root with relaxed precision
    [no]rsqrt       Perform reciprocal square root with relaxed precision
    [no]order       Allow expression reordering, including factoring
    -Mfprelaxed     -Mfprelaxed=div,rsqrt,sqrt (sqrt only for Intel targets)
-m[no-]fsgsbase     Enable [disable] FS/GS base instructions
-M[no]func32        Align functions on 32 byte boundaries
-m[no-]fxsr         Enable [disable] fxsave/fxrestore instructions
-Mgccbug[s]         Match behavior of certain gcc bugs
-m[no-]gfni         Enable [disable] Galois field arithmetic instructions
-M[no]large_arrays  Support for large arrays
-M[no]longbranch    Enable long branches
-m[no-]lwp          Enable [disable] LWP instructions
-m[no-]lzcnt        Enable [disable] LZCNT instructions
-m[no-]mmx          Enable [disable] MMX instructions
-m[no-]movdir64b    Enable [disable] MOVDIR64B instructions
-m[no-]movdiri      Enable [disable] MOVDIRI instructions
-m[no-]mpx          Enable [disable] MPX instructions
-m[no-]mwaitx       Enable [disable] MONITOR/MWAITX timer functionality
-m[no-]pclmul       Enable [disable] packed carry-less multiplication instructions
-m[no-]pku          Enable [disable] protection keys
-m[no-]popcnt       Enable [disable] population count instructions
-m[no-]prefetchwt1  Enable [disable] prefetch with intent to write and T1 hint instructions
-m[no-]rdrnd        Enable [disable] RDRND instructions
-M[no]reg_struct_return
                    Return small struct/union values in registers (32-bit only)
-m[no-]rtm          Enable [disable] RTM instructions
-M[no]second_underscore
                    Add second underscore to global symbols that already contain an underscore
-m[no-]sha          Enable [disable] SHA instructions
-m[no-]sse          Enable [disable] SSE instructions
-m[no-]sse2         Enable [disable] SSE2 instructions
-m[no-]sse3         Enable [disable] SSE3 instructions
-m[no-]sse4         Enable [disable] SSE 4.2 instructions
-m[no-]sse4.1       Enable [disable] SSE 4.1 instructions
-m[no-]sse4.2       Enable [disable] SSE 4.2 instructions
-m[no-]sse4a        Enable [disable] SSE 4a instructions
-m[no-]ssse3        Enable [disable] SSSE3 instructions
-m[no-]tbm          Enable [disable] TBM instructions
-m[no-]vaes         Enable [disable] promotion of selected AES instructions to AVX512/AVX registers
-M[no]varargs       Calls from Fortran to C will use varargs calling sequence
-m[no-]vpclmulqdq   Enable [disable] VPCLMULQDQ instructions
-Mwritable-constants
                    Store constants in the writable data segment
-Mwritable-strings  Store string constants in the writable data segment
-m[no-]xop          Enable [disable] XOP instructions
-m[no-]xsave        Enable [disable] XSAVE instructions
-m[no-]xsavec       Enable [disable] XSAVEC instructions
-m[no-]xsaveopt     Enable [disable] XSAVEOPT instructions
-m[no-]xsaves       Enable [disable] XSAVES instructions
-pc=32|64|80        Precision control in x87 stack
-ta=host|multicore|tesla:{cc30|cc35|cc50|cc60|cc62|cc70|cc72|cc75|ccall|cudaX.Y|fastmath|[no]flushz|[no]fma|keep|[no]lineinfo|llc|zeroinit|deepcopy|loadcache:{L1|L2}|maxregcount:<n>|pinned|[no]rdc|safecache|[no]unroll|managed|beta|autocompare|redundant}
                    Choose target accelerator
    host            Compile for serial execution on the host CPU
    multicore       Compile for parallel execution on the host CPU
    tesla           Compile for parallel execution on a Tesla GPU
     cc30           Compile for compute capability 3.0
     cc35           Compile for compute capability 3.5
     cc50           Compile for compute capability 5.0
     cc60           Compile for compute capability 6.0
     cc62           Compile for compute capability 6.2
     cc70           Compile for compute capability 7.0
     cc72           Compile for compute capability 7.2
     cc75           Compile for compute capability 7.5
     ccall          Compile for all supported compute capabilities
     cudaX.Y        Use CUDA X.Y Toolkit compatibility, where installed
     fastmath       Use fast math library
     [no]flushz     Enable flush-to-zero mode on the GPU
     [no]fma        Generate fused mul-add instructions (default at -O3)
     keep           Keep kernel files
     [no]lineinfo   Generate GPU line information
     zeroinit       Initialize allocated device memory with zero
     deepcopy       Enable Full Deepcopy support in OpenACC Fortran
     loadcache      Choose what hardware level cache to use for global memory loads
      L1            Use L1 cache
      L2            Use L2 cache
     maxregcount:<n>
                    Set maximum number of registers to use on the GPU
     pinned         Use CUDA Pinned Memory
     [no]rdc        Generate relocatable device code
     safecache      Allows variable-sized array sections in cache directives and assumes they fit into CUDA shared memory
     [no]unroll     Enable automatic inner loop unrolling (default at -O3)
     managed        Use CUDA Managed Memory
     beta           Enable beta code generation features
     autocompare    Automatically compare CPU/GPU results: implies redundant
     redundant      Redundant CPU/GPU execution
-tp=px|bulldozer|piledriver|zen|sandybridge|haswell|knl|skylake
                    Select target processor
    px              Generic x86 Processor
    bulldozer       AMD Bulldozer processor
    piledriver      AMD Piledriver processor
    zen             AMD Zen architecture (Epyc, Ryzen)
    sandybridge     Intel SandyBridge processor
    haswell         Intel Haswell processor
    knl             Intel Knights Landing processor
    skylake         Intel Skylake Xeon processor
```
these are the flags provided by pgi

rakati (approved):

saidctb (approved):

yguclu (approved):

Issue #458, opened 2020-10-05
Numpy arrays


yguclu:
As a USER, I would like to accelerate any Numpy array expression using Pyccel.

PR #939, opened 2021-08-13
Add transpose function

EmilyBourne:
Implement the `transpose` function. In fortran when used as a temporary variable the `transpose` function is used. When the variable is saved, the ordering is used to ensure pointers are created. In C the function is handled through unravelling or with a new function `transpose_alias_assign`.

The `transpose` function is also used to implement the ndarray property `.T` and method `.transpose()`

Related to #806 and #458


nhamidn:
i tried this python code:

```
from numpy.random import randint
import numpy as np


def f1(x : 'int[:,:]', out : 'int[:,:]'):
    from numpy import transpose
    y = transpose(x)
    print(y)
    for i in range(len(y)):
        for j in range(len(y[i])):
            out[i][j] = y[i][j]

x1 = np.array([[1, 2, 3],
                [4, 5, 6],
                [7, 8, 9]])

result_fo = np.zeros([3,3], dtype='int')

f1_epyc = epyccel(f1)
f1_epyc(x1, result_fo)
print(result_fo)
```
and this is what I got as a result:

```
  1                    2                    3                    4                    5                    6                    7                    8                    9

[[1 4 7]
 [2 5 8]
 [3 6 9]]
```
  
i think there is a problem with the ordering

EmilyBourne:
@nhamidn 
> i tried this python code:
> ...
> i think there is a problem with the ordering 

Can you confirm that this is a problem caused by the transpose function and not a poor implementation of the print function?

Ie what does the following give:

```
from numpy.random import randint
import numpy as np


def f1(x : 'int[:,:](order=F)', out : 'int[:,:]'):
    from numpy import transpose
    print(x)
    y = transpose(x)
    print(y)
    for i in range(len(y)):
        for j in range(len(y[i])):
            out[i][j] = y[i][j]

x1 = np.array([[1, 2, 3],
                [4, 5, 6],
                [7, 8, 9]])

result_fo = np.zeros([3,3], dtype='int')

f1_epyc = epyccel(f1)
f1_epyc(x1, result_fo)
print(result_fo)
```

EmilyBourne:
> [[1 4 7]
> [2 5 8]
> [3 6 9]]

This bit seems correct

nhamidn:
> @nhamidn
> 
> > i tried this python code:
> > ...
> > i think there is a problem with the ordering
> 
> Can you confirm that this is a problem caused by the transpose function and not a poor implementation of the print function?
> 
> Ie what does the following give:
> 
> ```
> from numpy.random import randint
> import numpy as np
> from pyccel.epyccel import epyccel
> 
> 
> def f1(x : 'int[:,:](order=F)', out : 'int[:,:]'):
>     from numpy import transpose
>     print(x)
>     y = transpose(x)
>     print(y)
>     for i in range(len(y)):
>         for j in range(len(y[i])):
>             out[i][j] = y[i][j]
> 
> x1 = np.array([[1, 2, 3],
>                 [4, 5, 6],
>                 [7, 8, 9]])
> 
> result_fo = np.zeros([3,3], dtype='int', order='F')
> 
> f1_epyc = epyccel(f1)
> f1_epyc(x1, result_fo)
> print(result_fo)
> ```

the print seems to be normal : 

this is the generated foortran code :

```
module trans

  implicit none

  contains

  subroutine f1(x, out)

    implicit none

    integer(C_INT64_T), intent(in), target :: x(0:,0:)
    integer(C_INT64_T), intent(inout) :: out(0:,0:)
    integer(C_INT64_T), pointer :: y(:,:)
    integer(C_INT64_T) :: i
    integer(C_INT64_T) :: j

    y(0:, 0:) => x
    print *, x
    print *, y
    do i = 0_C_INT64_T, size(y,1,C_INT64_T) - 1_C_INT64_T, 1_C_INT64_T
      do j = 0_C_INT64_T, size(y(i, :), kind=C_INT64_T) - 1_C_INT64_T, &
      1_C_INT64_T
        out(j, i) = y(i, j)
      end do
    end do

  end subroutine f1

end module trans


```

EmilyBourne:
> > @nhamidn
> > > i tried this python code:
> > > ...
> > > i think there is a problem with the ordering
> > 
> > 
> > Can you confirm that this is a problem caused by the transpose function and not a poor implementation of the print function?
> > Ie what does the following give:
> > ```
> > ...
> > ```
> 
> the print seems to be normal :
> 
> this is the generated foortran code :
> 
> ```
> ...
> ```

This is not a normal print for fortran ordering. If you run this code you will see that the output for `print *, x` does not match what you would expect. I have opened issue #940 about this

EmilyBourne:
@nhamidn 
> This is not a normal print for fortran ordering. If you run this code  you will see that the output for `print *, x` does not match what you would expect

Note that I made a mistake in my code. The `order='F'` argument should be passed to the input array `x1` not the output `result_fo`, I have edited my comment above

EmilyBourne:
@nhamidn To test the transpose, before #940 is fixed you can print the array row-wise manually to check the output. Ie:
```
from numpy.random import randint
import numpy as np


def f1(x : 'int[:,:]', out : 'int[:,:]'):
    from numpy import transpose
    y = transpose(x)
    for yi in y:
        print(yi)
    for i in range(len(y)):
        for j in range(len(y[i])):
            out[i][j] = y[i][j]

x1 = np.array([[1, 2, 3],
                [4, 5, 6],
                [7, 8, 9]])

result_fo = np.zeros([3,3], dtype='int')

f1_epyc = epyccel(f1)
f1_epyc(x1, result_fo)
print(result_fo)
```

nhamidn (approved):

saidctb (approved):

yguclu (approved):

yguclu:
@EmilyBourne How do you handle the case of Numpy arrays with rank > 2? Fortran's `transpose` function only works for 2D arrays, so I expect that `reshape` should be used instead of `transpose`...

EmilyBourne:
> @EmilyBourne How do you handle the case of Numpy arrays with rank > 2? Fortran's `transpose` function only works for 2D arrays, so I expect that `reshape` should be used instead of `transpose`...

I am currently using `transpose`. I wasn't sure about this but as the tests were passing I assumed it was ok. I will change it

yguclu:
> > @EmilyBourne How do you handle the case of Numpy arrays with rank > 2? Fortran's `transpose` function only works for 2D arrays, so I expect that `reshape` should be used instead of `transpose`...
> 
> I am currently using `transpose`. I wasn't sure about this but as the tests were passing I assumed it was ok. I will change it

Indeed I am puzzled by the fact that the 3D test works... We need to understand what is going on there!

EmilyBourne:
@yguclu 
> Indeed I am puzzled by the fact that the 3D test works... We need to understand what is going on there!

It is strange. [gfortran](https://gcc.gnu.org/onlinedocs/gfortran/TRANSPOSE.html) doesn't specify what happens if you pass something that is not 3d, so maybe they have implemented this to be helpful?

EmilyBourne:
> Fortran's `transpose` function only works for 2D arrays, so I expect that `reshape` should be used instead of `transpose`...

Fixed 018c0a1

yguclu (approved):

EmilyBourne:
> > @yguclu
> Possibly. Do you have access to an Intel or PGI compiler? At least one of them should raise an error...

I don't have a PGI compiler, but all the tests pass if I use `compiler='intel'` (the intel compiler is correctly called)

EmilyBourne:
> @EmilyBourne How do you handle the case of Numpy arrays with rank > 2? Fortran's `transpose` function only works for 2D arrays, so I expect that `reshape` should be used instead of `transpose`...

Hmm, it seems the tests are passing as there is no transpose required in most cases, it just changes the order of the resulting array. And in the case of an expression the transpose is often unrolled anyway. I have added a test case to force the use of the transpose function and it seems that the transpose is not correctly managed

EmilyBourne:
> Do you handle the method a.transpose() in addition to the property a.T?

No, I will add this

> Can the user pass the *axes parameters to the function and method?

No, I wanted to add this, but in python I believe it is possible to do this without necessarily allocating memory. This is not possible in Fortran unless we change all our element-wise access methods to carry `axes` instead of `order` (with `order='C'` equivalent to `(0,1,2,3,...)` and `order='F'` equivalent to `(rank-1,rank-2,,...)` ) so I thought it was probably complicated enough to leave for another PR.

yguclu:
> > Can the user pass the *axes parameters to the function and method?
> 
> No, I wanted to add this, but in python I believe it is possible to do this without necessarily allocating memory. This is not possible in Fortran unless we change all our element-wise access methods to carry `axes` instead of `order` (with `order='C'` equivalent to `(0,1,2,3,...)` and `order='F'` equivalent to `(rank-1,rank-2,,...)` ) so I thought it was probably complicated enough to leave for another PR.

I thought that even in Numpy a new array was created in all cases where it is not sufficient to change the ordering... I am going to check this.

EmilyBourne:
> I thought that even in Numpy a new array was created in all cases where it is not sufficient to change the ordering... I am going to check this.

No, as long as it can be handled by modifying the `steps` array numpy (and c for us) does not need to create a new array. E.g:
```
> import numpy as np
> x = np.random.randint(50, size=(3,2,4))
> y=x.transpose((1,2,0))
> y.flags
  C_CONTIGUOUS : False
  F_CONTIGUOUS : False
  OWNDATA : False
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False
```

yguclu:
> > I thought that even in Numpy a new array was created in all cases where it is not sufficient to change the ordering... I am going to check this.
> 
> No, as long as it can be handled by modifying the `steps` array numpy (and c for us) does not need to create a new array. E.g:
> 
> ```
> > import numpy as np
> > x = np.random.randint(50, size=(3,2,4))
> > y=x.transpose((1,2,0))
> > y.flags
>   C_CONTIGUOUS : False
>   F_CONTIGUOUS : False
>   OWNDATA : False
>   WRITEABLE : True
>   ALIGNED : True
>   WRITEBACKIFCOPY : False
>   UPDATEIFCOPY : False
> ```

You're right. Upon calling `transpose` Numpy recomputes the strides and stores them in the array struct.

yguclu:
For now I suggest to add the `transpose` method, and raise a neat error if the user passes the `*axes` parameters.

EmilyBourne:
> For now I suggest to add the `transpose` method, and raise a neat error if the user passes the `*axes` parameters.

Done. Should I open an issue for the `*axes` parameter?

yguclu:
@EmilyBourne I think the PR description should be expanded a bit (the method `transpose` and the property `T` should be mentioned). After that this can be merged.

Issue #856, opened 2021-04-14
Define shortcuts to make literals more readable


EmilyBourne:
**Describe the bug**
As mentioned in issues #487 and #604, our method for setting the precision of literals is quite verbose. It would be preferable to provide shorter more user friendly names

**Example**
The following code:
```python
import numpy as np
a = np.int32(1) + 2 + 3.5
```
currently generates:
```
program prog_tmp

use, intrinsic :: ISO_C_BINDING

implicit none

real(C_DOUBLE) :: a

a = 1_C_INT32_T + 2_C_INT64_T + 3.5_C_DOUBLE

end program prog_tmp
```
a more readable format for the line containing the equation would be:
```
a = 1_i32 + 2_i64 + 3.5_f64
```

**Expected behaviour**
As suggested in issue #604 
> rename C_LONG_LONG with
> 
> `use ISO_C_BINDING, only: i64 => C_LONG_LONG`
> 
> _Originally posted by @Beliavsky in https://github.com/pyccel/pyccel/issues/604#issue-771030427_

If the name `i64` is already used in the code then we should default back to the more verbose `C_LONG_LONG`

**Language**
Fortran

Issue #919, opened 2021-07-08
float128 support


AbdelGhafour69:
**What I would like to be added**

I would like to see numpy's <code>float128</code> added to Pyccel's supported data types.
I need it for mathematical operations and to round it back to <code>float64</code> data type which is already supported by Pyccel.

**Language**

It would be great if it's supported by both C and Fortran

EmilyBourne:
Blocked by #886

jalalium:
the closest thing C has to float128 is long double, but unfortunately it is only 80 bit long and will not do the job since it loses in precision compared to float128. GCC has a nonstandard type __float128 but i don't think we can use that. Another alternative would be creating our own custom type from two doubles and reassigning the bits to match with the IEEE754 standard.

PR #941, opened 2021-08-17
Add shortcuts to Fortran literals

jalalium:
This PR aims to resolve issue #856 . The following dictionary will be used instead of the original iso_c_binding dictionary whenever possible:
`iso_c_binding_shortcut = {
    "integer" : {
        1  : 'i8',
        2  : 'i16',
        4  : 'i32',
        8  : 'i64',
        16 : 'i128'}, #no supported yet
    "real"    : {
        4  : 'f32',
        8  : 'f64',
        16 : 'f128'},
    "logical" : {
        4  : "b4"}
}`
Checking if shortcuts can be implemented is done during the initialization of FCodePrinter through a systematic check of all the variable names in the namespace, this result is then stored as an attribute of the aforementioned class as an array of booleans of size 8 in self._namespace_restrictions.
The renaming is done in the print_kind method whenever it is possible.
finally, a line is added to state all the macros used in the module.

As an example, the Python code
```python
import numpy as np

a = np.int32(1) + 2 + 3.5
i32 = 5
print(i32)
print(a)
```
is translated to the Fortran code
```f90
program prog_t


      C_DOUBLE , C_INT32_T
  implicit none

  real(f64) :: a
  integer(i64) :: i32

  a = 1_C_INT32_T + 2_i64 + 3.5_f64
  i32 = 5_i64
  print *, i32
  print *, a

end program prog_t
```

### DIFF
--- a/pyccel/ast/datatypes.py
+++ b/pyccel/ast/datatypes.py
@@ -69,7 +69,7 @@ iso_c_binding = {
         2  : 'C_INT16_T',
         4  : 'C_INT32_T',
         8  : 'C_INT64_T',
-        16 : 'C_INT128'}, #no supported yet
+        16 : 'C_INT128_T'}, #no supported yet
     "real"    : {
         4  : 'C_FLOAT',
         8  : 'C_DOUBLE',
@@ -86,6 +86,7 @@ iso_c_binding_shortcut_mapping = {
     'C_INT16_T'             : 'i16',
     'C_INT32_T'             : 'i32',
     'C_INT64_T'             : 'i64',
+    'C_INT128_T'            : 'i128',
     'C_FLOAT'               : 'f32',
     'C_DOUBLE'              : 'f64',
     'C_LONG_DOUBLE'         : 'f128',
--- a/pyccel/ast/datatypes.py
+++ b/pyccel/ast/datatypes.py
@@ -81,7 +81,37 @@ iso_c_binding = {
     "logical" : {
         4  : "C_BOOL"}
 }
-
+iso_c_binding_shortcut = {
+    "integer" : {
+        1  : 'i8',
+        2  : 'i16',
+        4  : 'i32',
+        8  : 'i64',
+        16 : 'i128'}, #no supported yet
+    "real"    : {
+        4  : 'f32',
+        8  : 'f64',
+        16 : 'f128'},
+    "logical" : {
+        4  : "b4"}
+}
+iso_c_bindings = ['C_INT8_T', 'C_INT16_T', 'C_INT32_T', 'C_INT64_T', 'C_INT128',
+                  'C_FLOAT', 'C_DOUBLE', 'C_LONG_DOUBLE', "C_BOOL"]
+iso_c_binding_shortcuts = ['i8', 'i16', 'i32', 'i64', 'i128', 'f32', 'f64', 'f128', "b4"]
+iso_c_binding_check_index = {
+    "integer" : {
+        1  : 0,
+        2  : 1,
+        4  : 2,
+        8  : 3,
+        16 : 4}, #no supported yet
+    "real"    : {
+        4  : 5,
+        8  : 6,
+        16 : 7},
+    "logical" : {
+        4  : 8}
+}
 default_precision = {'real': 8,
                     'int': numpy.dtype(int).alignment,
                     'integer': numpy.dtype(int).alignment,
--- a/pyccel/ast/datatypes.py
+++ b/pyccel/ast/datatypes.py
@@ -81,36 +81,18 @@ iso_c_binding = {
     "logical" : {
         4  : "C_BOOL"}
 }
-iso_c_binding_shortcut = {
-    "integer" : {
-        1  : 'i8',
-        2  : 'i16',
-        4  : 'i32',
-        8  : 'i64',
-        16 : 'i128'}, #no supported yet
-    "real"    : {
-        4  : 'f32',
-        8  : 'f64',
-        16 : 'f128'},
-    "logical" : {
-        4  : "b4"}
-}
-iso_c_bindings = ['C_INT8_T', 'C_INT16_T', 'C_INT32_T', 'C_INT64_T', 'C_INT128',
-                  'C_FLOAT', 'C_DOUBLE', 'C_LONG_DOUBLE', "C_BOOL"]
-iso_c_binding_shortcuts = ['i8', 'i16', 'i32', 'i64', 'i128', 'f32', 'f64', 'f128', "b4"]
-iso_c_binding_check_index = {
-    "integer" : {
-        1  : 0,
-        2  : 1,
-        4  : 2,
-        8  : 3,
-        16 : 4}, #no supported yet
-    "real"    : {
-        4  : 5,
-        8  : 6,
-        16 : 7},
-    "logical" : {
-        4  : 8}
+iso_c_binding_shortcut_mapping = {
+    'C_INT8_T'              : 'i8',
+    'C_INT16_T'             : 'i16',
+    'C_INT32_T'             : 'i32',
+    'C_INT64_T'             : 'i64',
+    'C_FLOAT'               : 'f32',
+    'C_DOUBLE'              : 'f64',
+    'C_LONG_DOUBLE'         : 'f128',
+    'C_FLOAT_COMPLEX'       : 'C_FLOAT_COMPLEX',
+    'C_DOUBLE_COMPLEX'      : 'C_DOUBLE_COMPLEX',
+    'C_LONG_DOUBLE_COMPLEX' : 'C_LONG_DOUBLE_COMPLEX',
+    'C_BOOL'                : 'b4'
 }
 default_precision = {'real': 8,
                     'int': numpy.dtype(int).alignment,
--- a/pyccel/ast/datatypes.py
+++ b/pyccel/ast/datatypes.py
@@ -90,9 +90,9 @@ iso_c_binding_shortcut_mapping = {
     'C_FLOAT'               : 'f32',
     'C_DOUBLE'              : 'f64',
     'C_LONG_DOUBLE'         : 'f128',
-    'C_FLOAT_COMPLEX'       : 'C_FLOAT_COMPLEX',
-    'C_DOUBLE_COMPLEX'      : 'C_DOUBLE_COMPLEX',
-    'C_LONG_DOUBLE_COMPLEX' : 'C_LONG_DOUBLE_COMPLEX',
+    'C_FLOAT_COMPLEX'       : 'c32',
+    'C_DOUBLE_COMPLEX'      : 'c64',
+    'C_LONG_DOUBLE_COMPLEX' : 'c128',
     'C_BOOL'                : 'b4'
 }
 default_precision = {'real': 8,
--- a/pyccel/codegen/printing/fcode.py
+++ b/pyccel/codegen/printing/fcode.py
@@ -1,13 +1,14 @@
 # coding: utf-8
-# ------------------------------------------------------------------------------------------#
+#------------------------------------------------------------------------------------------#
 # This file is part of Pyccel which is released under MIT License. See the LICENSE file or #
 # go to https://github.com/pyccel/pyccel/blob/master/LICENSE for full license details.     #
-# ------------------------------------------------------------------------------------------#
+#------------------------------------------------------------------------------------------#
 # pylint: disable=R0201
 # pylint: disable=missing-function-docstring
 
 """Print to F90 standard. Trying to follow the information provided at
 www.fortran90.org as much as possible."""
+
 
 import string
 import re
@@ -22,28 +23,28 @@ from pyccel.ast.core import get_iterable_ranges
 from pyccel.ast.core import SeparatorComment, Comment
 from pyccel.ast.core import ConstructorCall
 from pyccel.ast.core import ErrorExit, FunctionAddress
-from pyccel.ast.internals import PyccelInternalFunction
+from pyccel.ast.internals    import PyccelInternalFunction
 from pyccel.ast.itertoolsext import Product
 from pyccel.ast.core import (Assign, AliasAssign, Declare,
                              CodeBlock, AsName,
                              If, IfSection, FunctionDef)
 
-from pyccel.ast.variable import (Variable, TupleVariable,
-                                 IndexedElement, HomogeneousTupleVariable,
-                                 InhomogeneousTupleVariable,
-                                 DottedName, PyccelArraySize, Constant)
+from pyccel.ast.variable  import (Variable, TupleVariable,
+                             IndexedElement, HomogeneousTupleVariable,
+                             InhomogeneousTupleVariable,
+                             DottedName, PyccelArraySize, Constant)
 
-from pyccel.ast.operators import PyccelAdd, PyccelMul, PyccelDiv, PyccelMinus, PyccelNot
-from pyccel.ast.operators import PyccelMod
+from pyccel.ast.operators      import PyccelAdd, PyccelMul, PyccelDiv, PyccelMinus, PyccelNot
+from pyccel.ast.operators      import PyccelMod
 
-from pyccel.ast.operators import PyccelUnarySub, PyccelLt, PyccelGt, IfTernaryOperator
+from pyccel.ast.operators      import PyccelUnarySub, PyccelLt, PyccelGt, IfTernaryOperator
 
-from pyccel.ast.core import FunctionCall, DottedFunctionCall
+from pyccel.ast.core      import FunctionCall, DottedFunctionCall
 
-from pyccel.ast.builtins import (PythonInt,
-                                 PythonPrint, PythonRange,
-                                 PythonFloat, PythonTuple)
-from pyccel.ast.builtins import PythonComplex, PythonBool, PythonAbs
+from pyccel.ast.builtins  import (PythonInt,
+                                  PythonPrint, PythonRange,
+                                  PythonFloat, PythonTuple)
+from pyccel.ast.builtins  import PythonComplex, PythonBool, PythonAbs
 from pyccel.ast.datatypes import is_pyccel_datatype
 from pyccel.ast.datatypes import is_iterable_datatype, is_with_construct_datatype
 from pyccel.ast.datatypes import NativeSymbol, NativeString, str_dtype
@@ -58,11 +59,11 @@ from pyccel.ast.datatypes import CustomDataType
 
 from pyccel.ast.internals import Slice
 
-from pyccel.ast.literals import LiteralInteger, LiteralFloat, Literal
-from pyccel.ast.literals import LiteralTrue
-from pyccel.ast.literals import Nil
+from pyccel.ast.literals  import LiteralInteger, LiteralFloat, Literal
+from pyccel.ast.literals  import LiteralTrue
+from pyccel.ast.literals  import Nil
 
-from pyccel.ast.mathext import math_constants
+from pyccel.ast.mathext  import math_constants
 
 from pyccel.ast.numpyext import NumpyEmpty
 from pyccel.ast.numpyext import NumpyMod, NumpyFloat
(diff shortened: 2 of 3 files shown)


pyccel/codegen/printing/fcode.py
```diff
@@ -74,112 +77,112 @@
...
-    'MathFactorial' : 'pyc_factorial',
-    'MathGcd'       : 'pyc_gcd',
-    'MathDegrees'   : 'pyc_degrees',
-    'MathRadians'   : 'pyc_radians',
-    'MathLcm'       : 'pyc_lcm',
+    'MathFactorial': 'pyc_factorial',
+    'MathGcd': 'pyc_gcd',
+    'MathDegrees': 'pyc_degrees',
+    'MathRadians': 'pyc_radians',
+    'MathLcm': 'pyc_lcm',
```
EmilyBourne, line 155: I'm guessing your editor did this automatically. We line these up purposely as we find it makes the code more readable. Could you find a way to turn off these automatic modifications in your editor please.

Having lots of trivial changes (e,g, whitespace changes) also makes it much harder to review your PR. According to github, you have modified 415 lines, which means the PR is much bigger than it needs to be

jalalium: I apologize for this, i will fix it right away!

EmilyBourne (changes requested):
Please fix the failing tests

pyccel/codegen/printing/fcode.py
EmilyBourne: Thanks 👍

pyccel/codegen/printing/fcode.py
```diff
@@ -206,6 +212,37 @@ def __init__(self, parser, prefix_module = None):
...
+
+    def print_macros(self):
+        """Prints the use line that indicates the macros used"""
+        macro = "use ISO_C_BINDING, only: "
+        rename = []
+        for idx, used in enumerate(self._macros):
+            if used:
+                rename.append(iso_c_binding_shortcuts[idx] + ' => ' + iso_c_bindings[idx])
+        macro += " , ".join(rename)
+        return macro
```
EmilyBourne, line 244: `C_INT`, etc are not macros, they are named constants (see [here](https://stackoverflow.com/tags/fortran-iso-c-binding/info) for more information). However here, you are not printing named constants either, you are printing the import of named constants. It is important to try and find a clear name so it is clear what is happening when the function is used elsewhere in the code. This means that other developers don't need to find the function and read the docstring to understand what is going on. For example this function could be called `_print_kind_imports`

pyccel/codegen/printing/fcode.py
```diff
@@ -285,7 +322,13 @@ def print_kind(self, expr):
...
         Prints the kind(precision) of a literal value
         """
-        return iso_c_binding[self._print(expr.dtype)][expr.precision]
+        if self._print(expr.dtype) == 'complex':
+            return iso_c_binding[self._print(expr.dtype)][expr.precision]
+        index = iso_c_binding_check_index[self._print(expr.dtype)][expr.precision]
+        if not self._namespace_restrictions[index]:
+            return iso_c_binding[self._print(expr.dtype)][expr.precision]
+        self._macros[index] = True
+        return iso_c_binding_shortcut[self._print(expr.dtype)][expr.precision]
```
EmilyBourne, line 331: `self._print(expr.dtype)` is called 5 times here! It would be better to call it once and save the result in a variable

pyccel/codegen/printing/fcode.py
```diff
@@ -206,6 +212,37 @@ def __init__(self, parser, prefix_module = None):
...
+                    valid[idx] = False
+        return valid
+
+    def print_macros(self):
+        """Prints the use line that indicates the macros used"""
+        macro = "use ISO_C_BINDING, only: "
+        rename = []
+        for idx, used in enumerate(self._macros):
+            if used:
+                rename.append(iso_c_binding_shortcuts[idx] + ' => ' + iso_c_bindings[idx])
```
EmilyBourne, line 242: If I understood you are only printing the import explicitly if the constant is renamed? Otherwise you are still using `use ISO_C_BINDING`? Does this not still cause collisions? `use MODULE_X` is equivalent to `from MODULE_X import *` in python

pyccel/codegen/printing/fcode.py
```diff
@@ -285,7 +322,13 @@ def print_kind(self, expr):
         """
         Prints the kind(precision) of a literal value
         """
-        return iso_c_binding[self._print(expr.dtype)][expr.precision]
+        if self._print(expr.dtype) == 'complex':
+            return iso_c_binding[self._print(expr.dtype)][expr.precision]
```
EmilyBourne, line 326: Is it necessary to treat complexes differently? Is it not possible to have entries for complexes in `iso_c_binding` but not in `iso_c_binding_shortcut`?

EmilyBourne (changes requested):
I think you should add a test where there is a collision to be sure that this is doing what you think it is doing

pyccel/ast/datatypes.py
```diff
@@ -81,7 +81,37 @@
...
+    "real"    : {
+        4  : 'f32',
+        8  : 'f64',
+        16 : 'f128'},
+    "logical" : {
+        4  : "b4"}
+}
+iso_c_bindings = ['C_INT8_T', 'C_INT16_T', 'C_INT32_T', 'C_INT64_T', 'C_INT128',
+                  'C_FLOAT', 'C_DOUBLE', 'C_LONG_DOUBLE', "C_BOOL"]
+iso_c_binding_shortcuts = ['i8', 'i16', 'i32', 'i64', 'i128', 'f32', 'f64', 'f128', "b4"]
```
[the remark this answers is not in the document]
jalalium: iso_c_binding_shortcut is used by print_kind to find the shortcut to use based on expr.dtype and expr.precision, whereas  iso_c_binding_shortcuts is used to print the use statement and it does not have access to the expr variable. The class keeps a boolean array of the shortcuts used and i thought it would be good to retrieve the name and shortcut from an array of strings instead of the dictionary that requires expr for look up.

pyccel/ast/datatypes.py
```diff
@@ -81,7 +81,37 @@
...
+        4  : 2,
+        8  : 3,
+        16 : 4}, #no supported yet
+    "real"    : {
+        4  : 5,
+        8  : 6,
+        16 : 7},
+    "logical" : {
+        4  : 8}
+}
```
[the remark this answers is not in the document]
jalalium: i chose this way to use arrays for restrictions and macros used, these additional structures represent a mapping between the dictionary keys and their index in the arrays. Shall i rely solely on the original dictionaries?

pyccel/ast/datatypes.py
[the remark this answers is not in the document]
EmilyBourne: I don't think the dictionaries are costly enough to be a problem. Having 2 objects with the same contents is usually a bad idea though (see [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principle).

>  shall i rely solely on the dictionaries?

Personally I think I would, it's clear and requires fewer modifications if you add a new datatype

> whereas iso_c_binding_shortcuts is used to print the use statement and it does not have access to the expr variable

You could also save a set of all the objects you used from `iso_c_binding_shortcut`

pyccel/ast/datatypes.py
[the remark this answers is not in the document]
EmilyBourne: See above. If you have a dictionary to find the index then I don't see how you are reducing the number of dictionary lookups

pyccel/codegen/printing/fcode.py
```diff
@@ -206,6 +212,37 @@ def __init__(self, parser, prefix_module = None):
...
+                if functionName == alias:
+                    valid[idx] = False
+        for variableName in self._namespace.imports['variables']:
+            for idx, alias in enumerate(iso_c_binding_shortcuts):
+                if variableName == alias:
+                    valid[idx] = False
+        for functionName in self._namespace.imports['functions']:
+            for idx, alias in enumerate(iso_c_binding_shortcuts):
+                if functionName == alias:
+                    valid[idx] = False
```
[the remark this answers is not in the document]
jalalium: Thank you, i did not know about this. I also was not aware about used_names. Using it instead of namespace will fix this problem.

pyccel/codegen/printing/fcode.py
```diff
@@ -206,6 +212,37 @@ def __init__(self, parser, prefix_module = None):
...
+        for functionName in self._namespace.functions:
+            for idx, alias in enumerate(iso_c_binding_shortcuts):
+                if functionName == alias:
+                    valid[idx] = False
+        for variableName in self._namespace.imports['variables']:
+            for idx, alias in enumerate(iso_c_binding_shortcuts):
+                if variableName == alias:
+                    valid[idx] = False
+        for functionName in self._namespace.imports['functions']:
```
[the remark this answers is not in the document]
jalalium: Thank you! i did not consider it. I should have used it from the start as it is exactly what i was looking for.

pyccel/codegen/printing/fcode.py
```diff
@@ -206,6 +212,37 @@ def __init__(self, parser, prefix_module = None):
 
         self.prefix_module = prefix_module
 
+    def check_restrictions(self):
+        """Checks if the namespace contains any of the shortcuts"""
+        valid = [True] * len(iso_c_binding_shortcuts)
+        for variableName in self._namespace.variables:
+            for idx, alias in enumerate(iso_c_binding_shortcuts):
+                if variableName == alias:
+                    valid[idx] = False
```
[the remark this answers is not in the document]
jalalium: noted! i will use what you showed me from now on!

pyccel/codegen/printing/fcode.py
```diff
@@ -198,6 +202,8 @@ def __init__(self, parser, prefix_module = None):
         super().__init__()
         self.parser = parser
         self._namespace = self.parser.namespace
+        self._namespace_restrictions = self.check_restrictions()
+        self._macros = [False] * len(iso_c_binding_shortcuts)
```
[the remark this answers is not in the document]
jalalium: what about `_used_kind_imports` ? basically, i have 8 possible shortcuts i can use, so this array of booleans keeps track of the shortcuts used. Maybe i should change what it stores as well?

pyccel/codegen/printing/fcode.py
jalalium: noted! i will change it!

pyccel/codegen/printing/fcode.py
jalalium: noted!

pyccel/codegen/printing/fcode.py
jalalium: sorry i did not understand this point.

pyccel/codegen/printing/fcode.py
jalalium: i will refactor the code to avoid treating them differently!

jalalium:
these are samples for testing:
this program has no restrictions so all shortcuts are used:
`import numpy as np
a = np.int32(1) + 2 + 3.5
print(a)
`
and it outputs:
`program prog_t

  use ISO_C_BINDING, only: i32 => C_INT32_T , i64 => C_INT64_T , f64 => &
      C_DOUBLE
  implicit none

  real(f64) :: a

  a = 1_i32 + 2_i64 + 3.5_f64
  print *, a

end program prog_t`
This second example has a variable named i32, so this shortcut won't be used:
`import numpy as np
a = np.int32(1) + 2 + 3.5
i32 = 5
print(i32)
print(a)
`
and it outputs:
`program prog_t

  use ISO_C_BINDING, only: i64 => C_INT64_T , f64 => C_DOUBLE
  implicit none

  real(f64) :: a
  integer(i64) :: i32

  a = 1_C_INT32_T + 2_i64 + 3.5_f64
  i32 = 5_i64
  print *, i32
  print *, a

end program prog_t
`

pyccel/codegen/printing/fcode.py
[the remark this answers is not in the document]
EmilyBourne: > what about `_used_kind_imports` ?

Sounds good to me

> Maybe i should change what it stores as well?

It's up to you, I imagine it depends what you do with your other modifications

pyccel/codegen/printing/fcode.py
EmilyBourne: Take the following code:
```
C_INT = 4.6
```
On the master branch this is translated to:
```
program prog_tmp

  implicit none

  real(C_DOUBLE) :: C_INT

  C_INT = 4.6_C_DOUBLE

end program prog_tmp
```
which raises the error:
```
/home/emily/Code/test/tmp/__pyccel__/tmp.f90:7:25:

    3 |   use, intrinsic :: ISO_C_BINDING
      |                   2      
......
    7 |   real(C_DOUBLE) :: C_INT
      |                         1
Error: Symbol ‘c_int’ at (1) conflicts with symbol from module ‘__iso_c_binding’, use-associated at (2)
/home/emily/Code/test/tmp/__pyccel__/tmp.f90:9:2:

    9 |   C_INT = 4.6_C_DOUBLE
      |  1
Error: Named constant ‘c_int’ in variable definition context (assignment) at (1)
```
If I modify the code to import explicitly as follows:
```
program prog_tmp

  use, intrinsic :: ISO_C_BINDING, only: C_DOUBLE

  implicit none

  real(C_DOUBLE) :: C_INT

  C_INT = 4.6_C_DOUBLE

end program prog_tmp
```
then the error disappears because `C_INT` is no longer imported. Your improvements have the potential to make this example work but currently it will raise the same error

EmilyBourne:
> I think you should add a test where there is a collision to be sure that this is doing what you think it is doing

Tests should be added to one of the tests/ folders so they are run as part of the testing suite. This prevents any of the other developers accidentally breaking your work in the future

pyccel/codegen/printing/fcode.py
jalalium: After investigating this issue, i found out that it all stems from 3 lines of code that print `imports += 'use, intrinsic :: ISO_C_BINDING\n'`in program, module and interface. I simply deleted these and only printed the constants i required and it did not seem to mess anything else since all tests passed.
This fixed the issue you described and outputted the following:
```
program prog_r


  use ISO_C_BINDING, only: f64 => C_DOUBLE
  implicit none

  real(f64) :: C_INT

  C_INT = 4.6_f64

end program prog_r
```
I will include this as well in my next push.

EmilyBourne: Great, just make sure you don't miss this case:
```
C_INT = 4.6
i64 = 3
```

jalalium: works as well!
```
program prog_r


  use ISO_C_BINDING, only: C_INT64_T
  implicit none

  integer(C_INT64_T) :: i64

  i64 = 3_C_INT64_T

end program prog_r
```

pyccel/ast/datatypes.py
```diff
@@ -81,7 +81,37 @@
...
+        8  : 'i64',
+        16 : 'i128'}, #no supported yet
+    "real"    : {
+        4  : 'f32',
+        8  : 'f64',
+        16 : 'f128'},
+    "logical" : {
+        4  : "b4"}
+}
+iso_c_bindings = ['C_INT8_T', 'C_INT16_T', 'C_INT32_T', 'C_INT64_T', 'C_INT128',
```
EmilyBourne, line 98: I realise this isn't your error, but is just copied from above, but would you mind correcting the literal constant for 128 bit integers. `C_INT128` is wrong. It should be `C_INT128_T` (see [here](https://gcc.gnu.org/onlinedocs/gcc-4.6.3/gfortran/ISO_005fC_005fBINDING.html))

pyccel/codegen/printing/fcode.py
EmilyBourne: Perfect :+1:

pyccel/ast/datatypes.py
jalalium: Noted! I will fix it. support for C_INT128_T is not  a priority?

EmilyBourne: It is not a priority, but it would be simple to add and is also a relatively simple issue. There isn't an issue open for it, but it could be done at the same time as issue #919 . This could be a good next issue for you, as long as PR #886 is merged before then

jalalium: Great!

EmilyBourne: The short version would therefore be `i128`

pyccel/ast/datatypes.py
```diff
@@ -81,7 +81,19 @@
     "logical" : {
         4  : "C_BOOL"}
 }
-
+iso_c_binding_shortcut_mapping = {
+    'C_INT8_T'              : 'i8',
+    'C_INT16_T'             : 'i16',
+    'C_INT32_T'             : 'i32',
+    'C_INT64_T'             : 'i64',
```
EmilyBourne, line 88: Missing `C_INT128_T` (as mentioned above this also needs correcting in `iso_c_binding`)

pyccel/codegen/printing/fcode.py
```diff
@@ -206,6 +208,20 @@ def __init__(self, parser, prefix_module = None):
 
         self.prefix_module = prefix_module
 
+    def print_constant_imports(self):
+        """Prints the use line for the constant imports used"""
+        macro = "use ISO_C_BINDING, only: "
```
EmilyBourne, line 213: I think we should keep the `intrinsic` tag (see [here](https://riptutorial.com/fortran/example/5650/intrinsic-modules) for details about what this does)

EmilyBourne (changes requested):
This looks good, and cleaner than before :). Please add a basic test to ensure that the code defaults to the long name correctly (and that the import is done correctly). This could be as simple as renaming one of the variables in one of the existing tests (e.g. [this file](https://github.com/pyccel/pyccel/blob/master/tests/pyccel/scripts/expressions.py)). Ideally we should add a specific test, but I'm not sure this issue is complicated enough to warrant that

EmilyBourne (approved):
Great work, looks good to me :)

pyccel/codegen/printing/fcode.py
```diff
@@ -206,6 +208,20 @@ def __init__(self, parser, prefix_module = None):
...
 
+    def print_constant_imports(self):
+        """Prints the use line for the constant imports used"""
+        macro = "use, intrinsic :: ISO_C_Binding, only : "
+        rename = []
+        for constant in self._constantImports:
+            if isinstance(constant, str):
+                rename.append(constant)
+            else:
+                rename.append(constant[0] + ' => ' + constant[1])
```
rakati, line 219: can you use list comprehension instead of forloop.
```python
rename = [c if isinstance(c, str) else c[0] + ' => ' + c[1] for c in self._constantImports]
```

pyccel/codegen/printing/fcode.py
```diff
@@ -285,7 +301,14 @@ def print_kind(self, expr):
         """
         Prints the kind(precision) of a literal value
```
rakati, line 302: can you update the description of the method to include the updated info in it?

pyccel/ast/datatypes.py
```diff
@@ -81,7 +81,20 @@
...
+    'C_INT128_T'            : 'i128',
+    'C_FLOAT'               : 'f32',
+    'C_DOUBLE'              : 'f64',
+    'C_LONG_DOUBLE'         : 'f128',
+    'C_FLOAT_COMPLEX'       : 'C_FLOAT_COMPLEX',
+    'C_DOUBLE_COMPLEX'      : 'C_DOUBLE_COMPLEX',
+    'C_LONG_DOUBLE_COMPLEX' : 'C_LONG_DOUBLE_COMPLEX',
```
rakati, line 95: can't we have shorter names for those types too?

EmilyBourne: We don't need shorter names for these types. The shorter names are there to make the code less verbose when writing literals (i.e. `3_C_INT`) however the only way to write a complex literal in is to write 2 real literals, so the complex datatypes are only used in declaration and type casting. Here it is not a problem if we are explicit rather than succinct

pyccel/codegen/printing/fcode.py
jalalium: noted!

pyccel/codegen/printing/fcode.py
jalalium: sure!

pyccel/ast/datatypes.py
rakati: ah, now I see.

rakati (approved):
Great Job @jalalium

jalalium:
> Great Job @jalalium

Thank you!

EmilyBourne:
@jalalium after the changes to the master branch there are now conflicts on this branch. Please can you fix this.

jalalium:
> @jalalium after the changes to the master branch there are now conflicts on this branch. Please can you fix this.

Sure! I am on it!

pyccel/ast/datatypes.py
yguclu: What about variables' declarations? Don't we want to use the shortcuts there too?

```f90
function add_complex(a, b) result(c)
    complex(c128), value :: a  ! shortcut
    complex(c128), value :: b  ! shortcut
    complex(c128) :: c         ! shortcut
    c = a + b
end function
```

pyccel/ast/datatypes.py
jalalium: I think support for a 128-bit integer through __int128 will depend on the target machine.
Should a check about it be performed?
Also, C does not allow the initialization of 128-bit integers, this should be done manually.
- Shall the lhs of the assignment be parsed as  a string?
- How shall this reflect on the generated code? can an assignment be represented as a function call?
- The same questions apply to __float128.

pyccel/ast/datatypes.py
EmilyBourne: If you think that this is better then we can. It will then at least all match. We didn't do this already because we were not certain of what the standard abbreviation was for complexes (different options were discussed on slack [here](https://pyccel.slack.com/archives/C018QDHENUU/p1629120620053900)). I didn't think it was a big problem if we were a little more verbose in the declarations

jalalium: So, shall i always abbreviate complex datatypes? shall i use c32, c64, and c128?

yguclu: I don't think it is important what abbreviation we use for complex numbers, as we can always change our dictionary in the future. But if our goal is to make the code less verbose, than I suggest that we do it systematically also for the variable declarations.

yguclu: > So, shall i always abbreviate complex datatypes? shall i use c32, c64, and c128?

Yes, I think that `c64`, `c128`, `c256`  and so forth should be clear enough and will not generate any confusion for the user.

yguclu:
> these are samples for testing:
> this program has no restrictions so all shortcuts are used:
> 
> ```
> import numpy as np
> a = np.int32(1) + 2 + 3.5
> print(a)
> ```
> 
> and it outputs:
> 
> ```
> program prog_t
> 
>   use, intrinsic :: ISO_C_BINDING
> 
>   use ISO_C_BINDING, only: i32 => C_INT32_T , i64 => C_INT64_T , f64 => &
>       C_DOUBLE
>   implicit none
> 
>   real(f64) :: a
> 
>   a = 1_i32 + 2_i64 + 3.5_f64
>   print *, a
> 
> end program prog_t
> ```
> 
> This second example has a variable named i32, so this shortcut won't be used:
> 
> ```
> import numpy as np
> a = np.int32(1) + 2 + 3.5
> i32 = 5
> print(i32)
> print(a)
> ```
> 
> and it outputs:
> 
> ```
> program prog_t
> 
>   use, intrinsic :: ISO_C_BINDING
> 
>   use ISO_C_BINDING, only: i64 => C_INT64_T , f64 => C_DOUBLE
>   implicit none
> 
>   real(f64) :: a
>   integer(i64) :: i32
> 
>   a = 1_C_INT32_T + 2_i64 + 3.5_f64
>   i32 = 5_i64
>   print *, i32
>   print *, a
> 
> end program prog_t
> ```

I do not think that we should have multiple `use` statement for `ISO_C_BINDING` in the same module. If we use `only` we must print every variable that is used:
```f90
use, intrinsic :: ISO_C_BINDING, only: &
    i64 => C_INT64_T, &
    f64 => C_DOUBLE, &
    C_INT32_T
```
Otherwise, we can just provide the renaming list without `only`:
```f90
use, intrinsic :: ISO_C_BINDING, &
    i64 => C_INT64_T, &
    f64 => C_DOUBLE
```
I personally prefer the first solution, which is more explicit. The complete syntax of the Fortran use statement can be checked [here](https://wg5-fortran.org/documents.html).

yguclu:
@jalalium Please disregard my last comment, which was based on an out-of-date code snippet. I've just checked the branch and the generated code exactly matches my suggestion! 😃

jalalium:
> @jalalium Please disregard my last comment, which was based on an out-of-date code snippet. I've just checked the branch and the generated code exactly matches my suggestion! smiley

Indeed, this change was made. i m working on making the change for complex literals. for some reason, a test fails i m trying to find out why.

yguclu (approved):

PR #948, opened 2021-08-25
Add shape of results to macro language to avoid unnecessary variable declarations

### LINKED ISSUE #927: Macro output must be declared before expanding the macro
**Describe the bug**
When a macro represents a function in python and a subroutine in fortran the output variables must be declared before the subroutine is called. In pyccel this means that you must also declare the variables before calling the function in python. For scalar quantities it should be trivial to add the declaration as the type can be deduced from the subroutine signature. For array quantities it would be nice to add something to the header language or macros to allow the size of these arrays to be specified

**To Reproduce**
For example the following compiles:
```python
from scipy.linalg.lapack import dgetrf
import numpy as np
def lu_decom(A: 'float32[:,:]'):
    n = len(A)
    piv = np.empty(n)
    info = -1
    A, piv, info = dgetrf(A)
```
but the following:
```python
from scipy.linalg.lapack import dgetrf
def lu_decom(A: 'float32[:,:]'):
    n = len(A)
    A, piv, info = dgetrf(A)
```
raises the error:
```
pyccel:
 |fatal [semantic]: lapack.py [6,4]| Undefined variable (piv)
```

Similarly the following:
```
from scipy.linalg.lapack import dgetrf
import numpy as np
def lu_decom(A: 'float32[:,:]'):
    n = len(A)
    piv = np.empty(n)
    A, piv, info = dgetrf(A)
```
also raises the error:
```
pyccel:
 |fatal [semantic]: lapack.py [7,4]| Undefined variable (info)
```
despite the fact that it should be possible to determine the type of `info` from the signature of `dgetrf`

**Expected behavior**
Scalar types should be declared by the macro

For array types, either a clearer error should be raised explaining why this command does not count as a variable definition, or something should be added to the macro language to allow the size to be inferred thus allowing the variable to be declared and allocated by the macro

**Additional context**
Related to #925

rmahjoubi:
**Declare results of a macro when necessary**

-Extend macro header syntax to also hold shape of the results
-Add the class `ShapedID` which represents the shape of a variable of the name ID
-Create the function `link_args` which link macro arguments to their appropriate `FunctionCall` arguments
-Extend macros in `stdlib/external` and in tests
-Remove dummy declarations from previous tests

Fixes #927.


pyccel/stdlib/external/lapack.py
```diff
@@ -16,10 +16,10 @@
 
 
 
-#$ header macro (ab, IPIV, info), dgetrf(ab) := dgetrf(ab.shape[0], ab.shape[1], ab, ab.shape[0], IPIV, info)
-#$ header macro (b, info), dgetrs(ab, piv, b, s='N') := dgetrs(s, ab.shape[1], 1, ab, ab.shape[0], piv, b, b.count, info)
+#$ header macro (double(ab.shape[0], ab.shape[1]), int(ab.shape[0]),int) :: (ab, IPIV, info), dgetrf(ab) := dgetrf(ab.shape[0], ab.shape[1], ab, ab.shape[0], IPIV, info)
+#$ header macro (double(b.shape[0], b.shape[1]), int) :: (b, info), dgetrs(ab, piv, b, s='N') := dgetrs(s, ab.shape[1], 1, ab, ab.shape[0], piv, b, b.count, info)
```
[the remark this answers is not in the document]
rmahjoubi: Can you please check this commit a9d90b5.
We still don't handle interfaces for macro functions, Is there an issue for this?

pyccel/parser/semantic.py
```diff
@@ -2131,27 +2131,26 @@ def _visit_Assign(self, expr, **settings):
...
                     lhs = [lhs]
-
-                results = []
-                for a in lhs:
-                    _name = _get_name(a)
-                    var = self.get_variable(_name)
-                    results.append(var)
+                for d_var, var in zip(d_results, lhs):
+                    if not var in args_names:
+                        tmp = self._assign_lhs_variable(var, d_var, None, new_expressions, None, **settings)
```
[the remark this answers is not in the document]
rmahjoubi: Fixed here 948a83b

pyccel/ast/headers.py
```diff
@@ -667,11 +674,50 @@ def apply(self, args, results=None):
...
                 if isinstance(arg, Macro):
                     d_arguments[i] = construct_macro(arg.name,
-                                      d_arguments[arg.argument])
+                                                     d_arguments[arg.argument])
                     if isinstance(arg, MacroShape):
                         d_arguments[i]._index = arg.index
+        return d_arguments
+
+    def apply_to_results(self, args):
+        """replace elements of the shape with appropriate values"""
```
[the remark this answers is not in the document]
rmahjoubi: Fiexed here 08db5ab

pyccel/ast/headers.py
```diff
@@ -627,11 +628,17 @@ def master_arguments(self):
     def results(self):
         return self._results
 
+    @property
+    def results_sh(self):
+        return self._results_sh
```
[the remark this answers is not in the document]
rmahjoubi: Fixed here ad49638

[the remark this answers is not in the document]
rmahjoubi: Should I change it to results_shapes?

pyccel/parser/syntax/headers.py
```diff
@@ -157,6 +157,62 @@ def expr(self):
...
+        d_var['is_pointer'] = False
+        d_var['precision']  = precision
+        d_var['is_const'] = False
+        if not(precision):
+            if dtype in ['double' ,'float','complex', 'int']:
+                d_var['precision'] = default_precision[dtype]
+
+        if d_var['rank']>1:
+            d_var['order'] = order
+        d_var['shape']=shape
```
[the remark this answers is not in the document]
rmahjoubi: Fixed here 2214a28

pyccel/ast/headers.py
```diff
@@ -589,22 +589,23 @@ def funcs(self):
 class MacroFunction(Header):
     """."""
```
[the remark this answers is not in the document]
rmahjoubi: Check this 17b932d
Any improvements you can suggest?

pyccel/errors/messages.py
```diff
@@ -152,3 +152,4 @@
 ARRAY_DEFINITION_IN_LOOP = 'Array definition in for loop may cause memory reallocation at each cycle. Consider creating the array before the loop'
 TEMPLATE_IN_UNIONTYPE = 'Cannot use templates in a union type'
 DUPLICATED_SIGNATURE = 'Same signature defined for the same function multiple times'
+UNVALID_MACRO_COMPOSITION = 'Unvalid macro composition'
```
EmilyBourne, line 155: `Unvalid`->`Invalid`

pyccel/parser/grammar/headers.tx
```diff
@@ -81,8 +81,15 @@ AvailableMacro:
 
 // TODO shall we add also results to the master call?
 
+//Extended Type with shape
+ExType: dtype=DataType ('*' prec=INT)? ('(' shape+=MacroValue[','] ')')? ( '(' 'order' '=' order = ID ')' )?;
+
+Shape: MacroValue | ':';
+
+MacroResult: name=ID ('(' shape+=Shape[','] ')')?;
```
EmilyBourne, line 89: I would like to choose a more generic name for this if possible.

I suspect this object will be used more generally very soon. For example for the function `sgesvx` defined here in [pyccel](https://github.com/pyccel/pyccel/blob/master/pyccel/stdlib/internal/lapack.pyh#L27) and [here](https://www.kite.com/python/docs/scipy.linalg.lapack.sgesvx) in scipy. As you can see there are default arguments which are arrays (used to store memory inside the function call. I suspect that we will need to use `MacroResult` to recreate this. In this case it is no longer just a result, hence my request for a more generic name

pyccel/ast/headers.py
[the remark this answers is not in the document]
EmilyBourne: Maybe a bit more description. I'm not sure that "Represents a macro function header in the code" would be enough to explain to someone who has never worked with the macros before, what this is. Maybe:
```
Represents a macro function header in the code.
A macro function header maps a defined function name, set of arguments and results to
a different function call. This is notably useful when mapping python functions to fortran
functions which do the same thing but do not have the same set of arguments/results
```

pyccel/stdlib/external/lapack.py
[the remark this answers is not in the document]
EmilyBourne: I don't think there is an issue. I don't know if it is worth creating one or not as I'm not sure it's actually needed. But as a macro can map to any function (including a python function) not just a fortran function, there is technically nothing stopping it. And there are several comments in the code about it

pyccel/parser/syntax/headers.py
```diff
@@ -157,6 +157,38 @@ def expr(self):
...
+
+    def __init__(self, **kwargs):
+        """Constructor for MacroResult
+
+        name: str
+            Name of the variable result
+
+        shape: list
+            a list representing the shape of the result
+        """
```
EmilyBourne, line 171: `__init__` functions do not need doc strings. In order to call a constructor you do:
```
MacroResult(..)
```
Therefore in order to see the docstring you would usually do;
```
MacroResult?
```
This shows the docstring for the class which is why the arguments for the `__init__` function should go in the class docstring

pyccel/parser/syntax/headers.py
```diff
@@ -157,6 +157,38 @@ def expr(self):
...
+    """class representing a MacroResult in the grammar."""
+
+    def __init__(self, **kwargs):
+        """Constructor for MacroResult
+
+        name: str
+            Name of the variable result
+
+        shape: list
+            a list representing the shape of the result
```
EmilyBourne, line 170: Missing
```
Parameters
------------
```

pyccel/parser/syntax/headers.py
```diff
@@ -157,6 +157,38 @@ def expr(self):
...
+        """
+        self._name   = kwargs.pop('name')
+        self._shape  = kwargs.pop('shape', [])
+
+        super().__init__(**kwargs)
+
+    @property
+    def expr(self):
+        """."""
```
EmilyBourne, line 179: This is not a docstring it is a pylint avoidance technique. Please use a real docstring

pyccel/ast/headers.py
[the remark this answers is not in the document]
EmilyBourne: Personally I would find that clearer

pyccel/ast/headers.py
```diff
@@ -667,11 +710,70 @@ def apply(self, args, results=None):
...
+                    else:
+                        new = arg
+                elif isinstance(arg, Macro):
+                    if arg.argument in argument_keys:
+                        new = d_arguments[arg.argument]
+                    else:
+                        raise ValueError('Unknown variable name')
+
+                    if isinstance(arg, MacroShape):
+                        new = construct_macro(arg.name, new, arg.index)
```
EmilyBourne, line 762: Is this not clearer if you call `MacroShape` directly ;)

pyccel/ast/headers.py
```diff
@@ -696,8 +798,8 @@ def apply(self, args, results=None):
                     new = d_results[arg]
                 else:
                     new = arg
-               #TODO uncomment later
-               #     raise ValueError('Unknown variable name')
+                #TODO uncomment later
+                #     raise ValueError('Unknown variable name')
```
EmilyBourne, line 802: Do we still need this? I couldn't work out what it was for

pyccel/parser/grammar/headers.tx
```diff
@@ -81,8 +81,15 @@ AvailableMacro:
 
 // TODO shall we add also results to the master call?
 
+//Extended Type with shape
+ExType: dtype=DataType ('*' prec=INT)? ('(' shape+=MacroValue[','] ')')? ( '(' 'order' '=' order = ID ')' )?;
```
EmilyBourne, line 85: Where is this used?

pyccel/parser/semantic.py
```diff
@@ -2131,27 +2131,33 @@ def _visit_Assign(self, expr, **settings):
...
-
                 master = macro.master
                 name = _get_name(master.name)
-
-                # all terms in lhs must be already declared and available
-                # the namespace
-                # TODO improve
+                results = []
+                args = [self._visit(i, **settings) for i in
+                            rhs.args]
```
EmilyBourne, line 2138: Can't this be 1 line?

pyccel/parser/semantic.py
```diff
@@ -2131,27 +2131,33 @@ def _visit_Assign(self, expr, **settings):
...
+                args_names = [arg.name for arg in args if isinstance(arg, Variable)]
 
                 if not sympy_iterable(lhs):
                     lhs = [lhs]
+                results_shapes = macro.get_results_shapes(args)
+                for i, result in enumerate(macro.results):
+                    if result in macro.master_arguments and not lhs[i] in args_names:
```
EmilyBourne, line 2145: You use `lhs[i]` a several times in this loop. Why not use `zip` ? Either:
```
for i, (lhs_i, result) in enumerate(zip(lhs, macro.results)):
```
or:
```
for result, shape, lhs_i in zip(macro.results, results_shapes, lhs):
```

rmahjoubi: Fixed here 0e3af3b

pyccel/parser/semantic.py
rmahjoubi: Fixed here c3e0ad8

pyccel/parser/grammar/headers.tx
rmahjoubi: Fixed here 813fd40

pyccel/ast/headers.py
rmahjoubi: Fixed here 9e3988d

pyccel/ast/headers.py
rmahjoubi: I was patiently waiting to steal from your branch h
Fixed here 9e3988d

pyccel/parser/syntax/headers.py
rmahjoubi: Sorry, Fixed here ee14144

pyccel/parser/syntax/headers.py
rmahjoubi: Fixed here 0717085

pyccel/parser/syntax/headers.py
rmahjoubi: Fixed here 0717085

pyccel/errors/messages.py
rmahjoubi: Fixed here d8b05e4

pyccel/parser/grammar/headers.tx
rmahjoubi: can you please suggest a name for this?

pyccel/ast/headers.py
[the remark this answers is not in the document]
rmahjoubi: Fixed here 041d53e

pyccel/ast/headers.py
```diff
@@ -627,11 +628,17 @@ def master_arguments(self):
     def results(self):
         return self._results
 
+    @property
+    def results_sh(self):
+        return self._results_sh
+
     # TODO: must be moved to annotation, once we add AliasVariables
     #       this is needed if we have to create a pointer or allocate a new
     #       variable to store the result
```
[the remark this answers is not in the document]
rmahjoubi: I don't think so, I will remove it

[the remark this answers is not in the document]
rmahjoubi: Removed here 34eb783

pyccel/ast/headers.py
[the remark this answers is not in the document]
rmahjoubi: Fixed here 4b6f12c

pyccel/ast/headers.py
EmilyBourne: Haha, mine got stuck waiting for someone to take the time to look at it. But as this is new code I think it will cause a collision anyway

pyccel/parser/semantic.py
```diff
@@ -1390,7 +1390,7 @@ def _assign_lhs_variable(self, lhs, d_var, rhs, new_expressions, is_augassign, *
                             else:
                                 status='allocated'
                             new_expressions.append(Allocate(var,
-                                shape=d_var['shape'], order=d_var['order'],
+                                shape=d_var['shape'], order=d_var.get('order'),
```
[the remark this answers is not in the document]
rmahjoubi: Fixed here 39a0f99

pyccel/parser/grammar/headers.tx
EmilyBourne: `MacroShapedParameter` ? `ShapedID` ?

pyccel/parser/semantic.py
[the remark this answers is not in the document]
EmilyBourne: If this doesn't break anything then I think we should change it back

pyccel/stdlib/external/lapack.py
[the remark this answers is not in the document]
rmahjoubi: Ok, then we'll take care of it when it is needed

pyccel/parser/grammar/headers.tx
rmahjoubi: Changed to `ShapedID` here bb2299d

pyccel/ast/headers.py
```diff
@@ -587,24 +587,52 @@ def funcs(self):
...
-        self._master           = master
-        self._master_arguments = master_args
-        self._results          = results
-        self._copies_required  = [a in self._results for a in self._arguments]
+        self._name              = name
+        self._arguments         = args
+        self._master            = master
+        self._master_arguments  = master_args
+        self._results           = results
+        self._results_shapes        = results_shapes
```
EmilyBourne, line 634: For some reason this line is aligned differently to the rest. Please could you fix it

pyccel/ast/headers.py
```diff
@@ -627,11 +655,26 @@ def master_arguments(self):
...
+        Parameters
+        ----------
+        args: iterable
+            The arguments provided at function call
+
+        Results
+        -------
+        d_arguments: dictionary
+            A Dictionary, where the keys are arguments from the macro
+            and values are arugments provided at function call
```
EmilyBourne, line 675: Typo `arugments`->`arguments`

pyccel/ast/headers.py
```diff
@@ -667,11 +710,70 @@ def apply(self, args, results=None):
...
+        for j, shape in enumerate(self.results_shapes):
+            newargs = []
+
+            for i, arg in enumerate(shape):
+                if str(arg) == ':':
+                    try:
+                        new = MacroShape(d_arguments[self.results[j]], i)
+                    except KeyError:
+                        msg = "Shape needs to be provided explicitly as it cannot be deduced"
+                        errors.report(msg, symbol=self.results[j],
```
EmilyBourne, line 747: I think `zip` makes more sense here than `enumerate`. You never use `j` directly

pyccel/ast/headers.py
```diff
@@ -667,11 +710,70 @@ def apply(self, args, results=None):
...
+                                      severity='error')
+
+                elif isinstance(arg, PyccelSymbol):
+                    if arg in argument_keys:
+                        new = d_arguments[arg]
```
EmilyBourne, line 752: Do you need the variable `argument_keys`? `arg in d_arguments` is strictly equivalent to `d_arguments.keys()`

pyccel/ast/headers.py
```diff
@@ -667,11 +710,70 @@ def apply(self, args, results=None):
...
+                else:
+                    new=arg
+
+                newargs.append(new)
+            newargs = tuple(newargs)
+            results_shapes.append(newargs)
+        return results_shapes
+
+    def apply(self, args, results=None):
+        """returns the appropriate arguments."""
```
EmilyBourne, line 774: XD github is making this useless docstring look like your fault. Can I suggest changing it to:
```
Converts the arguments provided to the macro to arguments appropriate for the FunctionCall to the associated function
```

pyccel/parser/semantic.py
```diff
@@ -2131,27 +2131,32 @@ def _visit_Assign(self, expr, **settings):
...
+                args = [self._visit(i, **settings) for i in rhs.args]
+                args_names = [arg.name for arg in args if isinstance(arg, Variable)]
 
                 if not sympy_iterable(lhs):
                     lhs = [lhs]
+                results_shapes = macro.get_results_shapes(args)
+                for m_result, shape, result in zip(macro.results, results_shapes, lhs):
+                    if m_result in macro.master_arguments and not result in args_names:
+                        index = macro.master_arguments.index(m_result)
+                        d_result = self._infere_type(master.arguments[index])
```
EmilyBourne, line 2146: Why can you not just do `d_result = self._infere_type(m_result)`? I'm not sure what `index` is adding here

pyccel/parser/semantic.py
```diff
@@ -2131,27 +2131,32 @@ def _visit_Assign(self, expr, **settings):
...
+                        d_result['shape'] = shape
+                        tmp = self._assign_lhs_variable(result, d_result, None, new_expressions, False, **settings)
+                        results.append(tmp)
+                    elif result in args_names:
+                        _name = _get_name(result)
+                        tmp = self.get_variable(_name)
+                        results.append(tmp)
+                    else:
+                        # TODO: check for result in master_results
```
EmilyBourne, line 2155: Is this TODO still valid?

pyccel/parser/syntax/headers.py
```diff
@@ -157,6 +157,39 @@ def expr(self):
...
+        """Returns a dictionary containing name and shape of result"""
+        d_var = {}
+        shape = []
+        for i in self._shape:
+            if isinstance(i, MacroStmt):
+                shape.append(i.expr)
+            else:
+                shape.append(PyccelSymbol(i))
+        d_var['name']   = self._name
+        d_var['shape']  = shape
```
EmilyBourne, line 189: I suspect this would be faster if you used comprehensions:
```
shape = [i.expr if isinstance(i, MacroStmt) else PyccelSymbol(i) for i in self._shape]
d_var = {'name': self._name, 'shape': shape}
```

rmahjoubi: Fixed b70d8ca

pyccel/ast/headers.py
rmahjoubi: sure, changed here 07f9270

pyccel/ast/headers.py
rmahjoubi: Fixed b3ac39e

pyccel/ast/headers.py
rmahjoubi: Fixed 18df3a1

pyccel/ast/headers.py
rmahjoubi: Fixed d21b82e

pyccel/ast/headers.py
rmahjoubi: Fixed 3ab5fab

pyccel/parser/semantic.py
rmahjoubi: I just added this TODO, I though we might want to have master_results in macros, this is suggested in another TODO

pyccel/parser/semantic.py
rmahjoubi: `m_result` is coming from the macro, it is still of type `PyccelSymbol`, that's why i need to get the index of m_result, to get the according argument from the `FunctionDef` `master`

pyccel/parser/semantic.py
EmilyBourne: ok :+1:

EmilyBourne (commented):
There is a codacy suggestion (now that it can recognise the dictionary). I will approve when that is fixed

rmahjoubi:
codacy error fixed

EmilyBourne (approved):

EmilyBourne:
There seems to be a problem with the merge

rmahjoubi:
I think we should make a distinction between `MacroFuntion` and `FunctionDef` when it comes to arguments.
For example in contrast to a `MacroFunction` argument, a `FunctionDef` argument cannot be a `LiteralInteger`, this means we can have all FunctionDef arguments of type `FunctionDefArgument`, in the other hand not all `MacroFunction` arguments are necessary of type `FunctionDefArgument`. In my opinion this contradicts the notions, I think we either keep all the arguments of a `MacroFunction` as `PyccelSymbols` and Literals, or we create another class that inherits from `FunctionDefArgument` dedicated to this kind of usage.
what do you think @EmilyBourne

EmilyBourne:
> I think we should make a distinction between `MacroFuntion` and `FunctionDef` when it comes to arguments.
> For example in contrast to a `MacroFunction` argument, a `FunctionDef` argument cannot be a `LiteralInteger`, this means we can have all FunctionDef arguments of type `FunctionDefArgument`, in the other hand not all `MacroFunction` arguments are necessary of type `FunctionDefArgument`. In my opinion this contradicts the notions, I think we either keep all the arguments of a `MacroFunction` as `PyccelSymbols` and Literals, or we create another class that inherits from `FunctionDefArgument` dedicated to this kind of usage.
> what do you think @EmilyBourne

I am not sure I follow your argument. Maybe you can provide an example? If a `MacroFunction` argument is a `LiteralInteger` then it seems to me that it is not an argument any more.
Taking this macro as an example:
```
#$ header macro (b(:,:), info), dgetrs(ab, piv, b, s='N') := dgetrs(s, ab.shape[1], 1, ab, ab.shape[0], piv, b, b.count, info)
```
I think macros have the following syntax:
```
#$ header macro (ShapedID, ShapedID), func_name(FunctionDefArgument(ab), ...,  FunctionDefArgument(s,default='N')) := FunctionCall(dgetrs, (FunctionCallArgument(s), FunctionCallArgument(ab.shape[1]), FunctionCallArgument(1), ...)
```

rmahjoubi:
> > I think we should make a distinction between `MacroFuntion` and `FunctionDef` when it comes to arguments.
> > For example in contrast to a `MacroFunction` argument, a `FunctionDef` argument cannot be a `LiteralInteger`, this means we can have all FunctionDef arguments of type `FunctionDefArgument`, in the other hand not all `MacroFunction` arguments are necessary of type `FunctionDefArgument`. In my opinion this contradicts the notions, I think we either keep all the arguments of a `MacroFunction` as `PyccelSymbols` and Literals, or we create another class that inherits from `FunctionDefArgument` dedicated to this kind of usage.
> > what do you think @EmilyBourne
> 
> I am not sure I follow your argument. Maybe you can provide an example? If a `MacroFunction` argument is a `LiteralInteger` then it seems to me that it is not an argument any more.
> Taking this macro as an example:
> 
> ```
> #$ header macro (b(:,:), info), dgetrs(ab, piv, b, s='N') := dgetrs(s, ab.shape[1], 1, ab, ab.shape[0], piv, b, b.count, info)
> ```
> 
> I think macros have the following syntax:
> 
> ```
> #$ header macro (ShapedID, ShapedID), func_name(FunctionDefArgument(ab), ...,  FunctionDefArgument(s,default='N')) := FunctionCall(dgetrs, (FunctionCallArgument(s), FunctionCallArgument(ab.shape[1]), FunctionCallArgument(1), ...)
> ```

In the current implementation we have:
#$ header macro (ShapedID, ShapedID), func_name(FunctionDefArgument(ab), ...,  FunctionDefArgument(s,default='N')) := FunctionDef(dgetrs, (FunctionDefArgument(s), FunctionDefArgument(ab.shape[1]), LiteralInteger(1)...)

EmilyBourne:
> In the current implementation we have:
> `#$ header macro (ShapedID, ShapedID), func_name(FunctionDefArgument(ab), ..., FunctionDefArgument(s,default='N')) := FunctionDef(dgetrs, (FunctionDefArgument(s), FunctionDefArgument(ab.shape[1]), LiteralInteger(1)...) `

Do you think this is what we should have or do you agree with my definition above? For me the right hand side of the macro declaration is what will appear in the code and should therefore be a `FunctionCall`. There will never be default arguments here, but there could be arguments passed by keyword, so it definitely seems to be a `FunctionCall` and not a `FunctionDef`. I must not have been paying enough attention when I coded that section

pyccel/ast/headers.py
```diff
@@ -621,9 +621,9 @@ def __init__(self, name, args, master, master_args, results=None, results_shapes
         if not isinstance(name, str):
             raise TypeError('name must be of type str or PyccelSymbol')
 
-        # master can be a string or FunctionDef
-        if not isinstance(master, (str, FunctionDef, Interface)):
-            raise ValueError('Expecting a master name of FunctionDef')
+        # master can be a string or FunctionCall
+        if not isinstance(master, (str, FunctionCall)):
+            raise ValueError('Expecting a function name, or a FunctionCall')
```
EmilyBourne: This must be how we ended up with `FunctionDefArgument`s, I thought this was just the name of the function used on the rhs, not the whole rhs

EmilyBourne (approved):

EmilyBourne (approved):

saidctb (approved):

EmilyBourne:
I have just had a conversation with @AbdelGhafour69 about lapack routines and we found a couple of problems.
He was testing the following function and getting wrong output:
```
def lu_decom(A: 'float64[:,:](order=F)'):
    n = len(A)
    piv = np.zeros(n)
    info = -1
    A, piv, info = dgetrf(A)
    print('Pivot array: \n',piv)
```
The solution was this:
```
def lu_decom(A: 'float64[:,:](order=F)'):
    n = len(A)
    piv = np.zeros(n, dtype=np.int32)
    info = -1
    A, piv, info = dgetrf(A)
    print('Pivot array: \n',piv)
```
On the master branch the first piece of code raises no errors. On this branch the situation is improved and an error is raised, however the following is accepted:
```
def lu_decom(A: 'float64[:,:](order=F)'):
    n = len(A)
    piv = np.zeros(n, dtype=np.int64)
    info = -1
    A, piv, info = dgetrf(A)
    print('Pivot array: \n',piv)
```
This means that this branch is not currently checking if the precision is modified. However if this is fixed we will encounter a different problem. Currently the contents of [this file](https://github.com/pyccel/pyccel/blob/master/pyccel/stdlib/internal/lapack.pyh#L14) are incorrect. All the integer types are declared as `int`, however this should correspond to the default fortran `int` (`int32`) not the default python int. This therefore also needs fixing so that checking the precision doesn't prevent you from running the correct code

EmilyBourne:
>This means that this branch is not currently checking if the precision is modified

Hmm, I've just looked into this and it seems it is a general problem not specific to this branch

pyccel/parser/syntax/headers.py
```diff
@@ -495,11 +520,12 @@ def expr(self):
...
-        results = self.results
-        if (results is None):
-            results = []
-
+        results = []
+        results_shapes = []
+        if self.results:
+            results = [res.expr['name'] for res in self.results]
+            results = [PyccelSymbol(r) for r in results]
+            results_shapes = [res.expr['shape'] for res in self.results]
```
yguclu, line 528: For clarity I recommend that we avoid redefining the same variable multiple times. And if we can make sure that `self.results` is always a list (even if empty - see my previous comment), this code reduces to just two lines:
```python
results = [PyccelSymbol(r.expr['name']) for r in self.results]
results_shapes = [r.expr['shape'] for r in self.results]
```

pyccel/parser/semantic.py
```diff
@@ -3511,17 +3514,18 @@ def _visit_MacroFunction(self, expr, **settings):
...
+                return FunctionCallArgument(master_arg)
 
-        master_args = [get_arg(a,m) if isinstance(m, PyccelSymbol) else m
-                        for a,m in zip(func.arguments, expr.master_arguments)]
+        master_args = [get_arg(a,m) for a,m in zip(func.arguments, expr.master_arguments)]
 
         results = expr.results
-        macro   = MacroFunction(name, args, func, master_args,
-                                  results=results)
+        results_shapes = expr.results_shapes
```
yguclu, line 3525: Are these two lines needed?

Can't you just pass `expr.results` and `expr.results_shapes` to `MacroFunction` below?

yguclu (approved):
Excellent job @rmahjoubi! I only have a couple of minor comments.

pyccel/parser/semantic.py
rmahjoubi: Fixed 6701afb

pyccel/parser/syntax/headers.py
rmahjoubi: Fixed 3d84be1

pyccel/parser/syntax/headers.py
```diff
@@ -449,8 +476,6 @@ def __init__(self, **kwargs):
 
         self.dotted_name = tuple(kwargs.pop('dotted_name'))
         self.results = kwargs.pop('results',None)
```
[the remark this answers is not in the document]
rmahjoubi: Fixed 3d84be1

yguclu (approved):

Issue #742, opened 2021-02-11
Modify module variables from Python using setters


yguclu:
As a USER, I would like that my Python code could modify the module variables of a C Python extension module generated by Pyccel.

EmilyBourne:
Blocked by #886

EmilyBourne:
Doc can be found [here](https://docs.python.org/3/c-api/typeobj.html#c.setattrfunc) and [here](https://docs.python.org/3/extending/newtypes.html#attribute-management)

EmilyBourne:
This does not seem possible. Python exposes module variables in the same way it exposes all variables. That is to say that it creates a label that can be pointed to any object. Variables can be added to a module using the [PyModule_AddObject](https://docs.python.org/3/c-api/module.html#c.PyModule_AddObject) method, however these variables are normal python objects which cannot modify an underlying variable of a basic type (it should still be possible to modify elements of an array).

This can be seen for example in the [math](https://github.com/python/cpython/blob/main/Modules/mathmodule.c#L3520) module. The math module exposes the `pi` variable as a `PyFloat` and uses it internally in functions such as the `sin` function
```
In [1]: import math                                                   

In [2]: math.pi                                                       
Out[2]: 3.141592653589793

In [3]: math.pi = True                                                

# math.pi value can be changed to anything
In [4]: math.pi                                                       
Out[4]: True

# sin function is unaffected 
In [5]: math.sin(3.141592653589793238462643383279502884197)          
Out[5]: 1.2246467991473532e-16
```

I cannot think of a way to work around this restriction without preventing the function from being called from another module

yguclu:
Here is an example for a directory that contains two files.

File **a.py**:
```python
pi = 3.14
```

File **b.pi**:
```python
from a import pi
```

Test with `ipython`:
```ipython
In [1]: import a                                                                

In [2]: a.pi                                                                    
Out[2]: 3.14

In [3]: a.pi = True                                                             

In [4]: import b                                                                

In [5]: b.pi                                                                    
Out[5]: True
```

EmilyBourne:
> Within the module scope, `math.pi` has the original float value. But within the global scope, any other module that imports `math` will see that `math.pi = True`.

Exactly, and this is the behaviour that we cannot replicate. It may be replicated on the python side, however if we translate a file which uses the constant then it will use the definition from the module scope not the global scope

Issue #741, opened 2021-02-11
Expose module variables to Python (read-only) through getters


yguclu:

EmilyBourne:
Blocked by #886

EmilyBourne:
Doc can be found [here](https://docs.python.org/3/c-api/typeobj.html#c.getattrfunc) and [here](https://docs.python.org/3/extending/newtypes.html#attribute-management)

EmilyBourne:
In order to expose arrays in Fortran a bind c getter function is required. This part is therefore blocked by #537

EmilyBourne:
Ideally we should detect when global variables are modified and update the equivalent python module variable (this will be automatic for arrays as the underlying memory is shared)

PR #949, opened 2021-08-26
Allow for instructions in a module

### LINKED ISSUE #653: Generate the module and the program in the same file
in the case of a program we currently generate two seperate files , one contains the functions and one contains the program.
**example
```python
def f():
    return 1

print(f())
```
this will generate two fortran files
The module
```fortran
module file1

implicit none

contains

function f() result(Out_0001)

  implicit none

  integer(C_INT64_T) :: Out_0001

  Out_0001 = 1_C_INT64_T
  return

end function f

end module file1

```
and the program
```fortran
program prog_prog_file1

use nparray, only: f

implicit none

print *, f()

end program prog_prog_file1
```
we would like to put them in one file as follows
```fortran
module file1

implicit none

contains

function f() result(Out_0001)

  implicit none

  integer(C_INT64_T) :: Out_0001

  Out_0001 = 1_C_INT64_T
  return

end function f
subroutine prog_prog_file1()
print *, f()
end subroutine
end module file1
```

### LINKED ISSUE #739: Allow instructions in Python module (outside of functions)
Description
--------------
As a USER, I would like to include instructions of various kind in a Python module, to be executed when the module is imported.

Acceptance criteria
------------------------
- Declare Python module variables in Fortran module, not program
- Move module instructions to `module_init` function, not program
- Only create a program if explicitly required by the user
- In Fortran module, add private variable
  ```fortran
  logical, private :: initialized = .false.
  ```
   which should be set to `.true.` by `module_init`
- Subsequent calls to `module_init` should do nothing if `initialized == True`
- `module_init` should call `module_init` on all other Pyccel-accelerated imported modules
- Make sure that `module_init` is called when the C extension module is imported

NOTE: all module variables will not be used by other modules at this stage.

Example
-----------
Let's consider the Python module `instructions.py`:

```python
def add(a: int, b: int):
    return a + b

print("Importing module 'instructions'")
x = 5
y = 6
print('x + y = ', add(x, y))
```
When the module is imported from IPython, the instructions are executed. But they are not executed a second time if the module is imported again:

```python
In [1]: import instructions                                                               
Importing module 'instructions'
x + y =  11

In [2]: import instructions                                                               

In [3]:
```
When calling `pyccel instruction.py -f` we should generate this Fortran file:

```fortran
module instructions

implicit none

logical, private :: initialized = .false.
integer(C_INT64_T) :: x
integer(C_INT64_T) :: y

contains

function add(a, b) result(Out_0001)

  implicit none

  integer(C_INT64_T) :: Out_0001
  integer(C_INT64_T), value :: a
  integer(C_INT64_T), value :: b

  Out_0001 = a + b
  return

end function add

subroutine module_init

  if (initialized .eqv. .true.) then
    return
  end if
  
  print *, 'Importing module ' // ACHAR(39) // 'instructions' // ACHAR(39)
  x = 5_C_INT64_T
  y = 6_C_INT64_T
  print *, 'x + y = ' // ' ' , add(x, y)

end subroutine module_init

end module instructions
```

### LINKED ISSUE #740: Expose module variables to other modules (Fortran side)
If a Python module uses variables from another module, Pyccel should create a Fortran module that imports those variables and makes them accessible to its functions.

NOTE: At this stage the module variables will only exist in Fortran, and not be exposed to Python.

EmilyBourne:
Allow for instructions in a module. Fixes #739, fixes #740 and fixes #653.
All statements which are not in a function are collected into an initialisation function. All statements which are bookmarked by:
```
if __name__ == '__main__':
```
are placed into a program. This program is printed in a separate file to the module to ensure everything can be compiled

**Commit Summary**
- Add .lock files to `.gitignore`
- Add `AsName.__repr__` function
- Allow `CodeBlock.insert2body` to insert multiple objects
- Add `init_func`, `free_func` and `program` to `Module` class
- Add `InProgram` class to represent `__name__ == '__main__'`
- Remove `ParserResult` which stored a `Module` and a `Program`, now (as in python) everything is in `Module`
- Add `is_private` property to `Variable`
- Fix compiling to include the fact that a file can have a program target and a module/shared library target
- Add `if __name__ == '__main__'` to all tests which generate executables
- Add `_print_Deallocate` to `FCodePrinter` to ensure modules can be deallocated
- Add test to pyccel/ folder to ensure that modules are correctly initialised when they are imported


pyccel/ast/variable.py
```diff
@@ -168,6 +172,10 @@ def __init__(
             raise TypeError('is_optional must be a boolean.')
         self._is_optional = is_optional
 
+        if not isinstance(is_private, bool):
+            raise TypeError('is_optional must be a boolean.')
```
rakati, line 176: fix the raised error to `'is_private must be a boolean.'`

pyccel/codegen/codegen.py
```diff
@@ -126,28 +117,22 @@ def modules(self):
 
     @property
     def is_module(self):
-        """Returns True if a Module."""
+        """Returns True if a Program."""
```
rakati, line 120: hmm, I don't why you changed that.

pyccel/codegen/codegen.py
```diff
@@ -207,52 +192,15 @@ def _collect_statements(self):
...
-        else:
-            raise NotImplementedError('TODO')
-
-
-        self._expr = expr
-
-        #  ...
+        self._is_program = self.ast.program is not None
+        # TODO: Expand this definition as we improve the cwrapper
+        self._is_module  = len(self.ast.funcs) + len(self.ast.interfaces) - bool(self.ast.init_func)>0
```
rakati, line 197: this is confusing for me, I know that Any Python file can be referenced as a module, so I think we should remove `is_module`, and let only `is_program` for the case of `if __name__ == '__main__'`.

pyccel/codegen/codegen.py
EmilyBourne: XD I deleted this function then put it back with copy paste, that's why it shows up as a stupid change

pyccel/codegen/codegen.py
EmilyBourne: I agree that any file can be referenced as a module, but our cwrapper doesn't work for empty modules.  I'm not sure if you can't wrap empty modules or if there's just an extra comma that shouldn't be there. In any case an empty module is not worth wrapping

pyccel/codegen/codegen.py
EmilyBourne: Fixed 5041be14

pyccel/ast/variable.py
EmilyBourne: Fixed 5041be14

pyccel/codegen/codegen.py
EmilyBourne: I've just checked, we can make an empty module. But I don't see why we should?

rakati: I think this can help in the case of a package, with an empty `__init__` file, even I don't know yet how we can use empty init files in pyccelizing packages, but only we need empty modules to be accepted by pyccel.

rakati: but one thing, if adding this feature can lead to a lot of extra changes, then I preferred to be in a separate PR.

EmilyBourne: No it's fine it barely needs any changes. Fixed here: a22420a
And `is_module` removed here: 062b5aa

EmilyBourne: Hmm, maybe not so easy. I am having problems linking with mpi

EmilyBourne: We found the problem, it was nothing to do with mpi. It is now fixed

EmilyBourne:
I have put this back to draft as the `free_func` doesn't work correctly. I forgot about this but it ought to be fixed before merging

pyccel/ast/core.py
```diff
@@ -268,6 +268,9 @@ def name(self):
     def target(self):
         return self._target
 
+    def __repr__(self):
+        return '{0} as {1}'.format(str(self.name), str(self.target))
```
rakati, line 272: can we have a better desc such as `rename old to new`?

pyccel/ast/core.py
```diff
@@ -661,16 +664,16 @@ def unravelled(self):
     def lhs(self):
         return self.body[-1].lhs
 
-    def insert2body(self, obj, back=True):
+    def insert2body(self, *obj, back=True):
         """ Insert an object to the body of the codeblock
```
rakati, line 668: I think this now support inserting many objects not only one, can you update this desc

pyccel/ast/core.py
EmilyBourne: So you would expect the object `Import(source, [AsName(old,new)])` to be printed as something like:
```
Import(from source import rename old to new)
```
?
Would this be better for a `repr`:
```
AsName(old as new)
```
?
I am not sure `rename old to new` makes the use case clear

pyccel/ast/core.py
EmilyBourne: Fixed in 9a158fcc

pyccel/ast/core.py
rakati: Ah, You are right, it's more readable than my example.

pyccel/ast/core.py
```diff
@@ -1130,7 +1153,7 @@ def __init__(
         if not iterable(variables):
             raise TypeError('variables must be an iterable')
         for i in variables:
-            if not isinstance(i, Variable):
+            if not isinstance(i, (Variable, Assign)):
                 raise TypeError('Only a Variable instance is allowed.')
```
rakati, line 1157: I think you forget to mention in the error that the Assign instance is also allowed.

pyccel/ast/core.py
```diff
@@ -1153,6 +1176,15 @@ def __init__(
             if not isinstance(i, Interface):
                 raise TypeError('Only a Inteface instance is allowed.')
 
+        if init_func is not None and not isinstance(init_func, FunctionDef):
```
rakati, line 1179: I suggest to use this way for these kind of tests
```python
if not instance(init_func, (type(None), FunctionDef)):
    # error
```

pyccel/ast/core.py
```diff
@@ -1153,6 +1176,15 @@ def __init__(
...
+            raise TypeError('init_func must be a FunctionDef')
+
+        if free_func is not None and not isinstance(free_func, FunctionDef):
```
rakati, line 1182: we can use it here too.

pyccel/ast/core.py
```diff
@@ -1153,6 +1176,15 @@ def __init__(
...
+            raise TypeError('free_func must be a FunctionDef')
+
+        if program is not None and not isinstance(program, (Program, CodeBlock)):
```
rakati, line 1185: and here

pyccel/ast/core.py
EmilyBourne: Actually I left this on purpose but I'm not sure the review process is going to let me leave it like that ^^'. Normally you should provide a `Variable`. The **only** exception is the `initialised` boolean which needs to be initialised outside of the module initialisation function, but is not a `Constant`. In this one case there needs to be some way of declaring the variable with an initial value, and this was the best that I found. If you can think of another way to do it then I suspect that would be better

pyccel/ast/core.py
```diff
@@ -1162,46 +1195,90 @@ def __init__(
...
         super().__init__()
 
     @property
     def name(self):
+        """ Name of the module
+        """
         return self._name
 
     @property
     def variables(self):
```
rakati, line 1216: can we say here module global variable instead of this desc?

pyccel/ast/core.py
EmilyBourne: Fixed fc326586. I don't know why there isn't a way to do this that doesn't involve calling `type`

pyccel/ast/core.py
EmilyBourne: Fixed fc326586

pyccel/ast/core.py
EmilyBourne: Fixed fc326586

pyccel/ast/core.py
EmilyBourne: Fixed 42b1ebf5

pyccel/ast/core.py
```diff
@@ -1302,29 +1378,44 @@ def __init__(
...
     def declarations(self):
+        """ Returns the declarations of the variables
+        """
         return [Declare(i.dtype, i) for i in self.variables]
 
+    def remove_import(self, name):
+        """ Remove an import with the given source name from the list
+        of imports
+        """
+        self._imports = tuple(i for i in self.imports if i.source != name)
```
rakati, line 1417: if the imports can be changed, why we store them in a tuple can't we use list for easy del and insert?

pyccel/codegen/printing/ccode.py
```diff
@@ -488,36 +486,56 @@ def _print_PythonLen(self, expr):
...
 
+    def _print_Header(self, expr):
+        return ''
+
     def _print_ModuleHeader(self, expr):
         name = expr.module.name
         # TODO: Add classes and interfaces
         funcs = '\n'.join('{};'.format(self.function_signature(f)) for f in expr.module.funcs)
 
+        variables = ''.join(['extern '+self._print(d) for d in expr.module.declarations if not d.variable.is_private])
```
rakati, line 497: I think `global_vars` may be a meaningful name than `varaible`.

pyccel/ast/core.py
EmilyBourne: Attribute nodes of `Basic` objects must be other `Basic` objects or tuples ([see here](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/basic.py#L61)). This simplifies things in `pyccel.ast.Basic` (I only need to check `if isinstance(obj, tuple)` rather than `if isinstance(obj, iterable)` or similar). But more importantly it discourages people from modifying these objects without thinking about it (which I apparently did) as we need to update the user nodes/attribute nodes. I have now fixed this: 14e0b1c7

pyccel/codegen/printing/ccode.py
EmilyBourne: Fixed 8e7cc1c0

pyccel/ast/core.py
EmilyBourne: It's also not a huge problem as this function is only called once (at least for now, and only when translating to python), so it doesn't matter if this function is not as optimised as it could be

pyccel/codegen/printing/ccode.py
```diff
@@ -488,36 +486,56 @@ def _print_PythonLen(self, expr):
                     PYCCEL_RESTRICTION_TODO,
                     symbol = expr, severity='fatal')
 
+    def _print_Header(self, expr):
+        return ''
```
rakati, line 490: if the _print_Header does nothing can we remove it?

pyccel/codegen/printing/ccode.py
```diff
@@ -488,36 +486,56 @@ def _print_PythonLen(self, expr):
...
+                        variables = global_variables,
                         funcs   = funcs)
 
     def _print_Module(self, expr):
+        self._current_module = expr.name
         body    = ''.join(self._print(i) for i in expr.body)
 
+        global_variables = ''.join([self._print(d) for d in expr.declarations])
+
+        if expr.program:
```
rakati, line 522: can we move this the buttom before return, and add the result to the code, the same of what we did in `fcode.py`
```python
# ...
if expr.program:
    code += self._print(expr.program)
return code
```

pyccel/codegen/printing/ccode.py
EmilyBourne: No, there are headers in the code. They don't need to be printed in C, but if there is no `_print_Header` function then we will get errors saying that the printer doesn't know how to print this object. I'm not sure why they still exist after the semantic stage. If we don't need them then we could return `EmptyNode()` from the `_print_XHeader` functions in the semantic stage, but I think checking if these are used anywhere or not is beyond the scope of this PR. I'm not sure why we didn't have this problem before. We must have had it at some point as there is already a `_print_Header` function in `fcode.py`

pyccel/codegen/printing/ccode.py
EmilyBourne: I think it needs to be before the printing of the imports so we collect any `additional_imports` although, looking at `_print_Program` it seems that it prints these itself. I will fix that

EmilyBourne: Do you want me to add a comment indicating this? Or is the comment on the next line:
```
```
sufficient?

pyccel/codegen/codegen.py
```diff
@@ -207,52 +186,13 @@ def _collect_statements(self):
     def _set_kind(self):
```
rakati, line 186: can we remove this, and move this line to init method,
then use is_program for checking if it is a program?

pyccel/codegen/pipeline.py
```diff
@@ -238,13 +238,8 @@ def handle_error(stage):
...
         return
 
-    if parser.module_parser:
-        parsers = [parser.module_parser, parser]
-        program_name = os.path.basename(os.path.splitext(parser.filename)[0])
-        module_names = [module_name, program_name]
-    else:
-        parsers = [parser]
-        module_names = [module_name]
+    parsers = [parser]
```
rakati, line 241: is it necessary to use list, we have only one item.
we may change also the name to `parser`

pyccel/codegen/pipeline.py
```diff
@@ -238,13 +238,8 @@ def handle_error(stage):
...
+    module_names = [module_name]
```
rakati, line 242: same here

pyccel/parser/semantic.py
```diff
@@ -2746,12 +2850,47 @@ def _visit_While(self, expr, **settings):
...
     def _visit_IfSection(self, expr, **settings):
-        cond = self._visit(expr.condition)
+        condition = expr.condition
+
+        name_symbol = PyccelSymbol('__name__')
+        main = LiteralString('__main__')
+        prog_check = isinstance(condition, PyccelEq) \
+                and all(a in (name_symbol, main) for a in condition.args)
+
+        if prog_check:
```
rakati, line 2860: can you check this test
```python
def func(x:int):
    if __name__ == '__main__':
        print('this is stupid thing')
    return x * x
```

pyccel/ast/core.py
rakati: I don't think I get your Idea, can you explain more.

pyccel/ast/core.py
rakati: yah, I don't why they didn't provide NoneType to use it directly.

pyccel/ast/core.py
rakati: hmm, now I see.

pyccel/codegen/printing/ccode.py
rakati: yes, this can help for better understanding.

pyccel/codegen/printing/ccode.py
rakati: I think because we are using `self._print(expr)` instead of calling directly the needed print.
anyway, this is beyond this PR, we will look at that after.
we will need to create an issue for clearing things (include unused declared variables).

EmilyBourne: > I think because we are using self._print(expr) instead of calling directly the needed print.

Yes, this is necessary to avoid an enormous number of if blocks

> anyway, this is beyond this PR, we will look at that after.

:+1:

> include unused declared variables

I'm not sure we want to do that given that we always try to produce code as close as possible to the user's code, and that the aim of this PR (and the related issues #740, #741, #742 ) is to allow modules such as:
```
g = 9.81
e = 3.14
```

pyccel/codegen/printing/ccode.py
EmilyBourne: Fixed 39e59eb5

pyccel/codegen/codegen.py
EmilyBourne: Fixed 0eb7808c6b4c02ff245f9683cf18d58fa8b6ca01

pyccel/codegen/pipeline.py
EmilyBourne: No, it is not necessary, but have fun reviewing the changes that come with getting rid of it ^^. See 14683d69362a2987b6534c0252ae63646553b3e9
I just removed these 2 lines and got rid of the loop, but it's going to look like a huge change. Don't feel like you need to check every line XD

pyccel/codegen/pipeline.py
EmilyBourne: See above

pyccel/parser/semantic.py
EmilyBourne: I get:
```
pyccel:
 |fatal [codegen]: main_odd.py| _print_InProgram is not yet implemented for language : Fortran
Pyccel has encountered syntax that has not been implemented yet. Please create an issue at https://github.com/pyccel/pyccel/issues and provide a small example of your problem. Do not forget to specify your target language (<pyccel.ast.core.InProgram object at 0x7fd07b8934c0>
```
So it is correctly creating this object in the function rather than the module and is then not unpacking the contents into the module. The statement `if __name__ == '__main__':` then raises an error when we try to print it.

I think this is acceptable behaviour for something that a user is very unlikely to do, but if you prefer I could add `_print_InProgram` methods to raise a more specific error?

pyccel/ast/core.py
EmilyBourne: Currently the `init_function` is created [here](https://github.com/pyccel/pyccel/blob/devel-issue739/pyccel/parser/semantic.py#L1653-L1663)
It always has the same contents, which are:
```
def init_function():
    if initialised:
        # module_body
```
It would be possible to create the function in `Module.__init__` from module_body. However keeping the name clash avoidance would be harder and you would still need to collect the function afterwards to add it to the scope

EmilyBourne: I think I found a better solution: 0b904013

pyccel/codegen/pipeline.py
rakati: great, sometimes we need huge changes to make things better.

pyccel/parser/semantic.py
rakati: while python letting that for the user, we may do too.
the function can have different behavior if it's imported or called inside the main block.
but as you said this is rare, so I think it's better to raise an error and let the user try another good way instead of this stupid method xD.

rakati (approved):

pyccel/parser/semantic.py
```diff
@@ -3452,6 +3594,13 @@ def _insert_obj(location, target, obj):
             else:
                 targets = expr.target
 
+            if import_init:
+                result  = FunctionCall(import_init,[],[])
+                targets.add(import_init.name)
```
saidctb, line 3599: shouldn't the import_init be added to the `init_func` imports and not to the  module imports, to avoid name clach

saidctb:
Thanks for the PR, this really great!
I have a question, if you could clarify this for me
let's say that we have two programs `mod1` and `mod2` , `mod2` import from `mod1`
do we have access to the imported variables or functions  in mod2 from mod1 when we create the shared library for mod2 ?

pyccel/parser/semantic.py
EmilyBourne: Hmm I hadn't considered name clashes here. I agree that `import_init` should perhaps be added to `init_func` if it is imported at the module level, but if it is imported inside a function then that is not equivalent. Maybe we could rename it if there is a collision?

EmilyBourne:
> Thanks for the PR, this really great!
> I have a question, if you could clarify this for me
> let's say we have two programs `mod1` and `mod2` , `mod2` import from `mod1`
> do we have access to the imported variables or functions in mod2 from mod1 when we create the shared library for mod2 ?

I am not certain what you mean. Assuming we have:
mod1.py:
```
def f():
   pass
```
mod2.py
```
from mod1 import f

def g():
    pass
```
Do you mean, can we do `from mod2 import f` after pyccelising? If so then no this is not possible, the information to implement it is available but this would unnecessarily bloat our cwrappers

pyccel/parser/semantic.py
saidctb: yes , we could for sure, I just thought it's cleaner and might avoid name clashes, but if it's a lot of works, then there is no need for the moment

saidctb:
> > Thanks for the PR, this really great!
> > I have a question, if you could clarify this for me
> > let's say we have two programs `mod1` and `mod2` , `mod2` import from `mod1`
> > do we have access to the imported variables or functions in mod2 from mod1 when we create the shared library for mod2 ?
> 
> I am not certain what you mean. Assuming we have:
> mod1.py:
> 
> ```
> def f():
>    pass
> ```
> 
> mod2.py
> 
> ```
> from mod1 import f
> 
> def g():
>     pass
> ```
> 
> Do you mean, can we do `from mod2 import f` after pyccelising? If so then no this is not possible, the information to implement it is available but this would unnecessarily bloat our cwrappers

yes that's what I mean , because this is the python behavior, but we can decide not to allow that

saidctb:
can we import module variables now with the shared library?

EmilyBourne:
> can we import module variables now with the shared library?

No. This is for the next PR. See issues #741 and #742. I thought this PR was large enough already XD

pyccel/parser/semantic.py
EmilyBourne: Potential name collisions have now been handled

saidctb (approved):

EmilyBourne:
While having `module__init` makes sense having `Program`s in the same file is problematic. For example it would prevent us from compiling the following:
mod1.py
```
def f():
   print("hello")

if __name__ == '__main__':
    f()
```
mod2.py
```
def g():
   from mod1 import f
   f()
   print("world")

if __name__ == '__main__':
    g()
```
As the compiler will raise an error about 2 mains. I am therefore going to separate this back into 2 files

saidctb:
> While having `module__init` makes sense having `Program`s in the same file is problematic. For example it would prevent us from compiling the following:
> mod1.py
> 
> ```
> def f():
>    print("hello")
> 
> if __name__ == '__main__':
>     f()
> ```
> 
> mod2.py
> 
> ```
> def g():
>    from mod1 import f
>    f()
>    print("world")
> 
> if __name__ == '__main__':
>     g()
> ```
> 
> As the compiler will raise an error about 2 mains. I am therefore going to separate this back into 2 files

the program will just call the module_init

EmilyBourne:
> > While having `module__init` makes sense having `Program`s in the same file is problematic. For example it would prevent us from compiling the following:
> > mod1.py
> > ```
> > def f():
> >    print("hello")
> > 
> > if __name__ == '__main__':
> >     f()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > mod2.py
> > ```
> > def g():
> >    from mod1 import f
> >    f()
> >    print("world")
> > 
> > if __name__ == '__main__':
> >     g()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > As the compiler will raise an error about 2 mains. I am therefore going to separate this back into 2 files
> 
> the program will just call the module_init

The program will call the `module_init` and do anything that is in a `if __name__ == '__main__':` block

yguclu (commented):
Excellent! Thank you Emily

pyccel/codegen/codegen.py
```diff
@@ -203,57 +176,12 @@ def _collect_statements(self):
...
-
-        #  ...
-
     def doprint(self, **settings):
         """Prints the code in the target language."""
         if not self._printer:
             self.set_printer(**settings)
-        self._code = self._printer.doprint(self.expr)
-        return self._code
+        return self._printer.doprint(self.ast)
```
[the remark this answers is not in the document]
EmilyBourne: I think the settings are mostly used to determine the language. It is a little strange that this is ignored, but this is only the case in the function `doprint` not in the function `export` below. The function `doprint` is only used in the codegen tests. It is not used by pyccel itself

pyccel/codegen/compiling/basic.py
```diff
@@ -143,10 +135,10 @@ def module_target(self):
         return self._module_target
 
     @property
-    def target(self):
+    def program_target(self):
         """ Returns the file to be generated by the compilation step
         """
-        return self._target
+        return self._prog_target
```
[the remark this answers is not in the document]
EmilyBourne: Fixed e4aba6d5

pyccel/codegen/codegen.py
```diff
@@ -265,15 +193,26 @@ def export(self, filename=None, **settings):
...
+
+        # print module
+        code = self._printer.doprint(self.ast)
         with open(filename, 'w') as f:
-            for line in self._code:
+            for line in code:
                 f.write(line)
 
-        return filename
+        prog_filename = None
```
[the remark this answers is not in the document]
EmilyBourne: Fixed e4aba6d5

pyccel/ast/core.py
```diff
@@ -1153,6 +1176,16 @@ def __init__(
...
 
+        NoneType = type(None)
+        if not isinstance(init_func, (NoneType, FunctionDef)):
+            raise TypeError('init_func must be a FunctionDef')
+
+        if not isinstance(free_func, (NoneType, FunctionDef)):
+            raise TypeError('free_func must be a FunctionDef')
+
+        if not isinstance(program, (NoneType, Program, CodeBlock)):
+            raise TypeError('program must be a Program')
```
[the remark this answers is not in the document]
EmilyBourne: Fixed e4aba6d5

yguclu (approved):

pyccel/codegen/codegen.py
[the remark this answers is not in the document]
yguclu: But could we possibly want to call `doprint()` multiple times with different languages?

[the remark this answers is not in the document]
EmilyBourne: I'm not sure. I actually tried to remove this function due to its lack of use. But then I came across the problem with the codegen tests. With the codegen tests set up as they are we shouldn't need to call `doprint()` multiple times with different languages. Elsewhere we should be using `export` anyway (which could plausibly call multiple languages if for example we wanted to provide a .h file for each bind c interface)

Issue #940, opened 2021-08-17
Printing nd arrays in Fortran


EmilyBourne:
**Describe the bug**
All arrays are printed the same way in Fortran regardless of their rank. For arrays with rank>1 this is not very clear. In particular it is misleading if the order is not C.

**To Reproduce**
Provide code to reproduce the behavior:
```python
import numpy as np

x = np.array([[1,2,3],[4,5,6],[7,8,9]], order='C')
y = np.array([[1,2,3],[4,5,6],[7,8,9]], order='F')

print(x)
print(y)
```

The output for python is:
```bash
[[1 2 3]
 [4 5 6]
 [7 8 9]]
[[1 2 3]
 [4 5 6]
 [7 8 9]]
```

The output of the pyccelized code is:
```
                    1                    2                    3                    4                    5                    6                    7                    8                    9
                    1                    4                    7                    2                    5                    8                    3                    6                    9
```
The generated code is:
```fortran
program prog_pr

  use, intrinsic :: ISO_C_BINDING

  implicit none

  integer(C_INT64_T), allocatable :: x(:,:)
  integer(C_INT64_T), allocatable :: y(:,:)

  allocate(x(0:2_C_INT64_T, 0:2_C_INT64_T))
  x = reshape([[1_C_INT64_T, 2_C_INT64_T, 3_C_INT64_T], [4_C_INT64_T, &
      5_C_INT64_T, 6_C_INT64_T], [7_C_INT64_T, 8_C_INT64_T, 9_C_INT64_T &
      ]], [3_C_INT64_T, 3_C_INT64_T])
  allocate(y(0:2_C_INT64_T, 0:2_C_INT64_T))
  y = transpose(reshape([[1_C_INT64_T, 2_C_INT64_T, 3_C_INT64_T], [ &
      4_C_INT64_T, 5_C_INT64_T, 6_C_INT64_T], [7_C_INT64_T, 8_C_INT64_T &
      , 9_C_INT64_T]], [3_C_INT64_T, 3_C_INT64_T]))
  print *, x
  print *, y

end program prog_pr
```

**Expected behaviour**
The array should be printed row-wise (as in C) to ensure that the matrix appears as expected

**Language**
Fortran

**Additional context**
This strange behaviour is causing confusion in PR #939

PR #908, opened 2021-06-11
Clear up ambiguity by removing ValuedVariable

EmilyBourne:
Remove `ValuedVariable` and replace it with `FunctionDefArgument` (formerly just `Argument`) and `FunctionCallArgument` which contain variables (fixes #876) . 

**Commit Summary**
- Remove `ValuedArgument` class. There is no advantage to having both `ValuedArgument` and `Argument`
- Add `FunctionCallArgument` class
- For consistency and clarity, rename `Argument` as `FunctionDefArgument`
- Allow `Variable` as name/var of an `Argument` (function def)
- Ensure all arguments to `FunctionCall` are of type `FunctionCallArgument`
- Add `Declare.__repr__` for debugging
- Remove `ValuedVariable`
- Update `Constant` to implement inherited methods (which are no longer inherited)
- Update parsers and printers to use new classes
- Activate python tests blocked by problematic use of `ValuedVariable`
- Activate C tests blocked by problematic use of `ValuedVariable`
- Remove `ValuedFunctionAddress ` class which is no longer used
- Handle `sep` and `end` parameters of `print()` in Python printer (fixes #874).

PR #953, opened 2021-08-31
Improve printing for Fortran

jalalium:
This PR aims to improve the printing capabilities of Fortran.

The printing of all expressions in Fortran was done through the use of `', '.join(['print *', *args])`. This approach is limited in the following ways:

- Padding was added automatically in many cases and this hinders readability.
- A new line is printed even if not sought.
- All arrays are printed the same way regardless of their rank.
- Although formatting can be achieved with print, it was not used.

All these issues were addressed with the use of `write` whenever it is possible. Expressions are processed as they would in C to generate a format for the arguments. this format will be passed to `write` for an output as close as possible to C and Python. In addition, arrays with rank > 1 are printed as they would in C.

Solves #982 (Can't print empty string).
Solves #940 (Printing nd arrays in Fortran).

# Examples
The following examples illustrate the changes:

## Example 1
```python
print("mcha bba lsbata")
```
**before:**
```fortran
  print *, 'mcha bba lsbata'
```
prints:
```fortran
  write(*, '(A)', advance="yes") 'mcha bba lsbata'
```

## Example 2
```python
print('The first sentence', end='. ')
print('The second sentence.')
```
prints:
```
The first sentence. The second sentence.
```
**before:**
```fortran
print *, 'The first sentence' // ' '
print *, 'The second sentence.'
```
prints:
```
 The first sentence 
 The second sentence.
```
**after:**
```fortran
write(*, '(A A)', advance="no") 'The first sentence' , '. '
      write(*, '(A)', advance="yes") 'The second sentence.'
```
prints:
```
The first sentence. The second sentence.
```

### Example 3
```python
import numpy as np
a = np.array([[1,2,3],[4,5,6]])
print(a)
```
prints:
```
[[1 2 3]
 [4 5 6]]
```
**before:**
```fortran
allocate(a(0:2_i64, 0:1_i64))
a = reshape([[1_i64, 2_i64, 3_i64], [4_i64, 5_i64, 6_i64]], [3_i64 &
        , 2_i64])
print *, a
```
prints:
```
                    1                    2                    3                    4                    5                    6
```
**after:**
```fortran
a = reshape([[1_i64, 2_i64, 3_i64], [4_i64, 5_i64, 6_i64]], [3_i64 &
            , 2_i64])
      write(*, '(A)', advance="no") '['
      do i = 0_i64, 0_i64, 1_i64
        write(*, '(A)', advance="no") '['
        do i_0001 = 0_i64, 1_i64, 1_i64
          write(*, '(I0 A)', advance="no") a(i_0001, i) , ' '
        end do
        write(*, '(I0 A)', advance="no") a(2_i64, i) , ']'
        write(*, '()', advance="yes")
      end do
      write(*, '(A)', advance="no") '['
      do i_0002 = 0_i64, 1_i64, 1_i64
        write(*, '(I0 A)', advance="no") a(i_0002, 1_i64) , ' '
      end do
      write(*, '(I0 A)', advance="no") a(2_i64, 1_i64) , ']'
      write(*, '(A)', advance="no") ']'
      write(*, '()', advance="yes")
```
prints:
```
[[1 2 3]
[4 5 6]]
```


EmilyBourne (changes requested):
Please merge in the master and fix any resulting collisions (if you need more detail about what is being merged see PR #908 )

EmilyBourne:
Some tests are failing because there is a variable called `size` in the python file. This masks the fortran intrinsic function `size`. This is described in issue #227 . Please find a new name for this variable to avoid the problem for now and get the tests to pass again. We will fix it at some point in a separate PR

pyccel/codegen/printing/fcode.py
```diff
@@ -452,28 +456,112 @@ def _print_Import(self, expr):
...
+        sep = ' '
+        code = ''
+        empty_end = FunctionCallArgument(LiteralString(''), 'end')
+        space_end = FunctionCallArgument(LiteralString(' '), 'end')
+        kwargs = [f for f in expr.expr if f.has_keyword]
+        for f in kwargs:
+            if f.keyword == 'sep':
+                sep = str(f.value)
+            elif f.keyword == 'end':
+                end = str(f.value)
```
rakati, line 469: I noticed that `kwargs` is used only in the forloop.
it may be better to use only
```python
for f in expr.expr:
     if not f.has_keyword:     continue
     if f.keyword == 'sep':    sep = str(f.value)
     elif f.keyword == 'end':  end = str(f.value)
```
without creating kwargs list only for this forloop

pyccel/codegen/printing/fcode.py
```diff
@@ -452,28 +456,112 @@ def _print_Import(self, expr):
...
+        def formatted_args_to_print(args_format, args, end):
+            if args_format == ['*']:
+                return ', '.join(['print *', *args]) + '\n'
+            args_format = ' A '.join(args_format)
+            new_line = "yes" if end.count('\n') > 0 else "no"
+            end = end.replace('\n', '')
+            args_code = ', " " ,'.join([*args])
+            if end != '':
+                return "write(*, '({},A)',advance=\"{}\") {}, \"{}\"\n".format(args_format, new_line, args_code, end)
+            return "write(*, '({})',advance=\"{}\") {}\n".format(args_format, new_line, args_code)
```
rakati: this code is broken for multiple line.
test with `print("alloha", 'toto', sep='performed by', end=''\n grr  \n  gr gr \n'')`.

EmilyBourne: @jalalium I have marked this PR as a draft as it seems fixing this is not trivial

jalalium: @rakati thank you, i will try to fix it.
@EmilyBourne thank you!

pyccel/codegen/printing/fcode.py
jalalium: @rakati i tried to keep the same structure as the one used in ccode.py. shall i change both of them?

rakati: you may not need, because Fortran and C are different, so there may be an easy way to do it instead of using the same structure as in `ccode`.

pyccel/codegen/printing/fcode.py
rakati: you may find some problems if the end or sep contains special characters, take a look at the  `_print_LiteralString`  method in `fcode.py` to know how to deal with that.

rakati: don't forget to activate the formated print tests. you will find the test related to that at `tests/pyccel/test_pyccel.py`.
just remove pytest xfail mark for the function `test_print_sp_and_end`.

pyccel/codegen/printing/fcode.py
```diff
@@ -2697,12 +2790,40 @@ def split_pos_code(line, endpos):
...
+                quote = 0
+                double_quote = 0
+                mine = []
+                for idx, c in enumerate(line):
+                    if c == '\'':
+                        quote += 1
+                    if c == '\"':
+                        double_quote += 1
+                    if c == ' ' and quote%2 == 0 and double_quote %2 == 0 and idx > 72 and (not pos or idx > pos[-1] + 65):
+                        pos.append(idx)
```
EmilyBourne, line 2806: I wonder if some of this can be accomplished with [regexes](https://docs.python.org/3/library/re.html)

pyccel/codegen/printing/fcode.py
```diff
@@ -2697,12 +2790,40 @@ def split_pos_code(line, endpos):
...
+                    hunk = line[:(cut-last_cut)].rstrip()
+                    line = line[(cut-last_cut):].lstrip()
+                    if line:
+                        hunk += trailing
+                    if cut == pos[0]:
+                        result.append(hunk)
+                        mine.append(hunk)
+                    else:
+                        result.append("%s%s"%("      " , hunk))
+                        mine.append("%s%s"%("      " , hunk))
```
EmilyBourne, line 2818: I think this could be written more clearly as:
```
result.append("      " + hunk)
```
I don't see what advantage string formatting is giving you here

pyccel/codegen/printing/fcode.py
jalalium: Sorry, i just wanted to make sure the tests worked before cleaning up the code. I will refactor it using regex

pyccel/codegen/printing/fcode.py
jalalium: that was the initial syntax. I ll rectify it.

pyccel/codegen/printing/fcode.py
EmilyBourne: No problem, do it in whatever order works best for you, I just thought it was worth mentioning regexes sooner rather than later because they can be a pain to use and will probably replace a large chunk of code, so it may not be worth working at fixing this bit if you are going to replace it as soon as it works

EmilyBourne: Looks like it's working now anyway :+1:

jalalium: Finally, i still don't think it's enough. Working on the parity of quote and double quote to know when it is safe to break a line will not account for the use of \' or \'' inside a quote or a double quote. I need to make sure I am only counting the actual quotes.

EmilyBourne: You need to activate the tests for this to make sure you are handling these cases

EmilyBourne: [This file](https://github.com/pyccel/pyccel/blob/master/tests/pyccel/scripts/print_strings.py) and [this file](https://github.com/pyccel/pyccel/blob/master/tests/pyccel/scripts/print_sp_and_end.py) contain lots of test cases that seem relevant to this PR. You need to make sure these are activated [here](https://github.com/pyccel/pyccel/blob/master/tests/pyccel/test_pyccel.py#L612-L623)

You could also add additional tests with single/double quotes and longer lines

EmilyBourne: I think it should be manageable with a regex that looks something like [this](https://stackoverflow.com/a/5696141)

EmilyBourne: I recommend using [raw strings](https://www.askpython.com/python/string/python-raw-strings) for this

pyccel/codegen/printing/fcode.py
```diff
@@ -460,32 +460,71 @@ def _print_PythonPrint(self, expr):
...
+                    for idx in range(len(mask_cuts) - 1):
+                        if mask_cuts[idx + 1] - mask_cuts[idx] < 8:
+                            valid = False
+                        mask_longest_chunk = max(mask_longest_chunk, mask_cuts[idx + 1] - mask_cuts[idx])
+                    if valid and mask_longest_chunk < longest_chunk:
+                        cuts = mask_cuts
+                        longest_chunk = mask_longest_chunk
+                if longest_chunk != length:
+                    return [line[a + 1 : b + 1] for a, b in zip([-1, *cuts], cuts)]
+                return [line[i:i + int(length / 3 + 1)] for i in range(0, length, int(length / 3 + 1))]
```
jalalium, line 497: This function will find the white spaces in a long string ( > 30 characters) then it will try all possible choices of white spaces to choose as a splitting point, out of these it will pick the splitting that minimizes the length of the longest chunk. Do you think we should choose a better heuristic maybe the number of chunks or the average length of all chunks? If no white spaces are present, then the string is evenly cut into 3 parts.

pyccel/codegen/printing/fcode.py
```diff
@@ -2790,39 +2828,40 @@ def split_pos_code(line, endpos):
...
-                        pos.append(idx)
+            elif len(line) > 72 and ('"' in line[72:] or "'" in line[72:]):
+                # finds all spaces outside double quotes
+                pattern = re.compile(' (?=(?:[^"]*"[^"]*")*[^"]*$)')
+                spaces_outside_dq = []
+                for match in pattern.finditer(line):
+                    spaces_outside_dq.append(match.start())
+                # finds all spaces outside quotes
+                pattern = re.compile(' (?=(?:[^\']*\'[^\']*\')*[^\']*$)')
+                spaces_outside_q = []
```
jalalium, line 2841: Is there a way to run these two regex simultaneously? Also are these two cases the only ones where breaking a line at a white space would be problematic?

EmilyBourne: I believe the syntax would be ` (?=(?:[^"]*"[^"]*")*[^"]*$)|(?=(?:[^\']*\'[^\']*\')*[^\']*$)`. You can use a [regex tester](https://regex101.com/) to check your regex (the one I linked to also includes an explanation of what each part of the regex is doing

pyccel/codegen/printing/fcode.py
EmilyBourne: So the following:
```
print("Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore ")
```
would be translated to:
```
write (*,continue=no) "Lorem ipsum dolor sit amet, consectetur adipiscing elit, "
write (*,continue=yes) "sed do eiusmod tempor incididunt ut labore et dolore "
```
?
This heuristic seems fine to me, if perhaps overly complicated. I would be fine with printing:
```
write (*,continue=no) "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor "
write (*,continue=yes) "incididunt ut labore et dolore "
```
i.e. just truncating at the last acceptable space. Testing all possible combinations is potentially quite costly and I'm not sure that the gain in readability is worth the extra cost

pyccel/codegen/printing/fcode.py
```diff
@@ -452,28 +455,151 @@ def _print_Import(self, expr):
...
+                    end = str(f.value)
+        args_format = []
         args = []
-        n = len(expr.expr)
-        for j, f in enumerate(expr.expr):
+        orig_args = [f for f in expr.expr if not f.has_keyword]
+
+        def formatted_args_to_print(fargs_format, fargs, end):
+
+            def split_long_argument(line):
```
EmilyBourne, line 475: Please add docstrings for these functions.

Functions in functions are usually used to handle small functionalities that are only pertinent to the parent function. Once they become to large to glance over they should be independent functions. Here I am struggling to tell which function ends where so I think these functions are causing readability problems

pyccel/codegen/printing/fcode.py
```diff
@@ -2697,12 +2828,41 @@ def split_pos_code(line, endpos):
         result = []
         trailing = ' &'
         for line in lines:
-            if len(line)>72 and ('"' in line[72:] or "'" in line[72:] or '!' in line[:72]):
+            if len(line) > 72 and '!' in line[:72]:
                 result.append(line)
-
-            elif len(line)>72:
+            elif len(line) > 72 and ('"' in line[72:] or "'" in line[72:]):
```
EmilyBourne, line 2833: Is this used given that you are handling the same case with the `split_long_argument` function? Do you need both?

pyccel/codegen/printing/fcode.py
```diff
@@ -452,28 +455,151 @@ def _print_Import(self, expr):
...
+                    for idx in range(len(mask_cuts) - 1):
+                        if mask_cuts[idx + 1] - mask_cuts[idx] < 8:
+                            valid = False
+                        mask_longest_chunk = max(mask_longest_chunk, mask_cuts[idx + 1] - mask_cuts[idx])
+                    if valid and mask_longest_chunk < longest_chunk:
+                        cuts = mask_cuts
+                        longest_chunk = mask_longest_chunk
+                if longest_chunk != length:
+                    return [line[a + 1 : b + 1] for a, b in zip([-1, *cuts], cuts)]
+                return [line[i:i + int(length / 3 + 1)] for i in range(0, length, int(length / 3 + 1))]
```
EmilyBourne, line 497: Please add comments to this function to explain what is happening

pyccel/codegen/printing/fcode.py
```diff
@@ -452,28 +455,151 @@ def _print_Import(self, expr):
...
+        def formatted_args_to_print(fargs_format, fargs, end):
+
+            def split_long_argument(line):
+                line = line[1 : len(line) - 1]
+                length = len(line)
+                longest_chunk = length
+                cuts = []
+                positions = [idx for idx in range(length) if line[idx] == ' ']
+                if len(positions) > 12:
+                    positions = positions[int(len(positions) / 4) : 12]
```
EmilyBourne, line 482: You can use the integer division operator `//` instead of casting results to integers, i.e. `len(positions)//4`

EmilyBourne, line 482: I don't understand what you are doing here. If there are more than 12 spaces in the line then only consider some of the spaces? What happens for a line like:
```
print("a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a a")
```
This line contains 52 spaces so after the lines above `positions` is an empty list

pyccel/codegen/printing/fcode.py
```diff
@@ -452,28 +455,151 @@ def _print_Import(self, expr):
...
+                for mask in range(1, 1<<len(positions)):
+                    mask_cuts = [positions[idx] for idx in range(len(positions)) if (1 << idx)&mask != 0 ]
```
EmilyBourne, line 484: Please add a comment to explain what this condition does.

Isn't it equivalent to: `(1 << idx) == mask` ?

EmilyBourne, line 484: Codacy seems to have missed this, but in python it would be better to use enumerate here, i.e.:
```
mask_cuts = [pos for idx,pos in enumerate(positions) if (1 << idx)&mask != 0 ]
```

pyccel/codegen/printing/fcode.py
```diff
@@ -452,28 +455,151 @@ def _print_Import(self, expr):
...
+                    mask_cuts.append(length)
+                    valid = True
+                    mask_longest_chunk = 0
+                    for idx in range(len(mask_cuts) - 1):
+                        if mask_cuts[idx + 1] - mask_cuts[idx] < 8:
+                            valid = False
+                        mask_longest_chunk = max(mask_longest_chunk, mask_cuts[idx + 1] - mask_cuts[idx])
```
EmilyBourne, line 491: I suspect there is a way to do this without the for loop using numpy arrays. E.g:
```
mask_cuts = np.array(mask_cuts)
mask_cut_diffs = mask_cuts[1:]-mask_cuts[:-1]
max_longest = max(mask_cut_diffs)
valid = any(mask_cut_diffs < 8)
```

pyccel/codegen/printing/fcode.py
```diff
@@ -452,28 +455,151 @@ def _print_Import(self, expr):
...
+                    if valid and mask_longest_chunk < longest_chunk:
+                        cuts = mask_cuts
+                        longest_chunk = mask_longest_chunk
```
EmilyBourne, line 494: Why do we need all the chunks to have at least 8 characters?

pyccel/codegen/printing/fcode.py
EmilyBourne, line 497: Do any of our existing tests test the entirety of this function?

pyccel/codegen/printing/fcode.py
```diff
@@ -452,28 +455,151 @@ def _print_Import(self, expr):
...
+                if longest_chunk != length:
+                    return [line[a + 1 : b + 1] for a, b in zip([-1, *cuts], cuts)]
+                return [line[i:i + int(length / 3 + 1)] for i in range(0, length, int(length / 3 + 1))]
+
+            if fargs_format == ['*']:
+                return ', '.join(['print *', *fargs]) + '\n'
```
EmilyBourne, line 500: What case is this handling? Isn't it missing the separator?

pyccel/codegen/printing/fcode.py
```diff
@@ -452,28 +455,151 @@ def _print_Import(self, expr):
...
+            separator = self._print_LiteralString(LiteralString(sep))
+            args_formatting = ''
+            args_code = ''
+            for a_f, a_c in zip(fargs_format, fargs):
+                if a_f == 'A' and len(a_c) > 30 and a_c[0] == '\'' and a_c.count('\'') == 2:
```
EmilyBourne, line 505: You can also write `'\''` as `"'"`. In python there is no difference between characters and strings

pyccel/codegen/printing/fcode.py
```diff
@@ -452,28 +455,151 @@ def _print_Import(self, expr):
...
+                    chunks = split_long_argument(a_c)
+                    whole_format = ' '.join(['A'] * len(chunks))
+                    arg_concat = ' , '.join('\'{}\''.format(chunk) for chunk in chunks)
+                    args_code += ('' if args_formatting == ''
+                                  else (' , ' + separator + ' , ' if separator != '' else ' , ')) + arg_concat
+                    args_formatting += ('' if args_formatting == ''
+                                        else (' A ' if separator != '' else ' ')) + whole_format
+                else:
+                    args_code += ('' if args_formatting == ''
+                                  else (' , ' + separator + ' , ' if separator != '' else ' , ')) + (a_c if a_c != '' else '\'\'')
```
EmilyBourne, line 515: I wonder if it is easier/clearer to add the `', '` after the for loop by collecting the `args_code` as a list and then using `join`. This line is quite hard to follow with all theses ifs

pyccel/codegen/printing/fcode.py
```diff
@@ -452,28 +455,151 @@ def _print_Import(self, expr):
...
+                    args_formatting += ('' if args_formatting == ''
+                                        else (' A ' if separator != '' else ' ')) + a_f
+            if args_formatting == '':
+                return "write(*, '(A)', advance=\"no\") {}\n".format(self._print_LiteralString(LiteralString(end)))
+            if end != '':
+                return "write(*, '({} A)', advance=\"no\") {} , {}\n".format(args_formatting, args_code, self._print_LiteralString(LiteralString(end)))
+            return "write(*, '({})', advance=\"no\") {}\n".format(args_formatting, args_code)
```
EmilyBourne, line 522: I don't see a line with `advance=\"yes\"`. Have I missed it? Or are the new lines handled with the `end` parameter? I think it would be better to use `advance=\"yes\"` to ensure that the output is flushed

pyccel/codegen/printing/fcode.py
```diff
@@ -452,28 +455,151 @@ def _print_Import(self, expr):
...
+        sep = ' '
+        code = ''
+        empty_end = FunctionCallArgument(LiteralString(''), 'end')
+        space_end = FunctionCallArgument(LiteralString(' '), 'end')
+        for f in expr.expr:
+            if f.has_keyword:
+                if f.keyword == 'sep':
+                    sep = str(f.value)
+                elif f.keyword == 'end':
+                    end = str(f.value)
```
EmilyBourne, line 468: I think we should add an `else` section which raises an error or a warning. Currently if I do:
```
print("hello world", flush=True)
```
Then pyccel executes without complaining and I am not aware that my keyword was ignored

pyccel/codegen/printing/fcode.py
```diff
@@ -452,28 +455,151 @@ def _print_Import(self, expr):
...
+                                  for_loop,
+                                  PythonPrint([FunctionCallArgument(f[max_index]), for_end])],
+                                 unravelled=True)
+                code += self._print(body)
+            else:
+                arg_format, arg = self.get_print_format_and_arg(f)
+                args_format.append(arg_format)
+                args.append(arg)
+        if args_format:
+            code += formatted_args_to_print(args_format, args, end)
```
EmilyBourne, line 575: Is this `if` block needed if you don't do:
```
if len(orig_args) == 0:
            return formatted_args_to_print(args_format, args, end)
```
above?

pyccel/codegen/printing/fcode.py
```diff
@@ -452,28 +455,151 @@ def _print_Import(self, expr):
...
-                args.append("'{}'".format(f))
+                arg_format, arg = self.get_print_format_and_arg(f)
+                args_format.append(arg_format)
+                args.append(arg)
             elif isinstance(f, PythonTuple):
-                for i in f:
-                    args.append(self._print(i))
+                arg_format, arg = self.get_print_format_and_arg(f)
+                args_format.append(arg_format)
+                args.append(arg)
```
EmilyBourne, line 539: We seem to have lost the loop here, are you sure that this doesn't pose any problems? I don't see any handling of tuples in the function `get_print_format_and_arg`

EmilyBourne, line 539: These if blocks do the same thing, why are they needed?

pyccel/codegen/printing/fcode.py
```diff
@@ -452,28 +455,151 @@ def _print_Import(self, expr):
...
-                for i in f:
-                    args.append(self._print(i))
-            elif f.dtype is NativeString() and j != n-1:
-                args.append("{} // ' ' ".format(self._print(f)))
-            else:
+                arg_format, arg = self.get_print_format_and_arg(f)
+                args_format.append(arg_format)
+                args.append(arg)
+            elif f.dtype is NativeString() and f != expr.expr[-1]:
+                args_format.append('A')
```
EmilyBourne, line 545: If I do:
```
a = 'hello'
print(a, " is the same as ", a)
```
then this condition will not be evaluated as you expect. You should test `i` not `f`

pyccel/codegen/printing/fcode.py
```diff
@@ -452,28 +455,151 @@ def _print_Import(self, expr):
...
                 args.append(self._print(f))
+            elif isinstance(f.rank, int) and f.rank > 0:
```
EmilyBourne, line 547: Have you found cases where the rank is not an int?

pyccel/codegen/printing/fcode.py
EmilyBourne, line 468: I am concerned about the casting using the `str()` function here.
I don't think the `sep` or `end` parameters have to be `LiteralString` objects. Plus the `__str__` function is used for readability of expressions so it does not guarantee that characters are written in a way that makes sense for fortran

pyccel/codegen/printing/fcode.py
```diff
@@ -205,6 +206,7 @@ def __init__(self, parser, prefix_module = None):
 
         self._additional_code = None
         self._additional_imports = set([])
+        self._additional_declare = []
```
EmilyBourne, line 209: Why do we need this? There are already ways to declare additional variables in this file. Search for `self._additional_code ` to see how this is handled elsewhere

pyccel/codegen/printing/fcode.py
```diff
@@ -452,28 +455,151 @@ def _print_Import(self, expr):
...
+    def get_print_format_and_arg(self,var):
+        type_to_format = {('real'): 'F0.12',
+                          ('complex'): '"(",F0.12," + ",F0.12,")"',
+                          ('integer'): 'I0',
+                          ('logical'): 'A',
+                          ('string'): 'A'}
+        var_type = self._print(var.dtype)
+        if var_type not in type_to_format:
+            if self._print(var.dtype).find("character") != -1:
+                return 'A', self._print(var)
```
EmilyBourne, line 587: Why do you recalculate `self._print(var.dtype)`? Can't you use `var_type.find("character")` ?

What are you trying to handle here? A class with character in the name?

pyccel/codegen/printing/fcode.py
```diff
@@ -452,28 +455,151 @@ def _print_Import(self, expr):
...
+
+        try:
+            arg_format = type_to_format[var_type]
+        except KeyError:
+            errors.report("{} type is not supported currently".format(var.dtype), severity='fatal')
+
+        if var.dtype is NativeComplex():
+            arg = '{}, {}'.format(self._print(NumpyReal(var)), self._print(NumpyImag(var)))
+        elif var.dtype is NativeBool():
+            arg = 'merge("True ", "False", {})'.format(self._print(var))
```
EmilyBourne, line 598: Perhaps we could also handle `LiteralTrue()` and `LiteralFalse()` explicitly, it seems strange to write:
```
merge("True ", "False", .True.)
```

EmilyBourne (changes requested):
I am not convinced that the existing tests are sufficient to test the code that you have written. Please add more

pyccel/codegen/printing/fcode.py
jalalium: does not seem to work. I tested it [here](https://regex101.com/r/wV5bD0/1) with no luck. I think that the or cannot work here because it looks ahead and may undo a " with a '.

pyccel/codegen/printing/fcode.py
jalalium: Noted, do you think i should make it independent?

pyccel/codegen/printing/fcode.py
jalalium: Yes, the `split_long_argument` only handles the arguments of the write call. The lines that are handled by this include the whole expression, But it is not enough on its own because we cannot guarantee that we will always have a space character to perform the split, we could have something like this:
`write(*, (A A A), advance="no") 'hi', 'looong string', 'bye`
the only suitable cuts here will be inside the long string literal expression, thus inside quotes so such a cut will cause problems. That's when the  `split_long_argument`  comes into play and provide the sought space character after splitting the 'looong string'

pyccel/codegen/printing/fcode.py
jalalium: sure!

pyccel/codegen/printing/fcode.py
jalalium: I don't think it will produce an empty list. I will iterate through all possible subsets of the positions to determine the ideal break. The complexity required will be n * 2 ^ n where n is the size of the positions. So, n cannot be too large. I chose 12 in this case, for a total of about 49000 iterations. this will run in about 4 microseconds even with python. To keep only 12 positions, i thought i would be better off skipping the first quarter of the string since those space characters will be too far off, then i just take the 12 positions from there to be my candidates.

pyccel/codegen/printing/fcode.py
jalalium: i apologize, i will thoroughly comment it.

pyccel/codegen/printing/fcode.py
jalalium: Sorry, you are right! i should have `int(len(positions) / 4) + 12` instead of `12`.

pyccel/codegen/printing/fcode.py
jalalium: Noted!

pyccel/codegen/printing/fcode.py
EmilyBourne: I'm actually not sure how this regex got so complicated. `("[^"]*")|('[^']*')` seems to do the trick

EmilyBourne: I'm not sure how this got so complicated. `("[^"]*")|('[^']*')` seems to do the trick. I would expect it to catch the gap between strings too, but in the regex tester this doesn't seem to be the case. If it does catch them then it is sufficient to check odd numbered groups

pyccel/codegen/printing/fcode.py
EmilyBourne: Yes I think so. I tried to review the function `_print_PythonPrint` and I kept having to skip over ~50 lines (which got worse with my comments on top), this makes it much harder to follow what is happening in the function

pyccel/codegen/printing/fcode.py
jalalium: If we want the chosen cut to be a space, truncating at the last acceptable space may not work. we could have something like this:
`print("Loremipsumdolorsitamet,consecteturadipiscinelit,seddoeiusmodtemporincididuntutlaboreet dolore ")`
cutting at the last space won't fix the problem. We could apply a blind approach only if we don't value a cut at a space character enough. Also, testing all possible combinations is not costly at all since we only test 2^12 combinations.

pyccel/codegen/printing/fcode.py
jalalium: Brilliant! thank you! i did not know this.

pyccel/codegen/printing/fcode.py
jalalium: because we don't want too small of a chunk. Minimizing the longest chunk will favor more cuts, thus the optimal combination will have as many cuts as possible, practically a cut at every white space. But we don't want that, that's why i limited the distance between two cuts.

pyccel/codegen/printing/fcode.py
jalalium: I don't think so. I will create more tests.

pyccel/codegen/printing/fcode.py
jalalium: Great, Thank you! i will refactor that part.

pyccel/codegen/printing/fcode.py
jalalium: Noted!

pyccel/codegen/printing/fcode.py
jalalium: I am not sure i follow, `split_long_argument` only acts on string literals involved in print statements. variable names should not be too long to encompass all the possible splitting points in a line that needs splitting. Not only that but even if that were the case and we had to split the variable name  ie `.... , variable, ....` got split to `...., vari& \n able, ....` this wont be a fatal problem since the & is not inside quotes.  `split_long_argument` sole role is to break string literals since `wrap_fortran` can't.

pyccel/codegen/printing/fcode.py
jalalium: Noted! i will fix it!

pyccel/codegen/printing/fcode.py
EmilyBourne: I am suggesting something like:
```
possible_breakpts = [match.start() for match in re.finditer('[^a-zA-Z0-9]', line)]
acceptable_breakpts = [pt for pr in possible_breakpts  if pt < MAX_SIZE]
if len(acceptable_breakpts)==0:
   # break in word at MAX_SIZE
else:
   # break at acceptable_breakpts[-1]
```

pyccel/codegen/printing/fcode.py
jalalium: oh thank you for mentioning that! So flushing should be performed only when the end is not specified as kwarg or whenever end contains a '\n'? I did not need to use  advance=\"yes\" because i format the ending myself and add it as an extra argument of write.

pyccel/codegen/printing/fcode.py
jalalium: I copied this part from the ccode printer. Shall i fix both of them?

pyccel/codegen/printing/fcode.py
jalalium: Oh i see, i over complicated things. So, I just get rid of all the combinations testing?

pyccel/codegen/printing/fcode.py
EmilyBourne: Yes please, don't forget to update the PR description after :+1:

pyccel/codegen/printing/fcode.py
EmilyBourne: Hmm, that's true. Actually it seems that python's `print` has a `flush` argument which defaults to `False` so maybe it is better like this

pyccel/codegen/printing/fcode.py
EmilyBourne: Ah ok. Yes splitting at every space would be excessive

pyccel/codegen/printing/fcode.py
EmilyBourne: I am not talking about names but actual variables, for example:
```
a = 'very long string'
```

EmilyBourne: > `split_long_argument` sole role is to break string literals since `wrap_fortran` can't.

This is what is causing my question. The line above is the start of the block which handles string literals and is in `wrap_fortran`. So when is this used?

pyccel/codegen/printing/fcode.py
EmilyBourne: Yes I think so. Sorry ^^

pyccel/codegen/printing/fcode.py
jalalium: You are right! something is not right here. Could you please suggest some python source code that will enter here?
Didn't your last PR render this part obsolete?
`a = (5, 'one',3)
print(a)
`
this code indicates that f is an instance of `FunctionCallArgument(InhomogeneousTupleVariable('a',dtype=NativeGeneric()))`.

jalalium: Sorry, i don't follow. isn't there any difference between str and PythonTuple?

EmilyBourne: There is no difference in your code. I think there should be a difference in which case this comment is not valid, but if they are identical then there shouldn't be multiple if blocks

pyccel/codegen/printing/fcode.py
jalalium: i don't think that check is necessary. I m not sure what it was checking for in the previous code.
Also, this example raised an issue i did not consider.
the output of this file will give:
`program prog_tupp



  implicit none

  character(len=280) :: a

  a = 'hello'
  write(*, '(A A A A A A)', advance="no") a , ' ' , ' is the same as ' , ' ' &
      , a , ACHAR(10)

end program prog_tupp
`
and this will print:
`hello                                                                                                                                                                                                                                                                                     is the same as  hello                                                                                                                                                                                                                                                                                   
`
with lots of extra white spaces due to the fact that the length of a is 280. Shouldn't we assign the exact length of a to get a nicer print?

pyccel/codegen/printing/fcode.py
EmilyBourne: Hmm, are you sure? I thought that the line `f = f.value` unpacked the `FunctionCallArgument` so that at this point `f` is an `InhomogeneousTupleVariable`. To print `a` we need to print each element individually, so something like:
```
for i in f:
    arg_format, arg = self.get_print_format_and_arg(i)
    args_format.append(arg_format)
    args.append(arg)
```

pyccel/codegen/printing/fcode.py
EmilyBourne: Strings are generally a mess at the moment (see epic #459 ). Fixing them is waaaay beyond the scope of this PR. You can try to improve the situation by using `trim` in the fortran code, but we are aware that this is a big mess. This is part of the reason why the tests never compare whitespace

pyccel/codegen/printing/fcode.py
jalalium: This is necessary to keep the declarations of the variables needed to loop over the matrix in the prelude. If they are added to the `self._additional_code`, they will raise the following error:
```
Error: Unexpected data declaration statement at (1)
.../__pyccel__/test.f90:16:26:

   16 |     integer(i64) :: i_0001
```

jalalium: This attribute is also used in ccode. I copied it from there.

pyccel/codegen/printing/fcode.py
jalalium: This is the tricky part i m not sure about.
I looked into the format specifiers for `write` in Fortran, and was only able to match these:
```
 def get_print_format_and_arg(self,var):
        type_to_format = {('real'): 'F0.12',
                          ('complex'): '"(",F0.12," + ",F0.12,")"',
                          ('integer'): 'I0',
                          ('logical'): 'A',
                          ('string'): 'A'}
```
if the dtype is not present in my dictionary i fall back to the default print with `print *, ...` . I am pretty sure i did not cover all cases. I did not find good enough of a documentation on the matter. 
```suggestion
            if self._print(var.dtype).find("character") != -1:
                return 'A', self._print(var)
```
this is to check if the type is a character then i could use  `A` with it as well.

pyccel/codegen/printing/fcode.py
EmilyBourne: Note that I did not say to place the declaration in `self._additional_code` but simply to search for the variable in the existing code. If you do this you will see that there are other places in the code where new variables are declared but that their declaration does not require the use of `self._additional_code` or `self._additional_declare`. I am not sure why declaration of variables is handled differently in the 2 printers. I suspect it is related to scoping

pyccel/codegen/printing/fcode.py
EmilyBourne: I agree that falling back to `print *, ...` is probably the safest

EmilyBourne: > this is to check if the type is a character then i could use A with it as well.

`character` is a fortran type. `var.dtype` returns the python type collected by pyccel. It will always be one of the types defined in [this file](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/datatypes.py) or the name of a class. I don't see anything named `character` here so I don't see the use case for your code

EmilyBourne:
Don't forget to merge in the master and to mark the PR as ready for review when you think you have addressed all the points raised above

jalalium:
I don't know why the Windows tests fail with the following message:
```
p_output = 'mcha bba lsbata\nvoilaaa\n3...2...1...Go!\nline1 \n line2\nline3\n\n \n\n\n\nAZUL\n\n\nat\nmid\n\n\'\'\'\'\'\n"\'\n\'...l fellawn\n\x08\x0b\x08\x0b\x08\n\x07\x08\n\x08\x0c\nline1\nline2\nline3\nline1\nline2\nline3\n\'\'\'\'"\'"\'///\\\\\n'
f_output = 'mcha bba lsbata\nvoilaaa\n3...2...1...Go!\nline1 \n line2\nline3\n\n \n\n\n\nAZUL\n\n\nat\nmid\n\n\'\'\'\'\'\n"\'\n\'...l fellawn\n\x08\x0b\x08\x0b\x08\n\x07\x08\n\x08\x0c\nline1\nline2\nline3\nline1\nline2\nline3\n\'\'\'\'"\'"\'///\\\\\n'
dtype = <class 'str'>, language = None

    def compare_pyth_fort_output_by_type( p_output, f_output, dtype=float, language=None):
    
        if dtype is str:
            p_list = [e.strip() for e in re.split('\n', p_output)]
            f_list = [e.strip() for e in re.split('\n', f_output)]
>           assert(p_list==f_list)
E           assert ['mcha bba ls... 'line3', ...] == ['mcha bba ls... 'line3', ...]
E             At index 28 diff: '' != "'"
E             Left contains one more item: ''
E             Use -v to get the full diff

pyccel\test_pyccel.py:139: AssertionError
```
the `p_output` and the 'f_output` are identical!

pyccel/codegen/printing/fcode.py
EmilyBourne: It is used to print tuples

tests/pyccel/test_pyccel.py
```diff
@@ -137,6 +137,8 @@ def compare_pyth_fort_output_by_type( p_output, f_output, dtype=float, language=
     if dtype is str:
         p_list = [e.strip() for e in re.split('\n', p_output)]
         f_list = [e.strip() for e in re.split('\n', f_output)]
+        print(p_list)
+        print(f_list)
```
rakati, line 141: remove debugging prints

EmilyBourne: Good catch! Please can you have a look at the comments that you left in your last review and decide if those conversations can be closed too.

EmilyBourne: Fixed 2b2dfb17

tests/pyccel/scripts/print_strings.py
```diff
@@ -19,7 +19,7 @@
     print('\a')
     print('\b')
     print('\f')
-    print('\r')
+    #print('\r')
```
rakati, line 22: I think `'\r'` was working before.
can you uncomment this line to be checked again?

tests/pyccel/test_pyccel.py
rakati: > Good catch! Please can you have a look at the comments that you left in your last review and decide if those conversations can be closed too.

Ok, sure I will

tests/pyccel/scripts/print_strings.py
EmilyBourne: Tests are passing d5b2f40a

EmilyBourne (approved):

pyccel/codegen/printing/fcode.py
```diff
@@ -547,31 +548,157 @@ def _print_Import(self, expr):
...
+        arg        : str
+                     The fortran code which represents var
+        """
+        type_to_format = {('float'): 'F0.12',
+                          ('complex'): '"(",F0.12," + ",F0.12,")"',
+                          ('int'): 'I0',
+                          ('bool'): 'A',
+                          ('string'): 'A',
+                          ('tuple'):  '*'
+                          }
```
yguclu, line 681: Why not saving this dictionary in a module variable, instead of creating it at each function call?

pyccel/codegen/printing/fcode.py
```diff
@@ -547,31 +548,157 @@ def _print_Import(self, expr):
...
+        var_type = str(var.dtype)
```
yguclu, line 682: I would save the variable `var_type = var.dtype`, which can be reused several time in this function.

pyccel/codegen/printing/fcode.py
```diff
@@ -547,31 +548,157 @@ def _print_Import(self, expr):
...
+        try:
+            arg_format = type_to_format[var_type]
```
yguclu, line 684: If you follow my previous suggestion, this line becomes
```python
arg_format = type_to_format[str(var_type)]
```

yguclu (changes requested):
This is very useful! Can you provide a few examples of print, with the old and new output?

pyccel/codegen/printing/fcode.py
EmilyBourne: Fixed e99c157daa06c5d43461be96cade1a057085611c

pyccel/codegen/printing/fcode.py
EmilyBourne: Fixed 415c87cf

pyccel/codegen/printing/fcode.py
```diff
@@ -547,31 +548,157 @@ def _print_Import(self, expr):
...
+                args_list    = [af for a in args_list for af in (a, separator)][:-1]
+
+            if fend_code not in ('ACHAR(10)', ''):
+                fargs_format.append('A')
+                args_list.append(fend_code)
+
+            args_code       = ' , '.join(args_list)
+            args_formatting = ' '.join(fargs_format)
+            return "write(*, '({})', advance=\"{}\") {}\n"\
+                .format(args_formatting, advance, args_code)
```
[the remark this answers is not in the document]
EmilyBourne: Fixed a9b0ce64

pyccel/codegen/printing/fcode.py
EmilyBourne: Fixed https://github.com/pyccel/pyccel/commit/e99c157daa06c5d43461be96cade1a057085611c

yguclu (approved):

PR #957, opened 2021-09-06
Wrapper part 2 (Treat C and Fortran arrays similarly in wrapper)

### LINKED ISSUE #801: as_static_function is called unnecessarily
**Describe the bug**
`as_static_function` is needed in order to modify the arguments of a fortran function so it is compatible with c. It is therefore called when printing `bind_c_X.f90`. This information is also needed in `X_wrapper.c` to write the function signature and call the function. It should be possible to reuse the functions created for the `bind_c_X.f90` file (especially now the file contains a module). However currently `CCodeWrapperPrinter` calls `as_static_function_call` (which creates a new function with a different body and then calls `as_static_function`) directly. This is unnecessary and could be costly for functions with large numbers of arguments

EmilyBourne:
Collect arrays into `ndarray`s in the cwrapper no matter the target language
Fixes #801

**Commit Summary**
- Update `numpy_wrapper` to get data and dimensions from `ndarray` instead of numpy array
- Move `internal_libs` to `codegen.utilities`
- Allow `t_ndarray` to be stored in a c pointer (necessary for optional arrays, which is available in fortran)
- Improve `stored_in_c_pointer` to handle more types
- Improve `_print_FunctionCall` to allow functions returning pointers to not need saving in temporary variables
- Use `CWrapperPrinter` to determine if `cwrapper_ndarrays` is needed
- Split `_get_static_function` into multiple functions:
    - `get_static_function`
    - `static_function_signature`
    - `get_static_args`
    - `get_static_declare_type`
- Collect `PyArrayObject` into a `ndarray` no matter the target language
- Provide functions to reduce code duplication
- Move array functions from `cwrapper.c` to `cwrapper_ndarrays.c` and add other useful functions

cc: @MenoIy


pyccel/ast/numpy_wrapper.py
```diff
@@ -66,15 +68,15 @@ def get_numpy_max_acceptable_version_file():
...
+pyarray_to_ndarray = FunctionDef(
+                name      = 'pyarray_to_ndarray',
                 arguments = [Variable(name = 'a', dtype = PyccelPyArrayObject(), is_pointer = True)],
                 body      = [],
                 results   = [Variable(name = 'array', dtype = NativeGeneric())])
 
 # numpy array check elements : function definition in pyccel/stdlib/cwrapper.c
 pyarray_checker = FunctionDef(
-                name      = 'pyarray_checker',
+                name      = 'pyarray_check',
```
rakati, line 79: I think you forget to change the name of the function too.
you change here only the name attribute.

EmilyBourne: I changed the name attribute to match the name in the c code. I'm not sure why this change was made so I didn't see the need to propagate it through the rest of the code. I can though if you think it's clearer

EmilyBourne: Fixed 330e8b6b

pyccel/ast/numpy_wrapper.py
```diff
@@ -84,25 +86,19 @@ def get_numpy_max_acceptable_version_file():
                 body      = [],
                 results   = [Variable(name = 'b', dtype = NativeBool())])
 
-# https://numpy.org/devdocs/reference/c-api/array.html#c.PyArray_DIMS
-numpy_get_ndims = FunctionDef(name      = 'PyArray_NDIM',
+# Return the shape in the n dimension : function definition in pyccel/stdlib/cwrapper.c
```
rakati, line 89: the function definition located at `pyccel/stdlin/cwrapper/cwrapper_ndarrays.c`

rakati, line 89: I think `Return the shape of the n'th dimension: ...` is more meaningful.

pyccel/stdlib/cwrapper/cwrapper_ndarrays.c
```diff
@@ -72,3 +199,77 @@ t_ndarray	pyarray_to_c_ndarray(PyArrayObject *a)
...
+ *		return 0 if object is NULL or shape at indexed dimension
+ * reference of the used c/numpy api function
+ * -------------------------------------------
+ * https://numpy.org/doc/1.17/reference/c-api.array.html#c.PyArray_DIM
+ */
+int     nd_ndim(t_ndarray *a, int n)
+{
+	if (a == NULL)
+		return 0;
+
```
rakati, line 250: I think we have to raise `index out of range` error in `n >= array.nd`

pyccel/ast/numpy_wrapper.py
```diff
@@ -84,25 +86,19 @@ def get_numpy_max_acceptable_version_file():
...
-
-# https://numpy.org/devdocs/reference/c-api/array.html#c.PyArray_DIM
-numpy_get_dim   = FunctionDef(name      = 'PyArray_DIM',
-                           body      = [],
-                           arguments = [Variable(dtype=PyccelPyArrayObject(), name = 'o', is_pointer=True),
+                           arguments = [Variable(dtype=NativeVoid(), name = 'o', is_pointer=True),
                                         Variable(dtype=NativeInteger(), name = 'idx')],
                            results   = [Variable(dtype=NativeInteger(), name = 'd')])
 
+# Return the data of ndarray : function definition in pyccel/stdlib/cwrapper.c
```
rakati, line 96: same here, the function definition is located at `pyccel/stdlib/cwrapper/cwrapper_ndarrays.c`.

pyccel/codegen/printing/ccode.py
```diff
@@ -791,8 +791,9 @@ def get_declare_type(self, expr):
                 if expr.rank > 15:
                     errors.report(UNSUPPORTED_ARRAY_RANK, severity='fatal')
                 self._additional_imports.add('ndarrays')
-                return 't_ndarray '
-            errors.report(PYCCEL_RESTRICTION_TODO, symbol="rank > 0",severity='fatal')
+                dtype = 't_ndarray'
```
rakati, line 794: just a question, do we need here treating the case of declared array pointer `t_ndarray*`?

pyccel/codegen/printing/ccode.py
```diff
@@ -1295,7 +1307,7 @@ def _print_Constant(self, expr):
 
     def _print_Return(self, expr):
         code = ''
-        args = [VariableAddress(a) if self.stored_in_c_pointer(a) else a for a in expr.expr]
+        args = [VariableAddress(a) if isinstance(a, Variable) and self.stored_in_c_pointer(a) else a for a in expr.expr]
```
rakati, line 1310: are sure that we need here to check that the `a` is instance of `variable`, we have this check in the function `stored_in_c_pointer`?

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -519,6 +646,14 @@ def get_PyBuildValue(self, used_names, variable):
     #                 _print_ClassName functions
     #--------------------------------------------------------------------
 
+    def _print_VariableAddress(self, expr):
```
rakati, line 649: is it necessary to overite this method? why can't we use `CCodePrinter`'s method for that?

pyccel/ast/numpy_wrapper.py
EmilyBourne: Good catch, nothing in this file should be located in `pyccel/stdlib/cwrapper.c`

pyccel/codegen/printing/ccode.py
EmilyBourne: Yes, this is important for optional arrays

pyccel/codegen/printing/ccode.py
EmilyBourne: We no longer have this check in `stored_in_c_pointer` as `Variable`s are not the only objects that can be stored in pointers. See e27b65dc3d48c52186d12125609cf7017a01e6eb

pyccel/stdlib/cwrapper/cwrapper_ndarrays.c
EmilyBourne: I don't think it is possible to have `n >= array.nd`. This function is only called by pyccel to pass an ndarray to a fortran function by doing (for example):
```
bind_c_array_int32_1d_scalar_add(nd_ndim(&x, 0), nd_data(&x), a);
```
When the array `x` is collected its rank is checked. We will only ask for indices where `n<x.rank` and as this is checked there is no danger of out of range problems

pyccel/ast/numpy_wrapper.py
EmilyBourne: Fixed a29cb8f

EmilyBourne: Fixed de6219e

pyccel/ast/numpy_wrapper.py
EmilyBourne: Fixed de6219e

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: I agree. Fixed: 18470e4

EmilyBourne: Hmm, it seems this is more complicated than expected as `stored_in_c_pointer` is not used as described everywhere. Especially when ndarrays are involved. I am trying to unravel the mess

pyccel/stdlib/cwrapper/cwrapper_ndarrays.c
rakati: hmm, I tried to reproduce the error for that case but I get  a different error:
```python
import numpy as np
a = np.array([[1, 2], [3, 4]])
n = 2
print(a.shape[n])
```
I get 
```bash
TypeError: tuple indices must be integers or slices, not variable
```

pyccel/codegen/printing/ccode.py
rakati: Ah now I see, thanks

pyccel/stdlib/cwrapper/cwrapper_ndarrays.c
EmilyBourne: This will not reproduce the error above as the function above is only used in the wrapper (hence it appearing in `cwrapper_ndarrays.c`). I don't think it is possible to reproduce the error which is why I don't think it needs handling. It would appear if you passed a 1d array to a function translated to fortran which expects a 2d array, if you comment the section that tests if the array that was passed is indeed a 2d array

pyccel/codegen/printing/cwrappercode.py
rakati: I notice that the `CCodePrinter`'s method doesn't add an `&` if `expr.variable.rank > 0`, while the one you wrote in `cwrappercode` does.

EmilyBourne: Yes exactly. This is really just a sloppy workaround though, it would be better to stick to the definition where `stored_in_c_pointer` does what it says and `VariableAddress` returns the C address of a variable. Currently that is the case for 0d objects but nd objects are all treated differently. This is fine as long as an `ndarray` cannot be a pointer, but optional ndarray arguments are pointers so this now needs handling properly

rakati (commented):
I think printing array pointers still need some fixes, for example this a sample ex:
````python
def func(x:'int[:]', y:'int[:]'=None):
    print(x)
    if y is not None:
        print(y)
````
still can print that.
you may to fix  _print_PyccelArraySize to:
```python
    if self.stored_in_c_pointer(expr.arg):
        return '{}->shape[{}]'.format(expr.arg, expr.index)
    return '{}.shape[{}]'.format(expr.arg, expr.index)
```
to catch the most of these kinds of errors, you will need to add more test cases for optinal variables.

pyccel/codegen/printing/cwrappercode.py
rakati: exactly, We need to be careful about that.

EmilyBourne:
> I think printing array pointers still need some fixes, for example this a sample ex:
> 
> ```python
> def func(x:'int[:]', y:'int[:]'=None):
>     print(x)
>     if y is not None:
>         print(y)
> ```
> 
> still can print that.
> you may to fix _print_PyccelArraySize to:
> 
> ```python
> def _print_PyccelArraySize(self, expr):
>     if self.stored_in_c_pointer(expr.arg):
>         return '{}->shape[{}]'.format(expr.arg, expr.index)
>     return '{}.shape[{}]'.format(expr.arg, expr.index)
> ```
> 
> to catch the most of these kinds of errors, you will need to add more test cases for optinal variables.

Ah you are right. This is the problem with not using Dotted objects :( I will add more tests

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: `VariableAddress` now does what it is supposed to

EmilyBourne:
I think activating optional arrays in C needs more work. It would be better to add some extra objects to `NumpyArrayClass`, but this is beyond the scope of this PR. As a result I have added a clear error for optional array arguments in C and am leaving the rest for another PR

pyccel/stdlib/cwrapper/cwrapper_ndarrays.c
rakati: yap, until now we are dealing only with user mistakes, what do you think in case of someone of us modified some part of the wrapper and get a `segfault` error, how much time does he need to figure out that problem?

EmilyBourne: Hmm, I don't know. Given that the function is only used in 1 line, if modifications are made to the code incrementally it should be reasonably easy to see where in the generated code you need to look and it should be fairly obvious that you're doing something stupid. This line is unlikely to be touched often. As the size has already been tested I am not convinced that the trade-off between catching very stupid developer behaviour and having 2 tests of the size is worth it. I'm also struggling to see how you can make this mistake.
The function is used [here](https://github.com/pyccel/pyccel/blob/wrapper_part_2/pyccel/codegen/printing/cwrappercode.py#L279)
so you would have to replace `rank` in:
```
[FunctionCall(array_get_dim, [arg_address, i]) for i in range(argument.rank)]
```
with a `LiteralInteger` or a sum or something very strange like that

rakati: yes, you are right.

pyccel/codegen/printing/ccode.py
```diff
@@ -1441,22 +1455,33 @@ def _print_AliasAssign(self, expr):
         lhs_var = expr.lhs
         rhs_var = expr.rhs
 
-        lhs = VariableAddress(lhs_var)
-        rhs = VariableAddress(rhs_var) if isinstance(rhs_var, Variable) else rhs_var
-
-        lhs = self._print(lhs)
-        rhs = self._print(rhs)
+        lhs_address = VariableAddress(lhs_var)
+        # Ensure all everything which can be stored in a VariableAddress is
```
rakati, line 1459: is this an incomplete comment or I didn't get the chance to understand it?

rakati (approved):

rakati:
> I think activating optional arrays in C needs more work. It would be better to add some extra objects to `NumpyArrayClass`, but this is beyond the scope of this PR. As a result I have added a clear error for optional array arguments in C and am leaving the rest for another PR

yes, I agree with you.

pyccel/codegen/printing/ccode.py
EmilyBourne: There is a word to many. It should be "Ensure everything which can be stored in a VariableAddress is"

EmilyBourne: To be understood as "Ensure everything which can be stored in a VariableAddress is stored in a VariableAddress"

EmilyBourne: Fixed 4f09939

saidctb (approved):

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -280,6 +372,22 @@ def _get_wrapper_name(self, used_names, func):
 
         return wrapper_name
 
+    def get_wrapper_arguments(self, used_names):
+        """
+        Create wrapper arguments
+        Parameters
+        ----------
```
yguclu, line 379: Please add an empty line before the `Parameters` section

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -190,35 +189,128 @@ def get_default_assign(self, arg, func_arg, value):
...
+        Get the declaration type of a variable, this function is used for
+        C/fortran binding using native C datatypes.
+
+        Parameters
+        ----------
+        variable : Variable
+            Variable holding information needed to choose the declaration type
+
+        Returns: String
+        --------
```
yguclu, line 298: Wrong formatting: `String` should be listed below the `---------` separator.

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -280,6 +372,22 @@ def _get_wrapper_name(self, used_names, func):
...
 
+    def get_wrapper_arguments(self, used_names):
+        """
+        Create wrapper arguments
+        Parameters
+        ----------
+        used_names : Set of strings
+            Set of variable and function names to avoid name collisions
+
+        Returns: List of variables
```
yguclu, line 383: Wrong formatting of `Returns` section

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -449,6 +556,44 @@ def _body_management(self, used_names, variable, collect_var, default_value = No
...
+        Parameters
+        ----------
+        wrapper_name    : string
+            The name of the C wrapper function
+
+        wrapper_args    : list of Variables
+            List of python object variables
+
+        wrapper_results : Variable
+            python object variable
```
yguclu, line 574: Shouldn't we say that `wrapper_args` contains the function arguments and `wrapper_results` the function results?

By the way, given line 591 below I think that `wrapper_results` is a list.

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -528,14 +673,11 @@ def _print_Interface(self, expr):
 
         # Find a name for the wrapper function
         wrapper_name = self._get_wrapper_name(used_names, expr)
+        used_names.add(wrapper_name)
 
         # Collect local variables
```
yguclu, line 678: Do we still collect local variables somewhere?

yguclu (changes requested):
Very good job @EmilyBourne !

I only have a few minor comments about the docstrings

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: Fixed : c21c497

pyccel/ast/cwrapper.py
```diff
@@ -2,10 +2,9 @@
 # This file is part of Pyccel which is released under MIT License. See the LICENSE file or #
 # go to https://github.com/pyccel/pyccel/blob/master/LICENSE for full license details.     #
 #------------------------------------------------------------------------------------------#
-# pylint: disable=missing-function-docstring
 
 """
-Handling the transitions between python code and C code.
+Handling the transitions between python code and C code using (Python/C Api).
```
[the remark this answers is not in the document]
EmilyBourne: Better ? 8faf4af

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: Fixed : 23851c9

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: Fixed : 23851c9

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: Fixed : 23851c9

pyccel/codegen/printing/cwrappercode.py
yguclu: Sorry, my question was too brief. I meant that while the variables `python_func_args`, `python_func_kwargs` and `python_func_selfarg` have been removed in this PR, the comment line `# Collect local variables` has remained. Was this intentional?

EmilyBourne: The lines have not been removed but have just been shifted to the function `get_wrapper_arguments` so this section still does the same thing. That said, I'm not sure this comment ever made sense. I have removed it

EmilyBourne: See c21c497

yguclu: Gotcha. Thanks

yguclu (approved):

PR #958, opened 2021-09-10
Add support for type function

### LINKED ISSUE #945: Type() function support
**Describe the bug**
In the use case of Templates, it would be helpful to know the type of a passed argument to a function.

**Test Code**
```python
from numpy import zeros
#$ header template T(float|int)
#$ header function test_fun(T)
def test_fun(a):
    x = zeros(10,dtype= type(a))
```

**Language**
C/Fortran

EmilyBourne:
Add support for type function. Fixes #945 . Also includes bug fixes

**Commit Summary**
- Add PythonType class to handle type function
- Allow type object to be passed as array dtype
- Allow type to be printed
- Add tests
- Add error test
- [BUG FIX] Ensure interface points at function with correct precision
- [BUG FIX] Ensure literals are printed with the correct precision in python
- [BUG FIX] Add a lock for the acquisition of locks to avoid deadlocks if 2 files have the same dependencies saved in different orders
- [BUG FIX] Avoid unnecessarily locking files while testing if compilation is necessary
- [BUG FIX] Correct location where python translations are stored for pyccel tests


MostafaMamouni:
i'm concerned about the output of a C/fortran code that is from the result of `type` function (which can be something like `<class typeN>`). in python, it's fine to see such an output because python is an OOP language so the keyword `class` has a meaning, but C/Fortran it just can be weird to see `<class 'int64'>`  or `<class 'float'>`. i suggest raising a warning whenever a user tries `print(type(obj/literal))`, since it will give something like `<class typeN>` which is helpful and has the meaning in python, the same output will be for C/Fortran but I don't think it has meaning concerning the languages and it can be helpful just for the devs.
python code:
`print(type(1))`
the output
<class 'int'>
C equivalent:
```
#include <stdlib.h>
#include <stdio.h>
int main()
{
    printf("%s\n", "<class \'int64\'>");
    return 0;
}
```
Fortran equivalent:
```
program prog_boo



  implicit none

  print *, '<class ' // ACHAR(39) // 'int64' // ACHAR(39) // '>'

end program prog_boo
```

EmilyBourne:
@MostafaMamouni 
> i'm concerned about the output of a C/fortran code that is from the result of `type` function (which can be something like `<class typeN>`). in python, it's fine to see such an output because python is an OOP language so the keyword `class` has a meaning, but C/Fortran it just can be weird to see `<class 'int64'>` or `<class 'float'>`. i suggest raising a warning whenever a user tries `print(type(obj/literal))`, since it will give something like `<class typeN>` which is helpful and has the meaning in python, the same output will be for C/Fortran but I don't think it has meaning concerning the languages and it can be helpful just for the devs.

To be clear the options are:
1. Reproduce the python output (or as close as possible)
2. Reproduce the python output (or as close as possible) but warn the user that this is not the actual type of the object
3. Print `int64_t` in C instead of `<class 'int64'>`

I am interested to know which choice other people think is best

MostafaMamouni:
> @MostafaMamouni
> 
> > i'm concerned about the output of a C/fortran code that is from the result of `type` function (which can be something like `<class typeN>`). in python, it's fine to see such an output because python is an OOP language so the keyword `class` has a meaning, but C/Fortran it just can be weird to see `<class 'int64'>` or `<class 'float'>`. i suggest raising a warning whenever a user tries `print(type(obj/literal))`, since it will give something like `<class typeN>` which is helpful and has the meaning in python, the same output will be for C/Fortran but I don't think it has meaning concerning the languages and it can be helpful just for the devs.
> 
> To be clear the options are:
> 
> 1. Reproduce the python output (or as close as possible)
> 2. Reproduce the python output (or as close as possible) but warn the user that this is not the actual type of the object
> 3. Print `int64_t` in C instead of `<class 'int64'>`
> 
> I am interested to know which choice other people think is best

For me option 3 is the best. because it's just `class` that should be removed.

EmilyBourne:
@MostafaMamouni
> For me option 3 is the best. because it's just `class` that should be removed.

Beware I wrote `int64_t` and not `int64`

pyccel/ast/builtins.py
```diff
@@ -774,6 +776,49 @@ def __str__(self):
...
+    __slots__ = ('_dtype','_precision','_obj')
+    _attribute_nodes = ('_obj',)
+
+    def __init__(self, obj):
+        self._dtype = obj.dtype
+        self._precision = obj.precision
+        self._obj = obj
+
+        if obj.rank > 0:
+            raise PyccelError("Python's type function doesn't return enough information about this object for pyccel to fully define a type")
```
rakati, line 792: I think we can do the same as python and return `numpy.ndarray` here.

EmilyBourne: If the aim is to print the type then this comment is valid, but the main use case for this is for doing something like:
```
x = zeros(10,dtype= type(a))
```
as suggested in the PR. This we cannot handle

pyccel/ast/builtins.py
```diff
@@ -774,6 +776,49 @@ def __str__(self):
...
+        return self._obj
+
+    @property
+    def print_string(self):
+        """ Return a literal string representing the type that
+        can be used in a print  statement
+        """
+        return LiteralString("<class '{dtype}{precision}'>".format(
+            dtype = str(self.dtype),
+            precision = self.precision*8 if self.precision else ''))
```
rakati, line 820: I don't know why we should print all this information instead of printing just the type same as python
```python
In [1]: type(1)
Out[1]: int

In [2]: a = 3.4

In [3]: type(a)
Out[3]: float

In [4]: import numpy as np

In [5]: type(np.array([2,3]))
Out[5]: numpy.ndarray

In [6]: def func():
   ...:     pass
   ...:

In [7]: type(func)
Out[7]: function

In [8]: type('Azul')
Out[8]: str

In [9]: t = type('c')

In [10]: type(t)
Out[10]: type
```

EmilyBourne: You are confusing the result of `type(a)` and the result of `print(type(a))` which is equivalent to `str(type(a))`. See:
```python
In [1]: print(type(1))                                                                                                                                                                                     
<class 'int'>

In [2]: a = 3.4                                                                                                                                                                                            

In [3]: print(type(a))                                                                                                                                                                                     
<class 'float'>

In [4]: import numpy as np                                                                                                                                                                                 

In [5]: print(type(np.array([2,3])))                                                                                                                                                                       
<class 'numpy.ndarray'>

In [6]: def func(): 
   ...:     pass 
   ...:                                                                                                                                                                                                    

In [7]: print(type(func))                                                                                                                                                                                  
<class 'function'>

In [8]: print(type('Azul'))                                                                                                                                                                                
<class 'str'>

In [9]: t = type('c')                                                                                                                                                                                      

In [10]: print(type(t))                                                                                                                                                                                    
<class 'type'>
```

EmilyBourne: This is related to the discussion with @MostafaMamouni above. It would be useful if you could give your opinion on this discussion

rakati: I just read your disscusion about that, for me, I see that the first option is more convenient, because we are dealing with python users seeking performance, not all users are familiar with C or Fortran, so I think we need to keep things the same or nearby as python is.

EmilyBourne: So then you are happy with the current output?

pyccel/ast/builtins.py
rakati: yes, I didn't pay attention to that, my fault.

pyccel/ast/builtins.py
rakati: yes.

rakati (approved):

jalalium:
I don't know what should be the expected behavior if the python source code defines a `type` function, but it looks like something is off.
```
def type(n : int):
    return 5

print(type(5))
```
this program creates a t module
```
module t


  implicit none

  contains

  function type(n) result(Out_0001)

    implicit none

    integer(i64) :: Out_0001
    integer(i64), value :: n

    Out_0001 = 5_i64
    return

  end function type

end module t
```
but ends up treating the `type` call in `prog_t` as if it mean the builtin `type` function.
```
program prog_prog_t

  use t, only: type


  implicit none

  print *, '<class ' // ACHAR(39) // 'int64' // ACHAR(39) // '>'

end program prog_prog_t
```

EmilyBourne:
> I don't know what should be the expected behavior if the python source code defines a `type` function, but it looks like something is off.

In python the behaviour is to override the builtin `types` function.
In pyccel, it is generally advised to only translate clean code (ie. one where pylint would not raise any errors or warnings). Your code raises the warning:
```
tmp.py:1:0: W0622: Redefining built-in 'type' (redefined-builtin)
```
I think the correct behaviour for pyccel should be to raise an error. This is related to issue #227

EmilyBourne:
> > I don't know what should be the expected behavior if the python source code defines a `type` function, but it looks like something is off.
> 
> In python the behaviour is to override the builtin `types` function.
> In pyccel, it is generally advised to only translate clean code (ie. one where pylint would not raise any errors or warnings). Your code raises the warning:
> 
> ```
> tmp.py:1:0: W0622: Redefining built-in 'type' (redefined-builtin)
> ```
> 
> I think the correct behaviour for pyccel should be to raise an error. This is related to issue #227

In any case I think this is beyond the scope of this PR as the following code encounters the same problems:
```
def float(n : int):
    return n+1

print(float(5))
```

jalalium (approved):

saidctb:
does this work?
```python
print(type(type(1)))
```

EmilyBourne:
> does this work?
> ```python
> print(type(type(1)))
> ```

No. Currently I suspect this will raise an AttributeError. I will add a check on the argument type to ensure that it is a `PyccelAstNode`

yguclu (changes requested):
This is a useful new feature! I have a couple of comments.

pyccel/codegen/printing/pycode.py
```diff
@@ -697,7 +699,23 @@ def _print_IfTernaryOperator(self, expr):
...
+            cast_func = DtypePrecisionToCastFunction[dtype.name][precision]
+            type_name = cast_func.__name__.lower()
+            is_numpy  = type_name.startswith('numpy')
+            cast_name = cast_func.name
+            name = self._aliases.get(cast_func, cast_name)
+            if is_numpy and name == cast_name:
+                self.insert_new_import(
+                        source = 'numpy',
+                        target = cast_name)
+            return '{}({})'.format(name, repr(expr.python_value))
```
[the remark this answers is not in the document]
EmilyBourne: If you do the following in python:
```
a = np.int32(3)
```
Then this is translated to the following pyccel nodes:
```
Assign(Variable(a), LiteralInteger(3, precision=4))
```
As the precision is stored in the literal object rather than saving the cast, the cast needs to be retrieved for printing in python. We can't create an actual cast node as the conversion to a literal is done in the `NumpyInt32.__new__` function.

This is therefore a bug fix for the python printer to ensure that literals are created with the correct precision

pyccel/ast/core.py
```diff
@@ -2600,6 +2600,8 @@ def point(self, args):
                                 or dtype2 in dtype1)
                 found = found and call_arg.rank \
                                 == func_arg.rank
+                found = found and call_arg.precision \
+                                == func_arg.precision
```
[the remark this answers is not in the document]
EmilyBourne: 8383f95aa2345620d04d9bee2d4541907c19dd6f

pyccel/codegen/printing/pycode.py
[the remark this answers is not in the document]
yguclu: Thanks for the explanation!

EmilyBourne:
There seems to be a deadlock somewhere in the code which is preventing the tests from passing (timeout errors). I thought of 1 place it could be and fixed that, but it seems there is still something more. Does anyone have any ideas where it could be?

PR #959, opened 2021-09-10
Pass single precision integers to LAPACK routines

AbdelGhafour69:
Correct `lapack.pyh` file to conform with LAPACK definitions and added tests for subroutines `dgetri` and `dgetrf`. Skip LAPACK tests on Windows (see issue #1091).


pyccel/stdlib/internal/lapack.pyh
```diff
@@ -11,228 +11,228 @@
...
 
 
-#$ header function sgesv(int, int, float [:,:](order=F), int, int [:], float [:,:](order=F), int, int)
-#$ header function dgesv(int, int, double [:,:](order=F), int, int [:], double [:,:](order=F), int, int)
+#$ header function sgesv(int32, int32, float [:,:](order=F), int32, int32 [:], float [:,:](order=F), int32, int32)
+#$ header function dgesv(int32, int32, double [:,:](order=F), int32, int32 [:], double [:,:](order=F), int32, int32)
 
-#$ header function sgesvx(str, str, str, int, int, float [:,:](order=F), int, float, float, int, int, int, float [:], float [:,:](order=F), int, float [:,:](order=F), int, float [:], int, int [:], int)
-#$ header function dgesvx(str, str, str, int, int, double [:,:](order=F), int, double, double, int, int, int, double [:], double [:,:](order=F), int, double [:,:](order=F), int, double [:], int, int [:], int)
+#$ header function sgesvx(str, str, str, int32, int32, float [:,:](order=F), int32, float, float, int32, int32, int32, float [:], float [:,:](order=F), int32, float [:,:](order=F), int32, float [:], int32, int32 [:], int32)
```
EmilyBourne, line 27: Can we also fix the float problem in this PR?

AbdelGhafour69: Yes we can, I think the whole file need reviewing.

EmilyBourne:
"Help wanted" label is added as the windows executable will not run. It fails with the error:
```
liblapack.dll: cannot open shared object file: No such file or directory
```
It is not clear what command should be used to avoid this error. If anyone has a windows machine they can test on then this would be useful

EmilyBourne (approved):

EmilyBourne:
It doesn't look like anyone is likely to pick this up and help diagnose the windows problem, so I have opened issue #1091 . I therefore think we can merge this PR

yguclu (approved):

PR #968, opened 2021-09-27
Extend tutorial

MostafaMamouni:
documenting the following features in the `tutorial` directory and updating the "hello world" example in Pyccel tutorial:

- [x] Functions as arguments
- [ ] `const` keyword
- [ ] Tuples
- [ ] Lists
- [x] Available decorators (`pure`, `elemental`, `stack_array`, `allow_negative_index`)
- [x] update the hello world example


MostafaMamouni:
i suggest making a separate file for how `__pyccel__` dir is structured because I think it's not that stable to repeatedly write it in the tutorial. The same thing for features that we are certain will get changed by time.
@yguclu @EmilyBourne

EmilyBourne:
I agree that we should not repeat the explanation of the `__pyccel__` dir structure. It should be explained once. Currently there are examples in the quickstart file (I don't know if there are any elsewhere?), but I haven't seen a description of what the files are anywhere. I don't think the structure of the `__pyccel__` folder is likely to change too much in the future.

By the way, I think you pushed the same file twice (`function-as-arguments.md` and `functions-as-arguments.md` )

MostafaMamouni:
`allow_negative_index` decorator makes no difference in the generated code. is that normal ? @EmilyBourne @yguclu

MostafaMamouni:
> `allow_negative_index` decorator makes no difference in the generated code. is that normal ? @EmilyBourne @yguclu

see [here](https://github.com/pyccel/pyccel/pull/968/files#diff-5d2c4817d8c3f3f15d61cce8c5ed1b446019113cf4c3164e1f2eeca5c59093f4R134)

EmilyBourne:
> `allow_negative_index` decorator makes no difference in the generated code. is that normal ? @EmilyBourne @yguclu

The `allow_negative_index` should make negative indexes work. I think negative literal indexes work with or without the decorator, but the following code should be different depending on whether or not the decorator is present:
```
def get_elem(f : 'float[:]', i : int, j : int):
    return f[i-j]
```

tutorial/decorators.md
```diff
@@ -0,0 +1,243 @@
+# Decorators
+
+Because Pyccel is converting a dynamically typed language (Python) to statically types ones, It imposes some *restrictions* on the code. One of them is using decorators that have certain options in order to specify how the functions or the variables should be treated in the conversion, Here are the available decorators.
```
EmilyBourne, line 3: I would not say that decorators are restrictions, rather they are optional markers which allow the user to benefit from adv advantages that are not available in python

tutorial/decorators.md
```diff
@@ -0,0 +1,243 @@
...
+
+## Stack array
+
+This decorator indicates that all arrays mentioned as argements (of the decorator) should be stored
```
EmilyBourne, line 7: Typo `argements`->`arguments`

tutorial/decorators.md
```diff
@@ -0,0 +1,243 @@
...
+on the stack.
+
+This example shows how the decorators can affect the conversion of the array between the supported languages, Pyccel here is told by the decorator `stack array` to store the array `arrat_in_stack` in the stack, For the array `array_in_heap` Pyccel is assuming that it should be stored in the heap:
```
EmilyBourne, line 10: Typo `arrat_in_stack`->`array_in_stack`

tutorial/decorators.md
```diff
@@ -0,0 +1,243 @@
...
+def fun1():
+
+     #/////////////////////////
+     #array stored in the stack
+     #////////////////////////
+     array_in_stack = [1,2,3]
+     #////////////////////////
+     #array stored in the heap
+     #////////////////////////
+     array_in_heap = [1,2,3]
```
EmilyBourne, line 25: I think we should store an array here rather than a list given that we don't really support lists (`np.array([1,2,3])`)

tutorial/decorators.md
```diff
@@ -0,0 +1,243 @@
...
+
+  end subroutine fun1
+  !........................................
+
+end module boo
+```
+
+## Allow negative index
+
+This one indicates that all arrays mentioned as argements (of the decorator) can be accessed with negative indexes. unlike fortran you can't use negative indexes in arrays in C, But with Pyccel you can work comfotabally with them in the Python code, For the generated C code Pyccel will take care of that.
```
EmilyBourne, line 105: Typos: `argements`->`arguments`   `comfotabally`->`comfortably `

Fortran allows negative indexes but they do not play the same role as they do in python. In fortran you can choose what number is the first index. For example:
```
allocate(my_array(-5:-3))
my_array = [1_i64, 2_i64, 3_i64]
print *, my_array(-5)
```
will print `1`. Negative indexing is therefore only allowed if the array was defined to have negative indexes.
When translating python code to fortran all our arrays have `0` as the first index therefore we cannot use negative indexes even in Fortran.

I don't think this description is very clear. I suggest:

```
In python negative indexes allow a user to index an array starting from the back (e.g. the index -1 is the
last element of the array). Pyccel recreates this behaviour for literal indexes. However when an array is
indexed with a variable or an expression, it is impossible to know at compile time whether the index is
positive or negative. As a result an if block must be added. This implies a (potentially large) performance
cost. Non-literal negative indexes are not especially common, therefore Pyccel does not add this costly
if block unless it is specifically requested. This can be done using the `allow_negative_index` decorator
```

tutorial/decorators.md
```diff
@@ -0,0 +1,243 @@
...
+
+  end subroutine fun1
+  !........................................
+
+end module boo
+```
+
+## Elemental and pure decorators
+
+This decorator indicates that the function bellow is an elemental one, An elemental function is a function with a single scalar operator and a scalar return value which can also be called on an array. When it is called on an array it returns the result of the function called elementwise on the array.
```
EmilyBourne, line 184: "This decorator" `pure` or `elemental`?

tutorial/decorators.md
```diff
@@ -0,0 +1,243 @@
...
+This decorator `pure` indicates that the function bellow is a pure one. So that function should return identical return values for identical arguments and has no side effects in its application.
```
EmilyBourne, line 185: Typo `bellow`->`below`

tutorial/decorators.md
```diff
@@ -0,0 +1,243 @@
...
+
+@pure
+@elemental
+@types(float)
+def square(x):
+    s = x*x
+    return s
+```
+
+This is the C generated code, this code can be generated even without these two decorators due to the C language itself that has not the elemenwise feature in the manipulation of the arrays nor a function prefix `pure`. So these decorators will not make a big deffirence in the Python/C conversion:
```
EmilyBourne, line 200: Typo `deffirence`->`difference`

The `pure` decorator does nothing in c (which does not have an equivalent keyword), but `elemental` does. You won't see it in the translation of the function itself, but in the translation of a call. E.g:
```
a = np.ones(5)
square(a)
```
Without `@elemental` for the function `square` this will not compile. It should raise a semantic error. With `@elemental` it will be translated to a for loop

tutorial/decorators.md
yguclu: True.

tutorial/decorators.md
yguclu: This description is definitely for `elemental` functions. `pure` has to do with the absence of side effects, so a "pure" function is the "standard" function in purely functional languages.

tutorial/decorators.md
yguclu: Also, `An` should not be capitalized as it follows a comma sign.

tutorial/decorators.md
yguclu: Typo `elemenwise` -> `elementwise` (missing `t`)

tutorial/decorators.md
EmilyBourne: Or rather, it should be a period `.`, not a comma `,` before `An`

EmilyBourne:
I wonder if we should split this into multiple PRs. That way any sections that are written can be made available to users, without waiting for everything to be written

MostafaMamouni:
> I wonder if we should split this into multiple PRs. That way any sections that are written can be made available to users, without waiting for everything to be written

perfect for me.

EmilyBourne:
This PR no longer modifies any code, so I am closing it.

EmilyBourne:
@MostafaMamouni Can we delete the branch?

PR #972, opened 2021-09-28
Bug fixes

EmilyBourne:
Bug fixes found by testing the pyccel-examples repository

**Commit Summary**
- Remove unnecessary line break in `_print_Return`
- Handle if with allocations where 1st block doesn't allocate all variables

### DIFF
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -2904,11 +2904,14 @@ class SemanticParser(BasicParser):
         variables = [v for branch in var_shapes for v in branch]
 
         for v in variables:
-            shape_branch1 = var_shapes[0][v]
-            if not all(v in branch_shapes.keys() for branch_shapes in var_shapes) \
-                    or not all(shape_branch1==branch_shapes[v] \
-                                for branch_shapes in var_shapes[1:]):
+            if v not in var_shapes[0]:
                 v.set_changeable_shape()
+            else:
+                shape_branch1 = var_shapes[0][v]
+                if not all(v in branch_shapes.keys() for branch_shapes in var_shapes) \
+                        or not all(shape_branch1==branch_shapes[v] \
+                                    for branch_shapes in var_shapes[1:]):
+                    v.set_changeable_shape()
 
         return If(*args)
 
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -2904,14 +2904,16 @@ class SemanticParser(BasicParser):
         variables = [v for branch in var_shapes for v in branch]
 
         for v in variables:
-            if v not in var_shapes[0]:
-                v.set_changeable_shape()
-            else:
+            all_shapes_set = all(v in branch_shapes.keys() for branch_shapes in var_shapes)
+            if all_shapes_set:
                 shape_branch1 = var_shapes[0][v]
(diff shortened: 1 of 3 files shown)


MostafaMamouni (approved):

rakati (commented):
I think we need to add a small test prove for that.

pyccel/parser/semantic.py
```diff
@@ -2904,11 +2904,14 @@ def _visit_If(self, expr, **settings):
...
-                    or not all(shape_branch1==branch_shapes[v] \
-                                for branch_shapes in var_shapes[1:]):
+            if v not in var_shapes[0]:
                 v.set_changeable_shape()
+            else:
+                shape_branch1 = var_shapes[0][v]
+                if not all(v in branch_shapes.keys() for branch_shapes in var_shapes) \
+                        or not all(shape_branch1==branch_shapes[v] \
+                                    for branch_shapes in var_shapes[1:]):
+                    v.set_changeable_shape()
```
rakati, line 2914: I think you can merge this if statement into one if.

EmilyBourne: Can you suggest how in a way that avoids indexing `var_shapes` more times than necessary?

rakati:
```python
if v not in var_shapes[0] or not all(v in branch_shapes.keys() for branch_shapes in var_shapes) \
                        or not all(var_shapes[0][v]==branch_shapes[v] \
                                    for branch_shapes in var_shapes[1:]):
   v.set_changeable_shape()
```

EmilyBourne: It was written as above to avoid doing:
```
all(var_shapes[0][v]==branch_shapes[v] \
                                    for branch_shapes in var_shapes[1:]):
```
which calls `var_shapes[0][v]` several times, when it only needs to be done once

rakati: I don't know if python will optimize that, and read the value of var_shapes[0][v] once.

EmilyBourne: XD normally "python" and "optimise" are not words that go together. I am happy to go back to this though, it may flip back later in the review though. It came from this discussion:
https://github.com/pyccel/pyccel/pull/921#discussion_r668750486

EmilyBourne: In fact this:
```
if v not in var_shapes[0] or not all(v in branch_shapes.keys() for branch_shapes in var_shapes) \
                        or not all(var_shapes[0][v]==branch_shapes[v] \
                                    for branch_shapes in var_shapes[1:]):
   v.set_changeable_shape()
```
Can be written more simply as:
```
if not all(v in branch_shapes.keys() for branch_shapes in var_shapes) \
                        or not all(var_shapes[0][v]==branch_shapes[v] \
                                    for branch_shapes in var_shapes[1:]):
   v.set_changeable_shape()
```
So it seems the discussion above was the cause of this bug

EmilyBourne: better: d0f141d0 ?

EmilyBourne:
@rakati 
> I think we need to add a small test prove for that.
Is this sufficient: ba525b0

rakati (approved):

yguclu (approved):

PR #977, opened 2021-09-30
Add 'omp end do' statement

EmilyBourne:
- Rename `nowait_expr` in `For` to the more general `end_annotation`
- Remove `_print_OMP_X` functions for which the class `OMP_X` does not exist
- Add `!$omp end do` statement to the end of all OpenMP for loops in Fortran
- Ignore any `#$omp end for` statements provided by the user and raise a warning indicating that the clause is ignored
- Ensure that the fst of AnnotatedComments is set


nhamidn:
what if the user specified the `#$omp end for `in the python code for example : 
```
     #$omp parallel num_threads(n)
     #$omp for
     for i in range(0, 10):
         print(i)
     #$omp end for
     #$omp end parallel
```
this will generate fortran  code with multiple end statement : 
```
    !$omp parallel num_threads(n)
    !$omp do
    do i = 0_i64, 9_i64, 1_i64
      print *, i
    end do
    !$omp end do
    !$omp end do
    !$omp end parallel
```
and generate C code with the end statement but in C we don't open the bracket the #$omp for construct:
```
     #pragma omp parallel num_threads(n)
     {
         #pragma omp for
         for (i = 0; i < 10; i += 1)
         {
             printf("%ld\n", i);
         }
     }
 }
```

i think we must ignore any `#$omp end for` pragma since we add our own.

nhamidn (approved):

yguclu:
@EmilyBourne In the PR description you say "Ignore any !$omp end do statements provided by the user", but I am not sure that I understand what you mean.

Since we changed our syntax from `!$omp do` to `!$omp for` a while back, I think that we should raise an error when we encounter OpenMP pragmas that do not fit our grammar.

EmilyBourne:
@yguclu 
> In the PR description you say "Ignore any !$omp end do statements provided by the user", but I am not sure that I understand what you mean.
> 
> Since we changed our syntax from `!$omp do` to `!$omp for` a while back, I think that we should raise an error when we encounter OpenMP pragmas that do not fit our grammar.

Sorry this was a typo that I have now updated. It should be "Ignore any `!$omp end for` statements provided by the user". This is done for example [here](https://github.com/pyccel/pyccel/blob/master/tests/internal/scripts/openmp/ex2.py). I discussed the usefulness of this with @nhamidn and he felt that it could be easier for users who are used to using openmp in languages (ie fortran) where end statements are necessary. If an end statement is provided a warning is raised to indicate that it is ignored

nhamidn:
> @yguclu
> 
> > In the PR description you say "Ignore any !$omp end do statements provided by the user", but I am not sure that I understand what you mean.
> > Since we changed our syntax from `!$omp do` to `!$omp for` a while back, I think that we should raise an error when we encounter OpenMP pragmas that do not fit our grammar.
> 
> Sorry this was a typo that I have now updated. It should be "Ignore any `!$omp end for` statements provided by the user". This is done for example [here](https://github.com/pyccel/pyccel/blob/master/tests/internal/scripts/openmp/ex2.py). I discussed the usefulness of this with @nhamidn and he felt that it could be easier for users who are used to using openmp in languages (ie fortran) where end statements are necessary. If an end statement is provided a warning is raised to indicate that it is ignored

you still writing it in the wrong way, it's `#$omp `not `!$omp` in python

pyccel/ast/omp.py
```diff
@@ -76,6 +76,11 @@ def __getnewargs__(self):
         args = (self.txt, self.combined)
         return args
 
+    def __str__(self):
+        instruction = [self.name, self.combined, self.txt]
+        instruction = ' '.join(i for i in instruction if i)
+        return '#$ omp '+instruction
```
yguclu, line 82: Sorry for being pedantic. Unless strictly necessary I would like us to avoid redefining variables, especially if their type changes as a result. Here the last two lines can be merged into one, and I think that the resulting code would be clearer.

pyccel/parser/grammar/openmp.tx
```diff
@@ -42,11 +42,11 @@ OmpConstructOrDirective:
...
-OmpTaskLoopConstruct:   name='taskloop'   clauses*=AllOmpClauses;
+OmpTaskLoopConstruct:   'taskloop'   clauses*=AllOmpClauses;
 OmpSingleConstruct:     'single'          clauses*=AllOmpClauses;
 OmpCriticalConstruct:   name='critical'   clauses*=AllOmpClauses;
 OmpMaskedConstruct:     'masked'          combined=MaskedCombined? clauses*=AllOmpClauses;
-OmpSimdConstruct:       name='simd'       clauses*=AllOmpClauses;
+OmpSimdConstruct:       'simd'       clauses*=AllOmpClauses;
 OmpAtomicConstruct:     name='atomic'     clauses*=AllOmpClauses; // Not all clauses were added (can't test them).
 OmpTaskConstruct:       name='task'       clauses*=AllOmpClauses;
 OmpCancelConstruct:     name='cancel'     clauses*=OmpCancelType;
```
yguclu, line 52: Could somebody explain why some constructs use `name='STRING'` and others just `'STRING'`?

pyccel/parser/syntactic.py
```diff
@@ -1034,6 +1034,7 @@ def _visit_CommentMultiLine(self, stmt):
 
                 txt = com[1:].lstrip()
                 exprs.append(Comment(txt))
+            exprs[-1].set_fst(stmt)
```
yguclu, line 1037: Is this a bug fix?

yguclu (changes requested):
Good job @EmilyBourne! I have a couple of comments and questions.

pyccel/parser/grammar/openmp.tx
EmilyBourne: @nhamidn

pyccel/parser/syntactic.py
EmilyBourne: A very small one yes. I have added it to the PR description

pyccel/ast/omp.py
EmilyBourne: So you prefer:
```
instruction = ' '.join(i for i in [self.name, self.combined, self.txt] if i)
```
?
I disagree that this is clearer. I can rename the list variable if you prefer?

yguclu: No, I mean:
```python
    def __str__(self):
        instruction = (self.name, self.combined, self.txt)
        return '#$ omp ' + ' '.join(i for i in instruction if i)
 ```

yguclu: And yes, the name `instruction` is not really appropriate for a tuple...

EmilyBourne: Ah of course sorry. Fixed: 20dfcbde

yguclu (approved):

pyccel/parser/grammar/openmp.tx
nhamidn: Constructs that don't accept any clauses like `#$omp master`, if it's grammar is : `OmpMasterConstruct:   'master'; ` will generate an error at this line https://github.com/pyccel/pyccel/blob/1621df5c8786fdaba5ec49b956a0ec980bb9c777/pyccel/parser/syntax/openmp.py#L863  saying `AttributeError: 'str' object has no attribute 'expr'`, and this is because textx will not create any object for this construct since we don't save any value in the grammar and will just return a string.

the `name= ` can be removed from the construct that save at least one value different than `name`.

PR #993, opened 2021-10-13
Raise errors for stack arrays with invalid shapes

### LINKED ISSUE #978: Pyccel allows creation of invalid stack arrays
**Describe the bug**
Pyccel does not check whether the requested stack arrays can actually fulfil that role. When targeting Fortran the compiler complains if the code is not valid, but when targeting C the code compiles without problems and just leads to segmentation faults

**Test Code**
The following code:
```python
import numpy as np

@stack_array('V')
def test_stack():
    nx = 5
    ny = 5

    V = np.zeros([nx,ny], dtype=int)
```
generates:
```c
void test_stack(void)
{
    int64_t nx;
    int64_t ny;
    int64_t array_dummy_0001[nx * ny];
    t_ndarray V = (t_ndarray){
        .nd_int64=array_dummy_0001,
        .shape=(int64_t[]){nx, ny},
        .strides=(int64_t[2]){0},
        .nd=2,
        .type=nd_int64,
        .is_view=false
    };
    stack_array_init(&V);
    nx = 5;
    ny = 5;
    array_fill((int64_t)0, V);
}
```

**Expected Behaviour**
An error should be raised at the semantic stage indicating that the shape of the array is incompatible with the stack_array attribute

bauom:
Raise errors when trying to create a stack array from a shape which is unknown when entering the function (and therefore at the array declaration). Fixes #978


EmilyBourne:
Stack arrays can also have the shape provided by function arguments

EmilyBourne:
Is this PR still active?

EmilyBourne (approved):

yguclu:
@bauom Can you please explain in some detail what behavior has been implemented by this PR?

Automatic arrays of non-literal size are somewhat controversial in the computer science community, but they are nevertheless supported by both the C and Fortran languages.

EmilyBourne:
> @bauom Can you please explain in some detail what behavior has been implemented by this PR?
> 
> Automatic arrays of non-literal size are somewhat controversial in the computer science community, but they are nevertheless supported by both the C and Fortran languages.

This PR adds errors for the situation described by issue #978 . Although automatic arrays of non-literal size are supported in Fortran and C, it is only possible if the size is known at the declaration. In Fortran all declarations are done at the top of the function (and we have also followed this convention in C). As a result non-literal sizes are only allowed if the size is constructed from the arguments or via a pure function. This PR adds errors if these restrictions are not followed, while still allowing us to do things like :
```fortran
@stack_array('tmp')
def test_degree(degree : int):
    from numpy import empty

    tmp = empty(degree+1, dtype=float)
    for i in range(degree+1):
        tmp[i]=0.
    return 1
```
(see [test](https://github.com/pyccel/pyccel/blob/master/tests/pyccel/scripts/runtest_degree_in.py))

yguclu:
@EmilyBourne Thanks for the explanation! The title and description of the PR should contain the same information.

EmilyBourne:
> @EmilyBourne Thanks for the explanation! The title and description of the PR should contain the same information.

Better?

pyccel/errors/messages.py
```diff
@@ -24,6 +24,8 @@
 INCOMPATIBLE_REDEFINITION = 'Incompatible redefinition'
 INCOMPATIBLE_REDEFINITION_STACK_ARRAY = 'Cannot change shape of stack array, because it does not support memory reallocation. Avoid redefinition, or use standard heap array.'
 STACK_ARRAY_DEFINITION_IN_LOOP = 'Cannot create stack array in loop, because if does not support memory reallocation. Create array before loop, or use standard heap array.'
+STACK_ARRAY_NON_LITERAL_SHAPE = 'Cannot create stack array from a shape which is unknown at function entry'
```
yguclu, line 27: I suggest to change the name of this variable, as it is a bit confusing for the developer. Maybe `STACK_ARRAY_UNKNOWN_SHAPE`?

pyccel/parser/semantic.py
```diff
@@ -964,6 +964,21 @@ def _assign_lhs_variable(self, lhs, d_var, rhs, new_expressions, is_augassign, *
                     if 'allow_negative_index' in decorators:
                         if lhs in decorators['allow_negative_index']:
                             d_lhs.update(allows_negative_indexes=True)
+                
+                # We cannot allow the definition of a stack array from a shape with non-literal integers
```
yguclu, line 968: This comment line is confusing, see previous comment

tests/epyccel/test_arrays_multiple_assignments.py
```diff
@@ -140,11 +140,14 @@ def f():
 
     # Check that we got exactly 1 Pyccel error
     assert errors.has_errors()
-    assert errors.num_messages() == 1
+    assert errors.num_messages() == 2
```
yguclu, line 143: The comment line should be updated: "1 Pyccel error" -> "2 Pyccel errors"

tests/epyccel/test_arrays_multiple_assignments.py
```diff
@@ -140,11 +140,14 @@ def f():
...
 
     # Check that the error is correct
     error_info = [*errors.error_info_map.values()][0][0]
     assert error_info.symbol  == 'x'
+    assert error_info.message == STACK_ARRAY_NON_LITERAL_SHAPE
+    error_info = [*errors.error_info_map.values()][0][1]
+    assert error_info.symbol  == 'x'
     assert error_info.message == STACK_ARRAY_DEFINITION_IN_LOOP
```
yguclu, line 151: The comment line should be updated: "the error is correct" -> "the errors are correct"

yguclu (changes requested):
Good job!

Before this PR, what kind of errors would we have gotten when running the new unit test? Compile or runtime errors? Is there a difference between C and Fortran?

EmilyBourne:
> Before this PR, what kind of errors would we have gotten when running the new unit test? Compile or runtime errors? Is there a difference between C and Fortran?

For the following example:
```python
@stack_array('x')
def f():
    import numpy as np
    i = 3
    x = np.ones(i, dtype=int)
    return x[0]
```
The fortran code is:
```fortran
function f() result(Out_0001)

    implicit none

    integer(i64) :: Out_0001
    integer(i64) :: i
    integer(i64) :: x(0:i - 1_i64)

    i = 3_i64
    x = 1_i64
    Out_0001 = x(0_i64)
    return

  end function f
```
which fails to compile and raises the error:
```
/home/emily/Code/test/tmp/__pyccel__/tmp.f90:16:24:

   16 |     integer(i64) :: x(0:i - 1_i64)
      |                        1
Error: Variable ‘i’ cannot appear in the expression at (1)
```

The C code is:
```c
int64_t f(void)
{
    int64_t i;
    int64_t array_dummy[i];
    t_ndarray x = (t_ndarray){
        .nd_int64=array_dummy,
        .shape=(int64_t[]){i},
        .strides=(int64_t[1]){0},
        .nd=1,
        .type=nd_int64,
        .is_view=false
    };
    stack_array_init(&x);
    i = 3;
    array_fill((int64_t)1, x);
    return GET_ELEMENT(x, nd_int64, (int64_t)0);
}
```
Which compiles however it will cause runtime errors as the value of `i` is undefined and can therefore be anything

pyccel/parser/semantic.py
EmilyBourne: Fixed [98096de](https://github.com/pyccel/pyccel/pull/993/commits/98096ded80323407e4a8468c0e0d69747973bc46)

tests/epyccel/test_arrays_multiple_assignments.py
EmilyBourne: Fixed [130f7ca](https://github.com/pyccel/pyccel/pull/993/commits/130f7ca137368cbf5575facac9c4f1c28a22a099)

tests/epyccel/test_arrays_multiple_assignments.py
EmilyBourne: Fixed [130f7ca](https://github.com/pyccel/pyccel/pull/993/commits/130f7ca137368cbf5575facac9c4f1c28a22a099)

pyccel/errors/messages.py
EmilyBourne: Fixed [b577d1a](https://github.com/pyccel/pyccel/pull/993/commits/b577d1afaeb07202a2e4bdab122a84b514c50572)

tests/epyccel/test_arrays_multiple_assignments.py
```diff
@@ -8,7 +8,7 @@
                                     ARRAY_DEFINITION_IN_LOOP,
                                     INCOMPATIBLE_REDEFINITION_STACK_ARRAY,
                                     STACK_ARRAY_DEFINITION_IN_LOOP,
-                                    ASSIGN_ARRAYS_ONE_ANOTHER, ARRAY_ALREADY_IN_USE)
+                                    ASSIGN_ARRAYS_ONE_ANOTHER, ARRAY_ALREADY_IN_USE, STACK_ARRAY_NON_LITERAL_SHAPE)
```
yguclu, line 11: We should now import `STACK_ARRAY_UNKNOWN_SHAPE`

yguclu (changes requested):
Can you please update our [tutorial](https://github.com/pyccel/pyccel/blob/master/tutorial/decorators.md#stack-array) with a brief explanation about the restrictions on the shape of the stack arrays?

tests/epyccel/test_arrays_multiple_assignments.py
EmilyBourne: Fixed [e11207b](https://github.com/pyccel/pyccel/pull/993/commits/e11207bc7ceaa38e20e92a9d1d4b8a53920e77cb)

yguclu (approved):

PR #996, opened 2021-10-15
Issue #983 | Add Round function and implement the conversion to C, Fortran, and Python

### LINKED ISSUE #983: Add round function to pyccel
**Describe the bug**
We would like to add the python builtin `round` to pyccel

**Test Code**
```python
if __name__ == '__main__':
     print(round(4.49))
```

nandiniraja348:
Add support for Python's `round` function which is implemented with a banker's round. An integer and a floating point implementation are both required to avoid floating point errors when handling integers. Fixes #983.

A minor bug fix is also included to ensure that the error mode is set back to 'user' when nothing is specified if the developer mode was used previously.


EmilyBourne:
Hi, how are you doing? Are you still working on this PR? Do you need a hand? I just wanted to remind you to keep your branch up to date with the master branch. There have been a few aesthetic changes recently (mainly PR #1012) which should be merged sooner rather than later to avoid accumulating conflicts

pyccel/ast/builtins.py
```diff
@@ -319,6 +320,26 @@ def arg(self):
...
+    """
+    name = 'round'
+    _dtype = NativeReal()
+    _precision = default_precision['real']
+    _rank = 0
+    _shape = ()
+    _order = None
+    _attribute_nodes  = ('_arg',)
+
+    def __init__(self, arg):
```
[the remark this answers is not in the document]
EmilyBourne: Maybe we should leave this improvement for another PR so we can merge this branch?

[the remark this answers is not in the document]
EmilyBourne: I have added the code. We are just missing the printer

EmilyBourne:
`round` in python has a very strange way of handling values that are equally close to 2 integers (see [here](https://docs.python.org/3/library/functions.html#round)):

> For the built-in types supporting [round()](https://docs.python.org/3/library/functions.html#round), values are rounded to the closest multiple of 10 to the power minus ndigits; if two multiples are equally close, rounding is done toward the even choice (so, for example, both round(0.5) and round(-0.5) are 0, and round(1.5) is 2). Any integer value is valid for ndigits (positive, zero, or negative). The return value is an integer if ndigits is omitted or None. Otherwise, the return value has the same type as number.

This needs to be handled, maybe we need to implement our own `round` function in `pyc_math` for [c](https://github.com/pyccel/pyccel/blob/master/pyccel/stdlib/math/pyc_math_c.h) and [fortran](https://github.com/pyccel/pyccel/blob/master/pyccel/stdlib/math/pyc_math_f90.f90)

EmilyBourne (approved):

pyccel/ast/builtins.py
```diff
@@ -354,6 +355,48 @@ def arg(self):
...
+    """ Represents a call to Python's native round() function.
+    """
+    __slots__ = ('_arg', '_ndigits', '_dtype', '_precision')
+    name = 'round'
+    _rank = 0
+    _shape = ()
+    _order = None
+    _attribute_nodes  = ('_arg','_ndigits')
+
+    def __init__(self, number, ndigits = None):
```
aihya, line 369: `self._ndigits` could be simply written as `self._ndigits = ndigits`

aihya (changes requested):

pyccel/ast/builtins.py
EmilyBourne: You're right. It's left over from a moment when I was assigning `0` to `self._ndigits`.

Fixed [af5f147](https://github.com/pyccel/pyccel/pull/996/commits/af5f14731a4a2a14f757e2dc94c18738c889037a)

aihya (approved):

aihya (dismissed):

aihya:
An Error is raised when I tried the following code:
```python3
if __name__ == "__main__":
    print(round(42, ndigits=5))
```

The error:
```
  File "/home/glitch/Documents/nandiniraja348/nandiniraja348_env/bin/pyccel", line 33, in <module>
  File "/home/glitch/Documents/nandiniraja348/pyccel/commands/console.py", line 241, in pyccel
    execute_pyccel(filename,
  File "/home/glitch/Documents/nandiniraja348/pyccel/codegen/pipeline.py", line 382, in execute_pyccel
    generated_program_filepath = src_compiler.compile_program(compile_obj=prog_obj,
  File "/home/glitch/Documents/nandiniraja348/pyccel/codegen/compiling/compilers.py", line 345, in compile_program
  File "/home/glitch/Documents/nandiniraja348/pyccel/codegen/compiling/compilers.py", line 429, in run_command
    raise RuntimeError(err_msg)
/home/glitch/Documents/nandiniraja348/__pyccel__/prog_test_round.f90:9:36:

    9 |   write(*, '(F0.12)', advance="yes") pyc_bankers_round(42_i64, 5_i64)
      |                                    1
Error: There is no specific function for the generic ‘pyc_bankers_round’ at (1)
```

yguclu:
@aihya Good catch! This is because the two functions implemented in Fortran require the first argument to be a float (single or double precision), while here you are passing an integer.

cc: @nandiniraja348 @EmilyBourne

aihya:
Incorrect rounded value when passing non-integer to `ndigits`:
```python
if __name__ == "__main__":
    print(round(42, ndigits=4.2))
```
Output: `41.999991113996`

```python
if __name__ == "__main__":
    print(round(42, ndigits=4.2))
```
Output: `4.200030659267`

EmilyBourne:
@aihya 
Pyccel doesn't translate code which doesn't run in python. Your example:
> ```python
> if __name__ == "__main__":
>     print(round(4.2, ndigits=4.2))
> ```
gives:
```python
<ipython-input-1-dd8329dcb3e7> in <module>
----> 1 print(round(4.2, ndigits=4.2))

TypeError: 'float' object cannot be interpreted as an integer
```

EmilyBourne:
I've just fixed the integer case. I am going to try and rebase this branch to see if we can get this merged

pyccel/stdlib/math/pyc_math_f90.f90
```diff
@@ -351,4 +367,58 @@ elemental function numpy_v2_sign_c64(x) result(Out_0001)
...
+
+    if (ndigits <= 0 .and. (diff == 0.5_f64 .or. diff == -0.5_f64)) then
+        rnd = nint(arg*0.5_f64, kind=i64)*2_i64
+    end if
+
+    rnd = rnd * 10._f64**(-ndigits)
+
+end function pyc_bankers_round_float
+
+pure function pyc_bankers_round_int_8(arg, ndigits) result(rnd)
```
EmilyBourne, line 395:
```suggestion
pure function pyc_bankers_round_int(arg, ndigits) result(rnd)
```

pyccel/stdlib/math/pyc_math_f90.f90
```diff
@@ -351,4 +367,58 @@ elemental function numpy_v2_sign_c64(x) result(Out_0001)
...
+        remainder = modulo(arg, mul_fact)
+        if ( remainder == pivot_point ) then
+            val = (mul_fact - remainder) / mul_fact
+            rnd = (val + IAND(val, 1_i64)) * mul_fact
+        else
+            rnd = ((arg + pivot_point) / mul_fact) * mul_fact
+        endif
+    endif
+
+end function pyc_bankers_round_int_8
```
EmilyBourne, line 422:
```suggestion
end function pyc_bankers_round_int
```

pyccel/ast/builtins.py
```diff
@@ -480,6 +481,51 @@ def arg(self):
     def __str__(self):
         return f'float({self.arg})'
 
+# ===========================================================================
+class PythonRound(PyccelFunction):
+    """
+    Class representing a call to Python's native round() function.
+
+    Class representing a call to Python's native round() function.
```
EmilyBourne, line 489:
```suggestion
    Class representing a call to Python's native round() function
    which rounds a float or integer to a given number of decimals.
```

pyccel/codegen/printing/fcode.py
```diff
@@ -1196,6 +1196,25 @@ def _print_PythonAbs(self, expr):
...
+        args : variable
+        """
+        arg = expr.arg
+        if not isinstance(arg.dtype.primitive_type, PrimitiveFloatingPointType):
+            arg = self._apply_cast(NumpyInt64Type(), arg)
+        ndigits = self._apply_cast(NumpyInt64Type(), expr.ndigits)
+        self.add_import(Import('pyc_math_f90', Module('pyc_math_f90',(),())))
+
+        arg_code = self._print(arg)
+        if ndigits:
```
EmilyBourne, line 1210:
```suggestion
        self.add_import(Import('pyc_math_f90', Module('pyc_math_f90',(),())))

        arg_code = self._print(arg)
        if expr.ndigits:
            ndigits = self._apply_cast(NumpyInt64Type(), expr.ndigits)
```

pyccel/codegen/printing/fcode.py
```diff
@@ -1196,6 +1196,25 @@ def _print_PythonAbs(self, expr):
...
+        if not isinstance(arg.dtype.primitive_type, PrimitiveFloatingPointType):
+            arg = self._apply_cast(NumpyInt64Type(), arg)
+        self.add_import(Import('pyc_math_f90', Module('pyc_math_f90',(),())))
+
+        arg_code = self._print(arg)
+        ndigits = self._apply_cast(NumpyInt64Type(), expr.ndigits) if expr.ndigits \
+                else LiteralInteger(0, NumpyInt64Type())
+        ndigits_code = self._print(ndigits)
+
+        code = pyc_bankers_round({arg_code}, {ndigits_code})
```
EmilyBourne, line 1210:
```suggestion
        code = f'pyc_bankers_round({arg_code}, {ndigits_code})'
```

pyccel/ast/builtins.py
```diff
@@ -480,6 +481,52 @@ def arg(self):
...
+    name = 'round'
+    _rank = 0
+    _shape = None
+    _order = None
+
+    def __init__(self, number, ndigits = None):
+        if ndigits is None:
+            self._class_type = PythonNativeInt()
+        else:
+            self._class_type = number.class_type
```
jalalium, line 509: I think `Booleans` should be handled separately.
```Python
if __name__ == '__main__':
    print(round(True))
```
does not work well in C.

jalalium (changes requested):
I think the round functions implemented, though sound, introduce errors due to floating point arithmetic:
```python

if __name__ == '__main__':
    print(round(2.675, 2))
```
this gives 2.68 instead of 2.67

EmilyBourne:
> I think the round functions implemented, though sound, introduce errors due to floating point arithmetic:
> 
> ```python
> if __name__ == '__main__':
>     print(round(2.675, 2))
> ```
> 
> this gives 2.68 instead of 2.67

This is a very surprising corner case. To me this looks like a Python rounding error rather than a C rounding error. Python implement's banker's round:
> Bankers' rounding takes its name from a history of being used in banks. It rounds x.5 to nearest even number. The reason is to evenly spread rounding of x.5 up and down. This increases the likelihood that the sum of many rounded numbers is closer to the true sum.

So I would expect `2.675` to round to `2.68`. However in Python I have:
```python
In [18]: print(f"{2.675:.16f}")
2.6749999999999998
```
0.49999999999998 therefore rounds down.

It seems that the correction comes from the multiplication by 100:
```python
In [20]: print(f"{2.675*100.0:.16f}")
267.5000000000000000
```
I can fix this. But I wonder if I should given that the C answer is more accurate :thinking:

pyccel/ast/builtins.py
EmilyBourne: Ah, I didn't realise that this would execute in Python :sweat_smile:

EmilyBourne: Fixed : [c618c2c](https://github.com/pyccel/pyccel/pull/996/commits/c618c2c3a85f858b4556bcc3b3257bfcd990b230)

And test added : f12a617

EmilyBourne:
Copy/pasting Python's `round` code from:
https://github.com/python/cpython/blob/0f56adb8d74e703974811620559d96c999334547/Objects/floatobject.c#L985
leads to the same results as our current implementation.

I suspect that the result with the rounding comes from this implementation:
https://github.com/python/cpython/blob/0f56adb8d74e703974811620559d96c999334547/Objects/floatobject.c#L927
which relies on printing floats to strings to truncate. Converting a float to a string, truncating and then converting back to a float would not be good for performance so I would rather keep the current implementation. This will give the same result in most cases. In some cases (such as `2.675, 2`) it may be closer to the expected result, there will probably be other cases where this method suffers from rounding errors instead but on balance I think this is no worse than Python's implementation.

jalalium (approved):
Well done!

PR #997, opened 2021-10-18
Provide Python implementations of OpenMP routines

### LINKED ISSUE #855: A code using the OpenMP Runtime library routines can't work in pure python mode
**Describe the bug**
Code using the OpenMP Runtime library routines can only work when compiled with pyccel, it can't work in pure python mode.

**To Reproduce**
For example this code :
```python
@types('int')
def set_num_threads(n):
    from pyccel.stdlib.internal.openmp import omp_set_num_threads
    omp_set_num_threads(n)

set_num_threads(4)
```

Generate this error when trying to run it in python mode:
```bash/fortran/c
from pyccel.stdlib.internal.openmp import omp_set_num_threads
ModuleNotFoundError: No module named 'pyccel.stdlib.internal.openmp'
```

**Language**
Python

EmilyBourne:
Allow Python calls of OpenMP routines in module `pyccel.stdlib.internal.openmp`, by providing Python implementations with default return values (`0`, `1`, `-1`, `True`, `False`, or `None`). Fixes #855.

### DIFF
--- a/tutorial/builtin-functions.md
+++ b/tutorial/builtin-functions.md
@@ -1,0 +1,75 @@
+# Built-in functions
+
+Python contains a limited number of builtin functions defined [here](https://docs.python.org/3/library/functions.html). Pyccel currently handles a small subset of those functions
+
+| Function | Supported |
+|----------|-----------|
+| **abs** | **Yes** |
+| all | No |
+| any | No |
+| ascii | No |
+| bin | No |
+| **bool** | **Yes** |
+| breakpoint | No |
+| bytearray | No |
+| bytes | No |
+| callable | No |
+| chr | No |
+| classmethod | No |
+| compile | No |
+| **complex** | **Yes** |
+| delattr | No |
+| dict | No |
+| dir | No |
+| divmod | No |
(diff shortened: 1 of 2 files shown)


EmilyBourne:
> > > In practice the new Python implementations are not tested, because we get different results in the serial case. You don't see any problem with that?
> > 
> > 
> > It is normal that we get different results in the serial case for tests that do things like choosing the number of parallel threads. I'm not sure how we could write a pertinent test that does anything more than call the function to check it exists. Do you have a suggestion?
> 
> Rather than skipping every Python test, I suggest that we remove Python from the languages produced by the fixture. Then we can add a single pure-Python test where we do exactly what you describe, i.e. checking that every OpenMP function returns what it is supposed to (either `None`, `0`, or `1`, I believe).

It is important to note that we do **not** skip every python test. Of the 36 tests in the file, only 9 are skipped. If you want a test that checks that the OpenMP functions return what I wrote in the body then I can add one. But personally I don't see the utility. It won't check that the values make sense, and if they need changing then the tests will also need changing

PR #1013, opened 2021-10-30
Add dummy variables where necessary to avoid aliasing in Fortran 

### LINKED ISSUE #981: Pyccel prints forbidden fortran aliasing
# The bug
Fortran has several restrictions on how arguments can be passed. Pyccel does not pay attention to these restrictions when printing. This can therefore lead to code which produces unexpected input

# Example
## Python code:
```python
def mod_values(x : int):
    return x+2,x+3

if __name__ == '__main__':
    a = 3
    b = 4
    c = 5
    a,d = mod_values(a)
    b,e = mod_values(b)
    c,f = mod_values(c)

    print(a)
    print(b)
    print(c)
    print(d)
    print(e)
    print(f)
```

## Fortran code
```fortran
module tmp


  implicit none

  contains

  subroutine mod_values(x, Out_0001, Out_0002)

    implicit none

    integer(i64), intent(out) :: Out_0001
    integer(i64), intent(out) :: Out_0002
    integer(i64), value :: x

    Out_0001 = x + 2_i64
    Out_0002 = x + 3_i64
    return

  end subroutine mod_values

end module tmp
```
```fortran
program prog_prog_tmp

  use tmp

  implicit none

  integer(i64) :: a
  integer(i64) :: b
  integer(i64) :: c
  integer(i64) :: d
  integer(i64) :: e
  integer(i64) :: f

  a = 3_i64
  b = 4_i64
  c = 5_i64
  call mod_values(a, Out_0001 = a, Out_0002 = d)
  call mod_values(b, Out_0001 = b, Out_0002 = e)
  call mod_values(c, Out_0001 = c, Out_0002 = f)
  print *, a
  print *, b
  print *, c
  print *, d
  print *, e
  print *, f

end program prog_prog_tmp
```

## Python Output
```bash
5
6
7
6
7
8
```

## Fortran Output
```bash
                    2
                    6
                    7
                    3
                    7
                    8
```

## Explanation
The variable `x` is assigned to 2 argument variables. One with intent `in`, and one with intent `out`. Intent `out` means that the current value of `x` can be discarded while `in, value` means that it is used and copied into the argument variable. These statements are contradictory 

# Expected behaviour
When a variable is passed to multiple arguments of a function and those arguments have conflicting intents (i.e. `out` and anything else), a copy must be made of the variable before/after the call

# Language
Fortran

EmilyBourne:
Add dummy variables if an argument is used as both an argument and a result (`intent(out)`). This prevents the out variable from being wiped before it is used as the argument. Fixes #981 

**Commit Summary**
- [BUGFIX] Don't check returns inside sub-functions when checking return compatibility


yguclu (commented):
I've noticed that the example Fortran code given in issue #981 can be fixed by simply changing `intent(out)` with `intent(inout)`. This approach does not wipe the input variable (which is passed by `value`) upon entry, and hence its value is effectively copied into the dummy argument before any calculation happens.

Such an approach appears correct for scalar arguments (where we pass read-only arguments by `value`), but would it work for array arguments where we pass `intent(in)` arguments by reference?

EmilyBourne:
> I've noticed that the example Fortran code given in issue #981 can be fixed by simply changing `intent(out)` with `intent(inout)`. This approach does not wipe the input variable (which is passed by `value`) upon entry, and hence its value is effectively copied into the dummy argument before any calculation happens.
> 
> Such an approach appears correct for scalar arguments (where we pass read-only arguments by `value`), but would it work for array arguments where we pass `intent(in)` arguments by reference?
> 
> See comment [here](https://github.com/pyccel/pyccel/issues/981#issuecomment-961141626)

Presumably you mean that we should therefore never print `intent(out)`? I agree that this should work for scalars as we pass by value in the case of `intent(in)` but I don't think it is sufficient for arrays and I think it is easier to implement a solution that works for both. If you prefer though I could use `intent(inout)` for scalar output and retain the copying implemented here for arrays.

Here is an example that would work with this branch (if issue #537 were fixed) but not with the master branch or with your suggested solution:
```
def f(a):
    b = np.array(a.shape[0]-1)
    for i in range(a.shape[0]-1):
        b[i] = np.sum(a[:i+1])
    return b

if __name__ == '__main__':
    c = np.random.randint(20)
    c = f(c)
```

yguclu:
> > I've noticed that the example Fortran code given in issue #981 can be fixed by simply changing `intent(out)` with `intent(inout)`. This approach does not wipe the input variable (which is passed by `value`) upon entry, and hence its value is effectively copied into the dummy argument before any calculation happens.
> > Such an approach appears correct for scalar arguments (where we pass read-only arguments by `value`), but would it work for array arguments where we pass `intent(in)` arguments by reference?
> > See comment [here](https://github.com/pyccel/pyccel/issues/981#issuecomment-961141626)
> 
> Presumably you mean that we should therefore never print `intent(out)`? I agree that this should work for scalars as we pass by value in the case of `intent(in)` but I don't think it is sufficient for arrays and I think it is easier to implement a solution that works for both. If you prefer though I could use `intent(inout)` for scalar output and retain the copying implemented here for arrays.
> 
> Here is an example that would work with this branch (if issue #537 were fixed) but not with the master branch or with your suggested solution:
> 
> ```
> def f(a):
>     b = np.array(a.shape[0]-1)
>     for i in range(a.shape[0]-1):
>         b[i] = np.sum(a[:i+1])
>     return b
> 
> if __name__ == '__main__':
>     c = np.random.randint(20)
>     c = f(c)
> ```

Nice counterexample! :+1:

EmilyBourne:
> Nice counterexample! +1

Does this mean I can put the `needs_inital_review` tag back?

yguclu (approved):

saidctb:
The solution looks good, but maybe in the future we should do something like
```fortran
program prog_prog_test_devel
  use test_devel
  implicit none

  integer(i64) :: a
  integer(i64) :: b
  integer(i64) :: c
  integer(i64) :: d
  integer(i64) :: e
  integer(i64) :: f

  a = 3_i64
  b = 4_i64
  c = 5_i64
  call mod_values(copy(a), Out_0001 = a, Out_0002 = d)
  call mod_values(copy(b), Out_0001 = b, Out_0002 = e)
  call mod_values(copy(c), Out_0001 = c, Out_0002 = f)
  print *, a
  print *, b
  print *, c
  print *, d
  print *, e
  print *, f

end program prog_prog_test_devel
```
where copy is defined as follows
```fortran
  pure elemental function copy(x) result(y)
    integer(i64), value :: x
    integer(i64)        :: y

    y = x
  end function
```

saidctb (approved):

saidctb:
In my computer, I have `gfortran version 7.5.0`, and it gives me the right results when I use the master branch, could this be a compiler issue rather?

EmilyBourne:
> In my computer, I have `gfortran version 7.5.0`, and it gives me the right results when I use the master branch, could this be a compiler issue ?

Some compilers handle this as you would expect, however it is not a compiler bug as the standard says that aliasing of this kind is not allowed. Therefore there is no guarantee that any compiler will do anything specific. On my computer it handles 2 of the 3 calls correctly, so I wonder if there is a race condition involved

saidctb:
> > In my computer, I have `gfortran version 7.5.0`, and it gives me the right results when I use the master branch, could this be a compiler issue ?
> 
> Some compilers handle this as you would expect, however it is not a compiler bug as the standard says that aliasing of this kind is not allowed. Therefore there is no guarantee that any compiler will do anything specific. On my computer it handles 2 of the 3 calls correctly, so I wonder if there is a race condition involved

can you solve the issue in your computer by doing
```fortran
ogram prog_prog_test_devel
  use test_devel
  implicit none

  integer(i64) :: a
  integer(i64) :: b
  integer(i64) :: c
  integer(i64) :: d
  integer(i64) :: e
  integer(i64) :: f

  a = 3_i64
  b = 4_i64
  c = 5_i64
  call mod_values(a+0, Out_0001 = a, Out_0002 = d)
  call mod_values(b+0, Out_0001 = b, Out_0002 = e)
  call mod_values(c+0, Out_0001 = c, Out_0002 = f)
  print *, a
  print *, b
  print *, c
  print *, d
  print *, e
  print *, f

end program prog_prog_test_devel
```

EmilyBourne:
> > > In my computer, I have `gfortran version 7.5.0`, and it gives me the right results when I use the master branch, could this be a compiler issue ?
> > 
> > 
> > Some compilers handle this as you would expect, however it is not a compiler bug as the standard says that aliasing of this kind is not allowed. Therefore there is no guarantee that any compiler will do anything specific. On my computer it handles 2 of the 3 calls correctly, so I wonder if there is a race condition involved
> 
> can you solve the issue in your computer by doing
> ....

Yes this also solves the problem. This is expected as you are no longer aliasing, however this new example is a less general solution which will not work for e.g. instances of classes

PR #1014, opened 2021-10-30
Fix default arguments for non-default precisions

### LINKED ISSUE #1003: Default arguments cannot be provided for non-default precisions
**Describe the bug**
Default arguments cannot be provided for non-default precisions as Pyccel does not seem to visit the `FunctionDef` arguments in the semantic stage

**To Reproduce**
```python
def f(a : 'int32' = 1):
    print(a)

def g():
    f()
```

```bash
pyccel:
 |error [semantic]: tmp.py [5,4]| Incompatible argument 1 (int(kind=8)) passed to function (expected int(kind=4)). Please cast the argument explicitly or overload the function (see https://github.com/pyccel/pyccel/blob/master/tutorial/headers.md for details) (f())
```

If we attempt to specify the correct precision with the following code:
```python
import numpy as np
def f(a : 'int32' = np.int32(1)):
    print(a)

def g():
    f()
```

then we get the error:
```bash
  File "/home/emily/Code/pyccel/pyccel/parser/semantic.py", line 2196, in _visit_FunctionCall
    return self._handle_function(expr, func, args, **settings)
  File "/home/emily/Code/pyccel/pyccel/parser/semantic.py", line 1089, in _handle_function
    self._check_argument_compatibility(new_expr.args, func.arguments,
  File "/home/emily/Code/pyccel/pyccel/parser/semantic.py", line 1036, in _check_argument_compatibility
    if incompatible(i_arg, f_arg):
  File "/home/emily/Code/pyccel/pyccel/parser/semantic.py", line 1023, in <lambda>
    (i_arg.dtype is not f_arg.dtype or \
AttributeError: 'DottedName' object has no attribute 'dtype'
```
The presence of `DottedName` shows that the `FunctionDefArgument` hasn't been visited

**Proposed Solution**
The `FunctionDefArgument` should be visited. When the default argument is a literal of the wrong precision it should be automatically cast to the correct precision

EmilyBourne:
Visit the default value of the `FunctionDefArgument` to allow specifying values non-trivially. Automatically convert literals to the correct precision. Fixes #1003


pyccel/codegen/printing/pycode.py
```diff
@@ -153,7 +153,10 @@ def _print_Variable(self, expr):
 
     def _print_FunctionDefArgument(self, expr):
         if expr.has_default:
-            return '{} = {}'.format(self._print(expr.name), self._print(expr.value))
+            if isinstance(expr.value, FunctionDef):
+                return '{} = {}'.format(self._print(expr.name), self._print(expr.value.name))
+            else:
+                return '{} = {}'.format(self._print(expr.name), self._print(expr.value))
```
yguclu, line 159: I am not sure that I understand this change. Can you provide an example?

EmilyBourne: It handles this case:
```
#$ header function high_valuedarg_1(int, (int)(int))
def high_valuedarg_1(a, function=f1):
    x = function(a)
    return x
```

The default argument is a function so without the correction we get errors like:
```
E     File "/home/runner/work/pyccel/pyccel/tests/epyccel/modules/__epyccel__/highorder_functions_3ukrf85c09cx_3ukrf85c09cx.py", line 49
E       def high_valuedarg_1(a, function = def f1(a):
E                                            ^
E   SyntaxError: invalid syntax
```
(see https://github.com/pyccel/pyccel/runs/4056643786?check_suite_focus=true)

yguclu: I see, thanks!

yguclu (approved):

PR #1019, opened 2021-11-01
Allow `import mod` syntax for a header file

### LINKED ISSUE #1018: Can't use `import mod` syntax for a header file
**Describe the bug**
The `import mod` syntax doesn't work for header files, as they check which targets are required

**To Reproduce**
```python
import numpy as np

def f(x: 'float32[:]'):
    import pyccel.stdlib.internal.blas as blas_mod
    n = np.int32(x.shape[0])
    incx = np.int32(1)
    return blas_mod.sasum (n, x, incx)
```

```bash
pyccel:
 |fatal [semantic]: mod.py [7,11]| Could not find sasum in imported module blas_mod (blas_mod.sasum(FunctionCallArgument('n'), FunctionCallArgument('x'), FunctionCallArgument('incx')))
```

**Expected behavior**
The header should not check for targets. This could also cause problems if the file is imported by 2 files in the tree as the parsing is only done once

EmilyBourne:
Allow `import mod` syntax for a header file. Fixes #1018 

**Commit Summary**
- Add test
- Ensure functions imported with ignore_at_import are correctly added to the namespace
- [BUGFIX] Ensure external function results are printed with correct precision

### DIFF
--- a/pyccel/codegen/printing/fcode.py
+++ b/pyccel/codegen/printing/fcode.py
@@ -1584,7 +1584,7 @@ class FCodePrinter(CodePrinter):
             scope = self.parser.namespace.sons_scopes[name]
             for key,f in scope.imports['functions'].items():
                 if isinstance(f, FunctionDef) and f.is_external:
-                    i = Variable(f.results[0].dtype, name=str(key))
+                    i = f.results[0].clone(str(key))
                     dec = Declare(i.dtype, i, external=True)
                     decs[i] = dec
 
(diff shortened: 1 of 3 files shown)


ratnania:
Can you add an example with a subroutine? I know it should work also from the code you already pushed.

EmilyBourne:
> Can you add an example with a subroutine? I know it should work also from the code you already pushed.

Done

ratnania (approved):

PR #1021, opened 2021-11-02
Fix external function call from function in function

### LINKED ISSUE #1020: External function results are not declared for functions within functions
**Describe the bug**
External functions are not declared for functions within functions

**To Reproduce**
```python
import numpy as np

def test_func():
    def blas_dnrm2(x: 'float64[:]',
                   incx: 'int32' = 1,
                  ):
        """
        Computes the Euclidean norm of a vector.
        """
        from pyccel.stdlib.internal.blas import dnrm2

        n = np.int32(x.shape[0])

        return dnrm2 (n, x, incx)

    y = np.ones(5)
    incx = np.int32(2)
    return blas_dnrm2(y,incx)
```

```fortran
module outer


  use, intrinsic :: ISO_C_Binding, only : i32 => C_INT32_T , i64 => &
      C_INT64_T , f64 => C_DOUBLE
  implicit none

  contains

  function test_func() result(Out_0002)

    implicit none

    real(f64) :: Out_0002
    real(f64), allocatable :: y(:)
    integer(i32) :: incx

    allocate(y(0:4_i64))
    y = 1.0_f64
    incx = 2_i32
    Out_0002 = blas_dnrm2(y, incx)
    if (allocated(y)) then
      deallocate(y)
    end if
    return

    contains
    !________________________________________________!
    !                                                !
    !        Computes the Euclidean norm of a vector.!
    !                                                !
    !________________________________________________!

    function blas_dnrm2(x, incx) result(Out_0001)

      implicit none

      real(f64) :: Out_0001
      real(f64), intent(inout) :: x(0:)
      integer(i32), value :: incx
      integer(i32) :: n

      n = Int(size(x, kind=i64), i32)
      Out_0001 = dnrm2(n, x, incx)
      return

    end function blas_dnrm2

  end function test_func

end module outer
```

**Expected behavior**
There should be a line `real(f64), external :: dnrm2` declaring the function result

**Language**
Fortran

**Additional context**
The same code when not inside a function translates correctly. Ie:
```python
def blas_dnrm2(x: 'float64[:]',
               incx: 'int32' = 1,
              ):
    """
    Computes the Euclidean norm of a vector.
    """
    import numpy as np
    from pyccel.stdlib.internal.blas import dnrm2

    n = np.int32(x.shape[0])

    return dnrm2 (n, x, incx)
```

cc: @ratnania

EmilyBourne:
Fix external function call from function in function (fixes #1020 ) by using the correct namespace. Reduce code duplication and add a test

### DIFF
--- a/pyccel/codegen/printing/fcode.py
+++ b/pyccel/codegen/printing/fcode.py
@@ -305,6 +305,19 @@ class FCodePrinter(CodePrinter):
             self._constantImports.add(constant_name)
         return constant_name
 
+    def _get_external_declarations(self):
+        """
+        Look for external functions and declare their result type
+        """
+        decs = {}
+        for key,f in self._namespace.imports['functions'].items():
+            if isinstance(f, FunctionDef) and f.is_external:
+                i = Variable(f.results[0].dtype, name=str(key))
+                dec = Declare(i.dtype, i, external=True)
+                decs[i] = dec
+
+        return decs
+
     # ============ Elements ============ #
     def _print_PyccelSymbol(self, expr):
         return expr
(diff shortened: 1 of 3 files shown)


ratnania (approved):
Great! I didn't pay attention to this case :)

PR #1022, opened 2021-11-03
Devel blas

ratnania:
The BLAS wrappers are implemented. They will be improved once the `inline` PR is merged

### DIFF
--- a/tests/epyccel/modules/blas.py
+++ b/tests/epyccel/modules/blas.py
@@ -1,0 +1,3317 @@
+# ==============================================================================
+#
+#                                  LEVEL 1
+#
+# ==============================================================================
+
+# ==============================================================================
+def blas_srotg(a: 'float32', b: 'float32',
+               c: 'float32' = 0.,
+               s: 'float32' = 0.,
+              ):
+
+    import numpy as np
+    import pyccel.stdlib.internal.blas as mod_blas
+
+    mod_blas.srotg (a, b, c, s)
+
+    return c, s
+
+# ==============================================================================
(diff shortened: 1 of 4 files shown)


tests/epyccel/modules/blas.py
```diff
@@ -0,0 +1,3317 @@
...
+    lda = m
+    # equation 2
+    if side:
+        flag_side = 'R'
+        lda = n
+    # ...
+
+    # ...
+    flag_uplo = 'U'
+    if lower : flag_uplo = 'L'
```
EmilyBourne, line 751: Why not use an If Ternary operator?

tests/epyccel/modules/blas.py
```diff
@@ -0,0 +1,3317 @@
...
+
+    mod_blas.drotg (a, b, c, s)
+
+    return c, s
+
+# ==============================================================================
+def blas_drotmg(d1: 'float64', d2: 'float64', x1: 'float64', y1: 'float64',
+                param: 'float64[:]'):
+
+    import numpy as np
```
EmilyBourne, line 834: There are lots of unused `import numpy as np` statements in this file. They should be flagged by codacy when it starts working again

EmilyBourne (commented):
Have I understood correctly that the aim is for the file `tests/epyccel/modules/blas.py` to become `pyccel/stdlib/external/blas.pyh` once PR #1017 is merged? And the file `tests/epyccel/test_blas.py` would then be rewritten to compile functions which call scipy?

EmilyBourne:
@rakati I see that you were assigned to this issue in March. Are you working on it at all or should you be removed from the assignees?

PR #1024, opened 2021-11-05
Add documentation for const keyword

MostafaMamouni:
This PR is about adding const keyword docs.

### DIFF
--- a/tutorial/const_keyword.md
+++ b/tutorial/const_keyword.md
@@ -1,0 +1,85 @@
+# Const keyword
+
+In case if you want to make sure that an argument will not change inside its function, Pyccel provides a `const` keyword which is a qualifier of the data type of the variable and which will be converted to its equivalents in C/Fortran, Here is a simple example of its usage:
+
+Here is a working Python code with `const` keyword:
+
+```python
+def func1(arr: 'const int[:]'):
+    # some code
+    return 0
+```
+
+The generated C code:
+
+```C
+#include "boo.h"
+#include "ndarrays.h"
+#include <stdint.h>
+#include <stdlib.h>
+
+
+/*........................................*/
+int64_t func1(t_ndarray arr)
+{
+    /*some code*/
+    return 0;
+}
+/*........................................*/
+```
+
+The fortran equevalent:
+
+```fortran
+module boo
+
+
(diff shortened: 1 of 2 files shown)


EmilyBourne:
As per the review process as described [here](https://github.com/pyccel/pyccel/blob/master/developer_docs/review_process.md), please add the `needs_initial_review` label when your PR is ready for review.

tutorial/const_keyword.md
```diff
@@ -0,0 +1,86 @@
+# Const keyword
+
+In case if you want to make sure that an argument will not change inside its function, Pyccel provides a `const` keyword which is a qualifier of the data type of the variable and which will be converted to its equivalents in C/Fortran, Here is a simple example of its usage:
```
EmilyBourne, line 3: `In case` is not necessary, you can start the sentence with `If`

EmilyBourne, line 3: I suggest using `the target language` instead of `C/Fortran`. It means one less thing to change if/when we add an additional language

EmilyBourne (changes requested):

EmilyBourne (approved):

tutorial/const_keyword.md
```diff
@@ -0,0 +1,86 @@
+# Const keyword
+
+If you want to make sure that an argument will not change inside its function, Pyccel provides a `const` keyword which is a qualifier of the data type of the variable and which will be converted to its equivalents in the target language, Here is a simple example of its usage:
```
yguclu, line 3: I suggest some improvements:
```suggestion
In order to make sure that a function argument is not modified by the function call, Pyccel provides the `const` keyword, which is converted to an equivalent datatype qualifier in the target language. Here is a simple example of its usage:
```

tutorial/const_keyword.md
```diff
@@ -0,0 +1,86 @@
...
+/*........................................*/
+int64_t func1(t_ndarray arr)
+{
+    /*some code*/
+    return 0;
+}
+/*........................................*/
+```
+
+The fortran equevalent:
```
yguclu, line 31: Typo: equevalent -> equivalent

yguclu (changes requested):
Thanks @MostafaMamouni! I have a couple of comments.

yguclu (approved):

PR #1034, opened 2021-11-15
Initialise ndarrays at NULL

### LINKED ISSUE #951: Ndarrays must be initialised at Null
**Describe the bug**
When a ndarray is freed, the contents of the `shape` pointer are checked to see if the call to `free` is necessary. This is done to handle cases where `free` is called twice. However, the `shape` pointer is only initialised if `array_create` is called. This could cause problems in cases where branching leaves a branch where the array is never initialised.

**To Reproduce**
```python
def f(b1 : bool, b2 : bool):
    import numpy as np
    if b1:
        if b2:
            x = np.zeros(3, dtype=int)
            n = x.shape[0]
        else:
            n = 0
    else:
        x = np.zeros(4, dtype=int)
        n = x.shape[0]
    return 5
```

```c
#include "if.h"
#include "ndarrays.h"
#include <stdlib.h>
#include <stdbool.h>
#include <stdint.h>

int64_t f(bool b1, bool b2)
{
    t_ndarray x;
    int64_t n;
    if (b1)
    {
        if (b2)
        {
            x = array_create(1, (int64_t[]){3}, nd_int64);
            array_fill((int64_t)0, x);
            n = 3;
        }
        else
        {
            n = 0;
        }
    }
    else
    {
        x = array_create(1, (int64_t[]){4}, nd_int64);
        array_fill((int64_t)0, x);
        n = x.shape[0];
    }
    free_array(x);

    return 5;
}
```

The contents of `free_array` are:
```
int32_t free_array(t_ndarray arr)
{
    if (arr.shape == NULL)
        return (0);
    free(arr.raw_data);
    arr.raw_data = NULL;
    free(arr.shape);
    arr.shape = NULL;
    free(arr.strides);
    arr.strides = NULL;
    return (1);
}
```

**Expected behavior**
The ndarray object should in initialised with a NULL pointer for the shape

**Language**
C

EmilyBourne:
Avoid potential problems when freeing data by initialising arrays at NULL. Fixes #951

[BUGFIX] Ensure deallocations are printed when avoiding creating unnecessary output variables in C

### DIFF
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -845,6 +845,9 @@ class CCodePrinter(CodePrinter):
 
         if expr.variable.is_stack_array:
             preface, init = self._init_stack_array(expr.variable,)
+        elif declaration_type == 't_ndarray ':
+            preface = ''
+            init    = ' = {.shape = NULL}'
         else:
             preface = ''
             init    = ''
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -1377,18 +1377,19 @@ class CCodePrinter(CodePrinter):
             # Check the Assign objects list in case of
             # the user assigns a variable to an object contains IndexedElement object.
             if not last_assign:
-                return 'return {0};\n'.format(self._print(args[0]))
-
-            # make sure that stmt contains one assign node.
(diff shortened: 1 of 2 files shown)


EmilyBourne:
The `free_array` which used to be correctly printed is now missing

yguclu (approved):

Issue #297, opened 2020-04-08
generate headers files to bypass double compilation


saidctb:
when a flag is passed we generate a header file to accelerate the compilation process if pyccel is called again on the same file, this will help us accelerate the compilation of a project or a folder

EmilyBourne:
Related to issue #218

EmilyBourne:
See branch `devel-issue297` which also fixes issue #344. Currently missing from the header syntax in order to fix this issue is:

- default values
- argument names (for keyword arguments)
- interfaces (specifically we need to be able to match the C name to the interface name)
- class variables
- Functions that are initialisers or freeing functions for modules

If we have to significantly modify the header language to allow argument names and default values, then it may be worth considering implementing issue #641 at the same time

saidctb:
we will face problems, when the default values are imported variables , like:
```python
from mod import x
def f(y=x):
    pass
```
so we might need to add imports in the header files

EmilyBourne:
After #1003 is fixed we may need imports anyway to do:
```
def f(a : 'int32' = np.int32(1)):
```
I think it is a good idea to include the imports anyway as it helps with compiling

EmilyBourne:
This should be done after #641 and stub files should be used rather than headers

PR #1046, opened 2021-11-24
Improve imports to allow importing inline functions

EmilyBourne:
Improve import infrastructure so that all imports are aware of what they are importing (rather than just knowing its name). This allows us to identify whether an imported object is an inline function. The import of a module is modified from being described as a dictionary to being described by a `Module` object. When printing imports import all necessary objects.

The inline functions are now exposed to python. As the wrappers call inline functions, the contents of the function are inlined into the wrapper.

When header files are added in the future (see issue #297 ) inline functions should be printed into the header file so they are available for inlining

**Commit Summary**
- [BUGFIX] Use Variables in `BindCFunctionDef` body not `FunctionDefArgument`s
- Ensure that the renamed object inside `AsName` is a pyccel object
- Add an internal dictionary to `Module` type to allow searching inside a `Module`
- Reorganise `Iterable` to remove the need for importing all possible future iterables (potentially causing circular imports)
- Add `global_funcs` to `InlineFunctionDef`
- Create `PyccelFunctionDef` class
- Ensure that `Import`s are aware of the object being imported after the syntactic stage
- Create `Decorator` class
- Save external libraries into `Module` instances
- Make `ast.numpyext.Shape` into a class so it can be stored in `PyccelFunctionDef`
- Move scipy information to its own file
- Add comparison magic methods to `DottedName`
- Import any objects necessary for inline methods
- Remove unused function `CWrapperCode._print_IndexedElement`
- Wrap inline functions
- [BUGFIX] Not all python `ast` object have line and column information
- [BUGFIX] In `get_filename_from_import` we were importing the python translation of a module, if it was available, rather than the original module because the current working directory is inside `__pyccel__`
- Save functions in a header file into its `Module`
- Remove duplicate functions in `stdlib.internal.lapack`. Keep the correct function signature
- Add a test for imports
- Remove restriction from documentation


tests/pyccel/test_pyccel.py
```diff
@@ -839,11 +839,12 @@ def test_function_aliasing():
...
+@pytest.mark.xdist_incompatible
 def test_inline(language):
     pyccel_test("scripts/decorators_inline.py", language = language)
 
 #------------------------------------------------------------------------------
-@pytest.mark.xfail(reason="Imported inline functions cannot import objects required for their contents")
+@pytest.mark.xdist_incompatible
 def test_inline_import(language):
     pyccel_test("scripts/runtest_decorators_inline.py",
             dependencies = ("scripts/decorators_inline.py"),
```
yguclu, line 850: Why are these functions incompatible with `xdist`?

EmilyBourne: Both functions translate the file `scripts/decorators_inline.py` there is therefore a chance of them trying to modify the same file at the same time. Or (as I saw before I remembered to add it), the python result collection for `test_inline_import` may try to use the pyccelized version of `scripts.decorators_inline`. This is not what is intended and indeed will not work until issue #741 is fixed

pyccel/ast/utilities.py
```diff
@@ -76,32 +70,44 @@ def builtin_function(expr, args=None):
...
+            Import('itertools', AsName(itertools_mod,'itertools')),
+            Import('math', AsName(math_mod,'math')),
+            Import('pyccel', AsName(pyccel_mod,'pyccel'))
+            ])
 if sys.version_info < (3, 10):
     from .builtin_imports import python_builtin_libs
 else:
     python_builtin_libs = set(sys.stdlib_module_names) # pylint: disable=no-member
 
 recognised_libs = python_builtin_libs.union(builtin_import_registery.keys())
```
yguclu, line 94: It has nothing to do with this PR, but... using the pipe operator `|` for the union is more Pythonic:
```suggestion
recognised_libs = python_builtin_libs | builtin_import_registery.keys()
```

EmilyBourne: Huh, I have never seen that before

pyccel/ast/variable.py
```diff
@@ -551,6 +551,15 @@ def name(self):
     def __str__(self):
         return """.""".join(str(n) for n in self.name)
 
+    def __eq__(self, other):
+        return str(self) == other
+
+    def __ne__(self, other):
+        return str(self) != other
```
yguclu, line 558: Wouldn't it be safer to write `str(other)`? Or is that a use case that should never happen?

EmilyBourne: I agree

EmilyBourne: Fixed 80b6fe9f

pyccel/ast/mathext.py
```diff
@@ -2,17 +2,20 @@
...
-from pyccel.ast.internals import PyccelInternalFunction
+from pyccel.ast.core      import PyccelFunctionDef, Module
 from pyccel.ast.datatypes import (NativeInteger, NativeBool, NativeFloat,
                                   default_precision)
+from pyccel.ast.internals import PyccelInternalFunction
+from pyccel.ast.variable  import Constant
 
 __all__ = (
+    'math_mod',
     'math_constants',
```
yguclu, line 18: Shouldn't we remove `math_constants` as well, given that they are contained in `math_mod` now?

EmilyBourne: No, the fortran and c printers seem to use this dictionary to get the values for printing

EmilyBourne: See [here](https://github.com/pyccel/pyccel/blob/master/pyccel/codegen/printing/ccode.py#L1748-L1759) and [here](https://github.com/pyccel/pyccel/blob/master/pyccel/codegen/printing/fcode.py#L159)

pyccel/ast/scipyext.py
```diff
@@ -0,0 +1,20 @@
...
+from .variable import Constant
+
+scipy_pi_const = Constant('float', 'pi', value=pi)
+
+scipy_mod = Module('scipy',
+        variables = (scipy_pi_const,),
+        funcs = (),
+        imports = [
+            Import('constants',Module('constants', (scipy_pi_const,), ()))
+            ])
```
yguclu, line 20: Does the `imports` argument in a `Module` constructor always refer to submodules?

EmilyBourne: Yes

EmilyBourne: No, they refer to anything imported in the module. E.g:
```
from numpy import empty

def f():
    a = empty(3)
```
Here the import is in the module not the function

pyccel/ast/utilities.py
```diff
@@ -76,32 +70,44 @@ def builtin_function(expr, args=None):
...
+    if source[0] in python_builtin_libs and source[0] not in builtin_import_registery.keys():
+        return True
+    else:
+        return source_name in builtin_import_registery
 
+#==============================================================================
+def collect_relevant_imports(func_module, targets):
+    """
+    Extract all objects necessary to create imports from a module given a list of targets
+    """
```
yguclu, line 110: Could you please expand this docstring and describe that the output of this function is a list of tuple pairs `(local_name, imported_object)`?

Also, is `func_module` of type `Module`? If yes, we could change its name to just `module`...

pyccel/codegen/printing/ccode.py
```diff
@@ -252,7 +253,12 @@ def __init__(self, parser, prefix_module = None):
 
     def get_additional_imports(self):
         """return the additional imports collected in printing stage"""
-        return self._additional_imports
+        return [i.source if isinstance(i, Import) else i for i in self._additional_imports]
+
+    def get_additional_import_objects(self):
+        """return the additional imports collected in printing stage"""
+        return [i if isinstance(i, Import) else Import(i, Module(i, (),())) \
+                for i in self._additional_imports]
```
yguclu, line 261: Does `_additional_imports` contain both `str` and `Module` objects now?

EmilyBourne: In `CCode` yes, this seemed easier for the developers than needing to remember to create an empty `Module` everywhere we want to import a C header. It could probably be improved by declaring all the modules we will use as global variables

EmilyBourne: Do you want me to do this?

pyccel/ast/mathext.py
yguclu: I see. That raises another question though: why don't we just print the Python value of those constants instead of raising a `NotImplementedError`?

EmilyBourne: I think that is a question for a separate issue. But I agree, the value in the python constant is not necessarily less precise than the C equivalent (e.g. `M_PI`) and even if it is, then I think it is better to print something than to refuse completely

pyccel/ast/scipyext.py
yguclu: That's what I thought. Then it is not correct to say that the `scipy` module imports the `constants` module. In fact if you try
```python
import scipy
c = scipy.constants
```
you will get
```python
AttributeError: module 'scipy' has no attribute 'constants'
```

It seems to me that we are missing the concept of package, which may contain sub-packages and modules.

EmilyBourne: Hmm, I think you are correct. But I am not sure we need to add packages now. In most packages you can access the sub-packages directly. E.g:
```
import numpy as np
r = np.random
```

pyccel/ast/utilities.py
EmilyBourne: Done cf9f6e76

pyccel/codegen/printing/ccode.py
yguclu: Sorry, I cannot understand your explanation. Why are you talking about headers? And what about global variables?

This is my simple observation: after introducing a new `Module` class, I would expect an homogeneous list of `str` to become an homogeneous list of `Module`. Mixing `str` and `Module` objects seems a bit messy to me.

EmilyBourne: Up till now all the imports in `_additional_imports` are imports which import a whole `Module` (or a C header file) i.e. `math`, `stdlib`, etc

I think the developers won't want to write functions like:
```
    def _print_MathIsnan(self, expr):
        """Convert a Python expression with a math isnan function call to C
        function call"""
        # add necessary include
        self._additional_imports.add(Import('math', Module('math', (),())))
        arg = expr.args[0]
        if arg.dtype is NativeInteger():
            code_arg = self._print(NumpyFloat(arg))
        else:
            code_arg = self._print(arg)
        return "isnan({})".format(code_arg)
```
I think this will be forgotten all the time.

My suggestion is to create `Import` objects for all the imports that are done in this file and to have them as global variables (as opposed to `_additional_imports` which is a class variable) then we can do:
```
self._additional_imports.add(math_import)
```

EmilyBourne: But it means I have to find them all

pyccel/ast/scipyext.py
yguclu: In your example the `numpy.__init__` module actually imports the sub-package `numpy.random`.

I also don't know if we need packages now. I am just worried that we are mixing two different concepts, namely "imported module" vs. "sub-package", which are a bit tricky to grasp in Python. This might become a problem in the future.

pyccel/codegen/printing/ccode.py
EmilyBourne: Fixed 6373f476

yguclu: Thanks for the explanation! It is a good idea and it looks clean.

pyccel/ast/core.py
```diff
@@ -248,27 +247,46 @@ class AsName(Basic):
 
     Parameters
     ==========
-    name   : str
-             original name of variable or function
+    obj    : Basic or BasicType
+             The variable, function, or module being renamed
```
yguclu, line 251: Could you please add an example to the docstring where `obj` is a `type`?

pyccel/ast/core.py
```diff
@@ -248,27 +247,46 @@ class AsName(Basic):
...
     _attribute_nodes = ()
 
-    def __init__(self, name, target):
-        self._name = name
+    def __init__(self, obj, target):
+        if PyccelAstNode.stage != "syntactic":
+            assert (isinstance(obj, Basic) and \
+                    not isinstance(obj, PyccelSymbol)) or \
+                   (isinstance(obj, type) and \
+                   Basic in obj.__mro__)
```
yguclu, line 263: Is there an advantage compared to `issubclass(obj, Basic)`?

pyccel/ast/core.py
```diff
@@ -2679,6 +2744,34 @@ def reinstate_presence_checks(self):
...
+    name : str
+           The name of the function
+    func_class : type inheriting from PyccelInternalFunction / PyccelAstNode
+                 The class which should be instantiated upon a FunctionCall
+                 to this FunctionDef object
+    """
+    def __init__(self, name, func_class):
+        assert isinstance(func_class, type) and \
+                (PyccelInternalFunction in func_class.__mro__ or \
+                 PyccelAstNode in func_class.__mro__)
```
yguclu, line 2768: Again, why not using `issubclass`?

pyccel/ast/numpyext.py
```diff
@@ -1311,81 +1317,101 @@ def is_elemental(self):
...
-    'randint'   : NumpyRandint,
-}
+#==============================================================================
+
+numpy_target_swap = {
+        numpy_funcs['full_like']  : numpy_funcs['full'],
+        numpy_funcs['empty_like'] : numpy_funcs['empty'],
+        numpy_funcs['zeros_like'] : numpy_funcs['zeros'],
+        numpy_funcs['ones_like']  : numpy_funcs['ones']
+    }
```
yguclu, line 1417: What is this dictionary for?

pyccel/ast/mathext.py
yguclu: Could you please create an issue for the constants? It sounds like a good first issue for new contributors.

pyccel/ast/numpyext.py
EmilyBourne: It is for replacing `Import('numpy',AsName(NumpyFullLike,'func'))` with `Import('numpy',AsName(NumpyFull,'func'))`. If I don't swap out the object of the `AsName` in the `Import` target then `_print_Import` prints:
```
from numpy import full_like as full
```

pyccel/ast/core.py
EmilyBourne: I believe the difference is that `Basic in obj.__mro__` is equivalent to : `issubclass(obj, Basic) or isinstance(obj, Basic)`
However this isn't needed here, the only reason that I didn't use `issubclass` is because it didn't show up when I googled how to do this ^^'

pyccel/ast/core.py
EmilyBourne: Fixed d0cbebc3

pyccel/ast/core.py
EmilyBourne: Fixed 0a810e67

pyccel/ast/core.py
EmilyBourne: Fixed 0a810e67

pyccel/ast/mathext.py
EmilyBourne: See #1047

pyccel/ast/core.py
yguclu: Thanks!

pyccel/ast/mathext.py
yguclu: Thanks!

pyccel/ast/numpyext.py
yguclu: Understood! Thanks

pyccel/ast/scipyext.py
yguclu: Should we move this topic to a new discussion?

yguclu (approved):

pyccel/ast/scipyext.py
EmilyBourne: Done. See https://github.com/pyccel/pyccel/discussions/1049

saidctb (approved):

Issue #956, opened 2021-09-06
Changing the precision of a variable doesn't raise an error


EmilyBourne:
**Describe the bug**
Changing the precision of a variable doesn't raise an error in [this function](https://github.com/pyccel/pyccel/blob/master/pyccel/parser/semantic.py#L1173-L1459), like changing the dtype does. The error should be raised in the semantic step in the indicated function

**To Reproduce**
The following code:
```python
import numpy as np

a = np.ones(4,dtype=np.int32)
a = np.zeros(4,dtype=np.int64)
```

Compiles and generates:
```bash/fortran/c
program prog_tmp


  use, intrinsic :: ISO_C_Binding, only : i32 => C_INT32_T , i64 => &
      C_INT64_T
  implicit none

  integer(i32), allocatable :: a(:)

  allocate(a(0:3_i64))
  a = 1_i32
  a = 0_i64

end program prog_tmp
```

**Expected behavior**
The following error should be raised:
```
ERROR at annotation (semantic) stage
pyccel:
 |error [semantic]: tmp.py [4]| Variable has already been defined with type 'int32'. Type 'int64' in assignment is incompatible (a=<pyccel.ast.numpyext.NumpyZeros object at 0x7fdca66610b0>)
```

**Additional Information**
This may break the compilation of certain files that used this (especially header files), however if something breaks this means that the result was incorrect already

PR #1052, opened 2021-12-19
Return numpy scalar types

### LINKED ISSUE #735: Pyccel can only return native python types for scalars

EmilyBourne:
Allow numpy scalar types to be returned. Fixes #735. Add -1 to indicate the default python precision. This allows us to differentiate between native python types and numpy types with the same precision (e.g. `float` and `float64`). This differentiation ensures that the final type matches the type returned by python exactly

**Commit Summary**
- Add `-1` to indicate the native python precision
- Add wrapper functions to convert to numpy scalar types
- [BUG FIX] Use `print_kind` for precision
- [BUG FIX] Map `numpy.absolute` to `numpy.abs`


yguclu:
@EmilyBourne Why is this PR blocked by issue #956?

EmilyBourne:
> @EmilyBourne Why is this PR blocked by issue #956?

Because I was lazy when I wrote it and didn't want to change the test twice. I went back and removed the dependence in commit 850386a733bb9b602cec87b3874c298772bcf602 but apparently I forgot to change the description

pyccel/ast/internals.py
```diff
@@ -262,3 +267,31 @@ def symbols(names):
...
+                else o.precision for o in objs if o.dtype is dtype)
+    else:
+        return max(default_precision[str(o.dtype)] if o.precision == -1 \
+                else o.precision for o in objs)
+
+def get_final_precision(obj):
+    """
+    Get the usable precision of an object
+    """
+    return default_precision[str(obj.dtype)] if obj.precision == -1 else obj.precision
```
yguclu, line 297: Please expand the docstring of this function. What do you mean with "final" and "usable" precision?

EmilyBourne: I mean the precision that you can use to print, ie 8 instead of -1 for a default precision `float`. I'm not sure how to express this. Should I just describe the if statement in the docstring?

yguclu: Thanks for the explanation. I would add it verbatim to the docstring, as well as a description of the if statement as you suggested.

Since this function is used all over the library, it is good to be verbose! 😉

yguclu:
Can you please explain why using `-1` instead of `default_precision[dtype]` helps in the calculation of the final type? Does this have to do with automatic casting in an expression?

Are there any unit tests that were failing before this PR, which could help with my understanding?

pyccel/ast/internals.py
EmilyBourne: Fixed : bbb4598

EmilyBourne:
> Can you please explain why using `-1` instead of `default_precision[dtype]` helps in the calculation of the final type? Does this have to do with automatic casting in an expression?
> 
> Are there any unit tests that were failing before this PR, which could help with my understanding?

Without a default precision which is indicated as such (i.e. -1) there is no way to differentiate between a python `float` and a numpy `float64`. For example the following test would fail:
```
def test_type():
     @types('float')
     @types('float64')
     def f(a):
          return a
          
     new_f = epyccel(f)
     x = np.float64(3.5)
     assert type(f(3.5)) is type(new_f(3.5))
     assert type(f(x)) is type(new_f(x))
```

I had the option to either always return numpy types, or differentiate between the 2. I thought it was better to differentiate, as I didn't see any reason for a code which does not use numpy to return a numpy type

yguclu:
Thanks, now I finally understand your changes! 😅
I think that the PR description is too succinct, can you please expand it a bit?

EmilyBourne:
> Thanks, now I finally understand your changes! 😅 I think that the PR description is too succinct, can you please expand it a bit?

Better?

pyccel/parser/semantic.py
```diff
@@ -1336,9 +1339,19 @@ def _assign_lhs_variable(self, lhs, d_var, rhs, new_expressions, is_augassign, *
...
+                        internal_precision != get_final_precision(var)):
+                    # Get type name from cast function (handles precision implicitly)
+                    try:
+                        d1 = DtypePrecisionToCastFunction[var.dtype.name][var.precision].name
+                    except KeyError:
+                        d1 = str(var.dtype)
+                    try:
+                        d2 = DtypePrecisionToCastFunction[dtype.name][precision].name
+                    except KeyError:
+                        d2 = str(var.dtype)
```
yguclu, line 1352: Is there no way to handle this without `try/except`?

yguclu: Ah sorry, never mind. You always raise an error after these checks.

pyccel/ast/builtins.py
```diff
@@ -817,9 +817,14 @@ def print_string(self):
...
-            dtype = str(self.dtype),
-            precision = self.precision*8 if self.precision else ''))
+        prec = self.precision
+        if prec == -1:
+            return LiteralString("<class '{dtype}'>".format(
+                dtype = str(self.dtype)))
+        else:
+            return LiteralString("<class '{dtype}{precision}'>".format(
+                dtype = str(self.dtype),
+                precision = prec*8 if prec else ''))
```
yguclu, line 827: Could we not modify the original code just a bit?
```python
        prec = self.precision
        return LiteralString("<class '{dtype}{precision}'>".format(
            dtype = str(self.dtype),
            precision = '' if prec in [None, -1] else prec*8))
```

pyccel/ast/cwrapper.py
```diff
@@ -399,9 +406,11 @@ def generate_datatype_error(variable):
     """
     dtype     = variable.dtype
 
-    if isinstance(dtype, NativeBool):
+    if variable.precision == -1:
+        precision = 'navive '
```
yguclu, line 410: Did you mean "native"?

EmilyBourne: Fixed 22f1f32

pyccel/ast/builtins.py
EmilyBourne: Fixed 88439ee

yguclu (approved):

yguclu:
There appears to be a small merge conflict

yguclu:
There is one failing test, which was expected

PR #1056, opened 2021-12-23
Return Numpy arrays

MostafaMamouni:
This pull request is about handling functions that return Numpy arrays and all the related aspects of this feature.

What will be done:
- [x] Return contiguous Numpy arrays in C
- [x] Return contiguous Numpy arrays in Fortran
- [x] Add essential unit tests
- [x] Check for memory leaks


pyccel/parser/semantic.py
```diff
@@ -6,7 +6,8 @@
 
 # pylint: disable=R0201, missing-function-docstring
 
-from itertools import chain
+from collections import OrderedDict
```
EmilyBourne, line 9: I don't see where you use this, but you probably don't need it. `dict` is ordered since python3.6 and python3.5 is now obsolete so we no longer need to use it

pyccel/codegen/printing/ccode.py
```diff
@@ -1344,9 +1344,12 @@ def _print_Interface(self, expr):
...
         if expr.is_inline:
             return ''
-
-        if len(expr.results) > 1:
+        results = expr.results
+        if len(results) > 1:
             self._additional_args.append(expr.results)
+        if isinstance(results[0], Variable) and results[0].allocatable:
+            results[0].allocatable = False
+            results[0].is_pointer = True
```
EmilyBourne, line 1352: Why do you need a pointer? Why only the 1st object that is returned? What happens if we do:
```
def f():
    import numpy as np
    a = np.ones(5)
    b = np.ones(3)
    return a,b
```

pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -214,12 +214,17 @@ int32_t free_array(t_ndarray arr)
...
 int32_t free_pointer(t_ndarray arr)
 {
-    if (arr.is_view == false || arr.shape == NULL)
+    if (arr.shape == NULL)
         return (0);
     free(arr.shape);
     arr.shape = NULL;
     free(arr.strides);
     arr.strides = NULL;
+    if (arr.is_view == false)
```
EmilyBourne, line 223: Can an object have `is_view ==false` and be a pointer?

EmilyBourne: I suspect you are following confusing Fortran implementations needed for the wrapper. I'm not sure this is needed in C. (or in Fortran outside of the wrapper)

pyccel/codegen/printing/ccode.py
MostafaMamouni: you're right, we have no need to use a pointer.
This code should be at the semantic stage, to prevent creating the allocate nodes for the assigned objects to functions that return an allocatable object (for both C and Fortran). that's why we should disable `allocatable` for the assigned objects by disabling it for the returned objects of functions, The problem is this breaks Fortran, causing a wrong data-typing for the return object.

Here is the current behavior of allocatable:
python:
```python
from numpy import ones

def create_array():
	z = ones(5)
	return z

def return_array():
	a = create_array()
	b = ones(5)
	return a

if __name__ == '__main__':
	c = return_array()
	print(c)
```
C:
module:
```C
#include "poo.h"
#include <stdlib.h>
#include "ndarrays.h"


t_ndarray create_array(void)
{
    t_ndarray z = {.shape = NULL};
    z = array_create(1, (int64_t[]){5}, nd_double);
    array_fill((double)1.0, z);
    return z;
}
t_ndarray return_array(void)
{
    a = array_create(1, (int64_t[]){5}, nd_double);
    a = create_array();
    b = array_create(1, (int64_t[]){5}, nd_double);
    array_fill((double)1.0, b);
    free_array(b);
    return a;
}
```
the program:
```C
#include "poo.h"
#include <stdlib.h>
#include "ndarrays.h"
#include <stdio.h>
#include <stdint.h>
int main()
{
    int64_t i;
    c = array_create(1, (int64_t[]){5}, nd_double);
    c = return_array();
    printf("%s", "[");
    for (i = 0; i < 4; i += 1)
    {
        printf("%.12lf ", GET_ELEMENT(c, nd_double, (int64_t)i));
    }
    printf("%.12lf]\n", GET_ELEMENT(c, nd_double, (int64_t)4));
    free_array(c);
    return 0;
}
```
the same problem in Fortran (creating a `deallocate` node for an object that is already assigned to a function that returns an allocatable object)

Here is what happens if we modify `allocatable` attribute in a `Variable` that will be returned in a function at the assigning stage (in this case, we have `a = create_array()` and `c = return_array()`)

the C code will be perfect, which means deallocate nodes won't be created for `a` (at `return_array`) and (`c` at the main function), and `a` and `c` will be declared properly.

for Fortran it's a bit problematic,
here is the generated code:
the module:
```Fortran
module poo


        C_INT64_T
  implicit none

  contains

  function create_array() result(z)

    implicit none

    real(f64) :: z(0:4_i64)

    allocate(z(0:4_i64))
    z = 1.0_f64
    return

  end function create_array

  function return_array() result(a)

    implicit none

    real(f64) :: a(0:4_i64)
    real(f64), allocatable :: b(:)

    a = create_array()
    allocate(b(0:4_i64))
    b = 1.0_f64
    if (allocated(b)) then
      deallocate(b)
    end if
    return

  end function return_array

end module poo
```

the program:
```Fortran
program prog_prog_poo

  use poo

        C_INT64_T
  implicit none

  real(f64) :: c(0:4_i64)

  c = return_array()
  print *, c
  if (allocated(c)) then
    deallocate(c)
  end if

end program prog_prog_poo
```

`a` and `c` are not declared properly, they are missing just the allocatable attribute

MostafaMamouni: i think we should modify `allocatable` after the object gets the right declaration format

pyccel/stdlib/ndarrays/ndarrays.c
MostafaMamouni: it's @bauom idea 😅 after working with pointer, I think we don't need it now since we don't need the returned object to be a pointer, I forget why needed it 🤔

pyccel/codegen/printing/ccode.py
EmilyBourne: > to prevent creating the allocate nodes for the assigned objects to functions that return an allocatable object

I presume you mean `deallocate` nodes?

I am a little confused. It seems you are setting `allocatable = False` (means size is known or variable points to another variable) in order to avoid the creation of a `deallocate` node. Doesn't [this line](https://github.com/pyccel/pyccel/blob/return_arrays/pyccel/parser/semantic.py#L2841) prevent the creation of the `deallocate` node?

I am not sure what has happened in the program to get an array that is not allocatable but has a `deallocate` node

pyccel/parser/semantic.py
```diff
@@ -2198,6 +2199,8 @@ def _visit_Assign(self, expr, **settings):
                 results = func.results
                 if results:
                     if len(results)==1:
+                        if isinstance(results[0], Variable) and results[0].allocatable:
+                           results[0].allocatable = False
```
EmilyBourne, line 2203: What is an array that is not `allocatable`, nor a `stack_array`, nor `is_pointer`? If it is not `allocatable` then you can't use `allocate` to define its size

pyccel/codegen/printing/ccode.py
MostafaMamouni: 🤦‍♂️ im so sorry for not mentioning "memory leaks", i was playing around it 😅.
in the examples above of the C code, these two lines
```C
    a = array_create(1, (int64_t[]){5}, nd_double);
    a = create_array();
```
are causing memory leaks, the same thing for
```C
    c = array_create(1, (int64_t[]){5}, nd_double);
    c = return_array();
```

MostafaMamouni: So we need to not allocate what is already allocated inside a function

EmilyBourne: Ah you are right. Isn't this happening **because** you have set `allocatable=False`? As the code knows that `allocatable` won't be called it is initialising the data straight away. It is the equivalent of `real(f64) :: c(0:4_i64)` in fortran where the data has also already been initialised to a known size

MostafaMamouni: for C it's working perfectly, but fortran seems to need extra work, I think we should keep the memory leaks fix at the semantic since it's for both C and Fortran, and I hope there is a way to fix that Fortran problem at the printer,
what do you think ?

EmilyBourne: If there are memory leaks then it is not working perfectly. I think we should have `allocatable = True`. I think having `allocatable = False` is causing the memory leaks as well as the fortran problems. The `deallocate` at the end of the function should be handled where it is created not by creating a new kind of array with:
- allocatable=False
- pointer=False
- stack_array=False

as I can't tell how this array is supposed to be initialised

tests/pyccel/scripts/test.json
```diff
@@ -0,0 +1,66 @@
+{
```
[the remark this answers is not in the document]
MostafaMamouni: yes indeed.

pyccel/codegen/printing/fcode.py
```diff
@@ -1819,6 +1819,10 @@ def _print_FunctionDef(self, expr):
             return ''
         self.set_scope(expr.scope)
 
+        for n in expr.results:
+            if isinstance(n, Variable) and n.is_ndarray:
+                raise NotImplementedError("Cannot convert a function ({}) that returns an array".format(expr.name))
```
EmilyBourne, line 1824: I think this should be in `_print_BindCFunctionDef`, otherwise you won't be able to add tests in `tests/pyccel`

tests/epyccel/test_epyccel_functions.py
```diff
@@ -111,29 +111,32 @@ def f2(x):
...
     @types('int [:]')
     def f3(x):
         from numpy import empty_like
         y = empty_like(x)
         y[:] = x - 1
         return y
 
-    with pytest.raises(PyccelError):
-        epyccel(f3)
+    f = epyccel(f3, language=language)
```
EmilyBourne, line 122: You should still catch the expected error

tests/epyccel/test_epyccel_functions.py
```diff
@@ -111,29 +111,32 @@ def f2(x):
 
 #------------------------------------------------------------------------------
 # Semantic error doesn't need testing in multiple languages
-def test_decorator_f3():
+def test_decorator_f3(language):
```
EmilyBourne, line 114: This comment can be removed

tests/epyccel/test_epyccel_functions.py
EmilyBourne: Actually you may only need to catch it for Fortran and C. Python should work out of the box as there is no wrapper

EmilyBourne:
Still needed:
- tests with function calls (e.g. executables in `tests/pyccel/`)
- tests with interfaces
- tests with multiple objects in the return statement

.travis.yml
```diff
@@ -94,6 +94,7 @@ jobs:
         - $HOME/Library/Caches/pip/
 
   - os: windows
+    if: branch = master
```
EmilyBourne, line 97: Pyccel no longer uses travis for its CI tests. To activate/deactivate certain tests you need to use this file:
https://github.com/pyccel/pyccel/blob/master/.github/workflows/Github_pytest.yml

tests/pyccel/scripts/return_numpy_arrays.py
```diff
@@ -0,0 +1,13 @@
+import numpy as np
+
+def single_return(x: 'int'):
+    return np.ones(x)
```
EmilyBourne, line 4: See issue #1130 . I think you will need to name the arrays before calling return for the tests to work

tests/pyccel/test_pyccel.py
```diff
@@ -652,6 +652,10 @@ def test_arrays_view(language):
             [int] * 6 + [int] * 10
     pyccel_test("scripts/arrays_view.py", language=language, output_dtype=types)
 
+def test_return_numpy_arrays(language):
+    types = [int] * 5 + [(type(np.array),) * 2]
```
EmilyBourne, line 656: These types concern the values that will be read from the terminal. It is important so that the test function knows how to parse the output (ints are parsed differently to floats and differently to strings). You cannot specify `type(np.array)` here as this does not tell it how to read it. Here you should have: `int` (all the values are ints so you shouldn't need to create a list of the right size)

.travis.yml
EmilyBourne: That said, unless you only have a problem on a specific platform it is better to test locally as you can pyccelize the file yourself and see what the error is. The pyccel tests give very little information when the compilation fails

EmilyBourne: You can get rid of all changes to this file using `git checkout master -- .travis.yml`

tests/pyccel/scripts/return_numpy_arrays.py
```diff
@@ -1,15 +1,23 @@
...
 
 a = single_return()
-b,c = multi_returns()
+b = f(1, 3)
+c = f(1., 3.)
+d = f(False, True)
+e = f(1+2j, 3+4j)
 
-print(a, b, c)
+print(a, b, c, d, e)
```
EmilyBourne, line 23: These statements need to be in a main for a program to be generated. I.e. this section should begin with:
```
if __name__ == '__main__':
```

tests/epyccel/test_epyccel_return_arrays.py
```diff
@@ -0,0 +1,67 @@
...
+import sys
+import pytest
+from numpy.random import randint, uniform
+import numpy as np
+
+sys.path.append('recognised_functions')
+
+from test_numpy_funcs import (min_int, max_int, min_int8, max_int8,
+                                min_int16, max_int16, min_int32, max_int32, max_int64, min_int64)
+from test_numpy_funcs import max_float, min_float, max_float32, min_float32,max_float64, min_float64
```
EmilyBourne, line 11: `from recognised_functions.test_numpy_funcs`

pyccel/ast/utilities.py
```diff
@@ -442,7 +442,7 @@ def collect_loops(block, indices, new_index, language_has_vectors = False, resul
 
             if assigns:
                 # For now we do not handle memory allocation in loop unravelling
-                if any(v.rank > 0 for v in func_vars1) or any(v.rank > 0 for v in func_vars1):
+                if any(v.rank > 0 for v in func_vars1) or any(v.rank > 0 for v in func_vars1) or any(v.rank > 0 for v in func_results):
```
EmilyBourne, line 445: The first 2 `any`s are the same. One should be removed

pyccel/codegen/printing/ccode.py
```diff
@@ -1092,7 +1092,10 @@ def _cast_to(self, expr, dtype, precision):
 
     def _print_DottedVariable(self, expr):
         """convert dotted Variable to their C equivalent"""
-        return '{}.{}'.format(self._print(expr.lhs), self._print(expr.name))
+        if self.stored_in_c_pointer(expr):
+            return '(*{}.{})'.format(self._print(expr.lhs), self._print(expr.name))
```
EmilyBourne, line 1096: Why not `{}->{}`?
Won't the `*` be added in `self._print(expr.lhs)`?

pyccel/codegen/printing/fcode.py
```diff
@@ -2842,7 +2844,7 @@ def _print_FunctionCall(self, expr):
                     newarg = arg
                 args.append(FunctionCallArgument(newarg, key))
             results = list(lhs_vars.values())
-            if len(func_results) == 1:
+            if len(func_results) == 1 and func_results[0].rank == 0:
```
EmilyBourne, line 2847: You could use `is_function` here

pyccel/parser/semantic.py
```diff
@@ -993,7 +995,8 @@ def _assign_lhs_variable(self, lhs, d_var, rhs, new_expressions, is_augassign, *
 
                 # ...
                 # Add memory allocation if needed
-                if lhs.allocatable:
+                array_declared_in_function = isinstance(rhs, FunctionCall) and not isinstance(rhs.funcdef, PyccelFunctionDef) and not rhs.funcdef.is_elemental and not isinstance(lhs, HomogeneousTupleVariable)
+                if lhs.allocatable and not array_declared_in_function and not isinstance(rhs, PyccelOperator):
```
EmilyBourne, line 999: Why do you need ` and not isinstance(rhs, PyccelOperator)` here ? That means you don't want to allocate `c` when you do:
```python
def f(a : 'int[:]', b : 'int[:]'):
    c = a+b
```
Doesn't it?
Isn't this what is causing all the seg faults in the tests?

pyccel/codegen/printing/ccode.py
MostafaMamouni: i don't remember adding this code

pyccel/parser/semantic.py
MostafaMamouni: to prevent creating an allocate node for `single_return` in `a = single_return() + 1`

pyccel/codegen/printing/ccode.py
```diff
@@ -1633,9 +1636,15 @@ def _print_AliasAssign(self, expr):
 
         lhs_address = VariableAddress(lhs_var)
         # Ensure everything which can be stored in a VariableAddress is
-        try:
+        if isinstance(rhs_var, (Variable, IndexedElement)):
             rhs_address = VariableAddress(rhs_var)
-        except TypeError:
+        elif isinstance(rhs_var, FunctionCall):
+            res = rhs_var.funcdef.results
+            if len(res)!= 0 or res[0].is_pointer:
```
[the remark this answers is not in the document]
MostafaMamouni: 😅 this too, i don't remember adding it

pyccel/codegen/printing/ccode.py
EmilyBourne: No this is right. It is `expr` that is stored in a c pointer not `expr.lhs`

tests/pyccel/scripts/return_numpy_arrays_xfails.py
```diff
@@ -0,0 +1,14 @@
+# pylint: disable=missing-function-docstring, disable=unused-variable, missing-module-docstring/
+#
+## This test should be moved to return_numpy_arrays.py once get fixed
```
EmilyBourne, line 3: Fixed for fortran. Won't fix for c

tests/pyccel/test_pyccel.py
```diff
@@ -659,7 +659,15 @@ def test_return_numpy_arrays(language):
     types += [bool]*2 # 2 bools for d
     pyccel_test("scripts/return_numpy_arrays.py", language=language, output_dtype=types)
 
-@pytest.mark.xfail()
+@pytest.mark.parametrize( 'language', (
+        pytest.param("fortran", marks = pytest.mark.fortran),
+        pytest.param("c", marks = [
+            pytest.mark.xfail(reason="Unravelling loops is not working yet."),
```
EmilyBourne, line 665: Can we change this and the one above. It is not that the unravelling isn't working, it's that we don't want a hidden allocation. This is the desired behaviour. So here we should have:
```
            pytest.mark.skip(reason="Hidden memory allocation triggers pyccel error"),
```

tests/errors/semantic/non_blocking/ARRAYS_IN_MULTIPLE_RETURNS.py
```diff
@@ -0,0 +1,7 @@
+# pylint: disable=missing-function-docstring, missing-module-docstring, no-value-for-parameter
```
EmilyBourne, line 1: Why do you disable `no-value-for-parameter`? Please don't disable pylint errors without good reason

tests/epyccel/test_epyccel_return_arrays.py
```diff
@@ -0,0 +1,148 @@
+# pylint: disable=wrong-import-position, disable=missing-function-docstring, missing-module-docstring/
```
EmilyBourne, line 1: Why do you disable `wrong-import-position`? Please don't disable pylint errors without good reason

`reimported` has a good reason. You can disable this

tests/epyccel/test_epyccel_return_arrays.py
```diff
@@ -0,0 +1,148 @@
...
+import sys
+from numpy.random import randint, uniform
+import numpy as np
+import pytest
+
+sys.path.append('recognised_functions')
```
EmilyBourne, line 7: Modifying the `sys.path` is hacky and dangerous. In addition you don't use the hack so I don't understand why you have done this

pyccel/ast/bind_c.py
```diff
@@ -317,8 +317,8 @@ def wrap_array(var, scope, persistent):
     assigns = [Assign(sizes[i], var.shape[i]) for i in range(var.rank)]
     variables = [bind_var, *sizes]
     if not persistent:
-        ptr_var = var.clone(scope.get_new_name(var.name+'_ptr'),
-                is_pointer=True, allocatable=False)
+        ptr_var = Variable(dtype=var.dtype, name=scope.get_new_name(var.name+'_ptr'),
+                memory_handling='alias', rank=var.rank, order=var.order, shape=var.shape)
         alloc = Allocate(ptr_var, shape=var.shape, order=var.order, status='unallocated')
         copy  = Assign(ptr_var, var)
```
EmilyBourne, line 323: Bad merge. Should be:
```python
    variables = [bind_var, *sizes]
    if not persistent:
        ptr_var = var.clone(scope.get_new_name(var.name+'_ptr'),
                memory_handling = 'alias')
```

pyccel/codegen/printing/ccode.py
```diff
@@ -1639,7 +1639,7 @@ def _print_AliasAssign(self, expr):
             rhs_address = VariableAddress(rhs_var)
         elif isinstance(rhs_var, FunctionCall):
             res = rhs_var.funcdef.results
-            if len(res)!= 0 and res[0].is_pointer:
+            if len(res)!= 0 and res[0].is_alias:
                 rhs_address = rhs_var
```
EmilyBourne, line 1643: Bad merge?

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -304,6 +306,78 @@ def get_static_declare_type(self, variable):
...
+                     self.scope.get_temporary_variable(NativeInteger(),
+                         result.name+'_size') for _ in range(result.rank)
+                     ]
+            nd_var = self.scope.get_temporary_variable(dtype_or_var = NativeVoid(),
+                    name = result.name,
+                    is_pointer = True,
+                    allocatable = False)
+            body.append(Allocate(result, shape = sizes, order = result.order,
+                status='unallocated'))
+            body.append(AliasAssign(DottedVariable(NativeVoid(), 'raw_data', is_pointer = True,
```
EmilyBourne, line 338: This should be raising an error. The `is_pointer` kwarg no longer exists

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -740,14 +814,15 @@ def get_module_exec_function(self, expr, exec_func_name):
...
+                    nd_var = self.scope.get_temporary_variable(dtype_or_var = v,
                             name = v.name,
                             memory_handling = 'alias'
                             )
                     alloc = Allocate(nd_var, shape=sizes, order=nd_var.order, status='unallocated')
                     body.append(alloc)
                     # Save raw_data into ndarray to obtain useable pointer
-                    set_data = Assign(DottedName(nd_var, 'raw_data'), VariableAddress(var))
+                    set_data = AliasAssign(DottedVariable(NativeVoid(), 'raw_data',
+                            is_pointer = True, lhs=nd_var), VariableAddress(var))
```
EmilyBourne, line 825: See above. `is_pointer` is gone

pyccel/codegen/printing/fcode.py
```diff
@@ -1229,19 +1229,18 @@ def _print_Declare(self, expr):
...
-        is_pointer = var.is_pointer
-        is_target = var.is_target
-        is_const = var.is_const
-        is_stack_array = var.is_stack_array
-        is_optional = var.is_optional
-        is_private = var.is_private
-        is_static = expr.static
-        is_external = expr.external
-        intent = expr.intent
-        intent_in = intent and intent != 'out'
```
EmilyBourne: This parameter (`intent_in`) seems to have been lost in the merge

pyccel/codegen/printing/fcode.py
```diff
@@ -1314,7 +1314,7 @@ def _print_Declare(self, expr):
             if is_alias:
                 allocatablestr = ', pointer'
 
-            elif allocatable and not intent_in:
+            elif on_heap and not intent:
```
EmilyBourne, line 1317: We should keep `intent_in`

pyccel/codegen/printing/ccode.py
```diff
@@ -1639,7 +1639,7 @@ def _print_AliasAssign(self, expr):
             rhs_address = VariableAddress(rhs_var)
         elif isinstance(rhs_var, FunctionCall):
             res = rhs_var.funcdef.results
-            if len(res)!= 0 and res[0].is_pointer:
+            if len(res)!= 0 and res[0].is_alias:
```
EmilyBourne, line 1642: Did you mean to lose this variable?

MostafaMamouni: it should behave the same as `is_pointer`, why it will be lost ?

EmilyBourne: I've just checked. It was my diff

pyccel/codegen/printing/ccode.py
```diff
@@ -1009,12 +1009,14 @@ def function_signature(self, expr, print_arg_names = True):
         String
             Signature of the function
         """
+        if len(expr.results) > 1:
+            self._additional_args.append(expr.results)
```
EmilyBourne, line 1013: If you `append` then you also need to `pop` somewhere

tests/pyccel/scripts/test.json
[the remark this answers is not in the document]
EmilyBourne: It's back

tests/pyccel/test_pyccel.py
```diff
@@ -652,6 +652,29 @@ def test_arrays_view(language):
             [int] * 6 + [int] * 10
     pyccel_test("scripts/arrays_view.py", language=language, output_dtype=types)
 
+def test_return_numpy_arrays(language):
+    types = [int]*4 # 4 ints for a
+    types += [int]*2 # 2 ints for b
+    types += [float]*2 # 2 floats for c
+    types += [bool]*2 # 2 bools for d
+    types += [complex]*2 # 2 complexs for e
+    types += [int]*5 # 5 ints for f
```
EmilyBourne, line 661: `f` contains floats

tests/pyccel/scripts/return_numpy_arrays.py
```diff
@@ -0,0 +1,30 @@
...
+def f(a, b):
+    c = np.array([a,b])
+    return c
+
+a = single_return()
+b = f(1, 3)
+c = f(1., 3.)
+d = f(False, True)
+e = f(1+2j, 3+4j)
+f,g = multi_returns()
```
EmilyBourne, line 27: It is a bad idea to use the same name (`f`) for the function and the variable. It seems that this doesn't raise an error. I will open an issue

EmilyBourne: #1156

pyccel/parser/semantic.py
```diff
@@ -988,7 +990,9 @@ def _assign_lhs_variable(self, lhs, d_var, rhs, new_expressions, is_augassign, *
 
                 # ...
                 # Add memory allocation if needed
-                if lhs.on_heap:
+                array_declared_in_function = isinstance(rhs, FunctionCall) and not isinstance(rhs.funcdef, PyccelFunctionDef) \
+                                            and not rhs.funcdef.is_elemental and not isinstance(lhs, HomogeneousTupleVariable)
```
EmilyBourne, line 994: This needs expanding to handle the case where the array is declared as the result of a function returning multiple elements (called from here : https://github.com/pyccel/pyccel/blob/master/pyccel/parser/semantic.py#L2423 or here : https://github.com/pyccel/pyccel/blob/master/pyccel/parser/semantic.py#L2431-L2432 )

EmilyBourne: I'm not sure if it can be done with the information currently passed to the function. You may need to add an extra argument

tests/pyccel/test_pyccel.py
EmilyBourne: @MostafaMamouni `h` is created in `multi_returns` with the command `x = np.ones(5)`. The default dtype is float so `h` contains floats not ints

pyccel/parser/semantic.py
EmilyBourne: @MostafaMamouni don't forget about this (very) old comment

tests/pyccel/test_pyccel.py
```diff
@@ -652,6 +652,29 @@ def test_arrays_view(language):
             [int] * 6 + [int] * 10
     pyccel_test("scripts/arrays_view.py", language=language, output_dtype=types)
 
+def test_return_numpy_arrays(language):
+    types = [int]*4 # 4 ints for a
+    types += [int]*2 # 2 ints for b
+    types += [float]*2 # 2 floats for c
+    types += [complex]*2 # 2 complexs for e
+    types += [bool]*2 # 2 bools for d
```
EmilyBourne, line 660: Isn't `d` printed before `e`?

pyccel/parser/semantic.py
```diff
@@ -944,7 +946,8 @@ def _assign_lhs_variable(self, lhs, d_var, rhs, new_expressions, is_augassign, *
 
             d_lhs = d_var.copy()
             # ISSUES #177: lhs must be a pointer when rhs is heap array
-            self._ensure_target(rhs, d_lhs)
+            if not arr_in_multirets:
+                self._ensure_target(rhs, d_lhs)
```
EmilyBourne, line 950: This `if` also needs adding round the other 2 places where this function is called

tests/epyccel/test_epyccel_return_arrays.py
```diff
@@ -0,0 +1,244 @@
...
+
+    assert np.allclose(f_integer16_output, test_int16_output)
+
+    f_integer32_output = epyccel_func(integer32, integer32)
+    test_int32_output = return_array(integer32, integer32)
+
+    assert np.allclose(f_integer32_output, test_int32_output)
+
+    # the if block should be removed after resolving (https://github.com/pyccel/pyccel/issues/735).
+    if sys.platform != 'win32':
```
EmilyBourne, line 193: Issue #735 was fixed in January

rmahjoubi (approved):

tests/epyccel/test_epyccel_return_arrays.py
```diff
@@ -0,0 +1,239 @@
...
+    epyccel_func = epyccel(return_array, language=language)
+
+    f_bl_true_output = epyccel_func(True, True)
+    test_bool_true_output = return_array(True, True)
+
+    f_bl_false_output = epyccel_func(False, False)
+    test_bool_false_output = return_array(False, False)
+
+    assert np.allclose(f_bl_true_output, test_bool_true_output)
+    assert np.allclose(f_bl_false_output, test_bool_false_output)
```
EmilyBourne, line 166: Please add dtype checks for this test

tests/epyccel/test_epyccel_return_arrays.py
```diff
@@ -0,0 +1,239 @@
...
+        def single_return():
+            from numpy import array
+            return array([1,2,3,4])
+        b = single_return()+1
+
+        return b
+
+    epyccel_function = epyccel(return_arrays_in_expression, language=language)
+
+    assert np.allclose(epyccel_function(), return_arrays_in_expression())
```
EmilyBourne, line 239: Please check return type

tests/ndarrays/leaks_check.py
```diff
@@ -50,6 +54,7 @@ def conditional_alloc(b1 : bool, b2 : bool):
 
     # testing garbage collecting in a Program
 
+    z = return_array()
```
EmilyBourne, line 57: Please add a leak check for multiple returns

tests/pyccel/scripts/return_numpy_arrays_xfails.py
EmilyBourne: This comment appears to be incorrect

EmilyBourne (changes requested):
Well done on this PR! It was a lot of work but we are nearly there :smile:

saidctb:
Great work!
I found a small bug when we have an example like
```python
import numpy as np
def f(n:int):
    a = np.ones((n,n))
    return a
    
def g():
    n = 4
    b = 2*f(n)
 ```
 it translates to
 ```fortran
 
  subroutine f(n, a) 

    implicit none

    real(f64), allocatable, intent(out) :: a(:,:)
    integer(i64), value :: n

    allocate(a(0:n - 1_i64, 0:n - 1_i64))
    a = 1.0_f64
    return

  end subroutine f

  subroutine g() 

    implicit none

    integer(i64) :: n
    real(f64), allocatable :: b(:,:)
    real(f64), allocatable :: Dummy_0000(:,:)

    n = 4_i64
    allocate(b(0:size(a, 1_i64, i64) - 1_i64, 0:size(a, 2_i64, i64) - &
          1_i64))
    call f(n, a = Dummy_0000)
    b = 2_i64 * Dummy_0000
    if (allocated(b)) then
      deallocate(b)
    end if

  end subroutine g
```
which doesn't compile because `a` is not defined

saidctb:
I have also noticed that we allocate the array twice in the `bind_c` function, so I made some tests to see the maximum memory allocation using Pyccel vs NumPy on my computer.
with NumPy I was able to allocate a 2D array of floats with the shape `(32000,35000)` and with Pyccel I was able to allocate up to `(21000,30000)` so there is a factor of `1.77`.
Maybe we can improve the way we handle it in the bind_c function and avoid the extra allocation by directly using the returned array.

EmilyBourne:
> Maybe we can improve the way we handle it in the bind_c function and avoid the extra allocation by directly using the returned array.

The problem is that `allocatable`s are automatically deallocated once they go out of scope. So if we point at the allocatable then by the time we get back to c we will be pointing at freed memory which is obviously dangerous

saidctb:
> > Maybe we can improve the way we handle it in the bind_c function and avoid the extra allocation by directly using the returned array.
> 
> The problem is that `allocatable`s are automatically deallocated once they go out of scope. So if we point at the allocatable then by the time we get back to c we will be pointing at freed memory which is obviously dangerous

Maybe this rule doesn't apply with bind(c) functions.
[here](https://www.ibm.com/docs/en/xl-fortran-aix/15.1.2?topic=29113-example-allocatable-pointer-arguments) we have an example with a returned `allocatable` array in a bind(c) function

EmilyBourne:
> > > Maybe we can improve the way we handle it in the bind_c function and avoid the extra allocation by directly using the returned array.
> > 
> > 
> > The problem is that `allocatable`s are automatically deallocated once they go out of scope. So if we point at the allocatable then by the time we get back to c we will be pointing at freed memory which is obviously dangerous
> 
> Maybe this rule doesn't apply with bind(c) functions. [here](https://www.ibm.com/docs/en/xl-fortran-aix/15.1.2?topic=29113-example-allocatable-pointer-arguments) we have an example with a returned `allocatable` array in a bind(c) function

In the example you are linking to the allocatable is `intent(out)` which is why it is not deallocated. However the only reason they can do this is because the C side is using `#include "ISO_Fortran_binding.h"`. This was introduced in [Fortran 2018] (https://fortranwiki.org/fortran/show/Fortran+2018). Pyccel usually follows the Fortran 2003 standard as Fortran compilers are very slow to be updated. It seems that support for ISO_Fortran_binding is now available for the GNU and intel compilers (see [here](https://fortranwiki.org/fortran/show/Fortran+2018+status) but not for PGI. So using this would make it impossible to use PGI to accelerate with pyccel.

If we don't use ISO_Fortran_binding then we can't return an array directly, instead we need to wrap it in a `c_ptr` which leads to the "result variable" going out of scope

saidctb:
> > > > Maybe we can improve the way we handle it in the bind_c function and avoid the extra allocation by directly using the returned array.
> > > 
> > > 
> > > The problem is that `allocatable`s are automatically deallocated once they go out of scope. So if we point at the allocatable then by the time we get back to c we will be pointing at freed memory which is obviously dangerous
> > 
> > 
> > Maybe this rule doesn't apply with bind(c) functions. [here](https://www.ibm.com/docs/en/xl-fortran-aix/15.1.2?topic=29113-example-allocatable-pointer-arguments) we have an example with a returned `allocatable` array in a bind(c) function
> 
> In the example you are linking to the allocatable is `intent(out)` which is why it is not deallocated. However the only reason they can do this is because the C side is using `#include "ISO_Fortran_binding.h"`. This was introduced in [Fortran 2018] (https://fortranwiki.org/fortran/show/Fortran+2018). Pyccel usually follows the Fortran 2003 standard as Fortran compilers are very slow to be updated. It seems that support for ISO_Fortran_binding is now available for the GNU and intel compilers (see [here](https://fortranwiki.org/fortran/show/Fortran+2018+status) but not for PGI. So using this would make it impossible to use PGI to accelerate with pyccel.
> 
> If we don't use ISO_Fortran_binding then we can't return an array directly, instead we need to wrap it in a `c_ptr` which leads to the "result variable" going out of scope

I see, one solution is to provide both ways of doing it because we can generate the two solutions, or we can do a more complicated one using the `save` attribute in a module variable
somehting like
```fortran
module dealloc_test

   use, intrinsic :: iso_c_binding, only : c_double, c_ptr, c_size_t, c_associated, c_loc

   implicit none

   real(c_double), allocatable, target, protected, save :: data_array(:) ! Named, protected target data object

contains

   function alloc(size) bind(C, name='alloc') result(rv)

      ! Argument list
      integer(c_size_t), value, intent(in) :: size
      ! Function result
      type (c_ptr) :: rv

      ! Local variables
      integer(c_size_t) :: i

      ! Handling elided for an existing allocation of data_array
      allocate( data_array(size) )

      rv = c_loc(data_array)

      return

   end function

   subroutine dealloc(ptr) bind(C, name='dealloc')

      ! Argument list
      type(c_ptr), value, intent(in) :: ptr

      ! Check if input pointer is associated with module variable
      if ( c_associated(ptr, c_loc(data_array)) ) then
         deallocate( data_array )
      end if

      return

   end subroutine

end module
```

EmilyBourne:
> > > > > Maybe we can improve the way we handle it in the bind_c function and avoid the extra allocation by directly using the returned array.
> > > > 
> > > > 
> > > > The problem is that `allocatable`s are automatically deallocated once they go out of scope. So if we point at the allocatable then by the time we get back to c we will be pointing at freed memory which is obviously dangerous
> > > 
> > > 
> > > Maybe this rule doesn't apply with bind(c) functions. [here](https://www.ibm.com/docs/en/xl-fortran-aix/15.1.2?topic=29113-example-allocatable-pointer-arguments) we have an example with a returned `allocatable` array in a bind(c) function
> > 
> > 
> > In the example you are linking to the allocatable is `intent(out)` which is why it is not deallocated. However the only reason they can do this is because the C side is using `#include "ISO_Fortran_binding.h"`. This was introduced in [Fortran 2018] (https://fortranwiki.org/fortran/show/Fortran+2018). Pyccel usually follows the Fortran 2003 standard as Fortran compilers are very slow to be updated. It seems that support for ISO_Fortran_binding is now available for the GNU and intel compilers (see [here](https://fortranwiki.org/fortran/show/Fortran+2018+status) but not for PGI. So using this would make it impossible to use PGI to accelerate with pyccel.
> > If we don't use ISO_Fortran_binding then we can't return an array directly, instead we need to wrap it in a `c_ptr` which leads to the "result variable" going out of scope
> 
> I see, one solution is to provide both ways of doing it because we can generate them both, or we can do a more complicated one using the `save` attribute in a module variable something like

I don't think using `save` or a module variable would be sufficient to fix the problem. If the function is called twice with different 
arguments and the result is saved in 2 different variables then the second call will overwrite the contents of the first array.

Generating different code depending on the available standard is an interesting idea. I would be for this as it would simplify a few things, but it is completely out of the scope of this PR. We should open a discussion to talk about this

pyccel/errors/messages.py
```diff
@@ -8,6 +8,7 @@
 	that do not have any parameters.
 """
 
+ARRAYS_IN_MULTIPLE_RETURNS = "returning an array in multiple returns is not supported yet."
```
EmilyBourne, line 11: This error is no longer needed

EmilyBourne: Nor is this one:
https://github.com/pyccel/pyccel/blob/71ac69a8719fd3aa9b6f73241fd8a56703f297d4/pyccel/errors/messages.py#L34

saidctb:
> > > > > > Maybe we can improve the way we handle it in the bind_c function and avoid the extra allocation by directly using the returned array.
> > > > > 
> > > > > 
> > > > > The problem is that `allocatable`s are automatically deallocated once they go out of scope. So if we point at the allocatable then by the time we get back to c we will be pointing at freed memory which is obviously dangerous
> > > > 
> > > > 
> > > > Maybe this rule doesn't apply with bind(c) functions. [here](https://www.ibm.com/docs/en/xl-fortran-aix/15.1.2?topic=29113-example-allocatable-pointer-arguments) we have an example with a returned `allocatable` array in a bind(c) function
> > > 
> > > 
> > > In the example you are linking to the allocatable is `intent(out)` which is why it is not deallocated. However the only reason they can do this is because the C side is using `#include "ISO_Fortran_binding.h"`. This was introduced in [Fortran 2018] (https://fortranwiki.org/fortran/show/Fortran+2018). Pyccel usually follows the Fortran 2003 standard as Fortran compilers are very slow to be updated. It seems that support for ISO_Fortran_binding is now available for the GNU and intel compilers (see [here](https://fortranwiki.org/fortran/show/Fortran+2018+status) but not for PGI. So using this would make it impossible to use PGI to accelerate with pyccel.
> > > If we don't use ISO_Fortran_binding then we can't return an array directly, instead we need to wrap it in a `c_ptr` which leads to the "result variable" going out of scope
> > 
> > 
> > I see, one solution is to provide both ways of doing it because we can generate them both, or we can do a more complicated one using the `save` attribute in a module variable something like
> 
> I don't think using `save` or a module variable would be sufficient to fix the problem. If the function is called twice with different arguments and the result is saved in 2 different variables then the second call will overwrite the contents of the first array.
> 
> Generating different code depending on the available standard is an interesting idea. I would be for this as it would simplify a few things, but it is completely out of the scope of this PR. We should open a discussion to talk about this

Both my suggestion can be done in another PR, I think it's fine to work with what we have now for the moment.
For my first suggestion, you are right it only works with one array and one allocation, to make it general we need to have a dictionary to keep track of all the allocated arrays, so each time the allocate function is called we add a new array to the dictionary, and when we want to deallocate it we can remove it from the dictionary.

EmilyBourne:
@saidctb 
> Great work! I found a small bug ...

This is now fixed. I have also added a test. See 435de3dc

saidctb:
> ```python
>     b = 2*f(n)
> ```

@EmilyBourne 
Is there a reason why  the statement
`b = 2*f(n)`
is translated to
```fortran
   call f(n, a = Dummy_0000)
   allocate(b(0:size(Dummy_0000, 1_i64, i64) - 1_i64, 0:size(Dummy_0000, 2_i64, i64) - &
         1_i64))
   b = 2_i64 * Dummy_0000
```
and not to 
```fortran
   b = 2_i64 * f(n)
```
because this will cause problems if the function is not pure

EmilyBourne:
> @EmilyBourne Is there a reason why the statement `b = 2*f(n)` is translated to
> 
> ```fortran
>    call f(n, a = Dummy_0000)
>    allocate(b(0:size(Dummy_0000, 1_i64, i64) - 1_i64, 0:size(Dummy_0000, 2_i64, i64) - &
>          1_i64))
>    b = 2_i64 * Dummy_0000
> ```
> 
> and not to
> 
> ```fortran
>    b = 2_i64 * f(n)
> ```

Unfortunately there doesn't seem to be a way to obtain the expected indexing for the allocatable if we do `b = 2_i64 * f(n)` or `b = f(n)`. So we end up with indexing starting at `1` which then means that all our interactions with the arrays are wrong.

> because this will cause problems if the function is not pure

What problems are you thinking of?

saidctb:
> > @EmilyBourne Is there a reason why the statement `b = 2*f(n)` is translated to
> > ```fortran
> >    call f(n, a = Dummy_0000)
> >    allocate(b(0:size(Dummy_0000, 1_i64, i64) - 1_i64, 0:size(Dummy_0000, 2_i64, i64) - &
> >          1_i64))
> >    b = 2_i64 * Dummy_0000
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and not to
> > ```fortran
> >    b = 2_i64 * f(n)
> > ```
> 
> Unfortunately there doesn't seem to be a way to obtain the expected indexing for the allocatable if we do `b = 2_i64 * f(n)` or `b = f(n)`. So we end up with indexing starting at `1` which then means that all our interactions with the arrays are wrong.
> 
> > because this will cause problems if the function is not pure
> 
> What problems are you thinking of?

let's say we have a function that is not pure where it updates a global variable every time you enter it
```python
def f():
    a[0] += 1
    return np.ones(10)
expr = f() + f()
```
this code gives different results than something like
```python
def f():
    a[0] += 1
    return np.ones(10)
temp = f()
expr = temp + temp
```

EmilyBourne:
> > > @EmilyBourne Is there a reason why the statement `b = 2*f(n)` is translated to
> > > ```fortran
> > >    call f(n, a = Dummy_0000)
> > >    allocate(b(0:size(Dummy_0000, 1_i64, i64) - 1_i64, 0:size(Dummy_0000, 2_i64, i64) - &
> > >          1_i64))
> > >    b = 2_i64 * Dummy_0000
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > and not to
> > > ```fortran
> > >    b = 2_i64 * f(n)
> > > ```
> > 
> > 
> > Unfortunately there doesn't seem to be a way to obtain the expected indexing for the allocatable if we do `b = 2_i64 * f(n)` or `b = f(n)`. So we end up with indexing starting at `1` which then means that all our interactions with the arrays are wrong.
> > > because this will cause problems if the function is not pure
> > 
> > 
> > What problems are you thinking of?
> 
> let's say we have a function that is not pure where it updates a global variable every time you enter it
> 
> ```python
> def f():
>     a[0] += 1
>     return np.ones(10)
> expr = f() + f()
> ```
> 
> this code gives different results than something like
> 
> ```python
> def f():
>     a[0] += 1
>     return np.ones(10)
> temp = f()
> expr = temp + temp
> ```

I would agree if that was how it was translated. But the translation is:
```
call f(Out_0001 = Dummy_0000)
call f(Out_0001 = Dummy_0001)
allocate(expr(0:9_i64))
expr = Dummy_0000 + Dummy_0001
```

saidctb:
> > > > @EmilyBourne Is there a reason why the statement `b = 2*f(n)` is translated to
> > > > ```fortran
> > > >    call f(n, a = Dummy_0000)
> > > >    allocate(b(0:size(Dummy_0000, 1_i64, i64) - 1_i64, 0:size(Dummy_0000, 2_i64, i64) - &
> > > >          1_i64))
> > > >    b = 2_i64 * Dummy_0000
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > and not to
> > > > ```fortran
> > > >    b = 2_i64 * f(n)
> > > > ```
> > > 
> > > 
> > > Unfortunately there doesn't seem to be a way to obtain the expected indexing for the allocatable if we do `b = 2_i64 * f(n)` or `b = f(n)`. So we end up with indexing starting at `1` which then means that all our interactions with the arrays are wrong.
> > > > because this will cause problems if the function is not pure
> > > 
> > > 
> > > What problems are you thinking of?
> > 
> > 
> > let's say we have a function that is not pure where it updates a global variable every time you enter it
> > ```python
> > def f():
> >     a[0] += 1
> >     return np.ones(10)
> > expr = f() + f()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > this code gives different results than something like
> > ```python
> > def f():
> >     a[0] += 1
> >     return np.ones(10)
> > temp = f()
> > expr = temp + temp
> > ```
> 
> I would agree if that was how it was translated. But the translation is:
> 
> ```
> call f(Out_0001 = Dummy_0000)
> call f(Out_0001 = Dummy_0001)
> allocate(expr(0:9_i64))
> expr = Dummy_0000 + Dummy_0001
> ```

ah okay, you are right, I thought that the expression was simplified

tests/epyccel/test_epyccel_return_arrays.py
```diff
@@ -0,0 +1,279 @@
...
+            return ones(n)
+        b = single_return(n)+1
+
+        return b
+
+    epyccel_function = epyccel(return_arrays_in_expression2, language=language)
+
+    n = randint(5)
+
+    assert np.allclose(epyccel_function(n), return_arrays_in_expression2(n))
```
EmilyBourne, line 279: One last missing `dtype` check. I think this was my fault. Sorry

EmilyBourne (approved):
This looks good now :smile: . I suspect @yguclu may want to update the version number for this PR.

EmilyBourne:
I have contacted Codacy's team about the failing test. I am not sure why it is trying to use the django plugin

yguclu:
> This looks good now smile . I suspect @yguclu may want to update the version number for this PR.

Indeed! I think it is time for Pyccel version 1.6! :wink:

yguclu (changes requested):
Great job! Please update the version to `1.6.0`

pyccel/codegen/printing/ccode.py
```diff
@@ -1029,9 +1031,13 @@ def get_var_arg(arg, var):
                 return code
 
             var_list = [a.var for a in args]
-            arg_code_list = [self.function_signature(var, False) if isinstance(var, FunctionAddress) else get_var_arg(arg, var) for arg, var in zip(args, var_list)]
+            arg_code_list = [self.function_signature(var, False) if isinstance(var, FunctionAddress)
+                                else get_var_arg(arg, var) for arg, var in zip(args, var_list)]
             arg_code = ', '.join(arg_code_list)
 
+        if self._additional_args :
+            self._additional_args.pop()
```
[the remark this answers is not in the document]
EmilyBourne: `_additional_args` stores all objects which are treated as arguments in C but are not arguments in python. In other words `out` arguments (maybe if it could be renamed to something clearer like `_return_args` if there are no other objects that would require this treatment). We need to be aware of these arguments as they are stored in c pointers. If  `_additional_args` is not set then the following function:
```python
def f():
    return 1, np.ones(3)
```
would be printed as:
```c
int f(int64_t Out_0001, t_ndarray Out_0002)
{
    Out_0001 = 1;
    Out_0002 = array_create(1, (int64_t[]){3}, nd_double, false);
    array_fill((double)1.0, Out_0002);
    return 0;
}
```
instead of
```c
int f(int64_t *Out_0001, t_ndarray *Out_0002)
{
    (*Out_0001) = 1;
    (*Out_0002) = array_create(1, (int64_t[]){3}, nd_double, false);
    array_fill((double)1.0, (*Out_0002));
    return 0;
}
```

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -307,6 +310,77 @@ def get_static_declare_type(self, variable):
...
+        expr        : FunctionDef
+                     The function being wrapped
+        static_func_args : List of arguments
+                     Arguments compatible with the static function
+        results     : List of results
+                     Results of the wrapped function
+
+        Returns     : List of Basic Nodes
+        -----------
+        """
```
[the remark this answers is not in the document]
EmilyBourne: Fixed cf6879c2

pyccel/codegen/printing/ccode.py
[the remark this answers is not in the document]
yguclu: Thanks for the recap!

I am a bit confused by the fact that we call `pop()` when `_additional_args` is not an empty list. Does that mean that we expect it to contain one element only? (Which I suspect is another list)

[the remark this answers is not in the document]
EmilyBourne: Indeed `_additional_args` is a list of lists. This allows it to handle functions in functions correctly

pyccel/codegen/printing/ccode.py
```diff
@@ -1105,9 +1105,9 @@ def _cast_to(self, expr, dtype, precision):
     def _print_DottedVariable(self, expr):
         """convert dotted Variable to their C equivalent"""
         if self.stored_in_c_pointer(expr.lhs):
-            code = '{}->{}'.format(self._print(ObjectAddress(expr.lhs)), self._print(expr.name))
+            code = f'{self._print(ObjectAddress(expr.lhs))}->{self._print(expr.name)}'
         else:
-            code =  '{}.{}'.format(self._print(expr.lhs), self._print(expr.name))
+            code =  f'{self._print(expr.lhs)}.{self._print(expr.name)}'
```
EmilyBourne, line 1110: This seems less readable to me like this. Please use intermediate variables to make it clearer. E.g:
```python
lhs_code = self._print(expr.lhs)
name_code = self._print(expr.name)
code = f'{lhs_code}.{name_code}'
```

pyccel/ast/variable.py
```diff
@@ -940,4 +940,4 @@ def __str__(self):
         return str(self.lhs)+'.'+str(self.name)
 
     def __repr__(self):
-        return '{}({}, dtype={})'.format(type(self).__name__, repr(self.lhs)+'.'+str(self.name), repr(self.dtype))
+        return f'{type(self).__name__}({repr(self.lhs)+'.'+str(self.name)}, dtype={repr(self.dtype)})'
```
EmilyBourne, line 943: Beware. You have strings inside strings which is breaking your f-string into:
```
f'{type(self).__name__}({repr(self.lhs)+' \
.  \
'+str(self.name)}, dtype={repr(self.dtype)})'
```
You need to either vary the quotes used:
```
f"{type(self).__name__}({repr(self.lhs)+'.'+str(self.name)}, dtype={repr(self.dtype)})"
```
or use intermediate variables to make this clearer

pyccel/codegen/printing/ccode.py
EmilyBourne: Both lines please

pyccel/codegen/printing/ccode.py
```diff
@@ -1098,9 +1104,18 @@ def _cast_to(self, expr, dtype, precision):
...
     def _print_DottedVariable(self, expr):
         """convert dotted Variable to their C equivalent"""
+
+        lhs_code = self._print(expr.lhs)
+        name_code = self._print(expr.name)
+        if self.stored_in_c_pointer(expr.lhs):
+            code = f'{self._print(ObjectAddress(expr.lhs))}->{name_code}'
+        else:
+            code = f'{lhs_code}.{name_code}'
+            code =  f'{self._print(expr.lhs)}.{self._print(expr.name)}'
```
EmilyBourne, line 1114: I don't think you meant to keep both of theses lines

pyccel/codegen/printing/ccode.py
```diff
@@ -1098,9 +1104,16 @@ def _cast_to(self, expr, dtype, precision):
 
     def _print_DottedVariable(self, expr):
         """convert dotted Variable to their C equivalent"""
+
+        lhs_code = self._print(expr.lhs)
+        name_code = self._print(expr.name)
+        code = f'{lhs_code}.{name_code}'
+        if self.stored_in_c_pointer(expr.lhs):
+            code = f'{self._print(ObjectAddress(expr.lhs))}->{name_code}'
```
EmilyBourne, line 1112: Personally I prefer the version with an if/else where `self._print(expr.lhs)`  is only calculated if it is used. Is there a reason why you prefer this version?

MostafaMamouni: no, but your suggestion is better.

pyccel/parser/semantic.py
```diff
@@ -828,7 +840,7 @@ def _handle_function(self, expr, func, args, **settings):
                         expr, func.is_elemental)
             return new_expr
 
-    def _create_variable(self, name, dtype, rhs, d_lhs):
+    def _create_variable(self, name, dtype, rhs, d_lhs, arr_in_multirets=False):
         """
         Create a new variable. In most cases this is just a call to
         Variable.__init__
```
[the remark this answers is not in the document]
MostafaMamouni: fixed here [cf5bb29](https://github.com/pyccel/pyccel/pull/1056/commits/cf5bb29bc7251977c2611fb2058ad8b658522108)

tests/epyccel/test_epyccel_return_arrays.py
```diff
@@ -0,0 +1,283 @@
...
+    epyccel_func = epyccel(return_array, language=language)
+
+    f_bl_true_output = epyccel_func(True, True)
+    test_bool_true_output = return_array(True, True)
+
+    f_bl_false_output = epyccel_func(False, False)
+    test_bool_false_output = return_array(False, False)
+
+    assert np.allclose(f_bl_true_output, test_bool_true_output)
+    assert np.allclose(f_bl_false_output, test_bool_false_output)
```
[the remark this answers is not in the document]
MostafaMamouni: fixed here [8050276](https://github.com/pyccel/pyccel/pull/1056/commits/805027603dfbd481ef11f2e68aeed389108ac155)

EmilyBourne (approved):

yguclu (approved):

PR #1058, opened 2021-12-27
In C files, never import inside a function

### LINKED ISSUE #1057: OpenMP Github actions are failing for MacOS
**Describe the bug**
OpenMP Github actions are failing for MacOS with errors such as:
```
>           raise RuntimeError(err_msg)
E           RuntimeError: Failed to build module
E           In file included from /Users/runner/work/pyccel/pyccel/tests/__epyccel__/__pyccel__/mod_em2wulvx49y1_em2wulvx49y1.c:9:
E           /usr/local/include/omp.h:483:31: error: unexpected OpenMP directive '#pragma omp begin declare variant'
E               #pragma omp begin declare variant match(device={kind(host)})
E                                         ^
E           /usr/local/include/omp.h:484:51: error: function definition is not allowed here
E               static inline int omp_is_initial_device(void) { return 1; }
E                                                             ^
E           /usr/local/include/omp.h:485:29: error: unexpected OpenMP directive '#pragma omp end declare variant'
E               #pragma omp end declare variant
E                                       ^
E           /usr/local/include/omp.h:486:31: error: unexpected OpenMP directive '#pragma omp begin declare variant'
E               #pragma omp begin declare variant match(device={kind(nohost)})
E                                         ^
E           /usr/local/include/omp.h:487:51: error: function definition is not allowed here
E               static inline int omp_is_initial_device(void) { return 0; }
E                                                             ^
E           /usr/local/include/omp.h:488:29: error: unexpected OpenMP directive '#pragma omp end declare variant'
E               #pragma omp end declare variant
E                                       ^
E           6 errors generated.
```

It is not clear what is causing this, but it seems to be related to the recent update of the Mac OS from 10.15 to 11.6.2

Updating gcc does not fix the issue.

cc : @nhamidn

EmilyBourne:
Put all imports in a C file at the module level instead of allowing imports in functions. Imports in functions are an anti-pattern and can cause problems as any inline functions in the header file become functions in functions which is not supported in C. This change fixes #1057 .

### DIFF
--- a/.github/workflows/Github_pytest.yml
+++ b/.github/workflows/Github_pytest.yml
@@ -5,65 +5,65 @@ on:
     branches: [ master ]
 
 jobs:
-  Linux:
+  #Linux:
 
-    runs-on: ubuntu-latest
+  #  runs-on: ubuntu-latest
 
-    steps:
-      - uses: actions/checkout@v2
-      - name: Set up Python 3.6
-        uses: actions/setup-python@v2
-        with:
-          python-version: 3.6
-      - name: Install dependencies
-        uses: ./.github/actions/linux_install
-      - name: Install python dependencies
-        uses: ./.github/actions/pip_installation
-      - name: Coverage install
-        uses: ./.github/actions/coverage_install
-      - name: Fortran/C tests with pytest
-        uses: ./.github/actions/pytest_run
-      - name: Python tests with pytest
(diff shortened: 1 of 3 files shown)


nhamidn (approved):

yguclu (approved):
Well done @EmilyBourne!

PR #1060, opened 2021-12-28
Expose scalar module variables to Python (read-only) through getters

EmilyBourne:
Expose scalar module variables to Python (read-only) through getters (part 1 of #741 )

### DIFF
--- a/pyccel/codegen/printing/cwrappercode.py
+++ b/pyccel/codegen/printing/cwrappercode.py
@@ -954,7 +954,7 @@ class CWrapperCodePrinter(CCodePrinter):
         return 'pyobject'
 
     def _print_PyccelPyArrayObject(self, expr):
-        self._additional_imports.add(cwrapper_ndarray_import)
+        self.add_import(cwrapper_ndarray_import)
         return 'pyarrayobject'
 
     def _print_PyArg_ParseTupleNode(self, expr):
@@ -1204,7 +1204,7 @@ class CWrapperCodePrinter(CCodePrinter):
 
         # Print imports last to be sure that all additional_imports have been collected
         imports  = module_imports.copy()
-        imports += self._additional_imports
+        imports += self._additional_imports.values()
         imports  = ''.join(self._print(i) for i in imports)
 
         return ('#define PY_ARRAY_UNIQUE_SYMBOL CWRAPPER_ARRAY_API\n'
--- a/pyccel/codegen/printing/cwrappercode.py
+++ b/pyccel/codegen/printing/cwrappercode.py
@@ -1128,7 +1128,9 @@ class CWrapperCodePrinter(CCodePrinter):
         return CCodePrinter._print_FunctionDef(self, wrapper_func)
 
     def _print_Module(self, expr):
+        # The initialisation and deallocation shouldn't be exposed to python
         funcs_to_wrap = [f for f in expr.funcs if f not in (expr.init_func, expr.free_func)]
+
         if self._target_language == 'fortran':
             vars_to_wrap_decs = [Declare(v.dtype, v.clone(v.name.lower()), module_variable=True) \
                                     for v in expr.variables if not v.is_private and v.rank == 0]
--- a/pyccel/codegen/printing/cwrappercode.py
+++ b/pyccel/codegen/printing/cwrappercode.py
@@ -15,10 +15,11 @@ from pyccel.ast.literals  import Nil
 from pyccel.ast.core import Assign, AliasAssign, FunctionDef, FunctionAddress
 from pyccel.ast.core import If, IfSection, Return, FunctionCall, Deallocate
 from pyccel.ast.core import create_incremented_string, SeparatorComment
-from pyccel.ast.core import Import, Module
-from pyccel.ast.core import AugAssign
+from pyccel.ast.core import Import, Module, Declare
+from pyccel.ast.core import AugAssign, CodeBlock
 
-from pyccel.ast.operators import PyccelEq, PyccelNot, PyccelOr, PyccelAssociativeParenthesis, PyccelIsNot
+from pyccel.ast.operators import PyccelEq, PyccelNot, PyccelOr, PyccelAssociativeParenthesis
+from pyccel.ast.operators import PyccelIsNot, PyccelLt, PyccelUnarySub
 
 from pyccel.ast.datatypes import NativeInteger, NativeBool, NativeFloat, str_dtype
 from pyccel.ast.datatypes import datatype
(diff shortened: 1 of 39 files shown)


yguclu:
@EmilyBourne Why does this PR have a **blocked** label?

EmilyBourne:
> @EmilyBourne Why does this PR have a **blocked** label?

All the PRs were blocked by #1057

tests/epyccel/test_epyccel_modules.py
```diff
@@ -104,3 +104,20 @@ def test_module_5(language):
...
+
+    atts = ('g', 'R0', 'rMin', 'rMax', 'skip_centre',
+            'method', 'compl', 'tiny')
+    for att in atts:
+        mod_att = getattr(mod, att)
+        modnew_att = getattr(modnew, att)
+        assert mod_att == modnew_att
+        #TODO: Remove .item() when #735 is fixed
+        assert type(mod_att) is type(modnew_att) or \
+                type(mod_att.item()) is type(modnew_att)
```
yguclu, line 121: You can remove `.item()` now, as #1052 was just merged.

EmilyBourne: Done 2c27bcd

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -1040,21 +1128,28 @@ def _print_FunctionDef(self, expr):
         return CCodePrinter._print_FunctionDef(self, wrapper_func)
 
     def _print_Module(self, expr):
-        funcs_to_wrap = expr.funcs
+        funcs_to_wrap = [f for f in expr.funcs if f not in (expr.init_func, expr.free_func)]
```
yguclu, line 1131: Are `init_func` and `free_func` also contained in the `funcs` list? I thought this was not the case...

EmilyBourne: Yes, the `funcs` are defined from the namespace where `init_func` and `free_func` exist (see [here](https://github.com/pyccel/pyccel/blob/master/pyccel/parser/semantic.py#L1701-L1707))

yguclu: But is that what we really want? Shouldn't we keep `init_func` and `free_func` separate from the other functions, given that they are special? Or do you see the need to include them in the same `funcs` list?

yguclu: In other words, do we really need these two lines?
- [`self.insert_function(init_func)`](https://github.com/pyccel/pyccel/blob/b632a0dd4b648e6f30059fac4a80374fd2c4b997/pyccel/parser/semantic.py#L1676)
- [`self.insert_function(free_func)`](https://github.com/pyccel/pyccel/blob/b632a0dd4b648e6f30059fac4a80374fd2c4b997/pyccel/parser/semantic.py#L1699)

EmilyBourne: I'm not sure. In most cases they need treating in the same way as the other functions
- They still need to be wrapped in the bind c file
- They still need to be printed in the same way in the original file
- Their header still needs to be printed
The only place where they need to be treated differently is in the C API interface where we don't want to expose the call to these 2 functions to python.

My impression is that both implementations would work with a similar number of exceptions in the code, however I suspect that the treatment of `init_func` and `free_func` is more likely to be forgotten. If it is included by default then it can be removed when you see the generated code, if not then there is no reminder

yguclu: I see your point, and I agree that the current solution is probably the safest. 

Can you please add some comment lines in the code as a reminder to the developer? Comments are useful when we insert those functions into `funcs` in the semantic parser, and when we later ignore them in the C wrapper printer.

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -672,6 +678,82 @@ def get_PyBuildValue(self, used_names, variable):
...
+        mod_name    : str
+                      The name of the module variable
+        var         : Variable
+                      The module variable
+        collect_var : Variable
+                      A PyObject* variable to store temporaries
+        """
+        if var.rank != 0:
+            errors.report("Arrays cannot currently be exposed to Python",
+                    severity='warning', symbol=var)
```
yguclu, line 696: I think this warning should be more specific. I suggest "Global arrays (defined at the module level) cannot [...]"

By the way, can we add the exact file location to the warning?

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -672,6 +678,82 @@ def get_PyBuildValue(self, used_names, variable):
...
+
+        Parameters
+        ----------
+        mod_name    : str
+                      The name of the module variable
+        var         : Variable
+                      The module variable
+        collect_var : Variable
+                      A PyObject* variable to store temporaries
+        """
```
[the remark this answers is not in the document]
EmilyBourne: Fixed d2f38bd, and `Returns` in 37663f7

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: File location is a bit trickier, but I have managed to do it via the assign node: 745a440

yguclu (commented):
Can you please explain the changes in 4da6799?

EmilyBourne:
> Can you please explain the changes in [4da6799](https://github.com/pyccel/pyccel/commit/4da679964563defc3e29b206377f36fb6a1d657a)?

Prior to this commit, the warning we discussed above was not shown when translating to C as the translated file did not compile (the first compilation is done before the generation of the wrapper). The error was:
```
In file included from /home/emily/Code/test/tmp/__pyccel__/tmp.c:1:
/home/emily/Code/test/tmp/__pyccel__/tmp.h:8:18: warning: ‘x’ initialized and declared ‘extern’
    8 | extern t_ndarray x = {.shape = NULL};
      |                  ^
/home/emily/Code/test/tmp/__pyccel__/tmp.c:8:11: error: redefinition of ‘x’
    8 | t_ndarray x = {.shape = NULL};
      |           ^
In file included from /home/emily/Code/test/tmp/__pyccel__/tmp.c:1:
/home/emily/Code/test/tmp/__pyccel__/tmp.h:8:18: note: previous definition of ‘x’ was here
    8 | extern t_ndarray x = {.shape = NULL};
      |                  ^
```
We should have:
```
extern t_ndarray x;
```
in the header file, and:
```
extern t_ndarray x = {.shape = NULL};
```
in the C file

yguclu (approved):

Issue #330, opened 2020-06-02
 Pull request #322 breaks all symbolic test cases


saidctb:

EmilyBourne:
Shouldn't this be multiple issues? Are all the tests broken by the same problem?
Can you provide more details

saidctb:
pyccel gives me the same error for all of them `pyccel.parser.errors.PyccelSemanticError: Undefined variable`

PR #1068, opened 2022-01-26
Clean up scope: Part 1

### LINKED ISSUE #668: Investigate and either fix or remove the `@python` decorator
**Describe the bug**
In `pyccel.decorators` there is a `@python` decorator which leads to `pyccel.ast.core.PythonFunction` objects being created. This decorator has no doc string to explain what it does. Old documentation states:

> due to the limitations of Pyccel the python decorator gives you the freedom to write whatever Python code  you want  and will be executed dynamically during the code generation process
> ```
>@python
> def f(x):
>     x = [None]*5
>     x = 5
>     x = 5.
>     return x
> ```

We should investigate the utility of this decorator. If it is not useful then it should be removed. If it is useful then we should investigate whether it still works and fix it if it doesn't. A proper doc-string should also be added to the decorator to explain its purpose to users

EmilyBourne:
In order to facilitate future development on issues such as #227, it is important to have a clean Scope class that is easy to work with from anywhere in pyccel. This is the first PR cleaning up the scope. This PR does not change the way pyccel works at all. It just moves the Scope to its own file and puts in place some of the functions which will be useful later. This will make future PRs easier to follow.

**Commit Summary**
- Move `Scope` class to its own file
- Change storage system inside `Scope` to avoid forgetting objects inside the `imports`
- Add accessor functions to `Scope` to interact with `Scope` directly rather than via the `SemanticParser`
- Remove static functions
- Remove `@python` decorator. Fixes #668


pyccel/parser/scope.py
```diff
@@ -0,0 +1,325 @@
...
+    def decorators(self):
+        """Dictionary of Pyccel decorators which may be
+        applied to a function definition in this scope."""
+        return self._locals['decorators']
+
+    @property
+    def static_functions(self):
+        """ A dictionary of static functions defined in this scope
+        """
+        return self._locals['static_functions']
```
EmilyBourne, line 141: @saidctb Do you know what a static function is?

pyccel/parser/scope.py
```diff
@@ -0,0 +1,325 @@
...
+        """ A dictionary of all the scopes contained within the
+        current scope
+        """
+        return self._sons_scopes
+
+    @property
+    def symbolic_functions(self):
+        """ A dictionary of symbolic functions defined in this scope
+        """
+        return self._locals['symbolic_functions']
```
EmilyBourne, line 161: @saidctb Do you know what a symbolic function is?

pyccel/parser/scope.py
saidctb: it's probably not needed anymore

pyccel/parser/scope.py
EmilyBourne: Is this related to issue #330 ?

pyccel/parser/scope.py
EmilyBourne: Fixed: 80886bf  af002f6

pyccel/parser/scope.py
EmilyBourne: I see this seems to be for the `@sympy` and `@python` decorators as well as lambdas

saidctb: yes this is for symbolic functions, `@sympy` and lambdas, `@python` should have a different property

EmilyBourne: Ah, ok. It looks like I removed the `@python` version : a4766d2
It wasn't used anywhere in the code. Probably this falls under issue #668

pyccel/parser/semantic.py
```diff
@@ -649,6 +412,8 @@ def create_new_function_scope(self, name, decorators):
             name = DottedName(self._current_function, name)
         self._current_function = name
 
+        return child
```
yguclu, line 415: Please update the docstring with a `Returns` section

pyccel/parser/semantic.py
```diff
@@ -1656,7 +1413,7 @@ def _visit(self, expr, **settings):
                 return obj
 
         # Unknown object, we raise an error.
-        errors.report(PYCCEL_RESTRICTION_TODO, symbol=type(expr),
+        return errors.report(PYCCEL_RESTRICTION_TODO, symbol=type(expr),
```
yguclu, line 1416: Recurring question about the need for a `return`

pyccel/parser/scope.py
```diff
@@ -0,0 +1,319 @@
...
+
+    def find_in_scope(self, name, location = None):
```
yguclu, line 157: Given that the class name is `Scope`, I find it a bit strange to call this method `find_in_scope` instead of just `find`. And what do you mean by `location`?

pyccel/parser/scope.py
```diff
@@ -0,0 +1,319 @@
...
+        # Walk up the tree of Scope objects, until the root if needed
+        if self.parent_scope:
+            return self.parent_scope.find_in_scope(name, location)
+        else:
+            return None
+
+    def get_all_from_scope(self, location):
+        """ Find and return the specified object in the scope
+        If the object cannot be found then None is returned
+        """
```
yguclu, line 177: Here I would simply change the method name from `get_all_from_scope` to `find_all`

yguclu (changes requested):
Thanks @EmilyBourne, this looks very clean and well structured!

I have a few comments and questions

pyccel/parser/scope.py
```diff
@@ -0,0 +1,319 @@
...
+from pyccel.ast.datatypes import DataTypeFactory
+from pyccel.ast.headers import MacroFunction, MacroVariable
+from pyccel.ast.headers import FunctionHeader, ClassHeader, MethodHeader
+from pyccel.ast.variable import Variable, DottedName
+
+from pyccel.errors.errors import Errors
+
+errors = Errors()
+
+class Scope(object):
```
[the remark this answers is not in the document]
EmilyBourne: No, it has been modified. The first commit moves the class from `parser/base.py` to its own file. The rest changes it.
You can therefore see the changes by comparing from this commit:
https://github.com/pyccel/pyccel/pull/1068/files/47c2d9057fd04b56e1fb5163011a15a936801f71..39c3a3265a79d12ad22eb62004e3f0bc361251ac

pyccel/parser/syntactic.py
```diff
@@ -728,8 +728,6 @@ def fill_types(ls):
                     txt += ' results(' + ','.join(results) + ')'
 
                 header = hdr_parse(stmts=txt)
-                if name in self.namespace.static_functions:
-                    header = header.to_static()
```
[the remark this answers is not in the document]
EmilyBourne: See https://github.com/pyccel/pyccel/pull/1068#discussion_r794465575
We don't know what static functions are and they are not used so this is unnecessary

pyccel/parser/semantic.py
```diff
@@ -311,27 +276,16 @@ def check_for_variable(self, name):
...
+        else:
+            return self.namespace.find_in_scope(name, 'variables')
 
     def get_variable(self, name):
         """ Like 'check_for_variable', but raise Pyccel error if Variable is not found.
         """
         var = self.check_for_variable(name)
         if var is None:
-            errors.report(UNDEFINED_VARIABLE, symbol=name,
+            return errors.report(UNDEFINED_VARIABLE, symbol=name,
```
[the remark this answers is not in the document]
EmilyBourne: Pylint raises an error:
> Either all return statements in a function should return an expression, or none of them should.

This can be fixed by adding either a `return` or a `raise` before `errors.report`. `return` seems safer to me because `raise` interrupts the flow, which `errors.report` already does internally

pyccel/parser/semantic.py
EmilyBourne: See above

pyccel/codegen/printing/pycode.py
```diff
@@ -797,6 +796,8 @@ def _print_Module(self, expr):
         if free_func:
             self._ignore_funcs.append(free_func.name)
 
+        imports += ''.join(self._print(i) for i in self.get_additional_imports())
```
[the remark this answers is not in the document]
EmilyBourne: Previously it was not picking up anything imported in the `init_func`

pyccel/ast/basic.py
```diff
@@ -442,3 +442,19 @@ def copy_attributes(self, x):
...
+
+    def __init__(self, scope = None):
+        self._scope = scope
+        super().__init__()
+
+    @property
+    def scope(self):
+        """ Scope object containing all available objects in this part of the code
+        """
+        return self._scope
```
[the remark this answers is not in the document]
EmilyBourne: A `ScopedNode` is a node which has a scope. I.e. which has variables, functions, etc which are accessible within it. This is not particularly related to `OpenMP` variables. In Python a `While` would not be a `ScopedNode`, but it can be in other languages where temporaries can be created inside a loop which is why a scope is created for it.

This is just a way to associate the existing scopes stored in `Parser.namespace` with the ast node that describes the scope.

As written when in a function `FunctionDef.scope` (as `FunctionDef` is a `ScopedNode`) corresponds to the **local** scope, `FunctionDef.scope.parent_scope` is the **enclosing** scope. No distinction is made between the **global** scope and the outermost **enclosing** scope

pyccel/parser/semantic.py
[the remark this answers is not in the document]
yguclu: This seems to be related to [this issue](https://github.com/PyCQA/pylint/issues/1782) and [this PR](https://github.com/PyCQA/pylint/pull/4304). So we should add a `NoReturn` type hint somewhere...

pyccel/parser/scope.py
yguclu: Given that this function recursively searches in all enclosing scopes, the word `local` sounds a bit misleading to me. I think that `find` is just fine.

As for `location`, maybe `category` is what we mean?

EmilyBourne: Ah sorry, I seem to have confused this function with a different one which doesn't check enclosing scopes

pyccel/parser/scope.py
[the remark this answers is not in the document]
yguclu: I've found the right command: https://github.com/pyccel/pyccel/compare/47c2d9057fd04b56e1fb5163011a15a936801f71..devel-issue1067

pyccel/ast/basic.py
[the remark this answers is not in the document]
yguclu: Thanks for the explanation! How about the `scope = None` case?

pyccel/parser/semantic.py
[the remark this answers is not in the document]
yguclu: It should be sufficient to declare `def report(self, ...) -> typing.NoReturn`. Please see [this test file](https://github.com/hippo91/pylint/blob/main/tests/functional/i/inconsistent/inconsistent_returns_noreturn.py) in the pylint repository.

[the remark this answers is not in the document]
yguclu: Actually I am mistaken, here (and in most other situations) we use `errors.report` to raise an exception, but it actually returns `None` in these two cases:

1.  `blocker = False`
2. `(self.mode == 'user') and (severity == 'internal')`

Hence we cannot say that this function never returns. Maybe our error report functions should be improved in the future? I think that your current solution is the most sensible for now.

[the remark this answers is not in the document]
EmilyBourne: Yes indeed. This only raises an exception if `blocker == True` or `severity = 'fatal'`

pyccel/ast/basic.py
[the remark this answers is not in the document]
EmilyBourne: Currently (the next PR will change this), there are no scopes at the syntactic stage. So at this point `scope = None`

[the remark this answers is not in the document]
EmilyBourne: It is also useful for some corner cases such as `PyccelFunctionDef` which inherits from `FunctionDef` but doesn't need a `Scope`

pyccel/parser/base.py
```diff
@@ -27,14 +27,14 @@
...
+from pyccel.parser.scope     import Scope
 from pyccel.parser.utilities import is_valid_filename_pyh, is_valid_filename_py
 
-from pyccel.errors.errors import Errors
+from pyccel.errors.errors   import Errors
+from pyccel.errors.messages import PYCCEL_UNFOUND_IMPORTED_MODULE
 
-# TODO - remove import * and only import what we need
-#      - use OrderedDict whenever it is possible
+# TODO - use OrderedDict whenever it is possible
```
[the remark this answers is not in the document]
EmilyBourne: I agree. Should we also remove the current usage of `OrderedDict`? I would assume that `dict` would be faster

pyccel/parser/scope.py
EmilyBourne: Fixed [c4ead6c](https://github.com/pyccel/pyccel/pull/1068/commits/c4ead6c1765b206ee9e13c6a5a787277321f4e75)

pyccel/parser/scope.py
EmilyBourne: Fixed [c4ead6c](https://github.com/pyccel/pyccel/pull/1068/commits/c4ead6c1765b206ee9e13c6a5a787277321f4e75)

pyccel/ast/core.py
```diff
@@ -971,7 +975,7 @@ def local_vars(self):
         return self._local_vars
 
 
-class With(Basic):
+class With(ScopedNode):
```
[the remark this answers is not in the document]
EmilyBourne: Fixed [e871379](https://github.com/pyccel/pyccel/pull/1068/commits/e8713792be3ac4147dc0a7734452c8c51ecc6707)

pyccel/parser/semantic.py
EmilyBourne: Fixed [ec774d4](https://github.com/pyccel/pyccel/pull/1068/commits/ec774d4b4b774aef629d12d90ac4d0b3e8dfa097)

pyccel/parser/base.py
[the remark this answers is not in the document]
yguclu: We have compared them for usage in Psydac, and in our experiments they are equally fast. We decided to replace `OrderedDict` with `dict` because the latter can be dumped to a YAML file natively: https://github.com/pyccel/psydac/pull/157

If it doesn't cost much effort, I would also prefer using `dict` in Pyccel because it is a more basic class.

pyccel/parser/scope.py
yguclu: Thanks. Maybe we should explain in the docstring which values are admitted for `category`? There is a `keys` local variable in the `__init__`

EmilyBourne: Should I make this variable a class variable? I don't want to add everything in the docstring because I suspect we will forget to maintain it

yguclu: Good idea. Then you can just point to it in the docstring

pyccel/parser/base.py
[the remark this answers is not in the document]
EmilyBourne: Fixed 22e379a3

pyccel/parser/scope.py
EmilyBourne: Fixed 11a09e9c

yguclu (approved):

pyccel/parser/scope.py
EmilyBourne: @yguclu @saidctb we need to decide what we want to do with issue #668 in order to finish this PR. If we want to remove `@python` then it should be done completely. If we want to keep it then I need to put the `python_functions` category back

yguclu: I think we can remove it completely for now. We will reconsider it in the future.

yguclu (approved):

yguclu:
Should we not remove the `@python` decorator as well?

pyccel/parser/scope.py
EmilyBourne: If I remove this then I should also clean up the places where it is used which means for example removing all code related to Lambdas. Lambdas don't work currently so this may not be a problem. I just want to make sure before removing it

EmilyBourne: Fixed:
[5be3529](https://github.com/pyccel/pyccel/pull/1068/commits/5be3529e0e105b0ac7db11c13ac791354c1727e5)
[47ae7e1](https://github.com/pyccel/pyccel/pull/1068/commits/47ae7e14c7a982ca972dd335019f1a3974808199)

PR #1071, opened 2022-02-01
Allow tuple slicing

### LINKED ISSUE #1070: Cannot slice shape tuple without assigning it to variable
Pyccel does not allow taking the slice `a.shape[:2]`, where `a` is a Numpy array. Hence the following code does not work

```python
import numpy as np

a = np.ones((5, 6, 7, 8))

(s0, s1) = a.shape[:2]
(s2, s3) = a.shape[2:]

print(s0, s1, s2, s3)
```

and it raises the error

```bash/fortran/c
$ pyccel test_tuple_indexing.py 
  File "/usr/local/bin/pyccel", line 33, in <module>
  File "$PYCCEL/pyccel/pyccel/commands/console.py", line 216, in pyccel
    execute_pyccel(filename,
  File "$PYCCEL/pyccel/pyccel/codegen/pipeline.py", line 217, in execute_pyccel
    parser.annotate(**settings)
  File "$PYCCEL/pyccel/pyccel/parser/parser.py", line 164, in annotate
    parser = SemanticParser(self._syntax_parser,
  File "$PYCCEL/pyccel/pyccel/parser/semantic.py", line 200, in __init__
    self.annotate()
  File "$PYCCEL/pyccel/pyccel/parser/semantic.py", line 248, in annotate
    ast = self._visit(ast, **settings)
  File "$PYCCEL/pyccel/pyccel/parser/semantic.py", line 1625, in _visit
  File "$PYCCEL/pyccel/pyccel/parser/semantic.py", line 1637, in _visit_Module
    body = self._visit(expr.program).body
  File "$PYCCEL/pyccel/pyccel/parser/semantic.py", line 1625, in _visit
  File "$PYCCEL/pyccel/pyccel/parser/semantic.py", line 1800, in _visit_CodeBlock
    line = self._visit(b, **settings)
  File "$PYCCEL/pyccel/pyccel/parser/semantic.py", line 1625, in _visit
  File "$PYCCEL/pyccel/pyccel/parser/semantic.py", line 2369, in _visit_Assign
    rhs = self._visit(rhs, **settings)
  File "$PYCCEL/pyccel/pyccel/parser/semantic.py", line 1625, in _visit
  File "$PYCCEL/pyccel/pyccel/parser/semantic.py", line 1927, in _visit_IndexedElement
    return self._extract_indexed_from_var(var, args)
  File "$PYCCEL/pyccel/pyccel/parser/semantic.py", line 866, in _extract_indexed_from_var
    return var[indices[0]]
  File "$PYCCEL/pyccel/pyccel/ast/builtins.py", line 420, in __getitem__
    return self._args[i]
TypeError: tuple indices must be integers or slices, not Slice
```

The code works correctly if we add the assignment `s = a.shape` and then proceed to slice the tuple `s`. Interestingly, a similar error happens when we write `s3 = a.shape[-1]`, which raises the exception

```bash/fortran/c
TypeError: tuple indices must be integers or slices, not PyccelUnarySub
```

**Expected behavior**
It should be possible to slice directly `a.shape` without the need for a temporary tuple variable. 

**Language**
The error happens at the semantic stage, hence it does not depend on the target language.

EmilyBourne:
Improve `__getitem__` for `PythonTuple`. Fixes #1070 . Add tests showing the issue is fixed

### DIFF
--- a/pyccel/ast/builtins.py
+++ b/pyccel/ast/builtins.py
@@ -471,8 +471,8 @@ class PythonTuple(PyccelAstNode):
 
     @property
     def allows_negative_indexes(self):
-        """ Indicates whether negative values can be
-        used to index this Variable
+        """ Indicates whether variables used to
+        index this Variable can be negative
         """
         return False
 
--- a/pyccel/ast/builtins.py
+++ b/pyccel/ast/builtins.py
@@ -426,7 +426,7 @@ class PythonTuple(PyccelAstNode):
             if a is None:
                 return None
             elif isinstance(a, PyccelUnarySub):
-                return -int(a.args[0])
+                return -a.args[0].python_value
             else:
                 return a
 
--- a/pyccel/ast/builtins.py
+++ b/pyccel/ast/builtins.py
@@ -419,11 +419,11 @@ class PythonTuple(PyccelAstNode):
     def __getitem__(self,i):
         if isinstance(i, (int, LiteralInteger)):
             return self._args[i]
-        elif isinstance(i, Slice):
-            if not all(isinstance(s, (LiteralInteger, type(None))) for s in (i.start, i.step, i.stop)):
-                raise NotImplementedError("Can't index PythonTuple with non-integer types")
-            else:
-                return PythonTuple(*self._args[i.start:i.stop:i.step])
+        elif isinstance(i, Slice) and \
+                all(isinstance(s, (LiteralInteger, type(None))) for s in (i.start, i.step, i.stop)):
+            return PythonTuple(*self._args[i.start:i.stop:i.step])
+        elif self.is_homogeneous:
+            return IndexedElement(self, i)
         else:
             raise NotImplementedError("Can't index PythonTuple with type {}".format(type(i)))
 
@@ -455,6 +455,10 @@ class PythonTuple(PyccelAstNode):
         """ Arguments of the tuple
         """
         return self._args
+
+    @property
+    def allows_negative_indexes(self):
+        return False
 
 #==============================================================================
 class PythonLen(PyccelInternalFunction):
--- a/pyccel/ast/builtins.py
+++ b/pyccel/ast/builtins.py
@@ -428,7 +428,7 @@ class PythonTuple(PyccelAstNode):
             elif isinstance(a, PyccelUnarySub):
                 return -int(a.args[0])
             else:
-                return int(a)
+                return a
 
         if is_int(i):
             return self._args[to_int(i)]
--- a/pyccel/ast/builtins.py
+++ b/pyccel/ast/builtins.py
@@ -16,7 +16,7 @@ from .basic     import Basic, PyccelAstNode
 from .datatypes import (NativeInteger, NativeBool, NativeFloat,
                         NativeComplex, NativeString, str_dtype,
                         NativeGeneric, default_precision)
-from .internals import PyccelInternalFunction, max_precision
+from .internals import PyccelInternalFunction, max_precision, Slice
 from .literals  import LiteralInteger, LiteralFloat, LiteralComplex, Nil
 from .literals  import Literal, LiteralImaginaryUnit, get_default_literal_value
 from .literals  import LiteralString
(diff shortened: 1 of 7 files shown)


pyccel/ast/builtins.py
```diff
@@ -417,7 +417,15 @@ def __init__(self, *args):
             self._shape     = (LiteralInteger(len(args)), ) + args[0].shape
 
     def __getitem__(self,i):
-        return self._args[i]
+        if isinstance(i, (int, LiteralInteger)):
+            return self._args[i]
```
yguclu, line 421: Shouldn't we also consider the case of a negative literal index here? Something like `PyccelUnarySub(LiteralInteger(1))`.

Maybe we need to provide the `__index__` method to `PyccelUnarySub`, and then simply check with
```suggestion
        if hasattr(i, '__index__'):
            return self._args[i]
```

pyccel/ast/builtins.py
```diff
@@ -448,6 +456,13 @@ def args(self):
         """
         return self._args
 
+    @property
+    def allows_negative_indexes(self):
+        """ Indicates whether negative values can be
+        used to index this Variable
+        """
+        return False
```
yguclu, line 464: The docstring does not fully reflect the actual meaning of this property. `allows_negative_indexes` performs a runtime check of an index which cannot be evaluated at compile time. We may need to describe this properly throughout Pyccel...

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -746,7 +746,8 @@ def get_module_exec_function(self, expr, exec_func_name):
             vars_to_wrap = [v.clone(v.name.lower()) for v in orig_vars_to_wrap]
             for v,w in zip(orig_vars_to_wrap,vars_to_wrap):
                 assign = v.get_user_nodes(Assign)[0]
-                w.set_fst(assign.fst)
+                if assign.fst:
+                    w.set_fst(assign.fst)
```
yguclu, line 750: Is this a bug fix?

tests/epyccel/modules/tuples.py
```diff
@@ -356,3 +356,17 @@ def my_tup():
...
+
+def tuple_variable_index():
+    a = 1
+    b = (1,2,3)[a]
+    return b
+
+def tuple_variable_slice():
+    a = 1
+    b = (1,2,3)[:a]
+    return b[0]
```
yguclu, line 372: Can you please check the following cases?
- negative index
- inhomogeneous tuple

yguclu (changes requested):
Well done @EmilyBourne! I have a few of comments

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: I'm not sure. I don't think we ought to have any cases where an assign doesn't have an fst. But they are occasionally generated in the code printer. I couldn't work out where this missing fst arose so I wasn't able to fix it without this if. However with the if we will occasionally have less detailed error messages (related to the wrapper)

pyccel/ast/builtins.py
EmilyBourne: You are right that we also need to consider a negative literal index, but I don't think we can provide the `__index__` method to `PyccelUnarySub` as there is no guarantee that the argument of `PyccelUnarySub` is a literal

pyccel/codegen/printing/cwrappercode.py
yguclu: I see, can you please add a comment in the code?

pyccel/ast/builtins.py
yguclu: I think that `__index__` should return an `int` if it can, otherwise just `self` which will cause a `TypeError`.

yguclu: Sorry, my suggestion makes no sense because `PyccelUnarySub` should be passed to `IndexedVariable` in this case...

yguclu: Maybe we need an `eval` method? And then we can check if it returns an `int`?

tests/epyccel/modules/tuples.py
EmilyBourne: Fixed:
0d62648
d53259c

pyccel/ast/builtins.py
EmilyBourne: Perhaps, but in this case I think this ought to be in all operators and is a pull request of its own. For now I am handling this locally: bdecf75

tests/epyccel/modules/tuples.py
yguclu: Perfect, thanks!

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: Done 1434e04f

pyccel/ast/builtins.py
EmilyBourne: Fixed 8178ff88

yguclu: Thanks

pyccel/ast/builtins.py
yguclu: Sounds good

yguclu (approved):

PR #1076, opened 2022-03-06
Add singleton class to access stage

### LINKED ISSUE #1065: Move PyccelAstNode.stage
The variable `PyccelAstNode.stage` is used in many places, including many places unrelated to `PyccelAstNode`. There is no reason for this variable to be stored here. It should be moved to a new location, e.g. the `utilities` folder, and accessed using the `global` keyword

EmilyBourne:
Add singleton class `PyccelStage` to save the stage instead of `PyccelAstNode.stage` as the stage is not restricted to `PyccelAstNode`s. Fixes #1065

### DIFF
--- a/pyccel/ast/basic.py
+++ b/pyccel/ast/basic.py
@@ -403,7 +403,6 @@ class Basic:
 class PyccelAstNode(Basic):
     """Class from which all nodes containing objects inherit
     """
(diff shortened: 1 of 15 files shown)


yguclu (approved):
Well done!

PR #1078, opened 2022-03-09
Fix index insertion into indexed arrays

### LINKED ISSUE #1077: Transposition doesn't work when the transposed array is assigned to inner indexes of a larger array
**Describe the bug**
The generated code for a transposition operation using `np.transpose` or `ndarray.T` is wrong when assigning a transposed 2D inside a 4D array. 

**To Reproduce**
```python
import numpy as np

a = np.zeros((1, 4, 3, 1))
b = np.zeros((3, 4))

a[0, :, :, 0] = np.transpose(b)
```

```fortran
module test_transpose


      C_DOUBLE , b4 => C_BOOL
  implicit none

  real(f64), allocatable :: a(:,:,:,:)
  real(f64), allocatable :: b(:,:)
  logical(b4), private :: initialised = .False._b4

  contains

  subroutine test_transpose__init() 

    implicit none

    integer(i64) :: i_0001

    if (.not. initialised) then
      allocate(a(0:0_i64, 0:2_i64, 0:3_i64, 0:0_i64))
      a = 0.0_f64
      allocate(b(0:3_i64, 0:2_i64))
      b = 0.0_f64
      do i_0001 = 0_i64, 3_i64, 1_i64
        a(0_i64, i_0001, :, 0_i64) = b(i_0001, :)
      end do
      initialised = .True._b4
    end if

  end subroutine test_transpose__init

  subroutine test_transpose__free() 

    implicit none

    if (initialised) then
      if (allocated(a)) then
        deallocate(a)
      end if
      if (allocated(b)) then
        deallocate(b)
      end if
      initialised = .False._b4
    end if

  end subroutine test_transpose__free

end module test_transpose
```

**Expected behavior**
The line: `a(0_i64, i_0001, :, 0_i64) = b(i_0001, :)` should be `a(0_i64, :, i_001, 0_i64) = b(i_0001, :)`.

**Language**
Fortran

EmilyBourne:
Fix index insertion into indexed arrays when the slices are not on contiguous data. Fixes #1077

### DIFF
--- a/tests/epyccel/test_epyccel_transpose.py
+++ b/tests/epyccel/test_epyccel_transpose.py
@@ -166,3 +166,16 @@ def test_force_transpose(language):
 
     f2_epyc = epyccel(f2, language=language)
     assert f2( x2 ) == f2_epyc( x2 )
+
+def test_transpose_to_inner_indexes(language):
+    def f1(x : 'int[:,:]'):
+        from numpy import transpose, empty
+        n,m = x.shape
+        y = empty((1,m,n,1))
+        y[0,:,:,0] = transpose(x)
+        return y[0,0,0,0], y[0,1,0,0], y[0,0,1,0], y[0,-1,0,0], y[0,0,-1,0]
+
+    x1 = randint(50, size=(2,5))
+
+    f1_epyc = epyccel(f1, language=language)
+    assert f1( x1 ) == f1_epyc( x1 )
--- a/pyccel/ast/utilities.py
+++ b/pyccel/ast/utilities.py
@@ -279,8 +279,11 @@ def insert_index(expr, pos, index_var):
     elif isinstance(expr, IndexedElement):
         base = expr.base
         indices = list(expr.indices)
-        while -pos<=expr.base.rank and not isinstance(indices[pos], Slice):
-            pos -= 1
+        i = -1
+        while i>=pos and -i<=expr.base.rank:
+            if not isinstance(indices[i], Slice):
+                pos -= 1
+            i -= 1
         if -pos>expr.base.rank:
             return expr
 


tests/epyccel/test_epyccel_transpose.py
```diff
@@ -166,3 +166,16 @@ def f2(x : 'int[:,:,:]'):
...
     f2_epyc = epyccel(f2, language=language)
     assert f2( x2 ) == f2_epyc( x2 )
+
+def test_transpose_to_inner_indexes(language):
+    def f1(x : 'int[:,:]'):
+        from numpy import transpose, empty
+        n,m = x.shape
+        y = empty((1,m,n,1))
+        y[0,:,:,0] = transpose(x)
+        return y[0,0,0,0], y[0,1,0,0], y[0,0,1,0], y[0,-1,0,0], y[0,0,-1,0]
```
yguclu, line 176: I suggest that we pass the array `y` as argument to `f1`, and then we check all components of it with `np.array_equal()`. This makes it easier to write other tests (see below).

Can you please add a few tests to verify other combinations of indices? E.g.
```
y[:, :, 0, 0]
y[:, 1, 0, :]
y[1, :, 0, :]
```
And maybe a 3D slice of a 5D array...

yguclu (changes requested):
Well done @EmilyBourne! That was fast!!

By the way, I cannot find any reference to `numpy.transpose()` or `ndarray.T` in our new documentation. Can you?

tests/epyccel/test_epyccel_transpose.py
EmilyBourne: Fixed 6da07428

Documentation has also been added

yguclu (approved):

Issue #1079, opened 2022-03-10
system-wide install and permission problem


econwang:
**Describe the bug**
`pyccel` needs write permission in the package install path, with a system-wide installation, it will raise the permission denied problem.  

@EmilyBourne proposed to install `pyccel` locally (https://github.com/pyccel/pyccel/issues/942#issuecomment-902121769), it is inconvenient to deploy and maintain on the cluster with local installation. Please provide the fix or suggestion that can change the folder.

EmilyBourne:
The objects for which pyccel currently requires write permission in the package install path are (hopefully I haven't missed anything):
1. [Json files containing default information for different compilers](https://github.com/pyccel/pyccel/blob/master/tutorial/compiler.md#user-defined-compiler)
2. [Pickled header files](https://github.com/pyccel/pyccel/blob/master/tutorial/header-files.md#pickling-header-files) for internal libraries (e.g. openmp, lapack)

There are several options for how this could be handled differently with different advantages and disadvantages:

1. The files could be generated during the installation/updates
    - On a cluster this means that the default compiler file will be hard to find (should we also copy this file into the `__pyccel__` folder?)
    - This means developers not using the versioning system (i.e. those who installed via git) must remember to regenerate the files manually after they have made changes
    - In theory this approach could be used without removing the operations that require write permission. As the files are already generated the calls to the write functions should never be required. Therefore people using the versioning system will see these files updated during updates (but not after as the code will not change, triggering the recreation of these files), and people who installed via git will continue to see these files regenerated as is currently the case
2. The files could be generated into a `~/.pyccel` folder
    - A `~/.pyccel.lock` file will also pollute the HOME
    - It is unconventional to store pickled files in a different location to the original file. This could make the compilation harder
3. The files could be generated into the `__pyccel__` folder
    - This loses the advantages of pickling
    - This could be helpful for keeping track of the compiler used
4. Hybrid approach

Personally I think I would prefer option 1 for the pickled header files, and either option 2 or option 1 plus a copy to the `__pyccel__` folder for the json files.

@yguclu @saidctb what do you think?

@econwang Do you have an opinion?

EmilyBourne:
Option 1 seems to be surprisingly tricky. The setup via pip is not designed to allow the creation of files. I therefore suggest an additional option:

[...]

PR #1080, opened 2022-03-14
Remove unnecessary json generation

EmilyBourne:
Generating JSON files is an unnecessary step which makes pyccel hard to use when it is installed system-wide (see #1079). This PR removes the generation and access the dictionaries directly. The flag `--export-compile-info` is added to provide the user with a summary of the compiler flags and options in use


tutorial/compiler.md
```diff
@@ -46,4 +46,4 @@ In addition, for each accelerator (mpi/openmp/openacc/python) that you will use
 
 Python is considered to be an accelerator and must additionally specify shared\_suffix.
 
-The default compilers can provide examples compatible with your system once pyccel has been executed at least. These json files can be found in the folder pyccel/compilers/
+The default compilers can provide examples compatible with your system once pyccel has been executed at least. To export the json file describing your setup, use the `--export-compile-info` flag or the `export_compile_info` argument and provide a target file name.
```
yguclu, line 49: Can you please provide an example of how this feature should be used in practice?

For instance, this does not work:
```bash
$ pyccel --export-compile-info comp.json
error: the following arguments are required: N
```

EmilyBourne: I didn't change the command significantly so we still need to be pyccelising something. Ie:
```
pyccel --export-compile-info comp.json my_file.py
```
Now I think about it, this seems like a bad idea. I will change it so that it works with the command you describe above. Do you still think I should give an example if it is fixed?

EmilyBourne: Fixed 2a548fc2

yguclu: Thanks for fixing this, it is much better now. Should we maybe enforce the `.json` extension of the preference file? That is, add `.json` to the file name if the extension is missing, and raise an error if the file name has the wrong extension.

EmilyBourne: Fixed 18e37e13

yguclu:
> Windows tests are currently stuck as [netlib.org](http://netlib.org) and the [windows build of the lapack libraries](http://icl.cs.utk.edu/lapack-for-windows/) are currently offline

I am dropping the Windows tests from the merge requirements, so that we can merge this PR

yguclu (approved):

PR #1081, opened 2022-03-16
Create a pyccel-init script to assist with installation in a read-only system

### LINKED ISSUE #1079: system-wide install and permission problem

EmilyBourne:
Create a pyccel-init script which pickles all internal files. Fixes #1079 . Clean up installation scripts

**Commit Summary**
- Only output pickle files if the files were not read from pickled files
- Catch `AttributeError`s which are raised when the pickle file was created with an outdated version of pyccel
- Create a `pyccel-init` script which creates all necessary pickle files. This command can then be run with sudo privileges to ensure that pyccel never needs write access to its own folder
- Raise a warning when pyccel tries to write a pickle file to a folder where it doesn't have write access to warn the user to use `pyccel-init`
- Rewrite the setup script to use the new `setup.cfg` workflow and to create the pickle files as a post-install step

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,7 +267,7 @@ Any required Python packages will be installed automatically from PyPI.
 
 ### On a read-only system
 
-If the folder where pyccel is saved is read only, it may be necessary to run an additional command:
+If the folder where pyccel is saved is read only, it may be necessary to run an additional command after installation or updating:
 ```sh
 sudo pyccel-init
 ```
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ If you are eager to try Pyccel out, we recommend reading our [quick-start guide]
 -   [Installation](#Installation)
     -   [From PyPi](#From-PyPi)
     -   [From sources](#From-sources)
+    -   [On a read-only system](#On-a-read-only-system)
 
 -   [Additional packages](#Additional-packages)
 
@@ -263,6 +264,13 @@ for a system-wide installation.
 
 this will install a _python_ library **pyccel** and a _binary_ called **pyccel**.
 Any required Python packages will be installed automatically from PyPI.
+
+### On a read-only system
+
+If the folder where pyccel is saved is read only without sudo privileges an additional command should be run:
+```
+sudo pyccel-init
+```
 
 ## Additional packages
 
--- a/README.md
+++ b/README.md
@@ -268,7 +268,7 @@ Any required Python packages will be installed automatically from PyPI.
 ### On a read-only system
 
 If the folder where pyccel is saved is read only without sudo privileges an additional command should be run:
-```
+```sh
 sudo pyccel-init
 ```
 
--- a/README.md
+++ b/README.md
@@ -272,8 +272,8 @@ If the folder where pyccel is saved is read only, it may be necessary to run an 
 sudo pyccel-init
 ```
 
-This step is necessary in order to [pickle header files](.tutorial/header-files.md#Pickling-header-files).
-If this command is not run then pyccel will still run correctly but may be slower when using [OpenMP](.tutoral/openmp.md) or other supported external packages.
+This step is necessary in order to [pickle header files](./tutorial/header-files.md#Pickling-header-files).
+If this command is not run then pyccel will still run correctly but may be slower when using [OpenMP](./tutorial/openmp.md) or other supported external packages.
 A warning, reminding the user to execute this command, will be printed to the screen when pyccelizing files which rely on these packages if the pickling step has not been executed.
 
 ## Additional packages
--- a/README.md
+++ b/README.md
@@ -267,10 +267,14 @@ Any required Python packages will be installed automatically from PyPI.
 
 ### On a read-only system
 
-If the folder where pyccel is saved is read only without sudo privileges an additional command should be run:
+If the folder where pyccel is saved is read only, it may be necessary to run an additional command:
 ```sh
 sudo pyccel-init
 ```
+
+This step is necessary in order to [pickle header files](.tutorial/header-files.md#Pickling-header-files).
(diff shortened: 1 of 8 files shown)


EmilyBourne:
The pyccel-init script seems to work but on my system, executing it with `sudo pyccel-init` gives:
```
sudo: pyccel-init: command not found
```
If I help it locate the file then it gives:
```
  File "/home/emily/.local/bin/pyccel-init", line 33, in <module>
    sys.exit(load_entry_point('pyccel', 'console_scripts', 'pyccel-init')())
  File "/home/emily/.local/bin/pyccel-init", line 22, in importlib_load_entry_point
    for entry_point in distribution(dist_name).entry_points
  File "/usr/lib/python3.8/importlib/metadata.py", line 503, in distribution
    return Distribution.from_name(distribution_name)
  File "/usr/lib/python3.8/importlib/metadata.py", line 177, in from_name
    raise PackageNotFoundError(name)
importlib.metadata.PackageNotFoundError: pyccel
```
Presumably if pyccel had been installed with sudo I would not have this problem, but I don't know

yguclu:
If I install `pyccel` in editable mode and then call `pyccel-init`, I get the following error:
```bash
$ pyccel-init
  File "/path/to/virtualenv/bin/pyccel-init", line 33, in <module>
    sys.exit(load_entry_point('pyccel', 'console_scripts', 'pyccel-init')())
  File "/path/to/pyccel/pyccel/commands/pyccel_init.py", line 24, in pyccel_init
    parser.parse(verbose=False)
  File "/path/to/pyccel/pyccel/parser/parser.py", line 141, in parse
    parser             = SyntaxParser(self._filename, **self._kwargs)
  File "/path/to/pyccel/pyccel/parser/syntactic.py", line 133, in __init__
    self.load()
  File "/path/to/pyccel/pyccel/parser/base.py", line 444, in load
    hs, version, parser = pickle.load(f)
AttributeError: Can't get attribute 'ParserResult' on <module 'pyccel.ast.core' from '/path/to/pyccel/pyccel/ast/core.py'>
```
We should make sure that this command works in editable mode too.

EmilyBourne:
> 

Hmm, I don't have this problem. It arises when the code is changed so that the pickle file is no longer valid. In 9ed3e60d, I avoid this by catching `AttributeError`s, but I don't know if there are other errors that out-of-date pickle files can cause

yguclu:
> > 
> 
> Hmm, I don't have this problem. It arises when the code is changed so that the pickle file is no longer valid. In [9ed3e60](https://github.com/pyccel/pyccel/commit/9ed3e60d853d06a33e90cdb8d9b88d9e07b944fa), I avoid this by catching `AttributeError`s, but I don't know if there are other errors that out-of-date pickle files can cause

This has fixed the problem on my computer, thanks!

Do you think it would be easy to add a print at pickle files creation, with a message to the user and the path to these files?

EmilyBourne:
> Do you think it would be easy to add a print at pickle files creation, with a message to the user and the path to these files?

Yes this is simple to add. But I the pickle file is printed every time the file is used. I'm not sure we want lots of pickle messages whenever we pyccelize a file which uses openmp?

EmilyBourne:
> the pickle file is printed every time the file is used

This is a problem for issue #1079 . I have modified this so that the pickle file is only created if the file wasn't itself loaded from a pickle file. This means that printing a message when one is created now makes sense. See [78d4764](https://github.com/pyccel/pyccel/pull/1081/commits/78d476444bda8e239eb28bd9a599fe0d44417a2d)

yguclu:
> > the pickle file is printed every time the file is used
> 
> This is a problem for issue #1079 . I have modified this so that the pickle file is only created if the file wasn't itself loaded from a pickle file. This means that printing a message when one is created now makes sense. See [78d4764](https://github.com/pyccel/pyccel/pull/1081/commits/78d476444bda8e239eb28bd9a599fe0d44417a2d)

Excellent! I had not realized that the pickle file was created multiple times...

README.md
```diff
@@ -264,6 +265,13 @@ for a system-wide installation.
 this will install a _python_ library **pyccel** and a _binary_ called **pyccel**.
 Any required Python packages will be installed automatically from PyPI.
 
+### On a read-only system
+
+If the folder where pyccel is saved is read only without sudo privileges an additional command should be run:
+```sh
+sudo pyccel-init
+```
```
yguclu, line 273: Is this still true? Won't `setup.py` create and install the pickle files for us?

I suspect that this command is only needed if pyccel is updated.

yguclu (changes requested):
Well done!!

I only have one comment in the README file. Please expand the PR explanation.

README.md
EmilyBourne: Fixed

EmilyBourne: I believe that `setup.py` should install the pickle files for us. But I suspect it is impossible to be sure until we have tested by pulling from pypi

yguclu: I almost forgot, please update the version

EmilyBourne: Fixed [4a4d0ab](https://github.com/pyccel/pyccel/pull/1081/commits/4a4d0ab36d32a612647c0d52476ae0b68dea91b0)

yguclu:
> Well done!!
> 
> I only have one comment in the README file. Please expand the PR explanation.

@EmilyBourne Can you please explain in more details the changes made in this PR?

EmilyBourne:
> @EmilyBourne Can you please explain in more details the changes made in this PR?

Done

yguclu (approved):

PR #1088, opened 2022-04-09
Correct nested generators bug

### LINKED ISSUE #1087: Nested functional generators do not always work as expected
**Describe the bug**
Nested functional generators do not always work as expected

**To Reproduce**
```python
from numpy import ones

a=ones((5,5),float)

a8 = min(max(a[i,k] for i in range(5)) for k in range(5))
```

```bash/fortran/c
pyccel:
 |fatal [semantic]: tmp3.py| Cannot determine final dtype of 'max' call with arguments of different types (float(8), float(8)). Please cast arguments to the desired dtype (max(FunctionCallArgument('Dummy_0001'), FunctionCallArgument('a'['i','k'])))
```

EmilyBourne:
- Fix #1087 (nested generators bug) caused by mismatched order in 0d/1d arrays. See 5e75b2ddc7152c78a7e994da4f5453816ee14fe8
- Correct naming bug which lead to nested generators not being tested
- Fix python printing of nested generators

### DIFF
--- a/pyccel/codegen/printing/pycode.py
+++ b/pyccel/codegen/printing/pycode.py
@@ -12,6 +12,7 @@ from pyccel.ast.builtins   import PythonMin, PythonMax
 from pyccel.ast.core       import CodeBlock, Import, Assign, FunctionCall, For, AsName, FunctionAddress
 from pyccel.ast.core       import IfSection, FunctionDef, Module
 from pyccel.ast.datatypes  import default_precision
+from pyccel.ast.functionalexpr import FunctionalFor
 from pyccel.ast.literals   import LiteralTrue, LiteralString
 from pyccel.ast.literals   import LiteralInteger, LiteralFloat, LiteralComplex
 from pyccel.ast.numpyext   import Shape as NumpyShape, numpy_target_swap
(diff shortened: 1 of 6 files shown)


pyccel/codegen/printing/pycode.py
```diff
@@ -119,14 +120,21 @@ def _find_functional_expr_and_iterables(self, expr):
...
-                body = body.body
+                body = list(body.body)
+                while isinstance(body[0], FunctionalFor):
+                    func_for = body.pop(0)
+                    for b in body:
+                        b.substitute(func_for.lhs, func_for)
                 if len(body) > 1:
                     if any(not(isinstance(b, Assign) and b.lhs is dummy_var) for b in body[1:]):
                         raise NotImplementedError("Pyccel has introduced unnecessary statements which it cannot yet disambiguate in the python printer")
                 body = body[0]
```
yguclu, line 131: Can you give me an example where the body of a `FunctionalFor` instance is a `CodeBlock`?

EmilyBourne: This happens in nested generators (`GeneratorComprehension` inherits from `FunctionalFor`). The `CodeBlock` is created at the semantic stage to contain the loop and the initialisation (e.g. `=inf` before running a `FunctionalMin`, `=0` before running a  `FunctionalSum` etc)

yguclu (approved):
Well done @EmilyBourne!

PR #1090, opened 2022-04-18
Use temporary PyccelAdd to type check AugAssign

### LINKED ISSUE #1054: Type checking for AugAssign does not report errors
**Describe the bug**
When checking type compatibility some exceptions are in place for `AugAssign`s. This is done to allow expressions such as:
```python
a = 2.5
a += 3
```
as `AugAssign`s are handled in the same way as `Assign`s and `a = 3` would not be allowed. However these exceptions mean that the correct type check is not done

**To Reproduce**
```python
a = 2
a += 3.5

if __name__ == '__main__':
    print(a)
```

This code compiles and produces the following output:
```bash/fortran/c
5
```

**Expected behaviour**
The code above should raise an error. The simplest way to do this is to save `a+=3.5` as `a=a+3.5` however this may lead to less readability in languages that support `AugAssign`. That said, given that Fortran, our standard translation language, doesn't support `AugAssign` this may not be a big loss.

EmilyBourne:
Use temporary PyccelAdd to type check AugAssign. This fixes #1054


pyccel/parser/semantic.py
```diff
@@ -1134,22 +1134,31 @@ def _assign_lhs_variable(self, lhs, d_var, rhs, new_expressions, is_augassign, *
...
-                        severity='error', blocker=False)
+                    if is_augassign:
+                        tmp_result = PyccelAdd(var, rhs)
+                        result_dtype = str(tmp_result.dtype)
+                        result_precision = get_final_precision(tmp_result)
+                        raise_error = (str(var.dtype) != result_dtype or \
+                                get_final_precision(var) != result_precision)
+                    else:
+                        raise_error = True
+                    if raise_error:
```
yguclu, line 1147: An empty line between these two would greatly increase readability.

yguclu (changes requested):
Nice fix!

I only have one tiny request...

pyccel/parser/semantic.py
EmilyBourne: Fixed [e7e5c34](https://github.com/pyccel/pyccel/pull/1090/commits/e7e5c34cb523104191f8ef85bbbf27c31d58ab98)

yguclu (approved):

PR #1094, opened 2022-04-18
Expose array module variables to Python (read-only) - C printer

EmilyBourne:
Allow Python to read module-level arrays from modules translated to C (part 2 of #741). Use Linux line endings in `pyccel/stdlib/ndarrays/ndarrays.h`.


pyccel/ast/cwrapper.py
```diff
@@ -378,10 +378,18 @@ def C_to_Python(c_object):
...
-        cast_function = c_to_py_registry[(c_object.dtype, c_object.precision)]
-    except KeyError:
-        errors.report(PYCCEL_RESTRICTION_TODO, symbol=c_object.dtype,severity='fatal')
+    if c_object.rank != 0:
+        if c_object.order == 'C':
+            cast_function = 'c_ndarray_to_pyarray'
+        elif c_object.order == 'F':
+            cast_function = 'fortran_ndarray_to_pyarray'
+        else:
+            cast_function = 'ndarray_to_pyarray'
```
yguclu, line 387: Shouldn't we distinguish between C and Fortran ordering only when `c_object.rank > 1`?

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -1172,7 +1178,9 @@ def _print_Module(self, expr):
         interface_funcs = [f.name for i in expr.interfaces for f in i.functions]
         funcs = [*expr.interfaces, *(f for f in funcs_to_wrap if f.name not in interface_funcs)]
 
+        self._in_header = True
         decs = ''.join('extern '+self._print(d) for d in vars_to_wrap_decs)
+        self._in_header = False
```
yguclu, line 1183: Is this a bug fix?

pyccel/ast/cwrapper.py
EmilyBourne: In theory `c_object.order` should be equal to `None` if `c_object.rank == 1`. The function `ndarray_to_pyarray` adds the C and Fortran flags if the rank is 1. See [here](https://github.com/pyccel/pyccel/blob/ebourne_expose_array/pyccel/stdlib/cwrapper_ndarrays/cwrapper_ndarrays.c#L226-L232)

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: Yes and no, these variables need to be declared as though they were in a header. But in reality the only place where that changes something is when printing arrays. In this case we need `self._in_header` to print:
`ndarray_t a;` instead of `ndarray_t a = {.shape = NULL};`

yguclu (approved):

PR #1096, opened 2022-04-19
Fix OpenMP on systems installed via anaconda

### LINKED ISSUE #1095: OpenMP does not work when python is installed via anaconda
**Describe the bug**
When Python is installed via anaconda an extra libdir is added to the compile command, but there is no equivalent `-Wl,rpath` command. As a result an error is raised on execution

**To Reproduce**
```python

@types('int')
def get_num_threads(n):
    from pyccel.stdlib.internal.openmp import omp_set_num_threads, omp_get_num_threads, omp_get_thread_num
    omp_set_num_threads(n)
    #$ omp parallel
    print("hello from thread number:", omp_get_thread_num())
    result = omp_get_num_threads()
    #$ omp end parallel
    return result
    
if __name__ == '__main__':
    x = get_num_threads(4)
    print(x)
```

Upon execution the following error message is obtained:
```bash
error while loading shared libraries: libgfortran.so.4: cannot open shared object file: No such file or directory
```

The generated compiler command for the executable is:
```
gfortran -O3 -funroll-loops -fPIC -fopenmp -c tmp.f90 -o tmp.o

gfortran -O3 -funroll-loops -fPIC -fopenmp -L /home/ebourne/anaconda3/lib tmp.o prog_tmp.f90 -otmp -lm -lgomp
```

**Expected behavior**

**Language**
Please specify which language the python code is translated to (Fortran by default)

**Additional context**
cc: @imadki

EmilyBourne:
Add `-Wl,rpath` command for default libdirs. Fixes #1095


imadki (commented):
It's working now. Many thanks

imadki (approved):
It's working. Many thanks

pyccel/codegen/compiling/compilers.py
```diff
@@ -298,14 +298,15 @@ def compile_program(self, compile_obj, output_folder, verbose = False):
...
         if self._info['language'] == 'fortran':
             j_code = (self._info['module_output_flag'], output_folder)
         else:
             j_code = ()
 
         cmd = [exec_cmd, *flags, *includes, *libdirs_flags,
-                *m_code, compile_obj.source,
+                 *linker_libdirs_flags, *m_code, compile_obj.source,
                 '-o', compile_obj.program_target,
                 *libs_flags, *j_code]
```
yguclu, line 311: Are we not duplicating most of the flags contained in `libdirs_flags`?

EmilyBourne: We should be duplicating all of them. The idea is to do: `-L libdir -Wl,-rpath libdir`. The first command `-L libdir` tells the compiler where to look for libraries. The second command `-Wl,-rpath libdir` tells the linker to save this path in the executable. Without the first command the file will not compile and will raise errors like:
`undefined reference to omp_get_thread_num_ `
Without the second command, the file will compile, but it will not know where to look for the libraries (only a problem when linking shared libraries). It will therefore raise errors like:
`cannot open shared object file: No such file or directory`
on execution

yguclu: I see, thanks!

yguclu (approved):

PR #1107, opened 2022-05-04
Make pyccel compatible with anaconda where possible

EmilyBourne:
Make pyccel compatible with anaconda where possible. This is possible for linux and macos, but not for Windows due to DLL hell (anaconda's DLLs take precedence and DLLs have no way to specify the location of their dependencies so this cannot be avoided)

**Commit Summary**
- Add github actions for a conda installation
- Update the README to specify that anaconda is possible but not advised
- Modify compiler tool to ignore conda paths when searching for a compiler
- [BUGFIX] module dependencies should come after the target so their contents are linked correctly
- Change library flags to correctly locate python library on a variety of platforms (including ones configured with anaconda)
- Specify minimum version for setuptools
- [BUGFIX] Change maximum values for arguments passed to `lcm` to avoid overflows on windows
- [BUGFIX] Increase tolerance to prevent failures in gamma test
- [BUGFIX] Use `isclose` to compare floating point values
- Add `external` pytest mark to indicate tests which will fail to find the correct DLL when running with conda on Windows


EmilyBourne:
Tests are passing for linux. But Windows is suffering from DLL hell

imadki:
It's fine, it works for me

pyccel/compilers/default_compilers.py
```diff
@@ -185,26 +185,42 @@ def change_to_lib_flag(lib):
...
+        linker_flags = [change_to_lib_flag(l) for l in
+                        config_vars.get("LIBRARY","").split() + \
+                        config_vars.get("LDSHARED","").split()[1:]]
+        python_info['python']['libs'] = [l[2:] for l in linker_flags if l.startswith('-l')]
+        python_info['python']['libdirs'] = [l[2:] for l in linker_flags if l.startswith('-L')] + \
+                            config_vars.get("LIBPL","").split()+config_vars.get("LIBDIR","").split()
+    elif possible_static_lib != '':
+        python_info['python']['dependencies'] = (possible_static_lib,)
+    else:
+        python_info['python']['dependencies'] = (possible_shared_lib,)
```
yguclu, line 224: Can you please explain what you are doing here?

yguclu (changes requested):

pyccel/codegen/compiling/compilers.py
```diff
@@ -363,7 +389,7 @@ def compile_shared_library(self, compile_obj, output_folder, verbose = False, sh
         file_out = os.path.join(compile_obj.source_folder, sharedlib_modname+ext_suffix)
 
         cmd = [exec_cmd, *flags, *includes, *libdirs_flags, *linker_libdirs_flags,
-                *m_code, compile_obj.module_target,
+                compile_obj.module_target, *m_code,
```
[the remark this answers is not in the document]
EmilyBourne: Static libraries are saved in the `m_code` variable. Libraries need to be placed in order such that any symbols are defined after they are used. Otherwise some compilers will not collect the symbols that it doesn't think are needed. So any static libraries need to come after `compile_obj.module_target`

pyccel/compilers/default_compilers.py
EmilyBourne: On some systems (notably with anaconda), the library mentioned in the configs does not exist but the equivalent static library does. This section of code searches for any libraries which have the expected name any actually exists. The static library is preferred as it means we don't need a lib dir argument which stops anaconda causing problems (eg linking scipy's blas library or h5py's hdf5 library). I have added comments: [a4139c6](https://github.com/pyccel/pyccel/pull/1107/commits/a4139c6a3b01bddd720cea70dccf45626819163c)

pyccel/codegen/compiling/compilers.py
```diff
@@ -28,6 +28,21 @@
     mac_target = '{}.{}'.format(*mac_version_tuple[:2])
     os.environ['MACOSX_DEPLOYMENT_TARGET'] = mac_target
 
+path_sep = ';' if platform.system() == 'Windows' else ':'
```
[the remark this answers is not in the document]
EmilyBourne: You're right. Previously this was used in a few places, but I managed to reduce the dependencies and just forgot this. Fixed a7653c90

pyccel/compilers/default_compilers.py
yguclu: Thanks!

yguclu (approved):

PR #1115, opened 2022-05-18
Avoid roundoff errors when testing generators

yguclu:
Use `randint` instead of `rand`, in order to avoid the effect of roundoff errors on the final result

### DIFF
--- a/tests/epyccel/test_epyccel_generators.py
+++ b/tests/epyccel/test_epyccel_generators.py
@@ -139,7 +139,7 @@ def test_nested_generators1(language):
     def f(a : 'float[:,:,:,:]'):
         return sum(sum(sum(a[i,k,o,2] for i in range(5)) for k in range(5)) for o in range(5))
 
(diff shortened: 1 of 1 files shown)


EmilyBourne (approved):

EmilyBourne:
Shouldn't the result (a float) be compared with `np.isclose` rather than `==`?

yguclu:
> Shouldn't the result (a float) be compared with `np.isclose` rather than `==`?

I thought about that, but then one needs to provide reasonable relative and absolute tolerances. I think that using integer numbers (represented as floats) is appropriate for testing the generator expressions in this module...

EmilyBourne:
> > Shouldn't the result (a float) be compared with `np.isclose` rather than `==`?
> 
> I thought about that, but then one needs to provide reasonable relative and absolute tolerances. I think that using integer numbers (represented as floats) is appropriate for testing the generator expressions in this module...

Issue #929, opened 2021-08-04
PGI/Nvidia/Intel fortran compilers cause seg faults when functions have optional arguments


EmilyBourne:
**Describe the bug**
As of Fortran 2018, to not pass a value to an optional arguments in a bind_c function in fortran, the user can pass `NULL`. Pyccel makes use of this fact to avoid large numbers of if/else blocks in the bind_X.f90 file. This capability is built into recent GNU implementations, however it is not available for intel/PGI/nvidia. As a result the files compile but when the function is called from python a segmentation fault is returned.

To fix this we should change how arrays are passed in the cwrapper. This change will also be necessary to fix issue #1117 .
Currently the wrapper for the following function:
```python
def f(a: 'float[:]' = None, b: 'float' = None):
    if a is not None:
        print(a)
    if b is not None:
        print(b)
```
is:
```fortran
  subroutine bind_c_f(n0_a, a, b) bind(c)

    implicit none

    integer(i64), value :: n0_a
    real(f64), intent(in), optional :: a(0:n0_a - 1_i64)
    real(f64), intent(in), optional :: b

    call f(a, b)

  end subroutine bind_c_f
```
Instead it should be:
```fortran
  subroutine bind_c_f(n0_a, a_ptr, b_ptr) bind(c)

    implicit none

    integer(i64), value :: n0_a
    type(c_ptr), value :: a_ptr
    type(c_ptr), value :: b_ptr

    real(f64), pointer :: a(:)
    real(f64), pointer :: b

    if (c_associated(a_ptr)) then
      call c_f_pointer(a_ptr, a, [n0_a])
      if (c_associated(b_ptr)) then
        call c_f_pointer(b_ptr, b)
        call f(a, b)
      else
        call f(a)
      end if
    else
      if (c_associated(b_ptr)) then
        call c_f_pointer(b_ptr, b)
        call f(b = b)
      else
        call f()
      end if
    end if

  end subroutine bind_c_f
```

EmilyBourne:
Some of the classes created in PR #1118 will be useful for this PR. The missing section is essentially the function which takes a `FunctionDef` (f) and returns the necessary if blocks

yguclu:
Do you plan to generate different codes depending on the target compiler? The additional code blocks become increasingly ugly as the number of optional arguments grows.

EmilyBourne:
> Do you plan to generate different codes depending on the target compiler? The additional code blocks become increasingly ugly as the number of optional arguments grows.

[...]

PR #1118, opened 2022-05-26
Improve cwrapper to handle slices as arguments

### LINKED ISSUE #1117: Strides not implemented for Fortran preventing passing slices to functions
I have the pyccelized function

```python
def fun(x : 'int[:]'):
    for i in range(len(x)):
    	print(x[i], ' ')
    print()
```

in `file.py` which just prints the entries of the integer input array `x`.

Running the code 

```python
import numpy as np
from file import fun

A = np.ones((5, 5))
nz = np.nonzero(A)

print('Array passed to pyccelized function : ', nz[0])
print('-----------------------------------------------')

# pass array nz[0]
print(nz[0].flags)
fun(nz[0])
print('-----------------------------------------------')

# pass a copy of the array nz[0]
print(nz[0].copy().flags)
fun(nz[0].copy())
```

reveals that passing the output `nz[0]` of `np.nonzero(A)` to the pyccelized function `fun` gives a wrong order because `nz[0]` is neither F nor C contiguous. Using a copy `nz[0].copy()` instead (which is both F and C contiguous) fixes the problem.

Output:
```
Array passed to pyccelized function :  [0 0 0 0 0 1 1 1 1 1 2 2 2 2 2 3 3 3 3 3 4 4 4 4 4]
-----------------------------------------------
  C_CONTIGUOUS : False
  F_CONTIGUOUS : False
  OWNDATA : False
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False

0 0 0 1 0 2 0 3 0 4 1 0 1 1 1 2 1 3 1 4 2 0 2 1 2 
-----------------------------------------------
  C_CONTIGUOUS : True
  F_CONTIGUOUS : True
  OWNDATA : True
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False

0 0 0 0 0 1 1 1 1 1 2 2 2 2 2 3 3 3 3 3 4 4 4 4 4
```

**Language**
Fortran

EmilyBourne:
Improve cwrapper to handle slices as arguments to functions translated to Fortran. Fixes #1117 

**Commit Summary**
- Add slices to `BindCFunctionDef` arguments
- Allow `IndexedElement` as argument in cwrapper (to store slice information in the argument
- Add cwrapper to pyccel stages
- Add helper `PyccelCIntegerDiv` class to only be used in the printer to avoid unnecessary casting/floor calls
- Simplify conditions in Fortran's `_print_Declare`
- Make `_check_pyarray_order` more permissive to just check if the dimensions are ordered as expected
- Add `nd_stride` accessor function
- Add tests


EmilyBourne:
To be solved after #929

PR #1122, opened 2022-05-31
Devel issue975

soukainahikma:
Added case when modulo equals 0 when one of the numbers is negative.

### DIFF
--- a/pyccel/stdlib/math/pyc_math_c.h
+++ b/pyccel/stdlib/math/pyc_math_c.h
@@ -20,6 +20,6 @@ int64_t         pyc_factorial(int64_t n);
 int64_t         pyc_gcd (int64_t a, int64_t b);
 int64_t         pyc_lcm (int64_t a, int64_t b);
 double          pyc_radians(double degrees);
-double          pyc_degrees(double radians);
+double          pyc_degrees(double radians); 
 
 #endif
--- a/pyccel/stdlib/math/pyc_math_c.h
+++ b/pyccel/stdlib/math/pyc_math_c.h
@@ -13,8 +13,8 @@
 ** operands of modulo operator.
 */
 
-#define MOD_PYC(N, M) ((N < 0 ^ M < 0) ? (N % M) + M : (N % M))
-#define FMOD_PYC(N, M) ((N < 0 ^ M < 0) ? fmod(N, M) + M : fmod(N, M))
+#define MOD_PYC(N, M) ((N < 0 ^ M < 0) && (N % M ) != 0 ? (N % M) + M : (N % M))
+#define FMOD_PYC(N, M) ((N < 0 ^ M < 0) && (N % M ) != 0  ? fmod(N, M) + M : fmod(N, M))
 
 int64_t         pyc_factorial(int64_t n);
 int64_t         pyc_gcd (int64_t a, int64_t b);
--- a/pyccel/stdlib/math/pyc_math_c.h
+++ b/pyccel/stdlib/math/pyc_math_c.h
@@ -20,6 +20,6 @@ int64_t         pyc_factorial(int64_t n);
 int64_t         pyc_gcd (int64_t a, int64_t b);
 int64_t         pyc_lcm (int64_t a, int64_t b);
 double          pyc_radians(double degrees);
-double          pyc_degrees(double radians); 
+double          pyc_degrees(double radians);
 
 #endif


EmilyBourne:
Is this PR ready for its initial review or is it still a draft? Both statuses are indicated at the moment

soukainahikma:
While solving this issue I discovered that the printing of all the integers translates to a “printf” with format specifier “ld” which works perfectly with positive numbers but with negative numbers it overflows.

EmilyBourne:
> While solving this issue I discovered that the printing of all the integers translates to a “printf” with format specifier “ld” which works perfectly with positive numbers but with negative numbers it overflows.

That's worrying! I'm not sure what would cause that. Please can you open an issue with an example to show the problem

EmilyBourne (changes requested):
Please could you add a test/tests which were failing before this PR but which pass now. This helps us make sure that your fix works and ensures that we will notice if future developments accidentally break something

soukainahikma:
Thank you so much for the reminder, I will pay attention next time and use the flags.
I think I fixed it and it’s ready for another review, I also added the test needed for this issue.

soukainahikma:
Alright, I will take care of it ! 😅

EmilyBourne (approved):
Thank you for adding all the tests. It looks like it works well :smiley:

bauom:
When using a complex number we got error in python and pyccel but the pyccel error is at compile time. shouldn't pyccel raise the error?
@EmilyBourne what do you think.

```python
import numpy as np

def f():
    a = np.array([12, 23 ,2], dtype=np.complex128)
    b = a[0]%2

if __name__ == '__main__':
    f()
```
<details>
  <summary>python error</summary>
 
```bash
  File "mod.py", line 8, in <module>
    f()
  File "mod.py", line 5, in f
    b = a[0]%2
TypeError: ufunc 'remainder' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''

```
</details>

<details>
  <summary>pyccel error</summary>
 
```bash
  File "/home/bauom/.local/bin/pyccel", line 11, in <module>
    load_entry_point('pyccel', 'console_scripts', 'pyccel')()
  File "/home/bauom/.pyccel/pyccel/commands/console.py", line 239, in pyccel
    execute_pyccel(filename,
  File "/home/bauom/.pyccel/pyccel/codegen/pipeline.py", line 368, in execute_pyccel
    src_compiler.compile_module(compile_obj=mod_obj,
  File "/home/bauom/.pyccel/pyccel/codegen/compiling/compilers.py", line 279, in compile_module
  File "/home/bauom/.pyccel/pyccel/codegen/compiling/compilers.py", line 403, in run_command
    raise RuntimeError(err_msg)
In file included from /home/bauom/dev/sandbox/tests/__pyccel__/mod.c:7:
/home/bauom/dev/sandbox/tests/__pyccel__/mod.c: In function ‘f’:
/home/bauom/dev/sandbox/tests/__pyccel__/math/pyc_math_c.h:17:28: error: invalid operands to binary < (have ‘complex double’ and ‘int’)
   17 | #define FMOD_PYC(N, M) ((N < 0 ^ M < 0) && fmod(N, M) != 0  ? fmod(N, M) + M : fmod(N, M))
      |                            ^
/home/bauom/dev/sandbox/tests/__pyccel__/mod.c:18:9: note: in expansion of macro ‘FMOD_PYC’
   18 |     b = FMOD_PYC(GET_ELEMENT(a, nd_cdouble, (int64_t)0), 2.0);
      |         ^~~~~~~~


```
</details>

bauom (approved):

yguclu (changes requested):

pyccel/stdlib/math/pyc_math_c.h
```diff
@@ -13,8 +13,8 @@
 ** operands of modulo operator.
 */
 
-#define MOD_PYC(N, M) ((N < 0 ^ M < 0) ? (N % M) + M : (N % M))
-#define FMOD_PYC(N, M) ((N < 0 ^ M < 0) ? fmod(N, M) + M : fmod(N, M))
+#define MOD_PYC(N, M) ((N < 0 ^ M < 0) && (N % M ) != 0 ? (N % M) + M : (N % M))
+#define FMOD_PYC(N, M) ((N < 0 ^ M < 0) && fmod(N, M) != 0  ? fmod(N, M) + M : fmod(N, M))
```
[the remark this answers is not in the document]
soukainahikma: I agree, saving the result into a variable is more efficient, I will make the necessary changes.

pyccel/stdlib/math/pyc_math_c.h
```diff
@@ -13,8 +13,11 @@
 ** operands of modulo operator.
 */
 
-#define MOD_PYC(N, M) ((N < 0 ^ M < 0) && (N % M ) != 0 ? (N % M) + M : (N % M))
-#define FMOD_PYC(N, M) ((N < 0 ^ M < 0) && fmod(N, M) != 0  ? fmod(N, M) + M : fmod(N, M))
+#define MOD_PYC(N, M) \
+    ({long int RES_MOD = (N % M); ((N < 0 ^ M < 0) && (RES_MOD ) != 0) ? (RES_MOD) + M : (RES_MOD);})
```
yguclu, line 17: Thanks for the changes.

I think there is still a problem with the parentheses. One is missing, and several are useless:

```suggestion
#define MOD_PYC(N, M) \
    ({long int RES_MOD = N % M; ((N < 0 ^ M < 0) && RES_MOD != 0 ? RES_MOD + M : RES_MOD);})
```

yguclu (changes requested):

pyccel/stdlib/math/pyc_math_c.h
yguclu: I also find that it is more natural to test for equality, rather than testing for inequality. What do you think?

yguclu: I am surprised that all tests pass, given the parenthesis problem... Are you 100% sure that you are properly testing this function?

soukainahikma: All my tests are added to the file "test_epyccel_mod.py" and i think i added all the cases to test the functionality of this function

soukainahikma: I will check again this function, fix the parenthesis problem and also see if I can test for the equality first .

EmilyBourne: Does this need to be a macro? If it is sufficiently complicated to need additional variables would an inline function be more pertinent?

soukainahikma: I thought of it, but I wasn't sure if this was the right thing to do since we are going to add just one variable, using the inline function will make the code more readable, so I am thinking of changing the macro into an inline function.

bauom: > @EmilyBourne If you write it as a function, how do you make sure that the compiler inlines it?

we can use the ```inline``` function specifier for C

yguclu: So you want to declare it as `inline` within the header itself?

yguclu: If we declare it as `extern`, I don't think that it will be inlined by the linker, unless we pass other flags to it. But I am fine with having the inline function declaration in the header.

yguclu: @soukainahikma Given the discussion above, please change these macros into `inline` functions. At the same time you could do the same with the functions `pyc_radians` and `pyc_degrees` which are currently in the file `pyccel/stdlib/math/pyc_math_c.c`.

EmilyBourne: > > we can use the inline function specifier for C
>
> So you want to declare it as inline within the header itself?

That is what I was suggesting yes. This should make it easier to read the parentheses too

bauom: we can't have only declaration in the header because the specifier inline will make it only available in the header file

bauom: if we add this to the header file
```C
inline int ab(int a, int b)
{
        return a + b;
}
```
and called it from another even if the header is included there we will get `undefined reference`

bauom: after discussing with @EmilyBourne we can use the `inline` specifier together with the `extern` specifier to define a function in a header file.

@soukainahikma for this issue you can add the inlined functions in the header file instead of the Macros using the example below.

ex:
```C
//function definition in Header file
extern inline func(void)
{
//body
}
```


<details>
 <summary>C standard about inline specifier</summary>

```
A function specifier may appear more than once; the behavior is the same as if it
appeared only once.
A function declared with an inline function specifier is an inline function. Making a
function an inline function suggests that calls to the function be as fast as possible.
Any function with internal linkage can be an inline function. For a function with external
linkage, the following restrictions apply: If a function is declared with an inline
function specifier, then it shall also be defined in the same translation unit. If all of the
file scope declarations for a function in a translation unit include the inline function
specifier without extern, then the definition in that translation unit is an inline
definition. An inline definition does not provide an external definition for the function,
and does not forbid an external definition in another translation unit. An inline definition
provides an alternative to an external definition, which a translator may use to implement
any call to the function in the same translation unit. It is unspecified whether a call to the
function uses the inline definition or the external definition.
```

</details>

@yguclu the standard state that the inline specifier will not always inline the function but will always try to call the function the fastest way possible.

pyccel/stdlib/math/pyc_math_c.c
```diff
@@ -7,7 +7,7 @@
 #include <math.h>
 
 /*---------------------------------------------------------------------------*/
-int64_t        pyc_factorial(int64_t n)
+int64_t                     pyc_factorial(int64_t n)
```
EmilyBourne, line 10: Is there any reason to add this giant spacing?

soukainahikma: I just wanted all the functions to be inlined vertically, to make the code readable after adding the `inline` keyword to some functions.

yguclu (approved):

Issue #338, opened 2020-06-08
Strings do not work


EmilyBourne:
Pyccel complains when trying to use strings as they do not have a precision

yguclu:
Does it complain because `precision=None`, or because the `precision` attribute is missing?

EmilyBourne:
It complains because `precision=None`:
```
File "/home/emily/.local/lib/python3.6/site-packages/pyccel/ast/core.py", line 2267, in __init__
    raise TypeError('precision must be an integer.')
```

EmilyBourne:
This can be fixed in 2 lines in PR #336 once PR #334 is merged

yguclu:
See Epic #459

PR #1125, opened 2022-06-08
Basic implementation of assert for C and Fortran

Pinkyboi:
This PR aim to add a basic assert implementation for pyccel:
- For the C implementation we will be using `assert.h` header.
- For the Fortran implementation we will be using an if block followed by the keyword `stop`.


EmilyBourne:
Hi, Thanks for opening your first Pull Request. Please make sure you read the developer documentation about how the [Review](https://github.com/pyccel/pyccel/blob/master/developer_docs/review_process.md) is organised in pyccel. This will tell you which labels to add and what to look out for to be sure that your PR has the best chance of passing the review stage quickly.

EmilyBourne (changes requested):
Please add a test. I would recommend doing a test in the tests/pyccel folder where you can run a program and check whether or not it has exit_code==0 (it should be 0 if everything is ok, and something else if the assert fails)

ohachim (commented):
I've tried assert with an expression comparing two strings:
```python
a = "ohachim"
b = "aihya"
assert a == b, "test"
```
we get: ```|fatal [semantic]: assertTest.py [2,8]| Pyccel has encountered syntax that has not been implemented yet. Please create an issue at https://github.com/pyccel/pyccel/issues and provide a small example of your problem. Do not forget to specify your target language
Type of Object : Nil cannot be infered```
The assertion seems to only be working for integral types right now, but maybe that's what's required for now?

EmilyBourne:
> I've tried assert with an expression comparing two strings:
> 
> ```python
> a = "ohachim"
> b = "aihya"
> assert a == b, "test"
> ```
> 
> we get: ` |fatal [codegen]: assertTest.py| Pyccel has encountered syntax that has not been implemented yet. Please create an issue at https://github.com/pyccel/pyccel/issues and provide a small example of your problem. Do not forget to specify your target language (string[kind = 0])` The assertion seems to only be working for integral types right now, but maybe that's what's required for now?

As you can see in the following issues:
- #338 
- #459 
- #605 

pyccel only has very minimal support for strings currently. In c it has even less support, only handling print statements. Given the error message you are describing, I'm guessing you were translating to c?

You can see the list of types fully supported by pyccel here: https://github.com/pyccel/pyccel/blob/master/tutorial/quickstart.md#type-annotations

EmilyBourne:
Please add a description to your PR describing the changes you made. This makes it easier for others to review and becomes the commit message when your PR is merged to master. You can look at the other open PRs for examples

EmilyBourne:
> I've got an issue when trying to use a module shared object in another python file:
> 
> * The module `test_module.py` python code:
> 
> ```python
> 
> @types('int','int')
> def func(a, b):
>     assert (b or a) and a == b, "this is a message"
>     print("-------------------------------------------")
> ```
> 
> * The python code in `main.py`:
> 
> ```python
>   from test_module import func
>   
>   func(1, 2)
> ```
> 
> When i try to use the compiled shared object of test_module this what happens:
> 
> Screen Shot 2022-06-30 at 12 57 23

This may be expected. An assert will kill the execution of the program which is then seen as a seg fault or a crash in python. Are you worried because you don't see your error message? What does the generated code look like?

> And when i try to run pyccel on `main.py` i get the following error:
> 
> Screen Shot 2022-06-30 at 13 00 06

This usually happens when you compile `main.py` without having first compiled `test_module.py`

EmilyBourne:
> Yes ! i just noticed that the `test_module.mod` wasn't generated, which is odd, `test_module.mod` isn't generated only when specifying `C` as the language to use, it works perfectly when using `Fortran`.

`.mod` is specific to Fortran. It is normal that this isn't generated when specifying `C`. It is generated by the compiler and is roughly equivalent to a `.h` file in C.

By default pyccel always generates pure fortran modules. The generated files are therefore not automatically compatible between languages. You should generate/compile your module and test in the same language

EmilyBourne:
> It seems like the segfault occurs even when the assert shouldn't be triggered.
> 
> * For the following code :
> 
> ```python
> from test_module import func
> 
> if __name__ == "__main__":
>     func(12, 12)
> ```
> 
> * The C code generated for the `test_module` is:
> 
> ```c
> #include "test_module.h"
> #include <stdlib.h>
> #include <stdint.h>
> #include "assert.h"
> #include <stdio.h>
> 
> 
> /*........................................*/
> void func(int64_t a, int64_t b)
> {
>     assert((b || a) && a == b);
>     printf("%s\n", "-------------------------------------------");
> }
> /*........................................*/
> ```
> 
> We again get the same Segfault as previously stated :
> 
> Screen Shot 2022-06-30 at 14 09 05

I can't reproduce this problem. Are you sure you saved your file properly?

Pinkyboi:
> .mod is specific to Fortran. It is normal that this isn't generated when specifying C. It is generated by the compiler and is roughly equivalent to a .h file in C.

>By default pyccel always generates pure fortran modules. The generated files are therefore not automatically compatible between languages. You should generate/compile your module and test in the same language

I see, it my bad then, i forgot that by default pyccel uses `Fortran`, its working in that case, thanks !

tests/pyccel/scripts/valid_assert_test.py
```diff
@@ -0,0 +1,11 @@
...
+
+if __name__ == "__main__":
+    assert True
+    a = 0
+    b = a
+    assert a == b
+    b = 1
+    assert a != b
+    assert a <= b
+    assert b >= a
```
EmilyBourne, line 11: Tests are failing due to the name of this file. See the [pytest documentation](https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#conventions-for-python-test-discovery):

> pytest implements the following standard test discovery:
> - search for test_*.py or *_test.py files, imported by their [test package name](https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#test-package-name).

If you rename this file so pytest doesn't think it contains a test for it to run then you should avoid this problem

Pinkyboi: Oh i see, i got me so confused as to why pytest tries to import that file.

EmilyBourne:
I removed the `needs_initial_review` label and marked this PR as a draft as you seemed to still be working on it. Once you are happy with it, don't forget to remove the draft status and put back the label

aihya (commented):
```python
if __name__ == "__main__":
    assert 1 == None, 'Zlaygaaa'
```
This test causes an issue:
```
  File "/home/pyccel/pyccel/ast/operators.py", line 437, in <listcomp>
    strs = [a for a in args if a.dtype is NativeString()]
AttributeError: 'Nil' object has no attribute 'dtype'
```

Pinkyboi:
> ```python
> if __name__ == "__main__":
>     assert 1 == None, 'Zlaygaaa'
> ```
> 
> This test causes an issue:
> 
> ```
>   File "/home/pyccel/pyccel/ast/operators.py", line 437, in <listcomp>
>     strs = [a for a in args if a.dtype is NativeString()]
> AttributeError: 'Nil' object has no attribute 'dtype'
> ```
I'm not sure if the problem comes from the assert, when testing with:

```python
if __name__ == "__main__":
    if 1 != None:
        print('Zlaygaaa')
```
I get the same error message that you mentioned.

I would assume that this is an entirely separated issue in `operators.py` that was, or has to be reported.

EmilyBourne:
> > ```python
> > if __name__ == "__main__":
> >     assert 1 == None, 'Zlaygaaa'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This test causes an issue:
> > ```
> >   File "/home/pyccel/pyccel/ast/operators.py", line 437, in <listcomp>
> >     strs = [a for a in args if a.dtype is NativeString()]
> > AttributeError: 'Nil' object has no attribute 'dtype'
> > ```
> 
> I'm not sure if the problem comes from the assert, when testing with:
> 
> ```python
> if __name__ == "__main__":
>     if 1 != None:
>         print('Zlaygaaa')
> ```
> 
> I get the same error message that you mentioned.
> 
> I would assume that this is an entirely separated issue in `operators.py` that was, or has to be reported.

I think if you run pylint on this code it will complain. Usually you should use `is` when comparing to `None` (see https://stackoverflow.com/a/14247383/5957195). Pyccel often assumes that you have clean code (ie you respect a lot of pylint's rules). I think this comes from there, but ideally it should probably be fixed.

aihya:
> 



> > > ```python
> > > if __name__ == "__main__":
> > >     assert 1 == None, 'Zlaygaaa'
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This test causes an issue:
> > > ```
> > >   File "/home/pyccel/pyccel/ast/operators.py", line 437, in <listcomp>
> > >     strs = [a for a in args if a.dtype is NativeString()]
> > > AttributeError: 'Nil' object has no attribute 'dtype'
> > > ```
> > 
> > 
> > I'm not sure if the problem comes from the assert, when testing with:
> > ```python
> > if __name__ == "__main__":
> >     if 1 != None:
> >         print('Zlaygaaa')
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I get the same error message that you mentioned.
> > I would assume that this is an entirely separated issue in `operators.py` that was, or has to be reported.
> 
> I think if you run pylint on this code it will complain. Usually you should use `is` when comparing to `None` (see https://stackoverflow.com/a/14247383/5957195). Pyccel often assumes that you have clean code (ie you respect a lot of pylint's rules). I think this comes from there, but ideally it should probably be fixed.

I just opened a new issue for case: https://github.com/pyccel/pyccel/issues/1152

aihya (approved):
Looks good to me

tests/pyccel/test_pyccel.py
```diff
@@ -849,6 +849,44 @@ def test_module_init( language ):
...
+    lang_out = get_lang_exit_value(output_test_file, language)
+    pyth_out = get_lang_exit_value(test_file, "python")
+    assert (not lang_out and not pyth_out) or (lang_out and pyth_out)
+
+
+def test_assert(language ):
+    pyccel_test_program_exit(language, "scripts/asserts/valid_assert.py")
+    pyccel_test_program_exit(language, "scripts/asserts/unvalid_assert1.py")
+    pyccel_test_program_exit(language, "scripts/asserts/unvalid_assert2.py")
+    pyccel_test_program_exit(language, "scripts/asserts/unvalid_assert3.py")
```
EmilyBourne, line 888: I suggest using `pytest.mark.parametrize` to make this into 3 tests. See `test_hope_benchmarks` for an example

EmilyBourne (changes requested):

pyccel/ast/core.py
```diff
@@ -3827,13 +3827,11 @@ class Assert(Basic):
     _attribute_nodes = ('_test',)
     #TODO add type check in the semantic stage
     def __init__(self, test):
-        #if not isinstance(test, (bool, Relational, sp_Boolean)):
-        #    raise TypeError('test %s is of type %s, but must be a Relational, Boolean, or a built-in bool.'
-        #                     % (test, type(test)))
-
+        if pyccel_stage == 'semantic':
+            if test.dtype is not NativeBool():
+                test = PythonBool(test)
```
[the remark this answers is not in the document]
Pinkyboi: Makes sense, i added it in db4a1f2221142d3f1669fd89d9f201f46e73d3cb :
https://github.com/pyccel/pyccel/blob/db4a1f2221142d3f1669fd89d9f201f46e73d3cb/pyccel/ast/core.py#L3830-L3832

tests/pyccel/test_pyccel.py
```diff
@@ -849,6 +849,44 @@ def test_module_init( language ):
...
 #------------------------------------------------------------------------------
+def get_lang_exit_value(abs_path, language, cwd=None):
+    abs_path = get_exe(abs_path, language)
+    if language == "python":
+        if cwd is None:
+            p = subprocess.Popen([sys.executable , "%s" % abs_path], stdout=subprocess.PIPE, universal_newlines=True)
+        else:
+            p = subprocess.Popen([sys.executable , "%s" % abs_path], stdout=subprocess.PIPE, universal_newlines=True, cwd=cwd)
+    else:
+        p = subprocess.Popen(["%s" % abs_path], stdout=subprocess.PIPE, universal_newlines=True)
```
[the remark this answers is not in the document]
Pinkyboi: Indeed we don't need those options, i removed them in 0cfbe851165b9611b28ba841d4eaee3757e64ce5 as well.
https://github.com/pyccel/pyccel/blob/0cfbe851165b9611b28ba841d4eaee3757e64ce5/tests/pyccel/test_pyccel.py#L856-L860

tests/pyccel/test_pyccel.py
Pinkyboi: Modified in 972b8a5d76c19b749bba3bd48cf7e75bc7f9afb3.
https://github.com/pyccel/pyccel/blob/972b8a5d76c19b749bba3bd48cf7e75bc7f9afb3/tests/pyccel/test_pyccel.py#L883-L890
is this better ?

tests/pyccel/test_pyccel.py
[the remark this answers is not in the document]
Pinkyboi: Yep, this one was basically a copy-paste from previous code so i left it as it is.

I removed the `%s` in  0cfbe851165b9611b28ba841d4eaee3757e64ce5.

pyccel/parser/semantic.py
```diff
@@ -2846,7 +2847,6 @@ def _visit_While(self, expr, **settings):
 
     def _visit_IfSection(self, expr, **settings):
         condition = expr.condition
-
```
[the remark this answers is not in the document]
Pinkyboi: I have nothing against spaces really 😅, those were unintentional, ill try to be more careful.
Fixed in db4a1f2221142d3f1669fd89d9f201f46e73d3cb as well.

pyccel/ast/core.py
```diff
@@ -3827,13 +3827,11 @@ class Assert(Basic):
...
         self._test = test
         super().__init__()
-
     @property
```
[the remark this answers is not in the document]
Pinkyboi: Added it back in db4a1f2221142d3f1669fd89d9f201f46e73d3cb.

pyccel/parser/semantic.py
```diff
@@ -3753,9 +3757,6 @@ def _visit_NumpyNonZero(self, func_call, **settings):
             arg = self._visit(new_symbol)
         return NumpyWhere(arg)
 
-#==============================================================================
-
-
 if __name__ == '__main__':
```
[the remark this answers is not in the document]
Pinkyboi: Added back in db4a1f2221142d3f1669fd89d9f201f46e73d3cb.

tests/pyccel/test_pyccel.py
EmilyBourne: Looks good. Is there any need for `test_assert` and `pyccel_test_program_exit` to be 2 files now? Do you think `pyccel_test_program_exit` could be useful for testing things other than asserts?

Pinkyboi: I combined `test_assert` and `pyccel_test_program_exit`  in b62266bfdc397d3fb1d330066c543c61eacb1a5a.

EmilyBourne (approved):

pyccel/ast/core.py
```diff
@@ -3827,10 +3827,9 @@ class Assert(Basic):
     _attribute_nodes = ('_test',)
     #TODO add type check in the semantic stage
```
yguclu, line 3828: Is this TODO comment still relevant?

yguclu: @Pinkyboi @EmilyBourne

EmilyBourne: It is no longer relevant as the argument is now wrapped in a `PythonBool` if it has the wrong type (a few lines after this comment)

yguclu: Then I am going to remove it, and merge this PR.

yguclu (approved):

PR #1127, opened 2022-06-08
Fix import of two files that import the same file

EmilyBourne:
The `d_parsers` may be updated during the parse stage however this information was not previously taken into account leading to the same file being parsed multiple times and sometimes overwritten.

### DIFF
--- a/pyccel/parser/parser.py
+++ b/pyccel/parser/parser.py
@@ -209,9 +209,11 @@ class Parser(object):
                 print ('>>> treating :: {}'.format(source))
 
             # get the absolute path corresponding to source
-
-            filename = get_filename_from_import(source, self._input_folder)
-            q = Parser(filename)
+            if source in d_parsers:
+                q = d_parsers[source]
+            else:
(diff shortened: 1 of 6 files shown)


saidctb (approved):

yguclu (approved):
Thanks @EmilyBourne for the bug fix!

@saidctb Do you need a new version of Pyccel on PyPI?

saidctb:
> Thanks @EmilyBourne for the bug fix!
> 
> @saidctb Do you need a new version of Pyccel on PyPI?

This bug was discovered in Struphy, but we found a way to make it work without fixing the bug
So we probably don't need a new version

yguclu:
> > Thanks @EmilyBourne for the bug fix!
> > @saidctb Do you need a new version of Pyccel on PyPI?
> 
> This bug was discovered in Struphy, but we found a way to make it work without fixing the bug. So we probably don't need a new version

I see. If this bug was effecting user code, it is probably a good idea to upload a new version. I am going to update the patch number.

PR #1132, opened 2022-06-16
Fix list comprehensions with anonymous iterators

### LINKED ISSUE #1131: List comprehensions with anonymous iterators don't work
**Describe the bug**
I just discovered pyccel and was testing it against a toy light switch problem where 100 (x) switches are toggled by 100 (x) people, where the first person toggles every first switch, the 2nd person toggles every second switch, 3rd person every 3rd, etc. After all 100 people have toggled the switches, how many are ON?

**To Reproduce**

```python
def switches(count: int) -> int:
    """
    >>> switches(100)
    10
    >>> switches(1000)
    31
    >>> switches(10000)
    100
    >>> switches(100000)
    316
    >>> switches(1000000)
    1000
    """
    switches = [False for _ in range(count)]
    switch = 0

    while switch < count:

        person = switch + 1
        sw = switch
        while sw < count:
            switches[sw] = not switches[sw]
            sw += person
        switch = person

    on = 0
    for i in switches:
        if i:
            on += 1
    return on

# The correct answers are in the doc strings above, 
# but pyccelized code gets the wrong answers < 10000 switches

>>> switches(100)
48
>>> switches(1000)
497
>>> switches(10000)
100
>>> switches(100000)
316
>>> switches(1000000)
1000
```

**Expected behavior**
The accelerated function should return the same values as the pure python code.

**Language**
C

**Additional context**
This is the first time I've ever done anything with pyccel, so I may well be doing something wrong!

EmilyBourne:
Fix list comprehensions with anonymous iterators (underscore `_`). Fixes #1131 

**Commit Summary**
- Change `substitute` function to only substitute exact (same id) matches instead of subbing any objects that are equal
- [BUGFIX] Print `ListComprehension` objects in Python in Numpy arrays to ensure indexation makes sense
- [BUGFIX] Correct docstring
- [BUGFIX] Ensure the names of `ListComprehension` indices are collisionless
- Add tests for issue #1131 
- Activate C and Python tests for list comprehensions
- [BUGFIX] Remove unused temporary variables created by `FunctionalFor` expressions
- [BUGFIX] Don't export temporary variables generated by Pyccel to the generated module
- Add an error message when trying to use the value of `_`


aihya (commented):
I don't know if these changes are not supposed to work with fortran, but when I compile the following code `a = [False for _ in range(5)]` if fails with:
`ERROR at shared library generation stage
  File "/home/pyccel/venv/bin/pyccel", line 33, in <module>
  File "/home/pyccel/pyccel/commands/console.py", line 239, in pyccel
    execute_pyccel(filename,
  File "/home/pyccel/pyccel/codegen/pipeline.py", line 386, in execute_pyccel
    generated_filepath = create_shared_library(codegen,
  File "/home/pyccel/pyccel/codegen/python_wrapper.py", line 99, in create_shared_library
    wrapper_code = wrapper_codegen.doprint(codegen.ast)
  File "/home/pyccel/pyccel/codegen/printing/codeprinter.py", line 53, in doprint
    lines = self._print(expr).splitlines(True)
  File "/home/pyccel/pyccel/codegen/printing/codeprinter.py", line 89, in _print
    obj = getattr(self, print_method)(expr)
  File "/home/pyccel/pyccel/codegen/printing/cwrappercode.py", line 1283, in _print_Module
    exec_func = self.get_module_exec_function(expr, exec_func_name)
  File "/home/pyccel/pyccel/codegen/printing/cwrappercode.py", line 758, in get_module_exec_function
    assign = v.get_user_nodes(Assign)[0]
IndexError: list index out of range`

EmilyBourne:
> I don't know if these changes are not supposed to work with fortran, but when I compile the following code `a = [False for _ in range(5)]` if fails with: `ERROR at shared library generation stage Traceback (most recent call last): File "/home/pyccel/venv/bin/pyccel", line 33, in <module> sys.exit(load_entry_point('pyccel', 'console_scripts', 'pyccel')()) File "/home/pyccel/pyccel/commands/console.py", line 239, in pyccel execute_pyccel(filename, File "/home/pyccel/pyccel/codegen/pipeline.py", line 386, in execute_pyccel generated_filepath = create_shared_library(codegen, File "/home/pyccel/pyccel/codegen/python_wrapper.py", line 99, in create_shared_library wrapper_code = wrapper_codegen.doprint(codegen.ast) File "/home/pyccel/pyccel/codegen/printing/codeprinter.py", line 53, in doprint lines = self._print(expr).splitlines(True) File "/home/pyccel/pyccel/codegen/printing/codeprinter.py", line 89, in _print obj = getattr(self, print_method)(expr) File "/home/pyccel/pyccel/codegen/printing/cwrappercode.py", line 1283, in _print_Module exec_func = self.get_module_exec_function(expr, exec_func_name) File "/home/pyccel/pyccel/codegen/printing/cwrappercode.py", line 758, in get_module_exec_function assign = v.get_user_nodes(Assign)[0] IndexError: list index out of range`

Huh this is a fun problem. It is happening because the module is trying to export the iterators of the list. But these are not module wide variable. This needs fixing. Thanks for finding this !

aihya:
I tried checking for the list generated by https://github.com/pyccel/pyccel/blob/6db5f78aef82b3b1ff889576f051bee5b90b7f20/pyccel/codegen/printing/cwrappercode.py#L758

EmilyBourne:
> I tried checking for the list generated by
> 
> https://github.com/pyccel/pyccel/blob/6db5f78aef82b3b1ff889576f051bee5b90b7f20/pyccel/codegen/printing/cwrappercode.py#L758
> 
> if it's empty before accessing its elements seems to fix the problem.

Yes, this is the easy fix. But changing this hides the underlying problem. Anything exposed in the wrapper should be assigned somewhere (we don't want to expose unintialised values) so I would prefer to find why we have uninitialised values here

ohachim:
When using the underscore '_'
```python
t = [_ for _ in range(3)]
```

we get: ```*|fatal [semantic]: listcmp.py [2,9]| Undefined variable (_)*``` , is this the wanted behaviour because we always assume that the underscore going to or should be temporary?
This works in python

EmilyBourne:
@ohachim 
> When using the underscore '_'
> 
> ```python
> t = [_ for _ in range(3)]
> ```
> 
> we get: `*|fatal [semantic]: listcmp.py [2,9]| Undefined variable (_)*` , is this the wanted behaviour because we always assume that the underscore going to or should be temporary? This works in python

In python there is a convention to use the name `_` for throwaway variables (i.e. variables whose value will never be used). Pyccel follows this convention so this is the expected behaviour. I have added a more specific error to make this clear. The error raised by your code is now:
```
pyccel:
 |fatal [semantic]: tmp.py [1,4]| Variables named '_' are assumed to be throwaways so their value cannot be used
```

ohachim (approved):

yguclu (approved):

PR #1138, opened 2022-06-27
Add `const` to function arguments in generated C code

### LINKED ISSUE #1031: C printer doesn't benefit from the const keyword feature
**Describe the bug**
As Pyccel can recognize a changed variable in the semantics and provide that information to the printers, so they can add the correspondent keywords to the generated C/Fortran code, those keywords are used by C/Fortran compliers to make further optimizations on the code.

**To Reproduce**
code to reproduce the behavior:
```python
def foo(x: 'int[:]', i: 'int'):
    #some code
    return x[I]
```

the generated code:
```c
int64_t foo(t_ndarray x, int64_t i)
{
    /*some code*/
    return GET_ELEMENT(x, nd_int64, i);
}
```

**Expected behavior**
the const keyword should be added to the dtype of the argument x.
```c
int64_t foo(const t_ndarray x, int64_t i)
{
    /*some code*/
    return GET_ELEMENT(x, nd_int64, i);
}
```
**Language**
C

azouiten:
Improve C code generation:
  - Refactor `function_signature` for more readability
  - Add `const` keyword in generated C code (closes #1031)


pyccel/ast/core.py
```diff
@@ -1915,6 +1916,20 @@ def has_default(self):
         (if not then it must be provided)
         """
         return self._value is not None
+    
+    @property
+    def is_const(self):
+        """ Indicates whether the argument has been modified
+        (if not then it can be declared as a constant)
+        """
+        return self._is_const
```
EmilyBourne, line 1925: I thought this information was already in the variable. Don't you just need to do `self._var.is_const`?

pyccel/parser/semantic.py
```diff
@@ -2967,6 +2967,26 @@ def _visit_FunctionDef(self, expr, **settings):
...
+        
+        def check_if_const(arguments, code_block):
+            lines = code_block.body
+            for line in lines:
+                if isinstance(line, For) :
+                    check_if_const(arguments, line.body)
+                    
+                if isinstance(line, Assign):
+                    for arg in arguments:
+                        if (isinstance(line.lhs, PyccelSymbol) and line.lhs == arg.var)\
```
EmilyBourne, line 2980: This should occur after the code has been visited in the semantic stage. If that is the case then there won't be any `PyccelSymbol`s left, they should have all been converted to `Variable`s

pyccel/parser/semantic.py
```diff
@@ -2967,6 +2967,26 @@ def _visit_FunctionDef(self, expr, **settings):
         if len(not_used) >= 1:
             errors.report(UNUSED_DECORATORS, symbol=', '.join(not_used), severity='warning')
 
+        # Check if an argument is modified
+        
+        def check_if_const(arguments, code_block):
+            lines = code_block.body
+            for line in lines:
+                if isinstance(line, For) :
+                    check_if_const(arguments, line.body)
```
EmilyBourne, line 2976: You should be able to use [`arg.get_user_nodes(Assign)`](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/basic.py#L92-L107) for this instead of recursing. This limits the chance of you missing cases, eg:
- if
- while

pyccel/parser/semantic.py
```diff
@@ -3292,7 +3291,6 @@ def _visit_FunctionDef(self, expr, **settings):
                         methods, superclass=cls.superclass))
 
             funcs += [func]
-
```
EmilyBourne, line 3295: Please try to avoid committing files where you have only changed the whitespace. It adds things to review while not actually doing anything helpful (everyone has a different opinion on what whitespace choices are the most readable). To undo these changes you can run:
```
git checkout master -- pyccel/parser/semantic.py
```

pyccel/codegen/printing/ccode.py
```diff
@@ -1024,7 +1024,7 @@ def function_signature(self, expr, print_arg_names = True):
         else:
             arg_code = ', '.join('{}'.format(self.function_signature(i.var, False))
                         if isinstance(i.var, FunctionAddress)
-                        else '{0}'.format(self.get_declare_type(i.var)) + (i.name if print_arg_names else '')
+                        else '{1}{0}'.format(self.get_declare_type(i.var), 'const ' if i.var.is_const else '') + (i.name if print_arg_names else '')
                         for i in args)
```
EmilyBourne, line 1028: I find this statement quite hard to read. It is really big! Can we break it up into more readable pieces?
E.g:
```
dec_types = [.. for i in args]
arg_code_list = [d + i.name for d,i in zip(dec_types, args)]
arg_code = ', '.join(..)
```

EmilyBourne:
This PR is marked as draft and also has the `needs_initial_review` flag. Is it ready for review or still being worked on?

EmilyBourne:
Please give this PR a more descriptive title

pyccel/codegen/printing/ccode.py
azouiten: makes more sense, I ll make the changes necessary and push it again

pyccel/codegen/printing/ccode.py
```diff
@@ -1022,10 +1022,18 @@ def function_signature(self, expr, print_arg_names = True):
...
-                        if isinstance(i.var, FunctionAddress)
-                        else '{0}'.format(self.get_declare_type(i.var)) + (i.name if print_arg_names else '')
-                        for i in args)
+            def get_var_arg(arg, var):
+                code = "const " if var.is_const else ""
+                code += self.get_declare_type(var)
+
+                if print_arg_names:
+                    code += arg.name
+                return code
```
aihya, line 1031: Just an opinion, this could be compactly written like this
```python3
qual = "const " * var.is_const
name = arg.name * print_arg_var
return qual + self.get_declare_type(var) + name
```

aihya (approved):

pyccel/parser/semantic.py
EmilyBourne: Same comment as last time

EmilyBourne (changes requested):

pyccel/parser/semantic.py
azouiten: I overlooked this, I ll change it asap.

EmilyBourne (approved):

Issue #1136, opened 2022-06-22
Ndarray properties can't be used for an optional array


EmilyBourne:
**Describe the bug**
Ndarray properties can't be used for an optional array.

**To Reproduce**
Provide code to reproduce the behaviour:
```python
def f(x : 'int[:]' = None):
    if x is not None:
        print(x.shape[0])
        x += 3
```

Provide the generated code, or the error message:
```bash/fortran/c
/home/emily/Code/test/tmp/__pyccel__/opt_arrays.c: In function ‘f’:
/home/emily/Code/test/tmp/__pyccel__/opt_arrays.c:14:26: error: ‘x’ is a pointer; did you mean to use ‘->’?
   14 |         printf("%ld\n", x.shape[0]);
      |                          ^
      |                          ->
/home/emily/Code/test/tmp/__pyccel__/opt_arrays.c:15:26: error: ‘x’ is a pointer; did you mean to use ‘->’?
   15 |         for (i = 0; i < x.shape[0]; i += 1)
      |                          ^
      |                          ->
```

**Expected behaviour**
The problem arises because the `_print` function isn't used to print the variable. This will also cause problems when #537 is fixed as we won't be able to do:
```python
a = my_func_returns_array().shape[0]
```
With the `_print` function call pyccel should generate `(*x).shape[0]` instead of `x.shape[0]`, however a solution which allows us to print `x->shape[0]` would be preferred. This may be possible with a class-wide variable but care should be taken not to break cases where multiple property accessors are handled in the same line. E.g:
```python
def f(b : 'int[:,:]' = None):
    a = b[my_func_returns_array().shape[0]].shape[0]
```

**Language**
C

aihya:
I wonder if there exist some tests for this bug already.

EmilyBourne:
I couldn't find one. There are lots of tests for optional arrays but they don't seem to use the ndarray properties (other than `GET_ELEMENT` which looks like a function so the `*` is correctly printed)

aihya:
Could `x` be a stack array in the first example?

EmilyBourne:
> Could `x` be a stack array in the first example?

No. For a variable to be allocated on the stack its size needs to be known and the function needs to be responsible for allocating it. Therefore it doesn't make much sense to apply this concept to arguments.

Let's take the example of integers in C++:
```
void f (int& my_var);

void main() {
    int stack_variable(0);
    int* heap_variable = new int(0);
    f(stack_variable);
[...]

PR #858, opened 2021-04-15
Numpy where

nhamidn:
Implement `numpy.where` with 3 arguments. Fixes #770 , fixes #745, fixes #1123.
In order to preserve the behaviour with 1 argument (rank 1 only) `numpy.nonzero`, `numpy.count_nonzero` and `numpy.size` are also implemented.

**Commit Summary**
- [BUGFIX] Cast booleans in arithmetic and comparison operators in Fortran to ensure compilation
- Add `numpy.nonzero`
- Add `numpy.count_nonzero`
- Add `numpy.size`
- Add `a.size` treatment
- Add tests
- Update documentation
- [BUGFIX] Correct tests with booleans in `tests/epyccel/recognised_functions/test_numpy_funcs.py` (previously only testing `False`)
- Activate existing tests
- Remove where tests with complex as complex can't be passed to comparison operator

PR #1147, opened 2022-06-30
function argument with optional value

aihya:
Fixes https://github.com/pyccel/pyccel/issues/1136
- [BUGFIX] Access Ndarray attributes with `->` instead of `.` when passing array arguments with a default value:
`arg.shape[0]` becomes `arg->shape[0]`
E.g:
```python
def f(x : 'int[:]' = None):
    if x is not None:
        print(x.shape[0])
        x += 3
```
results in:
```c
void f(t_ndarray *x)
{
    int64_t i;
    if (x != NULL)
    {
        printf("%ld\n", x->shape[0]);
        for (i = 0; i < x->shape[0]; i += 1)
        {
            GET_ELEMENT((*x), nd_int64, (int64_t)i) += 3;
        }
    }
}
```

### DIFF
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -1145,6 +1145,8 @@ class CCodePrinter(CodePrinter):
         return Slice(start, stop, step)
 
     def _print_PyccelArraySize(self, expr):
+        if expr.arg.is_optional:
+            return '{}->shape[{}]'.format(expr.arg, expr.index)
         return '{}.shape[{}]'.format(expr.arg, expr.index)
 
     def _print_Allocate(self, expr):
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -1144,10 +1144,12 @@ class CCodePrinter(CodePrinter):
 
         return Slice(start, stop, step)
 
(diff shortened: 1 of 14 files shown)


pyccel/codegen/printing/ccode.py
```diff
@@ -1145,6 +1145,8 @@ def _new_slice_with_processed_arguments(_slice, array_size, allow_negative_index
         return Slice(start, stop, step)
 
     def _print_PyccelArraySize(self, expr):
+        if expr.arg.is_optional:
+            return '{}->shape[{}]'.format(expr.arg, expr.index)
```
EmilyBourne, line 1149: Have a look at the more general `stored_in_c_pointer`. If you can find a solution that tests this instead of `.is_optional` this would be preferred as it will help fix other things (e.g. returning multiple arrays from a function)

aihya: Is this problem specific to `shape` or any other ndarray property?

EmilyBourne: It happens with other properties too if they are accessed like this. I don't remember if there are any others outside of the numpy where PR

EmilyBourne: #858 is now merged so there is now `size` as well as `shape` which has this problem

Issue #1148, opened 2022-07-01
Improve readability for pointer operations in C


EmilyBourne:
**Describe the bug**
Currently pointers are dereferenced whenever they are used in C.
E.g:
```python
def f(x : 'int[:]' = None):
    if x is not None:
        print(x.shape[0])
```
is translated to:
```c
void f(t_ndarray *x)
{
    if (x != NULL)
    {
        printf("%ld\n", (*x).shape[0]);
    }
}
```
This is easy to implement as `_print_Variable` takes care of the dereferencing internally.
However it is not the standard way of handling pointers in C. As a result it is a bit harder to parse. We would prefer the following syntax:
```c
void f(t_ndarray *x)
{
    if (x != NULL)
    {
        printf("%ld\n", x->shape[0]);
    }
}
```

**Language**
C

**Tasks**
- [x] Modify `VariableAddress` so it becomes `ObjectAddress`
    This means we can obtain the address of any object, not just a variable (e.g. the result of a FunctionCall)
    Potential pitfall : Address of address (this may be used in the cwrapper)
    - [x] Rename everywhere necessary
    - [x] Update the restrictions in `__init__`
    - [x] Consider moving this class to a different file
    - [x] Update the printer
- [x] Locate all functions which access class elements
- [x] Update the functions to store the base in an `ObjectAddress` before printing
    Potential pitfall : `DottedVariable` contains 2 objects which may be pointers and need to be referenced accordingly

EmilyBourne:
> I need to know if `FunctionAddress` requires the same changes?

What changes exactly are you referring to? You can't do pointer operations on a function address

PR #1151, opened 2022-07-02
Improve pointer operations readability in generated C code

aihya:
Change format `(*x).attribute` to `x->attribute`. Fix https://github.com/pyccel/pyccel/issues/1148.


pyccel/codegen/printing/cwrappercode.py
```diff
@@ -265,7 +267,7 @@ def get_static_args(self, argument):
         """
 
         if self._target_language == 'fortran' and argument.rank > 0:
-            arg_address = VariableAddress(argument)
+            arg_address = ObjectAddress(argument)
```
ohachim, line 270: *cwrappercode.py line 268*

```self._target_language = 'fortran'```,  maybe this should not be changed since the issue is C specific, or is this C specific anyway? since the file name is ```cwrappercode.py```.

Comment (EmilyBourne) [reply pyccel/codegen/printing/cwrappercode.py:270]: The file `cwrappercode.py` takes care of printing the wrapper which links the generated code (c or Fortran) to python. If we translate the file `file.py`, the file `cwrappercode.py` prints the file `file_wrapper.c`. This file is written in C because the python interpreter is written in C. You can find documentation about C wrappers [here](https://docs.python.org/3/c-api/index.html). This file needs to call functions from the generated code. But objects are not implemented in the same way in C and Fortran. This is usually due to arrays. Fortran has arrays in the language but C doesn't so we use the `ndarrays` class instead. This means that in the wrapper we need to call the Fortran functions differently. You can find information about the fortran/c interface [here](https://stackoverflow.com/tags/fortran-iso-c-binding/info).

As this file writes C code it is pertinent to use `VariableAddress` and its replacement `ObjectAddress`.

**Review by ohachim — APPROVED** [review]
Looks good to me.

File: pyccel/ast/object_address.py [hunk 2]
```diff
@@ -0,0 +1,38 @@
...
+    ObjectAddress(Variable('int','a'))                            is  &a
+    ObjectAddress(Variable('int','a', memory_handling='alias'))   is   a
+    """
+
+    __slots__ = ('_variable', '_rank', '_precision', '_dtype', '_shape', '_order')
+    _attribute_nodes = ('_variable',)
+
+    def __init__(self, variable):
+        if not isinstance(variable, (Variable, IndexedElement, ObjectAddress)):
+            raise TypeError("object must be a Variable, IndexedElement or ObjectAddress.") #TODO: to change
```
EmilyBourne, line 25: Shouldn't this PR remove this restriction?

EmilyBourne (changes requested):

pyccel/ast/object_address.py
aihya: Do you think it's a good idea to remove the attributes of `ObjectAddress`? because somewhere in the code we access the attributes `rank` or `order` for example, even though the target object doesn't have these attributes.
Like here: https://github.com/pyccel/pyccel/blob/6eb29eb0938354e1b65281c290464a1461089c84/pyccel/codegen/printing/ccode.py#L1656

EmilyBourne: I think this restriction should be replaced with :
```
if not isinstance(var, PyccelAstNode):
```
`PyccelAstNode`s always have these attributes and it doesn't make sense to take the address of something that is not a `PyccelAstNode` (eg an `If`)

pyccel/ast/object_address.py
```diff
@@ -0,0 +1,38 @@
...
+        self._precision = variable.precision
+        self._dtype     = variable.dtype
+        self._order     = variable.order
+        super().__init__()
+
+    @property
+    def variable(self):
+        """The object whose address is of interest
+        """
+        return self._variable
```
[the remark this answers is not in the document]
aihya: https://github.com/pyccel/pyccel/pull/1151/commits/babd7f27a5d646ea55537e2c34305af81cace4e2

pyccel/codegen/printing/ccode.py
```diff
@@ -1091,6 +1092,8 @@ def _cast_to(self, expr, dtype, precision):
 
     def _print_DottedVariable(self, expr):
         """convert dotted Variable to their C equivalent"""
+        if self.stored_in_c_pointer(expr):
+            return '{}->{}'.format(self._print(expr.lhs), self._print(expr.name))
```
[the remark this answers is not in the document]
aihya: solved here https://github.com/pyccel/pyccel/pull/1151/commits/421daabae7f09522066e6371d3e4a398301a46c1

pyccel/ast/object_address.py
EmilyBourne: Did you try removing the `try`/`except` mentioned above to see if anything breaks?

EmilyBourne (approved):

pyccel/ast/object_address.py
aihya: The same exception is raised

aihya: it was `PyBuildValueNode`

EmilyBourne: Hmm, this class probably ought to inherit from `PyccelAstNode`. It is basically a `FunctionCall` and that class does inherit from `PyccelAstNode`. The properties are then set to match the results of the function call, so for `PyBuildValueNode` that would be:
- dtype : `PyccelPyObject`
- precison : 0
- rank : 0
- shape : ()
- order : None

So then to make it work we would have to handle this class in `stored_in_c_pointer` (should return true)

EmilyBourne: Please could you make this change before merging. The handling in `stored_in_c_pointer` should be done in the `cwrappercode.py` file

EmilyBourne: By "this chang" I also meant making it inherit from `PyccelAstNode`.

EmilyBourne: The changes to be made are therefore:
- Make `PyBuildValueNode` inherit from `PyccelAstNode` (set class variables as described, add `_attribute_nodes` class property, don't forget to call superclass's `__init__` method)
- Correct `stored_in_c_pointer` to recognise this object as a pointer
- Remove the try/except clause

aihya: Same problem for `Nil`, I tried to apply same solution as for `PyBuildValueNode` but was causing a cyclic imports.

EmilyBourne: Cyclic imports via what? What imports did you need to add? Doesn't `Nil` inherit from `Literal`?

EmilyBourne: I've just checked the code. I'm still not sure how you managed to create a cyclic import but I see that `Nil` currently inherits from `Basic` and is a singleton. If all the `PyccelAstNode` attributes are class properties then I don't think that should create problems, but if it does then maybe this special class should be handled explicitly by `ObjectAddress`

EmilyBourne: Ah. That makes sense. I'm not sure what the dtype should be here, possibly `NativeGeneric`. It definitely shouldn't be `PyccelPyObject` though as null is equivalent to C's `NULL` pointer which can be used in many more places than a [PyObject](https://pythonextensionpatterns.readthedocs.io/en/latest/refcount.html#pyobjects-and-reference-counting)

EmilyBourne: The possible dtypes can be found [here](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/datatypes.py)

I would suggest one of:
- `NativeGeneric` - means any type
- `NativeVoid`

EmilyBourne: If you come across other types that need this treatment try and work out what the dtype should be. For example [PyBuildValueNode](http://web.mit.edu/people/amliu/vrut/python/ext/buildValue.html) has the dtype `PyccelPyObject` because it represents a call to the following function:
```c
PyObject *Py_BuildValue(char *format, ...);
```

For a function call the dtype matches the result type

aihya: fixed in https://github.com/pyccel/pyccel/pull/1151/commits/da63217bab5688b37c467f694024648c5f0e082c https://github.com/pyccel/pyccel/pull/1151/commits/7adb2d049247787cea7b17cecddc0de2718bf1f2

pyccel/ast/cwrapper.py
```diff
@@ -232,19 +234,24 @@ def arg_names(self):
...
-    __slots__ = ('_flags','_result_args',)
+    __slots__ = ('_flags','_result_args', '_rank', '_shape', '_precision', '_dtype', '_order')
     _attribute_nodes = ('_result_args',)
 
     def __init__(self, result_args = ()):
+        self._dtype = PyccelPyObject
+        self._rank = 0
+        self._precision = 0
+        self._shape = ()
+        self._order = None
```
EmilyBourne, line 254: As these values are the same for all instances of this class they can be attributes of the class rather than the instance (see `PyModule_AddObject` for an example)

pyccel/ast/literals.py
```diff
@@ -219,14 +219,22 @@ def python_value(self):
...
-    __slots__ = ()
+    __slots__ = ('_dtype', '_precision', '_rank', '_order', '_shape')
     _attribute_nodes = ()
 
+    def __init__(self):
+        self._dtype = NativeGeneric
+        self._precision = 0
+        self._rank = 0
+        self._shape = ()
+        self._order = None
```
EmilyBourne, line 235: See above, these can be class attributes

EmilyBourne (changes requested):

pyccel/ast/literals.py
```diff
@@ -7,7 +7,7 @@
 from pyccel.utilities.metaclasses import Singleton
 
 from .basic              import PyccelAstNode, Basic
-from .datatypes          import (NativeInteger, NativeBool, NativeFloat,
+from .datatypes          import (NativeGeneric, NativeInteger, NativeBool, NativeFloat,
                                   NativeComplex, NativeString, NativeFloat)
```
EmilyBourne, line 11: This import imports `NativeFloat` twice. Would you mind fixing it

pyccel/ast/literals.py
aihya: Do you think `__init__` should be removed from `Nil`?

EmilyBourne: Yes, once the object attributes are changed to class attributes it will no longer serve any purpose (I think codacy then complains)

pyccel/ast/literals.py
aihya: https://github.com/pyccel/pyccel/pull/1151/commits/b7c2e0fed930288786083369c581105b53034e76

pyccel/ast/literals.py
aihya: https://github.com/pyccel/pyccel/pull/1151/commits/1b5eeeabd856462734ac8912320e9392995723a8

pyccel/ast/cwrapper.py
aihya: https://github.com/pyccel/pyccel/pull/1151/commits/3feedbff8619014d4e0904741c57829903530059

pyccel/ast/cwrapper.py
```diff
@@ -232,17 +234,22 @@ def arg_names(self):
         """
         return self._arg_names
 
-class PyBuildValueNode(Basic):
+class PyBuildValueNode(PyccelAstNode):
     """
-    Represents a call to the function from Python.h which create a new value based on a format string
+    Represents a call to the, function from Python.h which create a new value based on a format string
```
EmilyBourne, line 239: Unnecessary comma

EmilyBourne (approved):

PR #1154, opened 2022-07-05
Allow None in equality operators

EmilyBourne:
Convert equality operators to `is` operators if one of the arguments is `None`. Allow the case where `None` is compared to a literal

### DIFF
--- a/pyccel/ast/operators.py
+++ b/pyccel/ast/operators.py
@@ -794,7 +794,7 @@ class PyccelNe(PyccelComparisonOperator):
     I.e:
         a != b
     is equivalent to:
-        PyccelEq(a, b)
+        PyccelNe(a, b)
 
     Parameters
     ----------
--- a/pyccel/ast/operators.py
+++ b/pyccel/ast/operators.py
@@ -2,8 +2,6 @@
 # This file is part of Pyccel which is released under MIT License. See the LICENSE file or #
 # go to https://github.com/pyccel/pyccel/blob/master/LICENSE for full license details.     #
 #------------------------------------------------------------------------------------------#
-# TODO [EB 12.03.21]: Remove pylint command with PR #797
-# pylint: disable=W0201
 """
 Module handling all python builtin operators
 These operators all have a precision as detailed here:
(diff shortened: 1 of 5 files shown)


aihya (approved):
Looks good to me. Just wondering if the code below should print the message `"passed"` or not, considering that functions in python return `None` by default.

```python3
def bruh():
    print('Bruh')

if __name__ == "__main__":
    if bruh() == None:
        print("passed")
```

EmilyBourne:
> Looks good to me. Just wondering if the code below should print the message `"passed"` or not, considering that functions in python return `None` by default.
> 
> ```python
> def bruh():
>     print('Bruh')
> 
> if __name__ == "__main__":
>     if bruh() == None:
>         print("passed")
> ```
> 
> because the generated program looks like this:
> 
> ```c
> int main()
> {
>     if (0)
>     {
>         printf("%s\n", "Waaahwah");
>     }
>     return 0;
> }
> ```

I can add this if you think it is useful. But the only place I can see it being used is if the return is optional. I don't think this is something that pyccel will support. It would require using lots more pointers which would make pyccel slower for the cases it is designed to tackle

pyccel/ast/operators.py
```diff
@@ -781,6 +779,12 @@ class PyccelEq(PyccelComparisonOperator):
     """
     __slots__ = ()
 
+    def __new__(cls, arg1, arg2, simplify = False):
```
Pinkyboi, line 782: I don't really understand the use of the function parameter simplify for both `PyccelEq` and `PyccelNe`, would you mind explain it to me ?

Pinkyboi:
Its not related to your changes but i think its better to fix this comment:
https://github.com/pyccel/pyccel/blob/3adbeda3c3499114fa4af346d98afb209f8bc71f/pyccel/ast/operators.py#L791-L797

I'd guess it should be `PyccelNe(a, b)` instead of `PyccelEq(a, b)` .

pyccel/ast/operators.py
EmilyBourne: This argument is inherited from the class `PyccelBinaryOperator`. It is used when we use pyccel to add operations (usually mathematical operations) and we don't want to print them fully.
E.g. if we have:
```
for i in range(5):
    a = 1+2
```
the operator `+` should be printed fully as it appears in the original code.
But in fortran a loop is printed with inclusive bounds. `do i = 0_i64, 5_i64 - 1_i64` is a bit unnatural, we prefer to print the more natural `do i = 0_i64, 4_i64` so here we don't want it printed fully. Instead we use `simplify = True` when creating this `PyccelMinus` operator to tell the class that it is allowed to make simplifications (e.g. literal operations simplify to their value, `a+0` simplifies to `a`, etc)

aihya (approved):

pyccel/ast/operators.py
Pinkyboi: Oh i understand, thank for this clarification !

yguclu (approved):

PR #1157, opened 2022-07-14
Use GitHub CI to lint parser/semantic.py with Pylint

### LINKED ISSUE #1140: pyccel/parser/semantic.py is too large to be analysed by codacy
`pyccel/parser/semantic.py` is too large to be analysed by codacy. We should think about breaking up the file. In the meantime we should be extra careful with pull requests.

cc: @saidctb @yguclu

EmilyBourne:
Fix #1140. Fix Codacy errors. Remove unused main from `parser/semantic.py`. 

Updates to Pylint mean that f-strings are now recommended


pyccel/parser/semantic.py
```diff
@@ -555,7 +572,8 @@ def _extract_indexed_from_var(self, var, indices, expr):
 
         if isinstance(var, PythonTuple):
             def is_literal_index(a):
-                is_int = lambda a: isinstance(a, (int, LiteralInteger)) or \
+                def is_int(a):
+                    return isinstance(a, (int, LiteralInteger)) or \
                         (isinstance(a, PyccelUnarySub) and \
                          isinstance(a.args[0], (int, LiteralInteger)))
```
yguclu, line 578: Does Pylint complain about the lambda function?

EmilyBourne: Yes I was getting the error [unnecessary-lambda-assignment](https://pylint.pycqa.org/en/latest/user_guide/messages/convention/unnecessary-lambda-assignment.html):

> pyccel/parser/semantic.py:573:25: C3001: Lambda expression assigned to a variable. Define a function using the "def" keyword instead. (unnecessary-lambda-assignment)

There is an explanation for why functions should be preferred [here](https://treyhunner.com/2018/09/stop-writing-lambda-expressions/#Misuse:_naming_lambda_expressions)

pyccel/parser/semantic.py
```diff
@@ -1992,7 +2020,8 @@ def _visit_PyccelPow(self, expr, **settings):
 
     def _visit_MathSqrt(self, expr, **settings):
         func = self.scope.find(expr.funcdef, 'functions')
-        arg, = self._handle_function_args(expr.args, **settings)
+        args = self._handle_function_args(expr.args, **settings)
+        arg = args[0]
```
yguclu, line 2024: I prefer the original code, because it was clear that `_handle_function_args` must return an iterable with only one element. Does Pylint complain about this too?

yguclu: In general this is a very useful warning! But here we know that the returned tuple has exactly 1 element, so there is no danger of a "Possible unbalanced tuple unpacking". The problem is that Pylint does not understand that `len(arg) == len(expr.args)` based on static code analysis.

I suppose that could make Pylint happy by writing this ugly code,
```python
arg, *_ = self._handle_function_args(expr.args, **settings)
```
but I recommend against this workaround which makes the code more obscure.

I actually prefer to disable Pylint on the specific lines where it complains about this.

yguclu (approved):

PR #1160, opened 2022-07-18
Correct indenting

EmilyBourne:
The indenting regex doesn't match the printing order which leads to incorrect printing if `@pure` and `@elemental` are used. This PR swaps the order to match what is done in `_print_FunctionDef`.

### DIFF
--- a/pyccel/codegen/printing/fcode.py
+++ b/pyccel/codegen/printing/fcode.py
@@ -185,7 +185,7 @@ type_to_print_format = {
 
 inc_keyword = (r'do\b', r'if\b',
                r'else\b', r'type\b\s*[^\(]',
-               r'(recursive )?(pure )?(elemental )?((subroutine)|(function))\b',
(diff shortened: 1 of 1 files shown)


yguclu:
Given that this PR is not connected to an issue, can you provide a small example with the wrong result (before this PR) and the correct one (after this PR)?

EmilyBourne:
Example:
```python
@pure
@elemental
def f(a : int):
    return a+1
```
`master` branch:
```fortran
module tmp


  implicit none

  contains

  elemental pure function f(a) result(Out_0001)

  implicit none

  integer(i64) :: Out_0001
  integer(i64), value :: a

  Out_0001 = a + 1_i64
  return

end function f

end module tmp
```
`neat_printing` branch:
```fortran
module tmp


  implicit none

  contains

  elemental pure function f(a) result(Out_0001)

    implicit none

    integer(i64) :: Out_0001
    integer(i64), value :: a

    Out_0001 = a + 1_i64
    return

  end function f

end module tmp
```

yguclu (approved):

PR #1166, opened 2022-07-20
Handle np.array type in the case of unspecified dtype

### LINKED ISSUE #1152: Comparison with None fails with AttributeError
**Describe the bug**
Comparison between `None` or `Nil` and another object causes an `AttributeError`.

**To Reproduce**
```python3
if __name__ == "__main__":
    print(None == 1)
```

**Error message**
```bash/fortran/c
  ...
  File "/home/pyccel/pyccel/parser/semantic.py", line 1922, in _visit_PyccelOperator
    return self._create_PyccelOperator(expr, args)
  File "/home/pyccel/pyccel/parser/semantic.py", line 648, in _create_PyccelOperator
    expr_new = type(expr)(*visited_args)
  File "/home/pyccel/pyccel/ast/operators.py", line 364, in __init__
    super().__init__(arg1, arg2)
  File "/home/pyccel/pyccel/ast/operators.py", line 128, in __init__
    self._set_shape_rank()
  File "/home/pyccel/pyccel/ast/operators.py", line 137, in _set_shape_rank
    self._shape, self._rank = self._calculate_shape_rank(*self._args)  # pylint: disable=no-member
  File "/home/pyccel/pyccel/ast/operators.py", line 437, in _calculate_shape_rank
    strs = [a for a in args if a.dtype is NativeString()]
  File "/home/pyccel/pyccel/ast/operators.py", line 437, in <listcomp>
    strs = [a for a in args if a.dtype is NativeString()]
AttributeError: 'Nil' object has no attribute 'dtype'
```

**Expected behavior**
It should print:
```bash/fortran/c
False
```

Pinkyboi:
Fix #1152: `np.array` now chooses the correct `dtype` from the type of its arguments:

- Extract `dtype` from `fill_value` without calling `process_dtype` in `NumpyFull` and `NumpyArray`.
- Change `NumpyZerosLike`, `NumpyOnesLike`, `NumpyEmptyLike`, `NumpyFullLike` to get `dtype` from passed argument if dtype not specified.
- Change `NumpyEmpty` to handle `NativeNumeric` dtype.

### DIFF
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -313,9 +313,9 @@ def process_dtype(dtype):
     elif isinstance(dtype, LiteralString):
         dtype = str(dtype).replace('\'', '').lower()
         if dtype not in dtype_registry:
-            raise TypeError('Unknown type of {}.'.format(dtype))
+            raise TypeError(f'Unknown type of {dtype}.')
     else:
-        raise TypeError('Unknown type of {}.'.format(dtype))
+        raise TypeError(f'Unknown type of {dtype}.')
     dtype, precision = dtype_registry[dtype]
     if precision == -1:
         precision        = default_precision[dtype]
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -313,9 +313,9 @@ def process_dtype(dtype):
     elif isinstance(dtype, LiteralString):
         dtype = str(dtype).replace('\'', '').lower()
         if dtype not in dtype_registry:
-            raise TypeError('Unknown type of  %s.' % dtype)
+            raise TypeError('Unknown type of {}.'.format(dtype))
     else:
-        raise TypeError('Unknown type of  %s.' % dtype)
+        raise TypeError('Unknown type of {}.'.format(dtype))
     dtype, precision = dtype_registry[dtype]
     if precision == -1:
         precision        = default_precision[dtype]
@@ -894,8 +894,9 @@ class NumpyFull(NumpyNewArray):
         # If there is no dtype, extract it from fill_value
         # TODO: must get dtype from an annotated node
         if not dtype:
-            dtype = fill_value.dtype
-
+            dtype = DtypePrecisionToCastFunction[fill_value.dtype.name][fill_value.precision]
+        if dtype in NativeNumeric:
+            dtype = DtypePrecisionToCastFunction[dtype.name][-1]
         # Verify dtype and get precision
         dtype, precision = process_dtype(dtype)
 
@@ -926,10 +927,9 @@ class NumpyAutoFill(NumpyFull):
         the fill_value is implicitly specified
     """
     __slots__ = ()
-    def __init__(self, shape, dtype='float', order='C'):
+    def __init__(self, shape, dtype=NativeFloat(), order='C'):
         if not dtype:
             raise TypeError("Data type must be provided")
-
         super().__init__(shape, Nil(), dtype, order)
 
 #==============================================================================
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -332,7 +332,7 @@ DtypePrecisionToCastFunction = {
 def process_dtype(dtype):
     """
     Returns the corresponding type and precision to the dtype.
-    
+
     Args:
         dtype     : Datatype
 
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -331,18 +331,29 @@ DtypePrecisionToCastFunction = {
 
 def process_dtype(dtype):
     """
-    Returns the corresponding type and precision to the dtype.
+    This function will take a dtype passed to a numpy array creation function
+    and depending on the type of the passed argument, it will do the necessary treatments
+    to extract the corresponding type and precision from it then return the result.
 
-    Args:
-        dtype     : Datatype
+    This function could be useful when working with numpy creation function having a dtype argument,
+    like numpy.array, nump.arrange, numpy.linspace...
+
+    Parameters:
+    ----------
+        dtype     : PythonType | PyccelFunctionDef | String
+        The actual dtype passed to the numpy function
 
     Raises:
-        TypeError: In the case of unrecognized argument type.
-        TypeError: In the case of passed string argument not recognized as valid dtype.
+    ----------
+        TypeError : In the case of unrecognized argument type.
+        TypeError : In the case of passed string argument not recognized as valid dtype.
 
     Returns:
+    ----------
         dtype     : Datatype
+        The Datatype corresponding to the passed dtype.
         precision : int
+        The precision corresponding to the passed dtype.
     """
 
     if isinstance(dtype, PythonType):
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -32,7 +32,7 @@ from .datatypes      import (dtype_and_precision_registry as dtype_registry,
 from .internals      import PyccelInternalFunction, Slice, max_precision, get_final_precision
 from .internals      import PyccelArraySize
 
-from .literals       import LiteralInteger, LiteralFloat, LiteralComplex, LiteralString, convert_to_literal
+from .literals       import LiteralInteger, LiteralFloat, LiteralComplex, convert_to_literal
 from .literals       import LiteralTrue, LiteralFalse
 from .literals       import Nil
 from .mathext        import MathCeil
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -331,28 +331,28 @@ DtypePrecisionToCastFunction = {
 
 def process_dtype(dtype):
     """
-    This function will take a dtype passed to a numpy array creation function
-    and depending on the type of the passed argument, it will do the necessary treatments
-    to extract the corresponding type and precision from it then return the result.
+    This function takes a dtype passed to a numpy array creation function,
+    processes it in different ways depending on its type, and finally extracts
+    the corresponding type and precision from the `dtype_registry` dictionary.
 
-    This function could be useful when working with numpy creation function having a dtype argument,
-    like numpy.array, nump.arrange, numpy.linspace...
+    This function could be useful when working with numpy creation function
+    having a dtype argument, like numpy.array, numpy.arrange, numpy.linspace...
 
-    Parameters:
+    Parameters
     ----------
-        dtype     : PythonType | PyccelFunctionDef | String
+    dtype: PythonType | PyccelFunctionDef | String
         The actual dtype passed to the numpy function
 
-    Raises:
-    ----------
-        TypeError : In the case of unrecognized argument type.
-        TypeError : In the case of passed string argument not recognized as valid dtype.
+    Raises
+    ------
+    TypeError: In the case of unrecognized argument type.
+    TypeError: In the case of passed string argument not recognized as valid dtype.
 
     Returns:
     ----------
-        dtype     : Datatype
+    dtype: Datatype
         The Datatype corresponding to the passed dtype.
-        precision : int
+    precision: int
         The precision corresponding to the passed dtype.
     """
 
@@ -361,10 +361,10 @@ def process_dtype(dtype):
     if isinstance(dtype, PyccelFunctionDef):
         dtype = dtype.cls_name
 
-    if dtype  in (PythonInt, PythonFloat, PythonComplex, PythonBool):
+    if dtype in (PythonInt, PythonFloat, PythonComplex, PythonBool):
         # remove python prefix from dtype.name len("python") = 6
         dtype = dtype.__name__.lower()[6:]
-    elif dtype  in (NumpyInt, NumpyInt8, NumpyInt16, NumpyInt32, NumpyInt64, NumpyComplex, NumpyFloat,
+    elif dtype in (NumpyInt, NumpyInt8, NumpyInt16, NumpyInt32, NumpyInt64, NumpyComplex, NumpyFloat,
 				  NumpyComplex128, NumpyComplex64, NumpyFloat64, NumpyFloat32):
         # remove numpy prefix from dtype.name len("numpy") = 5
         dtype = dtype.__name__.lower()[5:]
@@ -376,8 +376,8 @@ def process_dtype(dtype):
         raise TypeError(f'Unknown type of {dtype}.')
     dtype, precision = dtype_registry[dtype]
     if precision == -1:
-        precision        = default_precision[dtype]
-    dtype            = datatype(dtype)
+        precision = default_precision[dtype]
+    dtype = datatype(dtype)
 
     return dtype, precision
 
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -896,8 +896,6 @@ class NumpyFull(NumpyNewArray):
         if not dtype:
             dtype = DtypePrecisionToCastFunction[fill_value.dtype.name][fill_value.precision]
         # Verify dtype and get precision
-        if dtype in NativeNumeric:
-            dtype = DtypePrecisionToCastFunction[dtype.name][-1]
         dtype, precision = process_dtype(dtype)
 
         # Verify array ordering
@@ -927,7 +925,7 @@ class NumpyAutoFill(NumpyFull):
         the fill_value is implicitly specified
     """
     __slots__ = ()
-    def __init__(self, shape, dtype=PythonFloat, order='C'):
+    def __init__(self, shape, dtype='float', order='C'):
         if not dtype:
             raise TypeError("Data type must be provided")
         super().__init__(shape, Nil(), dtype, order)
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -376,7 +376,6 @@ class NumpyArray(NumpyNewArray):
             dtype = arg.dtype
             prec = get_final_precision(arg)
         else:
-            # Verify array ordering
             dtype, prec = process_dtype(dtype)
         # ... Determine ordering
         order = str(order).strip("\'")
@@ -902,7 +901,6 @@ class NumpyFull(NumpyNewArray):
             dtype = fill_value.dtype
             precision = get_final_precision(fill_value)
         else:
-            # Verify array ordering
             dtype, precision = process_dtype(dtype)
 
         # Verify array ordering
@@ -946,7 +944,7 @@ class NumpyEmpty(NumpyAutoFill):
 
     def __init__(self, shape, dtype='float', order='C'):
         if dtype in NativeNumeric:
-            precision = default_precision[str(dtype)]
+            precision = default_precision[str_dtype(dtype)]
             dtype = DtypePrecisionToCastFunction[dtype.name][precision]
         super().__init__(shape, dtype, order)
     @property
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -330,6 +330,21 @@ DtypePrecisionToCastFunction = {
 #==============================================================================
 
 def process_dtype(dtype):
+    """
+    Returns the corresponding type and precision to the dtype.
+    
+    Args:
+        dtype     : Datatype
+
+    Raises:
+        TypeError: In the case of unrecognized argument type.
+        TypeError: In the case of passed string argument not recognized as valid dtype.
+
+    Returns:
+        dtype     : Datatype
+        precision : int
+    """
+
     if isinstance(dtype, PythonType):
         return dtype.dtype, get_final_precision(dtype)
     if isinstance(dtype, PyccelFunctionDef):
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -948,7 +948,7 @@ class NumpyEmpty(NumpyAutoFill):
         if dtype in NativeNumeric:
             precision = default_precision[str(dtype)]
             dtype = DtypePrecisionToCastFunction[dtype.name][precision]
-        return super().__init__(shape, dtype, order)
+        super().__init__(shape, dtype, order)
     @property
     def fill_value(self):
         return None
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -994,7 +994,6 @@ class NumpyFullLike(PyccelInternalFunction):
         # NOTE: we ignore 'subok' argument
         if not dtype:
             dtype = DtypePrecisionToCastFunction[a.dtype.name][a.precision]
-        dtype = dtype
         order = a.order if str(order).strip('\'"') in ('K', 'A') else order
         shape = Shape(a) if shape is None else shape
         return NumpyFull(shape, fill_value, dtype, order)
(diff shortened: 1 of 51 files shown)


pyccel/ast/numpyext.py
```diff
@@ -367,7 +367,7 @@ def __init__(self, arg, dtype=None, order='C'):
 
         # Verify dtype and get precision
         if dtype is None:
-            dtype = arg.dtype
+            dtype = DtypePrecisionToCastFunction[str(arg.dtype).capitalize()][arg.precision]
```
EmilyBourne, line 370: In other places where this dictionary is accessed the key is: `arg.dtype.name` are you sure this will always give the same result as `str(arg.dtype).capitalize()`?

Pinkyboi: I didn't notice that attribute, changing `str(arg.dtype).capitalize()` for `arg.dtype.name` seems like the best thing to do.

aihya (approved):

Pinkyboi:
@EmilyBourne would it be a good idea to add a type check in `process_dtype` in case of unrecognized types ?
Something like the following:
```python
    if dtype not in dtype_registry:
        raise TypeError('Unknown type of  %s.' % dtype)
```

EmilyBourne:
> @EmilyBourne would it be a good idea to add a type check in `process_dtype` in case of unrecognized types ? Something like the following:
> 
> ```python
>     if dtype not in dtype_registry:
>         raise TypeError('Unknown type of  %s.' % dtype)
> ```

I think that could be useful. I would also keep this commit: https://github.com/pyccel/pyccel/pull/1166/commits/cb23ee541ae54d32867e341eb8df49ab44e20a34

Pinkyboi:
> I think that could be useful. I would also keep this commit: https://github.com/pyccel/pyccel/commit/cb23ee541ae54d32867e341eb8df49ab44e20a34

This commit caused some issues with pyccel tests that's why i discarded it.

EmilyBourne:
> > I think that could be useful. I would also keep this commit: https://github.com/pyccel/pyccel/commit/cb23ee541ae54d32867e341eb8df49ab44e20a34
> 
> This commit caused some issues with pyccel tests that's why i discarded it.

That is because there are several other classes which use the `process_dtype` function but which aren't clear about the requested precision. Ideally I would rather you fixed the default argument in these classes

azouiten (approved):

Pinkyboi:
@EmilyBourne I change `NumpyAutoFill` for it `dtype` to be recognized in `process_dtype` by making the default `dtype` a `NativeFloat` instead of an `str`.
I'm not fully satisfied by this fix, another way to do it is just to add str handling in `process_dtype` with something like:
```python
elif type(dtype) in [str, LiteralString]:
  dtype = str(dtype).replace('\'', '').lower()
```
Which one do you think is more suited ?

EmilyBourne:
> @EmilyBourne I change `NumpyAutoFill` for it `dtype` to be recognized in `process_dtype` by making the default `dtype` a `NativeFloat` instead of an `str`.
> I'm not fully satisfied by this fix, another way to do it is just to add str handling in `process_dtype` with something like:
> ```python
> elif type(dtype) in [str, LiteralString]:
>   dtype = str(dtype).replace('\'', '').lower()
> ```
> Which one do you think is more suited ?

I prefer the solution with `PythonFloat` but doing `elif isinstance(dtype, (str, LiteralString)):` sounds useful in any case

EmilyBourne:
> @EmilyBourne I change `NumpyAutoFill` for it `dtype` to be recognized in `process_dtype` by making the default `dtype` a `NativeFloat` instead of an `str`.
> I'm not fully satisfied by this fix, another way to do it is just to add str handling in `process_dtype` with something like:
> ```python
> elif type(dtype) in [str, LiteralString]:
>   dtype = str(dtype).replace('\'', '').lower()
> ```
> Which one do you think is more suited ?

I don't have a strong preference, but I would use `PythonFloat` rather than `NativeFloat`.
I think that adding `elif isinstance(dtype, (str, LiteralString))` could be useful even if you pick the first option

EmilyBourne (changes requested):
Please add at least 1 test that was failing before this PR

Pinkyboi:
@EmilyBourne  I added added a test  in [58e2d8b](https://github.com/pyccel/pyccel/pull/1166/commits/58e2d8bada983aabcaf8c306dc1eb77ffb1944ee).

Pinkyboi:
@EmilyBourne the test was firstly added in [58e2d8b](https://github.com/pyccel/pyccel/pull/1166/commits/58e2d8bada983aabcaf8c306dc1eb77ffb1944ee).

pyccel/ast/numpyext.py
```diff
@@ -890,9 +894,10 @@ def __init__(self, shape, fill_value, dtype=None, order='C'):
         # If there is no dtype, extract it from fill_value
         # TODO: must get dtype from an annotated node
         if not dtype:
-            dtype = fill_value.dtype
-
+            dtype = DtypePrecisionToCastFunction[fill_value.dtype.name][fill_value.precision]
```
EmilyBourne, line 897: Does `process_dtype` need to be called if the type is deduced from the array elements?

pyccel/ast/numpyext.py
```diff
@@ -890,9 +894,10 @@ def __init__(self, shape, fill_value, dtype=None, order='C'):
...
         # Verify dtype and get precision
+        if dtype in NativeNumeric:
+            dtype = DtypePrecisionToCastFunction[dtype.name][-1]
```
EmilyBourne, line 900: Why is this needed?

Pinkyboi: In the way i implemented it it's needed because `NumpyFill` uses a `NativeNumeric` as a default value, this could be avoided if i reverted the default value back to an `str`.

Pinkyboi: I think that would be more suited actually.

pyccel/ast/numpyext.py
Pinkyboi: Actually not.

pyccel/ast/numpyext.py
EmilyBourne: `PythonFloat` is not a `NativeNumeric`

EmilyBourne: Reverting the default value is also fine though. I think the original default is clearer

Pinkyboi: Oh yeah, that one's from the original change i forgot to remove.

pyccel/ast/numpyext.py
EmilyBourne: In that case I would rather we set the value in an if/else block. It is unnecessarily costly to use the desired dtype and precision to find the cast function which will ensure that `process_dtype` returns the desired dtype and precision

Pinkyboi: When actually coding that if else statement, i've found myself repeating a lot of what's done in `process_dtype` especially since  a different treatment of `dtype` is done depending on the precision (python types and numpy types) , so maybe dtype and precision calculation should still pass by `process_type` ?

pyccel/ast/numpyext.py
Pinkyboi: When i checked that condition again, i remembered that i left it because it solved the issue of classes like `NumpyImag`,
passing a `NativeNumeric` type as a specified dtype.

https://github.com/pyccel/pyccel/blob/a88889df050aed67343c06cc38532612b017ecca/pyccel/ast/numpyext.py#L216-L231

pyccel/ast/numpyext.py
EmilyBourne: Maybe you misunderstood me then. I was thinking of an if/else like:
```
if not dtype:
    dtype = fill_value.dtype
    precision = fill_value.precision
else:
    dtype, precision = process_type(dtype)
```

pyccel/ast/numpyext.py
EmilyBourne: Ah ok. That makes sense. Does `NumpyImag` set the correct precision though? Presumably this handling means it also uses default precision even if we are taking the imaginary part of a `complex64`?

We need to be sure if there is an error because we are missing the handling of `NumericNative` or because the function triggering the error isn't being clear enough about the requested precision

pyccel/ast/numpyext.py
Pinkyboi: Oh yeah that works ! assuming that the precision is valid.
```python
        if not dtype:
            dtype = fill_value.dtype
            precision = fill_value.precision
            if precision == -1:
                precision = default_precision[str(dtype)]
        else:
            dtype, precision = process_dtype(dtype)
```
Something like that maybe ?

EmilyBourne: Or maybe:
```python
if not dtype:
    dtype = fill_value.dtype
    precision = get_final_precision(fill_value.precision)
else:
    dtype, precision = process_type(dtype)
```

Pinkyboi: Yep !

Pinkyboi: Fixed in [acd42c3](https://github.com/pyccel/pyccel/pull/1166/commits/acd42c3d3e928a51a9188f6b868e3068f71b83ca).

pyccel/ast/numpyext.py
Pinkyboi: Fixed in [4d429e8](https://github.com/pyccel/pyccel/pull/1166/commits/4d429e8cee532f46d505f09b3c295b7852c9b27b).

pyccel/ast/numpyext.py
```diff
@@ -366,9 +372,12 @@ def __init__(self, arg, dtype=None, order='C'):
             raise TypeError('we only accept homogeneous arguments')
 
         # Verify dtype and get precision
-        if dtype is None:
+        if not dtype:
```
EmilyBourne, line 375: Why did you change this test? Are you sure that `dtype` can't have a value which could evaluate to `False` ?

EmilyBourne (changes requested):
You seem to have fixed a lot more bugs than you first planned to. Could you add some tests which check the dtype for these cases please

pyccel/ast/numpyext.py
Pinkyboi: Can it have a value that can be evaluated as `False` ? i was unaware of that, that would apply to all dtype checks i added.

EmilyBourne: `not dtype` is equivalent to `not bool(dtype)`. Some things return `False` when you do `bool(dtype)` e.g:
- empty lists
- `LiteralFalse`
- empty string?

So I just want to check if this handling is intentional or if a comparison with `None` would be more accurate

Pinkyboi: Now that you mention it, i can see that i would more accurate to use a comparison with `None` and throw an error for the case you mentioned (done by `process_dtype`).

Pinkyboi: Fixed in [bdb80e6](https://github.com/pyccel/pyccel/pull/1166/commits/bdb80e6707ae29560e8ae000d3ceb5394f4e3017).

Pinkyboi:
@EmilyBourne i added some additional tests in [6ac1b1c](https://github.com/pyccel/pyccel/pull/1166/commits/6ac1b1ce0b6f810658dce8e22ec12048660cf90d) but still not really sure why the Windows test fails.

EmilyBourne:
> @EmilyBourne i added some additional tests in [6ac1b1c](https://github.com/pyccel/pyccel/pull/1166/commits/6ac1b1ce0b6f810658dce8e22ec12048660cf90d) but still not really sure why the Windows test fails.

I don't see a failing test. Occasionally tests fail because the random numbers generated give a badly conditioned problem

EmilyBourne:
> @EmilyBourne i added some additional tests in [6ac1b1c](https://github.com/pyccel/pyccel/pull/1166/commits/6ac1b1ce0b6f810658dce8e22ec12048660cf90d) but still not really sure why the Windows test fails.

I don't see a failing test. Occasionally tests fail because the random numbers generated give a badly conditioned problem

Pinkyboi:
@EmilyBourne i did run the Windows test again and it solved the issue.

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -3127,84 +3451,150 @@ def create_zeros_like_shape_F(n,m):
...
+
     f_real_complex64   = epyccel(create_zeros_like_val_complex64, language = language)
-    assert(isclose(     f_real_complex64()  ,      create_zeros_like_val_complex64(), rtol=RTOL, atol=ATOL))
     assert matching_types(f_real_complex64(), create_zeros_like_val_complex64())
 
+    f_complex128_auto   = epyccel(create_zeros_like_val_complex128_auto, language = language)
+    assert matching_types(f_complex128_auto(), create_zeros_like_val_complex128_auto())
+
     f_real_complex128   = epyccel(create_zeros_like_val_complex128, language = language)
-    assert(isclose(     f_real_complex128() ,      create_zeros_like_val_complex128(), rtol=RTOL, atol=ATOL))
```
EmilyBourne, line 3207: Why did you decide to remove the value tests?

EmilyBourne (changes requested):

tests/epyccel/recognised_functions/test_numpy_funcs.py
Pinkyboi: that was a mistake, ill add it back.

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -1413,6 +1413,90 @@ def create_full_val_real_complex128(val):
...
+    @types('int')
+    def create_full_val_int(val):
+        from numpy import full
+        a = full(3,val)
+        return a[0]
+    @types('float')
+    def create_full_val_float(val):
+        from numpy import full
+        a = full(3,val)
+        return a[0]
```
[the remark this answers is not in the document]
Pinkyboi: I had troubles working with multiple type decorators with `python` as `language`, ill try to investigate on the cause of that but those multiple function worked like a workaround.

pyccel/ast/numpyext.py
```diff
@@ -934,6 +943,12 @@ class NumpyEmpty(NumpyAutoFill):
     """
     __slots__ = ()
     name = 'empty'
+
+    def __init__(self, shape, dtype='float', order='C'):
+        if dtype in NativeNumeric:
+            precision = default_precision[str(dtype)]
```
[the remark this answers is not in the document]
Pinkyboi: `default_precision[str_dtype(dtype)]` seem better suited ill change it.

tests/epyccel/recognised_functions/test_numpy_funcs.py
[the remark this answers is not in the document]
EmilyBourne: This is normal. There are several test which have `xfail` or `skip` tags for this case, and there is an issue (#885 ) to fix them one day. Translating to python doesn't accelerate anything, it just acts as a guide to see the internal changes pyccel may make, so skipping these tests is not a huge problem

tests/epyccel/recognised_functions/test_numpy_funcs.py
EmilyBourne: It primarily checks the type, but extra tests are never a problem ;)

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -1413,6 +1413,72 @@ def create_full_val_real_complex128(val):
     assert(isclose(     f_real_complex128(val_float)       ,      create_full_val_real_complex128(val_float), rtol=RTOL, atol=ATOL))
     assert matching_types(f_real_complex128(val_float), create_full_val_real_complex128(val_float))
 
+@pytest.mark.parametrize( 'language', (
+        pytest.param("fortran", marks = pytest.mark.fortran),
+        pytest.param("c", marks = pytest.mark.c),
+        pytest.param("python", marks = [
+            pytest.mark.skip,
```
EmilyBourne, line 1420: Please add a reason to your skip mark. See [here](https://github.com/pyccel/pyccel/blob/master/tests/epyccel/recognised_functions/test_numpy_funcs.py#L3528-L3537) for an example

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -1413,6 +1413,72 @@ def create_full_val_real_complex128(val):
...
+    @types('float64')
+    @types('complex64')
+    @types('complex128')
+    def create_full_val_auto(val):
+        from numpy import full
+        a = full(3,val)
+        return a[0]
+
+    integer   = randint(low = iinfo('int').min,   high = iinfo('int').max,   dtype=int)
+    integer32 = randint(low = iinfo('int32').min, high = iinfo('int32').max, dtype=np.int32)
```
EmilyBourne, line 1441: Values like `iinfo('int').min` are already defined at the top of this file. See [here](https://github.com/pyccel/pyccel/blob/master/tests/epyccel/recognised_functions/test_numpy_funcs.py#L11-L33). Please use them rather than recalculating them

EmilyBourne (approved):

yguclu:
@Pinkyboi Good job! Can you please extend the PR description with a list of all the relevant modifications?

tests/epyccel/modules/arrays.py
```diff
@@ -7,6 +7,22 @@
...
+@types('int8', 'int8')
+@types('int16', 'int16')
+@types('int32', 'int32')
+@types('int64', 'int64')
+@types('float', 'float')
+@types('float32', 'float32')
+@types('float64', 'float64')
+@types('complex64', 'complex64')
+@types('complex128', 'complex128')
+def array_return_first_element(a, b):
```
yguclu, line 21: Can't we handle all these cases using a template? See [here](https://github.com/pyccel/pyccel/blob/master/tutorial/templates.md#templates-using-decorators)

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -1413,6 +1413,73 @@ def create_full_val_real_complex128(val):
...
+def test_full_dtype_auto(language):
+    @types('int')
+    @types('float')
+    @types('complex')
+    @types('int32')
+    @types('float32')
+    @types('float64')
+    @types('complex64')
+    @types('complex128')
+    def create_full_val_auto(val):
```
yguclu, line 1436: Same comment about templates

pyccel/ast/numpyext.py
```diff
@@ -340,8 +342,12 @@ def process_dtype(dtype):
 				  NumpyComplex128, NumpyComplex64, NumpyFloat64, NumpyFloat32):
         # remove numpy prefix from dtype.name len("numpy") = 5
         dtype = dtype.__name__.lower()[5:]
+    elif isinstance(dtype, (LiteralString, str)):
+        dtype = str(dtype).replace('\'', '').lower()
+        if dtype not in dtype_registry:
+            raise TypeError(f'Unknown type of {dtype}.')
     else:
-        dtype            = str(dtype).replace('\'', '').lower()
+        raise TypeError(f'Unknown type of {dtype}.')
```
yguclu, line 350: I have just noticed that this function (`process_dtype`) does not have a docstring. Would you mind writing one?

Pinkyboi:
@yguclu I added what you requested, im not really confident about the new description or the docstring if you have any remarks please tell me about em.

tests/epyccel/modules/arrays.py
EmilyBourne: Fixed [5421f09](https://github.com/pyccel/pyccel/pull/1166/commits/5421f09d6e770384f5e0a2fad5dc8d9b72b025e5)

tests/epyccel/recognised_functions/test_numpy_funcs.py
EmilyBourne: Fixed [5421f09](https://github.com/pyccel/pyccel/pull/1166/commits/5421f09d6e770384f5e0a2fad5dc8d9b72b025e5)

pyccel/ast/numpyext.py
```diff
@@ -328,6 +330,21 @@ class NumpyComplex128(NumpyComplex):
...
+        dtype     : Datatype
+
+    Raises:
+        TypeError: In the case of unrecognized argument type.
+        TypeError: In the case of passed string argument not recognized as valid dtype.
+
+    Returns:
+        dtype     : Datatype
+        precision : int
+    """
```
EmilyBourne, line 346: We try to use the same formatting for all our docstrings which allows them to be recognised by doxygen if we ever have enough documentation to make it worth adding automated documentation. See [here](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/core.py#L164-L188) for an example of a docstring.

After reading the docstring we should understand:
- [x] what the purpose of the function is
- [ ] In what context the function should be used
- [ ] What the expected type of each of the arguments are
- [ ] What each of the arguments is for
- [x] What the expected type of each of the results are

Currently we haven't added a section for possible errors but I think this could be useful so I would keep this.

I think the information missing from your docstring is:
- Argument type (If `dtype` is a `DataType` then a `TypeError` is raised so the type that you have written is not the expected type)
- Argument role
- The description could be expanded to explain where this function should be used (i.e. to handle `dtype` arguments of array creation functions). The explanation should be clear enough to avoid the confusion that you had when you started tackling this issue.

Pinkyboi:
```
    """
    This function will take a dtype passed to a numpy array creation function
    and depending on the type of the passed argument, it will do the necessary treatments
    to extract the corresponding type and precision from it then return the result.

    This function could be useful when working with numpy creation function having a dtype argument,
    like numpy.array, nump.arrange, numpy.linespace...
    
    Parameters:
    ----------
        dtype     : PythonType | PyccelFunctionDef | String | Numpy Datatype | Python Builtin Datatype
        The actual dtype passed to the numpy function

    Raises:
    ----------
        TypeError : In the case of unrecognized argument type.
        TypeError : In the case of passed string argument not recognized as valid dtype.

    Returns:
    ----------
        dtype     : Datatype
        The Datatype corresponding to the passed dtype.
        precision : int
        The precision corresponding to the passed dtype.
    """
```
Is something like this better ?

EmilyBourne: That looks a lot better :)

Typo `linespace` -> `linspace`

What `PyccelFunctionDef` other than a `Numpy Datatype` or a `Python Builtin Datatype` can be used as an argument?

Pinkyboi: From what i know it can only be one of those two.

EmilyBourne: Then I don't think you need to mention `PyccelFunctionDef` separately.
You should have one of:
```
dtype     : PythonType | PyccelFunctionDef | String
```
```
dtype     : PythonType | String | Numpy Datatype | Python Builtin Datatype
```
```
dtype     : PythonType | PyccelFunctionDef (Numpy Datatype or Python Builtin Datatype) | String
```

EmilyBourne (approved):

yguclu (approved):

yguclu:
Fix #1152: `np.array` now chooses the correct `dtype` from the type of its arguments:

- Extract `dtype` from `fill_value` without calling `process_dtype` in `NumpyFull` and `NumpyArray`.
- Change `NumpyZerosLike`, `NumpyOnesLike`, `NumpyEmptyLike`, `NumpyFullLike` to get `dtype` from passed argument if dtype not specified.
- Change `NumpyEmpty` to handle `NativeNumeric` dtype.

Issue #1134, opened 2022-06-22
Complex type printing is wrong


EmilyBourne:
**Describe the bug**
Complex type printing is wrong as it misses a multiplication

**To Reproduce**
Provide code to reproduce the behavior:
```python
import numpy as np

if __name__ == '__main__':
    a = np.complex64(3)
    print(type(a))
```

Python output:
```
<class 'numpy.complex64'>
```

Pyccelised output:
```
<class 'complex32'>
```

**Expected behavior**
Pyccel should output:
```
<class 'complex64'>
```

**Additional context**
The value is set in [`PythonType.print_string`](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/builtins.py#L879-L886)

PR #1167, opened 2022-07-20
Fix wrong complex precision printing

aihya:
Multiplied complex precision by 2.
Now `<class 'complex64'>` is printed instead of `<class 'complex32'>` for the following code:
```python3
import numpy as np
if __name__ == "__main__":
    a = np.complex64(3)
    print(type(a))
```
Fixes https://github.com/pyccel/pyccel/issues/1134

### DIFF
--- a/pyccel/ast/builtins.py
+++ b/pyccel/ast/builtins.py
@@ -884,7 +884,7 @@ class PythonType(Basic):
         str_dtype = str(self.dtype)
         return LiteralString("<class '{dtype}{precision}'>".format(
             dtype = str_dtype,
-            precision = '' if prec in (None, -1) else (prec*8*[1, 2][str_dtype=="complex"])))
+            precision = '' if prec in (None, -1) else (prec * 8 * (2 if str_dtype=="complex" else 1))))
 
 #==============================================================================
 python_builtin_datatypes_dict = {
--- a/pyccel/ast/builtins.py
+++ b/pyccel/ast/builtins.py
@@ -881,9 +881,10 @@ class PythonType(Basic):
         can be used in a print  statement
         """
         prec = self.precision
+        str_dtype = str(self.dtype)
         return LiteralString("<class '{dtype}{precision}'>".format(
-            dtype = str(self.dtype),
-            precision = '' if prec in (None, -1) else prec*8))
+            dtype = str_dtype,
+            precision = '' if prec in (None, -1) else (prec*8*[1, 2][str_dtype=="complex"])))
 
 #==============================================================================
 python_builtin_datatypes_dict = {
--- a/pyccel/ast/builtins.py
+++ b/pyccel/ast/builtins.py
@@ -881,10 +881,9 @@ class PythonType(Basic):
         can be used in a print  statement
         """
         prec = self.precision
-        str_dtype = str(self.dtype)
         return LiteralString("<class '{dtype}{precision}'>".format(
-            dtype = str_dtype,
-            precision = '' if prec in (None, -1) else (prec * 8 * (2 if str_dtype=="complex" else 1))))
+            dtype = str(self.dtype),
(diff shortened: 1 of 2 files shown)


azouiten (commented):
the second line could use some readability, maybe a ternary operator would be a little bit better here.

pyccel/ast/builtins.py
```diff
@@ -881,9 +881,10 @@ def print_string(self):
         can be used in a print  statement
         """
         prec = self.precision
+        str_dtype = str(self.dtype)
         return LiteralString("<class '{dtype}{precision}'>".format(
-            dtype = str(self.dtype),
-            precision = '' if prec in (None, -1) else prec*8))
+            dtype = str_dtype,
+            precision = '' if prec in (None, -1) else (prec*8*[1, 2][str_dtype=="complex"])))
```
Pinkyboi, line 887: This like could be changed to something like this i think:
```python
precision = '' if prec in (None, -1) else (prec * 8 * (2 if str_dtype=="complex" else 1))))
```

aihya: fixed here: https://github.com/pyccel/pyccel/pull/1167/commits/8da5291f3e1bf05cf26c039b8f53ad1b16cd5a9d

Pinkyboi (approved):
Looks alright to me, although i don't fully understand why we should double the precision in the case of complex, i mean by that why do the 64 has a precision of 4 and 128 a precision of 8 to begin with.

aihya:
> Looks alright to me, although i don't fully understand why we should double the precision in the case of complex, i mean by that why do the 64 has a precision of 4 and 128 a precision of 8 to begin with.

I'm not quite sure why it was originally written with those precisions, but it might be related to the way it is printed for fortran.

azouiten (approved):

EmilyBourne:
@aihya @Pinkyboi 128 is because a complex requires 128 bits of storage, 64 for the real part, and 64 for the imaginary part. 64 bits= 8 bytes

Fortran and pyccel store precision in bytes. We use 8 for complex128 to show that the numbers have the same precision as an 8 byte float. They need twice as much memory though as there are two numbers with this precision

pyccel/ast/builtins.py
```diff
@@ -881,9 +881,10 @@ def print_string(self):
         can be used in a print  statement
         """
         prec = self.precision
+        str_dtype = str(self.dtype)
         return LiteralString("<class '{dtype}{precision}'>".format(
-            dtype = str(self.dtype),
-            precision = '' if prec in (None, -1) else prec*8))
+            dtype = str_dtype,
+            precision = '' if prec in (None, -1) else (prec * 8 * (2 if str_dtype=="complex" else 1))))
```
EmilyBourne, line 887: Why not `(16 if str_dtype == "complex" else 8)`?

EmilyBourne, line 887: I think it would be safer to check if `dtype is NativeComplex()` in case some wise guy in the future decides to name his class "complex"

EmilyBourne (changes requested):

pyccel/ast/builtins.py
aihya: fixed in https://github.com/pyccel/pyccel/pull/1167/commits/f2bc151f02f353d1d6ea6060edd3a641f5d3f669

aihya: fixed in https://github.com/pyccel/pyccel/pull/1167/commits/f2bc151f02f353d1d6ea6060edd3a641f5d3f669

EmilyBourne (approved):

EmilyBourne (changes requested):
Please add a test in this file: https://github.com/pyccel/pyccel/blob/master/tests/pyccel/scripts/runtest_type_print.py

yguclu (approved):

PR #1128, opened 2022-06-09
Creating array with list variables, implements fortran order, and nested arrays

ohachim:
Fixes #752, creating arrays with a list of variables
Fixes #1179
Fixes #682
Implements fortran order memory layout
Implements nested array creation

PR #1168, opened 2022-07-22
Create function to copy data from a sliced ndarray using strides.

soukainahikma:
- [x] Adding a condition to prevent variables from flattening when rank is greater than 1.
- [x] Adding a C function to copy  from a sliced array using strides.

Description of the solution: 

# N_offset calculation

This is a alt text.
Given an array A:

| 1 | 2 | 3 | 
|----|----|----|
| 4 | 5 | 6 |
| 7 | 8 | 9 |

    To move from 1 to 2 we jump with 8 bytes.
    To move from 1 to 4 we jump with 24 bytes.

Representation of A flattened (in one dimension), as it was set in the memory:

| 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 |
|----|----|----|----|----|----|----|----|----|

To access element 5, ```A[1][1]```:

	position = 1 * 24 + 1 * 8
    		= 32;
 
Which means we move with 32 bytes to access element 5;

Given the indexes we can calculate the position, but how can we calculate the indexes using the position and the shape of each dimension ?

I came up with an algorithm that reverses the calculation of the positions from the indexes, by observing the relationship between the positions, the shape of each dimension , and the indexes:

this is an example to explore, and expand it more:

Given an 3D array A:

	A = [[[1,2],[3,4]], [[5,6],[7,8]], [[9,10],[11,12]]]

| 1 | 2|  | 3 | 4 |  |  | 5 | 6 |  | 7 | 8 |  |  | 9 | 10 |  | 11 | 12 |
|----|----|----|----|----|----|----|----|----|----|----|----|----|----|----|----|----|----|----|

   position of ```A[i][j][k]```:
    
    position = i * strides_i + j * strides_j + k * strides_k
    
  Calculations of i, j and k:
  
  	shape_i = 3
  	shape_j = 2
  	shape_k = 2

    
 | position | i | j | k|
|----		|----|----|----|
| 0 		| 0 | 0 | 0 |
| 1 		| 0 | 0 | 1 |
| 2 		| 0 | 1 | 0 |
| 3 		| 0 | 1 | 1 |
| 4 		| 1 | 0 | 0 |
| 5 		| 1 | 0 | 1 |
| 6 		| 1 | 1 | 0 |
| 7 		| 1 | 1 | 1 |
| 8 		| 2 | 0 | 0 |
| 9 		| 2 | 0 | 1 |
| 10 		| 2 | 1 | 0 |
| 11 		| 2 | 1 | 1 |

k = position % shape_k

j = position / shape_k ```if j > shape_j ==> j = j % shape_j```

i = position / (shape_j * shape_k)  ```if i > shape_i ==> i = i % shape_i```


EmilyBourne:
Beware, this PR and #1128 both seem to be changing the same part of the code. You should speak to @ohachim

soukainahikma:
> 

That's right, thank you for mentioning it, we are currently discussing the changes on both issues

yguclu:
According to @bauom it will be fixed by #1280.

PR #1169, opened 2022-07-23
Add numpy.sign function

### LINKED ISSUE #1165: Add numpy.sign function
**Describe the bug**
Add numpy.sign function

**Test Code**
```python
a = sign(-1)
```

aihya:
Fix #1165: add support for the function `numpy.sign` for scalars and arrays.

For a scalar input `x`, the function returns a scalar `y`:
  - If `x > 0`, then `y = 1`
  - If `x == 0`, then `y = 0`
  - If `x < 0`, then `y = -1`

For an array input `x`, the function returns an array `y`, where `y[i]` is computed from `x[i]` following the rule above.

For efficiency and readability reasons, the new directory `pyccel/stdlib/numpy` contains the C and Fortran implementations of `numpy.sign` for various datatypes and precisions.


aihya:
@EmilyBourne do you have an idea how to fix the codacy error?

EmilyBourne:
> @EmilyBourne do you have an idea how to fix the codacy error?

No. We have the same problem in PR #1056 but we are not supposed to be using the django plugin. I have a ticket open at Codacy. Hopefully they will have time to look into it soon

ohachim (changes requested):
Using python lists directly causes errors in both C and Fortran if one of the values is `0` and when giving it a primitive list as a parameter:
Example:
```python
if __name__ == "__main__":
    import numpy as np
    x = np.array([1, 2, 3, 0])
    x[:] = np.sign([0])
```

aihya:
> Using python lists directly causes errors in both C and Fortran if one of the values is `0` and when giving it a primitive list as a parameter: Example:
> 
> ```python
> if __name__ == "__main__":
>     import numpy as np
>     x = np.array([1, 2, 3, 0])
>     x[:] = np.sign([0])
> ```

fixed in https://github.com/pyccel/pyccel/pull/1169/commits/6946249edd8194a5d5889f60e81a4540101702de

ohachim:
@aihya It seems that the issue persists, at least on a MAC, although using the same values, but changing the list with a tuple (example ```(0)``` instead of ```[0]```) works fine, I understand that per Pyccel's documentation, we should use tuples instead of lists, could this be affecting this?

ohachim (approved):

pyccel/ast/numpyext.py
```diff
@@ -1238,6 +1239,25 @@ class NumpyArctanh(NumpyUfuncUnary):
...
 
+
+class NumpySign(NumpyUfuncUnary):
+    """Represent a call to the sign function in the Numpy library"""
+    __slots__ = ('_x',)
+    name = 'sign'
+
+    def __init__(self, x):
+        super().__init__(x)
+        self._x = x
```
EmilyBourne, line 1250: `x` is a pyccel object so it should be saved in `_attribute_nodes`. Alternatively you could change the property to:
```python
@property
def x(self):
    return self._args[0]
```

pyccel/codegen/printing/fcode.py
```diff
@@ -1074,6 +1074,21 @@ def _print_NumpyArray(self, expr):
                 rhs_code = self._print(expr.arg)
         return rhs_code
 
+    def _print_NumpySign(self, expr):
+        str_x = self._print(expr.x)
+        kind  = self.print_kind(expr.x)
+
+        if expr.dtype == NativeComplex():
```
EmilyBourne, line 1081: I think this should be an identity comparison not just an equality so it would be better as:
```
expr.dtype is NativeComplex()
```
```
expr.dtype is NativeComplex()
```
or
```
isinstance(expr.dtype, NativeComplex)
```

pyccel/stdlib/ndarrays/ndarrays.h
```diff
@@ -140,4 +141,16 @@ int64_t         get_index(t_ndarray arr, ...);
 int64_t     *numpy_to_ndarray_strides(int64_t *np_strides, int type_size, int nd);
 int64_t     *numpy_to_ndarray_shape(int64_t *np_shape, int nd);
 
+/* numpy.sign for float, double and integers */
+inline double  sign(double x)
+{
+    return x ? x < 0 ? -1 : 1 : 0;
```
EmilyBourne, line 147: Please add parentheses to help read this instruction

tests/epyccel/modules/numpy_sign.py
```diff
@@ -0,0 +1,147 @@
...
+def literal_float_pos():
+    import numpy as np
+    b = np.sign(42.42)
+    return b
+
+def literal_float_neg():
+    import numpy as np
+    b = np.sign(-42.42)
+    return b
+
```
EmilyBourne, line 114: Please add tests for:
- argument is 0 :  e.g `np.sign(0)`/`np.sign(0.0)`
- complex argument with real part equal to 0 : e.g `np.sign(1j)`

tests/epyccel/test_epyccel_sign.py
```diff
@@ -0,0 +1,167 @@
...
+    x2_1d = np.copy(x1_1d)
+    x2_2d = np.copy(x1_2d)
+
+    f_1d(x1_1d)
+    f_2d(x1_2d)
+    f_1d_epyc(x2_1d)
+    f_2d_epyc(x2_2d)
+
+    assert np.array_equal(x1_1d, x2_1d)
+    assert np.array_equal(x1_2d, x2_2d)
```
EmilyBourne, line 167: Please check the result type in this file

EmilyBourne (changes requested):
Please expand the tests to cover the special cases

pyccel/ast/numpyext.py
aihya: I don't see why the property should be changed as we have only one argument `x` in the `__init__` for now. Unless you mean `__init__` must be written like this:
```python
def __init__(self, *args):
    self._args = args
    super().__init__(self._args[0])
```

EmilyBourne: The `_args` attribute is already set in the super class `PyccelInternalFunction`. I think that either this should be used:
```python
def __init__(self, x):
        super().__init__(x)

@property
def x(self):
    return self._args[0]
```
or if the argument is saved directly then it needs to be saved as an `_attribute_node` so that the `substitute` function is not broken. This can either be done by having `x` and `_args`:
```python
_attribute_nodes = ('_args', '_x')
def __init__(self, x):
        self._x = x
        super().__init__(x)

@property
def x(self):
    return self._x
```
or by just keeping `x`:
```python
_attribute_nodes = ('_x',)
def __init__(self, x):
        self._x = x
        super().__init__()

@property
def x(self):
    return self._x
```
In the second case you will need a specific `_print` function for `NumpySign` as `_print_NumpyUfuncBase` won't be able to use `.args`

aihya: Good to know, I always wondered when something is printed with `_print_NumpyUfuncBase` or a specific `_print_` function for numpy functions. So the existence of `_attribute_nodes` defines the `_print` method.

aihya: I didn't know about the `args` thingy, I think the class would be simpler now without having the property `x` or `__init__`.

EmilyBourne: > Good to know, I always wondered when something is printed with `_print_NumpyUfuncBase` or a specific `_print_` function for numpy functions. So the existence of `_attribute_nodes` defines the `_print` method.

Not at all. The choice of which `_print` function is explained in the developer docs here : https://github.com/pyccel/pyccel/blob/master/developer_docs/overview.md#function-naming-conventionsfile-navigation

All pyccel classes which inherit from `pyccel.ast.basic.Basic` have the `_attribute_nodes` property, but sometimes it is inherited from a superclass. It is used in [`pyccel.ast.basic.Basic`](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/basic.py#L28-L402) for functions such as:
- checking if an object is inside another (e.g. is the variable `a` somewhere inside the function `f`)
- replacing an object with another (e.g. replacing an array `b` with an element `b[i]` in the unravelling)

EmilyBourne: > I didn't know about the `args` thingy, I think the class would be simpler now without having the property `x` or `__init__`.

I agree the class could be reduced to:
```python
class NumpySign(NumpyUfuncUnary):
    """Represent a call to the sign function in the Numpy library"""
    __slots__ = ('_x',)
    name = 'sign'

    def _set_dtype_precision(self, x):
        self._dtype = x.dtype
        self._precision = default_precision[str_dtype(self._dtype)]
```

aihya: fixed here https://github.com/pyccel/pyccel/pull/1169/commits/dc70a1011e6f1c78dded7b279c0ccaa74cc63573

pyccel/stdlib/ndarrays/ndarrays.h
aihya: fixed here https://github.com/pyccel/pyccel/pull/1169/commits/621c3e82d16a51ea2afdd94a346fccc6a92c01f8

tests/epyccel/modules/numpy_sign.py
aihya: fixed here https://github.com/pyccel/pyccel/pull/1169/commits/3d87cc43c8f951a52ec7589eb230d3d605834360

pyccel/codegen/printing/fcode.py
aihya: https://github.com/pyccel/pyccel/pull/1169/commits/8336ee19c4181a9f391a62810e67b4d179406bfa

aihya:
> Please expand the tests to cover the special cases

I can't think of special cases other than those in the tests.

tests/epyccel/test_epyccel_sign.py
aihya: https://github.com/pyccel/pyccel/pull/1169/commits/36076c263bd4d21e926a97b45a064a3ba6dec39c

EmilyBourne:
> > Please expand the tests to cover the special cases
> 
> I can't think of special cases other than those in the tests.

That's what I meant, github just wouldn't let me request changes without leaving a message because I left the comments individually instead of starting a review 😅

EmilyBourne (approved):

azouiten (approved):

tests/epyccel/test_epyccel_sign.py
```diff
@@ -0,0 +1,185 @@
...
+    x2_1d = np.copy(x1_1d)
+    x2_2d = np.copy(x1_2d)
+
+    f_1d(x1_1d)
+    f_2d(x1_2d)
+    f_1d_epyc(x2_1d)
+    f_2d_epyc(x2_2d)
+
+    assert np.array_equal(x1_1d, x2_1d) and isinstance(x1_1d, type(x2_1d))
+    assert np.array_equal(x1_2d, x2_2d) and isinstance(x1_2d, type(x2_2d))
```
EmilyBourne, line 185: Please check the return types in this file

aihya: https://github.com/pyccel/pyccel/pull/1169/commits/0beda3e83578bfa9d98889ad2f1434fc97eacf17

tests/epyccel/test_epyccel_sign.py
```diff
@@ -0,0 +1,185 @@
...
+    f_pos = mod.literal_float_pos
+    f_neg = mod.literal_float_neg
+    f_nul = mod.literal_float_nul
+    f_pos_epyc = epyccel(f_pos, language = language)
+    f_neg_epyc = epyccel(f_neg, language = language)
+    f_nul_epyc = epyccel(f_nul, language = language)
+
+    assert f_pos_epyc() == f_pos()
+    assert f_neg_epyc() == f_neg()
+    assert f_nul_epyc() == f_nul()
```
EmilyBourne, line 123: These tests are also in this file and don't check dtype

aihya: I tried the following test and I was getting a weird behavior, if you can explain why that happens.
```python
import numpy as np

if __name__ == "__main__":
    b = np.sign(np.complex64(1+4j))
    print(b)
```

Inside the function `_print_NumpySign` I was printed at the beginning the `expr.args` and, logically speaking, I was expecting an object of type `NumpyComplex64` but I got `(Literal((1+4j)),)`

EmilyBourne: `NumpyComplex64` is a cast function. Casting a literal to a specific precision is equivalent to keeping the literal, but changing the precision

aihya: Well the kind of the real part of the complex number says `f64` instead of `f32`

yguclu: > Well the kind of the real part of the complex number says `f64` instead of `f32`. `(1.0_f64, 5.0_f64)`

That is indeed a bug. `numpy.complex64` is a complex number type composed of two 32-bit-precision floating-point numbers

EmilyBourne: Although this is a bug, it shouldn't lead to wrong results so you don't need to worry about it for this PR

pyccel/ast/numpyext.py
```diff
@@ -1277,6 +1278,15 @@ class NumpyArctanh(NumpyUfuncUnary):
...
 #=======================================================================================
 
+
+class NumpySign(NumpyUfuncUnary):
+    """Represent a call to the sign function in the Numpy library"""
+    __slots__ = ()
+    name = 'sign'
+    def _set_dtype_precision(self, x):
+        self._dtype     = x.dtype
+        self._precision = default_precision[str_dtype(self._dtype)]
```
EmilyBourne, line 1288: Are you sure that the precision should always be equal to the default precision? Similar code is used in lots of functions in this file, but I suspect these are bugs (similar to #1177 )

tests/epyccel/test_epyccel_sign.py
aihya: Tests were failing when I was checking for dtype, specifically for `numpy.complex64/128`

aihya: I suppose I should use that `get_final_precision` function too in the `NumpySign`.

EmilyBourne: I suspect so, but you need to check the behaviour of `NumpySign`. `get_final_precision` keeps the precision of the object you pass to it, but converts it to a positive value. This means, for example, that the function can't return a `float` (precision=-1), only a `float32` or `float64`. `default_precision` contains the numpy precision of the pure python type (`int`, `float`, `complex`)

tests/epyccel/test_epyccel_sign.py
```diff
@@ -0,0 +1,185 @@
...
+from pyccel.epyccel import epyccel
+
+def test_sign_complex(language):
+    f_pos = mod.complex_pos
+    f_neg = mod.complex_neg
+    f_pos_epyc = epyccel(f_pos, language = language)
+    f_neg_epyc = epyccel(f_neg, language = language)
+
+    assert f_pos_epyc() == f_pos()
+    assert f_neg_epyc() == f_neg()
```
EmilyBourne, line 15: Please add the requested type checks to this file

EmilyBourne (changes requested):

tests/epyccel/test_epyccel_sign.py
aihya: https://github.com/pyccel/pyccel/pull/1169/commits/ba06f25b8b31da097dad9a55ff2a75e1eaf93940

EmilyBourne (approved):

tests/epyccel/modules/numpy_sign.py
```diff
@@ -0,0 +1,177 @@
...
+def complex128_pos():
+    import numpy as np
+    b = np.sign(np.complex128(128+128j))
+    return b
+
+def complex128_neg():
+    import numpy as np
+    b = np.sign(np.complex128(-128-128j))
+    return b
```
yguclu, line 33: You have not tested the case where the real part is positive and the imaginary part is negative, or vice versa. I suspect that you may be up for a surprise!

yguclu (changes requested):

tests/epyccel/modules/numpy_sign.py
aihya: Fixed in the last commit

pyccel/stdlib/ndarrays/ndarrays.h
```diff
@@ -140,4 +141,16 @@ int64_t         get_index(t_ndarray arr, ...);
...
+inline double  sign(double x)
+{
+    return x ? (x < 0 ? -1 : 1) : 0;
+}
+
+/* numpy.sign for complex */
+inline double complex csign(double complex x)
+{
+    return x ? ((!creal(x) && cimag(x) < 0) || (creal(x) < 0) ? -1 : 1) : 0;
+}
```
yguclu, line 154: Do we handle the situation where the user has defined a variable named `sign` or `csign`?

CC: @EmilyBourne

yguclu (changes requested):

pyccel/stdlib/ndarrays/ndarrays.h
EmilyBourne: This case is not handled. To fix it you should add these keywords to the list [here](https://github.com/pyccel/pyccel/blob/master/pyccel/naming/cnameclashchecker.py#L16)

pyccel/codegen/printing/fcode.py
```diff
@@ -1076,6 +1076,21 @@ def _print_NumpyArray(self, expr):
...
+            ne_zero = f'(REALPART({str_x}) .ne. 0_{kind}) .or. (IMAGPART({str_x}) .ne. 0_{kind})'
+            lt_zero = f'((REALPART({str_x}) .eq. 0_{kind}) .and. IMAGPART({str_x}) .lt. 0_{kind}) .or. (REALPART({str_x}) .lt. 0_{kind})'
+        else:
+            ne_zero = f'({str_x} .ne. 0_{kind})'
+            lt_zero = f'({str_x} .lt. 0_{kind})'
+
+        merge_non_zero = f'merge(-1_{kind}, 1_{kind}, {lt_zero})'
+
+        return f'merge({merge_non_zero}, 0_{kind}, {ne_zero})'
+
```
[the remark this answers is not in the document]
EmilyBourne: I agree with the principle, but I think that this is yet another example of something that ought to be handled by a dedicated routine.

At one point we planned to add a section of pyccel to contain (usually numpy) routines that need to be written in both Fortran and C. To avoid writing these manually for both languages and for any languages we may want to use in the future, we planned to write these routines in python and translate them using pyccel. These routines were supposed to be translated during the installation.

Unfortunately we still have not implemented this. Previously when looking at things which would benefit from this mechanism we have often said that we will leave them to one side for now to avoid doing the work twice.
One of the main things holding up the discussion about this mechanism was the lack of support for functions returning arrays which meant that most of the functions we would want to translate couldn't be translated. This is no longer an issue. I think it is time we revisited this. We would probably need to organise a meeting to discuss the implementation details.

If there is a chance of this being implemented soon then I don't know if it is worth duplicating the work if we already have a working implementation in Fortran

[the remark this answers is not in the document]
yguclu: I agree that we should have a meeting about this mechanism soon. But in this specific case I am not sure that pyccelizing a function written in Python would lead to C and Fortran code that is as efficient as the code written here by hand. It is worth trying, though!

pyccel/stdlib/ndarrays/ndarrays.h
yguclu: @aihya: Please implement Emily's suggestion

tests/epyccel/test_epyccel_sign.py
```diff
@@ -0,0 +1,224 @@
...
+
+def test_sign_arr_complex(language):
+    f_1d = mod.array_complex_1d
+    f_2d = mod.array_complex_2d
+
+    f_1d_epyc = epyccel(f_1d, language = language)
+    f_2d_epyc = epyccel(f_2d, language = language)
+
+    x1_1d = np.array([0.+.1j, -2.-1.3j, -3.7j], dtype=np.complex64)
+    x1_2d = np.array([[0.+1.], [-2.-1.3j], [-3.7j]], dtype=np.complex64)
```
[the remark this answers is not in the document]
aihya: fixed here https://github.com/pyccel/pyccel/pull/1169/commits/4a9ec455c07d0fba7b98df9e42c5609305aae5dd

pyccel/stdlib/ndarrays/ndarrays.h
aihya: fixed here https://github.com/pyccel/pyccel/pull/1169/commits/0b718a47c15eb8e5595fe4ef3fb9dbcf27928a83

pyccel/codegen/printing/fcode.py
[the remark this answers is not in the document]
aihya: Ok, I don't know if this code should remain, but I managed to make a fortran module containing the sign functions for REAL, INTEGER and COMPLEX types, and it's working for arrays too.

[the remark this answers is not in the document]
aihya: The new fortran module is located [here](https://github.com/pyccel/pyccel/blob/numpy.sign/pyccel/stdlib/numpy/numpy_f90.f90)

pyccel/codegen/printing/ccode.py
```diff
@@ -1233,14 +1234,12 @@ def _print_NumpyUfuncBase(self, expr):
...
-        # add necessary include
+        self.add_import(c_imports['math'])
         type_name = type(expr).__name__
         try:
             func_name = numpy_ufunc_to_c_float[type_name]
-            if func_name in ('sign', 'csign'):
-                self.add_import(c_imports['ndarrays'])
-            else:
-                self.add_import(c_imports['math'])
+            if func_name in ('sign'):
```
EmilyBourne, line 1241: I don't think this will work. `('sign')` is not a tuple, it is a string. You are checking if the function name is one of `s`, `i`, `g`, or `n`. You should use `if func_name in ('sign',):`, or (more logically when only comparing with 1 value `if func_name == 'sign':`

pyccel/codegen/printing/ccode.py
```diff
@@ -1233,14 +1234,12 @@ def _print_NumpyUfuncBase(self, expr):
             numpy.cos(x) ==> cos(x)
 
         """
-        # add necessary include
+        self.add_import(c_imports['math'])
```
EmilyBourne, line 1237: Is the math library needed for all `NumpyUfuncBase` functions? For example, is it needed for `sign`?

pyccel/codegen/printing/ccode.py
```diff
@@ -1231,11 +1234,12 @@ def _print_NumpyUfuncBase(self, expr):
             numpy.cos(x) ==> cos(x)
 
         """
-        # add necessary include
         self.add_import(c_imports['math'])
         type_name = type(expr).__name__
         try:
             func_name = numpy_ufunc_to_c_float[type_name]
+            if func_name in ('sign'):
+                self.add_import(c_imports['numpy_c'])
```
EmilyBourne, line 1242: Are you sure that `NumpySign` is general enough to not need its own `_print_NumpySign` function? I'm not sure that the types are handled correctly here. `np.sign(3)` returns a `numpy.int64` but here you will have a double. This will cause problems if you translate:

```
print(np.sign(3))
```

pyccel/codegen/printing/fcode.py
```diff
@@ -2651,6 +2653,8 @@ def _print_NumpyUfuncBase(self, expr):
             func_name = numpy_ufunc_to_fortran[type_name]
         except KeyError:
             self._print_not_supported(expr)
+        if func_name.startswith('numpy_'):
+            self._additional_imports.add(Import('numpy_f90', Module('numpy_f90',(),())))
         args = [self._print(NumpyFloat(a) if a.dtype is NativeInteger() else a)\
 				for a in expr.args]
```
EmilyBourne, line 2659: See above. Would a `_print_NumpySign` function be useful?

EmilyBourne (changes requested):

pyccel/codegen/printing/ccode.py
EmilyBourne: When translating to fortran I obtain:
```
At line 9 of file /home/emily/Code/test/tmp/__pyccel__/prog_tmp2.f90 (unit = 6, file = 'stdout')
Fortran runtime error: Expected INTEGER for item 1 in formatted transfer, got REAL
(I0)
 ^

Error termination. Backtrace:
#0  0x7f9693748700 in ???
#1  0x7f9693749259 in ???
#2  0x7f9693749f3f in ???
#3  0x7f969396fcc1 in ???
#4  0x7f969397c9d2 in ???
#5  0x7f9693980085 in ???
#6  0x7f96939807e3 in ???
#7  0x562508f6e341 in ???
#8  0x562508f6e0ae in ???
#9  0x7f9693559082 in __libc_start_main
	at ../csu/libc-start.c:308
#10  0x562508f6e0ed in ???
#11  0xffffffffffffffff in ???
```
When translating to C I obtain a compilation warning:
```
/home/emily/Code/pyccel/pyccel/codegen/compiling/compilers.py:431: UserWarning: /home/emily/Code/test/tmp/__pyccel__/prog_tmp2.c: In function ‘main’:
/home/emily/Code/test/tmp/__pyccel__/prog_tmp2.c:8:15: warning: format ‘%ld’ expects argument of type ‘long int’, but argument 2 has type ‘double’ [-Wformat=]
    8 |     printf("%ld\n", sign(3.0));
      |             ~~^     ~~~~~~~~~
      |               |     |
      |               |     double
      |               long int
      |             %f

  warnings.warn(UserWarning(err))
```
then:
```
140725433417880
```

pyccel/stdlib/numpy/numpy_f90.f90
```diff
@@ -0,0 +1,86 @@
...
+! or go to https://github.com/pyccel/pyccel/blob/master/LICENSE for full license details. !
+! --------------------------------------------------------------------------------------- !
+
+module numpy_f90
+
+    use ISO_C_BINDING
+    
+    implicit none
+    
+    interface numpy_sign
```
EmilyBourne, line 12: This new keyword should also be added to the fortran name clash checker

pyccel/stdlib/numpy/numpy_f90.f90
```diff
@@ -0,0 +1,86 @@
...
+! This file is part of Pyccel which is released under MIT License. See the LICENSE file   !
+! or go to https://github.com/pyccel/pyccel/blob/master/LICENSE for full license details. !
+! --------------------------------------------------------------------------------------- !
+
+module numpy_f90
+
+    use ISO_C_BINDING
+    
+    implicit none
+
```
EmilyBourne, line 11: You should add private and public declarations to ensure that if you do `use numpy_f90` only `numpy_sign` is imported.

I.e.
```
module numpy_f90

    use ISO_C_BINDING

    implicit none

    private

    public :: numpy_sign


    interface numpy_sign
    ...
```

pyccel/codegen/printing/fcode.py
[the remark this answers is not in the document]
yguclu: Well done @aihya! I think we can use your new file for now. @EmilyBourne Is that fine with you?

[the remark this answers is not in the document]
EmilyBourne: Fine with me

pyccel/codegen/printing/ccode.py
yguclu: True! The fact that all tests pass suggests that we need much stricter unit testing.

pyccel/codegen/printing/fcode.py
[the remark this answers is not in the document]
EmilyBourne: Yes I assumed you were just asking if we could resolve this conversation

pyccel/codegen/printing/ccode.py
EmilyBourne: After a quick check it seems that `func_name in 'sign'` is not equivalent to `func_name in ('s','i','g','n')` so I don't think this problem can cause failing tests. It can just introduce unnecessary imports
```
In [3]: func_name in 'sign'                                                                                                                                                                                
Out[3]: True

In [4]: 's' in 'sign'                                                                                                                                                                                      
Out[4]: True
```

pyccel/codegen/printing/fcode.py
aihya: Do you suggest to put back the `_print_NumpySign`? Definitely it was working fine, but it seems like it wasn't readable.

pyccel/codegen/printing/ccode.py
aihya: I think the reason you get that value is because of the format string, but the actual result was fine.

EmilyBourne: You are correct. The format string matches the type that pyccel tells it to expect (the dtype/precision of `NumpySign`), but your function does not return the type that it says it does. Hence the wrong printing. This should be handled by your function **not** by `_print_PythonPrint`.

pyccel/codegen/printing/fcode.py
EmilyBourne: I don't remember what was unreadable. Was it this PR that used `x` but didn't have it saved in the `_attribute_nodes` ? I don't think it was unreadable, just possibly unnecessary as all it did was duplicate the code in `_print_NumpyUfuncBase`. However if different code is needed to solve the printing problem then a `_print_NumpySign` function becomes useful again

pyccel/stdlib/numpy/numpy_f90.f90
aihya: https://github.com/pyccel/pyccel/pull/1169/commits/f882cc173f3165b9bc46a35f4aec0fc3e411dee7

pyccel/stdlib/numpy/numpy_f90.f90
aihya: https://github.com/pyccel/pyccel/pull/1169/commits/0fcbccaba0791ab93f4070a2cca89a76263b973f

aihya: https://github.com/pyccel/pyccel/pull/1169/commits/0fcbccaba0791ab93f4070a2cca89a76263b973f https://github.com/pyccel/pyccel/pull/1169/commits/253338b4f6440c2e33a9566c7d459d7040bb8be8

pyccel/codegen/printing/fcode.py
aihya: https://github.com/pyccel/pyccel/pull/1169/commits/32e8a921d1a7389b253fb12f7cd4c4c3a5445182 I kept the fortran module and added separate printer

pyccel/codegen/printing/ccode.py
aihya: https://github.com/pyccel/pyccel/pull/1169/commits/d4745f9dc505e1f41dfad8e5055377eb82e30489 https://github.com/pyccel/pyccel/pull/1169/commits/da2619bf2fbda70678123e49267e344d601002d8 This issue is fixed new

pyccel/codegen/printing/ccode.py
aihya: https://github.com/pyccel/pyccel/pull/1169/commits/b4ce59313ab7007e6a6e695b0122457cd329f609 this is not needed anymore as moved it to `_print_NumpySign`

pyccel/codegen/printing/ccode.py
aihya: All these lines have been removed, as they are no longer needed.

pyccel/codegen/printing/ccode.py
EmilyBourne: Could you add some tests in the `tests/pyccel/` folder with a few different types to confirm this please

pyccel/codegen/printing/ccode.py
```diff
@@ -1253,7 +1254,6 @@ def _print_NumpyUfuncBase(self, expr):
             numpy.cos(x) ==> cos(x)
 
         """
-        # add necessary include
         self.add_import(c_imports['math'])
```
EmilyBourne, line 1257: Unnecessary removal of comment

pyccel/codegen/printing/ccode.py
```diff
@@ -1276,6 +1276,41 @@ def _print_NumpyUfuncBase(self, expr):
...
+        self.add_import(c_imports['numpy_c'])
+        func = ''
+        if isinstance(expr.dtype, NativeInteger):
+            func = 'isign'
+        elif isinstance(expr.dtype, NativeFloat):
+            func = 'fsign'
+        elif isinstance(expr.dtype, NativeComplex):
+            func = 'csign'
+        else:
+            errors.report(f'{PYCCEL_RESTRICTION_TODO} ({expr.dtype}[kind = {expr.precision}])', severity='fatal')
```
EmilyBourne, line 1310: Please add the `symbol` keyword to this error call (this will allow pyccel to determine the line where the error is coming from). As you are writing your own error message rather than relying on the auto-generated message raised when calling `NotImplementedError` I think it would be better to provide more information in the error. For example it is not obvious from this error that it is coming from the `sign` function

pyccel/codegen/printing/fcode.py
```diff
@@ -2657,6 +2657,30 @@ def _print_NumpyUfuncBase(self, expr):
...
+
+        Example
+        -------
+            import numpy
+
+            numpy.sign(x) => numpy_sign(x)
+            numpy_sign is an interface which calls the proper function depending on the data type of x
+
+        """
+        self._additional_imports.add(Import('numpy_f90', Module('numpy_f90',(),())))
```
EmilyBourne, line 2681: Can you import the `numpy_sign` function instead of all the module? I don't remember the syntax off the top of my head, but it should be something like:
```
self._additional_imports.add(Import('numpy_f90', Module('numpy_f90',('numpy_sign',),())))
```

pyccel/stdlib/ndarrays/ndarrays.h
```diff
@@ -9,6 +9,7 @@
 # include <complex.h>
 # include <stdbool.h>
 # include <stdint.h>
+# include <math.h>
```
EmilyBourne, line 12: Why does `ndarrays.h` need a new dependence if you haven't modified this file?

pyccel/stdlib/numpy/numpy_f90.f90
```diff
@@ -0,0 +1,145 @@
+! --------------------------------------------------------------------------------------- !
+! This file is part of Pyccel which is released under MIT License. See the LICENSE file   !
+! or go to https://github.com/pyccel/pyccel/blob/master/LICENSE for full license details. !
+! --------------------------------------------------------------------------------------- !
+
+module numpy_f90
+
+    use ISO_C_BINDING
```
EmilyBourne, line 8: As in the generated files it would be better to use `use, intrinsic :: ISO_C_Binding`

pyccel/stdlib/numpy/numpy_f90.f90
```diff
@@ -0,0 +1,145 @@
...
+
+    ! Implementation of numpy.sign function for real numbers
+    elemental function i8_sign(x) result(y)
+
+        implicit none
+
+        INTEGER(C_INT8_T), value :: x
+        INTEGER(C_INT8_T)        :: y
+        
+        y = merge(merge(-1, 1, x .lt. 0), 0, x .ne. 0)
```
EmilyBourne, line 37: It would be better to use type specifiers for the literals in this file:
```
y = merge(merge(-1_C_INT8_T, 1_C_INT8_T, x .lt. 0_C_INT8_T), 0_C_INT8_T, x .ne. 0_C_INT8_T)
```
As you can see this is hard to read, which is why we usually rename these values when we import:
```
y = merge(merge(-1_i8, 1_i8, x .lt. 0_i8), 0_i8, x .ne. 0_i8)
```
(It may be faster to write this file in python and get pyccel to give you the fortran version, you can then modify it to add the complex case following the example that you generate)

EmilyBourne (changes requested):

pyccel/codegen/printing/fcode.py
aihya: I tried to put it in `funcs` like this:
self._additional_imports.add(Import('numpy_f90', Module('numpy_f90',(),('numpy_sign',)))) but it doesn't work.

EmilyBourne: There is an example here: https://github.com/pyccel/pyccel/blob/master/pyccel/codegen/printing/fcode.py#L372

pyccel/stdlib/numpy/numpy_f90.f90
aihya: https://github.com/pyccel/pyccel/pull/1169/commits/b4ef1309aa2a48ad51b70675d90333c15ab270d4

pyccel/stdlib/numpy/numpy_f90.f90
aihya: https://github.com/pyccel/pyccel/pull/1169/commits/b4ef1309aa2a48ad51b70675d90333c15ab270d4

pyccel/stdlib/ndarrays/ndarrays.h
aihya: https://github.com/pyccel/pyccel/pull/1169/commits/5f170614f3bdfc8fa25ebc2e2a752d282dca8896

pyccel/codegen/printing/fcode.py
aihya: https://github.com/pyccel/pyccel/pull/1169/commits/11d147f49d0a936d3d6167cd71b4b6492dc2c903

pyccel/codegen/printing/ccode.py
aihya: https://github.com/pyccel/pyccel/pull/1169/commits/7e6dc6a384ba3dc2a0b2cca80ae60993d79eee08 checking for dtype inside the class now https://github.com/pyccel/pyccel/pull/1169/commits/0d6ac52998719e9972a6177a0033ca6413b8030a

pyccel/codegen/printing/ccode.py
aihya: https://github.com/pyccel/pyccel/pull/1169/commits/145ea58b2db75e69eb450efe233f7f1def24ca2f

tests/epyccel/modules/numpy_sign.py
```diff
@@ -0,0 +1,222 @@
...
+
+@types('complex64[:]')
+def array_complex_1d(x):
+    import numpy as np
+    x[:] = np.sign(x)
+
+@types('complex64[:,:]')
+def array_complex_2d(x):
+    import numpy as np
+    x[:,:] = np.sign(x)
```
EmilyBourne, line 222: As we are now capable of returning arrays I think we should add/modify the tests to have tests of the type:
```
@types('complex64[:,:]')
def array_complex_2d(x):
    import numpy as np
    y = np.sign(x)
    return y
```
This will let you check the dtype of the returned array

EmilyBourne (changes requested):

pyccel/codegen/printing/fcode.py
aihya: https://github.com/pyccel/pyccel/pull/1169/commits/869205ede015568fca2bc1e9f3fc30591954f916

EmilyBourne (approved):
Thanks for all the modifications :) this looks good now

pyccel/ast/numpyext.py
```diff
@@ -1322,6 +1323,17 @@ class NumpyArctanh(NumpyUfuncUnary):
...
+
+class NumpySign(NumpyUfuncUnary):
+    """Represent a call to the sign function in the Numpy library"""
+    __slots__ = ()
+    name = 'sign'
+    def _set_dtype_precision(self, x):
+        if not isinstance(x.dtype, (NativeInteger, NativeFloat, NativeComplex)):
+            raise TypeError(f'{x.dtype} not supported')
+        self._dtype     = x.dtype
+        self._precision = get_final_precision(x)
```
EmilyBourne, line 1335: Is this the expected behaviour? Windows doesn't seem to like it

EmilyBourne (changes requested):
There is also still a codacy error

pyccel/ast/numpyext.py
aihya: Yeah, for some reason the generated python array does not have the same dtype as pyccel's. I'm working on it.

aihya: I thought maybe I was using `long int isign` instead of `long long int isign` but it doesn't seem to be the issue.

EmilyBourne: It's a good job we modified those tests ;)

Given it's only appearing on Windows it's almost certainly got something to do with how int is handled. Int is int32 by default on Windows and int64 everywhere else but some functions like to do things differently. I'm guessing this line is linked to the problem anyway.

EmilyBourne: It's odd that only the arrays are complaining though

EmilyBourne:
Well done on finding the error ([def6005](https://github.com/pyccel/pyccel/pull/1169/commits/def60050e9325046c801e82cdba263dd8b349ed0)) next time you might want to check why a test is failing though ^^. The corrected version ran into an overflow error (in an unrelated test) which disappears if I rerun the test.

EmilyBourne (approved):

pyccel/stdlib/numpy/numpy_c.c
```diff
@@ -0,0 +1,24 @@
...
+double  fsign(double x)
+{
+    return SIGN(x);
+}
+
+/* numpy.sign for complex */
+double complex csign(double complex x)
+{
+    return x ? ((!creal(x) && cimag(x) < 0) || (creal(x) < 0) ? -1 : 1) : 0;
+}
```
yguclu, line 24: No newline at end of file

pyccel/stdlib/numpy/numpy_f90.f90
```diff
@@ -0,0 +1,156 @@
...
+      x_lt_zero = ((REALPART(x) .eq. 0_f64) .and. IMAGPART(x) .lt. 0_f64) &
+              .or. (REALPART(x) .lt. 0_f64)
+
+      Out_0001 = merge(merge(-1_c64, 1_c64, x_lt_zero), 0_c64, x_ne_zero)
+      return
+
+    end function numpy_sign_c64
+    !........................................
+  
+  end module numpy_f90
```
yguclu, line 156: No newline at end of file

yguclu (changes requested):
Great job @aihya! Please add `sign` to the list of supported functions [here](https://github.com/pyccel/pyccel/blob/master/tutorial/numpy-functions.md#other-functions).

pyccel/stdlib/numpy/numpy_f90.f90
aihya: https://github.com/pyccel/pyccel/pull/1169/commits/2a644c9e99378e351e50faea29acd54ef1e34f9a

pyccel/stdlib/numpy/numpy_c.c
aihya: https://github.com/pyccel/pyccel/pull/1169/commits/2a644c9e99378e351e50faea29acd54ef1e34f9a

aihya:
> Great job @aihya! Please add `sign` to the list of supported functions [here](https://github.com/pyccel/pyccel/blob/master/tutorial/numpy-functions.md#other-functions).

Added here: https://github.com/pyccel/pyccel/pull/1169/commits/3e87899dd0cc7de5db95fe45a31c13c5f78aa131

yguclu (approved):

PR #1177, opened 2022-08-04
Fix the precision of numpy.abs function

### LINKED ISSUE #1176: Wrong precision of elements of numpy.complex64 and numpy.complex128
**Describe the bug**
The precision of the real and imaginary parts of a complex number are wrongly set when `numpy.complex64`, or `numpy.complex128` are passed to another function.

If we print the precision of real and imaginary parts of `np.complex64(4+2j)` from inside the printer function of NumpyAbs (for example) we find `<class 'float64'>` instead of `<class 'float32'>`, 

**To Reproduce**
```python
if __name__ == "__main__":
    import numpy as np
    np.abs(np.complex64(4+2j))
```

**Expected behavior**
The precision of the real part should be `<class 'float32'>` for `np.complex64`, and `<class 'float64'>` for `np.complex128`

EmilyBourne:
Fix the precision of `numpy.absolute` function and the internal precision of the real and imaginary parts of a `LiteralComplex`. The latter caused no bugs so far as the precision of the complex was used to interpret the value and no literals with precisions higher than the default precision can be translated. Fixes #1176


aihya (approved):

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -170,6 +170,27 @@ def absolute_call_i(x):
...
+    assert matching_types(f1(x), absolute_call_c(x))
+
+    x = np.complex64(uniform(high=1e6)-1j*uniform(high=1e6))
+    assert(isclose(f1(x), absolute_call_c(x), rtol=RTOL, atol=ATOL))
+    assert matching_types(f1(x), absolute_call_c(x))
+
+    x = np.complex128(uniform(high=1e6)-1j*uniform(high=1e6))
+    assert(isclose(f1(x), absolute_call_c(x), rtol=RTOL, atol=ATOL))
+    assert matching_types(f1(x), absolute_call_c(x))
+
```
yguclu, line 193: Why do we need to use `isclose` to verify that the absolute value is correct? Assuming that the IEEE 754 standard is respected by all implementations, I would expect all digits to be correct after calling `abs`.

yguclu (changes requested):

tests/epyccel/recognised_functions/test_numpy_funcs.py
yguclu: Sorry. I have just realized that these tests are for complex numbers only. It makes sense to use a tolerance for comparison, then!

yguclu: The question remains for testing the functions `absolute_call_r` and `absolute_call_i`. But this not strictly related to this PR.

aihya (approved):

yguclu (approved):

PR #1178, opened 2022-08-06
Fix modulo in augassign

### LINKED ISSUE #1172: Modulo in augassign doesn't work in C
**Describe the bug**
Modulo in augassign doesn't work in C

**To Reproduce**
```python
a %= 2*pi
```

```bash/fortran/c
error: invalid operands to binary % (have ‘double’ and ‘double’)
```

**Expected behaviour**
The `AugAssign` should be converted to a normal `Assign` (`a = a%(2*pi)`) at the printer

**Language**
C

aihya:
Fix #1172: C printer now supports `%=` operation with doubles and floats using `fmod` and `fmodf` functions.

### DIFF
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -222,10 +222,6 @@ ndarray_type_registry = {
                   ('int',1)     : 'nd_int8',
                   ('bool',4)    : 'nd_bool'}
 
-modulo_fmod = {
-    'double': 'fmod',
-    'float' : 'fmodf'}
-
 import_dict = {'omp_lib' : 'omp' }
 
 c_imports = {n : Import(n, Module(n, (), ())) for n in
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -222,6 +222,10 @@ ndarray_type_registry = {
                   ('int',1)     : 'nd_int8',
                   ('bool',4)    : 'nd_bool'}
 
+modulo_fmod = {
+    'double': 'fmod',
+    'float' : 'fmodf'}
+
 import_dict = {'omp_lib' : 'omp' }
 
 c_imports = {n : Import(n, Module(n, (), ())) for n in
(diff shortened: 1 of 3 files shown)


azouiten (approved):

pyccel/codegen/printing/ccode.py
```diff
@@ -1632,10 +1636,22 @@ def _print_PyccelUnarySub(self, expr):
...
+        if op == '%':
+            # Two cases where this works: (NativeFloat and NativeFloat), (NativeFloat and NativeInt)
+            # The case (NativeInt and NativeFloat) would cause a type change error.
+            if isinstance(lhs.dtype, NativeFloat) or isinstance(rhs.dtype, NativeFloat):
+                self.add_import(c_imports['math'])
+                # Get (dtype, precision) pair for dtype_registry
+                key  = (str(lhs.dtype), get_final_precision(lhs))
+                # Get the proper modulo function (fmod for doubles, fmodf for floats)
+                func = modulo_fmod[dtype_registry[key]]
+                return "{0} = {1}({0}, {2});\n".format(lhs_code, func, rhs_code)
```
EmilyBourne, line 1653: Does this code duplicate what is in `_print_PyccelMod`? If it doesn't what is the difference and why is there is a difference?
If it does duplicate then this is dangerous practice (see [DRY principle](https://thevaluable.dev/dry-principle-cost-benefit-example/)). If anyone changes `_print_PyccelMod` in the future then they are likely to forget to also change this bit of code. Why not use something similar to Fortran's [`_print_AugAssign`](https://github.com/pyccel/pyccel/blob/master/pyccel/codegen/printing/fcode.py#L2066-L2068) to handle the modulo case?

EmilyBourne (changes requested):
Please add a test that was failing before this PR

pyccel/codegen/printing/ccode.py
aihya: Ok, I didn't know `pyc_modulo` and `pyc_fmodulo` exist already. That renders all changes I did useless 😄.

EmilyBourne: :laughing: that sounds like a good argument for mimicking Fortran's implemenation of `_print_AugAssign`

aihya: https://github.com/pyccel/pyccel/pull/1178/commits/aef0e90a9ae684bdb60f7a7e855bf8a7e9ee10de

EmilyBourne: As C actually has augassign operators I think it would be good to keep these where possible. I would just use something similar to Fortran's [_print_AugAssign](https://github.com/pyccel/pyccel/blob/master/pyccel/codegen/printing/fcode.py#L2066-L2068) to handle the modulo case which can't be written as an augmented assign

EmilyBourne (changes requested):
Please add a test that was failing before this PR

pyccel/codegen/printing/ccode.py
aihya: https://github.com/pyccel/pyccel/pull/1178/commits/815fed512c3e8f800f0a04aadcf334978beb913d

aihya:
> Please add a test that was failing before this PR

fixed https://github.com/pyccel/pyccel/pull/1178/commits/8018659b37f10d57a67c065871fbca5325503e95

EmilyBourne (approved):

EmilyBourne (approved):

yguclu (approved):

PR #1180, opened 2022-08-12
Correct wrong array order and several bugs returning fortran ordered arrays

### LINKED ISSUE #1171: Wrong array order
**Description**
When not specifying the order of an ndarray, Pyccel defaults to order C, which is correct, unless the array was created using and order F ndarray
**To Reproduce**
```python
if __name__ == "__main__":
    import numpy as np
    a = np.array([[1], [2], [4]], order="F")
    c = np.array(a)
```

```c
int main()
{
    a = array_create(2, (int64_t[]){3, 1}, nd_int64, false, order_f);
    int64_t array_dummy[] = {1, 2, 4};
    memcpy(a.nd_int64, array_dummy, a.buffer_size);
    c = array_create(2, (int64_t[]){3, 1}, nd_int64, false, order_c);
    array_copy_data(c, a, 0);
    /*fff fcf cfc ffc ccf ccc*/
    free_array(a);
    free_array(c);
    return 0;
}
```

**Expected behavior**
array_create should have ```order_f``` as a last parameter

**Additional context**
The problem is in the Variable classes order attribute, it should be parsed as order F.

EmilyBourne:
Correct wrong array order when initialising from a fortran ordered array (fixes #1171). Implemtation of tests showed several bugs when returning fortran ordered arrays. These are also fixed.

**Commit Summary**
- Update version patch number
- Add a test for `np.array` which returns an array and checks flags
- [BUGFIX] Ensure numpy is aware that it owns the data returned by a function to ensure that this memory is freed
- [BUGFIX] Only transpose 2d arrays saved to Fortran ordered objects if the transpose is required
- [BUGFIX] pyccel and numpy handle slicing differently which requires a more complicated `_ndarray_to_numpy_strides` function to return a fortran ordered array
- [BUGFIX] Correct numpy type when returning an array with rank>1


yguclu:
Hi @EmilyBourne, what's the status of this PR?

EmilyBourne:
This PR is supposed to fix a memory leak (I think some of the other bug fixes have been fixed in other branches since). It works perfectly for linux and mac but on Windows it is raising "double free". I need someone with a windows machine to run the test with a tool such as drmemory. I haven't been able to find anyone who has a windows machine and would be able to debug this. I managed to install pyccel on a very old Windows machine, but so far I haven't convinced it to run without getting stuck so I haven't been able to test myself. I can "fix" the problem with an `ifdef` macro but I am not sure that there wouldn't be a windows memory leak

EmilyBourne:
All issues identified here have been fixed/are being fixed elsewhere

PR #1181, opened 2022-08-15
Use correct format string when printing long integers in C

### LINKED ISSUE #1124: Printing literal integers uses incorrect static cast leading to wrong value for negatives
**Describe the bug**

 Printing integers translates to a ```printf``` with format specifier ```%ld``` which works perfectly with positive numbers but it overflows with negative numbers.

**To Reproduce**
```python
if __name__ == '__main__':
    print(-19)
```
Python output:
```
-19
```

Translated output:
```
4294967277
```

**Language**
C

**Additional context**

C translates the line to:
```c
printf("%ld\n", -19);
```

Hellio404:
Fix #1124: use correct static cast when printing integers in C.

List of modifications:

- Add `f` suffix to `float32` float literals
- Use macro `INT64_C` for appending suffix to `int64` integer literals
- Use macros `PRId64` , `PRId16`, `PRId8` to get correct conversion specifier (`%ld %lld %hd ...`) for `int64`, `int16` and `int8`
- Introduce new `CMacro` class representing a C Macro name
- Introduce new `CStringExpression` representing a holder for a list of `CMacro`s and `LiteralString`s


pyccel/codegen/printing/ccode.py
```diff
@@ -551,6 +551,10 @@ def _print_PythonBool(self, expr):
         return '({} != 0)'.format(value)
 
     def _print_Literal(self, expr):
+        if isinstance(expr, LiteralInteger) and get_final_precision(expr) == 8:
+            return f"{repr(expr.python_value)}l"
+        elif isinstance(expr, LiteralFloat) and get_final_precision(expr) == 4:
+            return f"{repr(expr.python_value)}l"
```
aihya, line 557: I think this could be simply written like:
```suggestion
        if isinstance(expr, (LiteralInteger, LiteralFloat)) and get_final_precision(expr) in (8, 4):
            return f"{repr(expr.python_value)}l"
```

aihya: In the case of `LiteralFloat`, why `get_final_precision` should return `4` and not `8` like for `LiteralInteger`?

aihya: If it's not the case, I think you should replace `l` with `f` in line 557.

Hellio404: yeah after I change from string formatting to f-string I didn't pay attention to the 'f'
I don't know if there is a way to explicitly get a float (kind=4) instead of the default float(kind=8)
```python
def fn(a: 'float32'):
    pass

fn(2.3)
```
the example above will not work and I wonder if there is a way to make it work with just literal

EmilyBourne: `2.3` has type `float`, for `float32` you need to do:
```python
import numpy as np

def fn(a: 'float32'):
    pass

fn(np.float32(2.3))
```

EmilyBourne: > I think this could be simply written like:
> 
> ```python
> if isinstance(expr, (LiteralInteger, LiteralFloat)) and get_final_precision(expr) in (8, 4):
>     return f"{repr(expr.python_value)}l"
> ```

I disagree. This doesn't do the same thing as the code above. It is equivalent to:
```python
if isinstance(expr, LiteralInteger) and get_final_precision(expr) in (4,8):
    return f"{repr(expr.python_value)}l"
elif isinstance(expr, LiteralFloat) and get_final_precision(expr) in (4,8):
    return f"{repr(expr.python_value)}l"
```
I don't think an `int32` should have an `l`

EmilyBourne:
Don't forget to add a test. You can use [test_print_strings](https://github.com/pyccel/pyccel/blob/master/tests/pyccel/test_pyccel.py#L633-L635) and its [associated script](https://github.com/pyccel/pyccel/blob/master/tests/pyccel/scripts/print_strings.py) as an example

Hellio404:
I did create some tests for printing integrals. master failed.

I wonder if this behavior is intended:
```python
print(int32(5))
print(int32(-5))
```
will generate in c
```c
printf("%d\n", 5);
printf("%d\n, (int32_t)(-5L));
```
why does the -5 got cast but not the 5

Hellio404:
@EmilyBourne is there a way to prevent Codacy from falling, like is there a linter or something like that?

EmilyBourne:
> @EmilyBourne is there a way to prevent Codacy from falling, like is there a linter or something like that?

Codacy runs pylint which is a linter. To see the errors that codacy will report locally you can run pylint with the rcfile .pylintrc from the home directory

EmilyBourne:
> I did create some tests for printing integrals. master failed.
> 
> I wonder if this behavior is intended:
> 
> ```python
> print(int32(5))
> print(int32(-5))
> ```
> 
> will generate in c
> 
> ```c
> printf("%d\n", 5);
> printf("%d\n, (int32_t)(-5L));
> ```
> 
> why does the -5 got cast but not the 5

I am not sure why this happens. It would be great if you could investigate. I would look for the problem in `_print_PythonPrint` and bear in mind that `5` is `LiteralInteger(5)`, but `-5` is `PyccelUnarySub(LiteralInteger(5))`

Hellio404:
> 

I did run the command `pylint --rcfile=.pylintrc  tests/pyccel/scripts/print_integrals.py` and it's said `Your code has been rated at 10.00/10` so am a bit confused

EmilyBourne:
> > 
> 
> I did run the command `pylint --rcfile=.pylintrc tests/pyccel/scripts/print_integrals.py` and it's said `Your code has been rated at 10.00/10` so am a bit confused

Hmm, I'm not sure why that would happen. Currently it's just failing because the last character in the file isn't a new line character. This is something that some editors care about and others don't

Hellio404:
> > > 
> > 
> > 
> > I did run the command `pylint --rcfile=.pylintrc tests/pyccel/scripts/print_integrals.py` and it's said `Your code has been rated at 10.00/10` so am a bit confused
> 
> Hmm, I'm not sure why that would happen. Currently it's just failing because the last character in the file isn't a new line character. This is something that some editors care about and others don't

Yeah but I did add a commit just to add that new line but it didn't got fixed weird

tests/pyccel/scripts/print_integrals.py
```diff
@@ -0,0 +1,25 @@
...
+    print(00)
+    print(1)
+    print(-1)
+    print(-0)
+    print(10000)
+    print(-10000)
+    print(2147483647)
+    print(-2147483648)
+    print(2147483648)
+    print(9223372036854775807)
```
EmilyBourne, line 16: Beware printing such large numbers. As they are not in cast statements they will automatically be treated as python ints. The default python ints are 64bit on linux/mac and 32bit on windows.
So a default int cannot be larger than 2147483647 or smaller than -2147483648 if you want windows tests to pass

Hellio404: owww okay didn't know that. Good to know thank you

EmilyBourne:
> > > > 
> > > 
> > > 
> > > I did run the command `pylint --rcfile=.pylintrc tests/pyccel/scripts/print_integrals.py` and it's said `Your code has been rated at 10.00/10` so am a bit confused
> > 
> > 
> > Hmm, I'm not sure why that would happen. Currently it's just failing because the last character in the file isn't a new line character. This is something that some editors care about and others don't
> 
> Yeah but I did add a commit just to add that new line but it didn't got fixed weird

I presume you mean [this commit](https://github.com/pyccel/pyccel/pull/1181/commits/aa170653ddcfdd8d93aaec23379d71608395731d)

I don't see a new line here. Maybe your editor is cleaning up? What os and editor are you using? If you can't convince it to save with an extra empty line I can push this fix for you

Hellio404:
am using vscode on macOS, I literally clone the repo and run the linter on the file and I there is the new line also when you edit the code on Github it shows you the new line.

and about the large numbers should I remove those test or just wrap them on int64()? what do you think is best

tests/pyccel/scripts/print_integrals.py
EmilyBourne: > and about the large numbers should I remove those test or just wrap them on int64()? what do you think is best

Wrapping them in int64 sounds like a good idea

EmilyBourne:
> am using vscode on macOS, I literally clone the repo and run the linter on the file and I there is the new line also when you edit the code on Github it shows you the new line.

Looking on codacy it looks like it's passing. So I suspect that it somehow got mixed up about which error came from which commit

Hellio404:
I think there is another issue when it came to casting with numpy.int32 or numpy.int64 as I told you before that there is a difference between `numpy.int32(5)` and `numpy.int32(-5)` as the first one output ``5`` and the second one outputs (int32_t)(-5L) so on windows and with Fortran:
```python
int64(-2147483649)
```
will output
```fortran
Int(-2147483649_i32, i64)
```
the `-2147483649_i32` is not valid because -2147483649 is out of range for 32bit int
which doesn't compile so in windows with Fortran there is no way to express a negative number smaller than -2147483648 as it's not valid writing

EmilyBourne:
> I think there is another issue when it came to casting with numpy.int32 or numpy.int64 as I told you before that there is a difference between `numpy.int32(5)` and `numpy.int32(-5)` as the first one output `5` and the second one outputs (int32_t)(-5L) so on windows and with Fortran:
> 
> ```python
> int64(-2147483649)
> ```
> 
> will output
> 
> ```fortran
> Int(-2147483649_i32, i64)
> ```
> 
> the `-2147483649_i32` is not valid because -2147483649 is out of range for 32bit int which doesn't compile so in windows with Fortran there is no way to express a negative number smaller than -2147483648 as it's not valid writing

That makes sense. When casting literals we have a special case where we change the underlying precision of the literal, whereas for everything else we cast the generated values. We should probably make a negative literal into another special case. That should be done [here](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/builtins.py#L372-L373)

Hellio404:
> > I think there is another issue when it came to casting with numpy.int32 or numpy.int64 as I told you before that there is a difference between `numpy.int32(5)` and `numpy.int32(-5)` as the first one output `5` and the second one outputs (int32_t)(-5L) so on windows and with Fortran:
> > ```python
> > int64(-2147483649)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > will output
> > ```fortran
> > Int(-2147483649_i32, i64)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > the `-2147483649_i32` is not valid because -2147483649 is out of range for 32bit int which doesn't compile so in windows with Fortran there is no way to express a negative number smaller than -2147483648 as it's not valid writing
> 
> That makes sense. When casting literals we have a special case where we change the underlying precision of the literal, whereas for everything else we cast the generated values. We should probably make a negative literal into another special case. That should be done [here](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/builtins.py#L372-L373)

Yes I agree it could be something like this:
```py
elif isinstance(arg, PyccelUnarySub) and isinstance(arg.args[0], LiteralInteger):
            return PyccelUnarySub(LiteralInteger(arg.args[0].python_value, precision = cls._precision))
```

Hellio404:
Honestly, I don't like the solution of replacing exactly `PRId64` with `%"PRId64"`  but don't see a way to insert a macro especially since [the format goes into LiteralString constructor then _print](https://github.com/pyccel/pyccel/blob/df2571feedd16b848089cd8095818a7215e45a3c/pyccel/codegen/printing/ccode.py#L848) which will escape all the double qq in the string

EmilyBourne:
> Honestly, I don't like the solution of replacing exactly `PRId64` with `%"PRId64"` but don't see a way to insert a macro especially since [the format goes into `LiteralString` constructor then `_print`](https://github.com/pyccel/pyccel/blob/df2571feedd16b848089cd8095818a7215e45a3c/pyccel/codegen/printing/ccode.py#L848) which will escape all the double qq in the string

Hmm, I agree. Let me think about it. Maybe we could have a `CFormatString` class?

Hellio404:
I don't see where the `CFormatString` can fit.
we could also change the `_print_LiteralString` to support C macros (I don't know if FORTRAN has something similar) with something like "hello world __C_MACRO__(THE_MACRO)!!" to be converted to something like "hello world "THE_DEFINE"!!"
so we can use it internally for PRId64 as it can be defined like "%__C_MACRO(PRId64)"
this is just an idea not sure how it will hold

Hellio404:
here is a PR of what I was talking about '__C_MACRO__' thing

pyccel/ast/literals.py
```diff
@@ -212,6 +212,15 @@ def __repr__(self):
 
     def __str__(self):
         return str(self.python_value)
+    
+    def __add__(self, o):
+        from .internals import CStringExpression, CMacro
```
Hellio404, line 217: I did have a problem with circular import so I had to import those classes here.
I don't know if it's the right thing to do or if there is a better solution

EmilyBourne: I can't think of anything. This happens occasionally, it usually means the 2 objects should be defined in the same file, but I'm not sure that I would think to look for `CMacro` or `CStringExpression` in `literals.py`. This workaround is fine as long as it doesn't happen too often

pyccel/ast/internals.py
```diff
@@ -245,6 +245,103 @@ def code(self):
...
+    _attribute_nodes  = ()
+
+    def __init__(self, *args):
+        super().__init__()
+        self._expression = []
+        for arg in args:
+            self += arg
+
+    def __repr__(self):
+        return ''.join([str(e) for e in self._expression])
```
EmilyBourne, line 263: Should be `return ''.join(repr(e) for e in self._expression)` . See [here](https://www.geeksforgeeks.org/str-vs-repr-in-python/) for the difference between `str` and `repr`

pyccel/codegen/printing/ccode.py
```diff
@@ -551,6 +553,11 @@ def _print_PythonBool(self, expr):
         return '({} != 0)'.format(value)
 
     def _print_Literal(self, expr):
+        if isinstance(expr, LiteralInteger) and get_final_precision(expr) == 8:
+            self.add_import(c_imports['stdint'])
+            return f"INT64_C({repr(expr.python_value)})"
+        elif isinstance(expr, LiteralFloat) and get_final_precision(expr) == 4:
+            return f"{repr(expr.python_value)}f"
         return repr(expr.python_value)
```
EmilyBourne, line 561: I think it would be clearer and safer to break this into `_print_LiteralInteger` and `_print_LiteralFloat`

pyccel/ast/internals.py
```diff
@@ -245,6 +245,103 @@ def code(self):
         """
         return self._code
 
+#------------------------------------------------------------------------------
+class CStringExpression(Basic):
+    """
+    Internal class used to hold a C string that has LiteralStrings and C macros
```
EmilyBourne, line 251: Please add Parameters to this docstring. E.g. what types are acceptable for `args`?

pyccel/ast/literals.py
EmilyBourne: I think the better solution is to use the magic method `__radd__` to handle this addition in the `CStringExpression` and `CMacro` classes. See [here](https://blog.finxter.com/python-__rxadd__-magic-method/) for details

pyccel/ast/internals.py
```diff
@@ -245,6 +245,103 @@ def code(self):
...
+
+    def __repr__(self):
+        return str(self._macro)
+
+    def __str__(self):
+        return str(self._macro)
+
+    def __add__(self, o):
+        if isinstance(o, (LiteralString, CStringExpression)):
+            return CStringExpression(self, o)
```
EmilyBourne, line 335: Would it not be simpler to do:
```
if isinstance(o, LiteralString):
    return CStringExpression(self, o)
elif isinstance(o, CStringExpression):
    return CStringExpression(self, *o)
```
and then `CStringExpression` can only contain instances of `LiteralString` and `CMacro`. I believe you can do this with the `__iter__` magic method

Hellio404: the CStringExpression can also contain instances of  CStringExpression for example
for example, if we have CStringExpression that has an array like:
[LiteralString("1"),  CStringExpression(LiteralString("2"), CMacro("3")),  CMacro("4")]
it's not the same as:
[LiteralString("1"),  LiteralString("2"), CMacro("3"),  CMacro("4")]
because the join method will not have the same output

EmilyBourne: Ah ok. What should each of these output with the join method? I'm not sure I see what example you are thinking of

pyccel/ast/literals.py
Hellio404: in `A() + B()` the website said that `__radd__` is used only when `__add__` is not implemented in the class A.
but here I will have an implementation in the `LiteralString` because I want `LiteralString("a") + LiteralString("B")` to return a value of `LiteralString("aB") ` so it will always use the `__add__` of the `LiteralString` class

EmilyBourne: If you scroll down it says:
> Python first tries to call the left object’s `__add__()` method `x.__add__(y)`. But this may fail for two reasons:
> 1. The method `x.__add__()` is not implemented in the first place, or
> 2. The method `x.__add__()` is implemented but returns a NotImplemented value indicating that the data types are incompatible.
>
> If this fails, Python tries to fix it by calling the `y.__radd__()` for reverse addition on the right operand y.

pyccel/codegen/printing/ccode.py
```diff
@@ -553,6 +554,17 @@ def _print_PythonBool(self, expr):
     def _print_Literal(self, expr):
         return repr(expr.python_value)
 
+    def _print_LiteralInteger(self, expr):
+        if isinstance(expr, LiteralInteger) and get_final_precision(expr) == 8:
+            self.add_import(c_imports['stdint'])
+            return f"INT64_C({repr(expr.python_value)})"
+        return repr(expr.python_value)
```
EmilyBourne, line 561: Do we support `int8` or `int16`? They have numpy conversion functions, but I don't remember if we have a way to print them. I think there may be a related issue.

pyccel/ast/c_concepts.py
```diff
@@ -34,3 +35,149 @@ def obj(self):
         """The object whose address is of interest
         """
         return self._obj
+
+#------------------------------------------------------------------------------
+class CStringExpression(Basic):
+    """
+    Internal class used to hold a C string that has LiteralStrings and C macros
+    Parameters:
+        *args: str or LiteralString or CMacro or CStringExpression
```
EmilyBourne, line 44: Can an arg still be `CStringExpression`?

pyccel/codegen/printing/ccode.py
Hellio404: the support for `int16` is fine as the `int16` and `int8` literals don't need a prefix and when printing `int16` the format is correct (`%hd`) but I just notice that the `int8` is printed as char (`%c`) [here](https://github.com/pyccel/pyccel/blob/02aba7c92344a30e28206b6abea6f13efac148cb/pyccel/codegen/printing/ccode.py#L803) I wonder if there is any reason for it

Hellio404: normally it should be `%hhd`

EmilyBourne: Looking at [this discussion](https://stackoverflow.com/questions/13052880/format-specifier-of-int8-data-type-in-c) I don't know to what degree `%hhd` is specified in the standard, which may explain how we ended up with `%c`. As they say though, the safest is to always use `inttypes.h` to be sure that there are no problems on strange machines

Hellio404: okay if that's the case we can use `PRIi8` and `PRIi16` for `int8` and `int16` for more portability as we did support macros anyways.

Hellio404: in my macOS the `PRId8` expanded to `%hhd` but on my Linux machine it's just `%d`

pyccel/codegen/printing/ccode.py
```diff
@@ -798,7 +810,7 @@ def get_print_format_and_arg(self, var):
                           ('complex',8) : '(%.12lf + %.12lfj)',
                           ('complex',4) : '(%.12f + %.12fj)',
                           ('int',4)     : '%d',
-                          ('int',8)     : '%ld',
+                          ('int',8)     : LiteralString("%") + CMacro('PRId64'),
```
aihya, line 814: Can you explain why you used the macro `PRId64` instead of a regular `%ld`?
`PRId64` literally expands to `ld`.

pyccel/ast/c_concepts.py
```diff
@@ -34,3 +35,149 @@ def obj(self):
...
+            return result
+        result += lst[0]
+        for elm in lst[1:]:
+            result += self
+            result += elm
+        return result
+
+    def get_flat_expression_list(self):
+        """
+        returns an list of LiteralStrings and CMacros after merging every
```
aihya, line 126: typo: `returns a list of...`

aihya (changes requested):

pyccel/ast/c_concepts.py
Hellio404: done

EmilyBourne: [024460f](https://github.com/pyccel/pyccel/pull/1181/commits/024460f409713f343bebaf106e9b7399c5b1f7d0)

aihya (approved):

EmilyBourne (changes requested):

pyccel/ast/c_concepts.py
```diff
@@ -34,3 +35,149 @@ def obj(self):
...
+            CMacro("m"),
+            CStringExpression(
+                LiteralString("the macro is: "),
+                CMacro("mc")
+            ),
+            LiteralString("."),
+        )
+    """
+    __slots__  = ('_expression',)
+    _attribute_nodes  = ()
```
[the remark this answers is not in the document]
Hellio404: I don't exactly know what the `_attribute_nodes` do
is there documentation for it?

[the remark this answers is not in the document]
EmilyBourne: Unfortunately not yet. It stores all pyccel objects which are contained in the object. It is used in the functions [`get_user_nodes`](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/basic.py#L92-L107), [`get_attribute_nodes`](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/basic.py#L122-L137), [`is_attribute_of`](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/basic.py#L169-L181), [`is_user_of`](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/basic.py#L184-L198) and [`substitute`](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/basic.py#L221-L237) which all have docstrings.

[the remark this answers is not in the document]
Hellio404: from what I can see the _attribute_nodes make the `super.__init__()` [converts any iterable ](https://github.com/pyccel/pyccel/blob/3dcd437333ea618ff2081f5c9d437113078f63b1/pyccel/ast/basic.py#L51)(the list `_expression`) to a tuple, so the `.append` is not working anymore, do you thing it's a good idea to use something like: 
```python
self._expression += (o,)
```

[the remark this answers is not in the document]
EmilyBourne: For example you could use the following to determine if `#include <inttypes.h>` is needed:
```
int64_macro = CMacro("PRId64")
...
if expr.is_user_of(int64_macro):
```
but if `_attribute_nodes` is empty then it won't recognise that `expr` uses `int64_macro`

pyccel/codegen/printing/ccode.py
```diff
@@ -798,9 +811,9 @@ def get_print_format_and_arg(self, var):
...
                           ('complex',4) : '(%.12f + %.12fj)',
                           ('int',4)     : '%d',
-                          ('int',8)     : '%ld',
-                          ('int',2)     : '%hd',
-                          ('int',1)     : '%c',
+                          ('int',8)     : LiteralString("%") + CMacro('PRId64'),
+                          ('int',2)     : LiteralString("%") + CMacro('PRId16'),
+                          ('int',1)     : LiteralString("%") + CMacro('PRId8'),
                           ('bool',4)    : '%s',
                           ('string', 0) : '%s'}
```
EmilyBourne, line 818: Would this dictionary be better as a global object rather than recreating it at every call to `get_print_format_and_arg`?

pyccel/ast/c_concepts.py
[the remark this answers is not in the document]
Hellio404: > from what I can see the _attribute_nodes make the `super.__init__()` [converts any iterable ](https://github.com/pyccel/pyccel/blob/3dcd437333ea618ff2081f5c9d437113078f63b1/pyccel/ast/basic.py#L51)(the list `_expression`) to a tuple, so the `.append` is not working anymore, do you thing it's a good idea to use something like:
> 
> ```python
> self._expression += (o,)
> ```

I tried this and it works fine but is not sure if this is how it should be done. what do you think?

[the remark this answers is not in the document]
EmilyBourne: This is done for 2 reasons. Firstly to ensure that objects are compatible rather than sometimes being lists and sometimes being tuples, and secondly to remind the developer to take care of the inheritance.
When you do:
```
self._expression += (o,)
```
you should also do:
```
o.set_current_user_node(self)
```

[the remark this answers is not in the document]
EmilyBourne: > > from what I can see the _attribute_nodes make the `super.__init__()` [converts any iterable ](https://github.com/pyccel/pyccel/blob/3dcd437333ea618ff2081f5c9d437113078f63b1/pyccel/ast/basic.py#L51)(the list `_expression`) to a tuple, so the `.append` is not working anymore, do you thing it's a good idea to use something like:
> > ```python
> > self._expression += (o,)
> > ```
> 
> I tried this and it works fine but is not sure if this is how it should be done. what do you think?

I think this is related to [this discussion](https://github.com/pyccel/pyccel/pull/1181#discussion_r970025146)

pyccel/ast/c_concepts.py
```diff
@@ -34,3 +35,149 @@ def obj(self):
...
+        append the argument o to the end of the list _expression
+        Parameter:
+            o: str or LiteralString or CMacro or CStringExpression
+        """
+        if isinstance(o, str):
+            o = LiteralString(o)
+        if not isinstance(o, (LiteralString, CMacro, CStringExpression)):
+            raise TypeError(f"unsupported operand type(s) for +: '{self.__class__}' and '{type(o)}'")
+        self.expression.append(o)
+        return self
```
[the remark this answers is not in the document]
Hellio404: yes, I see the issue here. I made a change so `+` will have the expected behavior of returning a new instance and I added an append function that will be used if we need to change the self without creating new instance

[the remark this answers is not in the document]
EmilyBourne: You can also use `__iadd__` to define the `+=` operator if you want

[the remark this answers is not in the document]
Hellio404: done

pyccel/codegen/printing/ccode.py
Hellio404: yes we could do that, that would match the [Fortarn printer](https://github.com/pyccel/pyccel/blob/b1cd3459d87c18a5267c31d0833654724b6a92ce/pyccel/codegen/printing/fcode.py#L176-L183) too

pyccel/ast/c_concepts.py
```diff
@@ -55,38 +55,50 @@ class CStringExpression(Basic):
...
-    _attribute_nodes  = ()
+    _attribute_nodes  = ('_expression',)
 
     def __init__(self, *args):
-        super().__init__()
         self._expression = []
+        super().__init__()
         for arg in args:
-            self += arg
+            self.append(arg)
```
EmilyBourne, line 65: Wouldn't this do the same thing if it was written as:
```
def __init__(self, *args):
    self._expression = args
    super().__init__()
```
?

Hellio404: Using the first one makes sure that you can only add LiteralString, CMacro or CStringExpression as args

pyccel/ast/c_concepts.py
```diff
@@ -146,7 +158,7 @@ def get_flat_expression_list(self):
     def expression(self):
         """ The list containing the literal strings and c macros
         """
-        return self._expression
+        return tuple(self._expression)
```
EmilyBourne, line 165: Is this still needed? Isn't `self._expression` a tuple now?

Hellio404: no it's not needed anymore I will remove it

pyccel/codegen/printing/ccode.py
EmilyBourne: It should be faster

EmilyBourne (approved):

pyccel/ast/c_concepts.py
```diff
@@ -34,3 +35,162 @@ def obj(self):
...
+    Example:
+        CStringExpression(
+            CMacro("m"),
+            CStringExpression(
+                LiteralString("the macro is: "),
+                CMacro("mc")
+            ),
+            LiteralString("."),
+        )
+    """
```
yguclu, line 56: Please use the `numpydoc` format of docstrings, which we try to use throughout the library.

The same comment applies to multiple locations in this file.

pyccel/ast/c_concepts.py
```diff
@@ -34,3 +35,162 @@ def obj(self):
...
+            o: str or LiteralString or CMacro or CStringExpression
+        """
+        if isinstance(o, str):
+            o = LiteralString(o)
+        if not isinstance(o, (LiteralString, CMacro, CStringExpression)):
+            raise TypeError(f"unsupported operand type(s) for append: '{self.__class__}' and '{type(o)}'")
+        self._expression += (o,)
+        o.set_current_user_node(self)
+
+    def intersperse(self, lst):
```
yguclu, line 106: It is more "Pythonic" to call this function `join`, don't you think?

tests/pyccel/scripts/print_integrals.py
```diff
@@ -0,0 +1,37 @@
+# pylint: disable=missing-function-docstring, missing-module-docstring/
```
yguclu, line 1: Why is this module named `print_integrals`? Did you mean `print_integers`?

yguclu (changes requested):
Good job @Hellio404!

I see that you have added some new classes to `c_concepts.py`. Please mention it in the PR description

pyccel/ast/c_concepts.py
Hellio404: I tried to follow the examples from other part of the code, if there is anything else about the doc format please let me know

yguclu: Just look at [this file](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/numpyext.py).

The documentation about the `numpydoc` format is [here](https://numpydoc.readthedocs.io/en/latest/format.html).

yguclu (approved):
Thanks for the modifications!

PR #1186, opened 2022-08-20
handle the unnamed function arguments 

azouiten:
trying to create temporary variables for the unnamed functions passed to functions so they are allocated properly if needed.


pyccel/parser/semantic.py
```diff
@@ -58,6 +58,7 @@
 from pyccel.ast.core import Decorator
 from pyccel.ast.core import PyccelFunctionDef
 from pyccel.ast.core import Assert
+from pyccel.ast.core import Declare
```
EmilyBourne, line 61: Declarations are added automatically for any variables which have been added to the scope. See https://github.com/pyccel/pyccel/blob/master/developer_docs/scope.md

pyccel/parser/semantic.py
```diff
@@ -1423,6 +1424,38 @@ def _assign_GeneratorComprehension(self, lhs_name, expr, **settings):
         expr_new.set_fst(expr.fst)
         return expr_new
 
+    def _handle_unamed_var(self, code_block, scope):
+        program_body = [line for line in code_block.body]
+        body = [item for item in program_body]
+        new_code_block = False
+        for ind, line in enumerate(body):
+                if isinstance(line, FunctionCall) or (isinstance(line, Assign) and isinstance(line.rhs, FunctionCall)):
```
EmilyBourne, line 1432: The `_additional_exprs` variable should allow you to call this function directly from `_visit_FunctionCall`

pyccel/parser/semantic.py
```diff
@@ -1423,6 +1424,38 @@ def _assign_GeneratorComprehension(self, lhs_name, expr, **settings):
...
+                    arg_changed = False
+                    fun_call = line if isinstance(line, FunctionCall) else line.rhs
+                    args = [arg for arg in fun_call.args]
+                    for arg_ind, arg in enumerate(args):
+                        if isinstance(arg.value, (PythonTuple, PythonList, NumpyArray)):
+                            print(type(arg.value), hasattr(arg.value, "dtype"))
+                            temp_var = scope.get_temporary_variable(dtype=arg.value.dtype, memory_handling='heap', \
+                                shape=arg.value.shape, order=arg.value.order, rank=arg.value.rank)
+                            program_body.insert(program_body.index(line, ind), Declare(temp_var.dtype, temp_var))
```
EmilyBourne, line 1441: The declaration is added in the printing stage. You don't need this

Hellio404:
As the Fortran already supports the unnamed array as function arguments what do you think if we can just create a tmp variable in the C printer something like [this](https://github.com/pyccel/pyccel/commit/880cd24295ed437dea2815986d720a1d34fb2346)

EmilyBourne:
What is the status of this PR? Has it been forgotten?

azouiten:
> What is the status of this PR? Has it been forgotten?

I haven't worked on it for a while but I ll definitely get back to it and finish it asap.

azouiten:
> As the Fortran already supports the unnamed array as function arguments what do you think if we can just create a tmp variable in the C printer something like [this](https://github.com/pyccel/pyccel/commit/880cd24295ed437dea2815986d720a1d34fb2346)

that does make sense, but I still don't think it should be solved from the printer, at the very least it would hurt the readability of the code as its something that you wouldn't expect to be in a printer , not sure how to do it tho

EmilyBourne:
> > As the Fortran already supports the unnamed array as function arguments what do you think if we can just create a tmp variable in the C printer something like [this](https://github.com/pyccel/pyccel/commit/880cd24295ed437dea2815986d720a1d34fb2346)
> 
> that does make sense, but I still don't think it should be solved from the printer, at the very least it would hurt the readability of the code as its something that you wouldn't expect to be in a printer , not sure how to do it tho

There is an ongoing debate in pyccel about adding a language-specific semantic layer to do this kind of thing. Currently any language specific things are done in the printer as we'd rather have less-readable pyccel code than less readable generated code (as is the case if you create temporaries in the semantic stage as they are then in the generated code even when they aren't necessary)

EmilyBourne:
@azouiten are you working on this at all? If not then can we close this PR? I'm not convinced that the code in this PR would be used if someone else decided to work on the issue.

Issue #1135, opened 2022-06-22
Precision calculation for mixed types is wrong


EmilyBourne:
We could have a complex `x` with low precision, and a float `y` with very high precision. With our current choice we lose precision. Here is an example:

```python
In [1]: import numpy as np

In [2]: a = np.complex64(3)

In [3]: a
Out[3]: (3+0j)

In [4]: b = np.float128(4)

In [5]: b
Out[5]: 4.0

In [6]: c = a + b

In [7]: c.dtype
Out[7]: dtype('complex256')

In [8]: c
Out[8]: (7+0j)
```

_Originally posted by @yguclu in https://github.com/pyccel/pyccel/pull/858#discussion_r903103628_

The following code reproduces the error:
```python
import numpy as np

a = np.complex64(3)
print(type(a))
b = np.float64(4)
c = a + b
print(type(c))
```

Python output:
```
<class 'numpy.complex64'>
<class 'numpy.complex128'>
```

Pyccelised output (see also #1134):
```
<class 'complex32'>
<class 'complex32'>
```

**Additional Information**
- This is due to the way precision is calculated in the [operators file](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/operators.py). A similar problem exists in the [numpyext file](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/numpyext.py)

EmilyBourne:
Fixed by #1187

Issue #1105, opened 2022-05-02
Operations on numpy arrays with python types return the wrong type


EmilyBourne:
**Describe the bug**
Arithmetic operations between a numpy object and a pure python object return objects with the wrong type. The numpy type should be prioritised

**To Reproduce**
Provide code to reproduce the behavior:
```python
import numpy as np

def get_int_array():
    a = np.ones(5, dtype=np.int32)
    b = a*2
    return b[0]
```
The type of the result should be `numpy.int32`

Provide the generated code, or the error message:
```fortran
  !........................................
  subroutine get_int_array()

    implicit none

    integer(i32), allocatable :: a(:)
    integer(i64), allocatable :: b(:)

    allocate(a(0:4_i64))
    a = 1_i32
    allocate(b(0:4_i64))
    b = a * 2_i64
    if (allocated(a)) then
      deallocate(a)
    end if
    if (allocated(b)) then
      deallocate(b)
    end if

  end subroutine get_int_array
  !........................................
```

**Expected behavior**
The numpy type should be prioritised

EmilyBourne:
The type is wrong because the precision is set wrong. The precision is selected in the `_calculate_dtype` functions in `pyccel/ast/operators.py`

Pinkyboi:
@bauom  I checked the `_calculate_dtype` functions and they seems like the problem comes from `max_precision` choosing the the type with the maximum precision, but since at this point of the execution both the dtypes for the numpy object and the pure python types are `NativeIntiger` in this case i was wondering if there's a way to know if one of them is a numpy type to prioritize it.

EmilyBourne:
> @bauom I checked the `_calculate_dtype` functions and they seems like the problem comes from `max_precision` choosing the the type with the maximum precision, but since at this point of the execution both the dtypes for the numpy object and the pure python types are `NativeIntiger` in this case i was wondering if there's a way to know if one of them is a numpy type to prioritize it.

Pure python types have default precision (-1), while numpy types have fixed precision (usually 4 or 8)

EmilyBourne:
@Pinkyboi I think this issue and #1135 are quite similar. You may want to fix both together to avoid making changes which turn out not to work in the more general case considered in #1135

Pinkyboi:
@EmilyBourne roger that ! ill look into both issues.

PR #1187, opened 2022-08-26
Fix arithmetic operations issues for Numpy arrays.

Pinkyboi:
Fixes: [#1105](https://github.com/pyccel/pyccel/issues/1105) and [#1135](https://github.com/pyccel/pyccel/issues/1135)

- Make Pyccel prioritize Numpy arrays types in arithmetic operations.
- Avoid precision loss when performing binary operations on variables having different types.
- Create temporary variable holding the value of arithmetic operations on array for function arguments.
- Fix name clash error occurring when doing an arithmetic operation between an array and function's return.


EmilyBourne:
> @EmilyBourne i think what i changed could fix both issues, but i wanted to know your opinion on the case of an `AugAssign` for example, how do you think that should be handled ?
> 
> Lets take for example one of the functions that fails during the tests:
> 
> ```python
> @types('int', 'int', 'int')
> @types('int8', 'int', 'int')
> @types('int16', 'int', 'int')
> @types('int32', 'int', 'int')
> @types('int64', 'int', 'int')
> @types('float', 'int', 'int')
> @types('float32', 'int', 'int')
> @types('float64', 'int', 'int')
> def get_linspace(start, steps, num):
>     stop = start + steps
>     b = linspace(start, stop, num)
>     x = 0.0
>     for bi in b:
>         x += bi
>     return x
> ```
> 
> We will get this error message:
> 
> ```
> ERROR at annotation (semantic) stage
> pyccel:
>  |error [semantic]: main.py [18,8]| Variable has already been defined with type 'float'. Type 'float32' in assignment is incompatible (x=bi)
> ```
> 
> That comes from one of my changes that prioritize the numpy type precision over the pure python type precision, is there a way to know if the lhs is already a defined variable so we can adapt the used precision to it instead of calculating it ?

If I run a quick test then I get:
```
In [1]: import numpy as np                                                      

In [2]: a = np.float32(4.5)                                                     

In [3]: b = 0.0                                                                 

In [4]: type(b)                                                                 
Out[4]: float

In [5]: type(a)                                                                 
Out[5]: numpy.float32

In [6]: b+= a                                                                   

In [7]: type(b)                                                                 
Out[7]: numpy.float64
```

So the error sounds right. I think it is the test that needs correcting. E.g:
```
def get_linspace(start, steps, num):
    stop = start + steps
    b = linspace(start, stop, num)
    x = np.float64(0.0)
    for bi in b:
        x += bi
    return x
```

Pinkyboi:
@EmilyBourne  That's what i thought too, ill change those tests.

tests/errors/semantic/non_blocking/INCOMPATIBLE_TYPES_IN_ASSIGNMENT_2.py
```diff
@@ -2,4 +2,4 @@
 import numpy as np
 
 a = np.float32(4)
-a += 5.0
+a = 5.0
```
EmilyBourne, line 5: Can you explain this change

tests/epyccel/test_arrays.py
```diff
@@ -3991,12 +3994,12 @@ def test_numpy_arange_with_dtype(language):
...
     f1 = arrays.arr_arange_5
     f2 = epyccel(f1, language = language)
-    np.testing.assert_array_almost_equal(f1(), f2(), decimal = 9)
+    np.testing.assert_allclose(f1(), f2())
 
 def test_numpy_arange_negative_step_2(language):
     f1 = arrays.arr_arange_6
     f2 = epyccel(f1, language = language)
-    np.testing.assert_array_almost_equal(f1(), f2(), decimal = 9)
+    np.testing.assert_allclose(f1(), f2())
```
EmilyBourne, line 4002: You should also check the return type (maybe this explains your precision problems). If the type is wrong then this problem might be linked to an issue caught by @aihya and fixed in PR #1169

Pinkyboi: If i understand correctly what @aihya explained to me is that the error could occur when converting an `np.array` to a python array the precision may go down in some cases, but i'm still not entirely sure about the implications of this.

Pinkyboi:
@EmilyBourne It seems like the you are right the precision become lower in windows, should add the fix that @aihya added directly to my branche or should i leave it as it is ?

EmilyBourne:
> @EmilyBourne It seems like the you are right the precision become lower in windows, should add the fix that @aihya added directly to my branche or should i leave it as it is ?

It's up to you and depends on how comfortable you are with git merges.

If you are comfortable then I would apply the fix locally to check that the problem is caused by this bug and not by pyccel assigning the wrong precision internally.

If you are not comfortable, then maybe wait until @yguclu has merged #1169

Pinkyboi:
@EmilyBourne before adding a valid test for #1105  that would check if we prioritized the `numpy` type over the pure python type, i wanted to be sure that this behavior is pyccel specific since this implementation doesn't mimic the python implementation.
In case of the python implementation the the precision would get promoted if the initial precision is lower than the pure python type precision.

EmilyBourne:
> @EmilyBourne before adding a valid test for #1105 that would check if we prioritized the `numpy` type over the pure python type, i wanted to be sure that this behavior is pyccel specific since this implementation doesn't mimic the python implementation. In case of the python implementation the the precision would get promoted if the initial precision is lower than the pure python type precision.

What exactly doesn't mimic the python implementation? The example provided in #1105?
What version of python do you have? On my computer (python 3.8.10) the precision does not get promoted:
```python
In [2]: import numpy as np 
   ...:  
   ...: def get_int_array(): 
   ...:     a = np.ones(5, dtype=np.int32) 
   ...:     b = a*2 
   ...:     return b[0] 
   ...:                                                                         

In [3]: type(get_int_array())                                                   
Out[3]: numpy.int32
```

Pinkyboi:
@EmilyBourne I just got confused since the treatment on the scalar precision and array precision isn't the same.

pyccel/ast/operators.py
```diff
@@ -393,11 +394,11 @@ def _calculate_dtype(cls, *args):
             return cls._handle_str_type(strs)
             assert len(integers + floats + complexes) == 0
```
ohachim, line 395: Will it ever pass to this line?

pyccel/ast/operators.py
```diff
@@ -384,6 +384,7 @@ def _calculate_dtype(cls, *args):
         e.g.
             1 + 2j -> PyccelAdd(LiteralInteger, LiteralComplex) -> complex
         """
+
         integers  = [a for a in args if a.dtype in (NativeInteger(),NativeBool())]
         floats    = [a for a in args if a.dtype is NativeFloat()]
```
ohachim, line 389: Are the list comprehension necessary, it doesn't look like they'll be used later, wouldn't it be better to just use `args[0].dtype`?

ohachim (changes requested):

pyccel/ast/operators.py
```diff
@@ -384,6 +384,7 @@ def _calculate_dtype(cls, *args):
         e.g.
             1 + 2j -> PyccelAdd(LiteralInteger, LiteralComplex) -> complex
         """
+
         integers  = [a for a in args if a.dtype in (NativeInteger(),NativeBool())]
```
azouiten, line 388: new line?

pyccel/ast/internals.py
```diff
@@ -292,8 +293,10 @@ def max_precision(objs : list, dtype = None, allow_native = True):
         return max(def_prec if o.precision == -1 \
                 else o.precision for o in objs if o.dtype is dtype)
     else:
-        return max(default_precision[str(o.dtype)] if o.precision == -1 \
-                else o.precision for o in objs)
+        if any(getattr(o, 'is_ndarray', False) for o in objs):
+            return get_final_precision(max(objs, key=attrgetter('precision')))
+        else:
+            return max(get_final_precision(o) for o in objs)
```
azouiten, line 299: I think these two statments are the same, so maybe you should just remove the if else?

tests/epyccel/recognised_functions/test_numpy_types.py
```diff
@@ -9,9 +9,121 @@
...
 from test_numpy_funcs import matching_types
 
-from pyccel.decorators import types
+from pyccel.decorators import types, template
 from pyccel.epyccel import epyccel
 
+def test_mult_numpy_python_type(language):
+
+    def mult_on_array_int8():
+        from numpy import ones, int8
```
azouiten, line 18: is there a reason why you import numpy in every function def?, I ve noticed that most tests have the same import statments.

tests/macro/scripts/MPI/gather.py
```diff
@@ -16,17 +16,17 @@
     block_length = nb_values / size
 
     # ...
-    values = [0]*block_length
+    values = [0] * block_length
```
azouiten, line 19: I remember emily once said that we should not make norm changes.

azouiten (dismissed):

tests/macro/scripts/MPI/gather.py
EmilyBourne: It's not that you shouldn't make norm changes. It's just that it makes it harder to review a PR when there are things in it that aren't relevant. It gives the impression that this PR changes something in MPI when it doesn't.
It's mostly courtesy. There's no need for all 3 reviewers to double check a file that didn't really change.

If we want to make changes to eg normalise the spacing conventions then we could create a PR for this and add another linter to keep it clean but this isn't a priority as the readability isn't very different and it doesn't affect the execution

EmilyBourne:
@Pinkyboi I see that you have re-requested reviews from @ohachim and @azouiten , but you don't seem to have addressed most of their comments. Please leave a message on open comments either linking to the commit where you changed something or explaining why the change is unnecessary/can't be done.

Pinkyboi:
@EmilyBourne it seems like i forgot to submit my comments, i didn't notice the the pending tag on my comments.

pyccel/ast/operators.py
EmilyBourne: @ohachim good catch. I suspect that's my error originally.
@Pinkyboi would you mind changing the order of these lines

pyccel/ast/operators.py
EmilyBourne: For example `3 + 4.0` will lead to a float with the current implementation, but would give an integer with your suggestion

tests/epyccel/test_arrays.py
EmilyBourne: @Pinkyboi Did you update the tests?

pyccel/ast/operators.py
Pinkyboi: @EmilyBourne wouldn't it be better to leave just the `TypeError` raised by  `_handle_str_type` ? in my opinion it's more descriptive of the error than just an `assert` although the error message `"unsupported operand type(s) for /: 'str' and 'str'"` could maybe be changed.

EmilyBourne: Ah if there is a `TypeError` then yes that is much better. In that case the `assert` can indeed be removed.
> `"unsupported operand type(s) for /: 'str' and 'str'"` could maybe be changed.

This doesn't seem like it's reporting the same error though, so we might need to add an additional `TypeError`. This is talking about doing `'ha'/'out'` which doesn't mean anything in python. Some of the other operators are supported but I think in that case they have their own classes which are not operators (e.g `Concatenate`).

tests/epyccel/test_arrays.py
Pinkyboi: @EmilyBourne i did revert back some changes there since i noticed that the problem was caused by my implementation, i left the test on the return type cuz i thought i may be useful still.

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5271,8 +5271,8 @@ def test_numpy_where_array_like_1d_with_condition(language):
     @types('float32[:]')
     @types('float64[:]')
     def get_chosen_elements(arr):
-        from numpy import where, shape
-        a = where(arr > 5, arr, arr*2)
+        from numpy import where, shape, float64
+        a = where(arr > 5, arr, arr * float64(2))
```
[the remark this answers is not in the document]
Pinkyboi: @EmilyBourne if i remember correctly there was a problem with fortran when doing binary arithmetic operations on the fly,  for `fortran`, `merge()` if we replaced the test with something like : 
```python
    def get_chosen_elements(arr):
        from numpy import where, shape
        
        arr1 = arr + 2
        a = where(arr > 5, arr, arr1)
        s = shape(a)
        return len(s), s[0], a[1], a[0]
```  
I wasn't sure if this is within the scope of my issue to be honest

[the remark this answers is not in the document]
EmilyBourne: Either a temporary can be created for this, or you should use the cast functions if they handle arrays

ohachim (approved):

pyccel/parser/semantic.py
```diff
@@ -707,6 +706,13 @@ def _handle_function_args(self, arguments, **settings):
             if isinstance(a.value, StarredArguments):
                 args.extend([FunctionCallArgument(av) for av in a.value.args_var])
             else:
+                if isinstance(a.value, PyccelArithmeticOperator):
+                    d_var = self._infere_type(a.value)
+                    d_var.pop('datatype')
+                    var = self.scope.get_temporary_variable(a.value.dtype, **d_var)
+                    assi = Assign(var, a.value)
+                    self._additional_exprs[-1].append(assi)
+                    a = self._visit(FunctionCallArgument(var), **settings)
```
EmilyBourne, line 715: Do you need to visit `FunctionCallArgument(var)`? The `var` object has a type/precision/rank/etc so doesn't it already have all necessary semantic behaviour?

pyccel/parser/semantic.py
```diff
@@ -1392,10 +1392,14 @@ def _assign_GeneratorComprehension(self, lhs_name, expr, **settings):
...
         # as the central expression
-        loops  = [self._visit(expr.loops, **settings)]
 
         # If necessary add additional expressions corresponding
         # to nested GeneratorComprehensions
+        loops = [self._visit(expr.loops, **settings)]
+        add_exprs  = [*self._additional_exprs[-1]]
+        loops = add_exprs + loops
+        self._additional_exprs.pop()
+        self._additional_exprs.append([])
```
EmilyBourne, line 1402: `self._additional_exprs[-1]` is already a list so upacking it into a list is pointless.
```python
add_exprs  = [*self._additional_exprs[-1]]
```
is equivalent to:
```python
add_exprs  = list(self._additional_exprs[-1])
```
or
```python
add_exprs  = self._additional_exprs[-1].copy()
```
But there is no need to make a copy here.

Also `pop()` returns the element that it is removing so this section could be written much more compactly as:
```python
loops = self._additional_exprs.pop() + loops
self._additional_exprs.append([])
```

Pinkyboi: Sure that looks cleaner, I'll fix that.

aihya (approved):

yguclu:
@aihya Please answer all the comments made by @EmilyBourne, pointing out which commit applies the requested change.

pyccel/parser/semantic.py
Pinkyboi: I ended up dropping the insertion in the loop list in [b26a937](https://github.com/pyccel/pyccel/pull/1187/commits/b26a937f05f32cee939135004c3977b37f9b462f).

pyccel/parser/semantic.py
Pinkyboi: i removed the visit to `FunctionCallArgument`  in [51fe564](https://github.com/pyccel/pyccel/pull/1187/commits/51fe56464f3b9dcc1d309dcd559434a087da4b05).

Pinkyboi:
@yguclu Really sorry for that, i added comments pointing in which commit the changes occurred.

tests/epyccel/recognised_functions/test_numpy_funcs.py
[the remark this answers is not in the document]
yguclu: @Pinkyboi Has this comment been addressed?

pyccel/ast/internals.py
```diff
@@ -292,8 +293,10 @@ def max_precision(objs : list, dtype = None, allow_native = True):
         return max(def_prec if o.precision == -1 \
                 else o.precision for o in objs if o.dtype is dtype)
     else:
-        return max(default_precision[str(o.dtype)] if o.precision == -1 \
-                else o.precision for o in objs)
+        ndarray_list = [o for o in objs if getattr(o, 'is_ndarray', False)]
+        if ndarray_list:
+            return get_final_precision(max(ndarray_list, key=attrgetter('precision')))
```
EmilyBourne, line 298: Is there a reason to prefer `attrgetter` to ` get_final_precision(max(ndarray_list, key=lambda l: l.precision))`?

pyccel/parser/semantic.py
```diff
@@ -708,6 +706,11 @@ def _handle_function_args(self, arguments, **settings):
             if isinstance(a.value, StarredArguments):
                 args.extend([FunctionCallArgument(av) for av in a.value.args_var])
             else:
+                if isinstance(a.value, PyccelArithmeticOperator) and a.value.rank:
+                    tmp_var = PyccelSymbol(self.scope.get_new_name(), is_temp=True)
+                    assign = self._visit(Assign(tmp_var, arg.value, fst= arg.value.fst))
+                    self._additional_exprs[-1].append(assign)
+                    a = FunctionCallArgument(self._visit(tmp_var))
```
EmilyBourne, line 713: Please add a test for this case

EmilyBourne (changes requested):

pyccel/ast/internals.py
Pinkyboi: I guess using the lambda function improves the readability a bit, i'll switch to that i guess.

EmilyBourne: I don't know if it's more readable, but it removes an unnecessary import

EmilyBourne: Actually it looks like your choice is faster: https://stackoverflow.com/a/54061395/5957195 :+1:

Pinkyboi: Alright i'll leave it as it is.

tests/epyccel/recognised_functions/test_numpy_funcs.py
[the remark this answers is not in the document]
Pinkyboi: @yguclu since all the commits after this message aim to fix that issue, i wasn't sure if i was supposed to address it in any way.

pyccel/parser/semantic.py
Pinkyboi: I added some tests in [2158507](https://github.com/pyccel/pyccel/pull/1187/commits/215850701c290467e536c5c1bae4755cc9b1c38a), is it alright to add the tests in scripts instead of epyccel, because i wanted to add tests for arithmetic operations with functions return.

EmilyBourne: Tests in the `tests/pyccel/` folder are fine, but you need to add them to the `test_pyccel.py` file in order for them to be executed. I don't see any tests testing this though. Is it not supposed to handle things like:
```
f(a+b)
```
?

Pinkyboi: Oh i guess i didnt add em, i was testing em in already defined functions like in `np.array` is it fine ?

Pinkyboi: @EmilyBourne i did add a test with an arithmetic add in [fdf0609](https://github.com/pyccel/pyccel/pull/1187/commits/fdf06091e95635660e96f393a40e239bf52e2805).

EmilyBourne: > Oh i guess i didnt add em, i was testing em in already defined functions like in np.array is it fine ?

I don't see this test. Can you point me to the line

EmilyBourne: Never mind I found it. For some reason I thought this code only applied to user defined functions

tests/epyccel/recognised_functions/test_numpy_types.py
```diff
@@ -9,9 +9,121 @@
...
+
+    epyccel_func = epyccel(mult_on_array_float32, language=language)
+    python_result = mult_on_array_float32()
+    pyccel_result = epyccel_func()
+    assert matching_types(pyccel_result, python_result)
+
+    epyccel_func = epyccel(mult_on_array_float64, language=language)
+    python_result = mult_on_array_float64()
+    pyccel_result = epyccel_func()
+    assert matching_types(pyccel_result, python_result)
```
EmilyBourne, line 81: It seems a shame not to also test the result in this test function

Pinkyboi: Surely, i will add that.

Pinkyboi: Done in [931707b](https://github.com/pyccel/pyccel/pull/1187/commits/931707bcf90417a380050074300fa42212e35785).

EmilyBourne (approved):

yguclu (approved):

PR #1188, opened 2022-09-12
Add support for sys.exit

### LINKED ISSUE #931: Add sys.exit to pyccel
**Describe the bug**
`sys.exit` does not work. It would be nice to have it

**Test Code**
```python
from sys import exit

exit(1)
```

Hellio404:
fix #931: adding support for sys.exit

List of modifications:

- support printing to the `stderr`
- `SysExit` class was added to represent a call to `sys.exit`
- adding support for `exit` in 3 different ways
  1. `exit(Integer)` ==> will exit the program with the `Integer` given in the argument
  2. `exit()` ==> will exit the program with 0
  3. `exit(Obj)` ==> will print the `Obj` to the `stderr` and exit with 1


pyccel/ast/sysext.py
```diff
@@ -0,0 +1,55 @@
...
+from .core import PyccelFunctionDef, Module
+from .internals import PyccelInternalFunction
+from .datatypes import NativeVoid
+from .internals import LiteralInteger
+from .operators import PyccelUnarySub
+
+class SysExit(PyccelInternalFunction):
+    """Represents a call to  sys.exit
+
+    arg : LiteralInteger, PyccelUnarySub(LiteralInteger)
```
rakati, line 18: as in [python docs for sys.exit](https://docs.python.org/3/library/sys.html#sys.exit) there are two stats:
- Success exit: we may have `exit( )` or `exit(0)`
- Error exit: `exit(non-zero-int)` or `exit(msg-error)`
I wonder if you can add the support for the case of empty arg and msg error?

Hellio404: I already added the support for empty calls but with the exit(object) we don't have support for printing to the standard error

rakati (commented):
great work Youness, I suggest adding tests to make sure that everything works well, especially testing exit within OpenMP too.

Hellio404:
I couldn't figure out how to test the exit error from Pytest I did try to use [this](https://stackoverflow.com/questions/62427205/in-python-3-using-pytest-how-do-we-test-for-exit-code-exit1-and-exit0-fo) but it doesn't catch the exit for the compiled version and the test just exits

pyccel/ast/sysext.py
rakati: > I couldn't figure out how to test the exit error from Pytest I did try to use [this](https://stackoverflow.com/questions/62427205/in-python-3-using-pytest-how-do-we-test-for-exit-code-exit1-and-exit0-fo) but it doesn't catch the exit for the compiled version and the test just exits

we can fairly see that the python exit function and C exit function have different behavior, I am confused about these two options: keep it simple like what you did, or use `PyExc_SystemExit` from python C Api.
@EmilyBourne what do you think?

rakati:
> I couldn't figure out how to test the exit error from Pytest I did try to use [this](https://stackoverflow.com/questions/62427205/in-python-3-using-pytest-how-do-we-test-for-exit-code-exit1-and-exit0-fo) but it doesn't catch the exit for the compiled version and the test just exits

we can fairly see that the python exit function and C exit function have different behavior, I am confused about these two options: keep it simple like what you did, or use PyExc_SystemExit from python C Api.
@EmilyBourne what do you think?

pyccel/ast/sysext.py
rakati: hmm, maybe we can use our c low-level function `write`, for writing to stderr, we will need just to import the libraries required to do that.
what do you think?

Hellio404: I don't think it's a good idea to use write as it's hard to format the output.
we can use `fprintf` for that and use the same already written _print_PythonPrint to generate the code for it.
but I couldn't figure out how to write to `stderr` on fortran, need someone familiar with it to add that

EmilyBourne:
> we can fairly see that the python exit function and C exit function have different behavior, I am confused about these two options: keep it simple like what you did, or use PyExc_SystemExit from python C Api.

I think it should just crash python. If it can print an error first that would be great. Otherwise we would have to rewrite all of pyccel so that functions always return bools indicating if it passed or not. That would not be efficient. I just want a way to make my code crash if I do something stupid

EmilyBourne:
> I couldn't figure out how to test the exit error from Pytest I did try to use [this](https://stackoverflow.com/questions/62427205/in-python-3-using-pytest-how-do-we-test-for-exit-code-exit1-and-exit0-fo) but it doesn't catch the exit for the compiled version and the test just exits

This should be tested in tests/pyccel in a similar way to the assert tests

EmilyBourne:
> > I couldn't figure out how to test the exit error from Pytest I did try to use [this](https://stackoverflow.com/questions/62427205/in-python-3-using-pytest-how-do-we-test-for-exit-code-exit1-and-exit0-fo) but it doesn't catch the exit for the compiled version and the test just exits
> 
> we can fairly see that the python exit function and C exit function have different behavior, I am confused about these two options: keep it simple like what you did, or use PyExc_SystemExit from python C Api. @EmilyBourne what do you think?

I would keep it simple, similarly to what we did for `assert`. I would even suggest only allowing `exit` statements in `if __name__ == '__main__'` blocks

pyccel/ast/sysext.py
EmilyBourne: https://stackoverflow.com/a/8508757/5957195

pyccel/codegen/printing/fcode.py
```diff
@@ -620,25 +626,25 @@ def _print_PythonPrint(self, expr):
...
 
-                for_body = [PythonPrint(print_body)]
+                for_body = [PythonPrint(print_body, output_unit=expr.output_unit)]
                 for_loop = For(for_index, for_range, for_body, scope=loop_scope)
                 for_end_char = LiteralString(']')
                 for_end = FunctionCallArgument(for_end_char,
                                                keyword='end')
 
-                body = CodeBlock([PythonPrint([FunctionCallArgument(LiteralString('[')), empty_end]),
+                body = CodeBlock([PythonPrint([FunctionCallArgument(LiteralString('[')), empty_end], output_unit=expr.output_unit),
```
rakati, line 635: I think @EmilyBourne will not be happy with this line because it's too long, as she said once if the line doesn't fits it screen you have to break the line.

pyccel/codegen/printing/fcode.py
```diff
@@ -220,12 +220,16 @@ def __init__(self, filename, prefix_module = None):
...
     def print_constant_imports(self):
         """Prints the use line for the constant imports used"""
-        macro = "use, intrinsic :: ISO_C_Binding, only : "
-        rename = [c if isinstance(c, str) else c[0] + ' => ' + c[1] for c in self._constantImports]
-        if len(rename) == 0:
-            return ''
-        macro += " , ".join(rename)
-        return macro
+        macros = []
+        for key in self._constantImports:
```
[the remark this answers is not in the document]
Hellio404: done

rakati (approved):
Well done Youness, for future PRs, if you faced a situation that required adding new feature (here printing to stderr) try to create an issue for it, mark the current PR as blocked by the issue you created, then solve the new issue in a sperate PR.

pyccel/ast/sysext.py
```diff
@@ -0,0 +1,57 @@
...
+# This file is part of Pyccel which is released under MIT License. See the LICENSE file or #
+# go to https://github.com/pyccel/pyccel/blob/master/LICENSE for full license details.     #
+#------------------------------------------------------------------------------------------#
+""" Module containing objects from the sys module understood by pyccel
+"""
+from .core import PyccelFunctionDef, Module
+from .internals import PyccelInternalFunction
+from .datatypes import NativeVoid
+from .internals import LiteralInteger
+
```
EmilyBourne, line 13: Missing `__all__`

pyccel/ast/builtins.py
```diff
@@ -588,7 +588,8 @@ class PythonPrint(Basic):
 
     expr : PyccelAstNode
         The expression to print
-
+    output_unit: String (Optional)
+        Select 'stdout' (default) or 'stderr' to print to
```
EmilyBourne, line 592: I think this argument should be called `file` to mimic python's behaviour

pyccel/codegen/printing/fcode.py
```diff
@@ -670,8 +680,11 @@ def _formatted_args_to_print(self, fargs_format, fargs, fend, fsep):
 
         args_code       = ' , '.join(args_list)
         args_formatting = ' '.join(fargs_format)
-        return "write(*, '({})', advance=\"{}\") {}\n"\
-            .format(args_formatting, advance, args_code)
+        if expr.output_unit == "stderr":
+            self._constantImports.setdefault('ISO_FORTRAN_ENV', set())\
+                .add(("stderr", "error_unit"))
+            return f"write(stderr, '({args_formatting})', advance=\"{advance}\") {args_code}\n"
+        return f"write(*, '({args_formatting})', advance=\"{advance}\") {args_code}\n"
```
EmilyBourne, line 687: Is there not an equivalent `stdout` instead of the default `*`?

pyccel/ast/sysext.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+    __slots__ = ()
+    name      = 'exit'
+    _dtype     = NativeVoid()
+    _precision = -1
+    _rank      = 0
+    _shape     = None
+    _order     = None
+
+    def __init__(self, arg=None):
```
EmilyBourne, line 32: I think this argument should be called `status`. Does this work if you translate:
```python
x = sys.exit(status=2)
```
?

tests/epyccel/utilities.py
```diff
@@ -2,7 +2,6 @@
 import numpy as np
 from pyccel.epyccel import epyccel
 
-
```
EmilyBourne, line 5: Please try to avoid committing "changes" to files which you haven't modified

pyccel/codegen/printing/fcode.py
```diff
@@ -2645,6 +2658,19 @@ def _print_ConstructorCall(self, expr):
         code = '{0}({1})'.format(name, code_args)
         return self._get_statement(code)
 
+    def _print_SysExit(self, expr):
+        code = ""
+        if not expr.arg.dtype is NativeInteger():
```
EmilyBourne, line 2663: More readable `if expr.arg.dtype is not NativeInteger():`

tests/pyccel/scripts/exits/positive_exit3.py
```diff
@@ -0,0 +1,7 @@
+# pylint: disable=missing-function-docstring, missing-module-docstring/
+
+import sys
+
+if __name__ == "__main__":
+    exit_code = 2147483647
+    sys.exit(exit_code)
```
EmilyBourne, line 7: Could you modify one of these tests to use `sys.exit`'s keyword please

EmilyBourne (changes requested):

tests/pyccel/scripts/exits/positive_exit3.py
EmilyBourne: By this I mean a test that does:
```
sys.exit(status=exit_code)
```
The tests you have added are also useful but to make codacy happy you will need to do
```
from sys import exit as sys_exit
```

Hellio404: so about that the `sys.exit` doesn't take any keyword arguments so for `python` it raises `TypeError`

EmilyBourne: Huh looks like you're right: https://docs.python.org/3/library/sys.html#sys.exit

I guess they forgot to update their docstring
image

In that case you can ignore this comment 👍

Hellio404: okay done

EmilyBourne (approved):

yguclu:
Upon resolving the conflict in `pyccel/codegen/printing/fcode.py`, please merge `master` into this branch.

yguclu (changes requested):

pyccel/ast/sysext.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/python
```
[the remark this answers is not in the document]
Hellio404: Done

pyccel/ast/sysext.py
```diff
@@ -0,0 +1,64 @@
...
+    _dtype     = NativeVoid()
+    _precision = -1
+    _rank      = 0
+    _shape     = None
+    _order     = None
+
+    def __init__(self, status=None):
+        if status is None:
+            status = LiteralInteger(0)
+        super().__init__(status)
```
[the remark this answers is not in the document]
Hellio404: Done

yguclu (approved):

PR #1192, opened 2022-09-19
Lists implementation

aihya:
Currently supported list methods:
- [x] append: insert an element to the end of the list.
- [ ] pop: remove an element from a specific index.
- [x] insert: insert an element to a specific index
- [x] clear: clear the elements of list
- [x] extend: append the elements of a list to another list
- [x] count: count number of occurrences of an element in a list
- [ ] index: get element by index
- [x] reverse: reverse the elements of a list
- [x] remove: remove first occurrence of an element from list (works on rank=1)
- [x] sort: sort elements of list
- [ ] indexing and slicing

Supported types:
- lists
- integers
- floats
- complex


EmilyBourne:
Don't forget to add documentation in the tutorial folder to explain what you are supporting

pyccel/stdlib/lists/lists.c
```diff
@@ -0,0 +1,207 @@
...
+
+    if (!(list = (t_list*)malloc(sizeof(t_list))))
+        return NULL;
+    list->size = size * tSizes[type];
+    list->type = type;
+    list->capacity = DEFAULT_CAP;
+    while (list->capacity <= list->size)
+        list->capacity *= 2;
+    if (list->size == 0)
+        list->elements = NULL;
```
aihya, line 26: @azouiten Do you think `list->elements` should be `NULL` in this case? It make sense if it's allocated with the `DEFAULT_CAP`.

pyccel/stdlib/lists/lists.c
```diff
@@ -0,0 +1,207 @@
...
+}
+
+void     free_list(t_list **list)
+{
+    free((*list)->elements);
+    free(*list);
+    *list = NULL;
+}
+
+void    append(t_list** list1, t_list* list2)
```
aihya, line 41: @azouiten We can definitely make use of the function `insert` here as it would make this function much simpler.

aihya: I think the second argument `t_list* list2` should be a `void *` instead of `t_list *` because we can append not only `t_list` type.

pyccel/stdlib/lists/lists.c
```diff
@@ -0,0 +1,207 @@
...
+    if (index >= (*list)->size)
+        index = (*list)->size;
+    char *elements = NULL;
+    if (totalSize > (*list)->capacity)
+    {
+        elements = malloc(totalSize);
+        memcpy(elements, &(*list)->elements, index);
+        memcpy(&elements[index], item, tSizes[(*list)->type]);
+        memcpy(&elements[index + tSizes[(*list)->type]], (int8_t*)(*list)->elements + index * (*list)->type, (*list)->size - index);
+        *list = allocate_list(totalSize / (*list)->type, (*list)->type, elements);
```
aihya, line 145: @azouiten It's not necessary to call `allocate_list` here as we only need to update `list->elements` to have new size. I think you can achieve this using `realloc` with the new size and shift the required bytes using `memcpy` or `memmove`.

pyccel/stdlib/lists/lists.c
azouiten: yes! that just an oversight.

pyccel/stdlib/lists/lists.c
azouiten: I was thinking we could skip this function altogether as we could simply call insert in the printing.

pyccel/stdlib/lists/lists.c
azouiten: yes that works too, but I dont think realloc is the best choice here because it could fail if it doesnt find enough space to expand into.

aihya: Well, `malloc` would fail too if it doesn't find enough space. The only reason why I thought of `realloc` is because it will copy the previous data for you, so you only need to call `memmove` on the bytes starting from `index`.

pyccel/stdlib/lists/lists.c
```diff
@@ -0,0 +1,212 @@
+#include "lists.h"
+
+
+t_list   *allocate_list(size_t size, t_type type, void *elemnts)
+{
+    t_list *list;
+    size_t tsize = tSizes[type];
```
aihya, line 7: @azouiten There's a slight problem regarding lists initialized with no elements, we have no specific type for them. I suggest to add `lst_unkown` to `enum e_type`, which could default to 1 byte in `tSizes` for example, and then reassign a new type to the list structure depending on the firsts elements type. This is involves the functions `append`, `pop` and `extend`.

pyccel/stdlib/lists/lists.c
```diff
@@ -0,0 +1,244 @@
...
+
+void     append(t_list* list, void* item)
+{
+    t_list listTmp;
+    listTmp.capacity = DEFAULT_CAP;
+    listTmp.elements = item;
+    listTmp.size = 1;
+    listTmp.type = list->type;
+    extend(list, &listTmp);
+}
```
aihya, line 85: @azouiten We can simply call `insert` in this case.
This code is not viable because we create an instance if t_list on the stack, which makes unreachable once out of the stack frame of the function `append`. This will cause segmentation fault.

pyccel/stdlib/lists/lists.c
```diff
@@ -0,0 +1,244 @@
...
+        index = (index > 0) ? index : 0;
+    }
+    return index;
+}
+
+void     insert(t_list* list, long int index, void* item)
+{
+    char * elements = list->elements;
+    size_t totalSize = list->size + 1;
+    size_t tsize = tsize;
```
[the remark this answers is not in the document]
azouiten: no idea how that happened

pyccel/stdlib/lists/lists.c
EmilyBourne: I don't think this is a good solution as there will be no way of implementing it in Fortran. I'm not sure what the work around should be though. Perhaps we could do this on the pyccel side so that by the time we start to print the type is known?

pyccel/stdlib/lists/lists.c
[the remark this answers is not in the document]
EmilyBourne: `append` is usually faster than `insert` (https://h1ros.github.io/posts/performance-difference-between-append-and-insert-in-python/) so I'm not sure if it makes sense to map the two together, but the current implementation is definitely problematic

pyccel/stdlib/lists/lists.c
```diff
@@ -0,0 +1,665 @@
...
+    t_list *list;
+    size_t tsize = tSizes[type];
+
+    if (!(list = (t_list*)malloc(sizeof(t_list))))
+        return NULL;
+    list->size = size;
+    list->type = type;
+    list->capacity = DEFAULT_CAP;
+    while (list->capacity <= list->size)
+        list->capacity *= 2;
```
EmilyBourne, line 32: Why `*=2`? And not `+= DEFAULT_CAP`? Are you not worried that this might lead to massive amounts of memory usage?

pyccel/stdlib/lists/lists.c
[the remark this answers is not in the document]
EmilyBourne: I would expect append to look something like:
```
if (list->size >= list->capacity) {
    list->capacity *= 2;
    char* elements = realloc(list->elements, list->capacity * tsize);
    list->elements = elements;
}
list->elements[list->size] = item;
list->size += 1;
```

pyccel/stdlib/lists/lists.c
```diff
@@ -0,0 +1,665 @@
...
+    while (index < list->size)
+    {
+        if (list->type == lst_list)
+        {
+            t_list *tmp = ((t_list**)elements)[index];
+            if (!(memcmp(item, tmp, sizeof(t_list))))
+                count += 1;
+        }
+        else if (!(memcmp(item, &(elements[index * tSizes[list->type]]), tSizes[list->type])))
+                count += 1;
```
EmilyBourne, line 131: Does `list.count` use `==` comparison or `is` comparison? This looks like an `is` comparison. Do you think that this is a problem? Or only if we add more complex objects such as classes later?

pyccel/stdlib/lists/lists.c
```diff
@@ -0,0 +1,665 @@
...
+        =======
+        index : size_t
+*/
+{
+    if (index >= size)
+        index = size;
+    else if (index < 0)
+    {
+        index += size + 1;
+        index = (index > 0) ? index : 0;
```
EmilyBourne, line 222: If you do the following in python:
```
a = [1,2,3]
b = a[-4]
```
You should get an `IndexError` not `b=a[0]` so I don't think line 222 is necessary

pyccel/stdlib/lists/lists.c
```diff
@@ -0,0 +1,665 @@
...
+                return index;
+        }
+        else if (!(memcmp(item, &elements[index * tSizes[list->type]], tSizes[list->type])))
+                return index;
+        index += 1;
+    }
+    return (-1);
+}
+
+size_t calculate_index(long int index, size_t size)
```
EmilyBourne, line 200: I wonder if this function is necessary. For arrays and in Fortran we do this in the printing. As it is very simple this function would be better inlined anyway

pyccel/stdlib/lists/lists.c
```diff
@@ -0,0 +1,244 @@
...
+    size_t ind = calculate_index(index, list->size) * tsize;
+
+    if (totalSize >= list->capacity)
+    {
+        list->capacity *= 2;
+        elements = realloc(elements, list->capacity * tsize);
+        list->elements = elements;
+    }
+    memmove(&(elements[ind + tsize]), &(elements[ind]), (list->size * tsize) - ind);
+    memcpy(&(elements[ind]), item, tsize);
```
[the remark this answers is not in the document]
EmilyBourne: Are you sure? `item` is an address

pyccel/stdlib/lists/lists.c
```diff
@@ -0,0 +1,665 @@
...
+    size_t tsize = tSizes[list->type];
+    char *elements = (char*)(list->elements);
+    t_pop_ret *ret_val = malloc(sizeof(t_pop_ret));
+
+    index = (index < 0) ? list->size + index : index;
+    if (index >= 0 && list->size > index)
+    {
+        ret_val->type = list->type;
+        if (list->type == lst_list)
+            ret_val->raw = (char *)((t_list**)elements)[index];
```
EmilyBourne, line 285: Why do you use `**` for lists?

pyccel/stdlib/lists/lists.c
```diff
@@ -0,0 +1,665 @@
...
+}
+
+void    free_pop(t_pop_ret** pop_val)
+/*
+        Deallocate the t_pop_ret object resulting from a pop operation
+
+        Parameters
+        ----------
+        pop_val : t_pop_ret**
+            double pointer on the t_pop_ret object to be deallocated
```
EmilyBourne, line 306: Why double pointer?

pyccel/stdlib/lists/lists.c
```diff
@@ -0,0 +1,665 @@
...
+            the t_list from where to remove the element
+
+        index : long int
+            the element to be removed
+*/
+{
+    int index;
+    t_pop_ret *ret;
+
+    if ((index = lst_index(list, item)) != -1)
```
EmilyBourne, line 333: I think this would be clearer as:
```c
int index = lst_index(list, item);
t_pop_ret *ret;

if (index != -1)
```

pyccel/stdlib/lists/lists.c
```diff
@@ -0,0 +1,665 @@
...
+    {
+        ret = pop(list, (size_t)index);
+        free_pop(&ret);
```
EmilyBourne, line 336: I think it would be better to write this function explicitly to avoid unnecessary allocation followed by immediate deallocation. If we `remove` a large object then this could be costly

pyccel/stdlib/lists/lists.c
```diff
@@ -0,0 +1,665 @@
...
+    void *buffer = malloc(tsize);
+
+    while (index + 1 <= list->size / 2)
+    {
+        memcpy(buffer, &elements[index * tsize], tsize);
+        memcpy(&elements[index * tsize], &elements[(list->size - index - 1) * tsize], tsize);
+        memcpy(&elements[(list->size - index - 1) * tsize], buffer, tsize);
+        index++;
+    }
+    free(buffer);
```
EmilyBourne, line 362: Have you tried using a stack array? `tsize` should be small so I think it should be ok to do:
`void buffer[tsize];`

pyccel/stdlib/lists/lists.c
```diff
@@ -0,0 +1,665 @@
...
+
+        index : long int
+            the element to be accessed
+*/
+{
+    char *elements = (char*)(list->elements);
+
+    if (list->size > index)
+        return ((void*)(&elements[index * tSizes[list->type]]));
+    return (NULL);
```
EmilyBourne, line 383: Do you plan to include error checking in the code? If not then what happens when `NULL` is returned?

EmilyBourne: Have you considered handling this with a macro as for ndarrays? Or with an inline function?

pyccel/stdlib/lists/lists.c
```diff
@@ -0,0 +1,665 @@
...
+
+/**
+ * @brief Set the length and type object for group allocation.
+ * 
+ * @param list 
+ * @param len 
+ * @param type 
+ */
+void set_length_and_type(t_list *list, size_t *len, int8_t *type)
+{
```
EmilyBourne, line 396: I don't understand what this function is for. The docstring does not make it clearer. What is meant by "group allocation". What is the difference between length (not used elsewhere in the file) and size?

pyccel/stdlib/lists/lists.c
```diff
@@ -0,0 +1,665 @@
...
+ * @brief Branch recursively through every element in the list and copy 
+ * it's value in the array pointed to by `int8_t *group`.
+ * Returns the number of elements copied on every recursion.
+ * 
+ * @param list 
+ * @param group 
+ * @param offset 
+ * @return size_t 
+ */
+size_t collect_data(t_list *list, int8_t *group, size_t offset)
```
EmilyBourne, line 419: Why `int8_t *`? Is this equivalent to the `char*` used elsewhere for addresses? I think it would be better to be consistent. Personally I prefer `void*` for addresses as it leads to less type confusion, but I don't know if this can cause problems in certain conditions?

pyccel/stdlib/lists/lists.c
EmilyBourne: Other than the size check is there a difference between this and `GET_INDEX`?

pyccel/stdlib/lists/lists.h
```diff
@@ -0,0 +1,74 @@
...
+#define _LISTS_
+#define DEFAULT_CAP 10
+
+#include <stdio.h>
+#include <stddef.h>
+#include <string.h>
+#include <stdlib.h>
+#include <complex.h>
+
+#define GET_INDEX(list, i) (list->elements + (i * tSizes[list->type]))
```
EmilyBourne, line 12: Are you sure this function returns the index? Not the address at that index?

EmilyBourne: If it is supposed to return the element, why do you do:
```
list->elements + (i * tSizes[list->type])
```
rather than
```
list->elements[i * tSizes[list->type]]
```
?

pyccel/stdlib/lists/lists.c
```diff
@@ -0,0 +1,665 @@
...
+            case lst_float:
+                return *((float *)GET_INDEX(list, i1)) - *((float *)GET_INDEX(list, i2));
+            case lst_double:
+                return *((double *)GET_INDEX(list, i1)) - *((double *)GET_INDEX(list, i2));
+            case lst_complex:
+                fprintf(stderr, "Ordering of complex numbers is not supported.");
+                exit(-1);
+        }
+    }
+
```
EmilyBourne, line 495: I would prefer an `else` here. That way the compiler will warn you if a `return` is forgotten in the `if`

pyccel/stdlib/lists/lists.c
```diff
@@ -0,0 +1,665 @@
...
+    cmp = 0;
+    len = 0;
+    type = 0;
+    if (list->type != lst_list)
+    {
+        switch (list->type)
+        {
+            case lst_int8:
+                return *((int8_t *)GET_INDEX(list, i1)) - *((int8_t *)GET_INDEX(list, i2));
```
EmilyBourne, line 479: Aren't you missing a loop? What about this case:
```
[0,1] < [0,2]
```
?

pyccel/stdlib/lists/lists.c
```diff
@@ -0,0 +1,665 @@
...
+                return *((double *)GET_INDEX(list, i1)) - *((double *)GET_INDEX(list, i2));
+            case lst_complex:
+                fprintf(stderr, "Ordering of complex numbers is not supported.");
+                exit(-1);
+        }
+    }
+
+    set_length_and_type(((t_list **)list->elements)[0], &len, &type);
+    group_1 = group_node_items(((t_list **)list->elements)[i1], len, type);
+    group_2 = group_node_items(((t_list **)list->elements)[i2], len, type);
```
EmilyBourne, line 498: Is the memory allocation necessary here? Couldn't `compare` just be recursive?

pyccel/stdlib/lists/lists.c
```diff
@@ -0,0 +1,665 @@
...
+ * and `i2`, putting them in two seperate allocated arrays, then compare 
+ * every value.
+ * Returns comparision value.
+ * 
+ * @param list 
+ * @param i1 
+ * @param i2 
+ * @return int 
+ */
+int compare(t_list *list, int i1, int i2)
```
EmilyBourne, line 463: What is this function comparing? I expected it to compare 2 lists but it seems that it compares 2 elements is that right? If so why is a function needed for this?

pyccel/stdlib/lists/lists.c
```diff
@@ -0,0 +1,665 @@
...
+    return cmp;
+}
+
+/**
+ * @brief Always take the last element of the list and use it as an anchor
+ * for sorting the elements with values less than it's value, while updating 
+ * the index where the anchor should exist, then swap the value in the index 
+ * `r` (the current position of the anchor) with new found position, this 
+ * ensures that the anchor is always in the right position.
+ * Returns the index of the anchor.
```
EmilyBourne, line 539: This docstring describes the method used, but it misses a description of what you are trying to do. What is `partition`?

pyccel/stdlib/lists/lists.c
aihya: I thought about it in the beginning, but I don't think its possible (not quite sure) because I would need two threads running simultaneously to be do the comparison. correct me if I'm wrong.

pyccel/stdlib/lists/lists.c
aihya: Ok, the list could contain numerical values or be a list of lists. If it happens that the list consist of numerical values (rank=1), the function would simply compare the elements in the indexes `i1` and `i2` after casting to the corresponding type. But for list of lists (rank>1), deciding whether the element in the index `i1` is bigger or smaller than `i2` won't be possible unless we compare the values in the innermost dimension.

pyccel/stdlib/lists/lists.c
aihya: This function will not remain in the code as I opted to replace it with the c function `qsort_r(3)`, which I think it uses the same quicksort algorithm, but a lot faster.

pyccel/stdlib/lists/lists.c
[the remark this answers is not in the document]
azouiten: yes `item` is already void*.

pyccel/stdlib/lists/lists.c
azouiten: I believe c++ vector container does the same, I think that a factor of 2 would minimize reallocations with the worst case scenario being 50% wasted memory. I also found this [https://stackoverflow.com/questions/1424826/why-is-vector-array-doubled](url)

pyccel/stdlib/lists/lists.c
azouiten: well for now we only support literals and lists themselves, as such this should be a == comparison as it compares he raw binary data, in the case of lists themselves I still want to add a functionality to iterate through the elements.
For more complex types tho I was thinking maybe we could a function table that holds type specific comparison functions,  with the indexes being the value of the enum corresponding to the specific type. 
something like this
```c
typedef (int)(compare_func)(t_list*, void*) compFuncs;
compFuncs[] = {compare_typeA, compare_typeB, ...}
```
I m not sure if this is the best way to do it.

pyccel/stdlib/lists/lists.c
azouiten: yes but it insert does except out of bound indexes and adjusts them. something like this works.
```
a = [1, 2, 3]
a.insert(-5, 4)
```
it outputs `[4, 1, 2, 3]`

pyccel/stdlib/lists/lists.c
azouiten: yeah I agree this could easily be calculated by pyccel

pyccel/stdlib/lists/lists.c
aihya: For multidimensional lists we store pointers over `t_list *` in `list->elements`. so instead of `(char *)((t_list *)elements)[index]` which will cause it to index with the size of the struct `t_list`, we index with the size of the pointer `t_list *`.

azouiten: int the case of a nested list it holds pointers to other t_lists, so I just cast it to access it using index only, I thought it would be a little bit more readable, but looking at it now maybe i should just revert it to `(char*)elements[index * tsize]`

pyccel/stdlib/lists/lists.c
azouiten: just so we can put NULL to the pointer we are holding, we could potentially use it to make checks.

pyccel/stdlib/lists/lists.c
azouiten: make sense, we could just add an extra parameter to specify if allocation is demanded.

pyccel/stdlib/lists/lists.c
EmilyBourne: :+1: thanks. I didn't know that vector did this

pyccel/stdlib/lists/lists.c
EmilyBourne: I am also not sure. I think it will depend how we plan to manage classes.

pyccel/stdlib/lists/lists.c
azouiten: yes error checking will be added, GET_INDEX and this function are the same
I think at some point @aihya added GET_INDEX to test his code.

pyccel/stdlib/lists/lists.c
EmilyBourne: I think this should be ok for now. As you say this should be sufficient for literals. Arrays don't have a comparison operator returning a scalar, and we haven't implemented magic methods for classes yet so it uses the default comparison (equivalent to `is`)

pyccel/stdlib/lists/lists.c
EmilyBourne: Huh, that's odd!

pyccel/stdlib/lists/lists.c
EmilyBourne: It's up to you. I think I'm just out of practice looking at C and forgot that `*` and `[]` are more or less equivalent

pyccel/stdlib/lists/lists.c
EmilyBourne: See conversation below

pyccel/stdlib/lists/lists.c
EmilyBourne: But if you want to compare an element of a list of lists to another element of a list of lists then this is equivalent to comparing a list.

I expected to see something like:
```c
/************************************
*  Function to compare 2 lists
*  Returns 1 if list1>list2
*  Returns -1 if list2>list1
*  Returns 0 if list1==list2
************************************/
int compare(t_list* list1, t_list* list2)
{
    const int list_len = MIN(list1->size, list2->size);
    for(int i(0); i<list_len; ++i)
    {
        if (list1->type == lst_list)
        {
            // assuming list2->type == lst_list for valid comparison
            int const comp = compare(list1->elements[i], list2->elements[i]);
            if (comp != 0)
                return comp;
        }
        else if (list1->elements[i] < list2->elements[i])
        {
            return -1;
        }
        else if (list1->elements[i] > list2->elements[i])
        {
            return 1;
        }
    }
    if (list1->size < list2->size) {
        return -1;
    }
    else if (list1->size > list2->size) {
        return 1;
    }
    else {
        return 0;
    }
}
```
Then if you want to compare 2 elements you do that directly. Either by knowing that they are both literals and comparing via that, or knowing that they are both lists and calling this function

aihya: This is odd, I don't understand why you want to compare the size of the lists. Haven't we agreed that the list will be homogeneous?

pyccel/stdlib/lists/lists.c
EmilyBourne: Has this been discussed? I thought there had been a discussion about error checking and we weren't sure we wanted it. I ought to have a discussion with Yaman about it at some point and see what his opinion is

pyccel/stdlib/lists/lists.c
EmilyBourne: Why would 2 unrelated homogeneous lists have the same size?

To be clear my function handles things like:
```
[1,2,3]<[1,2]
[1,2,3]<[1,2,4]
[5,3,1]<[7,32,1,9]
```
If you mean 2 elements of a list of list. I don't think homogeneous lists need to have this restriction. It is required for arrays because all the data is stored in the same memory block. This is not the case here

pyccel/stdlib/lists/lists.c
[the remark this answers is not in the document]
azouiten: [a94fed3](https://github.com/pyccel/pyccel/pull/1192/commits/a94fed3bae883f2aca1d263638b2c62600cce2ca)

pyccel/stdlib/lists/lists.c
azouiten: [f82da0e](https://github.com/pyccel/pyccel/pull/1192/commits/f82da0e8f2d8f25aef531034e2d152eb3ee0125b)

pyccel/stdlib/lists/lists.c
EmilyBourne: I think that will be the case

pyccel/stdlib/lists/lists.c
EmilyBourne: What do you mean by "imported to python"? Are you trying to **replace** python's `list` object?

pyccel/stdlib/lists/lists.c
aihya: https://github.com/pyccel/pyccel/pull/1192/commits/e7d7fce524d55ba8528fe125504613a3bdb7fb27

pyccel/stdlib/lists/lists.c
aihya: https://github.com/pyccel/pyccel/pull/1192/commits/e7d7fce524d55ba8528fe125504613a3bdb7fb27

pyccel/stdlib/lists/lists.c
aihya: https://github.com/pyccel/pyccel/pull/1192/commits/e7d7fce524d55ba8528fe125504613a3bdb7fb27

pyccel/stdlib/lists/lists.c
aihya: GET_INDEX is used to get an element from a list using an index, this function attempts to search for an element in the index.

pyccel/stdlib/lists/lists.h
aihya: https://github.com/pyccel/pyccel/pull/1192/commits/67142a65365551d27535ef09b33dc0d3a7717e69

aihya: > If it is supposed to return the element, why do you do:
> 
> ```
> list->elements + (i * tSizes[list->type])
> ```
> 
> rather than
> 
> ```
> list->elements[i * tSizes[list->type]]
> ```
> 
> ?

it's supposed to give the address of an element from the list, and then you can cast it to what ever type you want.

pyccel/stdlib/lists/lists.c
aihya: This function is removed from the code.

pyccel/stdlib/lists/lists.c
azouiten: actually no I don't recall having a discussion about it

pyccel/stdlib/lists/lists.c
azouiten: my bad, for some reason I was under the impression that we might need to import the library alone and use these methods on pylistobjects

pyccel/stdlib/lists/lists.c
azouiten: [09853cd](https://github.com/pyccel/pyccel/pull/1192/commits/09853cdde9bf961c41f209bef6d492397930248d)

pyccel/stdlib/lists/lists.c
azouiten: [288c34f](https://github.com/pyccel/pyccel/pull/1192/commits/288c34f8af9be150556748f66bfb1591213ed503)

pyccel/stdlib/lists/lists.c
aihya: @azouiten This function is not working as it should be. There was is an issue when I tried to count a elements in multidimensional lists.
Consider this:
```python
b = [[1, 2], [3, 4], [5, 6]]
print(b.count([3, 4]))
```
The code above translates to this C code:
```C
    t_list *b = NULL;
    t_list *Dummy_0000 = NULL;

    b = allocate_list(3, lst_list, (t_list*[]){
        allocate_list(1, lst_int64, (int64_t[]){
            INT64_C(1), INT64_C(2)
        }), 
        allocate_list(1, lst_int64, (int64_t[]){
            INT64_C(3), INT64_C(4)
        }), 
        allocate_list(1, lst_int64, (int64_t[]){
            INT64_C(5), INT64_C(6)
        })
    });

    Dummy_0000 = allocate_list(1, lst_int64, (int64_t[]){
        INT64_C(1)
    });

    printf("%"PRId64"\n", count(b, Dummy_0000));
```
This code will always print `0`, because if you check line 127 it will only compare the bytes of the two structures pointed to by `b` and `Dummy_0000` and not the actual data in elements.

EmilyBourne:
What is the status of this PR?

aihya:
> What is the status of this PR?

Sorry for taking too long to finish this PR, I'm trying to finish the necessary part and I will open other PRs to finish it.

EmilyBourne:
> > What is the status of this PR?
> 
> Sorry for taking too long to finish this PR, I'm trying to finish the necessary part and I will open other PRs to finish it.

Have you made any progress on this? Do you need a hand?

EmilyBourne:
@azouiten @aihya it seems that you are not following this PR anymore so I have unassigned you so someone else can pick it up

PR #1195, opened 2022-09-21
Change debugging output to show type precision

### LINKED ISSUE #1100: Error messages designed for Fortran users should be avoided
**Describe the bug**
Error messages designed for Fortran users should be avoided. Some error messages reference types using syntax such as `int(kind=8)`. This is aimed at Fortran developers. Instead the numpy name should be used (i.e. `int32`, `int64` etc)

**Test Code**
```python
from pyccel.stdlib.internal.openmp import omp_set_num_threads, omp_get_num_threads, omp_get_thread_num

def set_NT(n:'int32'):
    omp_set_num_threads(n)
    #$ omp parallel                                                                                                                                                                                        
    print("hello from thread number:", omp_get_thread_num())
    #$ omp single                                                                                                                                                                                          
    print("The best thread is number : ", omp_get_thread_num())
    #$ omp end single                                                                                                                                                                                      
    #$ omp end parallel
```

This raises:
```
pyccel:
 |error [semantic]: toto.py [4,4]| Argument 1 : n (int(kind=8)), passed to function omp_set_num_threads is incompatible (expected int(kind=4)). Please cast the argument explicitly or overload the function (see https://github.com/pyccel/pyccel/blob/master/tutorial/headers.md for details) (omp_set_num_threads(FunctionCallArgument('n')))
```
but the following would be clearer:
```
pyccel:
 |error [semantic]: toto.py [4,4]| Argument 1 : n (int64), passed to function omp_set_num_threads is incompatible (expected int32). Please cast the argument explicitly or overload the function (see https://github.com/pyccel/pyccel/blob/master/tutorial/headers.md for details) (omp_set_num_threads(FunctionCallArgument('n')))
```

yassine-alaoui:
fix #1100

changed the debugging message to show the precision such as `int64` instead of `int(kind(8))` and added conditions for complexes requested by @Hellio404 to print the proper precision since complexes have double size to cover the real and imaginary range.

running this code:
```javascript
import numpy as np
if __name__ == "__main__":
    def a(arg: 'complex128'):
        pass
    a(4)
```

used to show:
`Argument 1 : 4 (int(kind=8), passed to function a is incompatible (expected complex(kind=8)).`

now it shows this instead:
`Argument 1 : 4 (int64), passed to function a is incompatible (expected complex128)`

### DIFF
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -724,7 +724,7 @@ class SemanticParser(BasicParser):
         """
         dtype = var.dtype
         prec  = get_final_precision(var)
-        descr = f'{dtype}{(prec * 2 if dtype == NativeComplex() else prec) * 8 if prec else ""}'
+        descr = f'{dtype}{(prec * 2 if isinstance(dtype, NativeComplex) else prec) * 8 if prec else ""}'
         if include_rank and var.rank>0:
             dims = ','.join(':'*var.rank)
             descr += f'[{dims}]'
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -724,7 +724,7 @@ class SemanticParser(BasicParser):
         """
         dtype = var.dtype
         prec  = get_final_precision(var)
-        descr = f'{dtype}{prec * 8 if prec else ""}'
+        descr = f'{dtype}{(prec * 2 if dtype == NativeComplex() else prec) * 8 if prec else ""}'
         if include_rank and var.rank>0:
             dims = ','.join(':'*var.rank)
             descr += f'[{dims}]'
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -724,7 +724,7 @@ class SemanticParser(BasicParser):
         """
         dtype = var.dtype
         prec  = get_final_precision(var)
-        descr = f'{dtype}(kind={prec})'
+        descr = f'{dtype}{prec * 8 if prec else ""}'
         if include_rank and var.rank>0:
             dims = ','.join(':'*var.rank)
             descr += f'[{dims}]'


EmilyBourne:
Hi,
I'm not sure why you've requested my review on a draft PR without asking me a specific question? If you just want a first review, please update your branch, mark it ready for review and add the `needs_initial_review` tag as described in the [developer docs](https://github.com/pyccel/pyccel/blob/master/developer_docs/review_process.md). I am writing my thesis at the moment and need to hand it in next week so I am not reviewing PRs until next month at which point I will resume reviewing PRs which have reached the second stage of the review process (1 successful review, `ready_for_review` label has been added).

Sorry that I can't be more use at the moment

pyccel/parser/semantic.py
```diff
@@ -724,7 +724,7 @@ def get_type_description(self, var, include_rank = True):
         """
         dtype = var.dtype
         prec  = get_final_precision(var)
-        descr = f'{dtype}(kind={prec})'
+        descr = f'{dtype}{prec * 8 if prec else ""}'
```
Hellio404, line 727: For `complex` type will output the wrong precision. I don't exactly know why but I found [this](https://gcc.gnu.org/onlinedocs/gfortran/KIND-Type-Parameters.html#:~:text=The%20KIND%20value%20matches%20the%20storage%20size%20in%20bytes%2C%20except%20for%20COMPLEX%20where%20the%20storage%20size%20is%20twice%20as%20much)

you can see the issue with this code:
```python
def a(arg: 'complex128'):
    pass
a(4)
```
this will output 
```
Argument 1 : 4 (int64), passed to function a is incompatible (expected complex64)
```
but I think we should expect to see:
```
Argument 1 : 4 (int64), passed to function a is incompatible (expected complex128)
```

Hellio404 (changes requested):

pyccel/parser/semantic.py
EmilyBourne: See #1167 and #1134 for details

pyccel/parser/semantic.py
```diff
@@ -724,7 +724,7 @@ def get_type_description(self, var, include_rank = True):
         """
         dtype = var.dtype
         prec  = get_final_precision(var)
-        descr = f'{dtype}(kind={prec})'
+        descr = f'{dtype}{(prec * 2 if dtype == NativeComplex() else prec) * 8 if prec else ""}'
```
Hellio404, line 727: I think it's a good idea to use `isinstance` when trying to compare the identity of something

Hellio404 (changes requested):

pyccel/parser/semantic.py
Hellio404: you can just change the comparison to `isinstance(dtype, NativeComplex)`

Hellio404:
I wonder if this will be more readable
```python
if prec:
            descr = f'{dtype}{prec * 16 if isinstance(dtype, NativeComplex) else prec * 8}'
else:
    descr = f'{dtype}'
```

Hellio404 (approved):

EmilyBourne (approved):

yguclu (approved):

yguclu:
Fix #1100: print type debugging information using Numpy names instead of Fortran ones.

Issue #1200, opened 2022-10-05
Add sort function for lists


aihya:
Need to implement sorting function for lists.

**Language**
C is required for now.

PR #1201, opened 2022-10-05
Implement sort function for lists

aihya:
fixes https://github.com/pyccel/pyccel/issues/1200

The algorithm used for sorting is quicksort. It takes a pointer over `t_list` and sort the elements in ascending order.
There's a difficulty in sorting multidimensional lists which requires branching through the elements of end dimension to decide the ordering.


pyccel/stdlib/lists/lists.c
```diff
@@ -1,16 +1,5 @@
...
-            sizeof(int32_t), 
-            sizeof(int64_t), 
-            sizeof(float), 
-            sizeof(double), 
-            sizeof(void *)
-};
-
 t_list   *allocate_list(size_t size, t_type type, void *elemnts) // va_arg could alow us to take in multiple list of elements
 {
     t_list *list;
```
azouiten, line 5: you shouldn't t change my code as I m still modifying it, it would create unnecessary conflicts, especially that you are not using them in your code.

pyccel/stdlib/lists/lists.c
```diff
@@ -219,4 +210,228 @@ void*   array_subscripting(t_list *list, size_t index)
     if (list->size > index)
         return ((void*)(&elements[index * tSizes[list->type]]));
     return (NULL);
+}
+
+// Sorting ////////////////////////////
+
```
azouiten, line 216: it would be helpful if you can some documentation for these functions, and I should do the same with my functions.

aihya: https://github.com/pyccel/pyccel/pull/1201/commits/a885d1ba62455a7ecd5ec8e9bbb0d57400b87562

Issue #1196, opened 2022-10-04
Max precision arguments are incompatible with precision type interfaces 


Pinkyboi:
**Describe the bug**
When having multiple type interfaces for a function, and using default precision type indicators as argument of the `@types` decorator, an error is raised.

**To Reproduce**
```python
from pyccel.decorators import types
import numpy as np

@types('int[:]', 'int[:]')
@types('float[:]', 'float[:]')

def function(a, b):
    print("Hello World!")

if __name__ == '__main__':
    arr = np.array([0, 1, 2, 3, 4], dtype=float)
    function(arr, arr)
```

**Error message**
```bash/fortran/c
ERROR at annotation (semantic) stage
pyccel:
 |fatal [semantic]: test2.py| Arguments types provided to function_0001 are incompatible
```

**Additional context**
It works fine if we only have one `@types` decorator with a valid type indicators :
```python
from pyccel.decorators import types
import numpy as np

@types('float[:]', 'float[:]')

def function(a, b):
    print("Hello World!")

if __name__ == '__main__':
    arr = np.array([0, 1, 2, 3, 4], dtype=float)
    function(arr, arr)
```

Pinkyboi:
@EmilyBourne  I did run the test on the master branch.

EmilyBourne:
I think this error must be being caused because you can't have an array of default floats, they are automatically converted to the precise type. But I am not sure why it doesn't complain if it's not an Interface.

Presumably in the Interface case we are doing extra checks to tell the difference between `float` and `float64` which don't make sense for arrays. I think the fix to this should be simple. We just need to correct the precision in `Variable` for arrays

PR #1202, opened 2022-10-05
Default to using Numpy precision in interfaces for arrays

Pinkyboi:
Fixes [#1196](https://github.com/pyccel/pyccel/issues/1196):
Use the default numpy precision when constructing function's arguments in the case of non specified precision for numpy arrays.


EmilyBourne:
Please add a test that was failing before this PR

rmahjoubi:
can you add a couple more tests such

```python
import numpy as np

@types('int32[:]', 'int32[:]')
@types('float32[:]', 'float32[:]')

def function(a, b):
    print("Hello World!")

if __name__ == '__main__':
    arr = np.array([0, 1, 2, 3, 4], dtype=float)
    function(arr, arr)
```
I think a numeric promotion should be applied to the arguments in the following case, this should be discussed and handled in a different PR , but I think it's a good idea to add a test

```python
import numpy as np

@types('int[:]', 'int[:]')
@types('float[:]', 'float[:]')

def function(a, b):
    print("Hello World!")

if __name__ == '__main__':
    arr = np.array([0, 1, 2, 3, 4], dtype=np.float32)
    function(arr, arr)
```

Pinkyboi:
@rmahjoubi Alright i'll add that kind of test and mark them as expected to fail for now.

EmilyBourne:
> can you add a couple more tests
> 
> I think a numeric promotion should be applied to the arguments in the following case, this should be discussed and handled in a different PR , but I think it's a good idea to add a test

I'm not sure I understand what you mean here. One thing that we have taken a lot of care to avoid is array copies in the wrapper. This is something that f2py does but it's also something that can have a non-negligible effect on performance. That is why the wrapper raises an error explaining the type problem. I don't expect either of your tests to pass ever. Implicit numeric promotion should be discussed for scalars (https://github.com/pyccel/pyccel/discussions/1059), but it should be avoided for arrays.

Pinkyboi:
@nhamidn Any idea what could be causing the failed test when including the `openMP` library in `MacOSX` ?

EmilyBourne:
> @nhamidn Any idea what could be causing the failed test when including the `openMP` library in `MacOSX` ?

I don't see any reason why your PR would cause this. It looks like a change to the setup on mac. If this is the case then other PRs should encounter the same problem. `libomp` is installed on the system, but maybe we need reinstall to ensure it's correctly wired up

EmilyBourne:
> > @nhamidn Any idea what could be causing the failed test when including the `openMP` library in `MacOSX` ?
> 
> I don't see any reason why your PR would cause this. It looks like a change to the setup on mac. If this is the case then other PRs should encounter the same problem. `libomp` is installed on the system, but maybe we need reinstall to ensure it's correctly wired up

It looks like this is a git problem. Libomp has recently been updated from 14.0.6 to 15.0.2

rmahjoubi:
@pinkyboi could you please add similar tests to those I mentioned above, the first one is still failing on this branch

EmilyBourne:
> @Pinkyboi could you please add similar tests to those I mentioned above, the first one is still failing on this branch

@rmahjoubi Which test are you talking about? The first test here : https://github.com/pyccel/pyccel/pull/1202#issuecomment-1284692017 doesn't work on the master branch and would require copying an array

rmahjoubi:
@EmilyBourne I am talking about this test
```python
import numpy as np

@types('float64', 'float64')
@types('int', 'int')

def function(a, b):
    print("Hello World!")

if __name__ == '__main__':
    function(3.4,3.3)
```
It is not related directly to this issue but I think it's a simple fix.
The others I thought we may make sure an exception is raised.

EmilyBourne:
> @EmilyBourne I am talking about this test
> 
> ```python
> import numpy as np
> 
> @types('float64', 'float64')
> @types('int', 'int')
> 
> def function(a, b):
>     print("Hello World!")
> 
> if __name__ == '__main__':
>     function(3.4,3.3)
> ```
> 
> It is not related directly to this issue but I think it's a simple fix. The others I thought we may make sure an exception is raised.

I am not convinced that this is a simple fix. It is simple to fix it here (if you are familiar with the cwrapper). But this problem should be fixed in all the places it occurs at the same time otherwise it becomes much harder to explain in the documentation. Because of that I would rather leave this to its own PR, after discussion of #1059.

I agree that an error should be raised for the others though. Is that not the case?

rmahjoubi:
> > @EmilyBourne I am talking about this test
> > ```python
> > from pyccel.decorators import types
> > import numpy as np
> > 
> > @types('float64', 'float64')
> > @types('int', 'int')
> > 
> > def function(a, b):
> >     print("Hello World!")
> > 
> > if __name__ == '__main__':
> >     function(3.4,3.3)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It is not related directly to this issue but I think it's a simple fix. The others I thought we may make sure an exception is raised.
> 
> I am not convinced that this is a simple fix. It is simple to fix it here (if you are familiar with the cwrapper). But this problem should be fixed in all the places it occurs at the same time otherwise it becomes much harder to explain in the documentation. Because of that I would rather leave this to its own PR, after discussion of #1059.
> 
> I agree that an error should be raised for the others though. Is that not the case?

I can't see the relation between this test and #1059 as no casting is required there

EmilyBourne:
> I can't see the relation between this test and https://github.com/pyccel/pyccel/discussions/1059 as no casting is required there

In python there would be casting. If we were talking about integers then there would be casting on certain platforms. These are the kind of distinctions that would need to be explained in the docs if we fix this for cases where we know that in C there is no casting

rmahjoubi:
> > I can't see the relation between this test and #1059 as no casting is required there
> 
> In python there would be casting. If we were talking about integers then there would be casting on certain platforms. These are the kind of distinctions that would need to be explained in the docs if we fix this for cases where we know that in C there is no casting

@EmilyBourne Yes now I see

rmahjoubi (approved):

pyccel/ast/core.py
```diff
@@ -2818,8 +2819,9 @@ def point(self, args, use_final_precision = False):
             type_match = lambda dtype1, dtype2, call_arg, func_arg: \
                     (dtype1 in dtype2 or dtype2 in dtype1) \
                     and (call_arg.rank == func_arg.rank) \
-                    and call_arg.precision == func_arg.precision
-
+                    and (call_arg.precision == (func_arg.precision\
+                        if not func_arg.is_ndarray \
+                        else get_final_precision(func_arg)))
```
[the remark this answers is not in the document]
Pinkyboi: @rmahjoubi  It seems like that behaviour is wanted for interfaces to distinguish between `float` and `float64` for example,  or at least that's what i've understood from the comment that @EmilyBourne posted [here](https://github.com/pyccel/pyccel/issues/1196#issuecomment-1267253305)

pyccel/ast/core.py
```diff
@@ -2820,7 +2821,6 @@ def point(self, args, use_final_precision = False):
                     and (call_arg.rank == func_arg.rank) \
                     and call_arg.precision == func_arg.precision
 
-
```
EmilyBourne, line 2823: This file has not changed in a way that is useful for this PR. Please revert the changes to make the PR cleaner. (`git checkout master -- pyccel/ast/core.py`)

pyccel/ast/variable.py
```diff
@@ -123,6 +123,7 @@ def __init__(
         precision=0,
         is_argument=False,
         is_kwonly=False,
+        is_tuple=False,
```
EmilyBourne, line 126: Why do you need this argument? Shouldn't `is_tuple` be False unless this class is `TupleVariable`?

pyccel/ast/variable.py
```diff
@@ -212,6 +217,10 @@ def __init__(
         if self._rank < 2:
             self._order = None
 
+        if self.is_ndarray and not self.is_tuple\
```
EmilyBourne, line 220: Are there cases where `self.is_ndarray` and `self.is_tuple`? If so this seems like a bug that needs to be fixed

pyccel/ast/variable.py
```diff
@@ -401,6 +410,12 @@ def is_ndarray(self):
         return isinstance(self.dtype, (NativeInteger, NativeBool,
                           NativeFloat, NativeComplex))
 
+    @property
+    def is_tuple(self):
+        """user friendly method to check if the variable is a tuple:
+        """
+        return self._is_tuple
```
EmilyBourne, line 417: Shouldn't this return `False`? It should only be `True` if it is overwritten by `TupleVariable.is_tuple`

pyccel/parser/semantic.py
```diff
@@ -503,6 +503,17 @@ def _infere_type(self, expr, **settings):
...
 
+        elif isinstance(expr, IndexedElement):
+            d_var['datatype'   ] = expr.dtype
+            d_var['memory_handling'] = 'heap' if expr.rank > 0 else 'stack'
+            d_var['shape'      ] = expr.shape
+            d_var['rank'       ] = expr.rank
+            d_var['order'      ] = expr.order
+            d_var['precision'  ] = expr.precision
+            d_var['cls_base'   ] = expr.base.cls_base
+            d_var['is_tuple'   ] = expr.base.is_tuple
```
EmilyBourne, line 514: Couldn't this simply be ` d_var['is_tuple'   ] = isinstance(expr.base, TupleVariable)`?

pyccel/parser/semantic.py
```diff
@@ -503,6 +503,17 @@ def _infere_type(self, expr, **settings):
...
+            return d_var
```
EmilyBourne, line 515: I don't think that this is correct. If we have the following code:
```python
a = (1,2,3)
b = a[0]
```
then you will set `b.is_tuple` to True, but `b` is not a tuple

EmilyBourne (changes requested):
I am not entirely convinced by your fix. It seems better to me to create the variable with the correct precision where the problem occurs (arguments) rather than trying to correct the variable. This would reduce the number of times the `if` is encountered and reduces the potential for errors later. For example would it be possible to fix the problem [here](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/headers.py#L287-L288)?

I am not convinced that `is_tuple` is necessary or useful. If you can convince me that it is needed then the docstring of `Variable` needs updating

Pinkyboi:
@EmilyBourne  I totally get your idea, the fix look clanky, even when i was implementing it i had that idea that there must be a better way to do it ( i just didn't know where to modify so i went with parts of the code that i've already touched).

tests/epyccel/test_default_precision_template.py
```diff
@@ -0,0 +1,29 @@
...
+    @types('float[:]')
+    @types('complex[:]')
+    def return_array_element(array):
+        return array[0]
+
+    test_types = ['int', 'float', 'complex']
+    f1 = return_array_element
+    f2 = epyccel(f1, language=language)
+    for t in test_types:
+        d1 = randint(1, 15)
```
aihya, line 23: You sure this test works for all types of complex numbers? I mean `2+3j`, `3j` or `2`?

aihya: It seems like it's only using real part of the complex numbers that's why I asked.

Pinkyboi: I'm not sure that matters as long as we are testing with a complex type, but i may be wrong.

aihya (approved):

pyccel/ast/headers.py
```diff
@@ -284,6 +284,10 @@ def build_argument(var_name, dc):
 
             order = None
             shape = None
+
+            if rank and precision == -1:
+                precision = default_precision[dtype]
```
EmilyBourne, line 289: @aihya @azouiten I just wanted to point out that this syntax may also be used in pyccel to identify arrays, so you will also need to look out for this when handling lists

EmilyBourne (approved):

yguclu:
@Pinkyboi Thanks for this PR!

I am a bit confused by the title and description. In order to solve issue #1196 you added the lines

```python
            if rank and precision == -1:
                precision = default_precision[dtype]
```

to the local function `build_argument`, which is defined inside method `create_definition` of class `FunctionHeader`.

Here you actually used the **default precision** for Numpy arrays whenever the user does not specify the precision explicitly. So why do you use the words *"Default to using max precision"* and *"Convert default precision type specifiers to max precision"* ?

EmilyBourne:
> @Pinkyboi Thanks for this PR!
> 
> I am a bit confused by the title and description. In order to solve issue #1196 you added the lines
> 
> ```python
>             if rank and precision == -1:
>                 precision = default_precision[dtype]
> ```
> 
> to the local function `build_argument`, which is defined inside method `create_definition` of class `FunctionHeader`.
> 
> Here you actually used the **default precision** for Numpy arrays whenever the user does not specify the precision explicitly. So why do you use the words _"Default to using max precision"_ and _"Convert default precision type specifiers to max precision"_ ?

@yguclu I think what is meant by "default precision" in this PR is `-1`. Which is the precision of native python objects and tells us to use the default value. Perhaps it would be clearer to call this PR "Use numpy types inside numpy arrays"?

Pinkyboi:
@yguclu i've changed the title and description, is it more representative of the actual issue now ?

yguclu:
> @yguclu i've changed the title and description, is it more representative of the actual issue now ?

Yes, it is much clearer now! Thanks

tests/epyccel/test_default_precision_template.py
```diff
@@ -0,0 +1,29 @@
...
+        arr = np.ones(d1, dtype=t)
+        python_result = f1(arr)
+        pyccel_result = f2(arr)
+
+        assert isinstance(pyccel_result, type(python_result))
+        assert isclose(pyccel_result, python_result)
```
yguclu, line 29: In the unit tests we should explicitly define the relative and absolute tolerances for Numpy array comparison, by passing `rtol` and `atol`. See [here](https://github.com/pyccel/pyccel/blob/ca37095162bf364b63abf88f956d66e5bf9bb248/tests/epyccel/test_arrays.py#L3970-L4012) for example.

yguclu (changes requested):

yguclu (approved):

PR #1208, opened 2022-10-08
define the memory location for cuda array

framdani:
Let the user decides the memory location for cuda arrays, and make managed as the default behavior.

### DIFF
--- a/tests/internal/scripts/ccuda/kernel_launch.py
+++ b/tests/internal/scripts/ccuda/kernel_launch.py
@@ -10,7 +10,7 @@ def func(a):
 if __name__ == '__main__':
     threads_per_block = 5
     n_blocks = 1
-    arr = cuda.array([0, 1, 2, 3, 4])
+    arr = cuda.array([0, 1, 2, 3, 4], memory_location='device')
     cuda.deviceSynchronize()
     func[n_blocks, threads_per_block](arr)
     cuda.deviceSynchronize()
--- a/.github/workflows/Github_pytest.yml
+++ b/.github/workflows/Github_pytest.yml
@@ -2,7 +2,7 @@ name: Pyccel tests
 
 on:
   pull_request:
-    branches: [ master ]
+    branches: [ cuda_main_temp ]
 
 jobs:
   Linux:
--- a/pyccel/ast/cudaext.py
+++ b/pyccel/ast/cudaext.py
@@ -73,11 +73,11 @@ class CudaArray(CudaNewArray):
     arg : list, tuple, PythonList
 
     """
-    __slots__ = ('_arg','_dtype','_precision','_shape','_rank','_order')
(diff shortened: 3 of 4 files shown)


bauom:
you should merge with `cuda_main_temp` as it have now been updated.

pyccel/stdlib/cuda_ndarrays/cuda_ndarrays.cu
```diff
@@ -64,10 +64,26 @@ void cuda_array_fill_double(double c, t_ndarray arr)
 		arr.nd_double[i] = c;
 }
 
+void    device_location(void* devPtr, size_t size)
```
bauom, line 67: i think `memory` in place of `location` sounds better as in `device_memory`

pyccel/stdlib/ndarrays/ndarrays.h
```diff
@@ -65,6 +65,13 @@ enum e_types
         #endif
 };
 
+enum e_memory_locations
+{
+        managed = 0,
```
bauom, line 70: adding the numbers are not necessary if we start from one `0`

bauom (changes requested):
Could you add tests for each `memory_location`

tests/internal/scripts/ccuda/kernel_launch.py
```diff
@@ -10,7 +10,7 @@ def func(a):
 if __name__ == '__main__':
     threads_per_block = 5
     n_blocks = 1
-    arr = cuda.array([0, 1, 2, 3, 4])
+    arr = cuda.array([0, 1, 2, 3, 4], memory_location='host')
```
bauom, line 13: this array should be either `managed` or `device` as we are using it as a `kernel` argument

bauom (changes requested):
Can you add a tests that uses the `managed` memory?

tests/internal/scripts/ccuda/cuda_array_device.py
```diff
@@ -0,0 +1,17 @@
...
+@types('int[:]')
+def square(a):
+    index = cuda.blockIdx(0) * cuda.blockDim(0) + cuda.threadIdx(0)
+    a[index] = a[index] * a[index]
+
+if __name__ == '__main__':
+    threads_per_block = 5
+    n_blocks = 1
+    arr = cuda.array([0,1,2,3,4], memory_location = 'device')
+    #b = cuda.array([0,1,2,3,4], memory_location = 'device')
```
bauom, line 14: commented lines should be removed

bauom (changes requested):

tests/internal/scripts/ccuda/cuda_array_device.py
framdani: Should be good now

bauom (approved):
seems good to me now

PR #1210, opened 2022-10-08
Correct function call keyword to handle name collisions correctly

### LINKED ISSUE #1204: insensitive case naming problem in keyword argument
**Describe the bug**
We already handled the problem of the insensitive case in Fortran, but we still have an issue concerning working with optional parameter in function call.

**To Reproduce**
```python
def arr_mult_scalar(T: 'int[:]', t: int = 13):
    return T * t

def alias(T: 'int[:]', t: int):
    return arr_mult_scalar(T, t=t)
```
code generated
```fortran
  subroutine arr_mult_scalar(T, t_0001, Out_0001) 

    implicit none

    integer(i64), allocatable, intent(out) :: Out_0001(:)
    integer(i64), intent(in) :: T(0:)
    integer(i64), value :: t_0001

    allocate(Out_0001(0:size(T, kind=i64) - 1_i64))
    Out_0001 = T * t_0001
    return

  end subroutine arr_mult_scalar

  subroutine alias(T, t_0001, Out_0001) 

    implicit none

    integer(i64), allocatable, intent(out) :: Out_0001(:)
    integer(i64), intent(in) :: T(0:)
    integer(i64), value :: t_0001

    call arr_mult_scalar(T, t = t_0001, Out_0001 = Out_0001)
    return

  end subroutine alias
  !........................................```
```
**Erro raised**
```bash
    raise RuntimeError(err_msg)
/home/nouredine.ouhaddou/lab/__pyccel__/main.f90:34:32:

   34 |     call arr_mult_scalar(T, t = t_0001, Out_0001 = Out_0001)
      |                                1
Error: Keyword argument ‘t’ at (1) is already associated with another actual argument
```
**Expected**
the optional parameter should be renamed in the function call too.

**Language**
Fortran

EmilyBourne:
**Commit Summary**
- Reorder keyword arguments when creating a `FunctionCall` to match the order in the `Function Def` (Fixes an unknown bug in C when calling arguments out of order)
- Get collisionless keyword from scope when creating `FunctionCallArgument`s for `FunctionCall` objects (fixes #1204 )
- Add a test

### DIFF
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -2172,7 +2172,7 @@ class SemanticParser(BasicParser):
         # Correct keyword names if scope is available
         # The scope is only available if the function body has been parsed
         # (i.e. not for headers or builtin functions)
-        if func and func.scope:
+        if isinstance(func, FunctionDef) and func.scope:
             args = [a if a.keyword is None else \
                     FunctionCallArgument(a.value, func.scope.get_expected_name(a.keyword)) \
                     for a in args]
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -755,6 +755,22 @@ class SemanticParser(BasicParser):
                         get_final_precision(i_arg) != get_final_precision(f_arg) or
                         i_arg.rank != f_arg.rank)
 
+        # Sort arguments to match the order in the function definition
+        input_no_kwargs = [a for a in input_args if a.keyword is None]
+        nargs = len(input_no_kwargs)
+        input_kwargs = []
+        for ka in func_args[nargs:]:
+            key = ka.name
+            relevant_args = [a for a in input_args[nargs:] if a.keyword == key]
+            assert len(relevant_args) <= 1
+            if len(relevant_args) == 0 and ka.has_default:
+                input_kwargs.append(FunctionCallArgument(Nil(), key))
+            else:
+                input_kwargs.append(relevant_args[0])
+
+        input_args = input_no_kwargs + input_kwargs
+
+        # Compare each set of arguments
         for idx, (i_arg, f_arg) in enumerate(zip(input_args, func_args)):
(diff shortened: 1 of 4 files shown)


rakati (approved):
Nice way for handling this issue, I tried to test different approaches, and all of them worked fine :+1:

pyccel/parser/semantic.py
```diff
@@ -830,14 +831,31 @@ def _handle_function(self, expr, func, args, **settings):
...
+            for ka in func_args[nargs:]:
+                key = ka.name
+                relevant_args = [a for a in args[nargs:] if a.keyword == key]
+                assert len(relevant_args) <= 1
+                if len(relevant_args) == 0 and ka.has_default:
+                    input_kwargs.append(ka.default_call_arg)
+                elif len(relevant_args) == 1:
+                    input_kwargs.append(relevant_args[0])
+
+            args = input_no_kwargs + input_kwargs
```
yguclu, line 849: I wonder if this sorting could become potentially very expensive in a code containing many function calls with many arguments.

I don't know if this makes much of a difference, but we could use a single `input_args` list instead of two separate lists, in order to avoid the final concatenation. The code will still be readable and clear if we add some comments explaining what we are doing.

In any case we should make sure that `len(relevant_args)` is only called once. Now it is called 2 or 3 times.

yguclu (changes requested):
Good job! I have one minor comment

pyccel/parser/semantic.py
EmilyBourne: Fixed [416f118](https://github.com/pyccel/pyccel/pull/1210/commits/416f118a897982d506df1a90a15ea26730f076e4) and [096dcad](https://github.com/pyccel/pyccel/pull/1210/commits/096dcad82286a9ef0089e921c585f6291a6c33ba)

yguclu (approved):

PR #1211, opened 2022-10-08
Use loop unrolling to handle NumpyArange

### LINKED ISSUE #1205: assigning range of values to a row in 2d numpy array
**Describe the bug**
The problem here happens when you try to assign a range of values using `np.arange` to a row of a matrix.

**To Reproduce**
```python
def func(n: int, p: int):
    arr = np.zeros((n, p))
    for e in range(n):
        arr[e] = np.arange(e, e+p)
```
```c
void func(int64_t n, int64_t p)
{
    t_ndarray arr = {.shape = NULL};
    int64_t e;
    int64_t Dummy_0000;
    int64_t Dummy_0001;
    arr = array_create(2, (int64_t[]){n, p}, nd_double, false);
    array_fill((double)0.0, arr);
    for (e = INT64_C(0); e < n; e += INT64_C(1))
    {
        Dummy_0001 = INT64_C(0);
        for(Dummy_0000 = e; Dummy_0000 < e + p; Dummy_0000 += INT64_C(1))
        {
            arr[e, : ].nd_int64[Dummy_0001] = Dummy_0000;
            Dummy_0001 += INT64_C(1);

        }}
        free_array(arr);
    }
```
**Error**
```bash
/home/nouredine.ouhaddou/lab/__pyccel__/toto.c: In function ‘func’:
/home/nouredine.ouhaddou/lab/__pyccel__/toto.c:21:20: error: expected expression before ‘:’ token
   21 |             arr[e, : ].nd_int64[Dummy_0001] = Dummy_0000;
      |    
```
**Language**
C Language

EmilyBourne:
Use loop unrolling to handle NumpyArange. This fixes #1205 .

### DIFF
--- a/tests/epyccel/test_arrays.py
+++ b/tests/epyccel/test_arrays.py
@@ -3998,6 +3998,17 @@ def test_numpy_arange_negative_step_2(language):
     f2 = epyccel(f1, language = language)
     np.testing.assert_array_almost_equal(f1(), f2(), decimal = 9)
 
+def test_numpy_arange_into_slice(language):
+    f1 = arrays.arr_arange_7
+    f2 = epyccel(f1, language = language)
+    n = randint(2,10)
+    m = randint(2,10)
+    x = np.array(100*np.random.random((n,m)), dtype=int)
+    x_expected = x.copy()
+    f1(x_expected)
+    f2(x)
+    np.testing.assert_array_almost_equal(x, x_expected, decimal = 9)
+
 def test_iterate_slice(language):
     f1 = arrays.iterate_slice
     f2 = epyccel(f1, language = language)
--- a/tests/epyccel/test_arrays.py
+++ b/tests/epyccel/test_arrays.py
@@ -4009,7 +4009,7 @@ def test_numpy_arange_into_slice(language):
     x_expected = x.copy()
     f1(x_expected)
     f2(x)
-    np.testing.assert_allclose(f1(), f2(), rtol=RTOL, atol=ATOL)
+    np.testing.assert_allclose(x, x_expected, rtol=RTOL, atol=ATOL)
 
 def test_iterate_slice(language):
     f1 = arrays.iterate_slice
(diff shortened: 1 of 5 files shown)


EmilyBourne:
@nhamidn regarding [this comment](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/utilities.py#L475), could `__getitem__` be used to do this?

rakati (approved):

nhamidn:
> @nhamidn regarding [this comment](https://github.com/pyccel/pyccel/blob/master/pyccel/ast/utilities.py#L475), could `__getitem__` be used to do this?

I'll check it out

tests/epyccel/test_arrays.py
```diff
@@ -3998,6 +3998,17 @@ def test_numpy_arange_negative_step_2(language):
...
+def test_numpy_arange_into_slice(language):
+    f1 = arrays.arr_arange_7
+    f2 = epyccel(f1, language = language)
+    n = randint(2,10)
+    m = randint(2,10)
+    x = np.array(100*np.random.random((n,m)), dtype=int)
+    x_expected = x.copy()
+    f1(x_expected)
+    f2(x)
+    np.testing.assert_array_almost_equal(x, x_expected, decimal = 9)
```
yguclu, line 4010: I have just noticed that in this module we use `numpy.testing`, but according to [their own documentation](https://numpy.org/doc/stable/reference/testing.html#writing-your-own-tests) we should prefer the standard `assert` statements when using Pytest:

> pytest internally rewrites the `assert` statement to give informative output when it fails, so should be preferred over the legacy variant `numpy.testing.assert_`. Whereas plain assert statements are ignored when running Python in optimized mode with -O, this is not an issue when running tests with pytest.

yguclu (approved):

tests/epyccel/test_arrays.py
yguclu: After reading more on the subject and running some tests myself, I think it is a good idea to use `numpy.testing` because it provides a quite detailed error message. Nevertheless we should only use the functions listed [here](https://numpy.org/doc/stable/reference/routines.testing.html#asserts), and avoid these [deprecated functions](https://numpy.org/doc/stable/reference/routines.testing.html#asserts-not-recommended):

- `assert_`
- `assert_almost_equal`
- `assert_approx_equal`
- `assert_array_almost_equal`
- `print_assert_equal`

PR #1221, opened 2022-10-23
Cuda grid implementation

Pinkyboi:
This PR will implement the call the `cuda.grid()` by replacing it call with the corresponding arithmetic operation.

The `python` code:
```Python
from pyccel.decorators import kernel, types
from pyccel import cuda

@kernel
@types('int[:]')
def func(a):
    i,j,k = cuda.grid(2)
```
The `ccuda` code generated:
```cuda
#include "test.h"
#include <stdlib.h>
#include <stdint.h>
#include "ndarrays.h"


extern "C" __global__ void func(t_ndarray a)
{
    int64_t i;
    int64_t j;
    int64_t k;
    i = blockIdx.x * blockDim.x + threadIdx.x;
    j = blockIdx.y * blockDim.y + threadIdx.y;
    k = blockIdx.z * blockDim.z + threadIdx.z;
}
```

### DIFF
--- a/tests/internal/scripts/ccuda/cuda_grid.py
+++ b/tests/internal/scripts/ccuda/cuda_grid.py
@@ -1,0 +1,30 @@
+from pyccel.decorators import kernel, types
+from pyccel import cuda
+
+@kernel
+@types('int[:]')
+def func_1d(a):
+    i = cuda.grid(0)
+    print("1 dim :", a[i])
+
+@kernel
+@types('int[:]')
+def func_2d(a):
+    i, j = cuda.grid(1)
+    print("2 dim :", a[i], a[j])
+
+@kernel
+@types('int[:]')
+def func_3d(a):
+    i, j, k = cuda.grid(2)
+    print("3 dim :", a[i], a[j], a[k])
+
+if __name__ == '__main__':
+    threads_per_block = 5
+    n_blocks = 1
+    arr = cuda.array([0, 1, 2, 3, 4])
+    cuda.deviceSynchronize()
+    # Sice we dont support multi-dim n_block / threads_per_block
+    # func_2d and func_3d won't compile
+    func_1d[n_blocks, threads_per_block](arr)
+    cuda.deviceSynchronize()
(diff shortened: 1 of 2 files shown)


Pinkyboi:
@bauom could you please check if this is the behaviour wanted for the function call ( i didn't find much resources about the calculation done under the hood).

tests/internal/scripts/ccuda/cuda_grid.py
```diff
@@ -0,0 +1,30 @@
+from pyccel.decorators import kernel, types
+from pyccel import cuda
+
+@kernel
+@types('int[:]')
+def func_1d(a):
```
bauom, line 6: it would be better to call all of these `kernel` functions

bauom (changes requested):

tests/internal/scripts/ccuda/cuda_grid.py
Pinkyboi: Done.

tests/internal/scripts/ccuda/cuda_grid.py
```diff
@@ -24,7 +24,7 @@ def func_3d(a):
     n_blocks = 1
     arr = cuda.array([0, 1, 2, 3, 4])
     cuda.deviceSynchronize()
-    # Sice we dont support multi-dim n_block / threads_per_block
-    # func_2d and func_3d won't compile
     func_1d[n_blocks, threads_per_block](arr)
+    func_2d[n_blocks, threads_per_block](arr)
```
bauom, line 28: after the discussion that we had in the meeting I thought that we agreed to just use the `func_1d` for now until we add the support for multi-dim `n_block`  and `threads_per_block`.

bauom (changes requested):

tests/internal/scripts/ccuda/cuda_grid.py
Pinkyboi: I probably added it before our meeting, i'll revert it.

bauom (approved):
seems good to me.

PR #1222, opened 2022-10-26
Add cupy ravel

Pinkyboi:
This PR will implement the call the `cupy.ravel` by replacing it call with array creation of a flattened version of the passed array.

TODO:

- [ ] Add unit tests
- [ ] Create separate issue for `numpy.reshape`
- [ ] Set as blocked by issue above


pyccel/ast/cupyext.py
```diff
@@ -436,6 +444,40 @@ def __str__(self):
...
+            raise TypeError('Unknown type of  %s.' % type(arg))
+        if arg.rank == 0:
+            raise TypeError('Unknown type of  %s.' % type(arg))
+        return super().__new__(cls)
+
+    def __init__(self, arg, memory_location='managed'):
+        super().__init__(arg = arg, memory_location = memory_location)
+        shape = reduce((lambda x, y: x.python_value * y.python_value), self.shape)
+        self._shape = [shape if isinstance(shape, LiteralInteger) else LiteralInteger(shape)]
+        self._rank = len(self._shape)
```
bauom, line 473: can this be anything beside `1` ?

bauom (changes requested):

tests/internal/scripts/ccuda/cupy_ravel.py
```diff
@@ -0,0 +1,25 @@
+from pyccel.decorators import kernel, types
+from pyccel import cuda
+import cupy as cp
+
+@kernel
+@types('int[:]')
+def func(a):
```
bauom, line 7: can we have an example that is changing some elements of the array passed

tests/internal/scripts/ccuda/cupy_ravel.py
```diff
@@ -0,0 +1,25 @@
...
+    flat_cp_arr = cp.array([0, 1, 2, 3, 4])
+    
+    arr1 = cp.ravel(cp_arr)
+    arr2 = cp.ravel(flat_cp_arr)
+    
+    arr2 = cp.ravel([0, 1, 2, 3, 4])
+    arr3 = cp.ravel([[0, 1], [2, 3], [4, 2]])
+    
+    cuda.deviceSynchronize()
+    func[n_blocks, threads_per_block](arr)
```
bauom, line 24: the passed array does not exist.

tests/internal/scripts/ccuda/cupy_ravel.py
```diff
@@ -0,0 +1,25 @@
...
+    i = cuda.threadIdx(0) + cuda.blockIdx(0) * cuda.blockDim(0)
+    print("Hello World! ", a[i])
+
+if __name__ == '__main__':
+    threads_per_block = 5
+    n_blocks = 1
+    cp_arr = cp.array([[0, 1], [2, 3], [4, 2]])
+    flat_cp_arr = cp.array([0, 1, 2, 3, 4])
+    
+    arr1 = cp.ravel(cp_arr)
```
bauom, line 17: better to separate the `variable` and `literals` tests

pyccel/parser/semantic.py
```diff
@@ -902,11 +914,11 @@ def _handle_kernel(self, expr, func, args, **settings):
                         symbol = expr,
                         severity='fatal')
             # TODO : type check the NUMBER OF BLOCKS 'numBlocks' and threads per block 'tpblock'
-            if not isinstance(expr.numBlocks, LiteralInteger):
+            if not isinstance(expr.numBlocks, (LiteralInteger, PyccelSymbol)):
```
bauom, line 917: we either check exactly on if the expr is a int `variable` if we can or we just limit it to `LiteralInteger` for now

pyccel/parser/semantic.py
```diff
@@ -902,11 +914,11 @@ def _handle_kernel(self, expr, func, args, **settings):
...
                 errors.report("Invalid Block number parameter for Kernel call",
                         symbol = expr,
                         severity='error')
-            if not isinstance(expr.tpblock, LiteralInteger):
+            if not isinstance(expr.tpblock, (LiteralInteger, PyccelSymbol)):
```
bauom, line 921: same as above.

bauom (changes requested):

tests/internal/scripts/ccuda/cupy_ravel.py
Pinkyboi: Sure i'll add that.

tests/internal/scripts/ccuda/cupy_ravel.py
Pinkyboi: Good catch ill do that.

tests/internal/scripts/ccuda/cupy_ravel.py
Pinkyboi: Alright !

pyccel/parser/semantic.py
Pinkyboi: For this one i added PyccelSymbol because it seems like you are variables as number or blocks and number of thread per blocks in other cuda test.

pyccel/parser/semantic.py
Pinkyboi: Same reason as the above.

tests/internal/scripts/ccuda/cupy_ravel.py
Pinkyboi: Done in [dcdc81f](https://github.com/pyccel/pyccel/pull/1222/commits/dcdc81fb9230c8977b7563a658c1bedc2d1037b1)

tests/internal/scripts/ccuda/cupy_ravel.py
Pinkyboi: Done in [dcdc81f](https://github.com/pyccel/pyccel/pull/1222/commits/dcdc81fb9230c8977b7563a658c1bedc2d1037b1)

tests/internal/scripts/ccuda/cupy_ravel.py
Pinkyboi: done in [dcdc81f](https://github.com/pyccel/pyccel/pull/1222/commits/dcdc81fb9230c8977b7563a658c1bedc2d1037b1)

Pinkyboi:
@bauom I have an interrogation concerning the current way things are handled in Allocation, for `cupy.ravel` to be sent to a kernel it need to be on the heap which make sense, but at the same time it should be a view, for now i kinda ignore the second condition which will result in a memory leak because the array buffer is allocated in `array_create`, what could be the best way to handle that in your opinion ?

pyccel/ast/cupyext.py
```diff
@@ -436,6 +442,39 @@ def __str__(self):
...
+
+    Parameters
+    ==========
+    arg : PyccelAstNode
+            A PyccelAstNode of unknown shape
+    memory_location : string
+            The location where the new array memory should be allocated
+    """
+    name   = 'ravel'
+    __slots__ = ('_arg','_dtype','_precision','_shape','_rank','_order', '_memory_location')
```
EmilyBourne, line 458: See https://towardsdatascience.com/understand-slots-in-python-e3081ef5196d for more details about `__slots__`. There is a section on inheritance.

`__slots__` are inherited. By repeating them you create multiple objects `CupyArray.arg` and `CupyRavel.arg` in the same object. This tends to break inheritance as line 470 sets `self(CupyRavel)._shape` not `self(CupyArray)._shape` so functions in `CupyArray` don't have access to this value.

```suggestion
    __slots__ = ()
```

pyccel/stdlib/cuda_ndarrays/cuda_ndarrays.cu
```diff
@@ -132,6 +132,19 @@ t_ndarray   cuda_array_create(int32_t nd, int64_t *shape,
...
 
+void cupy_ravel(t_ndarray *dest, t_ndarray2 src, enum e_memory_locations location)
+{
+    *dest = src;
+    void (*fun_ptr_arr[])(void*, size_t) = {managed_memory, host_memory, device_memory};
+
+    dest->nd = 1;
+    (*fun_ptr_arr[location])(&(dest->shape), sizeof(int64_t));
+    (*fun_ptr_arr[location])(&(dest->strides), sizeof(int64_t));
+    *(dest->shape) = src.lenght;
```
EmilyBourne, line 143: Type `lenght` -> `length`. Does this mean that this function isn't tested or that there is a typo throughout the code base?

EmilyBourne:
@Pinkyboi I fixed your line endings to fix the Codacy tests. Don't forget to pull before making changes to fix the failing codacy tests

pyccel/stdlib/cuda_ndarrays/cuda_ndarrays.cu
Pinkyboi: Thats a typo from my side, i can't really test my changes locally for now since i don't have access to a machine with a Nvidia GPU for now, so i mostly only verify the printing side and count on @bauom to verify the compilation.

EmilyBourne: Ah. Are there no compilation tests then? You don't need a GPU to compile, just to execute

pyccel/stdlib/cuda_ndarrays/cuda_ndarrays.cu
```diff
@@ -132,6 +132,19 @@ t_ndarray   cuda_array_create(int32_t nd, int64_t *shape,
     return (arr);
 }
 
+void cupy_ravel(t_ndarray *dest, t_ndarray2 src, enum e_memory_locations location)
```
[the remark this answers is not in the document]
Pinkyboi: Good catch, thank you !

pyccel/stdlib/cuda_ndarrays/cuda_ndarrays.cu
Pinkyboi: Pretty much i didn't test compilation, i'll try to install nvcc this time and see how it goes.

pyccel/stdlib/cuda_ndarrays/cuda_ndarrays.cu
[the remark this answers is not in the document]
Pinkyboi: Fixed in [27c4250](https://github.com/pyccel/pyccel/pull/1222/commits/27c42509cd49ed0d43dc62a1788e1bec1b4d5c12).

pyccel/stdlib/cuda_ndarrays/cuda_ndarrays.cu
Pinkyboi: Fixed in [27c4250](https://github.com/pyccel/pyccel/pull/1222/commits/27c42509cd49ed0d43dc62a1788e1bec1b4d5c12).

pyccel/ast/cupyext.py
```diff
@@ -436,6 +444,39 @@ def __str__(self):
...
+            raise TypeError('Unknown type of  %s.' % type(arg))
+        if arg.rank == 0:
+            raise TypeError('Unknown type of  %s.' % type(arg))
+        return super().__new__(cls)
+
+    def __init__(self, arg):
+        super().__init__(arg = arg)
+        shape = reduce((lambda x, y: x.python_value * y.python_value), self.shape)
+        self._shape = [shape if isinstance(shape, LiteralInteger) else LiteralInteger(shape)]
+        self._rank = 1
```
EmilyBourne, line 473: If this is always 1 can it be a class attribute? Or is it here due to the inheritance from `CupyArray`?

pyccel/ast/cupyext.py
```diff
@@ -436,6 +444,39 @@ def __str__(self):
...
+    Parameters
+    ==========
+    arg : PyccelAstNode
+            A PyccelAstNode of unknown shape
+    memory_location : string
+            The location where the new array memory should be allocated
+    """
+    name   = 'ravel'
+    __slots__ = ()
+    _attribute_nodes = ('_arg',)
```
EmilyBourne, line 461: Could you omit this line: `_attribute_nodes = ('_arg',)` and just let it be inherited from `CupyArray` to avoid risking it getting out of date?

pyccel/codegen/printing/ccudacode.py
```diff
@@ -521,19 +552,18 @@ def copy_CudaArray_Data(self, expr):
...
+            cpy_data = f"cudaMemcpy({lhs}.{dtype}, {arg}.{dtype}, {lhs}.buffer_size, cudaMemcpyHostToDevice);"
             return '%s\n' % (cpy_data)
         else :
+            if arg.rank > 1:
+                arg = self._flatten_list(arg)
             arg = ', '.join(self._print(i) for i in arg)
             dummy_array = "%s %s[] = {%s};\n" % (declare_dtype, dummy_array_name, arg)
-            cpy_data = "cudaMemcpy({0}.raw_data, {1}, {0}.buffer_size, cudaMemcpyHostToDevice);".format(self._print(lhs), dummy_array_name, dtype)
+            target_array_name = self._print(lhs)
+            cpy_data = f"cudaMemcpy({target_array_name}.{dtype}, {dummy_array_name}, {target_array_name}.buffer_size, cudaMemcpyHostToDevice);"
```
EmilyBourne, line 566: Beware, this way of writing the code does not allow for optional arrays

tests/internal/scripts/ccuda/cupy_ravel_variable.py
```diff
@@ -0,0 +1,24 @@
...
+@kernel
+@types('int[:]', 'int[:]', 'int[:]')
+def func(a, b, c):
+    i = cuda.threadIdx(0) + cuda.blockIdx(0) * cuda.blockDim(0)
+    print("Hello World! ", a[i], b[i], c[i])
+
+if __name__ == '__main__':
+    threads_per_block = 32
+    n_blocks = 1
+    c =[[1,2],[1,3]]
```
EmilyBourne, line 16: Could you use tuples here as lists are not officially supported:
```suggestion
    c =((1,2),(1,3))
```

EmilyBourne (changes requested):

pyccel/ast/cupyext.py
Pinkyboi: We need that `_rank` to be inherited from `CupyArray` since we will use the normal `CupyArray` printer in some cases.

pyccel/codegen/printing/ccudacode.py
Pinkyboi: I'm not really sure to understand what you mean by "optional arrays".

pyccel/ast/cupyext.py
Pinkyboi: Yeah we can't really put its as class attribute since we inherit from `CupyArray` and we use `CupyArray`'s printer in some cases.

pyccel/ast/cupyext.py
Pinkyboi: Done in [be2fd09](https://github.com/pyccel/pyccel/pull/1222/commits/be2fd091c234f25b4ceaddf4fb8ac361e6c16966).

tests/internal/scripts/ccuda/cupy_ravel_variable.py
Pinkyboi: Done in [3fb754f](https://github.com/pyccel/pyccel/pull/1222/commits/3fb754ffb4b4eb90b92fcb423f01a41c6843fa50).

pyccel/codegen/printing/ccudacode.py
EmilyBourne: See the discussion on this PR: https://github.com/pyccel/pyccel/pull/1280#discussion_r1058301708

pyccel/ast/cupyext.py
```diff
@@ -436,6 +444,38 @@ def __str__(self):
...
+
+    def __init__(self, arg):
+        super().__init__(arg = arg)
+        shape = reduce((lambda x, y: x.python_value * y.python_value), self.shape)
+        self._shape = [shape if isinstance(shape, LiteralInteger) else LiteralInteger(shape)]
+        self._rank = 1
+
+    @property
+    def arg(self):
+        return self._arg
```
EmilyBourne, line 476: How is this different from the function you are overriding?

pyccel/ast/cupyext.py
EmilyBourne: Ok. So `CupyRavel` is a special case of `CupyArray`? I'm surprised using `CupyArray`'s printer doesn't cause problems with the `arg` property. I would have expected `Ravel` to be more or less equivalent to `_flatten_list`

pyccel/ast/cupyext.py
Pinkyboi: This one is my bad, i'll remove it.

Pinkyboi: Removed in [def5e4e](https://github.com/pyccel/pyccel/pull/1222/commits/def5e4eb1e30a14de57c60d1033c70f4d0561c20).

pyccel/ast/cupyext.py
Pinkyboi: In some cases we will need to copy the array that's where we use `copy_CudaArray_Data` like in `CupyArray`, but in some other cases we need to create a view instead.

EmilyBourne: If it can be a view then how can it be a special case of `CupyArray`? I am not convinced by this inheritance

Pinkyboi: It's just that if the original array isn't a `device` array, it will create a copy the same way we handle it in a regular `CupyArray`, what do you suggest instead ?

pyccel/codegen/printing/ccudacode.py
Pinkyboi: I'm sorry i took this late to answer.
if i understand correctly, this wont work if i had a pointer instead of a regular array ?
this is the case where the argument is not a variable and so it can't be a pointer, but i tried for the case of a variable.
Like the following case:
```python
@types('int[:]')
def func(arr = None):
    if arr is None:
        arr2 = cp.array([0, 1, 2, 3, 4])
    else:
        arr2 = cp.ravel(arr)
    return arr2 
```
I got the following result:
```C
extern "C" void func(t_ndarray *arr)
{
    t_ndarray arr2 = {.shape = NULL};
    if (arr == NULL)
    {

        int64_t tmp_shape[] = {5};
        arr2 = cuda_array_create(1, tmp_shape, nd_int64, false, allocateMemoryOnDevice);
        int64_t cuda_array_dummy[] = {0, 1, 2, 3, 4};
        cudaMemcpy(arr2.nd_int64, cuda_array_dummy, arr2.buffer_size, cudaMemcpyHostToDevice);
    }
    else
    {

        int64_t tmp_shape_0001[] = {arr->shape[0]};
        arr2 = cuda_array_create(1, tmp_shape_0001, nd_int64, false, allocateMemoryOnDevice);
        cudaMemcpy(arr2.nd_int64, (*arr).nd_int64, arr2.buffer_size, cudaMemcpyHostToDevice);
    }
}
```
I don't think there's a problem with this transpiration.

EmilyBourne: > this is the case where the argument is not a variable and so it can't be a pointer, but i tried for the case of a variable.

That's why I just put "Beware" and didn't ask you to change it. I think it is better practice to always write it in a way that makes the pointer status implicit. That way you don't need to remember all the places where you need to use pointers and much less code needs updating if we add a case.

E.g. when code is added to be able to return cuda arrays, returning 2 arrays will make those arrays into pointers.

> I don't think there's a problem with this transpiration.

I see 2 problems:
1.  (Unrelated to this PR) The function says `return arr2` but `arr2` is not returned
2.  `(*arr).nd_int64` : this is an unidiomatic way of writing this. If class objects were used this would be written more idiomatically as `arr->nd_int64`. This is also a problem in the non-cuda code though. I need to open a PR and go through and sort it out. Probably I will do this when I write the codegen stage docs

EmilyBourne: (By the way, transpiration is the liquid that your body generates when you're hot. I think you mean translation)

Pinkyboi: I meant transpilation, i guess I clicked the auto correction without noticing 😅.

bauom (changes requested):
as the tests aim to test `cupy.ravel` and these tests are only for compiling they should be striped to the minimum compiled code possible.

pyccel/stdlib/cuda_ndarrays/cuda_ndarrays.cu
```diff
@@ -132,6 +132,17 @@ t_ndarray   cuda_array_create(int32_t nd, int64_t *shape,
     return (arr);
 }
 
+void cupy_ravel(t_ndarray *dest, t_ndarray src)
+{
+    *dest = src;
+    dest->nd = 1;
+    managed_memory(&(dest->shape), sizeof(int64_t));
+    managed_memory(&(dest->strides), sizeof(int64_t));
```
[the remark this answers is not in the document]
Pinkyboi: Fixed in [c970bf7](https://github.com/pyccel/pyccel/pull/1222/commits/c970bf7c2a765b2460d73a623965a1bd1f052511).
I did clean up the tests for `cupy.ravel` here too [a8c5db4](https://github.com/pyccel/pyccel/pull/1222/commits/a8c5db4779b6f2030eb1da75ce1191a1ee3df333).

pyccel/ast/cupyext.py
EmilyBourne: Hmm, I'm not sure. Maybe `CupyRavel` to act as the reshape, but if `memory_location != 'device'` then do `self._arg = CupyArray(arg)` ? But that will create an unnecessary temporary. Can `CupyArray` take a ravel object as an argument?

I'm almost tempted to say that it should be something like:
```python
class CupyRavel:
    def __new__(cls, arg):
        if arg.memory_location == 'device':
            return Reshape(arg)
        else:
            return Reshape(CupyArray(arg))
```

Do cupy's arrays have a `reshape` attribute like numpy's?

Pinkyboi: I think its a good idea, but it would be better for it to have it own PR.

EmilyBourne: Maybe, I definitely agree that `Reshape` should have its own PR,

EmilyBourne:
@Pinkyboi Please could you migrate this PR to the `pyccel/pyccel-cuda` repo. You can transfer your branch to that repo using the following commands:
```
git remote add cuda git@github.com:pyccel/pyccel-cuda.git
git checkout GPU_ddocs
git push -u cuda GPU_ddocs
```
cc: @bauom

PR #1223, opened 2022-10-27
Fix Macos openmp compilation problem

EmilyBourne:
Update mac compilation to correctly include and link openmp. Update gfortran to a more recent version (gfortran-10 is no longer installed by default).


yguclu:
Superseded by PR #1224

PR #1225, opened 2022-10-28
Add cuda.copy function

framdani:
Add `cuda.copy` function which copies  data from `host` or `device` and places it into a newly created array.

The function takes three parameters:
- The `src`; an existing cudaArray.
- The memory location of the `dst`; the newly created array.
- The third is a boolean which indicates whether the copy is `asynchronous` or not.

### DIFF
--- a/pyccel/ast/cudaext.py
+++ b/pyccel/ast/cudaext.py
@@ -199,7 +199,7 @@ class CudaCopy(CudaNewArray):
             raise TypeError('unknown type of  %s.' % type(arg))
         
         # Verify the memory_location of src
-        if arg._memory_location not in ('device', 'host'):
+        if arg.memory_location not in ('device', 'host'):
             raise ValueError("The direction of the copy should be from 'host' or 'device'")
 
         # Verify the memory_location of dst
@@ -211,11 +211,11 @@ class CudaCopy(CudaNewArray):
             raise TypeError('is_async must be boolean')
         
         self._arg             = arg
-        self._shape           = arg._shape
-        self._rank            = arg._rank
-        self._dtype           = arg._dtype
-        self._order           = arg._order
-        self._precision       = arg._precision
+        self._shape           = arg.shape
+        self._rank            = arg.rank
+        self._dtype           = arg.dtype
+        self._order           = arg.order
+        self._precision       = arg.precision
         self._memory_location = memory_location
         self._is_async        = is_async
         super().__init__()
@@ -228,7 +228,9 @@ class CudaCopy(CudaNewArray):
     def memory_location(self):
         return self._memory_location
 
-
+    @property
+    def is_async(self):
+        return self._is_async
 
 class CudaThreadIdx(CudaInternalVar)        : pass
 class CudaBlockDim(CudaInternalVar)         : pass
--- a/pyccel/ast/cudaext.py
+++ b/pyccel/ast/cudaext.py
@@ -199,11 +199,11 @@ class CudaCopy(CudaNewArray):
             raise TypeError('unknown type of  %s.' % type(arg))
         
         # Verify the memory_location of src
-        if arg.memory_location not in ('device', 'host'):
+        if arg.memory_location not in ('device', 'host', 'managed'):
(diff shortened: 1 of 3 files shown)


bauom (approved):
this PR seems done to me.
i forgot about it, could you request a review next time so i can notice it easily.

pyccel/ast/cudaext.py
```diff
@@ -174,6 +174,62 @@ def __init__(self, dim=None):
...
+    
+    is_async: bool
+        Indicates whether the copy is asynchronous or not [Default value: False]
+
+    """
+    __slots__ = ('_arg','_dtype','_precision','_shape','_rank','_order','_memory_location', '_is_async')
+
+    def __init__(self, arg, memory_location, is_async=False):
+
+        if not isinstance(arg, (PythonTuple, PythonList, Variable)):
```
Pinkyboi, line 198: I'm not really sure if we can accept all those types, for example only `Variable` has the attribute `memory_location` in you implementation something like this should be possible:
``` python
if __name__ == '__main__':
    arr = cuda.array([1,2,3,4], memory_location='device')
    cpy = cuda.copy([1,2,3,4], 'host', is_async=True)
```

pyccel/ast/cudaext.py
```diff
@@ -174,6 +174,62 @@ def __init__(self, dim=None):
...
+            raise TypeError('unknown type of  %s.' % type(arg))
+        
+        # Verify the memory_location of src
+        if arg._memory_location not in ('device', 'host'):
```
Pinkyboi, line 202: Why not use the getter of `_memory_location` instead ?

pyccel/ast/cudaext.py
```diff
@@ -174,6 +174,62 @@ def __init__(self, dim=None):
     def dim(self):
         return self._dim
 
+
+class CudaCopy(CudaNewArray):
```
Pinkyboi, line 178: Any reason to have inherit from `CudaNewArray` ? shouldn't this be a `PyccelInternalFunction` instead ?

Pinkyboi (changes requested):

framdani:
> this PR seems done to me. i forgot about it, could you request a review next time so i can notice it easily.

Well noted.

pyccel/ast/cudaext.py
framdani: Yes you right , it will raise an error in this case. 
It would be better to restrict the copy only from existing cuda arrays for now.
Fixed in 28f3ecf8916671459508cdbd9b75fcca28a692f4

pyccel/ast/cudaext.py
framdani: I didn't pay attention since I could access it within this class.
It has now been updated 55ac74160903acf9799f1a795cbab1eb035dcf0c

pyccel/ast/cudaext.py
framdani: Due to the creation of a new array. And As stated in the docstring, cudaNewArray is the base class for all the cuda functions that imply a call to allocate.

pyccel/ast/cudaext.py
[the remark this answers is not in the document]
framdani: Already updated here 3b6c82f950a072652c41b27ce080802b2b5d6ac9

pyccel/ast/cudaext.py
Pinkyboi: But unless i missed something your `CudaCopy` doesn't allocate anything by itself, just copies the data from one array to another,  if your Class Inherited from `PyccelInternalFunction` for example would it change anything ?

framdani: Yes, `CudaCopy` only copies the data from one array to a new cuda array that will be created. 
In case of `PyccelInternalFunction` as base class, I should add the instance of `CudaCopy` in the type inference which is the same as `CudaNewArray`.
I think that `CudaNewArray `as base class is wokring just fine.

Pinkyboi: Sure that works too.

Pinkyboi (approved):

bauom (approved):

Issue #798, opened 2021-03-09
OpenMP GPU Commands do not compile correctly


EmilyBourne:
**Describe the bug**
To compile OpenMP GPU commands additional compiler flags may be necessary (e.g. `-foffload`). This needs investigating to ensure that files compile.

Currently the implemented tests test the following OpenMP GPU commands:
- omp_set_default_device
- omp_get_default_device
- omp_get_num_devices
- omp_is_initial_device
- omp_get_initial_device
- omp_get_team_num
- omp_get_num_teams

but they do not compile reliably with the current flags and libraries

EmilyBourne:
This should be easier to test after PR #880 as the errors will be visible

PR #1226, opened 2022-11-01
Rewrite pyccel omp handling for better support for other versions

Hellio404:
This PR reintroduces OpenMP in Pyccel as a standalone plugin, interfacing with Pyccel via a plugins utility.

## Plugins Utility

The plugins module provides the basic classes from which every plugin should inherit, and it's the main interface to the plugins from Pyccel.

**Core Functionality**
- Via the plugins module, a Pyccel class instance can register itself to be patched with the plugin's functionalities, then each plugin picks its targets from the registered objects
- Currently, the plugins are loaded and used in no particular order; when a plugin fails to load, a warning is raised and the loader resumes with the rest
- When an object registers itself with the plugins, every plugin that is targeting the object will patch it in the order the loading happened
- When an object unregisters itself from the plugins, every plugin that is targeting the object undoes its patches in the order the loading happened
- Some plugins may need to change their behavior after `set_options` is called; for the changes to take place, `set_options` should be called with `refresh` set to `True`

## OpenMP Plugin

**Plugin Behavior**
- The plugin keeps track of any applied versions and only applies other versions when specified in the options; any applied version persists until the unregister method is called

**Classes**
- Treating TextX objects is done in the syntactic stage via classes that inherit from `OmpTxNode`, which itself inherits from `OmpNode`; visiting these classes results in instances of their respective classes that inherit from `OmpNode`
- Each OpenMP version has dedicated syntactic/semantic parsers and printers
- Printers use the raw statement of an OpenMP object provided in the comment; the segmentation of these statements allows for the suppression or tweaking of printing of segments correlated to the desired OpenMP object


Hellio404:
Hello @EmilyBourne We just started rewriting the way we handle OpenMP annotated comments.

We want to support multiple versions. with the previous implementation, it was really hard and there were a lot of issues and hardcoded cases.

With this method, we used a similar method that Pyccel use (the visitor design pattern) so each Construct / Clause has its own way to check and make sure its arguments are correct.

we also use inheritance to eliminate code duplication when we try to support another version, we only need to override a small set of functions/variables from one version to version.

I just want an initial review of the implementation we used before we start implementing the code for Constructs/Clauses

pyccel/parser/grammar/openmp.tx
```diff
@@ -1,184 +1,102 @@
...
-ThreadIndex: (ID | INT);
-TaskIndex: (ID | INT);
-TeamIndex: (ID | INT);
-PriorityIndex: (ID | INT);
-NotaStmt: /.*$/;
+//              Omp Versions
 ////////////////////////////////////////////////////
+OMP_4_5: '';
+OMP_5_0: '';
+OMP_5_1: '';
```
EmilyBourne, line 102: Am I right in thinking that any new OpenMP versions will lead to big `textx` errors until the new constructs are implemented?

Hellio404: I don't see where the problem will be. When a new version is out we don't need to change anything if we don't want to support that version yet. if the user did use syntax from the new version he will get an error that the thing he uses is unknown the same as if he uses a wrong syntax and that's it

pyccel/parser/syntax/omp_versions/version_4_5.py
```diff
@@ -0,0 +1,585 @@
...
+    OmpEndConstruct,
+)
+from pyccel.ast.variable import Variable
+from pyccel.parser.extend_tree import extend_tree
+from pyccel.ast.datatypes import NativeInteger, NativeVoid
+
+__all__ = ("Openmp",)
+
+
+class Openmp:
```
EmilyBourne, line 20: Am I right that this class is behaving as a namespace?

pyccel/parser/grammar/openmp.tx
EmilyBourne: No, textx errors are now caught https://github.com/pyccel/pyccel/blob/rewrite-omp/pyccel/parser/syntax/openmp.py#L59

pyccel/parser/syntax/omp_versions/version_4_5.py
Hellio404: Yes, it is. it's just a class that groups classes from the same version together and also help new version to inherit the same classes of the previous version

so for example version_5_0 can just have class Openmp(version_4_5.Openmp) and this will make version_5_0 Openmp class fully ready to be used by Textx and if we can only override the classes that got changes in version_5_1

yguclu:
TODO: resolve conflicts with `master` branch

EmilyBourne:
@rmahjoubi I see that you were assigned to this PR in March. Have you worked on it at all? @Hellio404 would it be possible to give a quick summary of what is missing from this PR?

rmahjoubi:
@EmilyBourne I took a look at what @Hellio404 did, I think with this design it's reasonably easy to add the new versions of openmp.
There are some elements missing from this PR, some are mentioned in todo comments, including some limitations in the grammar, and the ability to get the version of openmp from the compiler, which i think on most compilers is unfeasible, since implementations of new versions of openmp comes in chunks.
Also there are  edge cases and bugs such as when we have Consecutive Construcs, they would get treated as CommentMultiLine, which causes failure in the semantics of the Constructs, also nested directives of the same kind doesn't work.
The major problem I found is that the directives are not aware of the CodeBlock they control, and this will cause a lot unexpected behavior in the generated code. for example:
```
import numpy as np 

x = np.array([1,2,3,4])
y = np.array([1,2,3,4])
#$ pragma omp parallel
#$ pragma omp single
z  = x + y
#$ pragma omp end single
#$ pragma omp single 
t  = x + y
#$ pragma omp end single
```
In the code we generate we'll use the same tmp variables in both single blocks which will create unexpected race condtions.
Finally the majority of the constructs and clauses needs to be implemented

EmilyBourne:
@rmahjoubi Thanks for looking into this. It sounds like the PR is still useful in that case. I suspect the safest way to specify the OpenMP version would be in the compiler config file (https://github.com/pyccel/pyccel/blob/devel/docs/compiler.md#user-defined-compiler).

> The major problem I found is that the directives are not aware of the CodeBlock they control, and this will cause a lot unexpected behavior in the generated code. for example:

Is this a new problem compared to the previous implementation?

Do you plan on working on this?

EmilyBourne:
> > Is this a new problem compared to the previous implementation?
> 
> I think both implementations have this problem. Yes I will work on this.

Ok great. It looks like it's been a while since this branch was merged. Let me know if there's anything unfamiliar that you need a hand merging in

rmahjoubi:
Hello @EmilyBourne,

I'm looking for an overall feedback this new design, which is still in need of some polishing, documentation, and refactoring. Below is an overview of the current design and its status:

### Status
- **Grammar**: Majority of OpenMP 4.5 features are now supported.
- **Syntactic Parsing**: Compatible with OpenMP 4.5 and OpenMP 5.0.
- **Semantic Parsing**: Basic functionality has been implemented. I'm uncertain if implementing all semantic validations of OpenMP is feasible.
- **Printers**: Compatible with OpenMP 4.5 and OpenMP 5.0.

### Design Overview
- **OmpAnnotatedComment**: This component groups functionalities common to all OpenMP objects.
- **OmpConstruct**: Represents a directive associated with a code block and, optionally, an end directive.
- **Version-Specific Tools**: Each OpenMP version has dedicated syntactic/semantic parsers and printers.
- **Parsing**: Utilizes classes that dynamically inherit from Pyccel's parsers and those from an OpenMP version.
- **Specialized Directives and Clauses**: OpenMP versions can define more specialized directive and clause classes to handle exceptional cases.
- **Inheritance Hierarchy**:
  - Directive classes should inherit from `OmpDirective`.
  - Clause classes should inherit from `OmpClause`.
- **Printers Functionality**: These use the raw statement of an OpenMP object provided in the comment. The segmentation of these statements allows for the suppression or tweaking of printing of segments correlated to the desired OpenMP object.
- **Printers**: Similar to parsers, classes dynamically inherit from Pyccel's printers and those specific to an OpenMP version.

EmilyBourne (commented):
Posting old comments

max-models:
I tried compiling some of my kernels with this branch and I am getting this error:

bar.py:

```
def f(x: 'float[:]'):
    N: int = x.shape[0]
    #$ omp target teams distribute parallel for map(tofrom: x[:])
    for i in range(N):
        x[i] += 1.0
```

```
❯ pyccel --language=fortran --openmp bar.py

pyccel:
 |fatal [syntactic]: bar.py [3,4]| Expected 'if' or 'device' or 'private' or 'firstprivate' or 'map' or 'is_device_ptr' or 'defaultmap' or 'nowait' or 'depend' or EOF
```

max-models:
I also tried the example in the docs

```
from numpy import zeros
from pyccel.stdlib.internal.openmp import omp_get_team_num
n = 8
threadlimit = 4
a = zeros(n, dtype=int)
#$ omp target
#$ omp teams num_teams(2) thread_limit(threadlimit)
#$ omp distribute
for i in range(0, n):
    a[i]    = omp_get_team_num()
#$ omp end teams
#$ omp end target

for i in range(0, n):
    print("Team num :", a[i])
```


```
❯ pyccel --language=fortran --openmp bar.py --verbose

pyccel:
 |fatal [syntactic]: bar.py [7]| Expected 'num_teams'
```

rmahjoubi:
> ```python
> #$ omp target
> #$ omp teams num_teams(2) thread_limit(threadlimit)
> #$ omp distribute
> for i in range(0, n):
>     a[i]    = omp_get_team_num()
> #$ omp end teams
> #$ omp end target
> ```

There was an error in the grammar, It is fixed now.

> ```
> def f(x: 'float[:]'):
>     N: int = x.shape[0]
>     #$ omp target teams distribute parallel for map(tofrom: x[:])
>     for i in range(N):
>         x[i] += 1.0
> ```

There are constructs that are not defined in the grammar yet including this one, It's added, but I still didn't figure out how to support array sectioning yet, so for now only the following does work:
```python
def f(x: 'float'):
    #$ omp target teams distribute parallel for map(tofrom: x)
    for i in range(10):
        x += 1.0
```

EmilyBourne:
> There are constructs that are not defined in the grammar yet including this one, It's added, but I still didn't figure out how to support array sectioning yet

Given the size of this PR it's probably better to not think about this too much here. Having a first version should be the priority. We can make a list of improvements like array sectioning that can be done in separate PRs. Potentially these can then be tackled in parallel if they are independent tasks

max-models:
> > ```python
> > #$ omp target
> > #$ omp teams num_teams(2) thread_limit(threadlimit)
> > #$ omp distribute
> > for i in range(0, n):
> >     a[i]    = omp_get_team_num()
> > #$ omp end teams
> > #$ omp end target
> > ```
> 
> There was an error in the grammar, It is fixed now.
> 
> > ```
> > def f(x: 'float[:]'):
> >     N: int = x.shape[0]
> >     #$ omp target teams distribute parallel for map(tofrom: x[:])
> >     for i in range(N):
> >         x[i] += 1.0
> > ```
> 
> There are constructs that are not defined in the grammar yet including this one, It's added, but I still didn't figure out how to support array sectioning yet, so for now only the following does work:
> 
> ```python
> def f(x: 'float'):
>     #$ omp target teams distribute parallel for map(tofrom: x)
>     for i in range(10):
>         x += 1.0
> ```

Ok, thanks for the fix. I have a small question about your last code snippet, should it be possible to compile it without the `--openmp` flag? Currently, it is not for me:

```
(env_struphy) [maxlin@viper12 struphy]$ cat foo.py
def f(x: 'float'):
    #$ omp target teams distribute parallel for map(tofrom: x)
    for i in range(10):
        x += 1.0
(env_struphy) [maxlin@viper12 struphy]$ pyccel --openmp foo.py
(env_struphy) [maxlin@viper12 struphy]$ pyccel foo.py

pyccel:
 |fatal [syntactic]: foo.py [2,4]| Annotated comments must start with omp, acc or header
```

rmahjoubi:
> Ok, thanks for the fix. I have a small question about your last code snippet, should it be possible to compile it without the `--openmp` flag? Currently, it is not for me (it seems to also be the case for any `#$ omp` annotation):
> 
> ```
> (env_struphy) [maxlin@viper12 struphy]$ cat foo.py
> def f(x: 'float'):
>     #$ omp target teams distribute parallel for map(tofrom: x)
>     for i in range(10):
>         x += 1.0
> (env_struphy) [maxlin@viper12 struphy]$ pyccel --openmp foo.py
> (env_struphy) [maxlin@viper12 struphy]$ pyccel foo.py
> 
> ERROR at parsing (syntax) stage
> pyccel:
>  |fatal [syntactic]: foo.py [2,4]| Annotated comments must start with omp, acc or header
> ```

You are welcome. It's not possible without the --openmp flag, the error message should be changed. The flag tells pyccel to plug openmp functionality, to unplug openmp we should use the Plugins.handle_loading, for example:

```python

def omp_arraysum_combined(x : 'int[:]'):
    func_result = 0
    #$ omp parallel for reduction (+:func_result)
    for i in range(0, 5):
        func_result += x[i]
    return func_result

f1 = epyccel(test, fflags = '-Wall',accelerators=['openmp'], language='c') #openmp is loaded
from pyccel.utilities.plugins import Plugins
plugin.handle_loading({'clear': True}) #openmp is unloaded
```

```python
def f(x: 'float'):
    #$ omp target teams distribute parallel for map(tofrom: x)
    for i in range(10):
        x += 1.0
```

A reduction clause is needed for the above to work, and I think this will default to the host device because of #798

EmilyBourne:
> It's not possible without the --openmp flag, the error message should be changed.

This is strange. If the `--openmp` flag is not provided then `#$ omp` should be seen as a comment as in all other languages. Maybe a warning could be raised though to warn the user that their openmp command is being ignored

rmahjoubi:
> This is strange. If the `--openmp` flag is not provided then `#$ omp` should be seen as a comment as in all other languages. Maybe a warning could be raised though to warn the user that their openmp command is being ignored

This is the behavior for every comment that starts with #$, I changed the behavior to raise a warning if it's an omp comment, otherwise it's seen as a comment.

rmahjoubi:
/bot run spelling pyccel_lint pylint docs coverage

pyccel/plugins/Openmp/plugin.py
```diff
@@ -0,0 +1,130 @@
...
+    """
+    __slots__ = ("_options", "_loaded_versions")
+    DEFAULT_VERSION = 4.5
+
+    VERSION_MODULES = {
+        4.5: openmp_4_5,
+        5.0: openmp_5_0
+    }
+
+    PARSER_TYPES = [SyntaxParser, SemanticParser, CCodePrinter, FCodePrinter, PythonCodePrinter]
```
EmilyBourne, line 43: This will be the cause of your race conditions.
It would be better to add the methods to an instance of the class instead of adding them to the type.

For example each parser could call a get_plugins method in the constructor. Or they could be passed as arguments. They can then pass themselves as arguments to apply_patches

rmahjoubi: E       FileNotFoundError: [Errno 2] No such file or directory: '/opt/hostedtoolcache/Python/3.9.22/x64/lib/python3.9/site-packages/pyccel/plugins/Openmp/grammar/openmp.tx'


this is what causing the tests to fail, so i added the .tx files here 47a5dcd to be included in the build, but didn't seem  to fix it.
but i think what you suggested is better anyways

EmilyBourne: I suspect you have both problems.  Once you ran in parallel you had a lot fewer failures didn't you?

rmahjoubi:
/bot run spelling pyccel_lint pylint docs coverage

docs/openmp.md
```diff
@@ -1,5 +1,42 @@
...
+Specify the OpenMP version during compilation using the `--omp-version` flag:
+
+```bash
+--omp-version=4.5
+# or
+--omp-version=5.0
+```
+
+### Epyccel Integration
+When using Epyccel, configure OpenMP through the accelerators' parameter:
```
EmilyBourne, line 17:
```suggestion
When using `epyccel`, configure OpenMP through the accelerators' parameter:
```

rmahjoubi:
/bot run pylint linux spelling docs coverage intel macosx windows

rmahjoubi:
/bot run pylint linux spelling docs coverage intel macosx windows

pyccel/extensions/Openmp/ast.py
```diff
@@ -0,0 +1,402 @@
...
+            # fetch the omp version from the root model, it will not be fetched in after the syntactic stage
+            p = self
+            while hasattr(p, 'parent'):
+                p = p.parent
+            if hasattr(p, 'omp_version'):
+                self._omp_version = p.omp_version
+            else:
+                raise NotImplementedError(
+                    "OpenMP version not set"
+                )
```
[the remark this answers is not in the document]
rmahjoubi: fixed here ea900b6

pyccel/extensions/Openmp/ast.py
```diff
@@ -0,0 +1,402 @@
...
+
+    @property
+    def line(self):
+        """
+        returns the line of an omp object from the root parent OmpDirective.
+        """
+        p = self
+        while not isinstance(p, OmpDirective):
+            p = p.parent
+        return p.python_ast.lineno
```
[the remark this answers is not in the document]
rmahjoubi: fixed here https://github.com/pyccel/pyccel/blob/ea900b60416f922359d4e7ee86ade111dbe378ff/pyccel/plugins/Openmp/omp.py#L315

pyccel/extensions/Openmp/ast.py
```diff
@@ -0,0 +1,402 @@
...
+    """
+    Represents an every OpenMP Directive.
+
+    Parameters
+    ----------
+    name: str
+          The name of the directive
+
+    require_end_directive bool
+          True if the directive is syntactically incorrect without a corresponding end directive
```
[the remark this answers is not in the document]
rmahjoubi: This code no longer exists

pyccel/extensions/Openmp/extension.py
```diff
@@ -0,0 +1,84 @@
...
+        omp_version = self._options.get('omp_version', self._default_version)
+        mixin = openmp_5_0.SemanticParser if omp_version == 5.0 else openmp_4_5.SemanticParser
+
+        class Extended(sp, mixin):
+            def __init__(self, *args, **kwargs):
+                mixin.__init__(self)
+                sp.__init__(self, *args, **kwargs)
+            def _visit(self, expr):
+                if expr in self.omp_reserved_nodes:
+                    return EmptyNode()
```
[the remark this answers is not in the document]
rmahjoubi: This code no longer exists

pyccel/extensions/Openmp/extension.py
```diff
@@ -0,0 +1,84 @@
...
+                else:
+                    return sp._visit(self, expr)
+        return Extended
+
+    def extend_printer(self, printer):
+        if 'openmp' not in self._options.get('accelerators', []):
+                return printer
+
+        #don't extend
+        omp_version = self._options.get('omp_version', self._default_version)
```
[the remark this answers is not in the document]
rmahjoubi: I went with an environment variable

https://github.com/pyccel/pyccel/blob/ea900b60416f922359d4e7ee86ade111dbe378ff/pyccel/plugins/Openmp/plugin.py#L121

pyccel/extensions/Openmp/ast.py
```diff
@@ -0,0 +1,402 @@
...
+class OmpAnnotatedComment(PyccelAstNode):
+    """Parent class for all openmp classes, including classes used in the grammar."""
+
+    __slots__ = ("_parent", "_raw", "_position", "_line", "_version", "_deprecated", "_omp_version")
+    _attribute_nodes = ()
+
+    def __init__(self, **kwargs):
+        super().__init__()
+        self._parent = kwargs.pop("parent", None)
+        self._omp_version = kwargs.pop("omp_version", None)
```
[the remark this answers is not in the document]
rmahjoubi: This code no longer exists

pyccel/plugins/Openmp/plugin.py
rmahjoubi: fixed here 94277ae

EmilyBourne:
@rmahjoubi maybe we should close this PR and open another one based on the same branch so you are the official author and @Hellio404 doesn't get spammed?
I think it would make more sense for credit anyway.

pyccel/codegen/pipeline.py
```diff
@@ -119,6 +121,8 @@ def execute_pyccel(fname, *,
         This can allow certain constants to be defined outside of the function passed to epyccel.
     """
     start = time.time()
+    plugins = Plugins()
+    plugins.set_options({'accelerators':accelerators, 'omp_version':omp_version, 'language':language})
```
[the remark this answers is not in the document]
rmahjoubi: > For me a plugin implies that external plugins whose contents are unknown could be supported in the future. > > Does your framework allow for this?

Yes this was my intention for the framework. I thought I'd allow the plugins to define their CLI arguments and as you mentioned have any extra arguments passed to epyccel as `**kwargs`, but there are arguments that are needed by both pyccel and the plugins which made it a little complicated, like `openmp`  argument which also control the compilation flags, which ideally should be done via the plugin.

pyccel/errors/messages.py
```diff
@@ -161,3 +161,10 @@
 WRONG_LINSPACE_ENDPOINT = 'endpoint argument must be boolean'
 NON_LITERAL_KEEP_DIMS = 'keep_dims argument must be a literal, otherwise rank is unknown'
 NON_LITERAL_AXIS = 'axis argument must be a literal, otherwise pyccel cannot determine which dimension to operate on'
+
+#openmp
+OMP_VERSION_NOT_SUPPORTED = "OPENMP {} is not supported. Defaulting to OPENMP {}."
+
+#plugins
+PLUGIN_DIRECTORY_NOT_FOUND = "Plugin directory not found: {}."
```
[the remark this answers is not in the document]
rmahjoubi: yes they are both used two times each

pyccel/utilities/plugins.py
```diff
@@ -0,0 +1,205 @@
...
+    "PatchRegistry",
+    "Plugin",
+    "Plugins",
+)
+
+errors = Errors()
+
+
+@dataclass
+class PatchInfo:
```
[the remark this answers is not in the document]
rmahjoubi: I used it for classes that are only meant to store data

Issue #1209, opened 2022-10-08
Missing temporary creation on return for C


EmilyBourne:
**Describe the bug**
When returning an array expression in C, a temporary must be created to generate correct loop unravelling.

**To Reproduce**
The following example fails:
```python
def arr_mult_scalar(T: 'int[:]', t: int = 13):
    return T * t
```
but this succeeds:
```python
def arr_mult_scalar(T: 'int[:]', t: int = 13):
    x = T * t
    return x
```

In the failing case the generated code is:
```c
t_ndarray arr_mult_scalar(t_ndarray T, int64_t t_0001)
{
    Out_0001 = array_create(1, (int64_t[]){T.shape[INT64_C(0)]}, nd_int64, false);
    return T * t_0001;
}
```

**Expected behaviour**
A temporary should be created and the loop unravelling functions should be called on the expression

**Language**
C

PR #1231, opened 2022-11-14
Add Missing temporary creation on return for C

Pinkyboi:
Fixes [#1209](https://github.com/pyccel/pyccel/issues/1209).
- Add missing temporary array creation on return, for the cases that involve array loop unraveling.

### DIFF
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -14,7 +14,7 @@ from pyccel.ast.builtins  import PythonRange, PythonComplex
 from pyccel.ast.builtins  import PythonPrint, PythonType
 from pyccel.ast.builtins  import PythonList, PythonTuple
 
-from pyccel.ast.core      import For, CodeBlock
+from pyccel.ast.core      import Declare, For, CodeBlock
 from pyccel.ast.core      import FuncAddressDeclare, FunctionCall, FunctionCallArgument, FunctionDef
 from pyccel.ast.core      import Deallocate
 from pyccel.ast.core      import FunctionAddress, FunctionDefArgument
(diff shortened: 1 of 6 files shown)


Pinkyboi:
@EmilyBourne i not really sure about that fix i did, could you, when you are free, see for me if it could cause problem ?

EmilyBourne:
> @EmilyBourne i not really sure about that fix i did, could you, when you are free, see for me if it could cause problem ?

This looks right to me

pyccel/codegen/printing/ccode.py
```diff
@@ -1600,13 +1600,13 @@ def _print_Return(self, expr):
...
                 variables = last_assign.rhs.get_attribute_nodes(Variable)
-                unneeded_var = not any(b in vars_in_deallocate_nodes for b in variables)
+                unneeded_var = not any(b in vars_in_deallocate_nodes or b.is_ndarray for b in variables)
                 if unneeded_var:
                     code = ''.join(self._print(a) for a in expr.stmt.body if a is not last_assign)
                     return code + 'return {};\n'.format(self._print(last_assign.rhs))
                 else:
-                    code = ''+self._print(expr.stmt)
                     last_assign.lhs.is_temp = False
+                    code = ''+self._print(expr.stmt)
```
EmilyBourne, line 1609: I'm not sure why you moved this line, but as you did you must have looked at it. Do you know why it is `code = ''+self._print(expr.stmt)` and not simply `code = self._print(expr.stmt)`?

EmilyBourne (approved):

pyccel/codegen/printing/ccode.py
Pinkyboi: I swapped the line order because the last assign should become a non temporary variable for it to be declared, if a loop unraveling happens during the printing the last assign lhs will become an indexed element that doesn't have the attribute is_temp, that's why i swapped order.
And i agree it should be `code = self._print(expr.stmt)` i'll fix that.

Pinkyboi: Fixed in [52d0f45](https://github.com/pyccel/pyccel/pull/1231/commits/52d0f4520b9dfafb8462427d007880650990f96a).

yguclu (approved):

PR #1234, opened 2022-11-18
Auto-trigger benchmarks

EmilyBourne:
Add empty commit to pyccel-benchmarks repo to trigger benchmarks.

Use utf-8 encoding instead of ascii to handle accents

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -2,7 +2,7 @@ name: Benchmarks
 
 on:
   push:
-    branches: [ master ]
+    branches: [ bench ]
 
 jobs:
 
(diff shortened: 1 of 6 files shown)


.github/workflows/bench.yml
```diff
@@ -11,30 +11,18 @@ jobs:
...
-        working-directory: ./.
-      - name: Add & Commit
-        uses: EndBug/add-and-commit@v9.0.0
-        with:
-          message: 'Update performance comparison'
-          add: 'performance.md'
-          default_author: github_actions
+          git config user.email "yaman.guclu@gmail.com"
+          git config user.name "Yaman Güçlü"
+          git commit --allow-empty -m "Benchmark of pyccel@${GITHUB_SHA}"
```
yguclu, line 27: I've just checked and GitHub generates the wrong hyperlink from `GITHUB_SHA`. 

For example, in commit [459c58b](https://github.com/pyccel/pyccel-benchmarks/commit/459c58b9fef51989cf2fc4feffde67fc7d05174f), instead of [this correct link](https://github.com/pyccel/pyccel/commit/dfeca799e36bc0f44ce290bc6e8b94fb5361b94c) it creates [this wrong link](https://github.com/pyccel/pyccel-benchmarks/commit/dfeca799e36bc0f44ce290bc6e8b94fb5361b94c).

We probably need to be more specific when we construct this link.

yguclu (changes requested):

.github/workflows/bench.yml
EmilyBourne: Fixed [381e485](https://github.com/pyccel/pyccel/pull/1234/commits/381e485c7bd24a02b3b254a60dfc40eeb0c03283)

yguclu (approved):

PR #1235, opened 2022-11-18
Remove temporary variable for CudaDeviceSynchronize

Pinkyboi:
Remove the Temporary variable that is created when calling any void `PyccelInternalFunction`.


Pinkyboi:
@bauom Is there any way to activate basic pyccel tests on this branch too ?

bauom:
> @bauom Is there any way to activate basic pyccel tests on this branch too ?

the checks should be already activated.
i think its just because you changed the base branch.
you will see the test on the next commit.

EmilyBourne:
I agree. This should be possible in the master branch. You will need different tests though. The tests are run when you push to a PR which targets master. If you change the target and update the branch they should trigger. It may be easier to create a new branch though?

Pinkyboi:
@EmilyBourne I will add a new PR to master with that fix then.

Pinkyboi:
@bauom @EmilyBourne sorry i kinda forgot about this PR, but i wanted to signal that i don't think those changes can be done in master, since there's no `PyccelInternalFunction` with a type of void, so in the meantime i could just fix it for the cuda branch.

EmilyBourne:
> @bauom @EmilyBourne sorry i kinda forgot about this PR, but i wanted to signal that i don't think those changes can be done in master, since there's no `PyccelInternalFunction` with a type of void there, so the coverage test will fail, i would suggest fixing it only in the cuda branch for now.

Ah you're right. That makes it hard to add tests. In that case I am ok with this staying in the cuda branch

EmilyBourne:
@Pinkyboi please update your branch and fix the collisions

pyccel/parser/semantic.py
```diff
@@ -2574,6 +2574,8 @@ def _visit_Assign(self, expr, **settings):
                         bounding_box=(self._current_fst_node.lineno, self._current_fst_node.col_offset),
                         severity='error')
                     return None
+            if lhs.is_temp and rhs.dtype is NativeVoid():
```
bauom, line 2577: tests are failing on this condition.

bauom (changes requested):

pyccel/parser/semantic.py
Pinkyboi: Noticed that, on it.

Pinkyboi: I think i fixed it in [42c9942](https://github.com/pyccel/pyccel/pull/1235/commits/42c994203c1136fb2642cb9972c317d421389383).

EmilyBourne (approved):

bauom (changes requested):
tests should only include the specified part as to not fail because of a unrelated issue

tests/internal/scripts/ccuda/cuda_array_managed.py
```diff
@@ -11,7 +11,7 @@ def square(a):
     threads_per_block = 5
     n_blocks = 1
     a = cuda.array([0,1,2,3,4], memory_location = 'managed')
-    cuda.synchronize()
-    square[n_blocks, threads_per_block](a)
-    cuda.synchronize()
+    cuda.deviceSynchronize()
+    square[5, 1](a)
+    cuda.deviceSynchronize()
```
bauom, line 16: the naming changed in #1274 from `deviceSynchronize` to `synchronize`

bauom (changes requested):
tests should only compile the specified part as to not fail because of a unrelated issue

tests/internal/scripts/ccuda/cuda_array_managed.py
Pinkyboi: Fixed in [8ef5990](https://github.com/pyccel/pyccel/pull/1235/commits/8ef59906f98230ce1b03545e6817e8279ce751d3).

Pinkyboi:
> tests should only include the specified part as to not fail because of a unrelated issue

@bauom i added in [5f32e4f](https://github.com/pyccel/pyccel/pull/1235/commits/5f32e4fd0222f299587db3a043bba69379ce0d17) a test containing only `cuda.synchronize()` i suppose that solves the issue ?

tests/internal/scripts/ccuda/cuda_synchronize.py
```diff
@@ -0,0 +1,5 @@
+from pyccel import cuda
+
+if __name__ == '__main__':
+    cuda.synchronize()
+    cuda.synchronize()
```
bauom, line 5: why are you  using 2 calls to `cuda.synchronize()`

bauom (changes requested):

tests/internal/scripts/ccuda/cuda_synchronize.py
Pinkyboi: No reason in particular, its just they are supposed to be used by pair when calling a kernel launch i thought that maybe leaving this way wasn't a problem.

Pinkyboi: Removed in [2a851ed](https://github.com/pyccel/pyccel/pull/1235/commits/2a851edd0ce5fd41a0251345c43b7ab0b462fd1c).

EmilyBourne:
Following the discussion [here](https://github.com/pyccel/pyccel/pull/1294#discussion_r1068628380), this PR could be merged to master, if `PythonPrint` is modified to inherit from `PyccelInternalFunction` and behave like every other function in pyccel

PR #1236, opened 2022-11-18
Enhancements to cuda array creation

Pinkyboi:
For now in copy_CudaArray_Data the cudaMemcpyKind field of cudaMemcpy is hardcoded, this PR aims to make it so that field can be deduced from the variable's memory location and the current context of the function, as well as other fixes to make array creation in cuda easier.

- [x]  Fixes compilation issues caused by the previous cuda_memcopy PR.
- [x]  Raise an error when creating a `Host` array in `Device` function.
- [x] Create temporary variable when using `numpy.array`, `cuda.array`, `cupy.array` calls as arguments.


Pinkyboi:
@bauom for this PR i added some function `numba` functions, do you think that its good to leave them in `codaext` for now or should i make and new extension for `numba` ?

bauom:
> @bauom for this PR i added some function `numba` functions, do you think that its good to leave them in `codaext` for now or should i make and new extension for `numba` ?

if you are using the exact `numba` syntax it would be much better to have them in `numbaext` so we can start building that up too.

Pinkyboi:
@bauom I removed the two `numba` functions from this PR, i'll create a new PR so we can start building a `numba` extension.

EmilyBourne:
Marked as draft as tests are failing

Pinkyboi:
@EmilyBourne Those failed test are related to openMP, not related to my PR, i'm not sure about why that occurs.

tests/internal/scripts/ccuda/cuda_to_host_device.py
```diff
@@ -0,0 +1,23 @@
...
+if __name__ == '__main__':
+    threads_per_block = 5
+    n_blocks = 1
+    a = np.array([0,1,2,3,4])
+    b = cuda.to_device(a)
+    c = cuda.to_device(np.array([4,3,2,1,0]))
+    cuda.deviceSynchronize()
+    mult[n_blocks, threads_per_block](a, c)
+    cuda.deviceSynchronize()
+    c = cuda.to_host(b)
```
bauom, line 21: if you could split these into separated tests.

bauom (changes requested):

EmilyBourne:
> @EmilyBourne Those failed test are related to openMP, not related to my PR, i'm not sure about why that occurs.

Are the tests also failing in `cuda_main_temp`? If so it would probably be a good idea to open an issue and/or mark then as `xfail`.
If not then the failures must be caused by your changes.

It looks like openmp is not installed but it says that it is. Can you try putting 
`brew reinstall libomp` [here](https://github.com/pyccel/pyccel/blob/master/.github/actions/macos_install/action.yml#L9)

EmilyBourne:
> @EmilyBourne Those failed test are related to openMP, not related to my PR, i'm not sure about why that occurs.

The codacy tests are also failing which is not related to OpenMP

.github/actions/macos_install/action.yml
```diff
@@ -6,7 +6,7 @@ runs:
     - name: Install MPI, OpenMP
       run: |
         brew install open-mpi
-        brew install libomp
+        brew reinstall libomp
```
EmilyBourne, line 9: I made this change, but it didn't fix the problem. It seems that the problem has been fixed upstream though, so this change can be reverted

pyccel/parser/semantic.py
```diff
@@ -718,7 +720,8 @@ def _handle_function_args(self, arguments, **settings):
             if isinstance(a.value, StarredArguments):
                 args.extend([FunctionCallArgument(av) for av in a.value.args_var])
             else:
-                if isinstance(a.value, PyccelArithmeticOperator) and a.value.rank:
+                if isinstance(a.value, PyccelArithmeticOperator) and a.value.rank\
+                    or isinstance(a.value, (NumpyNewArray, CudaNewArray)):
```
EmilyBourne, line 724: Please add parentheses to separate the and/or statements for readability

pyccel/parser/semantic.py
```diff
@@ -823,6 +826,19 @@ def _handle_function(self, expr, func, args, **settings):
                         severity = 'fatal')
 
             args, kwargs = split_positional_keyword_arguments(*args)
+            if func in (CupyArray, CudaArray, NumpyArray):
+                if 'device' in self.scope.decorators or 'kernel' in self.scope.decorators:
```
EmilyBourne, line 830: I don't see these decorators in the `decorators.py` file. Are these cupy objects?

pyccel/parser/semantic.py
```diff
@@ -823,6 +826,19 @@ def _handle_function(self, expr, func, args, **settings):
...
+                    current_context = 'device'
+                    if ('memory_location' in kwargs.keys() and kwargs['memory_location'] == 'host')\
```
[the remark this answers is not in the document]
EmilyBourne: Or you could use `getattr` if there is an obvious default value:
```suggestion
                    if (kwargs.get('memory_location', current_context)== 'host') \
```

pyccel/parser/semantic.py
```diff
@@ -495,6 +496,7 @@ def _infere_type(self, expr, **settings):
             d_var['datatype'   ] = expr.dtype
             d_var['memory_handling'] = 'heap' if expr.rank > 0 else 'stack'
             d_var['memory_location'] = expr.memory_location
+            d_var['current_context'] = expr.current_context
```
EmilyBourne, line 499: Do you think this option should be added throughout this function?

tests/internal/scripts/ccuda/cuda_array_variable.py
```diff
@@ -0,0 +1,19 @@
...
+@kernel
+@types('int[:]')
+def square(a):
+    index = cuda.blockIdx(0) * cuda.blockDim(0) + cuda.threadIdx(0)
+    a[index] = a[index] * a[index]
+
+if __name__ == '__main__':
+    threads_per_block = 5
+    n_blocks = 1
+    arr_var = [0,1,2,3,4]
```
EmilyBourne, line 15: Please try to use officially supported syntax in tests. Here you can use a tuple.

```suggestion
    arr_var = (0,1,2,3,4)
```

pyccel/parser/semantic.py
```diff
@@ -95,7 +96,7 @@
 from pyccel.ast.numpyext import NumpyNewArray, NumpyNonZero
 from pyccel.ast.numpyext import DtypePrecisionToCastFunction
 
-from pyccel.ast.cudaext import CudaNewArray, CudaThreadIdx, CudaBlockDim, CudaBlockIdx, CudaGridDim
+from pyccel.ast.cudaext import CudaArray, CudaNewArray, CudaThreadIdx, CudaBlockDim, CudaBlockIdx, CudaGridDim
```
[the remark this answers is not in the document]
Pinkyboi: `CupyArray` inherits from `CudaArray`, `CupyArray` is used in the `Cupy` extension that aims to replicate the standard `Cupy` syntax, the `CUDA` extension on the other hand gives more freedom to the user but uses functions that we came up with (no standard like `Cupy`).

pyccel/parser/semantic.py
Pinkyboi: It's odd, i think it should be added to `decorators.py`.

pyccel/parser/semantic.py
Pinkyboi: To be honest i think that the current_context should be checked in semantic, in a similar way to how `PyccelFunctionDef` are visited, i was going to ask for you opinion on that.

.github/actions/macos_install/action.yml
Pinkyboi: Done in [9ea3abf](https://github.com/pyccel/pyccel/pull/1236/commits/9ea3abfa46b3d032e2791954f51364c4f7cb8c54).

pyccel/parser/semantic.py
Pinkyboi: Done in [bec781c](https://github.com/pyccel/pyccel/pull/1236/commits/bec781ca41cf5261a9fa7a5010ff9052b8502d62).

pyccel/ast/cudaext.py
```diff
@@ -237,6 +250,11 @@ class CudaBlockDim(CudaInternalVar)         : pass
 class CudaBlockIdx(CudaInternalVar)         : pass
 class CudaGridDim(CudaInternalVar)          : pass
 class CudaGrid(PyccelAstNode)               :
+    """Represents a call to cuda.grid for code generation.
+        dim  : int
+        The number of dimensions of requested, it should correspond
+        to the number of dimensions declared when instantiating the kernel
```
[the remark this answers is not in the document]
Pinkyboi: I changed it here [8b89bba](https://github.com/pyccel/pyccel/pull/1236/commits/8b89bba94a271b78c15be15a9aa9ba4024e2a1b7), i'm not sure if that's what you meant by numpy's convention.

pyccel/parser/semantic.py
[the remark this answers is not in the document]
Pinkyboi: Done here [6f16b4e](https://github.com/pyccel/pyccel/pull/1236/commits/6f16b4e6984efe175879501cda8c5f9421d5a20a).

tests/internal/scripts/ccuda/cuda_array_variable.py
Pinkyboi: Does that mean that i should replace all lists with tuples, or only the ones assigned to variables ?

Pinkyboi: Changed it to tuple here [b8643d2](https://github.com/pyccel/pyccel/pull/1236/commits/b8643d26a6fe5c53a8fe34eb79d67b376d89109c).

pyccel/ast/cudaext.py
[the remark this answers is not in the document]
EmilyBourne: @jalalium is the expert, but I believe this is correct.

pyccel/parser/semantic.py
EmilyBourne: Hmmm, ok. I also realised the `kernel` decorator is there, it's just out of order in the `__all__` variable in decorators.py (the rest are alphabetical)

tests/internal/scripts/ccuda/cuda_array_variable.py
EmilyBourne: Generally tuples should be used instead of lists, but we tolerate lists when there is a strong convention (e.g `np.array([1,2,3])`, it is very rare to use `np.array((1,2,3))` so it may be harder to recognise immediately)

pyccel/parser/semantic.py
EmilyBourne: I'm not sure what you mean

Pinkyboi: I'm not sure if i responded to your interrogation then, where should it be added instead in your opinion ?

EmilyBourne: My question was whether every `if` branch of `_infere_type` should set the `current_context` key.

I would still like to know what you were considering though

Pinkyboi: It's only needed for `CUDA` arrays i don't think its relevant for the others.
I was considering that maybe the way i'm handling `current_context` now is not optimal, maybe if we could "visit" the `CUDA` objects that need the current context (to raise errors or to allocate in the right manner), i would be better instead of adding if statements in function that are not supposed to handle those cases, i'll probably discuss that with @bauom during our next meeting.

EmilyBourne: > It's only needed for CUDA arrays i don't think its relevant for the others.

I'm not sure I agree. It is relevant for `Variable` creation. However I think you are right that the others don't need to specify it and can rely on the default value.

Visiting the CUDA objects definitely sounds like a good idea. I agree that it would be better to have a `_visit_CupyArray` function etc. I'm not sure what the best way to handle the `NumpyArray` case would be though. I suspect the most obvious would be to add a condition in `_assign_lhs_variable`:
```python
if 'device' in self.scope.decorators or 'kernel' in self.scope.decorators and get(d_var, 'memory_location', 'host') == 'host':
                        errors.report("Host arrays cannot be allocated on the Device",
                            symbol = expr,
                            severity = 'fatal')
```

That way you also handle other array creation functions and are reasonably future-proof

tests/internal/scripts/ccuda/cuda_to_host_device.py
Pinkyboi: Done in [cb5c4cc](https://github.com/pyccel/pyccel/pull/1236/commits/cb5c4cc6c22caa5184122375a4936c0ac9763418).

EmilyBourne (changes requested):
Please merge in the changes from the branch above and fix the failing tests. You have requested my review but unless you have a specific question I'm not sure it's worth me going through the code when so much is failing

pyccel/codegen/printing/ccudacode.py
```diff
@@ -521,19 +521,25 @@ def copy_CudaArray_Data(self, expr):
         declare_dtype = self.find_in_dtype_registry(self._print(rhs.dtype), rhs.precision)
         dtype = self.find_in_ndarray_type_registry(self._print(rhs.dtype), rhs.precision)
         arg = rhs.arg if isinstance(rhs, (CudaArray, CupyArray)) else rhs
+
+        memcpy_kind_src = str(rhs.current_context).capitalize()
+        memcpy_kind_dest = 'Host' if rhs.memory_location == 'host' else 'Device'
```
EmilyBourne, line 526: Could this be:
```python
        memcpy_kind_src = str(rhs.current_context).capitalize()
        memcpy_kind_dest = rhs.memory_location.capitalize()
```
?
I'm not sure why you're using 2 different methods here.

EmilyBourne (approved):

pyccel/codegen/printing/ccudacode.py
Pinkyboi: Oh that's to not have a "Managed" `cudaMemcpy` destination, since the only valid value are "Host" and "Device", and it name imply the `managed` memory location can work with any of those two values.

EmilyBourne: Ok perfect :+1:

Pinkyboi:
@bauom when you have time please review this PR.

EmilyBourne:
@Pinkyboi Please could you migrate this PR to the `pyccel/pyccel-cuda` repo. You can transfer your branch to that repo using the following commands:
```
git remote add cuda git@github.com:pyccel/pyccel-cuda.git
git checkout GPU_ddocs
git push -u cuda GPU_ddocs
```
cc: @bauom

PR #1255, opened 2022-12-12
issue_1254 - Do not use exact comparison in tests involving floating …

ceciver:
This pull request updates the tests in the codebase to use approximate comparisons when testing floating point numbers. Exact comparisons with floating point numbers can be unreliable due to the inherent imprecision of representing decimal numbers using a finite number of binary digits.

By using approximate comparisons, we can specify a tolerance value that allows for some numerical error in the comparison, making the tests more robust and reliable.

This change replaces all uses of the `==` operator with floating point numbers with calls to the `numpy.isclose()` function, which provides support for approximate comparisons.


tests/epyccel/recognised_functions/test_numpy_types.py
```diff
@@ -112,22 +112,22 @@ def add_numpy_to_numpy_type(np_s_l, np_s_r):
...
     python_result = add_numpy_to_numpy_type(fl64, complex64)
-    assert pyccel_result == python_result
+    assert np.isclose(pyccel_result, python_result)
     assert isinstance(pyccel_result, type(python_result))
 
     pyccel_result = epyccel_func(complex128, fl64)
     python_result = add_numpy_to_numpy_type(complex128, fl64)
-    assert pyccel_result == python_result
+    assert np.isclose(pyccel_result, python_result)
     assert isinstance(pyccel_result, type(python_result))
```
yguclu, line 131: Be careful! You should not use `numpy.isclose`, but rather `numpy.allclose` (see [docs](https://numpy.org/doc/stable/reference/generated/numpy.allclose.html#numpy.allclose)).

Further you should provide relative and absolute tolerances (`RTOL`, `ATOL`) which can be defined at the beginning of the test file. Using a value of `1e-14` for both should be adequate. See [this file](https://github.com/pyccel/pyccel/blob/master/tests/epyccel/test_epyccel_mod.py) as an example.

yguclu (changes requested):

tests/epyccel/recognised_functions/test_numpy_types.py
yguclu: Sorry, when comparing scalar values `isclose` is equivalent to `allclose`, so they can both be used here.

ceciver: Thank you! I have updated my PR to include the relative and absolute tolerances (RTOL, ATOL).

jalalium (approved):

tests/epyccel/recognised_functions/test_numpy_types.py
```diff
@@ -12,6 +12,13 @@
 from pyccel.decorators import types, template
 from pyccel.epyccel import epyccel
 
+if sys.platform == 'win32':
+    RTOL = 1e-13
+    ATOL = 1e-14
+else:
+    RTOL = 2e-14
+    ATOL = 1e-15
```
yguclu, line 20: We do not need to treat Windows differently unless it is really needed. Here we are only adding two numbers, so hopefully we can use the same (smaller) tolerance for all operating systems.

yguclu (changes requested):

tests/epyccel/recognised_functions/test_numpy_types.py
ceciver: @rakati I understand your suggestion, but I would like to point out that the types used in this situation are numpy.float64 and numpy.complex128. Therefore, we only need to use one tolerance value for both types. It is not necessary to have separate tolerance values for 32 precision in this case.

rakati: All right, I suggest naming it `RTOL_64`. If you already found a suitable one for 32 bits too, you can add it for future use.

EmilyBourne:
I suspect this is not the only place in the tests with this problem. Was any other code examined to find similar issues?

ceciver:
> I suspect this is not the only place in the tests with this problem. Was any other code examined to find similar issues?

@EmilyBourne Yes, I will update the tests that use exact comparison for floats.

tests/epyccel/test_epyccel_IfTernaryOperator.py
```diff
@@ -1,9 +1,13 @@
...
 
 import pytest
-
+import numpy as np
 from pyccel.epyccel import epyccel
 from pyccel.decorators import types
+
+RTOL = 2e-14
+ATOL = 1e-15
+
```
yguclu, line 10: The tests in this module should not introduce any roundoff errors

tests/epyccel/test_epyccel_generators.py
```diff
@@ -5,6 +5,9 @@
 
 from pyccel.epyccel import epyccel
 
+RTOL = 2e-14
+ATOL = 1e-15
+
```
yguclu, line 10: I suspect that the tests in this module should not introduce any roundoff. We usually start from a random sequence of small integers, convert it to float, and then apply simple operators.

tests/epyccel/test_epyccel_python_annotations.py
```diff
@@ -73,7 +76,7 @@ def test_array_real_1d_scalar_add(language):
     f1(x1, a)
     f2(x2, a)
 
-    assert np.array_equal( x1, x2 )
+    assert np.allclose(x1, x2, rtol=RTOL, atol=ATOL)
```
yguclu, line 79: Adding small integers in floating-point representation should be exact, hence not affected by roundoff errors.

tests/epyccel/test_epyccel_python_annotations.py
```diff
@@ -88,7 +91,7 @@ def test_array_real_2d_F_scalar_add(language):
...
```
yguclu, line 94: See comment above.

tests/epyccel/test_epyccel_python_annotations.py
```diff
@@ -102,7 +105,7 @@ def test_array_real_2d_F_add(language):
...
```
yguclu, line 108: See comment above.

tests/epyccel/test_epyccel_python_annotations.py
```diff
@@ -130,7 +133,7 @@ def test_array_real_1d_complex_3d_expr(language):
...
```
yguclu, line 136: See comment above.

tests/epyccel/test_epyccel_types.py
```diff
@@ -6,6 +6,9 @@
 from modules import types
 from pyccel.epyccel import epyccel
 
+RTOL = 2e-14
+ATOL = 1e-15
+
```
yguclu, line 11: All functions in the [**types**](https://github.com/pyccel/pyccel/blob/master/tests/epyccel/modules/types.py) module are simple identities, hence it is appropriate to use exact comparison here.

tests/epyccel/test_generic_functions.py
```diff
@@ -6,68 +6,71 @@
 import modules.generic_functions_2 as mod2
 from pyccel.epyccel import epyccel
 
+RTOL = 2e-14
+ATOL = 1e-15
+
```
yguclu, line 11: This module does not use any random number. If all tests pass now, I don't see the need to change all exact comparisons to approximate ones.

CC: @EmilyBourne

yguclu (changes requested):

tests/epyccel/test_epyccel_complex_func.py
```diff
@@ -7,6 +7,9 @@
 import modules.complex_func as mod
 from pyccel.epyccel import epyccel
 
+ATOL = 1e-15
+RTOL = 2e-14
+
```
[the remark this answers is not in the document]
EmilyBourne: I'm not sure I agree. While machines implementing IEEE standards should give equivalent results for casts and simple operations, I don't think that we can guarantee that all machines that pyccel could encounter will implement this everywhere. Personally, I would use tolerances throughout.

If we really want to check exact equality then this can be done where there is casting, and maybe where there are very simple operations. But I suspect mixing equality and tolerance is a recipe for getting them mixed up sometimes. It's harder to standardise the handling with two different setups and creates more work if the test is modified slightly. It also prevents us from creating a universal `test_results` function

ceciver:
Hello @EmilyBourne I fixed the falling tests that are relevant to my PR, but still some tests on windows/macos that fails but not related to my PR.

EmilyBourne:
@ceciver This is normal. As you can see from any other PRs merging to master this is a problem due to #1272 which is fixed in #1277. We are waiting for @yguclu to get back from holiday to finish merging the fix

EmilyBourne (approved):

EmilyBourne:
@ceciver Please merge the master branch and fix the conflicts

tests/epyccel/test_epyccel_complex_func.py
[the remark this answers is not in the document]
yguclu: Personally I would prefer to switch from exact to approximate comparison only in the cases where this is necessary, but you are making some solid arguments here. If using approximate comparison for floats ultimately means more unit tests which are more maintainable, I must agree.

yguclu (approved):

PR #1256, opened 2022-12-13
Improve np.array implementation

### LINKED ISSUE #1241: Wrong use of transpose in np.array() in fortran
**Describe the bug**
Wrong behaviour when using steps in views with np.array() bug in fortran translation

**To Reproduce**
```python
import numpy as np
if __name__ == '__main__':
    a = np.array([[ 1,  2,  4,  8, 16],
       [ 1,  2,  4,  8, 16],
       [ 1,  2,  4,  8, 16],
       [ 1,  2,  4,  8, 16],
       [ 1,  2,  4,  8, 16]], order="F")
    tmp = a[::2]
    b = np.array(tmp, order="F")
    print(b)
```

```c
program prog_prog_test

  use test_0001

  implicit none

  integer(i64), allocatable, target :: a(:,:)
  integer(i64), pointer :: tmp(:,:)
  integer(i64), allocatable :: b(:,:)
  integer(i64) :: i
  integer(i64) :: i_0001
  integer(i64) :: i_0002

  allocate(a(0:4_i64, 0:4_i64))
  a = transpose(reshape([[1_i64, 2_i64, 4_i64, 8_i64, 16_i64], [1_i64, &
        2_i64, 4_i64, 8_i64, 16_i64], [1_i64, 2_i64, 4_i64, 8_i64, &
        16_i64], [1_i64, 2_i64, 4_i64, 8_i64, 16_i64], [1_i64, 2_i64, &
        4_i64, 8_i64, 16_i64]], [5_i64, 5_i64]))
  tmp(0:, 0:) => a(::2_i64, :)
  allocate(b(0:size(tmp, 1_i64, i64) - 1_i64, 0:size(tmp, 2_i64, i64) - &
        1_i64))
  b = transpose(tmp)
  do i = 0_i64, size(b, 1_i64, i64) - 1_i64 - 1_i64, 1_i64
    do i_0001 = 0_i64, size(b, 2_i64, i64) - 1_i64 - 1_i64, 1_i64
      write(stdout, '(I0 A)', advance="no") b(i, i_0001) , ' '
    end do
    write(stdout, '(I0 A)', advance="no") b(i, size(b, 2_i64, i64) - &
          1_i64) , ']'
  end do
  do i_0002 = 0_i64, size(b, 2_i64, i64) - 1_i64 - 1_i64, 1_i64
    write(stdout, '(I0 A)', advance="no") b(size(b, 1_i64, i64) - 1_i64, &
          i_0002) , ' '
  end do
  write(stdout, '(I0 A)', advance="no") b(size(b, 1_i64, i64) - 1_i64, &
        size(b, 2_i64, i64) - 1_i64) , ']'
  write(stdout, '(A)', advance="no") ']'
  if (allocated(a)) then
    deallocate(a)
  end if
  if (allocated(b)) then
    deallocate(b)
  end if

end program prog_prog_test
```

**Expected behavior**
Should print
```
[[1 2 4 8 16]
[1 2 4 8 16]
[1 2 4 8 16]]
```
Instead prints:
```
[[16 129 16]
[16 1 1]
[16 2 2]
[16 4 4]
[16 8 8]]
```

**Language**
fortran

 **Additional info**
Doesn't happen in 1D arrays

jalalium:
- Ensure `np.array` can be called with one variable argument (previously this caused an infinite loop over the elements of the array).
- Simplify `_print_NumpyArray` to reduce duplication and fix #1241 and fix duplicate #1257.
- Improve `np.array` implementation to correct the default order and add the `ndmin` parameter.
- Add tests for new code.
- Clean up `test_arrays.py` so failures reference the relevant issue, tuples use `==` instead of `np.array_equal`, and a more comprehensive check is carried out on returned arrays including checking the datatype and ordering.

### DIFF
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -502,12 +502,9 @@ class NumpyNewArray(PyccelInternalFunction):
         return order
 
 #==============================================================================
-# TODO [YG, 18.02.2020]: accept Numpy array argument
-# TODO [YG, 18.02.2020]: use order='K' as default, like in numpy.array
-# TODO [YG, 22.05.2020]: move dtype & prec processing to __init__
 class NumpyArray(NumpyNewArray):
     """
-    Represents a call to  numpy.array for code generation.
+    Represents a call to `numpy.array` for code generation.
 
     A class representing a call to the NumPy `array` function.
 
@@ -521,12 +518,16 @@ class NumpyArray(NumpyNewArray):
 
     order : str
         The ordering of the array (C/Fortran).
+
+    ndmin : LiteralInteger, int, optional
+        The minimum number of dimensions that the resulting array should
+        have.
     """
     __slots__ = ('_arg','_dtype','_precision','_shape','_rank','_order')
     _attribute_nodes = ('_arg',)
     name = 'array'
 
-    def __init__(self, arg, dtype=None, order='C'):
+    def __init__(self, arg, dtype=None, order='K', ndmin=None):
 
         if not isinstance(arg, (PythonTuple, PythonList, Variable)):
             raise TypeError('Unknown type of  %s.' % type(arg))
@@ -538,6 +539,17 @@ class NumpyArray(NumpyNewArray):
         if not (is_homogeneous_tuple or is_array):
             raise TypeError('we only accept homogeneous arguments')
 
+        if not isinstance(order, (LiteralString, str)):
+            raise TypeError("The order must be specified explicitly with a string.")
+        elif isinstance(order, LiteralString):
+            order = order.python_value
+
+        if ndmin is not None:
+            if not isinstance(ndmin, (LiteralInteger, int)):
+                raise TypeError("The minimum number of dimensions must be specified explicitly with an integer.")
+            elif isinstance(ndmin, LiteralInteger):
+                ndmin = ndmin.python_value
+
         init_dtype = dtype
 
         # Verify dtype and get precision
@@ -546,11 +558,13 @@ class NumpyArray(NumpyNewArray):
             prec = get_final_precision(arg)
         else:
             dtype, prec = process_dtype(dtype)
-        # ... Determine ordering
-        order = str(order).strip("\'")
 
         shape = process_shape(False, arg.shape)
         rank  = len(shape)
+
+        if ndmin and ndmin>rank:
+            shape += (LiteralInteger(1),)*(ndmin-rank)
+            rank = ndmin
 
         if rank < 2:
             order = None
@@ -561,9 +575,8 @@ class NumpyArray(NumpyNewArray):
             if order not in ('K', 'A', 'C', 'F'):
                 raise ValueError(f"Cannot recognize '{order}' order")
 
-            # TODO [YG, 18.02.2020]: set correct order based on input array
             if order in ('K', 'A'):
-                order = 'C'
+                order = arg.order
             # ...
 
         self._arg   = arg
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -563,7 +563,7 @@ class NumpyArray(NumpyNewArray):
         rank  = len(shape)
 
         if ndmin and ndmin>rank:
-            shape += (LiteralInteger(1),)*(ndmin-rank)
+            shape = (LiteralInteger(1),)*(ndmin-rank) + shape
             rank = ndmin
 
         if rank < 2:
@@ -577,6 +577,8 @@ class NumpyArray(NumpyNewArray):
 
             if order in ('K', 'A'):
                 order = arg.order
+                if order is None:
+                    order = 'C'
             # ...
 
         self._arg   = arg
--- a/pyccel/ast/numpyext.py
+++ b/pyccel/ast/numpyext.py
@@ -576,9 +576,7 @@ class NumpyArray(NumpyNewArray):
                 raise ValueError(f"Cannot recognize '{order}' order")
 
             if order in ('K', 'A'):
-                order = arg.order
-                if order is None:
-                    order = 'C'
+                order = arg.order or 'C'
             # ...
 
         self._arg   = arg
--- a/pyccel/codegen/printing/fcode.py
+++ b/pyccel/codegen/printing/fcode.py
@@ -1078,47 +1078,38 @@ class FCodePrinter(CodePrinter):
         expr_args = (expr.arg,) if isinstance(expr.arg, Variable) else expr.arg
         # If Numpy array is stored with column-major ordering, transpose values
         # use reshape with order for rank > 2
-        if expr.order == 'F':
-            if expr.rank == 2:
-                rhs_code = self._print(expr.arg)
-                if expr.arg.order != expr.order:
-                    rhs_code = 'transpose({})'.format(rhs_code)
-            elif expr.rank > 2:
-                args     = [self._print(a) for a in expr_args]
-                new_args = []
-                for ac, a in zip(args, expr_args):
-                    if a.order == 'C':
-                        shape    = ', '.join(self._print(i) for i in a.shape)
-                        order    = ', '.join(self._print(LiteralInteger(i)) for i in range(a.rank, 0, -1))
-                        ac       = 'reshape({}, [{}], order=[{}])'.format(ac, shape, order)
-                    new_args.append(ac)
-
-                args     = new_args
-                rhs_code = '[' + ' ,'.join(args) + ']'
-                shape    = ', '.join(self._print(i) for i in expr.shape)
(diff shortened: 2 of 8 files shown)


EmilyBourne:
@jalalium Thanks for your pull request. Please have a quick look at the [review process in the developer docs](https://github.com/pyccel/pyccel/blob/master/developer_docs/review_process.md) to see how we organise PRs in pyccel and add the appropriate labels. Thanks!

yguclu:
@jalalium Please add a unit test which reproduces issue #1241. This should fail before your changes.

jalalium:
> @jalalium Please add a unit test which reproduces issue #1241. This should fail before your changes.

While trying to add tests for both `epyccel` and `pyccel`, I noticed that their behavior are inconsistent. `Epyccel` performs an extra transposition of the `fortran` code. Basically, the output of the `epyccel` function and the output one would get from generating `fortran` code are inconsistent.
**For the following function:**
```    
def f():
    import numpy as np
    a = np.array([[1,  2,  4,  8, 16],
             [1,  2,  4,  8, 16],
             [1,  2,  4,  8, 16],
             [1,  2,  4,  8, 16],
             [1,  2,  4,  8, 16]], order="F")
    b = np.array(a, order="F")
    return b
```
**epyccel outputs the following**
```
f2 = epyccel(f, language='fortran')
print(f2())
```
```
[[ 1  1  1  1  1]
 [ 2  2  2  2  2]
 [ 4  4  4  4  4]
 [ 8  8  8  8  8]
 [16 16 16 16 16]]
 ```
 While the code generated in `Fortran` outputs the expected result.

**Comment by EmilyBourne** [comment]
I think there are 2 bugs here. I am opening issues

**Comment by EmilyBourne** [comment]
Unfortunately after looking into this I don't think you can add tests until #1128 and #1260 are merged. I am therefore marking this PR as blocked. Hopefully @ohachim can get back to PR #1128 soon so we can unblock these issues.

**Comment by EmilyBourne** [comment]
Blocked by #1258

File: pyccel/codegen/printing/fcode.py [hunk 1]
```diff
@@ -1060,7 +1060,8 @@ def _print_NumpyArray(self, expr):
         if expr.order == 'F':
             if expr.rank == 2:
                 rhs_code = self._print(expr.arg)
-                rhs_code = 'transpose({})'.format(rhs_code)
+                if expr.arg.order != expr.order:
+                    rhs_code = 'transpose({})'.format(rhs_code)
```
EmilyBourne, line 1064: I think there are other places where this function behaves badly. Eg:
```
a = np.ones((3,4), order='C')
b = np.ones((3,4), order='F')
a[:] = b
```

EmilyBourne:
> It appears that issue #1496 has been fixed by PR #1477. If that is the case, I suppose that the conflicts in `pyccel/codegen/printing/cwrappercode.py` can be resolved by simply discarding any changes in this branch. @EmilyBourne Do you agree? 

That is correct. I needed a workaround to get the tests to pass but these changes can be discarded

pyccel/ast/numpyext.py
```diff
@@ -561,9 +575,10 @@ def __init__(self, arg, dtype=None, order='C'):
             if order not in ('K', 'A', 'C', 'F'):
                 raise ValueError(f"Cannot recognize '{order}' order")
 
-            # TODO [YG, 18.02.2020]: set correct order based on input array
             if order in ('K', 'A'):
-                order = 'C'
+                order = arg.order
+                if order is None:
+                    order = 'C'
```
yguclu, line 581: A more Pythonic one-liner:
```suggestion
                order = arg.order or 'C'
```

tests/epyccel/test_arrays.py
```diff
@@ -3808,60 +3818,144 @@ def test_array_view_steps_F_1D_1(language):
...
+# TEST: Array with ndmin argument
+#==============================================================================
+
+@pytest.mark.parametrize( 'language', (
+        pytest.param("fortran", marks = pytest.mark.fortran),
+        pytest.param("c", marks = pytest.mark.c),
+        pytest.param("python", marks = [
+            pytest.mark.skip(reason=("Template results in wrong ordered arrays")),
+            pytest.mark.python]
+        )
```
yguclu, line 3865: Can you please explain what goes wrong in the Python printer? A corresponding open issue should be referenced here.

tests/epyccel/test_arrays.py
```diff
@@ -3808,60 +3818,144 @@ def test_array_view_steps_F_1D_1(language):
...
+    check_array_equal(f1(d), f2(d))
+    check_array_equal(f1(e), f2(e))
+
+@pytest.mark.parametrize( 'language', (
+        pytest.param("fortran", marks = pytest.mark.fortran),
+        pytest.param("c", marks = pytest.mark.c),
+        pytest.param("python", marks = [
+            pytest.mark.skip(reason=("Template results in wrong ordered arrays")),
+            pytest.mark.python]
+        )
```
yguclu, line 3890: Same comment as above.

tests/epyccel/test_arrays.py
```diff
@@ -3808,60 +3818,144 @@ def test_array_view_steps_F_1D_1(language):
...
```
yguclu, line 3915: Same comment as above.

tests/epyccel/test_arrays.py
```diff
@@ -3808,60 +3818,144 @@ def test_array_view_steps_F_1D_1(language):
...
```
yguclu, line 3940: Same comment as above.

tests/epyccel/test_epyccel_return_arrays.py
```diff
@@ -778,3 +778,141 @@ def return_f_array(b : 'T'):
...
+        assert pyth_out.flags.c_contiguous == pycc_out.flags.c_contiguous
+        assert pyth_out.flags.f_contiguous == pycc_out.flags.f_contiguous
+
+@pytest.mark.parametrize( 'language', (
+        pytest.param("fortran", marks = pytest.mark.fortran),
+        pytest.param("c", marks = pytest.mark.c),
+        pytest.param("python", marks = [
+            pytest.mark.xfail(reason="Order not printed in Python. See #1260"),
+            pytest.mark.python
+        ])
```
yguclu, line 808: #1260 was merged last week.

yguclu (changes requested):
Looks good, but I have a couple of questions.

tests/epyccel/test_epyccel_return_arrays.py
```diff
@@ -778,3 +778,141 @@ def return_f_array(b : 'T'):
...
+    )
+)
```
EmilyBourne, line 810:
```suggestion
```

tests/epyccel/test_arrays.py
EmilyBourne: The problem is that the code is different for the different orders. This is because the order is printed explicitly. This could be avoided by saving the initial order as we now do for the dtype. But I'm not convinced it is worth it as this use case is not common.

There is no existing relevant issue. There was one but it was converted to discussion #1334. This could be referenced but I am not sure that it helps clarify the reason for the xfail. Additionally when a fix is chosen and implemented it should fix all Python xfail marks so referencing the discussion shouldn't make any difference for finding xfails to remove

tests/epyccel/test_epyccel_return_arrays.py
EmilyBourne: Removed

tests/epyccel/test_arrays.py
yguclu: Could you please give me an example of how the code differs based on the order of the input array?

EmilyBourne: Eg here:
```python
@template('T', ['int[:]', 'int[:,:]', 'int[:,:,:]', 'int[:,:](order=F)', 'int[:,:,:](order=F)']) 
 def array_ndmin_1(x : 'T'): 
     from numpy import array 
     y = array(x, ndmin=1) 
     return y
```
The argument types `int[:,:]`, and `int[:,:](order=F)` don't work at the same time as the code generated for the first type is something like:
```python
 def array_ndmin_1(x : 'int[:,:]'): 
     from numpy import array 
     y = array(x, dtype=x.dtype, order='C') 
     return y
```
(You may have to double check the actual generated code). NumPy allows properties like the order to be changed when copying data

yguclu: I see. The problem is that we compute the value of `order` explicitly, instead of keeping the original argument `order='K'`, which is indeed the default value for `numpy.array`.

EmilyBourne: Yes. This makes our life a lot easier when working out the order of any results for printing in C/Fortran, but in cases like this it causes differences in printing

yguclu (approved):

PR #1266, opened 2022-12-20
Fix deallocate printer

framdani:


pyccel/stdlib/cuda_ndarrays/cuda_ndarrays.h
```diff
@@ -24,4 +24,13 @@ void _cuda_array_fill_double(double c, t_ndarray arr);
 t_ndarray   cuda_array_create(int32_t nd, int64_t *shape, enum e_types type, bool is_view, enum e_memory_locations location);
 int32_t         cuda_free_array(t_ndarray dump);
 int32_t         cuda_free_pointer(t_ndarray dump);
+
+__host__ __device__
+int32_t cuda_free_host(t_ndarray arr);
```
bauom, line 29: I don't see a need to decorate `cuda_free_host` with `__host__` as there is not way for the `device` to call it on a `host` array or memory.

bauom (changes requested):

pyccel/stdlib/cuda_ndarrays/cuda_ndarrays.h
framdani: Fixed b9cff43a4cd4ed5cb85c4a84cc28c9cf74d24c05

bauom:
if you could add some tests that use cuda_free_pointer as i don't think we have any.
everything else seems good to me.

tests/internal/scripts/ccuda/deallocate.py
```diff
@@ -0,0 +1,19 @@
...
+    """
+    Takes an array 'arr' as input and raise each element to the power 2.
+    """
+    i = cuda.grid(0)
+    arr[i] = math.pow(arr[i], 2)
+
+if __name__ == '__main__':
+    a = cuda.array([1,2,3,4], memory_location='device')
+    b = cuda.array([3,4,5,6], memory_location='host')
+    c = b
```
bauom, line 16: i think it would be better to have the name of the file indicate which free are we targeting with the tests.
then we can know if `cuda_free` or `cuda_free_pinter` is the one that have the error.

bauom (changes requested):

EmilyBourne:
Codacy tests are failing

pyccel/codegen/printing/ccudacode.py
```diff
@@ -344,15 +344,15 @@ def _print_Allocate(self, expr):
         return '{}\n{}\n{}\n'.format(free_code, shape_Assign, alloc_code)
 
     def _print_Deallocate(self, expr):
```
AbdelkhalekBAROUDI, line 346: Missing docstring in this function.

EmilyBourne: We don't usually insist on docstrings in `_visit_X` and `_print_X` functions as they would all contain the same thing with only the name of the type changed

AbdelkhalekBAROUDI: Okay, I got it.

pyccel/stdlib/cuda_ndarrays/cuda_ndarrays.h
bauom: I noticed that the commit doesn't seem to have the changes I mentioned.

framdani: Yeah it was a mistake. Fixed 243ea719deb4621de9031638ff1e293a03faa280

bauom:
@EmilyBourne is this `CoverageChecker` error intended and we should not create a variable like that.

bauom:
@framdani i think in `_print_Deallocate` you can just return the Values because I don't see the need for `dealloc_code` Variable.

EmilyBourne:
> @EmilyBourne is this `CoverageChecker` error intended and we should not create a variable like that.

Huh, I thought I'd pulled the fix for that bug. It's not intended. There's a file where I accidentally save the line instead of the line index. I'll fix it once I find somewhere I can get my computer out

EmilyBourne:
> > @EmilyBourne is this `CoverageChecker` error intended and we should not create a variable like that.
> 
> Huh, I thought I'd pulled the fix for that bug. It's not intended. There's a file where I accidentally save the line instead of the line index. I'll fix it once I find somewhere I can get my computer out 

Ah sorry, wrong PR. The fix is in #1279. You can either pull the changes from there or wait until that PR is merged and update

pyccel/codegen/printing/ccudacode.py
```diff
@@ -344,15 +344,15 @@ def _print_Allocate(self, expr):
...
+        dealloc_code = ''
         if expr.variable.is_alias:
-            return '{}free_pointer({});\n'.format(cuda, self._print(expr.variable))
-        return '{}free_array({});\n'.format(cuda, self._print(expr.variable))
+            dealloc_code =  f"cuda_free_pointer({expr.variable});"
+        else:
+            if expr.variable.memory_location == 'host':
+                dealloc_code = f"cuda_free_host({expr.variable});"
+            else:
+                dealloc_code = f"cuda_free({expr.variable});"
```
EmilyBourne, line 354: You should never print an object using `str(expr)` (this is what you are doing in an f-stream), you need to call the `self._print` function:
```
        if expr.variable.is_alias:
            dealloc_code =  f"cuda_free_pointer({var_code});"
        elif expr.variable.memory_location == 'host':
            dealloc_code = f"cuda_free_host({var_code});"
        else:
            dealloc_code = f"cuda_free({var_code});"
```

pyccel/codegen/printing/ccudacode.py
```diff
@@ -344,15 +344,15 @@ def _print_Allocate(self, expr):
...
+        return '%s\n' % (dealloc_code)
```
EmilyBourne, line 355: Why not put the `\n` in the original definition of `dealloc_code`?

EmilyBourne (changes requested):

pyccel/codegen/printing/ccudacode.py
framdani: Fixed a6341bb0fbb29da2b6290d8421fab40577c423f2

pyccel/codegen/printing/ccudacode.py
framdani: Updated a6341bb0fbb29da2b6290d8421fab40577c423f2

bauom (commented):
everything seems good to me now.
will merge in the correct order to fix the failing tests.

EmilyBourne (changes requested):
Please fix the codacy errors

EmilyBourne:
The f-strings are now reported on the codacy page: https://app.codacy.com/gh/pyccel/pyccel/pullRequest?prid=10805918

tests/cuda_test/test_print_deallocate.py
```diff
@@ -0,0 +1,40 @@
...
+    output = printer._print_Deallocate(expr)
+    expected_output = "cuda_free(y);\n"
+    assert output == expected_output
+    #Test case 4: variable is an alias
+    var = Variable('int', 'y', memory_handling='alias')
+    expr = Deallocate(var)
+    printer = CcudaCodePrinter('')
+    output = printer._print_Deallocate(expr)
+    expected_output = "cuda_free_pointer((*y));\n"
+    assert output == expected_output
```
EmilyBourne, line 40: I assume this test was added to avoid the Coverage Checker failure? I don't think this is a good idea. It tests the code which is good. But not having an error indicating that this code isn't tested hides the fact that Pyccel doesn't  seem to be generating a `Deallocate` node when the code requires it. I would prefer that the memory leak was fixed rather than just ensuring the printer code is tested

tests/cuda_test/test_print_deallocate.py
```diff
@@ -0,0 +1,40 @@
+# pylint: disable=missing-function-docstring, missing-module-docstring/
+# pylint: disable=protected-access
```
EmilyBourne, line 2: What is this pylint error? Why are you disabling it?

tests/cuda_test/test_print_deallocate.py
EmilyBourne: That said, this definitely should fix the coverage checker and it doesn't seem to. I'm going to play around with your branch briefly and see if I can find the problem

framdani: Yes, the tests were added to address the Coverage Checker failure. Regarding the memory leaks tests, you are correct, and I'll work on fixing it.

tests/cuda_test/test_print_deallocate.py
framdani: The pylint error `# pylint: disable=protected-access` is related to accessing a protected member. I am disabling it to test the `_print_Deallocate` method of the printer class which is a protected method.

tests/cuda_test/test_print_deallocate.py
EmilyBourne: By "fix the memory leak" I meant that if the deallocate node isn't being printed then there is probably a memory leak there. But don't worry about it. It looks like the coverage doesn't work properly. I'm working on fixing it

tests/cuda_test/test_print_deallocate.py
EmilyBourne: This shouldn't be necessary. You should be able to do `printer.doprint(expr)` for the same result

framdani:
> I think the problem is that the tests are being run from the wrong folder (`.` see [here](https://github.com/pyccel/pyccel/blob/cuda_main_temp/.github/actions/pytest_run_cuda/action.yml#L16)) instead of (`tests` see [here](https://github.com/pyccel/pyccel/blob/master/.github/actions/pytest_run/action.yml#L30)). This change was made by @bauom here : [452d2e3](https://github.com/pyccel/pyccel/commit/452d2e39333a78e150e491ea14ca063f84fc70b1) .
> 
> As a result you are using the local `pyccel/` folder instead of the installed one. But the coverage checks against the installed folder so it doesn't see that you have used the code. @framdani this is not your fault. I will open a PR to fix the problem which you can merge into your branch.

Thank you for pointing this out. I appreciate your help in identifying the problem.

EmilyBourne:
> > I think the problem is that the tests are being run from the wrong folder (`.` see [here](https://github.com/pyccel/pyccel/blob/cuda_main_temp/.github/actions/pytest_run_cuda/action.yml#L16)) instead of (`tests` see [here](https://github.com/pyccel/pyccel/blob/master/.github/actions/pytest_run/action.yml#L30)). This change was made by @bauom here : [452d2e3](https://github.com/pyccel/pyccel/commit/452d2e39333a78e150e491ea14ca063f84fc70b1) .
> > As a result you are using the local `pyccel/` folder instead of the installed one. But the coverage checks against the installed folder so it doesn't see that you have used the code. @framdani this is not your fault. I will open a PR to fix the problem which you can merge into your branch.
> 
> Thank you for pointing this out. I appreciate your help in identifying the problem.

There is a second problem. I think I have just found the solution. Sorry for borrowing your branch for so long for debugging

tests/cuda_test/test_print_deallocate.py
framdani: Oh I see. I thought you were talking about adding tests to check the memory leaks ...

tests/cuda_test/test_print_deallocate.py
framdani: Updated here 7f36c249828e9d131eb20db0e35963811350519c

EmilyBourne: Thanks :) but is this file still needed?
If the code is tested properly in the other tests then I think that is enough. Those tests are more complete anyway as they also check that the code compiles

tests/pyccel/scripts/test.json
```diff
@@ -0,0 +1,64 @@
...
+        "shared_suffix": ".cpython-310-darwin.so",
+        "libs": [
+            "python3.10"
+        ],
+        "libdirs": [
+            "/opt/homebrew/opt/python@3.10/Frameworks/Python.framework/Versions/3.10/lib/python3.10/config-3.10-darwin",
+            "/opt/homebrew/opt/python@3.10/Frameworks/Python.framework/Versions/3.10/lib"
+        ]
+    }
+}
```
EmilyBourne, line 64: Did you mean to commit this file?

tests/pyccel/scripts/test2.json
```diff
@@ -0,0 +1,64 @@
...
+        "shared_suffix": ".cpython-310-darwin.so",
+        "libs": [
+            "python3.10"
+        ],
+        "libdirs": [
+            "/opt/homebrew/opt/python@3.10/Frameworks/Python.framework/Versions/3.10/lib/python3.10/config-3.10-darwin",
+            "/opt/homebrew/opt/python@3.10/Frameworks/Python.framework/Versions/3.10/lib"
+        ]
+    }
+}
```
EmilyBourne, line 64: Did you mean to commit this file?

EmilyBourne (changes requested):

EmilyBourne:
Please could you merge `cuda_main_temp` into this branch. The diff is quite misleading at the moment

EmilyBourne:
Fixes https://github.com/pyccel/pyccel-cuda/issues/8 ?

tests/pyccel/scripts/test.json
framdani: This file was generated automatically and committed by mistake. It has been removed 9f712d936cc0a7893ae65ef321000c868b018749

framdani:
Yes this PR fixes https://github.com/pyccel/pyccel-cuda/issues/8

framdani:
> Fixes [pyccel-cuda#8](https://github.com/pyccel/pyccel-cuda/issues/8) ?

Yes this PR fixes [pyccel-cuda#8](https://github.com/pyccel/pyccel-cuda/issues/8)

framdani:
> Please could you merge `cuda_main_temp` into this branch. The diff is quite misleading at the moment

Fixed here 2a03c9c9c778ec1af554e83c2485bb0b8440bbc1

EmilyBourne (approved):

pyccel/codegen/printing/ccudacode.py
```diff
@@ -344,15 +344,14 @@ def _print_Allocate(self, expr):
...
+        var_code = self._print(expr.variable)
         if expr.variable.is_alias:
-            return '{}free_pointer({});\n'.format(cuda, self._print(expr.variable))
-        return '{}free_array({});\n'.format(cuda, self._print(expr.variable))
+            return f"cuda_free_pointer({var_code});\n"
+        else:
+            if expr.variable.memory_location == 'host':
+                return f"cuda_free_host({var_code});\n"
+            else:
+                return f"cuda_free({var_code});\n"
```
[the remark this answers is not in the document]
bauom: `cuda_free` calls `cudaFree` which can free `device` or `managed` memory.

Issue #1271, opened 2022-12-21
changing cuda.devicesynchronize to cuda.synchronize


bauom:
For now to call the `cudaDeviceSyncronize()` we use `cuda.cudaDeviceSyncronize()` , we could use `cuda.syncronize()` like used in `torch` or `numba`.

PR #1274, opened 2022-12-21
Cuda synchronize

Pinkyboi:
Fixes: #1271.
Renames `cuda.deviceSynchronize` to `cuda.synchronize`.
Fixes a known bug that prevented the creation of a `kernel` or `device` function.

### DIFF
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -812,7 +812,7 @@ class SemanticParser(BasicParser):
             func = func.cls_name
             if func in (CudaThreadIdx, CudaBlockDim, CudaBlockIdx, CudaGridDim):
                 if 'kernel' not in self.scope.decorators\
-                    or 'device' not in self.scope.decorators:
+                    and 'device' not in self.scope.decorators:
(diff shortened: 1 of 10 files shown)


pyccel/parser/semantic.py
```diff
@@ -812,7 +812,7 @@ def _handle_function(self, expr, func, args, **settings):
             func = func.cls_name
             if func in (CudaThreadIdx, CudaBlockDim, CudaBlockIdx, CudaGridDim):
                 if 'kernel' not in self.scope.decorators\
-                    or 'device' not in self.scope.decorators:
+                    and 'device' not in self.scope.decorators:
```
bauom, line 815: can you add this to the description so we can keep track of it.

bauom (changes requested):
except that change from using `or` to `and` that needs to be mentioned in the description
everything seems good to me

pyccel/parser/semantic.py
Pinkyboi: Fixed.

bauom (approved):

Issue #1272, opened 2022-12-21
Tests don't run with numpy 1.24.0


EmilyBourne:
### Describe the bug
Numpy has been warning for a while that it will deprecate `np.bool`, `np.int`, `np.float`, etc. It seems that as of 1.24.0 this is now deprecated. As a result some of the tests now fail.

### To Reproduce
E.g. : https://github.com/pyccel/pyccel/blob/master/tests/epyccel/recognised_functions/test_numpy_types.py#L145

### Expected behavior
We should remove deprecated behaviour or prevent these tests from running if a more recent numpy version is used

PR #1276, opened 2022-12-24
Add documentation for treatment of `order` in NumPy arrays

ohachim:
Create order documentation.
This will help us decide how `order` should be implemented in `C` code.


ohachim:
@bauom, there are some import errors in the Windows/MacOs checks, it's probably not from my side since this PR is based on `master` and only adds a doc file.
Not sure how to handle this

EmilyBourne:
> @bauom, there are some import errors in the Windows/MacOs checks, it's probably not from my side since this PR is based on `master` and only adds a doc file.
> Not sure how to handle this

See #1272

developer_docs/order_docs.md
```diff
@@ -0,0 +1,427 @@
...
+`arr.strides` is the variable that helps us navigate the array (indexing/printing), it tells us how many bytes we have to skip in memory to move to the next position along a certain axis (dimension). For example for `memory_layout_of_a = [1 4 7 2 5 8]` and `strides_of_a = (8, 24)`, we have to skip 8 bytes (1 element for `int64`) to move to the next row, but 24 bytes (3 elements for `int64`) to get to the same position in the next column of `a`.  
+`a[2][1]` would give us `'8'`, using the `strides`: `2 * 8 + 1 * 24 = 40`, which means that in the flattened array, we would have to skip `40` bytes to get the value of `a[2][1]`, each element is 8 bytes, so we would have to skip `40 / 8 = 5` elements to get to `'8'`
+
+## Pyccel's C code
+
+In `Pyccel`'s C code, we try to clone `Numpy`'s indexing/printing and memory layout conventions.
+
+### Ordering in C code
+
+The arrays in `C` code, are flattened into a one dimensional string, `strides` and `shape` are used to navigate the array (unlike `Numpy`. `Pyccel`'s strides use 'number of elements' instead of 'number of bytes' as a unit)
```
EmilyBourne, line 80: A string? I don't think so. What object are you talking about here?

developer_docs/order_docs.md
```diff
@@ -0,0 +1,427 @@
...
+
+  `order_c` creation  
+   &nbsp;&nbsp;&nbsp;&nbsp;1. allocate/create `order_c ndarray`  
+   &nbsp;&nbsp;&nbsp;&nbsp;2. copy values to `ndarray`  
+
+  `order_f` creation  
+   &nbsp;&nbsp;&nbsp;&nbsp;1. allocate/create temporary `order_c ndarray`  
+   &nbsp;&nbsp;&nbsp;&nbsp;2. copy values to temporary `ndarray`  
+   &nbsp;&nbsp;&nbsp;&nbsp;3. allocate/create `order_f ndarray`  
+   &nbsp;&nbsp;&nbsp;&nbsp;4. copy temporary `ndarray` elements to final `order_f ndarray` using `strides` and `shape`, this will create a column-major version of the temporary `order_c ndarray`
```
EmilyBourne: I think this step is only needed for the `np.array` function. As `np.array` is not the only way to populate an array (e.g. `np.ones`), I think this statement is a bit misleading

developer_docs/order_docs.md
```diff
@@ -0,0 +1,427 @@
...
+    free_array(a);
+    free_array(b);
+    free_array(c);
+    return 0;
+}
+```  
+
+### `order_f` array creation example
+
+If the data is one dimensional, all we would need is one copy operation, same as an `order_c ndarray`.
```
EmilyBourne, line 205: 1D objects always have `order=None` in pyccel (in numpy `c_order = True` and `f_order = True`)

EmilyBourne:
I am working on fixing #1272. In the meantime I have set the maximum numpy version to avoid the problem.

I made some changes to the Fortran docs and left some comments for C. I wonder if we should use `row` and `col` for the indexes instead of `i` and `j`?

developer_docs/order_docs.md
ohachim: By `string` I meant a `series of something` since that's one of its definitions, but I can see how that can be confusing in programming docs!
We could say `array` instead

EmilyBourne: 1d array? Flattened array?

bauom:
> I am working on fixing #1272. In the meantime I have set the maximum numpy version to avoid the problem.
> 
> I made some changes to the Fortran docs and left some comments for C. I wonder if we should use `row` and `col` for the indexes instead of `i` and `j`?

I think it make sense because we only have examples for 2D arrays but it will be more confusing in bigger dimensions.

developer_docs/order_docs.md
ohachim: Oh, I only had `np.array` in mind when writing this!
This means I might have to make more changes.

developer_docs/order_docs.md
ohachim: Yeh, we change it from `are flattened into a one dimensional string` to ` are flattened into a one dimensional arry`

developer_docs/order_docs.md
ohachim: Okey, I'll mention this and remove those lines.
For C code, we pass `order_c` to `array_create` in this situation, is that okey?
We could change it for organizational purposes, am not sure if that is very useful

ohachim:
@EmilyBourne, I am a bit confused in this part:

ohachim:
Note the changes to the shape and the indexing, which make this code closer to the following intermediate representation:

```python
import numpy as np

if __name__ == '__main__':
    A_mem = np.array([1,2,3,4,5,6])
    A = np.array(A_mem.reshape(2,3), order='C')
    B_mem = np.array([1,2,3,4,5,6])
    B = np.array(B_mem.reshape(3,2), order='C')
    print(A.shape[::-1])  # (3,2)
    print(B.shape)   # (3,2)
    # print(A.ravel('K')) # array([1,2,3,4,5,6])
    # print(B.ravel('K')) # array([1,2,3,4,5,6])

    # Index optimally for A in F layout:
    for i in range(3):
        for j in range(2):
            A[j,i] = ...

    # Index optimally for B in F layout:
    for j in range(2):
        for i in range(3):
            B[i,j] = ...
```
does `Index optimally for A in F layout` mean if A was `order="F"`, if that's the case I find it a bit confusing to have the arrays created with `order="C"`

developer_docs/order_docs.md
ohachim: https://github.com/pyccel/pyccel/pull/1276/commits/1d6a5d7eb2baa0efcca3d1818bb257cd4656323b
Should I mention that `array_create` default to `order_c` as a parameter if the array is one dimensional?

EmilyBourne: > For C code, we pass order_c to array_create in this situation, is that okey?

I think so. If we never check it then it doesn't matter. `order_c` just means "last dimension is contiguous" so that is true for a 1D array

> We could change it for organizational purposes, am not sure if that is very useful

I agree I don't think it matters

> Should I mention that array_create default to order_c as a parameter if the array is one dimensional?

I wouldn't. As it is a default parameter the developer doesn't need to know this.

EmilyBourne:
> > I am working on fixing #1272. In the meantime I have set the maximum numpy version to avoid the problem.
> > I made some changes to the Fortran docs and left some comments for C. I wonder if we should use `row` and `col` for the indexes instead of `i` and `j`?
> 
> I think it make sense because we only have examples for 2D arrays but it will be more confusing in bigger dimensions.

I agree for more dimensions we would have to use i, j, k, etc. But I think here it makes things clearer

EmilyBourne:
> @EmilyBourne, I am a bit confused in this part:

That's normal. Most of my description was back to front :facepalm: I have re-written my section.

developer_docs/order_docs.md
EmilyBourne: Maybe just make it clear what you are talking about ?

Example:
 To translate the following:
```python
a = np.array([[1, 2, 3], [4, 5, 6]], order=?)
```

**`order_c`**
1.    allocate/create `order_c ndarray`
2.    copy values to `ndarray`

**`order_f`**
1.    allocate/create temporary `order_c ndarray` (contains the list)
2.    copy values to temporary `ndarray` (initialises the list)
3.    allocate/create `order_f ndarray`
4.    copy temporary `ndarray` elements to final `order_f ndarray` using `strides` and `shape`, this will create a column-major version of the temporary `order_c ndarray`

ohachim: @EmilyBourne, @bauom, are there any other functions that change behaviour depending on the order?
If that's the case then we might need to talk about them in here?
For now, this talks generally about order, then goes into the specifics of array creation depending on the order (`np.array`).

EmilyBourne: Nothing comes to mind apart from copying data. But that is the same case as here. I'm not sure I would even classify `array` as a function changing its behaviour depending on the order. Both orders do:
- Allocate memory
- Copy data to memory

It's just that for the second step we are copying a C-ordered array to a F-ordered array so there needs to be some reorganisation

EmilyBourne (approved):
I am happy with the docs as they are, but I would like to get as many reviews as possible before merging to make sure this is all understandable for developers who are less familiar with the code base.

@ohachim if you agree, please add the `needs_inital_review` message and ask people in the discord to review this

ceciver (approved):

EmilyBourne:
@ohachim please fix typos and update the PR description

PR #1277, opened 2022-12-26
Deactivate tests using deprecated numpy syntax

### LINKED ISSUE #1272: Tests don't run with numpy 1.24.0

EmilyBourne:
- Deactivate tests that use deprecated numpy syntax, fixes #1272 
- Correct tests to use bound that don't generate `RuntimeWarning` messages about overflows
- Make type printing match python completely
- Stop deleting `.lock` file during folder overwrite (potential cause of mac parallel bug)

### DIFF
--- a/tests/epyccel/recognised_functions/test_numpy_types.py
+++ b/tests/epyccel/recognised_functions/test_numpy_types.py
@@ -306,10 +306,10 @@ else:
 def test_numpy_int_scalar(language, function_boundaries):
 
     integer8 = randint(min_int8, max_int8, dtype=np.int8)
-    integer16 = randint(min_int16, max_int16, dtype=np.int16)
-    integer = randint(min_int, max_int, dtype=int)
-    integer32 = randint(min_int32, max_int32, dtype=np.int32)
-    integer64 = randint(min_int64, max_int64, dtype=np.int64)
+    integer16 = randint(min_int8, max_int8, dtype=np.int16)
+    integer = randint(min_int8, max_int8, dtype=int)
+    integer32 = randint(min_int8, max_int8, dtype=np.int32)
+    integer64 = randint(min_int8, max_int8, dtype=np.int64)
 
     get_int = function_boundaries[0]
     # Modifying a global variable in a scop will change it to a local variable, so it needs to be initialized.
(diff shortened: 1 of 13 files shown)


tests/epyccel/recognised_functions/test_numpy_types.py
```diff
@@ -292,15 +295,21 @@ def get_int8(a):
...
 
     integer8 = randint(min_int8, max_int8, dtype=np.int8)
-    integer16 = randint(min_int16, max_int16, dtype=np.int16)
-    integer = randint(min_int, max_int, dtype=int)
-    integer32 = randint(min_int32, max_int32, dtype=np.int32)
-    integer64 = randint(min_int64, max_int64, dtype=np.int64)
+    integer16 = randint(min_int8, max_int8, dtype=np.int16)
+    integer = randint(min_int8, max_int8, dtype=int)
+    integer32 = randint(min_int8, max_int8, dtype=np.int32)
+    integer64 = randint(min_int8, max_int8, dtype=np.int64)
```
bauom, line 312: shouldn't these use min and max for the associated types?

tests/epyccel/recognised_functions/test_numpy_types.py
```diff
@@ -652,10 +666,10 @@ def test_numpy_float_scalar(language, get_float):
     integer32 = randint(min_int32, max_int32, dtype=np.int32)
     integer64 = randint(min_int64, max_int64, dtype=np.int64)
 
-    fl = uniform(min_float / 2, max_float / 2)
+    fl = uniform(min_float32 / 2, max_float32 / 2)
     fl32 = uniform(min_float32 / 2, max_float32 / 2)
     fl32 = np.float32(fl32)
-    fl64 = uniform(min_float64 / 2, max_float64 / 2)
+    fl64 = uniform(min_float32 / 2, max_float32 / 2)
```
bauom, line 672: why are we using `float32` instead of `float64` in these examples?

bauom (changes requested):

tests/epyccel/recognised_functions/test_numpy_types.py
EmilyBourne: No. The test converts integers of precision X to int8 objects. If the original int is larger than the target int then there will be overflow warnings. The latest version of numpy flags these overflow warnings with a `DeprecationError`, indicating that in a few versions time, this code will fail if we pass an int that is out of the bounds `[min_int8, max_int8]` to the function `int8`

tests/epyccel/recognised_functions/test_numpy_types.py
EmilyBourne: As above. We are converting to  `float32`, so the `float64` can't contain a float which is too big to be cast without overflow/underflow errors

tests/epyccel/recognised_functions/test_numpy_types.py
bauom: ah ok i see :+1:

bauom (approved):

yguclu (approved):

PR #1279, opened 2022-12-27
Fix kernel launch configuration errors

framdani:
This PR is about raising an error in case of : 
- Missing kernel launch configuration.
- Providing invalid parameters (blocks per grid, threads per block).
- Calling non-kernel functions with parameters.


pyccel/parser/semantic.py
```diff
@@ -903,14 +904,27 @@ def _handle_kernel(self, expr, func, args, **settings):
...
             # TODO : type check the NUMBER OF BLOCKS 'numBlocks' and threads per block 'tpblock'
             if not isinstance(expr.numBlocks, LiteralInteger):
-                errors.report("Invalid Block number parameter for Kernel call",
+                try:
+                    numBlocks = self.scope.variables[expr.numBlocks]
+                    if not isinstance(numBlocks.dtype, NativeInteger):
+                        errors.report("Invalid Block number parameter for Kernel call",
+                        symbol = expr,
+                        severity='error')
+                except KeyError:
```
EmilyBourne, line 913: Not IndexError?

framdani: No, because I'm looking for items in a dictionary of variables in this scope.
I think it would be better if I used method get instead of square braces.

EmilyBourne: Ok great. I was just confused by the error message in that case. I don't see any reason to use `get` instead of square braces. Square braces for dictionaries are completely standard in python. The only reason I thought it was a list/array rather than a dictionary was because you say "invalid block number"

Pinkyboi (approved):
Maybe you could add to `pyccel.errors.messages` the new error messages you introduced with this PR but other than that it looks all good to me.

pyccel/errors/messages.py
```diff
@@ -163,4 +163,6 @@
 KERNEL_STACK_ARRAY_ARG = "A variable allocated on the stack can't be passed to a Kernel function"
 NON_KERNEL_FUNCTION_CUDA_VAR = 'Cuda internal variables should only be used in Kernel or Device functions'
 UNVALID_KERNEL_CALL_BLOCK_NUM = 'Invalid Block number parameter for Kernel call'
-UNVALID_KERNEL_CALL_TP_BLOCK = 'Invalid Thread per Block parameter for Kernel call'
\ No newline at end of file
+UNVALID_KERNEL_CALL_TP_BLOCK = 'Invalid Thread per Block parameter for Kernel call'
```
bauom: i think you forgot to use these.

bauom (changes requested):

pyccel/errors/messages.py
framdani: Fixed 0d7b325d0a6c14dde370d98ad3934ab812361a6d

bauom (approved):

bauom (changes requested):
check the error message in this file `cuda_test/test_kernel_semantic.py`
as i think it was changed in this PR.
and if you could add some tests.

EmilyBourne (changes requested):
Tests are failing

tests/internal/scripts/ccuda/invalid_kernel_config.py
```diff
@@ -15,4 +15,4 @@ def opp(a, b):
     #invalid parameters
     opp[1.2, 7.0](a, b)
```
bauom, line 16: this test should be done in the same way as the tests in `tests/cuda_test/test_kernel_semantic.py`
you can just move these to `tests/cuda_test/test_kernel_semantic.py` maybe having all the error test in one file will be easier to maintain.

tests/internal/scripts/ccuda/missing_kernel_config.py
```diff
@@ -15,4 +15,4 @@ def opp(a, b):
     #call kernel launch without specifying config
     opp(a, b)
```
bauom, line 16: this test should be done in the same way as the tests in `tests/cuda_test/test_kernel_semantic.py`
you can just move these to `tests/cuda_test/test_kernel_semantic.py` maybe having all the error test in one file will be easier to maintain.

bauom (changes requested):

tests/internal/scripts/ccuda/missing_kernel_config.py
framdani: Fixed 4f61a29cc5b955aac66a97da402068ae7eeb4e55

tests/internal/scripts/ccuda/invalid_kernel_config.py
framdani: Fixed 7e71c89c30bf895c7d72a29fea81ae8e794b7fd6.

bauom:
we should just know why the tests are failing now and the PR should be good to be merged 👍

EmilyBourne:
> we should just know why the tests are failing now and the PR should be good to be merged +1

I think the failure is due to the removal of the "No new line at end of file". I will try and fix it for you

EmilyBourne:
@framdani Fixed. It is now failing for the reason it is supposed to fail.

https://github.com/pyccel/pyccel/actions/runs/3840365360/attempts/1#summary-10442401931

pyccel/parser/semantic.py
```diff
@@ -812,7 +812,9 @@ def _handle_function(self, expr, func, args, **settings):
         =======
         new_expr : FunctionCall or PyccelInternalFunction
         """
-
+        if isinstance(func, FunctionDef):
+            if 'kernel' in func.decorators:
```
EmilyBourne, line 816: why do you need 2 lines for this? Why not:
```suggestion
        if isinstance(func, FunctionDef) and 'kernel' in func.decorators:
```

EmilyBourne: Is there a way to add the kernel launch configuration? Why is the error always raised?

pyccel/parser/semantic.py
```diff
@@ -925,14 +927,27 @@ def _handle_kernel(self, expr, func, args, **settings):
                         severity='fatal')
             # TODO : type check the NUMBER OF BLOCKS 'numBlocks' and threads per block 'tpblock'
             if not isinstance(expr.numBlocks, LiteralInteger):
-                errors.report("Invalid Block number parameter for Kernel call",
+                try:
+                    numBlocks = self.scope.variables[expr.numBlocks]
```
EmilyBourne, line 931: This is not a good way to search for variables as it will fail if the variable is renamed due to collisions. You should use `self.scope.find(expr.numBlocks, category= 'variables')`.

pyccel/parser/semantic.py
```diff
@@ -925,14 +927,27 @@ def _handle_kernel(self, expr, func, args, **settings):
...
+                        symbol = expr,
+                        severity='error')
+                except KeyError:
+                    errors.report(INVALID_KERNEL_CALL_BP_GRID,
                         symbol = expr,
                         severity='error')
             if not isinstance(expr.tpblock, LiteralInteger):
-                errors.report("Invalid Thread per Block parameter for Kernel call",
+                try:
+                    tpblock = self.scope.variables[expr.tpblock]
```
EmilyBourne, line 942: See above, this is not the right way to search for variables

EmilyBourne (changes requested):

pyccel/parser/semantic.py
framdani: Did you mean that we should define default values for the parameters blocks per grid and threads per blocks if they are not specified, instead of raising an error? Could you please provide more clarification on this idea?

EmilyBourne: > Did you mean that we should define default values for the parameters blocks per grid and threads per blocks if they are not specified, instead of raising an error?

As I understand it, currently an error is raised whether parameters are specified or not. It seems strange to me to add code to handle a case that is not yet implemented

EmilyBourne: Ah ok! I understand. Sorry I thought I was looking for arguments to the `@kernel` decorator. So the blocks per grid and threads per block are provided by the index in the function call?

framdani: They should be provided as shown in the example below:
`launch_kernel[blocks_per_grid, threads_per_block](arguments...)`

pyccel/parser/semantic.py
EmilyBourne: Actually there is a function for this which also takes care of raising the error. See `get_variable`

tests/cuda_test/test_kernel_semantic.py
```diff
@@ -121,4 +121,103 @@ def kernel_call():
...
+@pytest.mark.parametrize( 'language', [
+        pytest.param("ccuda", marks = pytest.mark.ccuda)
+    ]
+)
+def test_invalid_kernel_config(language):
+    def invalid_kernel_config():
+        @kernel
+        def kernel_call():
+            pass
+        kernel_call[1.0, 2.0]()
```
EmilyBourne, line 135: Why are these indices floats?

tests/cuda_test/test_kernel_semantic.py
```diff
@@ -121,4 +121,103 @@ def kernel_call():
...
+        pytest.param("ccuda", marks = pytest.mark.ccuda)
+    ]
+)
+def test_invalid_thread_per_block(language):
+    def invalid_thread_per_block():
+        @kernel
+        def kernel_call():
+            pass
+        threads_per_block = 5.0
+        kernel_call[1, threads_per_block]()
```
EmilyBourne, line 211: Why is the number of threads a float?

tests/cuda_test/test_kernel_semantic.py
framdani: The indices should be integers. I intentionally provided a float values as an invalid types to test that the appropriate error is raised.

EmilyBourne: :facepalm: of course. Sorry

tests/cuda_test/test_kernel_semantic.py
framdani: The same here, I provided invalid type to raise the appropriate error.

pyccel/parser/semantic.py
framdani: Thank you for pointing that out. Fixed e5d7ba0ed13698404e6914ea4ea1b945f8d06c15

pyccel/parser/semantic.py
framdani: Fixed e5d7ba0ed13698404e6914ea4ea1b945f8d06c15

EmilyBourne (approved):

EmilyBourne:
Please merge the updated branch `cuda_main_temp` into this branch

EmilyBourne:
@bauom If you are happy with the changes, I recommend that we merge this branch as it fixes half of the problems in the CUDA tests

EmilyBourne:
To avoid triggering Codacy every time you use a kernel. I recommend using a different definition of the decorator:
```python
def kernel(f):
    return np.array([[f]])
```
This will allow the kernel to be indexed with `[0,0]` in pure python code.

bauom:
> To avoid triggering Codacy every time you use a kernel. I recommend using a different definition of the decorator:
> 
> ```python
> def kernel(f):
>     return np.array([[f]])
> ```
> 
> This will allow the kernel to be indexed with `[0,0]` in pure python code.

The problem with that is a kernel can't return.
it would be incorrect to define it as such.

EmilyBourne:
> > To avoid triggering Codacy every time you use a kernel. I recommend using a different definition of the decorator:
> > ```python
> > def kernel(f):
> >     return np.array([[f]])
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This will allow the kernel to be indexed with `[0,0]` in pure python code.
> 
> The problem with that is a kernel can't return. it would be incorrect to define it as such.

This is not the definition of the kernel. This is the definition of the `@kernel` decorator. Decorators take a function as an argument and return something, usually a function, here an array of functions

tests/cuda_test/test_kernel_semantic.py
```diff
@@ -121,4 +121,126 @@ def kernel_call():
...
+    # Check that we got exactly 2 Pyccel error
+    assert errors.has_errors()
+    assert errors.num_messages() == 2
+
+    # Check that the errors are correct
+    error_info = [*errors.error_info_map.values()][0]
+    assert error_info[0].symbol.func == 'kernel_call'
+    assert error_info[1].symbol.func == 'kernel_call'
+    assert INVALID_KERNEL_CALL_BP_GRID == error_info[0].message
+    assert INVALID_KERNEL_CALL_TP_BLOCK == error_info[1].message
```
bauom, line 152: isn't these error checked separately in the below tests?

bauom (changes requested):

bauom:
> > > To avoid triggering Codacy every time you use a kernel. I recommend using a different definition of the decorator:
> > > ```python
> > > def kernel(f):
> > >     return np.array([[f]])
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This will allow the kernel to be indexed with `[0,0]` in pure python code.
> > 
> > 
> > The problem with that is a kernel can't return. it would be incorrect to define it as such.
> 
> This is not the definition of the kernel. This is the definition of the `@kernel` decorator. Decorators take a function as an argument and return something, usually a function, here an array of functions

oh I see to indicate that if function is decorated by `@kernel` is will be indexed.

Is there a way beside disabling the codacy error to write the test when you try to index a function but it does not have the `@kernel` decorator.
as we raise a `Pyccel-GPU` error for that because in the GPU integration we made indexing a function exclusive to kernels.

EmilyBourne:
> oh I see to indicate that if function is decorated by `@kernel` is will be indexed.
> 
> Is there a way beside disabling the codacy error to write the test when you try to index a function but it does not have the `@kernel` decorator. as we raise a `Pyccel-GPU` error for that because in the GPU integration we made indexing a function exclusive to kernels.

I don't think so. I would disable the codacy error locally. You are testing incorrect syntax, it is normal that codacy won't like it

tests/cuda_test/test_kernel_semantic.py
framdani: Yes. Actually, I wasn't sure what caused the checker coverage test to fail (untested lines) so I thought that I should add this case where both of the parameters failed.

EmilyBourne: The coverage problem is happening because the execution of the cuda tests is run separately to the test which tells the coverage what code ran. This was discussed in our last meeting. It is fixed in #1291

framdani: Thank you for bringing this to my attention. I have already understood this during our previous meeting, so I will go ahead and remove this case as it is no longer necessary.

framdani: Fixed f33701bcd3d1507b53b6c8127205a6ba06b54ae8

EmilyBourne:
I am happy to approve once the decorator definition is fixed

tests/cuda_test/test_kernel_semantic.py
```diff
@@ -203,7 +203,7 @@ def test_invalid_function_call(language):
     def invalid_function_call():
         def non_kernel_func():
             pass
-        non_kernel_func[1, 2]()
+        non_kernel_func[1, 2]() # pylint: disable=E1136
```
EmilyBourne, line 206: Is the pylint command needed? The updated definition of the decorator should have fixed this

framdani: This test checks the case where a non-kernel function is called with a defined launch configuration.Therefore, this command is necessary as the non-kernel function is not decorated with the kernel decorator.

EmilyBourne: Ah of course sorry! :+1:

EmilyBourne (approved):

bauom (approved):

bauom:
@EmilyBourne
The parallel test is complaining because the docker container is running as a sudoer.
should we just make it so it ignore that problem.
I Don't see the need to redo the tests for linux inside the container while we run the separated linux tests too.

EmilyBourne:
> @EmilyBourne The parallel test is complaining because the docker container is running as a sudoer. should we just make it so it ignore that problem. I Don't see the need to redo the tests for linux inside the container while we run the separated linux tests too.

I think this should be fixed in

PR #1283, opened 2022-12-29
Fix return Constants bug

### LINKED ISSUE #1282: Can't return an imported Variable
### Describe the bug
Imported `Constant`s cannot be returned as a `Variable` is not created for the `FunctionDef` result.

### To Reproduce
```python
def ret_pi():
    from math import pi
    return pi
```

### Error details
```fortran
  function ret_pi() result(pi)

    implicit none

    real(f64) :: pi

    return

  end function ret_pi
```

This also leads to weird behaviour in the wrapper

### Expected behaviour
The following code should be generated:
```fortran
  function ret_pi() result(Out_0001)

    implicit none

    real(f64) :: Out_0001

    Out_0001 = 3.141592653589793_f64
    return

  end function ret_pi
```

EmilyBourne:
Ensure results variables exist locally. Fixes #1282

### DIFF
--- a/pyccel/parser/syntactic.py
+++ b/pyccel/parser/syntactic.py
@@ -786,12 +786,14 @@ class SyntaxParser(BasicParser):
         results = []
         result_counter = 1
 
+        local_symbols = self.scope.local_used_symbols
+
         for r in zip(*returns):
             r0 = r[0]
 
             pyccel_symbol  = isinstance(r0, PyccelSymbol)
             same_results   = all(r0 == ri for ri in r)
-            name_available = all(r0 != a.name for a in arguments)
+            name_available = all(r0 != a.name for a in arguments) and r0 in local_symbols
 
             if pyccel_symbol and same_results and name_available:
                 result_name = r0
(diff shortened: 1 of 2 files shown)


EmilyBourne:
Blocked by #1277

pyccel/parser/syntactic.py
```diff
@@ -786,12 +786,14 @@ def fill_types(ls):
...
 
+        local_symbols = self.scope.local_used_symbols
+
         for r in zip(*returns):
             r0 = r[0]
 
             pyccel_symbol  = isinstance(r0, PyccelSymbol)
             same_results   = all(r0 == ri for ri in r)
-            name_available = all(r0 != a.name for a in arguments)
+            name_available = all(r0 != a.name for a in arguments) and r0 in local_symbols
```
bauom, line 796: just asking if i understood correctly.

we are checking if we have the name in the `local_symbols` if yes then just return it without creating a `tmp` variable?

EmilyBourne: Yes exactly. For example:
```python
def f():
    a = 4
    return a
```
`a` is a local symbol so we don't need to do:
```python
def f():
    a = 4
    Out_0001 = a
    return Out_0001
```
to ensure that a locally declared variable is returned

bauom (approved):

yguclu (approved):
A very clean bug fix!

Issue #692, opened 2021-01-20
Test and document the @sympy decorator


EmilyBourne:
**Describe the bug**
The `@sympy` decorator is supposed to evaluate sympy expressions in a function and generate the python function. This function is then accelerated. E.g;
```python
@sympy
def f(x : 'float'):
    from sympy import tan
    return tan(x).diff
```
pyccel should generate:
```python
def f(x : 'float'):
    return tan(x)**2 + 1
```

This decorator is not tested. It should be tested, maybe fixed (are the headers handled?) and documented.

**Acceptance Criteria**
- [ ] Working tests are added
- [ ] Documentation is written

EmilyBourne:
This is superceded by the lambdify function

PR #1284, opened 2022-12-29
Add developer docs for the syntactic stage

EmilyBourne:
Add developer docs for the syntactic stage. As per the docs, fatal errors should be used sparingly to give the user as much information as possible. This was not the case, so this PR also reduces the severity of most errors issued by the syntactic stage


EmilyBourne:
@yguclu The remaining Codacy error is due to RemarkLint not recognising markdown's footnote syntax.

developer_docs/syntactic_stage.md
```diff
@@ -0,0 +1,82 @@
+# Syntactic Stage
+
+The syntactic stage is described by the file [pyccel.parser.syntactic](../pyccel/parser/syntactic.py)
+
+The syntactic stage serves 4 main purposes:
+1.  Convert Python's [AST](https://docs.python.org/3/library/ast.html) (abstract syntax tree) representation of the python file to Pyccel's AST representation (objects of the classes in the folder [pyccel.ast](../pyccel/ast))
```
jalalium, line 6: I think it is not clear how this conversion is done. It would be ideal if we could map each of the following to its equivalent in pyccel.
````
-- ASDL's 4 builtin types are:
-- identifier, int, string, constant

module Python
{
    mod = Module(stmt* body, type_ignore* type_ignores)
        | Interactive(stmt* body)
        | Expression(expr body)
        | FunctionType(expr* argtypes, expr returns)

    stmt = FunctionDef(identifier name, arguments args,
                       stmt* body, expr* decorator_list, expr? returns,
                       string? type_comment)
          | AsyncFunctionDef(identifier name, arguments args,
                             stmt* body, expr* decorator_list, expr? returns,
                             string? type_comment)

          | ClassDef(identifier name,
             expr* bases,
             keyword* keywords,
             stmt* body,
             expr* decorator_list)
          | Return(expr? value)

          | Delete(expr* targets)
          | Assign(expr* targets, expr value, string? type_comment)
          | AugAssign(expr target, operator op, expr value)
          -- 'simple' indicates that we annotate simple name without parens
          | AnnAssign(expr target, expr annotation, expr? value, int simple)

          -- use 'orelse' because else is a keyword in target languages
          | For(expr target, expr iter, stmt* body, stmt* orelse, string? type_comment)
          | AsyncFor(expr target, expr iter, stmt* body, stmt* orelse, string? type_comment)
          | While(expr test, stmt* body, stmt* orelse)
          | If(expr test, stmt* body, stmt* orelse)
          | With(withitem* items, stmt* body, string? type_comment)
          | AsyncWith(withitem* items, stmt* body, string? type_comment)

          | Match(expr subject, match_case* cases)

          | Raise(expr? exc, expr? cause)
          | Try(stmt* body, excepthandler* handlers, stmt* orelse, stmt* finalbody)
          | TryStar(stmt* body, excepthandler* handlers, stmt* orelse, stmt* finalbody)
          | Assert(expr test, expr? msg)

          | Import(alias* names)
          | ImportFrom(identifier? module, alias* names, int? level)

          | Global(identifier* names)
          | Nonlocal(identifier* names)
          | Expr(expr value)
          | Pass | Break | Continue

          -- col_offset is the byte offset in the utf8 string the parser uses
          attributes (int lineno, int col_offset, int? end_lineno, int? end_col_offset)

          -- BoolOp() can use left & right?
    expr = BoolOp(boolop op, expr* values)
         | NamedExpr(expr target, expr value)
         | BinOp(expr left, operator op, expr right)
         | UnaryOp(unaryop op, expr operand)
         | Lambda(arguments args, expr body)
         | IfExp(expr test, expr body, expr orelse)
         | Dict(expr* keys, expr* values)
         | Set(expr* elts)
         | ListComp(expr elt, comprehension* generators)
         | SetComp(expr elt, comprehension* generators)
         | DictComp(expr key, expr value, comprehension* generators)
         | GeneratorExp(expr elt, comprehension* generators)
         -- the grammar constrains where yield expressions can occur
         | Await(expr value)
         | Yield(expr? value)
         | YieldFrom(expr value)
         -- need sequences for compare to distinguish between
         -- x < 4 < 3 and (x < 4) < 3
         | Compare(expr left, cmpop* ops, expr* comparators)
         | Call(expr func, expr* args, keyword* keywords)
         | FormattedValue(expr value, int conversion, expr? format_spec)
         | JoinedStr(expr* values)
         | Constant(constant value, string? kind)

         -- the following expression can appear in assignment context
         | Attribute(expr value, identifier attr, expr_context ctx)
         | Subscript(expr value, expr slice, expr_context ctx)
         | Starred(expr value, expr_context ctx)
         | Name(identifier id, expr_context ctx)
         | List(expr* elts, expr_context ctx)
         | Tuple(expr* elts, expr_context ctx)

         -- can appear only in Subscript
         | Slice(expr? lower, expr? upper, expr? step)

          -- col_offset is the byte offset in the utf8 string the parser uses
          attributes (int lineno, int col_offset, int? end_lineno, int? end_col_offset)

    expr_context = Load | Store | Del

    boolop = And | Or

    operator = Add | Sub | Mult | MatMult | Div | Mod | Pow | LShift
                 | RShift | BitOr | BitXor | BitAnd | FloorDiv

    unaryop = Invert | Not | UAdd | USub

    cmpop = Eq | NotEq | Lt | LtE | Gt | GtE | Is | IsNot | In | NotIn

    comprehension = (expr target, expr iter, expr* ifs, int is_async)

    excepthandler = ExceptHandler(expr? type, identifier? name, stmt* body)
                    attributes (int lineno, int col_offset, int? end_lineno, int? end_col_offset)

    arguments = (arg* posonlyargs, arg* args, arg? vararg, arg* kwonlyargs,
                 expr* kw_defaults, arg? kwarg, expr* defaults)

    arg = (identifier arg, expr? annotation, string? type_comment)
           attributes (int lineno, int col_offset, int? end_lineno, int? end_col_offset)

    -- keyword arguments supplied to call (NULL identifier for **kwargs)
    keyword = (identifier? arg, expr value)
               attributes (int lineno, int col_offset, int? end_lineno, int? end_col_offset)

    -- import name with optional 'as' alias.
    alias = (identifier name, identifier? asname)
             attributes (int lineno, int col_offset, int? end_lineno, int? end_col_offset)

    withitem = (expr context_expr, expr? optional_vars)

    match_case = (pattern pattern, expr? guard, stmt* body)

    pattern = MatchValue(expr value)
            | MatchSingleton(constant value)
            | MatchSequence(pattern* patterns)
            | MatchMapping(expr* keys, pattern* patterns, identifier? rest)
            | MatchClass(expr cls, pattern* patterns, identifier* kwd_attrs, pattern* kwd_patterns)

            | MatchStar(identifier? name)
            -- The optional "rest" MatchMapping parameter handles capturing extra mapping keys

            | MatchAs(pattern? pattern, identifier? name)
            | MatchOr(pattern* patterns)

             attributes (int lineno, int col_offset, int end_lineno, int end_col_offset)

    type_ignore = TypeIgnore(int lineno, string tag)
}
```
This would also give an overview of has been implemented so far.
Maybe write an asdl for pyccel too

EmilyBourne:
@yguclu The remaining coverage problems cannot be tested as they are there for future proofing

developer_docs/syntactic_stage.md
```diff
@@ -0,0 +1,82 @@
...
+The key line of the function `parse` is the call to `self._visit(self.fst)`.
+All elements of the tree must be visited.
+The `_visit` function internally calls a function named `_visit_X`, where `X` is the type of the object.
+This function must have the form:
+```python
+def _visit_ClassName(self, stmt):
+    ...
+```
+Each of these functions should internally call the `_visit` function on each of the elements of the object to obtain Pyccel AST nodes which can be combined to create a Pyccel AST node representing the current object.
+
```
jalalium, line 27: I think we should do the same regarding nodes and their respective _visit functions. This may seem unnecessary but it will allow the redirection to the docs of all visit functions.

developer_docs/syntactic_stage.md
```diff
@@ -0,0 +1,82 @@
...
+A child scope can be created using one of the following functions (for more details see the docstrings in [pyccel.parser.scope](../pyccel/parser/scope.py):
+-   `Scope.new_child_scope`
+-   `Scope.create_new_loop_scope`
+-   `Scope.create_product_loop_scope`
+
+Occasionally it is necessary to create objects in the syntactic stage.
+The `Scope` functions should be used for this purpose to avoid name collisions.
+See the [scope](scope.md) docs for more details.
+In all cases it is preferable to delay this stage as much as possible to ensure that as much information is known about the scope as possible.
```
jalalium, line 72: to delay this stage as ~~much~~ **long** as possible

developer_docs/syntactic_stage.md
```diff
@@ -0,0 +1,82 @@
+# Syntactic Stage
+
+The syntactic stage is described by the file [pyccel.parser.syntactic](../pyccel/parser/syntactic.py)
+
+The syntactic stage serves 4 main purposes:
+1.  Convert Python's [AST](https://docs.python.org/3/library/ast.html) (abstract syntax tree) representation of the python file to Pyccel's AST representation (objects of the classes in the folder [pyccel.ast](../pyccel/ast))
+2.  Raise an error for any syntax used that is not yet supported by pyccel
+3.  Convert header comments from strings to Pyccel's AST representation
```
jalalium, line 8: Can this line be used to reference the CommentLine and the CommentMultiline classes?

developer_docs/syntactic_stage.md
```diff
@@ -0,0 +1,82 @@
...
+
+The key line of the function `parse` is the call to `self._visit(self.fst)`.
+All elements of the tree must be visited.
+The `_visit` function internally calls a function named `_visit_X`, where `X` is the type of the object.
+This function must have the form:
+```python
+def _visit_ClassName(self, stmt):
+    ...
+```
+Each of these functions should internally call the `_visit` function on each of the elements of the object to obtain Pyccel AST nodes which can be combined to create a Pyccel AST node representing the current object.
```
jalalium, line 26: An example is greatly appreciated here. Maybe a walktrough of the visits for a function definition

developer_docs/syntactic_stage.md
```diff
@@ -0,0 +1,82 @@
...
+This function takes several arguments (see docstring for more details).
+The most important arguments are:
+-   _message_ : Describe the issue that lead to the error
+
+-   _symbol_ : The Python ast object should be passed here. This object contains information about its position in the file (line number, column) which ensures the user can more easily locate their error
+
+-   _severity_ : The severity level must be one of the following:
+    -   _warning_ : An error will be printed but Pyccel will continue executing
+    -   _error_ : An error will be printed but Pyccel will continue executing the syntactic stage
+    -   _fatal_ : An error will be printed and Pyccel will stop executing. This level should rarely be needed in the syntactic stage as a failure in one function should not affect the execution of another. It is preferable to show the users all errors at once
```
jalalium, line 42: Is execution the right word here? it gives the impression that we are not in the syntactic stage, it has a _runtimy_ feel to it.

developer_docs/syntactic_stage.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Headers
+
+The headers (type declarations/openmp pragmas/etc) also have their own syntax which cannot be parsed by Python's ast module.
+The module [textx](http://textx.github.io/textX/stable/) is used to parse these statements.
+The files describing the _textx_ grammar are found in the folder [pyccel.parser.grammar](../pyccel/parser/grammar).
+From these files _textx_ generates instances of the classes found in the folder [pyccel.parser.syntax](../pyccel/parser/syntax).
+These instances can then be inserted into the annotated syntax tree.
```
jalalium, line 50: This is the first mention of _annotated syntax tree_. Do you think it should have been passively introduced above?

developer_docs/syntactic_stage.md
EmilyBourne: I'm not sure what you mean? Are you perhaps confusing `pyccel` and `epyccel`? The executable `pyccel` stops executing. This is at runtime (although in python there is not really an alternative as there is no compile time)

developer_docs/syntactic_stage.md
EmilyBourne: Oops. It should be _abstract syntax tree_
Fixed 046b880d

developer_docs/syntactic_stage.md
EmilyBourne: Fixed 8bca0416

developer_docs/syntactic_stage.md
EmilyBourne: I don't think it should be clear how the conversion is done at this point. This is an overview of how the stage works. Not a description of everything in the file. And at this point, it is a very abstract overview to give the aims. I have added a link to make it clear that more explanations will be provided later in the file: c83a90cd

What you describe could be placed in the relevant section, but personally, I tend to think that long lists like this are only read if the reader is looking for something specific, otherwise, they just interrupt the flow of the reading.
So, while a list could be useful somewhere, I don't think this is the place for it. I think there should instead be a link to a page where this is available. However, if we have a link, then I don't think it necessarily needs to link to our docs. Letting python document python code ensures that the reader can easily find the documentation for their version of python. Just as an example, in your list above `NamedExpr` only exists from python 3.8 onwards.

I would be interested to know what other people think though.  @pyccel/pyccel-dev do you prefer docs to have everything in one place? Or to have a clear overview with more detailed information available if necessary?

I am also against maintaining any kind of list like this unless it is auto-generated.
If it is not auto-generated then we will forget to maintain it.
I am not familiar with ADSL so I don't know if there is a way to do this?
Is there something in sphinx?

>  It would be ideal if we could map each of the following to its equivalent in pyccel.

You mean copy/paste the `_visit_X` functions into a file? I don't know that I see this as particularly useful. Is it really easier for developers to search for this information in text rather than inside a file? Again I'm happy to go with the majority if other people disagree @pyccel/pyccel-dev but I would rather this was in another (ideally auto-generated) file

developer_docs/syntactic_stage.md
EmilyBourne: > I think we should do the same 

Meaning? Write a list of all `_visit` functions? Is this not what the doc of all visit functions is for?

developer_docs/syntactic_stage.md
EmilyBourne: I would rather do that in the header section than in the overview. But otherwise yes, it can be updated to do this. Should I point at the code given that the sphinx docs are not yet ready?

developer_docs/syntactic_stage.md
EmilyBourne: A `FunctionDefinition` is one of the most complicated objects in pyccel, so I'm not sure it is well adapted to an introductory example. Maybe a for loop?

EmilyBourne: Better fb39e982 ?

developer_docs/syntactic_stage.md
jalalium: Thank you! I think we should come up with a scheme to reference missing documentation so it can be written as soon as possible. Maybe distinguish between pointing to autogenerated doc / hand written doc. With the sphinx doc being what it is currently, i don't see how we could point to specific classes and methods. pages are generated for modules only.

developer_docs/syntactic_stage.md
jalalium: Yes, thank you very much!

developer_docs/syntactic_stage.md
jalalium: No, I meant that `execution of another` implies that a function is being executed, when in actuality it is rather syntacticly parsed or its syntactic parsing is executed not the function itself.  the following sounds better to me:
> This level should rarely be needed in the syntactic stage as a failure in __parsing__ one function should not affect the __parsing__ of another. It is preferable to show the users all errors at once

developer_docs/syntactic_stage.md
jalalium: Nevermind then, it stemmed from the same need to have everything in one place.

developer_docs/syntactic_stage.md
EmilyBourne: For now I am not pointing to anything that does not exist yet. Pointing to code is equally hard as you need to keep the line numbers updated.

Does the sphinx HTML not contain ids that you can link to? Maybe only once you have the settings right so everything has its own section?

developer_docs/syntactic_stage.md
EmilyBourne: Hmmm, now I see what you mean. But the functions I am talking about here are the `_visit_X` functions not the python functions being translated.  E.g. the failure of `_visit_Dict` shouldn't stop `_visit_List` working. You can have both a list and a dictionary in the same function

EmilyBourne: How about:
> This level should rarely be needed in the syntactic stage as a failure in one visitation function should not affect the execution of another

jalalium: Sorry, I misunderstood. I thought you were taking a function as an example of what's beiing parsed. Now, I totally agree with you. What do you think of clearly stating that we are talking about `_visit_X` functions?

EmilyBourne: I agree. That is what I tried to do by saying "visitation function". Is this not clear? Do you have a suggestion?

EmilyBourne: > This level should rarely be needed in the syntactic stage as a failure in one visitation function (`_visit_X`) should not affect the execution of another

?

jalalium: Perfect!

EmilyBourne: Fixed : d607b56b
(By the way, it is for you to decide when I have adequately responded to your comment and resolve the conversations once you're happy)

aihya (approved):
Looks good to me.

Pinkyboi (approved):
Looks all good to me.

developer_docs/syntactic_stage.md
AbdelkhalekBAROUDI, line 26: which functions does **these functions ** refer to ?

EmilyBourne: Better b71dd369 ?

AbdelkhalekBAROUDI: Yes, it's much more explicit now. Thank you !

developer_docs/syntactic_stage.md
```diff
@@ -0,0 +1,139 @@
...
+2.  [**Errors**](#Errors) : Raise an error for any syntax used that is not yet supported by pyccel
+3.  [**Headers**](#Headers) : Convert header comments from strings to Pyccel's AST representation
+4.  [**Scoping**](#Scoping) : Collect the name of all variables in each scope (see [scope](scope.md) for more details) to ensure no name collisions can occur if pyccel generates Variable names
+
+## Navigation and AST Creation
+
+The entry point for the class `SyntaxParser` is the function `parse`.
+This function is called from the constructor to examine a FST (Full Syntax Tree) object.
+The FST is the output of Python's `ast.parse` function with comments inserted using the [`pyccel.parser.extend_tree.extend_tree`](../pyccel/parser/extend_tree.py).
+Python's ast module discards all comments, so the function [`pyccel.parser.extend_tree.extend_tree`](../pyccel/parser/extend_tree.py) is needed to read the file and reinsert these comments at the correct location.
```
AbdelkhalekBAROUDI, line 16: I find that pyccel.parser.extend_tree.extend_tree is redundant. Wouldn't be better to write : 

```suggestion
The FST is the output of Python's `ast.parse` function with comments inserted at the correct location using the [`pyccel.parser.extend_tree.extend_tree`](../pyccel/parser/extend_tree.py), since Python's ast module discards all comments.
```
instead of lines 15 and 16 ?

EmilyBourne: I agree. Better: a098fb89 ?

AbdelkhalekBAROUDI: yes, much better ?

AbdelkhalekBAROUDI (approved):
This is very insightful !!

jalalium (approved):

developer_docs/overview.md
```diff
@@ -86,4 +86,4 @@ If the error prevents further translation (e.g. the type of an object is now unk
 
 ## Getting Help
 
-While discussions within the associated GitHub issue are often sufficient, should you require more help do not hesitate to ask one of the other developers to add you to our slack: <pyccel.slack.com>
+While discussions within the associated GitHub issue are often sufficient, should you require more help do not hesitate to ask one of the other developers to add you to our Slack channel <pyccel.slack.com>.
```
EmilyBourne, line 89: I think this line needs modifying to mention discord instead anyway

yguclu: True, but do we have a public channel on Discord yet? cc: @bauom

bauom: the `GENERAL` category have 2 channels `general` and `random` these can be viewed by anyone.
should we rename for something else ?

pyccel/parser/syntactic.py
```diff
@@ -176,10 +176,9 @@ def _treat_comment_line(self, line, stmt):
                     self._metavars[str(expr.name)] = expr.value
                     expr = EmptyNode()
             else:
-
-                raise errors.report(PYCCEL_INVALID_HEADER,
+                errors.report(PYCCEL_INVALID_HEADER,
                               symbol = stmt,
-                              severity='error')
+                              severity='fatal')
```
yguclu, line 181: Why did you increase the severity in this case?

pyccel/parser/syntactic.py
```diff
@@ -741,11 +737,11 @@ def fill_types(ls):
 
         body = stmt.body
 
-        if 'sympy' in decorators.keys():
+        if 'sympy' in decorators:
             # TODO maybe we should run pylint here
             stmt.decorators.pop()
             func = SympyFunction(name, arguments, [],
-                    [stmt.__str__()])
+                    [str(stmt)])
```
yguclu, line 744: This should fit in a single line, which is more readable

pyccel/parser/syntactic.py
```diff
@@ -210,8 +209,8 @@ def _visit(self, stmt):
             return result
 
         # Unknown object, we raise an error.
-        errors.report(PYCCEL_RESTRICTION_UNSUPPORTED_SYNTAX, symbol=stmt,
-                      severity='fatal')
+        return errors.report(PYCCEL_RESTRICTION_UNSUPPORTED_SYNTAX, symbol=stmt,
+                      severity='error')
```
yguclu, line 213: I find it a bit strange to call `errors.report` within a `return` statement. I suppose that `errors.report == None`, but `return None` is automatically added at the end of every Python function...

pyccel/parser/syntax/headers.py
```diff
@@ -72,7 +72,8 @@ def expr(self):
         if any(isinstance(d_type, FuncType) for d_type in self.dtypes):
             msg = 'Functions in a template are not supported yet'
             errors.report(msg,
-                        severity='fatal')
+                        severity='error')
+            return EmptyNode()
```
yguclu, line 76: Why do you need an `EmptyNode` here, instead of writing `return errors.report(msg, severity='error')`?

developer_docs/overview.md
bauom: we maybe should write a bigger message in the `welcome` channel and that were we should link.
here is a permanent link to the welcome channel 
https://discord.gg/2Q6hwjfFVb

bauom: It is automatic now them will have access to the public channels which now are 
GENERAL : `general`, `random`
MEETING ROOMS: `room1`, `room2`


There is a way to make access temporary if we want.

pyccel/parser/syntactic.py
EmilyBourne: Fixed 772f82f8

pyccel/parser/syntactic.py
EmilyBourne: This usually happens with `severity='fatal'` (so the return doesn't actually happen) to avoid codacy's errors about [inconsistent-return-statements](https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/inconsistent-return-statements.html).

This shouldn't be used here. Fixed 721fce9d

pyccel/parser/syntax/headers.py
EmilyBourne: Added explanation: d63e27fb

developer_docs/overview.md
```diff
@@ -56,18 +66,18 @@ Therefore the print functions which are acceptable for visiting this object are:
...
 The objects as understood by Pyccel are each described by classes which inherit from [pyccel.ast.basic.Basic](../pyccel/ast/basic.py).
 These classes are found in the [ast](../pyccel/ast) folder.
 The objects in the Abstract Syntax Tree (AST) are described in several files.
-There is one file for each supported extension module and files to group concepts, e.g. literals/operators/built-in functions
+There is one file for each supported extension module and files to group concepts, e.g. literals/operators/built-in functions.
 
 ## Error System
 
-Pyccel tries to fail cleanly and raise readable errors for users. This is managed using the error handling module found in the [errors](../pyccel/errors) folder. In order to raise an error 2 things must be done:
+Pyccel tries to fail cleanly and raise readable errors for users. This is managed using the error handling module found in the [errors](../Pyccel/errors) folder. In order to raise an error 2 things must be done:
```
EmilyBourne, line 80: This change is wrong. Pyccel here is the name of a folder that is being linked to. Capitalisation will break the link

pyccel/parser/syntactic.py
EmilyBourne: Not sure how I got round this before, probably because of the broken line. In any case I now have a coverage error but I don't know enough about what this is supposed to do to add a test

pyccel/parser/syntax/headers.py
EmilyBourne: Hmm, actually I've just had a look at this and `None` is the one other value that is acceptable. Would you prefer I use `return None`?

pyccel/parser/syntactic.py
EmilyBourne: See #692. There were tests but they are deactivated, see #330

pyccel/parser/syntax/headers.py
yguclu: Yes. We do not use `EmptyNode` extensively: https://github.com/pyccel/pyccel/search?q=EmptyNode

pyccel/parser/syntactic.py
yguclu: All right. Can we ask the CoverageChecker to ignore this line?

pyccel/parser/syntax/headers.py
EmilyBourne: fixed [54e8ab2](https://github.com/pyccel/pyccel/pull/1284/commits/54e8ab2188091f2da7c023fc59d2286ce42ef32c)

pyccel/parser/syntactic.py
yguclu: Good point.

From what I see [here](https://github.com/pyccel/pyccel/blob/26a14bfa9cae3e21b5175892b77f7bd1f7d334e2/pyccel/errors/errors.py#L296-L304) we now raise a `PyccelSyntaxError` which is a proper exception, while we were raising a `None` earlier.

EmilyBourne: Yes that sounds right

pyccel/parser/syntax/headers.py
yguclu: Sorry but I was probably too quick with my comment. According to the [docstring of `EmptyNode`](https://github.com/pyccel/pyccel/blob/261c1d9027efc3e7a2c33b6a1e7c081c48f0a602/pyccel/ast/core.py#L3651-L3674), objects of this class should be used as much as possible instead of `None` :sweat_smile:

EmilyBourne: Oops, I removed the commit

developer_docs/overview.md
yguclu: This sounds good to me. I let you handle the server access as you see fit :wink:

yguclu (approved):
Excellent explanations! These docs will be extremely useful

pyccel/parser/syntactic.py
yguclu: Not a problem, the coverage check is not required anyway

PR #1286, opened 2022-12-30
Fix wrong interpretation of AugAssign to AliasAssign

### LINKED ISSUE #1281: `AugAssign` interpreted as `AliasAssign`
### Describe the bug
When calling an `AugAssign` operator on a pointer, pyccel interprets this as an `AliasAssign`. This is probably due to the fact that `AugAssign` inherits from `Assign` and this difference is not taken into consideration when the `AliasAssign` is created. The problem exists in all languages so it must be a semantic problem.

### To Reproduce
```python
import numpy as np

if __name__ == '__main__':
    a = np.ones(5)
    b = a
    b += 3
```

### Error details
Here is the generated code in Fortran:
```fortran
program prog_prog_tmp4

  use tmp4

        C_DOUBLE
  implicit none

  real(f64), allocatable, target :: a(:)
  real(f64), pointer :: b(:)

  allocate(a(0:4_i64))
  a = 1.0_f64
  b(0:) => a
  b(0:) => 3_i64
  if (allocated(a)) then
    deallocate(a)
  end if

end program prog_prog_tmp4
```
C:
```c
#include "tmp4.h"
#include <stdlib.h>
#include "ndarrays.h"
#include <stdint.h>
int main()
{
    a = array_create(1, (int64_t[]){INT64_C(5)}, nd_double, false);
    array_fill((double)1.0, a);
    alias_assign(&b, a);
    free_pointer(b);
    b = INT64_C(3);
    free_array(a);
    free_pointer(b);
    return 0;
}
```
and python:
```python
from numpy import ones

if __name__ == "__main__":
    a = ones((5), dtype=float)
    b = a
    b = 3
```

### Expected behaviour
The codes should be:
Fortran:
```fortran
program prog_prog_tmp4

  use tmp4

        C_DOUBLE
  implicit none

  real(f64), allocatable, target :: a(:)
  real(f64), pointer :: b(:)

  allocate(a(0:4_i64))
  a = 1.0_f64
  b(0:) => a
  b(0:) = b(0:) + 3_i64
  if (allocated(a)) then
    deallocate(a)
  end if

end program prog_prog_tmp4
```
C:
```c
#include "tmp4.h"
#include <stdlib.h>
#include "ndarrays.h"
#include <stdint.h>
int main()
{
    a = array_create(1, (int64_t[]){INT64_C(5)}, nd_double, false);
    array_fill((double)1.0, a);
    alias_assign(&b, a);
    for (i = INT64_C(0); i < INT64_C(5); i += INT64_C(1))
    {
        GET_ELEMENT(b, nd_double, (int64_t)i) += INT64_C(3);
    }
    free_array(a);
    free_pointer(b);
    return 0;
}
```
and python:
```python
from numpy import ones

if __name__ == "__main__":
    a = ones((5), dtype=float)
    b = a
    b += 3
```

aihya:
Fixes #1281

This issue is solving the wrong interpretation of `AugAssign` into `AliasAssign`.

```python
import numpy as np

if __name__ == '__main__':
    a = np.ones(5)
    b = a
    b += 3
```
The code above, when converted to C, used to generate this:
```C
#include "test.h"
#include <stdlib.h>
#include "ndarrays.h"
#include <stdint.h>
int main()
{
    a = array_create(1, (int64_t[]){INT64_C(5)}, nd_double, false);
    array_fill((double)1.0, a);
    alias_assign(&b, a);
    free_pointer(b);
    b = INT64_C(3);
    free_array(a);
    free_pointer(b);
    return 0;
}
```
Now it generates this:
```C
#include "test.h"
#include <stdlib.h>
#include "ndarrays.h"
#include <stdint.h>
#include <stdio.h>
#include <inttypes.h>
int main()
{
    int64_t i;
    int64_t i_0001;
    a = array_create(1, (int64_t[]){INT64_C(5)}, nd_double, false);
    array_fill((double)1.0, a);
    alias_assign(&b, a);
    for (i = INT64_C(0); i < b.shape[INT64_C(0)]; i += INT64_C(1))
    {
        GET_ELEMENT(b, nd_double, (int64_t)i) += INT64_C(3);
    }
    free_array(a);
    free_pointer(b);
    return 0;
}
```
which removes an unnecessary de-allocation of `b` after the call to `alias_assign(&b, a)` and replaces `b = INT64_C(3)` with for loop to broadcast `+=` operation on all elements of `b`, which point on `a`.

For Fortran:
```fortran
program prog_prog_test

  use test_0001

        C_INT64_T
  implicit none

  real(f64), allocatable, target :: a(:)
  real(f64), pointer :: b(:)

  allocate(a(0:4_i64))
  a = 1.0_f64
  b(0:) => a
  b(0:) => 3_i64
  if (allocated(a)) then
    deallocate(a)
  end if

end program prog_prog_test
```
the line `b(0:) => 3_i64` is replaced with `b = b + 3_i64`


aihya:
@EmilyBourne the fix I did seems to be working fine for linux. For some reason there's an `ImportError` raised for both windows and macos and it doesn't look it has any relationship to the fix in anyway.

EmilyBourne:
> @EmilyBourne the fix I did seems to be working fine for linux. For some reason there's an `ImportError` raised for both windows and macos and it doesn't look it has any relationship to the fix in anyway.

The MacOS and Windows errors are on all branches. See issue #1272 and the associated PR. We're just waiting for someone with merge rights to fix it ;)

tests/epyccel/test_epyccel_augassign.py
```diff
@@ -0,0 +1,170 @@
...
+    x1_int     = np.ones(5, dtype=int)
+    x1_float   = np.ones(5, dtype=float)
+    x1_complex = np.ones(5, dtype=complex)
+    x2_int     = np.ones(5, dtype=int)
+    x2_float   = np.ones(5, dtype=float)
+    x2_complex = np.ones(5, dtype=complex)
+
+    assert f_int(x1_int) == f_int_epyc(x2_int) == 42
+    assert f_float(x1_float) == f_float_epyc(x2_float) == 4.2
+    assert f_complex(x1_complex) == f_complex_epyc(x2_complex) == (4.0 + 2.0j)
```
EmilyBourne, line 28: It is not necessary to specify the  value that you think will be returned the pure python execution of the function is there to obtain this value.

By the way you need:
```python
    x1_int     = np.zeros(5, dtype=int)
    x1_float   = np.zeros(5, dtype=float)
    x1_complex = np.zeros(5, dtype=complex)
    x2_int     = np.zeros(5, dtype=int)
    x2_float   = np.zeros(5, dtype=float)
    x2_complex = np.zeros(5, dtype=complex)
```

to obtain the result you expect.

It would also be good to check that the pointer is working correctly and that the argument has indeed changed

aihya: > It would also be good to check that the pointer is working correctly and that the argument has indeed changed
What pointer? I don't see what you mean.

EmilyBourne:
```
b = a
```
is not a copy. `b` points at `a`. So when `b` is modified, `a` should also be modified. It would be good to check that:
```python
np.all_equal(x1_int, x2_int)
```

EmilyBourne:
@aihya I see you have marked the PR as ready for review, but have not added any labels. Please see the [review process docs](https://github.com/pyccel/pyccel/blob/master/developer_docs/review_process.md#review-stage-labels) for details about how PRs are managed in pyccel

pyccel/parser/semantic.py
```diff
@@ -1210,7 +1210,7 @@ def _ensure_infered_type_matches_existing(self, dtype, d_var, var, is_augassign,
                     self._current_fst_node.col_offset),
                         severity='error', symbol=var.name)
 
-        elif var.is_ndarray and var.is_alias:
+        elif var.is_ndarray and var.is_alias and not is_augassign:
             # we allow pointers to be reassigned multiple times
             # pointers reassigning need to call free_pointer func
             # to remove memory leaks
```
Pinkyboi, line 1216: Maybe i'm missing something but why do we need to free aliases ?

EmilyBourne: Aliases contain shape and strides arrays which are independent of the original array. This allows you to do:
```
a = np.ones(4,6)
b = a[:2,:3]
a.shape # 4,6
b.shape # 2,2
```
These arrays need freeing when the alias goes out of scope

Pinkyboi: Oh i see so we free the shape and strides in that case makes sense.

Pinkyboi: Well it's seems fine to me then.

EmilyBourne: In that case, you should `approve` the PR

Pinkyboi (approved):

pyccel/parser/semantic.py
```diff
@@ -2632,7 +2632,10 @@ def _visit_Assign(self, expr, **settings):
...
 
             if is_pointer_i:
-                new_expr = AliasAssign(l, r)
+                if isinstance(expr, AugAssign):
+                    new_expr = AugAssign(l, expr.op, r)
+                else:
+                    new_expr = AliasAssign(l, r)
 
             elif isinstance(expr, AugAssign):
                 new_expr = AugAssign(l, expr.op, r)
```
EmilyBourne, line 2641: Why do you duplicate this line of code:
```python
new_expr = AugAssign(l, expr.op, r)
```
rather than re-ordering the `if`?
```python
if isinstance(expr, AugAssign):
    new_expr = AugAssign(l, expr.op, r)
elif is_pointer_i:
    new_expr = AliasAssign(l, r)
```

tests/epyccel/test_epyccel_augassign.py
```diff
@@ -0,0 +1,212 @@
...
+    f_int(x1_int)
+    f_float(x1_float)
+    f_complex(x1_complex)
+    f_int_epyc(x2_int)
+    f_float_epyc(x2_float)
+    f_complex_epyc(x2_complex)
+
+    assert np.array_equal(x1_int, x2_int)
+    assert np.array_equal(x1_float, x2_float)
+    assert np.array_equal(x1_complex, x2_complex)
```
EmilyBourne, line 35: Why do you not test the output of the function as well as the argument?

EmilyBourne (changes requested):

tests/epyccel/test_epyccel_augassign.py
aihya: I don't think there is a need for that, `np.array_equal` checks the elements of the arrays.

EmilyBourne: But `a` and `b` are not the same object. When everything is working correctly I agree that `np.array_equal` is enough, but if something is broken, having both allows us to tell the difference between a broken `AugAssign` and a broken pointer

tests/epyccel/modules/augassign.py
```diff
@@ -0,0 +1,155 @@
...
+def augassign_mul_1d_float(a):
+    b = a
+    b *= 4.2
+    return b[0]
+
+@types('complex[:]')
+def augassign_mul_1d_complex(a):
+    b = a
+    b *= (4.0 + 2.0j)
+    return b[0]
```
ohachim, line 99: Checking the first elements is enough?

ohachim: nvm, you are using `array_equal`, but that means the return is not necessary?

pyccel/parser/semantic.py
aihya: fixed https://github.com/pyccel/pyccel/pull/1286/commits/0345a8d7484f597364948d4d65eff6a70405ff2a https://github.com/pyccel/pyccel/pull/1286/commits/f8459aa46b3f9da260c48f8c45604567712663ad

tests/epyccel/modules/augassign.py
aihya: It's use now in the test functions: https://github.com/pyccel/pyccel/pull/1286/commits/7f2400caf3017a9be96cca9df9f72595526e50ac

tests/epyccel/test_epyccel_augassign.py
aihya: https://github.com/pyccel/pyccel/pull/1286/commits/7f2400caf3017a9be96cca9df9f72595526e50ac

EmilyBourne (approved):

ohachim (approved):

yguclu (approved):
Well done!

PR #1291, opened 2023-01-06
Add coverage for cuda tests

EmilyBourne:
Move cuda tests to run with linux tests. Once they have run the coverage results from both are combined to test the coverage. This PR also includes the bug fixes for the coverage checker scripts


.github/actions/coverage_collection/action.yml
```diff
@@ -6,7 +6,6 @@ runs:
     - name: Coverage collection
       run: |
         coverage combine
-        coverage xml
```
bauom, line 9: why did we remove this but we still used it in the Cuda coverage collection

.github/workflows/Github_pytest.yml
```diff
@@ -95,10 +89,94 @@ jobs:
...
+        with:
+          name: cuda-coverage-artifact
+          path: .coverage
+          retention-days: 1
+
+  CoverageCollection:
+
+    runs-on: ubuntu-latest
+    needs: [Linux, Cuda]
+    if: ${{ always() && needs.Linux.result == 'success' && needs.Cuda.result != 'failure' }}
```
bauom, line 130: is there more that `success`, `failure` in the result ?

bauom (changes requested):

.github/actions/coverage_collection/action.yml
EmilyBourne: `coverage combine` combines `.coverage.X` files into a `.coverage` file. Once we have a `.coverage` file describing all the tests, we can call `coverage xml` which converts a `.coverage` file into an XML summary.

If I leave this here then the output is the XML summary not the `.coverage` file and then it is much harder to combine the outputs from the linux tests and the cuda tests

.github/workflows/Github_pytest.yml
EmilyBourne: It can also be `skipped`

.github/actions/coverage_collection/action.yml
bauom: Oh I see thank you.

bauom (approved):

PR #1292, opened 2023-01-06
Adding git actions for automating documentation validation and coverage

jalalium:
Create new  `doc_coverage` workflow which automates documentation validation and coverage in two ways:

1. Use `docstr-coverage` to make sure that all new functions and classes have a docstring.
2. Use `git diff` to list the files updated by the PR, then parse these files to extract all their objects. Afterwards, the docstrings are validated to make sure they follow `numpydoc` standards.


ci_tools/summarise_doccoverage.py
```diff
@@ -0,0 +1,62 @@
...
+        for obj in added_obj:
+            print(f'\t * {obj}')
+        print()
+    sys.exit(1)
+else:
+    print('Success!')
+    print('Base Branch Summary')
+    print(results['base_summary'])
+    print('Compare Branch Summary')
+    print(results['compare_summary'])
```
EmilyBourne, line 62: Why not output to the summary file so you can have a neat markdown output of this?

ci_tools/list_changed_files.py
```diff
@@ -0,0 +1,23 @@
+""" Script to list the python files changed by the PR
+"""
+import argparse
+
+parser = argparse.ArgumentParser(
+    description='Collect the files that have been modified by the PR to check their docstrings')
```
EmilyBourne, line 6: Can you reuse `git_evaluation_tools.py` for this? That way we don't need to maintain 2 codes which do very similar things

ci_tools/summarise_doccoverage.py
jalalium: I was confused about this point. I am working under the assumption that marking a test as failing is by exiting with an error code. That's why i thought that i should output the results then exit. Is there another way of marking the test as failing?

ci_tools/list_changed_files.py
jalalium: Thank you for pointing that out! i will refactor my code. I just wanna touch bases on the overall approach here. Do we concur on having the pull requester fix all the docstrings in the files he has modified? or should we limit it to the actual objects he has modified only. That way, instead of checking the files that have been changed i should rather check the lines and get solely the objects to whom these lines belong.

EmilyBourne: For me the ideal approach would be to have the pull requester fix any docstrings he has modified and any errors that are new. That way if his changes cause example code elsewhere to become outdated he has to fix that too.

Some of the files in pyccel are really big so fixing everything seems a bit mean. In addition, we are likely to have lots of people changing docstrings at the same time as some files are worked on by everyone often.

Let me know if you want to discuss how to implement this

ci_tools/summarise_doccoverage.py
EmilyBourne: See [here](https://github.com/pyccel/pyccel/blob/master/.github/workflows/pyccel_lint.yml) or [here](https://github.com/pyccel/pyccel/blob/master/.github/workflows/spelling.yml) for an example.

Tests are indeed marked as failing by exiting with an error code. But if you write to the file in the $GITHUB_STEP_SUMMARY variable then it appears as a markdown summary (this happens pass or fail). See for example https://github.com/pyccel/pyccel/actions/runs/3892223663

ci_tools/list_changed_files.py
jalalium: > For me the ideal approach would be to have the pull requester fix any docstrings he has modified and any errors that are new. That way if his changes cause example code elsewhere to become outdated he has to fix that too.
The `sphinx` extension `doctest` will take care of this. I will create an action for it once `sphinx doc` is ready. It will make it very easy to detect the kind of errors you mentioned.
What do you think about the [new](https://github.com/pyccel/pyccel/pull/1292/commits/5d12c239a8632cd6770cc9de2d3f68ae31a43607) approach here. it only checks the objects that were modified, not the whole file. However, if a method is modified, it will also include its class docstring. The module docstring is always included. Please let me know if I understood your comment correctly.

ci_tools/summarise_doccoverage.py
EmilyBourne: Can you point me at an example?

ci_tools/list_changed_files.py
```diff
@@ -13,7 +13,8 @@
...
     for file,changes in results.items():
-        for line in changes['addition']:
-            print(file, line, sep=' ', file=out)
-        for line in changes['deletion']:
-            print(file, line, sep=' ', file=out)
+        if file.startswith('pyccel/'):
+            for line in changes['addition']:
+                print(file, line, sep=' ', file=out)
+            for line in changes['deletion']:
+                print(file, line, sep=' ', file=out)
```
EmilyBourne, line 20: Are you sure you want to check removed lines? Won't that make it impossible to get rid of errors?

EmilyBourne, line 20: Why does this need to be a separate file. Is it not simpler to obtain `changes['addition']` directly in `list_docs_tovalidate.py`?

pyccel/ast/basic.py
```diff
@@ -384,7 +384,7 @@ def remove_user_node(self, user_node, invalidate = True):
         assert(user_node in self._user_nodes)
         self._user_nodes.remove(user_node)
         if self.is_unused and invalidate:
-            self.invalidate_node()
+            self.invalidate_node( )
```
EmilyBourne: Why this change?

ci_tools/list_changed_files.py
EmilyBourne: Ok we can do this later if you prefer :+1:

This seems ok for now but it doesn't really handle my comment. For example imagine that the interface of `Variable` or `Assign` is changed. The docstring of `FunctionDef` may now be wrong.

I was hoping that you could do something like:

- Get current problems in base branch as `base.json`
- Get current problems in new branch as `new.json`
- Report anything in `new.json` which isn't in `base.json`

Of course the exact solution will depend what the output of sphinx and co looks like. I don't know if there is a json option? Or something else that would be easy to manipulate?

pyccel/ast/basic.py
jalalium: It was only to test that changed are caught by the tool, this change is outdated

EmilyBourne: Ah sorry, I didn't check the full diff. Just the commit you pointed me at

ci_tools/list_changed_files.py
jalalium: You are absolutely right! i will merge the two files

jalalium: You are right! I need to think about this more carefully!

ci_tools/list_changed_files.py
jalalium: > This seems ok for now but it doesn't really handle my comment. For example imagine that the interface of Variable or >Assign is changed. The docstring of FunctionDef may now be wrong.

When you say that `FunctionDef` docstring may now be wrong. Do you mean the example/s present in it? or the actual description, parameters and return values? If it is the later, I do not know of a tool to check it.

ci_tools/summarise_doccoverage.py
jalalium: [here](https://github.com/pyccel/pyccel/actions/runs/3928835228) is one that fails, and [another](https://github.com/pyccel/pyccel/actions/runs/3997716493) that passes

ci_tools/list_changed_files.py
EmilyBourne: I mean the examples. I can't think of a reason why a change in the `Variable` interface would lead to a change in the arguments/results/definition of a `FunctionDef` without any code breaking

EmilyBourne: That's why I said we can do this later if you haven't yet added the code that checks the examples

ci_tools/summarise_doccoverage.py
EmilyBourne: Why does the failure start with "Success!"?

EmilyBourne: Otherwise it looks good

jalalium: The first part is a success, it did not add any new functions without a doctring. It is the second part that checks `numpydoc` validity that fails. I changed the second failure to be a header as well.

ci_tools/list_changed_files.py
jalalium: Okay great!

ci_tools/summarise_doccoverage.py
EmilyBourne: :+1:

yguclu:
@jalalium What is the status of this PR? I see two open conversations

ci_tools/summarise_doccoverage.py
yguclu: @EmilyBourne Has this conversation been resolved?

jalalium:
> @jalalium What is the status of this PR? I see two open conversations

I need to fix Emily's remarks. I will do it right now!

jalalium:
@yguclu Should be good now!

ci_tools/list_docs_tovalidate.py
```diff
@@ -0,0 +1,53 @@
...
+                    if any((node.lineno <= x <= node.end_lineno
+                            for x in line_nos)):
+                        objects.append('.'.join([prefix, node.name]))
+            if isinstance(node, ClassDef):
+                if any((node.lineno <= x <= node.end_lineno
+                        for x in line_nos)):
+                    objects.append('.'.join([prefix, node.name]))
+                for child in node.body:
+                    if isinstance(child, ast.FunctionDef):
+                        child.parent = node
```
EmilyBourne, line 50: To reduce code duplication and increase readability I suggest:
```suggestion
            if isinstance(node, (FunctionDef, ClassDef)):
                code_changed = any((node.lineno <= x <= node.end_lineno
                            for x in line_nos))
                if code_changed:
                    objects.append('.'.join([prefix, node.name]))
                    
                # Update member function names
                if isinstance(node, ClassDef):
                    for child in node.body:
                        if isinstance(child, ast.FunctionDef):
                            child.name = f'{node.name}.{child.name}'
```
(I have not tested this though so best to double check it works correctly)

EmilyBourne (changes requested):

ci_tools/list_docs_tovalidate.py
```diff
@@ -0,0 +1,53 @@
...
+parser.add_argument('gitdiff', metavar='gitdiff', type=str,
+                        help='Output of git diff between PR branch and base branch')
+parser.add_argument('output', metavar='output', type=str,
+                        help='File to save the output to')
+args = parser.parse_args()
+
+results = gdj(args.gitdiff)
+changes = {}
+for file, upds in results.items():
+    if file.startswith('pyccel/'):
```
EmilyBourne, line 18: It is not necessary here as we know that this file will be run on linux but it is good practice to use the [`pathlib`](https://docs.python.org/3/library/pathlib.html) module (or the [`os.path`](https://docs.python.org/3/library/os.path.html#module-os.path) module) to handle anything file related.
For example on Windows you would never enter this if as the file would start with `pyccel\\`

So here you could have:
```python
if pathlib.PurePath(file).parents[0] == 'pyccel':
```

and below you could have:
```python
prefix = '.'.join(pathlib.PurePath(file).with_suffix('').parts)
```

ci_tools/list_docs_tovalidate.py
```diff
@@ -0,0 +1,53 @@
...
+from git_evaluation_tools import get_diff_as_json as gdj
+
+parser = argparse.ArgumentParser(description='List the objects with docstrings in the files provided')
+parser.add_argument('gitdiff', metavar='gitdiff', type=str,
+                        help='Output of git diff between PR branch and base branch')
+parser.add_argument('output', metavar='output', type=str,
+                        help='File to save the output to')
+args = parser.parse_args()
+
+results = gdj(args.gitdiff)
```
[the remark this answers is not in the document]
jalalium: Got it!

ci_tools/list_docs_tovalidate.py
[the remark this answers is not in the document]
jalalium: You are right! Yes, I am collecting all objects whose contents have been changed. The "parent" attribute here is used to differentiate between functions and class methods. The class name of a method is needed to get the full name of the method object. This will not take care of inner classes, but I don't think there is a need for that, is there?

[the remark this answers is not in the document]
EmilyBourne: Not that I am aware of. I think we're safe for now. Although what about functions in functions?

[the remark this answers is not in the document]
jalalium: It is not handled either. I will refactor the code and mimic what you exposed in the following comment to take care of those cases as well. It is not hard to keep the absolute name of the object at all times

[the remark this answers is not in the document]
EmilyBourne: Ah no, it should be handled thanks to walk. It just won't have a helpful name

ci_tools/list_docs_tovalidate.py
EmilyBourne: obviously you can also do `from pathlib import PurePath` for shorter commands

EmilyBourne: Sorry I think this should be `parts` not `parents`

jalalium: True, I will fix it!

ci_tools/list_docs_tovalidate.py
```diff
@@ -48,11 +48,14 @@
...
-                    for x in line_nos)):
+                    for x in line_nos)) and node.name != 'inner_function':
                 objects.append('.'.join([prefix, node.name]))
             for child in node.body:
                 if isinstance(child, (FunctionDef, ClassDef)):
-                    child.name = '.'.join([node.name, child.name])
+                    if isinstance(node, ClassDef):
+                        child.name = '.'.join([node.name, child.name])
+                    else:
+                        child.name = 'inner_function'
```
EmilyBourne, line 58: I'm not a huge fan of this solution as it will run faster as:
```
if isinstance(node, ClassDef):
    for child in node.body:
           if isinstance(child, (FunctionDef, ClassDef)):
                child.name = '.'.join([node.name, child.name])
else:
    for child in node.body:
           if isinstance(child, (FunctionDef, ClassDef)):
                child.name = 'inner_function'
```
but then this feels like code duplication. Maybe:
```
if isinstance(node, (FunctionDef, ClassDef)):
            if any((node.lineno <= x <= node.end_lineno
                    for x in line_nos)) and not node.name.startswith('inner_function.'):
                objects.append('.'.join([prefix, node.name]))
            prefix = node.name if isinstance(node, ClassDef) else 'inner_function'
            for child in node.body:
                if isinstance(child, (FunctionDef, ClassDef)):
                    child.name = '.'.join([prefix, child.name])
```
which also avoids evaluating the property multiple times?

jalalium: > 
Perfect

EmilyBourne (approved):

pyccel/ast/core.py
```diff
@@ -4228,3 +4228,4 @@ def doit(expr, targets):
 
 #==============================================================================
 
+
```
yguclu, line 4231: Useless whitespace change

yguclu (changes requested):
Excellent job!

Only one small thing needs fixing...

pyccel/ast/core.py
jalalium: Thank you! I will fix it right away

yguclu: My point was that this file should not be modified at all by this PR. I've fixed it.

yguclu (approved):

PR #1299, opened 2023-01-09
Fix array slicing

### LINKED ISSUE #1288: C functions don't correctly handle arrays with strides
### Describe the bug
Passing a non-contiguous array to a function which has been translated to C gives unexpected results.

### To Reproduce
```python
import numpy as np

def f(a : 'float[:]'):
    a[:] = 2

if __name__ == '__main__':
    arr = np.ones((2,3,4))

    f(arr[0,:,3])
    print(arr)
```

### Error details
Python output:
```bash
[[[1. 1. 1. 2.]
  [1. 1. 1. 2.]
  [1. 1. 1. 2.]]

 [[1. 1. 1. 1.]
  [1. 1. 1. 1.]
  [1. 1. 1. 1.]]]
```
C output:
```bash
[[[1.000000000000 1.000000000000 1.000000000000 2.000000000000]
[1.000000000000 1.000000000000 1.000000000000 1.000000000000]
[1.000000000000 1.000000000000 1.000000000000 1.000000000000]]
[[1.000000000000 1.000000000000 1.000000000000 1.000000000000]
[1.000000000000 1.000000000000 1.000000000000 1.000000000000]
[1.000000000000 1.000000000000 1.000000000000 1.000000000000]]]
```

The generated code for the function is:
```c
void f(t_ndarray a)
{
    int64_t i;
    for (i = INT64_C(0); i < a.shape[INT64_C(0)]; i += INT64_C(1))
    {
        GET_ELEMENT(a, nd_double, (int64_t)i) = INT64_C(2);
    }
}
```

so either `GET_ELEMENT` is incorrectly implemented, or `array_slicing` is incorrectly implemented.

### Language
C

### Additional context
This problem may also exist for other code snippets using different structures

aihya:
Handle non-contiguous array views in C (fixes #1288). Update Pyccel version to 1.7.2.


bauom (changes requested):
I can't find any test that reproduce the test in the issue #1288

```Python
import numpy as np

def f(a : 'float[:]'):
    a[:] = 2

if __name__ == '__main__':
    arr = np.ones((2,3,4))

    f(arr[0,:,3])
    print(arr)
```
try passing views to the test function.

bauom:
> [83a93f5](https://github.com/pyccel/pyccel/commit/83a93f58cc984bfa006aedd2eab33d2377c2a28d)

@aihya 
tests are failing now

aihya:
> > [83a93f5](https://github.com/pyccel/pyccel/commit/83a93f58cc984bfa006aedd2eab33d2377c2a28d)
> 
> @aihya tests are failing now

It's weird how the C tests works fine but not the fortran ones. I think the failing tests are related to the issue https://github.com/pyccel/pyccel/pull/1256.

yguclu:
Hi @aihya, what's the status of this PR? Is it blocked by another issue?

CC: @EmilyBourne @bauom @jalalium

EmilyBourne:
> > > [83a93f5](https://github.com/pyccel/pyccel/commit/83a93f58cc984bfa006aedd2eab33d2377c2a28d)
> > 
> > @aihya tests are failing now
> 
> It's weird how the C tests works fine but not the fortran ones. I think the failing tests are related to the issue https://github.com/pyccel/pyccel/pull/1256.

I disagree. Issue #1241 concerns arrays being unnecessarily transposed in Fortran code. Here we have 1D arrays so there cannot be a transpose.

EmilyBourne:
I think the error is due to issue #1117. You should add parametrization to the tests to mark the Fortran tests with xfail or skip

EmilyBourne:
> I think the error is due to issue #1117. You should add parametrization to the tests to mark the Fortran tests with xfail or skip

@yguclu I have started work on a fix for this. Let me know if you need it quickly for the JOSS paper otherwise it will probably appear in a few weeks

yguclu:
> > I think the error is due to issue #1117. You should add parametrization to the tests to mark the Fortran tests with xfail or skip
> 
> @yguclu I have started work on a fix for this. Let me know if you need it quickly for the JOSS paper otherwise it will probably appear in a few weeks

Good question. I don't think that we need it for the updated tests, but I am not 100% sure. How complicated is the fix? If it takes less than a day then I would give it a high priority... :sweat_smile:

aihya:
> I think the error is due to issue #1117. You should add parametrization to the tests to mark the Fortran tests with xfail or skip



> Hi @aihya, what's the status of this PR? Is it blocked by another issue?
> 
> CC: @EmilyBourne @bauom @jalalium

The tests pass for now.

pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -257,21 +258,23 @@ t_ndarray array_slicing(t_ndarray arr, int n, ...)
     for (int32_t i = 0; i < arr.nd ; i++)
     {
         slice = va_arg(va, t_slice);
+        if (slice.end - slice.start > 1 && !offset)
+            offset = i;
```
EmilyBourne, line 262: Are the strides properly considered if the offset is provided?

EmilyBourne (commented):
Does this work for a view with strides?
`a[2,::3,5]`?

EmilyBourne (approved):

pyccel/stdlib/ndarrays/ndarrays.c
aihya: The test seems to work fine

aihya: https://github.com/pyccel/pyccel/pull/1299/commits/5cd16d8e7f6cae925b9446e20305fcdd89a43d29

pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -245,6 +245,7 @@ t_ndarray array_slicing(t_ndarray arr, int n, ...)
     va_list  va;
     t_slice slice;
     int32_t start = 0;
+    int32_t offset = 0;
```
yguclu, line 248: Can you please explain what `offset` represents?

aihya: `offset` is the position of the first slice with the step bigger than 1, it just gives an idea of what dimension from which the copying begins.

pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -238,8 +238,8 @@ t_slice new_slice(int32_t start, int32_t end, int32_t step)
     slice.step = step;
     return (slice);
 }
-
-t_ndarray array_slicing(t_ndarray arr, int n, ...)
+#include <stdio.h>
+t_ndarray array_slicing(t_ndarray arr, int n, int n_slices, ...)
```
yguclu, line 242: I do not understand why we need both `n` and `n_slices`.

I think that a bit of documentation is really needed here. Could you please add a comment block using Doxygen style?

yguclu:
I find it curious that different tests fail on different platforms:
   - `pyccel/test_pyccel.py::test_arrays_view[c]` fails on Linux and macOS 
   - `epyccel\test_higherorder_functions.py::test_euler[c]` fails on Windows

@aihya Do you have any clue of why it is so?

EmilyBourne:
Beware, the tests are not necessarily run in the same order and they exit once something fails so this may explain the different failing tests

yguclu:
> Beware, the tests are not necessarily run in the same order and they exit once something fails so this may explain the different failing tests

One may need to run `valgrind --tool=memcheck` to find the culprit here.

EmilyBourne:
> I find it curious that different tests fail on different platforms:
> 
>     * `pyccel/test_pyccel.py::test_arrays_view[c]` fails on Linux and macOS
> 
>     * ~~`epyccel\test_higherorder_functions.py::test_euler[c]`~~ `epyccel\test_higherorder_functions.py::test_real_3[c]` fails on Windows
> 
> 
> @aihya Do you have any clue of why it is so?

I can confirm that Windows didn't run the test `test_arrays_view[c]` and mac doesn't run `test_real_3[c]` so this goes some way to explaining why different tests fail

pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -239,45 +239,47 @@ t_slice new_slice(int32_t start, int32_t end, int32_t step)
...
-        free(view.strides);
-        view.strides = tmp_strides;
-        view.shape = tmp_shape;
     }
+    for (int32_t i = 0; i < n; i++)
+    {
+        view.shape[i] = (slice.end - slice.start + (slice.step - 1)) / slice.step;
+        view.strides[i] = arr.strides[offset + i] * slice.step;
+        slice = va_arg(va, t_slice);
+    }
```
EmilyBourne, line 280: I get the impression that this only handles consecutive removed dimensions. I suspect something like `[:,0,:,2,:]` would not work

pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -239,45 +239,47 @@ t_slice new_slice(int32_t start, int32_t end, int32_t step)
...
-t_ndarray array_slicing(t_ndarray arr, int n, ...)
+t_ndarray array_slicing(t_ndarray arr, int n, int n_slices, ...)
 {
     t_ndarray view;
     va_list  va;
     t_slice slice;
     int32_t start = 0;
+    int32_t offset = 0;
 
     view.nd = n;
```
EmilyBourne, line 250: I am not convinced you have the right number of dimensions

pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -239,45 +239,47 @@ t_slice new_slice(int32_t start, int32_t end, int32_t step)
     return (slice);
 }
 
-t_ndarray array_slicing(t_ndarray arr, int n, ...)
+t_ndarray array_slicing(t_ndarray arr, int n, int n_slices, ...)
```
EmilyBourne, line 242: Don't you always have `n_slices == arr.nd`?

tutorial/ndarrays.md
```diff
@@ -278,7 +278,7 @@ Some examples:
             a = array_create(2, (int64_t[]){2, 4}, nd_int64);
             int64_t array_dummy_0001[] = {1, 2, 3, 4, 5, 6, 7, 8};
             memcpy(a.nd_int64, array_dummy_0001, a.buffer_size);
-            b = array_slicing(a, 1, new_slice(1, 2, 1), new_slice(0, a.shape[1], 1));
+            b = array_slicing(a, 1, 2, new_slice(1, 2, 1), new_slice(0, a.shape[1], 1));
```
EmilyBourne, line 281: @nhamidn  How do we differentiate between `a[:,0:1,:,2]` and `a[:,0,:,1:2]`

EmilyBourne (changes requested):

pyccel/stdlib/ndarrays/ndarrays.c
aihya:
```python3
if __name__ == "__main__":
     x1 = np.ones((2, 3, 4, 5, 6), dtype=int)
     a = x1[:,0,:,2,:]
     print(1)
```
I tried this test and it worked fine. It gave the same result as python.

pyccel/stdlib/ndarrays/ndarrays.c
aihya: The only reason why I added `n_slices` is because there was a test written like this where there 1 as number of slices but two slices were given:
`b = array_slicing(a, 1, new_slice(1, 2, 1), new_slice(0, a.shape[1], 1));`

EmilyBourne: `n` is not the number of slices. I believe it is the number of dimensions in the view

aihya: Could it be possible the number of slices be bigger than the size of `arr`?

EmilyBourne: As there is no doc you will have to check that in the c printer or speak with the person that developed it (git blame says @nhamidn).

My understanding is that the number of slices is always equal to the number of dimensions in arr. Which may be more than `n` the number of dimensions in `view`

aihya:
@EmilyBourne @yguclu @bauom I think the new `array_slicing` might be a little bit general. What I do is I separate literals from slices in the function `new_slice` using a fourth parameter `slice_type`, which could be 1 for `[..., :, ...]` or 0 for `[..., 2, ...]`. The idea is that if we have a syntax like this `arr[:, 2, :, 0, :]`, the slices passed to `array_slicing` would contain a fourth parameter having the values `1, 0, 1, 0, 1` consecutively. So, I use that value instead of checking the range of the slice (`slice.end - slice.start`) to determine whether to include that dimension in the `view.shape` or not.

aihya:
```python3
if __name__ == "__main__":
     x1 = np.ones((2, 3, 4, 5, 6), dtype=int)
     a = x1[:, 2, :, 3, :]
     print(a.shape)
```
The code above used to give the shape `(4, 1, 6)` in the previous `array_slicing`, now it gives `(2, 4, 6)`, as not it considers only the `:` part, which is correct.

tests/epyccel/test_arrays.py
```diff
@@ -3474,6 +3474,85 @@ def test_iterate_slice(language):
...
+    assert np.array_equal(x1, x2)
+    f1(x1[::2, 3, 4], 5)
+    f2(x2[::2, 3, 4], 5)
+    assert np.array_equal(x1, x2)
+    f1(x1[2, ::3, 4], 6)
+    f2(x2[2, ::3, 4], 6)
+    assert np.array_equal(x1, x2)
+    f1(x1[2, 3, ::4], 7)
+    f2(x2[2, 3, ::4], 7)
+    assert np.array_equal(x1, x2)
```
yguclu, line 3554: Can you please add a check there two slices are used instead of just one?

tests/epyccel/modules/arrays.py
```diff
@@ -1828,6 +1828,14 @@ def iterate_slice(i : int):
         res += ai
     return res
 
+#==============================================================================
+# TESTS: VIEW
+#==============================================================================
+
+@types('int[:]', 'int')
+def array_1d_view_assign(x, a):
+    x[:] = a
```
yguclu, line 1837: Can you please add a function that works on 2D views?

yguclu (changes requested):
Good job @aihya, this looks very promising! I think we need more tests though...

pyccel/ast/internals.py
```diff
@@ -137,13 +137,17 @@ class Slice(Basic):
...
-    __slots__ = ('_start','_stop','_step')
-    _attribute_nodes = ('_start','_stop','_step')
+    __slots__ = ('_start','_stop','_step', '_slice_type')
+    _attribute_nodes = ('_start','_stop','_step', '_slice_type')
 
-    def __init__(self, start, stop, step = None):
+    Range = LiteralInteger(1)
+    Element = LiteralInteger(0)
+
+    def __init__(self, start, stop, step = None, slice_type = Element):
```
EmilyBourne, line 146: Why is the default slice type `Element`?

pyccel/ast/internals.py
```diff
@@ -173,6 +179,18 @@ def step(self):
...
+    def slice_type(self):
+        """ The type of the slice (Range or Element)
+        Range   <=> [:]
+        Element <=> [3]
+        """
+        return self._slice_type
+
+    @slice_type.setter
+    def slice_type(self, value):
+        self._slice_type = value
```
EmilyBourne, line 192: Where do you need the setter as well as the constructor argument?

pyccel/codegen/printing/ccode.py
```diff
@@ -1106,8 +1106,10 @@ def _print_IndexedElement(self, expr):
                     if isinstance(ind, Slice):
                         inds[i] = self._new_slice_with_processed_arguments(ind, PyccelArraySize(base, i),
                             allow_negative_indexes)
+                        inds[i].slice_type = Slice.Range
                     else:
                         inds[i] = Slice(ind, PyccelAdd(ind, LiteralInteger(1), simplify = True), LiteralInteger(1))
+                        inds[i].slice_type = Slice.Element
```
EmilyBourne, line 1112: Shouldn't the constructor argument be sufficient to fix the `slice_type`?

pyccel/codegen/printing/ccode.py
```diff
@@ -1106,8 +1106,10 @@ def _print_IndexedElement(self, expr):
                     if isinstance(ind, Slice):
                         inds[i] = self._new_slice_with_processed_arguments(ind, PyccelArraySize(base, i),
                             allow_negative_indexes)
+                        inds[i].slice_type = Slice.Range
```
EmilyBourne, line 1109: Shouldn't the `_new_slice_with_processed_arguments` function set the `slice_type` to avoid duplication?

pyccel/codegen/printing/ccode.py
```diff
@@ -1246,7 +1248,8 @@ def _print_Slice(self, expr):
         start = self._print(expr.start)
         stop = self._print(expr.stop)
         step = self._print(expr.step)
-        return 'new_slice({}, {}, {})'.format(start, stop, step)
+        slice_type = self._print(expr.slice_type)
+        return 'new_slice({}, {}, {}, {})'.format(start, stop, step, slice_type)
```
EmilyBourne, line 1252: Not sure why codacy doesn't care currently but this should probably be an f-string

EmilyBourne (changes requested):
This looks like a much cleaner and simpler implementation. Great job!

pyccel/ast/internals.py
aihya: I just had to initialize it with something. Although, I'm not sure if the naming is correct, I put `Range` to represent `[..., :, ...]`  and `Element` for `[..., 2, ...]` for example.

EmilyBourne: The name seems fine to me.

My concern is not the fact that there is an initialisation, but the fact that you initialise with `Element`. I believe `Slice` objects are also created in the semantic stage. With a default value you obviously don't see an error for this, but these slices are then all considered to be `Element`s. This seems wrong to me given that in python a slice is always a `slice_type = Range` and the `Slice`s created in the semantic stage are the equivalent of pure python slices.

tests/epyccel/test_arrays.py
```diff
@@ -3474,6 +3474,85 @@ def test_iterate_slice(language):
...
+    f2(x2[2, :, 3], 3)
+    assert np.array_equal(x1, x2)
+    f1(x1[:, 2, 4], 4)
+    f2(x2[:, 2, 4], 4)
+    assert np.array_equal(x1, x2)
+    f1(x1[::2, 3, 4], 5)
+    f2(x2[::2, 3, 4], 5)
+    assert np.array_equal(x1, x2)
+    f1(x1[2, ::3, 4], 6)
+    f2(x2[2, ::3, 4], 6)
```
EmilyBourne, line 3550: Could you please add a test with a slice of length 1

aihya: https://github.com/pyccel/pyccel/pull/1299/commits/8a6e2b3cd1ae93ebef9c0a56fcc55f1efcee9c5c

pyccel/codegen/printing/ccode.py
aihya: https://github.com/pyccel/pyccel/pull/1299/commits/ee8cf11658638a9af0b93edc9f4abc71323b9f89

pyccel/codegen/printing/ccode.py
aihya: https://github.com/pyccel/pyccel/pull/1299/commits/ee8cf11658638a9af0b93edc9f4abc71323b9f89

pyccel/codegen/printing/ccode.py
aihya: https://github.com/pyccel/pyccel/pull/1299/commits/6331896f3ba794243b8f89d73913440e41f7b0f2

pyccel/ast/internals.py
aihya: https://github.com/pyccel/pyccel/pull/1299/commits/6331896f3ba794243b8f89d73913440e41f7b0f2

tests/epyccel/modules/arrays.py
aihya: https://github.com/pyccel/pyccel/pull/1299/commits/8a6e2b3cd1ae93ebef9c0a56fcc55f1efcee9c5c

tests/epyccel/test_arrays.py
aihya: https://github.com/pyccel/pyccel/pull/1299/commits/8a6e2b3cd1ae93ebef9c0a56fcc55f1efcee9c5c

pyccel/ast/internals.py
```diff
@@ -173,6 +179,14 @@ def step(self):
         """
         return self._step
 
+    @property
+    def slice_type(self):
+        """ The type of the slice (Range or Index)
+        Range <=> [..., :, ...]
+        Index <=> [..., 3, ...]
+        """
+        return self._slice_type
```
yguclu, line 188: On line 144 you call it `Element`, not `Index`

pyccel/stdlib/ndarrays/ndarrays.h
```diff
@@ -21,11 +21,15 @@
                                         float complex : _array_fill_cfloat,\
                                         double complex : _array_fill_cdouble)(c, arr)
 
+#define RANGE 1
+#define ELEMENT 0
```
yguclu, line 25: Is there a reason for not using an enumeration?

yguclu (changes requested):

pyccel/stdlib/ndarrays/ndarrays.c
yguclu: @EmilyBourne Can we close this conversation?

tutorial/ndarrays.md
yguclu: @EmilyBourne Can we close this conversation?

pyccel/stdlib/ndarrays/ndarrays.c
yguclu: @EmilyBourne Can we close this conversation?

pyccel/stdlib/ndarrays/ndarrays.c
yguclu: Good job for avoiding the double allocation!

yguclu:
@aihya Please update the Pyccel version to `1.7.2` in [/pyccel/version.py](https://github.com/pyccel/pyccel/blob/fix_ndarrays_strides/pyccel/version.py). We'll make a new release once this PR is merged.

pyccel/ast/internals.py
```diff
@@ -153,6 +157,8 @@ def __init__(self, start, stop, step = None):
             raise TypeError('Slice stop must be Integer or None')
         if step is not None and not (hasattr(step, 'dtype') and isinstance(step.dtype, NativeInteger)):
             raise TypeError('Slice step must be Integer or None')
+        if slice_type is not None and not (hasattr(slice_type, 'dtype') and isinstance(slice_type.dtype, NativeInteger)):
```
EmilyBourne, line 160: What happens if `slice_type is None`?
The second half can be written more compactly as
`isinstance(getattr(slice_type, 'dtype', None), NativeInteger))`

But should this not be:
`if slice_type not in (Range, Element):`
?

EmilyBourne (changes requested):

pyccel/ast/internals.py
aihya: https://github.com/pyccel/pyccel/pull/1299/commits/7a51001eff28302d544896356e850fa6f21bf8e9

pyccel/ast/internals.py
EmilyBourne: `self.Range, self.Element` to pass

yguclu (approved):
Great job!

EmilyBourne (approved):

Issue #1293, opened 2023-01-06
[CUDA] Cuda tests do not run


EmilyBourne:
### Describe the bug
Cuda tests have only been running when merging to `cuda_main` (see [here](https://github.com/pyccel/pyccel/blob/cuda_main_temp/.github/workflows/Cuda_pytest.yml#L5)), however cuda branches merge to `cuda_main_temp`. As a result the tests haven't been run for 2 months and breaking changes have been merged to `cuda_main_temp`. The tests now fail.
Tests have been reactivated, they now need to be fixed.

### Error details
See https://github.com/pyccel/pyccel/actions/runs/3856177210/jobs/6572072053

### Language
Cuda

EmilyBourne:
@bauom can this issue be closed? Or should we wait until the cuda pipeline has been restored?

bauom:
> @bauom can this issue be closed? Or should we wait until the cuda pipeline has been restored?

as far as i know this problem is now is fixed No?
we did activate the tests in all the Cuda branches.
and @framdani is adding run time tests in #1328
so this problem should be fixed for good.
we might need to add a local check in the future until we have some way to run the `cuda` tests remotely.

EmilyBourne:
@bauom Ok then I am closing the issue. I don't think all the tests are active because it is still not set up to merge to `cuda_devel` yet. But that is probably a separate issue. Once that pipeline is set up it should force the downstream branches to update I think.

PR #1300, opened 2023-01-10
Fix type inference for Cupy array functions

### LINKED ISSUE #1273: Cupy arrange doesn't have memory_location
When trying to pyccelize the following code :
```python
from pyccel.decorators import kernel, types
from pyccel import cuda
import cupy as cp

@kernel
@types('int[:]')
def func(a):
    i = cuda.threadIdx(0) + cuda.blockIdx(0) * cuda.blockDim(0)
    print("Hello World! ", a[i])

if __name__ == '__main__':
    threads_per_block = 32
    n_blocks = 1
    arr = cp.arange(32)
    cuda.deviceSynchronize()
    func[n_blocks, threads_per_block](arr)
    cuda.deviceSynchronize()
```
we get the following message:
```shell
  File "/pyccel/parser/semantic.py", line 1540, in _visit
  File "/pyccel/parser/semantic.py", line 1761, in _visit_CodeBlock
    line = self._visit(b, **settings)
  File "/pyccel/parser/semantic.py", line 1540, in _visit
  File "/parser/semantic.py", line 2544, in _visit_Assign
    d_var  = self._infere_type(rhs, **settings)
  File "/pyccel/parser/semantic.py", line 496, in _infere_type
    d_var['memory_location'] = expr.memory_location
AttributeError: 'CupyArange' object has no attribute 'memory_location'
```

Pinkyboi:
Fixes #1273.

### DIFF
--- a/pyccel/ast/cupyext.py
+++ b/pyccel/ast/cupyext.py
@@ -63,9 +63,8 @@ class CupyNewArray(CudaNewArray):
     """ Class from which all Cupy functions which imply a call to Allocate
     inherit
     """
-    __slots__ = ('_memory_location')
+    _memory_location = 'device'
     def __init__(self):
-        self._memory_location = 'device'
         super().__init__()
 
     @property
--- a/pyccel/ast/cupyext.py
+++ b/pyccel/ast/cupyext.py
@@ -36,7 +36,7 @@ from .literals       import Nil
 from .mathext        import MathCeil
 from .operators      import broadcast, PyccelMinus, PyccelDiv
 from .variable       import (Variable, Constant, HomogeneousTupleVariable)
-from .cudaext        import CupyNewArray, CudaArray
+from .cudaext        import CudaNewArray, CudaArray
 from .numpyext       import process_dtype, process_shape, DtypePrecisionToCastFunction
 
 errors = Errors()
@@ -63,7 +63,9 @@ class CupyNewArray(CudaNewArray):
     """ Class from which all Cupy functions which imply a call to Allocate
     inherit
     """
+    __slots__ = ('_arg','_dtype','_precision','_shape','_rank','_order', '_memory_location')
     def __init__(self):
+        self._memory_location = 'device'
         super().__init__()
 
 #==============================================================================
@@ -122,7 +124,6 @@ class CupyArray(CupyNewArray):
         self._dtype = dtype
         self._order = order
         self._precision = prec
-        self._memory_location = 'device'
         super().__init__()
 
     def __str__(self):
@@ -176,7 +177,6 @@ class CupyArange(CupyNewArray):
 
         self._shape = (MathCeil(PyccelDiv(PyccelMinus(self._stop, self._start), self._step)))
         self._shape = process_shape(False, self._shape)
-        self._memory_location = 'device'
         super().__init__()
 
     @property
@@ -255,7 +255,7 @@ class CupyFull(CupyNewArray):
         self._dtype = dtype
         self._order = CupyNewArray._process_order(self._rank, order)
         self._precision = precision
-        self._memory_location = 'device'
+
         super().__init__(fill_value)
 
     #--------------------------------------------------------------------------
--- a/pyccel/ast/cupyext.py
+++ b/pyccel/ast/cupyext.py
@@ -68,6 +68,12 @@ class CupyNewArray(CudaNewArray):
         self._memory_location = 'device'
         super().__init__()
 
+    @property
+    def memory_location(self):
+        """ Indicates whether a Variable has a dynamic size
+        """
+        return self._memory_location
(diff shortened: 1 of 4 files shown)


EmilyBourne:
Don't you also need the property:
```python
@property
def memory_location(self):
    return self._memory_location
```
in `CudaNewArray`?

EmilyBourne:
This should fix #1293 once #1279 is merged

pyccel/ast/cupyext.py
```diff
@@ -142,7 +158,6 @@ class CupyArange(CudaNewArray):
         The type of the output array, if dtype is not given,
         infer the data type from the other input arguments.
     """
-    __slots__ = ('_start','_step','_stop','_dtype','_precision','_shape')
     _attribute_nodes = ('_start','_step','_stop')
     _rank = 1
     _order = None
```
EmilyBourne, line 163: Having `_rank` and `_order` as class properties is good practice, however they cannot also be in `CupyNewArray`s slots

pyccel/ast/cupyext.py
```diff
@@ -157,7 +163,7 @@ class CupyArange(CupyNewArray):
...
     name = 'arange'
 
-    def __init__(self, start, stop = None, step = None, dtype = None, memory_location = 'device'):
+    def __init__(self, start, stop = None, step = None, dtype = None):
```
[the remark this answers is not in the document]
Pinkyboi: Because for the case of `Cupy` array the memory location will always be `device`

pyccel/ast/cupyext.py
```diff
@@ -58,14 +59,29 @@
 )
 
 #==============================================================================
-class CupyArray(CudaNewArray):
+class CupyNewArray(CudaNewArray):
+    """ Class from which all Cupy functions which imply a call to Allocate
+    inherit
+    """
+    __slots__ = ('_arg','_dtype','_precision','_shape','_rank','_order', '_memory_location')
```
EmilyBourne, line 66: I don't think this class should define all these slots. Why does a `NewArray` function only have 1 argument? Why can't any of these be class objects in sub classes?

Pinkyboi: Fixed in [07b3260](https://github.com/pyccel/pyccel/pull/1300/commits/07b3260ba94aa741a57d5dee713ad0fd1b7b3039).

pyccel/ast/cupyext.py
Pinkyboi: Fixed in [07b3260](https://github.com/pyccel/pyccel/pull/1300/commits/07b3260ba94aa741a57d5dee713ad0fd1b7b3039).

tests/internal/scripts/ccuda/cupy_arange.py
```diff
@@ -12,6 +12,6 @@ def func(a):
     threads_per_block = 32
     n_blocks = 1
     arr = cp.arange(32)
-    cuda.deviceSynchronize()
+    cuda.synchronize()
     func[n_blocks, threads_per_block](arr)
-    cuda.deviceSynchronize()
\ No newline at end of file
+    cuda.synchronize()
```
bauom: I think this test should only call `cupy.arange` so it will only fails when needed we don't need a kernel call or the use of `cuda.threadIdx(0)` and decorators in this test

bauom (changes requested):

tests/internal/scripts/ccuda/cupy_arange.py
Pinkyboi: Done in [bd6faa6](https://github.com/pyccel/pyccel/pull/1300/commits/bd6faa6f343762e484a7cea6ce6098f5a7ca5ff5).

bauom (approved):
everything seems fine to me now.

pyccel/ast/cupyext.py
```diff
@@ -63,14 +63,14 @@ class CupyNewArray(CudaNewArray):
     """ Class from which all Cupy functions which imply a call to Allocate
     inherit
     """
-    __slots__ = ('_arg','_dtype','_precision','_shape','_rank','_order', '_memory_location')
+    __slots__ = ('_memory_location')
     def __init__(self):
         self._memory_location = 'device'
```
EmilyBourne, line 68: Can `_memory_location` ever by anything else? If not it could be a class attribute

Pinkyboi: Yeah seems about right it can be an attribute.

Pinkyboi: Done in [17180eb](https://github.com/pyccel/pyccel/pull/1300/commits/17180ebcb81a2cce22594b086ea3abcd588e45c9).

EmilyBourne (approved):

PR #1301, opened 2023-01-10
cleaning compile stage Cuda tests.

bauom:
Changing the tests to remove unnecessary code to make the tests more indicating on the test file name.

- striping tests from unnecessary code to test the intended behavior.


Pinkyboi (approved):
Although the issues that you did avoids are the ones that will be fixed by this PR #1279 it seems all good to me.

bauom:
@EmilyBourne I think This should be merged before #1300 as it should fix #1293 too.
The tests that were failing were not related to the tests themselves, and that specific test that failed all of them can be tested in #1279

bauom:
> Although the issues that did avoids are the ones that will be fixed by this PR #1279 it seems all good to me.

I am aware of that but I think it should be better to separate the tests so they can be more specific and only fails on the intended tests.

EmilyBourne:
I don't see the point in this PR. There are already 2 open PRs which fix the failing tests and the tests are still failing on this PR

It is really hard to have a test that only fails because of one thing. In pyccel I don't think it's possible.
Have you checked to make sure that removing these tests doesn't reduce coverage of something that should be tested?
If your aim is to separate the tests then I would expect this PR to replace tests not remove them entirely.

I also don't understand why these failing tests are considered to be for things that don't need tests. If they haven't been fully implemented then they shouldn't have been merged. As they are merged it is much better to be aware of the problem.

If your aim is just to get the coverage running asap it would be better to mark these as xfail then we can keep track of the problem

bauom:
> I don't see the point in this PR. There are already 2 open PRs which fix the failing tests and the tests are still failing on this PR
> 
> It is really hard to have a test that only fails because of one thing. In pyccel I don't think it's possible. Have you checked to make sure that removing these tests doesn't reduce coverage of something that should be tested? If your aim is to separate the tests then I would expect this PR to replace tests not remove them entirely.
> 
> I also don't understand why these failing tests are considered to be for things that don't need tests. If they haven't been fully implemented then they shouldn't have been merged. As they are merged it is much better to be aware of the problem.
> 
> If your aim is just to get the coverage running asap it would be better to mark these as xfail then we can keep track of the problem

I did not remove the tests but just for example a test with the name of `cuda_array` do not need to launch a kernel and the tests that we had they all tried to launch a kernel.

I just matched the file names on what the test contains.

EmilyBourne:
> I did not remove the tests but just for example a test with the name of cuda_array do not need to launch a kernel and the tests that we had they all tried to launch a kernel.

I disagree. There are things that were being tested but aren't anymore. So tests have been removed.
I think I understand what you mean though. You want to test just the array creation, but these tests are all doing much more than that.

EmilyBourne:
I don't think your PR description is accurate though

bauom:
> I don't think your PR description is accurate though
I will Try to make the description cleaner.
are we on the same page that the tests should not test more than what is desired from them for example `cuda_array_managed` should have a call to `cuda.array` in the minimal way possible.
Or do you think this is not helpful.

EmilyBourne:
> > I don't think your PR description is accurate though
>
> I will Try to make the description cleaner.
> are we on the same page that the tests should not test more than what is desired from them for example `cuda_array_managed` should have a call to `cuda.array` in the minimal way possible.
> Or do you think this is not helpful.

Ideally tests should build upon each other so it is possible to work out what's failing.
Given you don't have a wrapper, array creation is reasonably fundamental so I agree that these tests should contain just the array creation

EmilyBourne (approved):

EmilyBourne:
> @EmilyBourne I think This should be merged before #1300 as it should fix #1293 too. The tests that were failing were not related to the tests themselves, and that specific test that failed all of them can be tested in #1279

I agree. This should be merged before #1279. Please update and fix other failing tests first though (I have activated the new tests for `cuda_main_temp`)

EmilyBourne (approved):

PR #1307, opened 2023-01-12
sync cuda_devel with cuda_main

bauom:


EmilyBourne (approved):

bauom:
@EmilyBourne does the codacy errors shows as i can't see what's the problem.

EmilyBourne:
@bauom The problems were caused by line endings (see https://github.com/pyccel/pyccel/pull/1312). However now it seems that Codacy is just reporting an error because there was one at some point. I am in contact with them and have an open bug report. I asked them if they want to have a look and find the problem, otherwise I suspect I can fix it for you with some rebases or by closing this PR and opening a new one

EmilyBourne:
> @bauom The problems were caused by line endings (see #1312). However now it seems that Codacy is just reporting an error because there was one at some point. I am in contact with them and have an open bug report. I asked them if they want to have a look and find the problem, otherwise I suspect I can fix it for you with some rebases or by closing this PR and opening a new one

I tried fixing with some rebases but it seems this is not sufficient. I suspect the presence of wrong line endings in the diff (the endings that are removed) is still causing problems

EmilyBourne:
@bauom codacy have now fixed the problem. You should be able to merge

PR #1308, opened 2023-01-16
Fix compilation without static library

### LINKED ISSUE #1184: Import Error when no static library is available
**Describe the bug**
Pyccel tries to link the python library using its exact path rather than providing a `-L` argument. This is done to avoid accidentally importing other python libraries (e.g. hdf5) in the future. To help with this pyccel prefers the static library. However on systems where there is no static library available, the compiler succeeds (due to the exact path), but the library cannot be found at runtime leading to the following error:
```
E   ImportError: libpython3.9.so.1.0: cannot open shared object file: No such file or directory
```
cc: @dominikbell @FrederikSchnack

EmilyBourne:
Ensure library directories are provided when a shared library must be used. Fixes #1184. Update Pyccel patch version to 1.7.1


yguclu (approved):

yguclu:
Since this PR updates the patch version I would like to wait for a couple of other PRs before merging it.

EmilyBourne:
> Since this PR updates the patch version I would like to wait for a couple of other PRs before merging it.

I agree. I just thought something was missing from the PR page as it doesn't pass everything and there was no flag. So I added one so I don't keep forgetting and checking :sweat_smile:

Issue #1126, opened 2022-06-08
Can't pass an unnamed array to a function in C


EmilyBourne:
**Describe the bug**
When passing an array created on the fly to a function, the code will not translate to C.

**To Reproduce**
Provide code to reproduce the behavior:
```python
import numpy as np

def f(a : 'int[:]'):
    print(a)

if __name__ == '__main__':
    f(np.array([1,2,3]))
```

Provide the generated code, or the error message:
```bash/fortran/c
ERROR at code generation stage
pyccel:
 |fatal [codegen]: tmp.py [7,6]| _print_NumpyArray is not yet implemented for language : C
Pyccel has encountered syntax that has not been implemented yet. Please create an issue at https://github.com/pyccel/pyccel/issues and provide a small example of your problem. Do not forget to specify your target language ((1, 2, 3))
```

**Expected behavior**
The C printer should create a temporary variable in which the array can be stored before calling the function

**Language**
C

**Additional information**
The above code translates to Fortran correctly. The code below translates to both Fortran and C:
```python
import numpy as np

def f(a : 'int[:]'):
    print(a)

if __name__ == '__main__':
    x = np.array([1,2,3])
    f(x)
```

Pinkyboi:
@azouiten are you still planning to work on this issue ?

azouiten:
> @azouiten are you still planning to work on this issue ?
> 

Yes, I apologise for the delay.

PR #1309, opened 2023-01-20
Add support for numpy.sum

jalalium:
This PR adds support for `np.sum`


EmilyBourne:
There should be some existing tests for this in the pyccel/epyccel folder that can be reactivated by removing the `xfail`/`skip` mark

EmilyBourne:
https://github.com/pyccel/pyccel/blob/85cfff336bb408fa34a41985ff24884493b6c193/tests/epyccel/test_arrays.py#L228-L268

tests/epyccel/test_arrays.py
```diff
@@ -226,9 +226,7 @@ def test_array_int32_1d_sub_augassign(language):
     assert np.array_equal( x1, x2 )
 
 @pytest.mark.parametrize( 'language', [
-        pytest.param("c", marks = [
-            pytest.mark.skip(reason="Numpy sum not yet implemented for C language"),
-            pytest.mark.c]),
+        pytest.param("c", marks = pytest.mark.c),
         pytest.param("fortran", marks = pytest.mark.fortran)
     ]
 )
```
EmilyBourne, line 232: This mark should not be needed anymore. You can rely on the feature to provide the `language` argument

EmilyBourne:
> Please remove the parametrization here:

I suspect there are other locations in this folder where tests can be activated. Please search for them

tests/epyccel/test_arrays.py
jalalium: Fixed! thank you!

tests/epyccel/test_arrays.py
```diff
@@ -3516,322 +2774,160 @@ def test_array_2d_F_slice_stride_23(language):
...
 def test_array_2d_C_slice_stride_15(language):
     a = arrays.a_2d_c
 
     f1 = arrays.array_2d_C_slice_stride_15
     f2 = epyccel(f1, language = language)
     assert np.array_equal(f1(a), f2(a))
 
-@pytest.mark.parametrize( 'language', [
-        pytest.param("c", marks = [
-            pytest.mark.skip(reason="sum and len not implemented in c"),
```
EmilyBourne, line 3753: Is `len` implemented in C now? (If it isn't I think it's a 1or 2 line fix so probably worth adding here)

jalalium: I think it is impelemented! i tested it and it worked! I activated these tests too.

jalalium:
Some tests fail not because of sum but because slices and strides are not working as intended.

EmilyBourne:
> Some tests fail not because of sum but because slices and strides are not working as intended.

I think there is already an issue about this. Could you put the xfail back for these cases and correct the reason for the error (ideally referencing the issue)

tests/epyccel/test_arrays.py
EmilyBourne: Great! Apparently I did that ^^'. I obviously forgot to update the skip reasons.

jalalium:
> > Some tests fail not because of sum but because slices and strides are not working as intended.
> 
> I think there is already an issue about this. Could you put the xfail back for these cases and correct the reason for the error (ideally referencing the issue)

I changed the reason for the error for all tests.
There is another issue relative to negative steps as well!

pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -373,3 +373,240 @@ int64_t     *numpy_to_ndarray_shape(int64_t *np_shape, int nd)
...
+{
+    int64_t elements = 1;
+    for (int i = 0; i <arr.nd; i++)
+        elements *= arr.shape[i];
+    int64_t *tmp = (int64_t*)malloc(sizeof(int64_t) * arr.nd);
+    memset(tmp, 0, sizeof(tmp));
+    int64_t output = 0;
+    for (int32_t i = 0; i < elements; i++)
+    {
+        output += arr.nd_int64[get_index_from_array(arr, tmp)];
```
jalalium, line 509: is there a cleaner way of looping over the elements of an `nd_array`?

jalalium:
In the benchmarks, the `laplace 2d` test uses `np.sum(np.abs(p)`. However, this is not supported by Pyccel. Numpy unary functions do not work as arguments for function calls. Is there an issue for that? Is it necessary to fix that as well for the benchmarks?

pyccel/stdlib/ndarrays/ndarrays.c
EmilyBourne: Can't we just do:
```c
int64_t numpy_sum_int64(t_ndarray arr)
{
    int64_t sum = 0;
    for(int32_t idx(0); idx<arr.length; ++idx)
    {
        sum += arr.nd_int64[idx];
    }
    return sum;
}
```
?

EmilyBourne:
> In the benchmarks, the `laplace 2d` test uses `np.sum(np.abs(p)`. However, this is not supported by Pyccel. Numpy unary functions do not work as arguments for function calls. Is there an issue for that? Is it necessary to fix that as well for the benchmarks?

I think there is an issue for this to auto-create a temporary, or to raise an error so there is no hidden allocation. I don't remember which

pyccel/stdlib/ndarrays/ndarrays.c
jalalium: No, this only works for arrays that are not views.

EmilyBourne: Ah of course. I'm completely forgetting about strides ^^

pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -373,3 +373,240 @@ int64_t     *numpy_to_ndarray_shape(int64_t *np_shape, int nd)
...
+        case 7:
+            return GET_INDEX(arr, tmp[0], tmp[1], tmp[2], tmp[3], tmp[4], tmp[5], tmp[6]);
+        case 8:
+            return GET_INDEX(arr, tmp[0], tmp[1], tmp[2], tmp[3], tmp[4], tmp[5], tmp[6], tmp[7]);
+        case 9:
+            return GET_INDEX(arr, tmp[0], tmp[1], tmp[2], tmp[3], tmp[4], tmp[5], tmp[6], tmp[7], tmp[8]);
+        case 10:
+            return GET_INDEX(arr, tmp[0], tmp[1], tmp[2], tmp[3], tmp[4], tmp[5], tmp[6], tmp[7], tmp[8], tmp[9]);
+    }
+}
```
EmilyBourne, line 405: This raises a warning:
```
  /Users/runner/hostedtoolcache/Python/3.10.9/x64/lib/python3.10/site-packages/pyccel/codegen/compiling/compilers.py:431: UserWarning: /Users/runner/work/pyccel/pyccel/tests/internal/scripts/openmp/__pyccel__/ndarrays/ndarrays.c:405:1: warning: non-void function does not return a value in all control paths [-Wreturn-type]
```

As per this:
https://github.com/pyccel/pyccel/blob/f47756a0eb0c9b6da96037ff077df0e06e84b9c3/pyccel/stdlib/ndarrays/ndarrays.h#L31-L45
Pyccel can handle up to 15 dimensions. The default should raise an error before so you could use something like:
```
default:
    return -1;
```

That said, I don't know if macro+switch is more efficient than just using a for loop to calculate the index.

jalalium: Sorry, I will add the default return and i will go up to 15 dimensions. The problem with using for loops is that we only know the number of for loops at runtime. How can we generate such a loop dynamically in C?

EmilyBourne: I don't think we need multiple loops here:
```
static int64_t     get_index_from_array(t_ndarray arr, int64_t *tmp)
{
    int64_t idx = 0;
    for (int dim(0); dim<arr.nd; ++dim) {
        idx += arr.strides[dim] * (tmp[dim]);
    }
    return idx;
}
```

jalalium: That's much better! thank you!

jalalium:
@EmilyBourne Tests fail in MacOS because of negative indices and not because of `np.sum`.
For example, `test_array_2d_C_slice_19` fails because the value of `b[-1][-1]` is wrong.
What should I do about it?

pyccel/stdlib/ndarrays/ndarrays.c
EmilyBourne: Is `elements` equivalent to `arr.length`?

EmilyBourne:
> @EmilyBourne Tests fail in MacOS because of negative indices and not because of `np.sum`. For example, `test_array_2d_C_slice_19` fails because the value of `b[-1][-1]` is wrong. What should I do about it?

It is worrying that this doesn't fail in linux. Open an issue explaining the bug and add an xfail/skip mark quoting the issue.

When adding things to Pyccel, if you discover a bug you should always open an issue. If the bug is not caused by your PR then there is no reason to fix it in your PR. This is what the xfail/skip marks are for.

jalalium:
> > @EmilyBourne Tests fail in MacOS because of negative indices and not because of `np.sum`. For example, `test_array_2d_C_slice_19` fails because the value of `b[-1][-1]` is wrong. What should I do about it?
> 
> It is worrying that this doesn't fail in linux. Open an issue explaining the bug and add an xfail/skip mark quoting the issue.
> 
> When adding things to Pyccel, if you discover a bug you should always open an issue. If the bug is not caused by your PR then there is no reason to fix it in your PR. This is what the xfail/skip marks are for.

Noted! Thank you!

pyccel/stdlib/ndarrays/ndarrays.c
yguclu: @jalalium I have the same question

EmilyBourne:
Is it possible that your failing tests are actually caused by issue #1288 ? There are tests elsewhere which successfully use slices (see here : https://github.com/pyccel/pyccel/blob/master/tests/pyccel/scripts/arrays_view.py ) so I am not certain that your examples are failing for the reasons you think

pyccel/stdlib/ndarrays/ndarrays.c
EmilyBourne: There is a recursive option, but I'm not sure that it is faster or cleaner:
```c
int64_t     numpy_sum_int8(t_ndarray arr, int nd = -1, int offset = 0)
{
    if (nd == -1) {
        nd = arr.nd
    }
    int64_t stride = arr.strides[arr.nd-nd];
    int64_t shape = arr.shape[arr.nd-nd];
    int64_t output = 0;
    if (nd == 1) {
        for (int32_t j = 0; j < shape; j++)
        {
            ouput += arr.nd_int8[offset + j*stride];
        }
    }
    else {
        for (int32_t j = 0; j < shape; j++)
        {
            ouput += numpy_sum_int8(arr, nd-1, offset + j*stride);
        }
    }
    return output;
}
```

EmilyBourne: Assuming the strides contain what they are supposed to (I'm not sure about this, this may be the cause of #1288). What about:

```c
int64_t     numpy_sum_int8(t_ndarray arr)
{
    int64_t local_stride[arr.nd];
    local_stride[arr.nd-1] = 1;
    for (int32_t i = arr.nd-2; i >= 0; --i) {
        local_stride[i] = local_stride[i+1]*arr.shape[i+1];
    }
    int64_t output = 0;
    for (int32_t view_index = 0; view_index < arr.length; view_index++)
    {
        int32_t i = view_index;
        int32_t array_index = 0;
        for (int32_t dim = 0; dim < arr.nd; dim++)
        {
            array_index += (i_copy / local_stride[dim]) * arr.stride[dim];
            i -= local_stride[dim] * index[dim];
        }
        output += arr.nd_int8[array_index];
    }
    return output;
}
```

EmilyBourne:
@jalalium 
> > In the benchmarks, the `laplace 2d` test uses `np.sum(np.abs(p)`. However, this is not supported by Pyccel. Numpy unary functions do not work as arguments for function calls. Is there an issue for that? Is it necessary to fix that as well for the benchmarks?
> 
> I think there is an issue for this to auto-create a temporary, or to raise an error so there is no hidden allocation. I don't remember which

The issue is #1126

EmilyBourne:
@jalalium Don't forget to add the extra tests to satisfy the coverage checker

jalalium:
> Is it possible that your failing tests are actually caused by issue #1288 ? There are tests elsewhere which successfully use slices (see here : https://github.com/pyccel/pyccel/blob/master/tests/pyccel/scripts/arrays_view.py ) so I am not certain that your examples are failing for the reasons you think

Now all tests pass, there was in fact a problem with my implementation. But you are correct! #1288 is relevant to this PR as well, however, there are no activated tests that check for the issue.

EmilyBourne:
> > Is it possible that your failing tests are actually caused by issue #1288 ? There are tests elsewhere which successfully use slices (see here : https://github.com/pyccel/pyccel/blob/master/tests/pyccel/scripts/arrays_view.py ) so I am not certain that your examples are failing for the reasons you think
> 
> Now all tests pass, there was in fact a problem with my implementation. But you are correct! #1288 is relevant to this PR as well, however, there are no activated tests that check for the issue. 

By "failing tests" I meant the ones that you flagged as being related to #1310

jalalium:
> 

Those fail because `@allow_negative_index('a')` works for getting a single element but not for slices. All the tests that i flagged fail solely because the slice contains a variable. if one were to change the variable for its actual value, it would work!

pyccel/stdlib/ndarrays/ndarrays.c
jalalium: This is exactly what I am doing your `local_stride` array is my `tmp` array, I should declare it on the stack as you did. And you start with the array full and subtract from the indices while i do the opposite. I also tested the recursive method but it gave me a wrong output. I was not sure that the number of elements in a view is equivalent to `arr.length` so i will just use it and refactor the code, document it and clean it up

jalalium:
> @jalalium
> 
> > > In the benchmarks, the `laplace 2d` test uses `np.sum(np.abs(p)`. However, this is not supported by Pyccel. Numpy unary functions do not work as arguments for function calls. Is there an issue for that? Is it necessary to fix that as well for the benchmarks?
> > 
> > 
> > I think there is an issue for this to auto-create a temporary, or to raise an error so there is no hidden allocation. I don't remember which
> 
> The issue is #1126 . It is not exactly the same thing, but it is very relevant

I totally agree, both cases fall under the umbrella of function compositions with functions that need to create temporary arrays.

jalalium:
> @jalalium Don't forget to add the extra tests to satisfy the coverage checker

How do you link untested lines with their respective test?

pyccel/stdlib/ndarrays/ndarrays.c
EmilyBourne: I don't think so. Your `tmp` is a list of indexes which is why you can pass it to `get_index_from_array`. My solution avoids the `if` which may help

> And you start with the array full and subtract from the indices while i do the opposite. 

Not sure what you mean by "start with the array full and subtract from the indices". I iterate in order.

> I also tested the recursive method but it gave me a wrong output.

I didn't do any tests for either implementation, there may be some minor bugs

EmilyBourne:
> > @jalalium Don't forget to add the extra tests to satisfy the coverage checker
> 
> How do you link untested lines with their respective test?

You don't. If the line appears you are not testing what you think you are. Which test/line are you looking at?

EmilyBourne:
> > 
> 
> Those fail because `@allow_negative_index('a')` works for getting a single element but not for slices. All the tests that i flagged fail solely because the slice contains a variable. if one were to change the variable for its actual value, it would work!

Ok. Then we need an issue for this. I don't think #1311 or #1310 describe this problem correctly

pyccel/codegen/printing/ccode.py
```diff
@@ -1438,6 +1438,28 @@ def _print_NumpyRandint(self, expr):
...
+        '''
+        Convert a call to numpy.sum to the equivalent function in C.
+        '''
+        if not isinstance(expr.arg, (NumpyArray, Variable, IndexedElement)):
+            raise TypeError(f'Expecting a NumpyArray, given {type(expr.arg)}')
+        dtype, prec, name = (expr.arg.dtype,
+                             expr.arg.precision,
+                             self._print(expr.arg))
+        if prec == -1:
+            prec = 8
```
EmilyBourne, line 1451: You should use the `default_precision` from `ast.datatypes` not define your own precision. The current method will probably lead to failures for windows with integers as you will see when you add tests ;)
```suggestion
            prec = default_precision[self._print(dtype)]
```

pyccel/ast/numpyext.py
```diff
@@ -563,8 +563,15 @@ def __init__(self, arg):
...
-        self._precision = get_final_precision(arg)
+        if isinstance(arg.dtype, (NativeBool, NativeInteger)):
+            self._dtype = NativeInteger()
+            self._precision = 8
+        elif isinstance(arg.dtype, NativeFloat):
+            self._dtype = NativeFloat()
+            self._precision = default_precision[str(arg.dtype)]
+        elif isinstance(arg.dtype, NativeComplex):
+            self._dtype = NativeComplex()
+            self._precision = default_precision[str(arg.dtype)]
```
EmilyBourne, line 574: Are you sure about the precision here? Numpy has a `np.float128` type. When I use `np.sum` on an array of this type I get a `np.float128`.
Maybe it should be:
```suggestion
        if isinstance(arg.dtype, NativeBool):
            self._dtype = NativeInteger()
        else:
            self._dtype = arg.dtype
        self._precision = max(arg.precision, default_precision[str(self._dtype)]
```

EmilyBourne (changes requested):

pyccel/ast/numpyext.py
jalalium: This is much better, thank you! Ideally, we would like `precision` to be one larger than `arg.precision` to avoid overflow issues. This applies to all precisions except the largest one if each datatype. I will fix this.

EmilyBourne: No. We would like the precision to match what is done in numpy. They can decide what is best adapted. Imagine if the precision increased every time you did a calculation, you would soon run out of memory!

EmilyBourne (approved):

tests/epyccel/test_arrays.py
```diff
@@ -4032,6 +3474,20 @@ def test_iterate_slice(language):
...
     assert f1(i) == f2(i)
 
+#==============================================================================
+# NUMPY SUM
+#==============================================================================
+
+def test_arr_bool_sum(language):
+    f1 = arrays.arr_bool_sum
+    f2 = epyccel(f1, language = language)
+    assert f1() == f2()
```
EmilyBourne, line 3484: It would be good to check that the return type is as expected:
```
assert isinstance(f1(), type(f2()))
```

jalalium: Done!

yguclu: @EmilyBourne Is this point resolved?

pyccel/ast/numpyext.py
yguclu: @EmilyBourne Is this point resolved?

pyccel/stdlib/ndarrays/ndarrays.c
yguclu: @EmilyBourne Is this point resolved?

tests/epyccel/test_arrays.py
EmilyBourne: It is fixed here, but I don't think there are any tests for the dtype of sums of any other types.

pyccel/stdlib/ndarrays/ndarrays.c
EmilyBourne: I am happy with the current implementation. It is clearer with the comments. But I didn't ask the question you(@yguclu ) and @jalalium asked if there was a cleaner way of looping over the elements. I have suggested some possible alternatives. I think it is now for you two to decide if you are happy with the current implementation or if you prefer an alternative

pyccel/ast/numpyext.py
EmilyBourne: Yes I just left the conversation open so @jalalium could see my reply

tests/epyccel/test_arrays.py
yguclu: By the way, I think it is better if we check exact type equality rather than using `isinstance`. In the latter case `type(f1())` could be a subclass of `type(f2())`, which is not what we want. I recommend instead
```python
assert type(f1()) == type(f2())
```

EmilyBourne: I think it should be:
```
assert type(f1()) is type(f2()) #pylint: disable=unidiomatic-typecheck
```

yguclu: And we should disable the pylint check on the whole test file...

pyccel/stdlib/ndarrays/ndarrays.c
EmilyBourne: Ah. It is definitely supposed to be. Whether it always is may be related to PR #1299 . In any case the code now uses `arr.length` and still passes the tests

tests/epyccel/test_arrays.py
yguclu: The unit tests can be improved in a separate PR

yguclu (approved):

PR #1312, opened 2023-01-24
Use the same line endings throughout

EmilyBourne:
Developers do not all work on the same OS. This sometimes leads to different line endings in different files. In addition to sometimes causing confusing diffs, this can cause problems for Codacy (old mac systems use `\r` for line endings but git only supports `\n` and `\r\n` which means that `git diff` can give unexpected results).
The `.gitattribute` file ensures that line endings are homogeneous over commits from different systems as detailed here:
https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings

This PR adds the `.gitattribute` file and corrects files with unusual line endings


yguclu (approved):

EmilyBourne:
> Good job @EmilyBourne! I suppose this PR can be merged anytime?

Yes. Probably sooner rather than later to avoid the problem reoccurring

PR #1316, opened 2023-01-26
Always use LF line endings

EmilyBourne:
PR #1312 ensured that line endings are git compatible but did not specify which line endings should be used. As a result files with lots of changes in Windows (e.g see PR #1309) are changing line ending leading to big diffs.

This PR fixes the line endings to LF (line feed) format, which is the standard in Unix.


yguclu:
The PR description seems to be missing some text at the end.

EmilyBourne:
> The PR description seems to be missing some text at the end.

Oops. Fixed

yguclu (approved):

PR #1319, opened 2023-01-26
Allow declaration of multiple np.linspace arrays in Fortran

### LINKED ISSUE #1318: Cannot declare two np.linspace arrays when using Fortran
### Describe the bug
Fortran code generator tries to declare `integer(i64) :: linspace_index` variable each time an `np.linspace` array is created without checking if the variable has already been added to the scope. This throws an error in the `Scope.insert_variable` method

### To Reproduce
```python
import numpy as np
    
if __name__ == '__main__':
    x = np.linspace(0, 2, 128)
    y = np.linspace(0, 2, 128)
    print(x)
    print(y)
   
```

### Error details
```Fortan
 line 278, in insert_variable
    raise RuntimeError('New variable {} already exists in scope'.format(name))
RuntimeError: New variable linspace_index already exists in scope
```

### Expected behavior
The two variables should have declared without a problem.
### Language
Fortran

### Additional context
This bug is urgent since it is used in `laplace_2d_mod` test file in the benchmarks

jalalium:
Avoid name clashes in Fortran when creating two or more 1D Numpy arrays in the same scope through multiple calls to `np.linspace`. Fixes #1318.


yguclu (approved):
Looks good to me!

pyccel/codegen/printing/fcode.py
```diff
@@ -954,7 +954,8 @@ def _print_NumpyLinspace(self, expr):
         else:
             template = '[(({start} + {index}*{step}), {index} = {zero},{end})]'
             var = Variable('int', 'linspace_index')
-            self.scope.insert_variable(var)
+            if var.name not in self.scope.variables:
```
EmilyBourne, line 957: I am not sure why this is needed. Is it because the name is not being chosen by the scope (as described in the developer docs) leading to name collisions?

EmilyBourne (changes requested):

pyccel/codegen/printing/fcode.py
EmilyBourne: The name should be chosen by the scope to avoid collisions. If you want to reuse the variable then you should collect it from the scope not create a new one. I don't think it's worth putting this in this PR though. I plan to adapt the function `get_temporary_variable` to allow reuse at some point, so if you use this then I should find it when I get round to that

jalalium: Exactly, do you think it is better to get a new suffixed variable for each linspace declaration? Since it is a temporary variable, I think just one will do.

yguclu: This is a good point. I think that we should reuse the same temporary variable if it already exists in the scope, but how can we be sure that `linspace_index` was not created by the user for another purpose?

yguclu: -- as usual GitHub does not update the conversation in real time, so the comments are not reported in the correct order --

EmilyBourne: We cannot be sure which is why we must use either:
- `get_new_temporary_variable` to create a temporary
- `get_new_name` + `Variable`

However the problem is that we then cannot be certain of the variable name on the second pass which makes it hard to search for it in the scope.

That's why I prefer using `get_new_temporary_variable` to create a `Dummy_XXX` Variable that hopefully one day we will be able to reuse in various contexts not just in the linspace index

yguclu: @EmilyBourne I agree with you, we should indeed use `get_new_temporary_variable` here.

I suppose the same bug appears in several other places across Pyccel....

jalalium: Noted, I will use `get_new_temporary_variable` instead.

jalalium: is [this](https://github.com/pyccel/pyccel/pull/1319/commits/7dbc6c4f8333f95a64ba835c306cd99e34fcbad8) better?

yguclu: Again, it looks good to me 😅 

@EmilyBourne I've just noticed that the function is called `get_temporary_variable` (without `new`). Such a name suggests that an existing temporary variable could be provided if available, although that is not the case.

EmilyBourne: Ah sorry. I checked the name and still got it wrong! That impression is intended. I hope this function could be tweaked to allow that in the future. I should create an issue to describe this

EmilyBourne (approved):

PR #1324, opened 2023-02-08
Clean up Doc Coverage action

EmilyBourne:
- Ensure errors are not raised by trying to parse non-python files as python files
- Correct collection of objects to avoid repeated objects
- Don't check the docstrings of magic methods
- Update the developer docs to describe the new actions triggered in the review process
- Stop raising an error when optional sections are missing


ci_tools/process_results.py
```diff
@@ -0,0 +1,48 @@
...
+                        help='Github step summary')
+args = parser.parse_args()
+
+warning_codes = ['EX01', 'SA01']
+
+errors = {}
+warnings = {}
+with open(args.report, 'r', encoding='utf-8') as f:
+    for line in f:
+        file_name, code, msg = line.split(':')
```
EmilyBourne, line 19: @jalalium This command does not work in my branch. Any idea why? Does it fail if I add a function?

PR #1325, opened 2023-02-14
Ignore certain objects at numpydoc validation

jalalium:
This PR aims to ignore the following objects from numpydoc validation:

- If class is `SemanticParser` and `name.startswith('_visit_')`
- If class is `SyntacticParser` and `name.startswith('_visit_')`
- if file in `codegen.printing` and `name.startswith('_print_')`

Additionally, the function `should_ignore` can be used should one need to extend the objects to ignore.

### DIFF
--- a/ci_tools/list_docs_tovalidate.py
+++ b/ci_tools/list_docs_tovalidate.py
@@ -20,14 +20,20 @@ def should_ignore(name):
         bool
             True if object should be ignored, False otherwise.
     '''
-    patterns_to_ignore = ['SemanticParser._visit_',
-                          'SyntacticParser._visit_',
-                          'codegen.printing._print_']
     obj_name = name.split('.')[-1]
+    #ignore magic methods
     if obj_name.startswith('__') and obj_name.endswith('__'):
         return True
-    if any(pattern in name for pattern in patterns_to_ignore):
+    #ignore _visit_ methods in the SemanticParser class
+    if 'SemanticParser._visit_' in name:
         return True
+    #ignore _visit_ methods in the SyntaxParser class
+    if 'SyntaxParser._visit_' in name:
+        return True
+    #ignore _print_ methods in the codegen.printing module
+    if '_print_' in name and 'pyccel.codegen.printing' in name:
+        return True
+
     return False
 
 if __name__ == '__main__':
--- a/ci_tools/list_docs_tovalidate.py
+++ b/ci_tools/list_docs_tovalidate.py
@@ -33,7 +33,6 @@ def should_ignore(name):
     #ignore _print_ methods in the codegen.printing module
     if '_print_' in name and 'pyccel.codegen.printing' in name:
         return True
-
     return False
 
 if __name__ == '__main__':
@@ -76,7 +75,7 @@ if __name__ == '__main__':
             # inside a function or a class is updated to include
             # the name of the parent object
             if isinstance(node, (FunctionDef, ClassDef)):
-                if should_ignore(node.name):
+                if should_ignore('.'.join(prefix, node.name)):
                     continue
                 if any((node.lineno <= x <= node.end_lineno
                         for x in line_nos)) and not node.name.startswith('inner_function'):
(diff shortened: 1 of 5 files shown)


EmilyBourne (approved):

EmilyBourne:
@jalalium Is this tested?

jalalium:
> @jalalium Is this tested?

Yes, as you can see from [f7a2f36](https://github.com/pyccel/pyccel/pull/1325/commits/f7a2f361318bcbaf463680b82f9f0f01ac7c42b2), the validator ignores the objects you mentioned

ci_tools/process_results.py
```diff
@@ -10,25 +10,38 @@
                         help='Github step summary')
 args = parser.parse_args()
 
+all_codes = ['GL01', 'GL02', 'GL03', 'GL05', 'GL06', 'GL07', 'GL08', 'GL09',
+             'GL10', 'SS01', 'SS02', 'SS03', 'SS04', 'SS05', 'SS06', 'ES01',
+             'PR01', 'PR02', 'PR03', 'PR04', 'PR05', 'PR06', 'PR07', 'PR08',
+             'PR09', 'PR10', 'RT01', 'RT02', 'RT03', 'RT04', 'RT05', 'YD01',
+             'SA01', 'SA02', 'SA03', 'SA04', 'EX01']
+
 warning_codes = ['EX01', 'SA01']
```
yguclu, line 19: Right now `all_codes` contains both error and warning codes, and after specifying `warning_codes` you know that the remaining codes must be errors.

Instead I find it clearer to define two disjoint sets, `error_codes` and `warning_codes`. You can use an assertion to make sure there is no duplicate:

```python
warning_codes = {'EX01', 'SA01'}
error_codes = {'GL01', ..., 'SA03'}

assert warning_codes & error_codes == set()
```
And later in the loop you can write clearly:

```python
if code in warning_codes:
    ...
elif code in error_codes:
    ...
else:
    parsing_errors.append(line)
    # no need for 'continue' statement
```

ci_tools/process_results.py
```diff
@@ -10,25 +10,38 @@
...
+                else:
+                    warnings[file_name].append(msg)
             else:
-                errors[file_name].append(msg)
+                if file_name not in errors:
+                    errors[file_name] = [msg]
+                else:
+                    errors[file_name].append(msg)
+        except ValueError:
+            parsing_errors.append(line)
```
yguclu, line 42: When is this exception raised?

ci_tools/list_docs_tovalidate.py
```diff
@@ -6,55 +6,86 @@
...
+
+        Parameters
+        ----------
+        name : str
+            Absolute name of the inspected object.
+
+        Returns
+        -------
+        bool
+            True if object should be ignored, False otherwise.
```
yguclu, line 21: Talking about numpydoc... isn't there an excessive indent in this docstring? :sweat_smile:

yguclu (changes requested):
Good job! I have a couple of comments

ci_tools/process_results.py
EmilyBourne: It is raised if `split(':')` fails to find the correct number of colons

ci_tools/list_docs_tovalidate.py
EmilyBourne: numpydoc doesn't complain about that. But it does complain about a lack of an extended summary

ci_tools/process_results.py
yguclu: Like in this case?
```
>>> a, b, c = (1, 2)
  File "<stdin>", line 1, in <module>
ValueError: not enough values to unpack (expected 3, got 2)
```
A comment line would be useful.

ci_tools/process_results.py
jalalium: got it! i will fix it!

ci_tools/process_results.py
jalalium: Noted!

yguclu (approved):

PR #1328, opened 2023-02-20
Fix allocation mishandling in CUDA array creation

### LINKED ISSUE #1327: Bug in CUDA array creation due to allocation mishandling
### Describe the bug
The allocation of the CUDA array is not being properly handled, leading to a segmentation fault.

### To Reproduce
For example :
```python
ffrom pyccel import cuda

if __name__ == '__main__':
    arr = cuda.array([0,1,2,3,4], memory_location = 'device')
    print(arr)
```

### Language
Cuda

framdani:
This PR fixes a bug where the allocation of the CUDA array was not being properly handled. Fixes [pyccel-cuda#2](https://github.com/pyccel/pyccel-cuda/issues/2).

**Changes Made:**
- Fixed the allocation of the CUDA array to prevent segmentation fault.
- Added tests to ensure proper handling of the CUDA array allocation.

### DIFF
--- a/.github/actions/pytest_run_cuda/action.yml
+++ b/.github/actions/pytest_run_cuda/action.yml
@@ -10,7 +10,7 @@ runs:
   steps:
     - name: Ccuda tests with pytest
       run: |
-        python -m pytest -n auto -rx -m "not (parallel or xdist_incompatible) and ccuda" --ignore=symbolic --ignore=ndarrays
+        python -m pytest -n auto -rx -m "not (parallel or xdist_incompatible) and ccuda" --ignore=symbolic --ignore=ndarrays --ignore=cuda_ndarrays
         pyccel-clean
       shell: ${{ inputs.shell_cmd }}
       working-directory: ./tests
--- a/tests/cuda_ndarrays/conftest.py
+++ b/tests/cuda_ndarrays/conftest.py
@@ -1,5 +1,4 @@
-# pylint: disable=missing-module-docstring/
-# pylint: disable=arguments-differ, inconsistent-return-statements
+# pylint: disable=missing-function-docstring, missing-module-docstring/
 import subprocess
 import os
 import pathlib
@@ -17,23 +16,24 @@ def pytest_collect_file(parent, path):
         if NEEDS_FROM_PARENT:
             return CTestFile.from_parent(path=pathlib.Path(path), parent=parent)
         return CTestFile(parent=parent, path=pathlib.Path(path))
+    return None
 
 
 class CTestFile(pytest.File):
     """
-    A custom file handler class for C unit test files.
+    A custom file handler class for CCuda unit test files.
     """
 
     @classmethod
-    def from_parent(cls, **kwargs):
-        return super().from_parent(**kwargs)
+    def from_parent(cls, parent, **kwargs):
+        return super().from_parent(parent=parent, **kwargs)
 
     def collect(self):
         """
         Overridden collect method to collect the results from each
-        C unit test executable.
+        CCuda unit test executable.
         """
-        # Run the exe that corresponds to the .c file and capture the output.
+        # Run the exe that corresponds to the .cu file and capture the output.
         test_exe = os.path.splitext(self.fspath.basename)[0]
         rootdir = self.config.rootdir.strpath
         test_exe = os.path.relpath(test_exe)
@@ -93,8 +93,8 @@ class CTestItem(pytest.Item):
         self.test_result = test_result
 
     @classmethod
-    def from_parent(cls, *, test_result, **kwargs):
-        return super().from_parent(test_result=test_result, **kwargs)
+    def from_parent(cls, parent, **kwargs):
+        return super().from_parent(parent=parent, **kwargs)
 
     def runtest(self):
         """The test has already been run. We just evaluate the result."""
@@ -105,13 +105,14 @@ class CTestItem(pytest.Item):
         """"Called to display header information about the test case."""
         return self.fspath, self.test_result["line_number"], self.name
 
-    def repr_failure(self, exception):
+    def repr_failure(self, excinfo, style=None):
         """
         Called when runtest() raises an exception. The method is used
         to format the output of the failed test result.
         """
-        if isinstance(exception.value, CTestException):
-            return (f"Test failed : {self.test_result['file_name']}:{self.test_result['line_number']} {self.test_result['function_name']} < {self.test_result['DSCR']} >\n INFO : {self.test_result['INFO']}")
+        if isinstance(excinfo.value, CTestException):
+            return f"Test failed : {self.test_result['file_name']}:{self.test_result['line_number']} {self.test_result['function_name']} < {self.test_result['DSCR']} >\n INFO : {self.test_result['INFO']}"
+        return None
 
 
 
--- a/tests/cuda_ndarrays/conftest.py
+++ b/tests/cuda_ndarrays/conftest.py
@@ -12,7 +12,6 @@ NEEDS_FROM_PARENT = hasattr(pytest.Item, "from_parent")
 def pytest_collect_file(parent, path):
     """
     A hook to collect test_*.cu test files.
-
     """
     if path.ext == ".cu" and path.basename.startswith("test_"):
         if NEEDS_FROM_PARENT:
@@ -22,7 +21,6 @@ def pytest_collect_file(parent, path):
 def pytest_collection_modifyitems(items):
     """
     a hook to modify the items before the tests for C Cuda unit test files.
-
     """
     for item in items:
         if item.fspath.ext == ".cu":
@@ -31,7 +29,6 @@ def pytest_collection_modifyitems(items):
 class CTestFile(pytest.File):
     """
     A custom file handler class for C unit test files.
-
     """
 
     @classmethod
@@ -42,7 +39,6 @@ class CTestFile(pytest.File):
         """
         Overridden collect method to collect the results from each
         C unit test executable.
-
         """
         # Run the exe that corresponds to the .c file and capture the output.
         test_exe = os.path.splitext(str(self.fspath))[0]
@@ -96,7 +92,6 @@ class CTestItem(pytest.Item):
     """
     Pytest.Item subclass to handle each test result item. There may be
     more than one test result from a test function.
-
     """
 
     def __init__(self, *, test_result, **kwargs):
@@ -121,7 +116,6 @@ class CTestItem(pytest.Item):
         """
         Called when runtest() raises an exception. The method is used
         to format the output of the failed test result.
-
         """
         if isinstance(exception.value, CTestException):
             return (f"Test failed : {self.test_result['file_name']}:{self.test_result['line_number']} {self.test_result['function_name']} < {self.test_result['DSCR']} >\n INFO : {self.test_result['INFO']}")
--- a/tests/cuda_ndarrays/conftest.py
+++ b/tests/cuda_ndarrays/conftest.py
@@ -1,5 +1,5 @@
-# pylint: disable=missing-function-docstring, missing-module-docstring/
-# pylint: disable=arguments-differ, inconsistent-return-statements, protected-access, abstract-method/
+# pylint: disable=missing-module-docstring/
+# pylint: disable=arguments-differ, inconsistent-return-statements
 import subprocess
 import os
 import pathlib
@@ -18,13 +18,6 @@ def pytest_collect_file(parent, path):
             return CTestFile.from_parent(path=pathlib.Path(path), parent=parent)
         return CTestFile(parent=parent, path=pathlib.Path(path))
 
-def pytest_collection_modifyitems(items):
-    """
-    a hook to modify the items before the tests for C Cuda unit test files.
-    """
-    for item in items:
-        if item.fspath.ext == ".cu":
-            item._nodeid = item.nodeid + " < " + item.test_result["DSCR"] + " >"
 
 class CTestFile(pytest.File):
     """
@@ -81,10 +74,10 @@ class CTestFile(pytest.File):
                 test_results[-1][token] = data
         for test_result in test_results:
             if NEEDS_FROM_PARENT:
-                yield CTestItem.from_parent(name = test_result["function_name"],
+                yield CTestItem.from_parent(name = test_result["function_name"] + " < " + test_result["DSCR"] + " >",
                         parent = self, test_result = test_result)
             else:
-                yield CTestItem(name = test_result["function_name"], parent = self,
+                yield CTestItem(name = test_result["function_name"] + " < " + test_result["DSCR"] + " >", parent = self,
                         test_result = test_result)
 
 
--- a/tests/cuda_ndarrays/conftest.py
+++ b/tests/cuda_ndarrays/conftest.py
@@ -1,0 +1,132 @@
+# pylint: disable=missing-function-docstring, missing-module-docstring/
+# pylint: disable=arguments-differ, inconsistent-return-statements, protected-access, abstract-method/
+import subprocess
+import os
+import pathlib
+import sys
+import shutil
+import pytest
+
+NEEDS_FROM_PARENT = hasattr(pytest.Item, "from_parent")
+
+def pytest_collect_file(parent, path):
+    """
+    A hook to collect test_*.cu test files.
+
+    """
+    if path.ext == ".cu" and path.basename.startswith("test_"):
+        if NEEDS_FROM_PARENT:
+            return CTestFile.from_parent(path=pathlib.Path(path), parent=parent)
+        return CTestFile(parent=parent, path=pathlib.Path(path))
+
+def pytest_collection_modifyitems(items):
+    """
+    a hook to modify the items before the tests for C Cuda unit test files.
+
+    """
+    for item in items:
+        if item.fspath.ext == ".cu":
+            item._nodeid = item.nodeid + " < " + item.test_result["DSCR"] + " >"
+
+class CTestFile(pytest.File):
+    """
+    A custom file handler class for C unit test files.
+
+    """
+
+    @classmethod
+    def from_parent(cls, **kwargs):
+        return super().from_parent(**kwargs)
+
+    def collect(self):
+        """
+        Overridden collect method to collect the results from each
+        C unit test executable.
+
+        """
+        # Run the exe that corresponds to the .c file and capture the output.
+        test_exe = os.path.splitext(str(self.fspath))[0]
+        rootdir = str(self.config.rootdir)
+        test_exe = os.path.relpath(test_exe)
+        ndarray_path =  os.path.join(rootdir , "pyccel", "stdlib", "ndarrays")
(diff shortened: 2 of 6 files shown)


EmilyBourne (changes requested):
This looks right to me. I assume there are already tests which test this part of the code? Do you have any idea what additional test might have caught this earlier (e.g. we use valgrind for ndarrays, I don't know if that is possible for cuda)?

Please can you expand the PR description to explain the changes that you made.

bauom (changes requested):
we should start populating a new test file that can run locally for now so we do not have the same problem of run time errors that slips by between commits.

can you create a file that does the same as what in the file below just for the array creation tests.
https://github.com/pyccel/pyccel/blob/master/tests/ndarrays/test_ndarrays.c

framdani:
> This looks right to me. I assume there are already tests which test this part of the code? Do you have any idea what additional test might have caught this earlier (e.g. we use valgrind for ndarrays, I don't know if that is possible for cuda)?
> 
> Please can you expand the PR description to explain the changes that you made.

There were no tests that covered this part of the code previously. I believe the tests I've added now ensure that the changes I made do not introduce any new bugs.

EmilyBourne:
@framdani The tests look good, but I think you are missing something so that they are compiled and run in the CI. Ie. the equivalent of this file: https://github.com/pyccel/pyccel/blob/master/tests/ndarrays/conftest.py and/or this file : https://github.com/pyccel/pyccel/blob/cuda_devel/.github/actions/valgrind_run/action.yml

framdani:
> we should start populating a new test file that can run locally for now so we do not have the same problem of run time errors that slips by between commits.
>

EmilyBourne:
> @EmilyBourne you're right , the tests I added were intended to be run locally for now. I'll do some research and ensure that the necessary configuration are added so that they can run in the CI.

@framdani Ok. No need to reinvent the wheel, I think it should be as simple as copying this file into your folder : https://github.com/pyccel/pyccel/blob/master/tests/ndarrays/conftest.py  and:
- modifying it to collect `.cu` files instead of `.c` files
- Fixing the compilation command : https://github.com/pyccel/pyccel/blob/master/tests/ndarrays/conftest.py#L52-L53

bauom:
> > @EmilyBourne you're right , the tests I added were intended to be run locally for now. I'll do some research and ensure that the necessary configuration are added so that they can run in the CI.
> 
> @framdani Ok. No need to reinvent the wheel, I think it should be as simple as copying this file into your folder : https://github.com/pyccel/pyccel/blob/master/tests/ndarrays/conftest.py and:
> 
>     * modifying it to collect `.cu` files instead of `.c` files
> 
>     * Fixing the compilation command : https://github.com/pyccel/pyccel/blob/master/tests/ndarrays/conftest.py#L52-L53

that should work but we should ignore them in GitHub action for now as we can not run them.

EmilyBourne:
> > > @EmilyBourne you're right , the tests I added were intended to be run locally for now. I'll do some research and ensure that the necessary configuration are added so that they can run in the CI.
> > 
> > 
> > @framdani Ok. No need to reinvent the wheel, I think it should be as simple as copying this file into your folder : https://github.com/pyccel/pyccel/blob/master/tests/ndarrays/conftest.py and:
> > ```
> > * modifying it to collect `.cu` files instead of `.c` files
> > 
> > * Fixing the compilation command : https://github.com/pyccel/pyccel/blob/master/tests/ndarrays/conftest.py#L52-L53
> > ```
> 
> that should work but we should ignore them in GitHub action for now as we can not run them.



> > > @EmilyBourne you're right , the tests I added were intended to be run locally for now. I'll do some research and ensure that the necessary configuration are added so that they can run in the CI.
> > 
> > 
> > @framdani Ok. No need to reinvent the wheel, I think it should be as simple as copying this file into your folder : https://github.com/pyccel/pyccel/blob/master/tests/ndarrays/conftest.py and:
> > ```
> > * modifying it to collect `.cu` files instead of `.c` files
> > 
> > * Fixing the compilation command : https://github.com/pyccel/pyccel/blob/master/tests/ndarrays/conftest.py#L52-L53
> > ```
> 
> that should work but we should ignore them in GitHub action for now as we can not run them.

Oh yes! I keep forgetting that we still can't run these tests. In that case I wonder if it's possible to split it into 2 tests?
1. Compile with `-Wall -Werror` in the CI
2. Execute (not run on CI but available for testing before merging)

tests/cuda_ndarrays/test_cuda_ndarrays.cu
```diff
@@ -0,0 +1,537 @@
...
+    assert_int32(arr.nd, 1, "testing the number of dimensions", __func__, __FILE__, __LINE__);
+    assert_int64(arr.shape[0], tmp_shape[0], "testing the shape", __func__, __FILE__, __LINE__);
+    for(int i = 0; i < tmp_shape[0]; i++)
+    {
+        assert_double(arr.nd_double[i], cuda_array_dummy[i], "testing the data", __func__, __FILE__, __LINE__);
+    }
+    cuda_free_host(arr);
+}
+
+void test_cuda_array_create_managed_double()
```
bauom, line 113: why should we create another array to check the `managed_memory` array

bauom (changes requested):

tests/cuda_ndarrays/test_cuda_ndarrays.cu
framdani: Actually, there is no need to create an extra array if I only needed to access the data. I just forgot to update it.
My original intention was to have a separate copy of the data in host memory and apply some operations on it before applying the same operations on the original array from a kernel. What do you think should I proceed , or keeping it simple for this specific test may be sufficient ?

bauom: I think just doing the same as the test for `host_memory` without using the additional array.
the other tests will be added when we add `kernel` tests.

.github/actions/pytest_run_cuda/action.yml
```diff
@@ -10,7 +10,7 @@ runs:
   steps:
     - name: Ccuda tests with pytest
       run: |
-        python -m pytest -n auto -rx -m "not (parallel or xdist_incompatible) and ccuda" --ignore=symbolic --ignore=ndarrays
+        python -m pytest -n auto -rx -m "not (parallel or xdist_incompatible) and ccuda" --ignore=tests/symbolic --ignore=tests/ndarrays --ignore=tests/cuda_ndarrays
         pyccel-clean
       shell: ${{ inputs.shell_cmd }}
-      working-directory: ./tests
+      working-directory: ./
```
EmilyBourne, line 16: Please revert the dangerous changes in this file (which I assume come from a bad merge). See Issue #1344 and PR #1348 for more details

.github/workflows/Github_pytest.yml
```diff
@@ -200,7 +184,7 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: 3.7
-      - name: Install Python dependencies
+      - name: Install python dependencies
```
EmilyBourne, line 187: Bad merge. Please use the version of this file from `cuda_main_temp`

setup.cfg
```diff
@@ -27,7 +27,6 @@ test =
     pytest >= 2.7
     scipy >= 1.5.0
     mpi4py >= 3.0.0
-    tblib >= 1.7.0
     coverage >= 3.1
     astunparse >= 1.6.0
     pytest-xdist >= 1.16
```
EmilyBourne, line 32: Bad merge. Please use the version of this file from `cuda_main_temp`

EmilyBourne (changes requested):
It looks like there was some confusion when fixing the collisions. Please can you fix this

EmilyBourne:
Fixes [pyccel-cuda#2](https://github.com/pyccel/pyccel-cuda/issues/2) ?

tests/cuda_ndarrays/conftest.py
```diff
@@ -0,0 +1,132 @@
...
+
+def pytest_collect_file(parent, path):
+    """
+    A hook to collect test_*.cu test files.
+
+    """
+    if path.ext == ".cu" and path.basename.startswith("test_"):
+        if NEEDS_FROM_PARENT:
+            return CTestFile.from_parent(path=pathlib.Path(path), parent=parent)
+        return CTestFile(parent=parent, path=pathlib.Path(path))
```
EmilyBourne, line 20: I agree with the commented pylint error

tests/cuda_ndarrays/conftest.py
```diff
@@ -0,0 +1,132 @@
+# pylint: disable=missing-function-docstring, missing-module-docstring/
+# pylint: disable=arguments-differ, inconsistent-return-statements, protected-access, abstract-method/
+import subprocess
+import os
+import pathlib
+import sys
+import shutil
+import pytest
+
+NEEDS_FROM_PARENT = hasattr(pytest.Item, "from_parent")
```
EmilyBourne, line 10: What does this do? Does it depend on the pylint version?

tests/cuda_ndarrays/conftest.py
```diff
@@ -0,0 +1,132 @@
...
+    """
+    if path.ext == ".cu" and path.basename.startswith("test_"):
+        if NEEDS_FROM_PARENT:
+            return CTestFile.from_parent(path=pathlib.Path(path), parent=parent)
+        return CTestFile(parent=parent, path=pathlib.Path(path))
+
+def pytest_collection_modifyitems(items):
+    """
+    a hook to modify the items before the tests for C Cuda unit test files.
+
```
EmilyBourne, line 25: Unnecessary empty line in docstring

tests/cuda_ndarrays/conftest.py
```diff
@@ -0,0 +1,132 @@
...
+    """
+    for item in items:
+        if item.fspath.ext == ".cu":
+            item._nodeid = item.nodeid + " < " + item.test_result["DSCR"] + " >"
+
+class CTestFile(pytest.File):
+    """
+    A custom file handler class for C unit test files.
+
```
EmilyBourne, line 34: Unnecessary empty line in docstring

tests/cuda_ndarrays/conftest.py
```diff
@@ -0,0 +1,132 @@
...
+    @classmethod
+    def from_parent(cls, **kwargs):
+        return super().from_parent(**kwargs)
+
+    def collect(self):
+        """
+        Overridden collect method to collect the results from each
+        C unit test executable.
+
```
EmilyBourne, line 45: Unnecessary empty line in docstring

tests/cuda_ndarrays/conftest.py
```diff
@@ -0,0 +1,132 @@
...
+        """
+        # Run the exe that corresponds to the .c file and capture the output.
+        test_exe = os.path.splitext(str(self.fspath))[0]
```
EmilyBourne, line 48: What is the type of `self.fspath`? I'm surprised that it needs to be converted to a string. I would have expected it to either already be a string, or be a type which has functions for handling this kind of operation

EmilyBourne: Same question for `self.config.rootdir` on the next line

tests/cuda_ndarrays/conftest.py
```diff
@@ -0,0 +1,132 @@
...
+        ndarray_path =  os.path.join(rootdir , "pyccel", "stdlib", "ndarrays")
+        cuda_ndarray_path = os.path.join(rootdir, "pyccel", "stdlib", "cuda_ndarrays")
+        comp_cmd = [shutil.which("nvcc"), test_exe + ".cu",
+                    os.path.join(ndarray_path, "ndarrays.c"),
+                    os.path.join(cuda_ndarray_path, "cuda_ndarrays.cu"),
+                    "-I", ndarray_path,
+                    "-I", cuda_ndarray_path,
+                    "-o", test_exe,]
+
+        subprocess.run(comp_cmd, check= 'TRUE')
```
EmilyBourne, line 60: Are you sure the `check` argument takes a string?

tests/cuda_ndarrays/conftest.py
```diff
@@ -0,0 +1,132 @@
...
+        return self.fspath, self.test_result["line_number"], self.name
+
+    def repr_failure(self, exception):
+        """
+        Called when runtest() raises an exception. The method is used
+        to format the output of the failed test result.
+
+        """
+        if isinstance(exception.value, CTestException):
+            return (f"Test failed : {self.test_result['file_name']}:{self.test_result['line_number']} {self.test_result['function_name']} < {self.test_result['DSCR']} >\n INFO : {self.test_result['INFO']}")
```
EmilyBourne, line 127: Have you come across other Exception types which should be ignored? Which ones? Would it not be safer to handle these types separately?

tests/cuda_ndarrays/test_cuda_ndarrays.cu
```diff
@@ -0,0 +1,505 @@
...
+
+#define getname(X) #X
+
+void assert_double(double v1 , double v2, const char *dscr,
+        const char * func, const char *file, int32_t line)
+{
+    if (v1 != v2)
+    {
+        printf("[FAIL] %s:%d:%s\n", file, line, func);
+        printf("[INFO] %s:%f != %s:%f\n", getname(v1), v1, getname(v2),v2);
```
EmilyBourne, line 16: Am I right that this is equivalent to:
```c
printf("[INFO] v1:%f != v2:%f\n", v1, v2);
```
?
Can you explain the advantage of using `getname` here?

EmilyBourne (changes requested):

tests/cuda_ndarrays/conftest.py
framdani: `self.fspath` is an attribute of type `py._path.local.localPath` which why it needs to be converted to a string. As an alternative I can use the attribute `basename` to get the filename with the extension as a string without the need for conversion.

framdani: Same for `self.config.rootdir`. It does have also its own way of handling the conversion  which is the attribute `strpath`.

framdani: Updated here 4fd629acda3bc1e8540ead05ed2162c4e8c3530a

tests/cuda_ndarrays/conftest.py
framdani: Fixed 2a376d0a156d1ca4ca39fdd6c244c4b0b145c000

tests/cuda_ndarrays/conftest.py
framdani: Fixed 2a376d0a156d1ca4ca39fdd6c244c4b0b145c000

tests/cuda_ndarrays/conftest.py
framdani: Fixed 2a376d0a156d1ca4ca39fdd6c244c4b0b145c000

tests/cuda_ndarrays/conftest.py
framdani: It is checking if `pytest.Item` class has the `from_parent` attribute. The purpose is to ensure the compatibility with different versions of pytest

tests/cuda_ndarrays/conftest.py
framdani: Sorry, should be a boolean not a string

tests/cuda_ndarrays/test_cuda_ndarrays.cu
framdani: Yes, that's correct. Since the specific args v1 and v2 will always be named as such in the function, when the assertion fails, the printf statement using `getname` will always print "v1" and "v2". Updated here c899cf6873aa2ef11c2f9d53ce3874463a5f1415

tests/cuda_ndarrays/conftest.py
bauom: this was left from my testing to raise different error messages.
there should not be any other exceptions possible because we are the one raising them.

tests/cuda_ndarrays/conftest.py
EmilyBourne: Have you found versions of pytest which don't have the attribute?

tests/cuda_ndarrays/conftest.py
EmilyBourne: If there is a typo in the code which raises the error then a variety of errors (e.g. `NameError`) could be raised. This is very similar to [this error](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/broad-exception-caught.html). In this case as I understand it pytest will call this function and the error will be ignored. That is why it is dangerous to ignore all other errors. I suggest:

```suggestion
        if isinstance(exception.value, CTestException):
            return (f"Test failed : {self.test_result['file_name']}:{self.test_result['line_number']} {self.test_result['function_name']} < {self.test_result['DSCR']} >\n INFO : {self.test_result['INFO']}")
        else:
             return super().repr_failure(exception)
```

tests/cuda_ndarrays/conftest.py
EmilyBourne: Actually I don't see this attribute in the documentation so maybe my question is rather where have you seen this attribute?

EmilyBourne: (maybe the answer is in the `ndarrays` folder?)

bauom: [this](https://docs.pytest.org/en/7.1.x/deprecations.html#fspath-argument-for-node-constructors-replaced-with-pathlib-path) shows that it was deprecated and I think I added it for it to be backward compatible.

EmilyBourne: Ok. That is fine with me :+1:

tests/cuda_ndarrays/conftest.py
```diff
@@ -0,0 +1,132 @@
+# pylint: disable=missing-function-docstring, missing-module-docstring/
+# pylint: disable=arguments-differ, inconsistent-return-statements, protected-access, abstract-method/
```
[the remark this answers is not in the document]
EmilyBourne: So I'm clear here. My problem is with the second line. In pyccel we have agreed to disable the docstring checks in the tests, but nothing else

tests/cuda_ndarrays/conftest.py
EmilyBourne: Is there no `super().repr_failure` function?

framdani: I did make the requested changes, I just wanted to run some tests before committing them. `super.repr_failure()` is available and I have modified the code here d65ca5218b54097980aab6c7f8b4efa35193b531.

EmilyBourne: Perfect thanks very much for fixing all this. I didn't realise the other conftest file was such a mess. We mustn't have reviewed it properly :(

EmilyBourne (approved):

tests/cuda_ndarrays/conftest.py
framdani: No problem at all :sweat_smile:

bauom (approved):

PR #1345, opened 2023-03-17
Ensure cuda files are installed with pyccel

### LINKED ISSUE #1344: Cuda tests run from the wrong folder results in wrong coverage
The cuda tests are run from the root directory instead of the tests directory. As a result the installed version of pyccel is not used which results in wrong coverage. This change was made in commit 452d2e39333a78e150e491ea14ca063f84fc70b1 . I suspect this was done to fix the fact that the cuda_ndarrays folder was not available. This may be due to an incorrect installation process (.cu files are not in the MANIFEST) or due to the import of cuda_ndarrays not being specified correctly as described in the [docs](https://github.com/pyccel/pyccel/blob/master/developer_docs/codegen_stage.md#imports).

> I think the problem is that the tests are being run from the wrong folder (`.` see [here](https://github.com/pyccel/pyccel/blob/cuda_main_temp/.github/actions/pytest_run_cuda/action.yml#L16)) instead of (`tests` see [here](https://github.com/pyccel/pyccel/blob/master/.github/actions/pytest_run/action.yml#L30)). This change was made by @bauom here : https://github.com/pyccel/pyccel/commit/452d2e39333a78e150e491ea14ca063f84fc70b1 .
>
> As a result you are using the local `pyccel/` folder instead of the installed one. But the coverage checks against the installed folder so it doesn't see that you have used the code. @framdani this is not your fault. I will open a PR to fix the problem which you can merge into your branch.

_Originally posted by @EmilyBourne in https://github.com/pyccel/pyccel/issues/1266#issuecomment-1473452318_

### LINKED ISSUE #1346: Cuda coverage tests don't work
The coverage tests report using an absolute path. As the cuda tests use a docker (and a different python version) compared to the linux tests the source folder is different for these two cases. Therefore when they are combined (with the source explicitly specified) the cuda results are ignored. Coverage has a `relative_path` flag which is designed to handle problems where coverage is tested in different environments (exactly our problem). The use of this flag should fix the problem.

*as seen in #1266*

EmilyBourne:
Include cuda files in the manifest to ensure that they are installed with pyccel. Fixes #1344 . This should help with the coverage as the python files that are run are the installed versions not the local versions.

Use coverage's `[paths]` options to rename files tested on different systems. Fixes #1346 .

Pull some updates from master regarding the installation and test setup

### DIFF
--- a/.github/actions/coverage_install/action.yml
+++ b/.github/actions/coverage_install/action.yml
@@ -13,7 +13,7 @@ runs:
         SITE_DIR=$(dirname ${INSTALL_DIR})
         echo "SITE_DIR = ${SITE_DIR}"
         echo -e "import coverage; coverage.process_startup()" > ${SITE_DIR}/pyccel_cov.pth
-        echo -e "[run]\nparallel = True\nrelative_files = True\nsource = ${INSTALL_DIR}\ndata_file = $(pwd)/.coverage\n[report]\ninclude = ${INSTALL_DIR}/*\n[xml]\noutput = cobertura.xml" > .coveragerc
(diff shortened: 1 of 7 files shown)


framdani (approved):
The issue with coverage seems to be fixed now. Thank you

bauom (approved):

PR #1347, opened 2023-03-22
WIP: Try to use reduce memory usage for Python 3.11

EmilyBourne:


EmilyBourne:
See https://github.com/pyccel/pyccel/actions/runs/4364605044/jobs/7632126695 for a failure with an output message

PR #1363, opened 2023-04-18
Fix minor bot issues

EmilyBourne:
Fix some remaining issues with the bot.

- Fix base reference for lint test
- Raise a helpful message if tests are requested which are not recognised
- Move pull_request_review bot to separate file to reduce list of skipped jobs
- Use `get_job_information` to check job results to prevent mix-ups when multiple jobs are run simultaneously on a non-draft PR
- Fix wrong tag when passing to "ready_for_review" stage


EmilyBourne:
Tests were run on the fork : https://github.com/EmilyBourne/pyccel/pull/16

yguclu:
Once again macOS is failing as it cannot find the object file `ndarrays.o`. Do you have an idea of why this is happening?

EmilyBourne:
> Once again macOS is failing as it cannot find the object file `ndarrays.o`. Do you have an idea of why this is happening?

It is definitely a race condition as I ran a test to check that `filelock` is working correctly on macOS. But I have no idea where the race condition is

.github/workflows/pyccel_lint.yml
```diff
@@ -39,9 +42,8 @@ jobs:
         shell: bash
       - name: Collect diff information
         run: |
-          BASE_BRANCH=$GITHUB_BASE_REF
           git fetch
-          git diff origin/${BASE_BRANCH}..HEAD --no-indent-heuristic --unified=0 --output=pull_diff.txt --no-color
+          git diff origin/${{ inputs.base }}..HEAD --no-indent-heuristic --unified=0 --output=pull_diff.txt --no-color
```
bauom, line 46: could you explain this change.

EmilyBourne: Previously `$BASE_BRANCH` was defined 2 lines higher as `BASE_BRANCH=$GITHUB_BASE_REF`. The variable `$GITHUB_BASE_REF` is only defined when the test is triggered by a pull request event (open/synchronize/mark_as_draft/etc) as documented [here](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables)) which means that it is not defined when the test is triggered via a commit using `/bot run lint` or `/bot mark as ready`. To fix this I use an input which is defined from a bot output. The bot uses the github CLI to investigate the PR and work out what the base branch is.

I removed the intermediate variable `BASE_BRANCH` as I can't see what it adds

yguclu:
@bauom What is the status of your review?

bauom (approved):
> @bauom What is the status of your review?
@yguclu sorry didn't notice the reply.

everything seems fine to me.

yguclu (approved):

PR #1369, opened 2023-04-26
Improve coverage check

### LINKED ISSUE #1368: Coverage is no longer optional
Until a few weeks ago having full coverage of a PR was encouraged but not enforced. This is no longer the case, however there are occasionally sections of code which cannot be tested on the runner (e.g. compiler settings for unusual setups) so this should not be mandatory.

cc: @yguclu

EmilyBourne:
Ensure that coverage is not compulsory for the bot to mark the PR as ready. This is done by asking the user to explicitly tell the bot to ignore failing coverage tests. In the same message the user is also asked to explain the failures and open appropriate issues. Fixes #1368 .
Additionally, discard code in `__str__` or `__repr__` functions from coverage output. These functions are used for debugging so the testing requirements do not need to be as stringent.

The bot is also patched so the flags are removed when it knocks a PR back to draft due to failing tests

### DIFF
--- a/ci_tools/bot_interaction.py
+++ b/ci_tools/bot_interaction.py
@@ -219,14 +219,20 @@ def mark_as_ready(pr_id):
 
     job_data = [j for j in data if j['name'] not in ('Bot', 'CleanUpBot')]
 
-    print(job_data)
-
-    failures = [j['name'] for j in job_data if j['conclusion'] in ('cancelled', 'failed')]
+    failures = [j['name'] for j in job_data if j['conclusion'] in ('cancelled', 'failure')]
 
     if any(j['name'] == 'Codacy Static Code Analysis' and j['conclusion'] in ('FAILURE', 'ACTION_REQUIRED') for j in running_job_data):
         failures.append('Codacy Static Code Analysis')
 
-    if failures:
+    ignore_coverage = accept_coverage_failure(pr_id)
+    coverage_failed = any('coverage' in f for f in failures) and not ignore_coverage
+    others_failed = any('coverage' not in f for f in failures)
+
+    if coverage_failed and not others_failed:
+        set_draft(pr_id)
+        leave_comment(pr_id, message_from_file('warn_coverage.txt'))
+
+    elif others_failed:
         set_draft(pr_id)
         message = message_from_file('set_draft_failing.txt')
         for f in failures:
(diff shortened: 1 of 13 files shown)


ci_tools/bot_interaction.py
```diff
@@ -217,13 +217,23 @@ def mark_as_ready(pr_id):
 
     job_data = [j for j in data if j['name'] not in ('Bot', 'CleanUpBot')]
 
-    failures = [j['name'] for j in job_data if j['conclusion'] in ('cancelled', 'failed')]
+    failures = [j.get('name',"") for j in job_data if j['conclusion'] in ('cancelled', 'failure')]
 
     if any(j['name'] == 'Codacy Static Code Analysis' and j['conclusion'] in ('FAILURE', 'ACTION_REQUIRED') for j in running_job_data):
```
yguclu, line 222: When you created the list `job_data` just above, you used `j['name']`, therefore every `j` dictionary in the list has the key `name`. (If not, a `KeyError` exception would have been raised.) So why do you have to use `j.get('name', "")` to build the list `failures`?

ci_tools/check_new_coverage.py
```diff
@@ -20,7 +20,18 @@
...
+
+print(new_untested)
+
+new_untested = cov.allow_untested_error_calls(new_untested)
+
+print(new_untested)
+
+new_untested = cov.allow_untested_debug_code(new_untested)
+
+print(new_untested)
```
yguclu, line 34: Is this debug code, or do you want to print to terminal for the user?

yguclu (changes requested):

ci_tools/check_new_coverage.py
EmilyBourne: You are right, this is debug code. It has been removed: fb603b7

ci_tools/bot_interaction.py
EmilyBourne: I made this change as I saw a `KeyError` in the output of a test at one point (I'm not sure exactly what leads to a nameless test). I must have got the lines mixed up. I have fixed it : [771eed8](https://github.com/pyccel/pyccel/pull/1369/commits/771eed8fd8fc09f0d89790fa347c65552ab3db16)

yguclu (approved):
Nice!

Issue #1370, opened 2023-04-26
Pip optional argument fails with zsh


EmilyBourne:
### Description

The command `pip3 install -e .[test]` fails in the environment with the error:
```
zsh: no matches found: .[test]
```
This error and the fix is reported here : https://github.com/mu-editor/mu/issues/852 . The doc should be updated to recommend the use of the more general command `pip3 install -e ".[test]"`

### Environment information

Python version : 3.9

OS version : Ubuntu 11.3

Conda : 21.1.0

cc: @mkaddani

PR #1374, opened 2023-04-27
Fix installation instructions

mkaddani:
updating the Readme.md as mentioned in #1370 to fix Pip optional argument fails with zsh

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -19,3 +19,4 @@ Contributors
 * Abdechahid Ihya
 * Youness Farini
 * Mohamed Jalal Maaouni
+* Mohammed Kaddani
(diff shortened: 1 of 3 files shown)


EmilyBourne:
@mkaddani I have cancelled your tests. They will be rerun when you press the "Ready for review" button (or ask the bot to mark as ready). As you haven't changed any code I don't see any reason to run these before deciding that the PR is ready

mkaddani:
@EmilyBourne ah okey, i was following (exploring) the bot instructions.
Thank you!

EmilyBourne:
> Yes, that makes sense. We should probably expand the bot description here : https://github.com/pyccel/pyccel/blob/master/ci_tools/bot_messages/bot_commands.txt so it says that these tests are run. Feel free to do that on this branch. It is also docs

I will fix this in #1381 . Please can you take a look and let me know if you think the messages are clearer

EniddeallA (approved):
Looks good!

EmilyBourne (approved):

PR #1380, opened 2023-05-02
Ensure temporary arrays created for function calls are correctly declared in loops.

### LINKED ISSUE #1379: absolute value of difference between two arrays is wrong when inside a loop
### Describe the bug
if the absolute value of the difference between two arrays is taken inside a loop, the value is wrong

### To Reproduce
```python
    ii_a  = np.zeros(2)
    ii_b  = np.zeros(2)
    ii_c  = np.zeros(2)
    ii_d1 = np.zeros(2)
    ii_d2 = np.zeros(2)
    
    ii_a[:] = 0

    for m in range(1):
    # if True:
        for d in range(2):
            ii_b[d] = d
        ii_c[:] = ii_b - ii_a
        ii_d1[:] = np.abs(ii_c)
        ii_d2[:] = np.abs(ii_b - ii_a)
        print("ii_d1 = ", ii_d1)
        print("ii_d2 = ", ii_d2) 
```

### Error details
```sh
ii_d1 =  [.000000000000 1.000000000000]
ii_d2 =  [.000000000000 .000000000000]
```

### Expected behavior
We would expect 
```sh
ii_d1 =  [.000000000000 1.000000000000]
ii_d2 =  [.000000000000 .000000000000]
```
Note: this good result is obtained when the loop is disabled
(commenting  `for m in range(1):` and uncommenting  `if True:`)

### Language
Fortran

### Additional context
the same problem occurs with arrays of integers

EmilyBourne:
When a function call takes an array as an argument and that array is the result of an arithmetic operation a temporary variable is created. The creation of new arrays is inserted before each line in a `CodeBlock`, however despite the fact that `For.body` contains a `CodeBlock` it didn't call `_visit_CodeBlock`. As a result the insertion of the line was done before the `For` block instead of before the relevant line. This PR fixes the problem by using `_visit_CodeBlock`. A test which fails on the master branch has been added.

Fixes #1379


yguclu:
I think that we should update the patch version with this PR. @EmilyBourne Do you agree, or do you prefer to wait?

EmilyBourne:
> I think that we should update the patch version with this PR. @EmilyBourne Do you agree, or do you prefer to wait?

That makes sense to me. The CHANGELOG seems quite full, and I don't see any imminent PRs worth waiting for. There is a bigger one in the pipeline but that will warrant a minor version change anyway. No need to mix in too much other stuff

yguclu (approved):

PR #1389, opened 2023-05-12
New lists implementation

MostafaMamouni:
This pull request concerns a new lists implementation that introduces a data structure that is similar in Cpython,
As you may know, every object in Python can be represented with PyObject object which is a generic type for python's object, here is the definition of the PyObject struct from the current Cpython:
```C
struct _object {
    _PyObject_HEAD_EXTRA
    union {
       Py_ssize_t ob_refcnt;
#if SIZEOF_VOID_P > 4
       PY_UINT32_T ob_refcnt_split[2];
#endif
    };
    PyTypeObject *ob_type;
};
```
In short, PyObject contains only the reference count and the type of the object, `_PyObject_HEAD_EXTRA` is for debugging purposes, for more details PyObject is defined [here](
https://github.com/python/cpython/blob/ce4eecf989e23fa26c033de78f1ca8b035a979cb/Include/object.h#L166) explained [here](https://github.com/python/cpython/blob/ce4eecf989e23fa26c033de78f1ca8b035a979cb/Include/object.h#L166) and renamed (its original name is _object) [here](https://github.com/python/cpython/blob/ce4eecf989e23fa26c033de78f1ca8b035a979cb/Include/pytypedefs.h#L18).

a list in Python is represented by the following struct [here](https://github.com/python/cpython/blob/ce4eecf989e23fa26c033de78f1ca8b035a979cb/Include/cpython/listobject.h#L5):
```C
typedef struct {
    PyObject_VAR_HEAD
    /* Vector of pointers to list elements.  list[0] is ob_item[0], etc. */
    PyObject **ob_item;

    /* ob_item contains space for 'allocated' elements.  The number
     * currently in use is ob_size.
     * Invariants:
     *     0 <= ob_size <= allocated
     *     len(list) == ob_size
     *     ob_item == NULL implies ob_size == allocated == 0
     * list.sort() temporarily sets allocated to -1 to detect mutations.
     *
     * Items must normally not be NULL, except during construction when
     * the list is not yet visible outside the function that builds it.
     */
    Py_ssize_t allocated;
} PyListObject;
```
`PyObject_VAR_HEAD` contains the number of items in a variable object (list in this case). [here](https://github.com/python/cpython/blob/main/Include/object.h#L180)

you can notice that manipulating lists in Cpython involves recursion and every item in the list can be any object including lists themselves. In the same way i suggest to make a similar approach with the following:
```C
typedef struct 
{
    size_t value;
    void *pointer;
    char *type;
}   GenericObject;
```
This struct can represent a single value of any type (to be supported) or any other objects (lists for now), So it is not that abstracting but it serves the goal. In Cpython you can get the actual value of an`int` object through `PyObject` casted to `PyIntObject`
```C
typedef struct {
    PyObject_HEAD
    long ob_ival;
} PyIntObject;
```
and then ob_ival.

for lists i suggest the following:
```C
typedef struct
{
    GenericObject **items;
    size_t noi;
    size_t capacity;
}   PyccelList;
```
With this approach, we have more convenience in converting PyListObject's to PyccelList's and vice versa, plus more readability considering the simplicity of the two structs, For performance, We can always enhance our functions to parallel operation on `GenericObject **items` that is in addition to the C code, I'm also thinking of making the conversion in wrapping parallel.

NOTE: the C code (lib, C wrapping) can be tested standalone for now.

what to be merged:
- [ ] basic functions for handling the data structure (allocating, freeing, filling).
- [ ] append function.
- [ ] the C wrapping of the append function.
- [ ] add `list` as a new supported dtype by pyccel.

### DIFF
--- a/pyccel/stdlib/lists/lists.c
+++ b/pyccel/stdlib/lists/lists.c
@@ -82,6 +82,7 @@ void    free_genericobject(GenericObject *go)
     free(go);
 }
 
+/*recursively copy all the items in src into dest*/
 int pyccel_copylist(PyccelList **src, PyccelList **dest)
 {
     for (size_t i = 0; i < (*src)->noi; i++)
@@ -108,6 +109,7 @@ int pyccel_expandlist(PyccelList **list)
     if ((*list) == NULL)
         return 1;
 
+    //initialize_list will double the given number
     newlist = initialise_list((*list)->capacity);
     pyccel_copylist(list, &newlist);
     free_list(*list);
--- a/pyccel/stdlib/lists/lists.c
+++ b/pyccel/stdlib/lists/lists.c
@@ -1,0 +1,136 @@
+#include "lists.h"
+
+PyccelList *initialise_list(size_t noi)
+{
+    PyccelList *list;
+
+    list = (PyccelList *)malloc(sizeof(PyccelList));
+    list->capacity = noi*2;
+    list->items = (GenericObject **)malloc(sizeof(GenericObject*)*list->capacity);
+    
+    for (size_t i = 0; i < list->capacity; i++)
+    {
+        list->items[i] = (GenericObject *)malloc(sizeof(GenericObject));
+        initialise_genericobject(&(list->items[i]), "v", 0, NULL);
+    }
+
+    list->noi = noi;
+    return list;
+}
+
+void    initialise_genericobject(GenericObject **obj, char  *type, size_t value, void *pointer)
+{
+    (*obj)->type = type;
+    (*obj)->pointer = pointer;
+    (*obj)->value = value;
+}
+
+void print_list(PyccelList *list)
+{
+    size_t i = 0;
+    write(1, "[", 1);
+    while (i < list->noi)
+    {
+        if (!strcmp(list->items[i]->type, "pl"))
+        {
+            print_list(list->items[i]->pointer);
+        }
+        else
+        {
+            printf("%zu", list->items[i]->value);
+            fflush(stdout);
+        }
+        i++;
+        if (i < list->noi)
+            write(1, ",", 1);
+    }
+    write(1, "]", 1);
+}
+
+void fill_itemswithvalues(PyccelList **list, size_t value)
+{
+    size_t i = 0;
+
+    while (i < (*list)->noi)
+    {
+        (*list)->items[i]->value = value;
+        (*list)->items[i]->type = "v";
+        i++;
+    }
+}
+
+void    free_list(PyccelList *list)
+{
+    for (size_t i = 0; i < list->capacity; i++)
+    {
+        if (!strcmp((list)->items[i]->type, "pl"))
+        {
+            free_list(list->items[i]->pointer);
+            free_genericobject(list->items[i]);
+        }
+        else
+        {
+            free_genericobject(list->items[i]);
+        }
+    }
+    free(list->items);
+    free(list);
+}
+
+void    free_genericobject(GenericObject *go)
+{
+    free(go);
+}
+
+int pyccel_copylist(PyccelList **src, PyccelList **dest)
+{
+    for (size_t i = 0; i < (*src)->noi; i++)
+    {
+        if (!strcmp((*src)->items[i]->type, "pl"))
+        {
+            PyccelList *item = (PyccelList *)((*src)->items[i]->pointer);
+            (*dest)->items[i]->pointer = initialise_list(item->noi);
+            (*dest)->items[i]->type = "pl";
+            pyccel_copylist(&item, (PyccelList **)(&((*dest)->items[i]->pointer)));
+        }
+        else
+        {
+            initialise_genericobject(&((*dest)->items[i]), (*src)->items[i]->type, (*src)->items[i]->value, (*src)->items[i]->pointer);
+        }
+    }
+    return 0;
+}
+
+int pyccel_expandlist(PyccelList **list)
+{
+    PyccelList *newlist;
+
+    if ((*list) == NULL)
+        return 1;
+
+    newlist = initialise_list((*list)->capacity);
+    pyccel_copylist(list, &newlist);
+    free_list(*list);
+    *list = newlist;
+
+    return 0;
+}
+
+int pyccel_append(PyccelList **list, size_t value, void *pointer)
+{
+    if (list == NULL)
+        return 1;
+    GenericObject *item_to_fill;
+
+    if ((*list)->capacity == (*list)->noi)
+        pyccel_expandlist(list);
+
+    item_to_fill = (*list)->items[(*list)->noi];
+    if (pointer != NULL)
+        initialise_genericobject(&item_to_fill, "pl", 0, pointer);
+    else
+        initialise_genericobject(&item_to_fill, "v", value, NULL);
+
+    (*list)->noi++;
+    return 0;
+}
--- a/pyccel/stdlib/lists/lists.h
+++ b/pyccel/stdlib/lists/lists.h
@@ -1,0 +1,37 @@
+#ifndef LISTS_H
+#define LISTS_H
+
+#include <stdlib.h>
+#include <Python.h>
+#include <stdio.h>
+#include <strings.h>
+#include <unistd.h>
+
+typedef struct 
+{
+    size_t value;
+    void *pointer;
+    char *type;
+}   GenericObject;
+
+typedef struct
+{
+    GenericObject **items;
+    size_t noi;
+    size_t capacity;
+}   PyccelList;
+
+PyccelList *initialise_list(size_t noi);
+void    initialise_genericobject(GenericObject **obj, char  *type, size_t value, void *pointer);
+
+int     pyccel_append(PyccelList **list, size_t value, void *pointer);
+int     pyccel_expandlist(PyccelList **list);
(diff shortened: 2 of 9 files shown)


EmilyBourne:
Am I correct that this is an implementation for non-homogeneous lists? Do you plan to have a separate implementation for homogeneous lists? As discussed here : https://github.com/pyccel/pyccel/discussions/1183#discussioncomment-3423209 and here : https://github.com/pyccel/pyccel/discussions/1183#discussioncomment-3424198 for the HPC applications that Pyccel targets homogeneous lists are **much** more important. Further the can be optimised much more.

As a follow-on question, how do you plan to handle Fortran? As discussed here : https://github.com/pyccel/pyccel/discussions/1183#discussioncomment-3423234 we should try and implement something which can be supported in all target languages. Having different docs for each language is unmanageable long term

EmilyBourne:
Also, how would you handle the following code:
```
a = [1, 2.0, 'a']
i = np.random.randint(2)
b = 3+a[i]
```
?
What is the type of `b`? I don't think there is any way to know at compile time

MostafaMamouni:
> Am I correct that this is an implementation for non-homogeneous lists? Do you plan to have a separate implementation for homogeneous lists? As discussed here : [#1183 (comment)](https://github.com/pyccel/pyccel/discussions/1183#discussioncomment-3423209) and here : [#1183 (comment)](https://github.com/pyccel/pyccel/discussions/1183#discussioncomment-3424198) for the HPC applications that Pyccel targets homogeneous lists are **much** more important. Further the can be optimised much more.
> 
> As a follow-on question, how do you plan to handle Fortran? As discussed here : [#1183 (comment)](https://github.com/pyccel/pyccel/discussions/1183#discussioncomment-3423234) we should try and implement something which can be supported in all target languages. Having different docs for each language is unmanageable long term

isn't homogeneous lists will be basicaly arrays, we are already working with lists as array in pyccel right now,
if we want to work only with homogeneous lists, why not just add missing functionalities from lists like append, pop etc ... to
ndarray in pyccel. but if we want to support lists as a data structure  i don't think homogeneous is the right approach, as far as i know lists are meant to be used for heterogeneous usage and arrays for homogeneous.

EmilyBourne:
> isn't homogeneous lists will be basicaly arrays, we are already working with lists as array in pyccel right now,

Not quite the main difference between the two is the use of append, pop, insert, etc

> if we want to work only with homogeneous lists, why not just add missing functionalities from lists like append, pop etc ... to
ndarray in pyccel.

We should not add these to ndarray. In order to optimise for append, insert etc additional memory should be allocated beyond what is strictly necessary. This is not the case for ndarrays.

> but if we want to support lists as a data structure i don't think homogeneous is the right approach, as far as i know lists are meant to be used for heterogeneous usage and arrays for homogeneous.

I disagree. In my experience, especially in HPC applications we have the following cases:
1. Fixed size homogeneous arrays
    - `np.ndarray` in Python
    - `std::array` in C++
    - `ndarray_t` in Pyccel generated C
    - `allocatable` in Fortran
2. Variable sized homogeneous arrays
    - `list` in Python
    - `std::vector` in C++
    - `t_list` (see #1192) in Pyccel generated C
    - does not exist in Fortran?
3. Small inhomogeneous arrays
    - `tuple` in Python
    - `std::tuple` in C++
    - `InhomogeousTupleVariable` in Pyccel

That just leaves the final case which you are aiming at here:
4. Large inhomogeneous arrays

I don't know of a use case for this in HPC applications which makes it harder to determine the right approach. It is telling that this is also the only case for which C++ doesn't have a structure.

As a corner case, case (4) should be treated last and we definitely shouldn't limit our optimisations for the other 3 use cases simply in order to allow it.
If you can work out a solution to the code typing problem I mentioned above (other than throwing out all optimisations and just writing a new Python interpreter where no types are known) then we can discuss handling this case but I would rather #1192 was handled first as I suspect this implementation will not be significantly faster than pure Python anyway.

To summarise:

Case 1) optimises for fast array access and minimises memory

Case 2) optimises for speed of appending/popping elements

Case 3) is small enough to not require optimisations

Case 4) is extremely difficult to optimise for, Python (which is also written in C) already does a very good job at this

There are other cases which may also need other structures, eg:

Case 5) optimises for random access insertion (linked lists, see `collections` in Python and `std::list` in C++)

Case 6) optimises for access to the last element (queue, see `collections` in Python and `std::queue` in C++)

Etc

I think that the case that even the C++ committee doesn't think can be done easily should be one of the last developments that we tackle. If ever.

MostafaMamouni:
> 2. Variable sized homogeneous arrays
>    
>    * `list` in Python
>    * `std::vector` in C++
>    * `t_list` (see [Lists implementation #1192](https://github.com/pyccel/pyccel/pull/1192)) in Pyccel generated C
>    * does not exist in Fortran?

are you sure `t_list` and `list` in Python are the same, they have different data structures and thier manupilation is very different from eachother, `list` in Python is not strictly homogeneous, even if it some times can look homogeneous (eg [1,2,3]) you won't get the expected performance due to
- type checking, every element in the list is PyObject that will ba casted to PyIntObject (not sure about the correct the name) after the type checking and then you get the actual value.
- copying data, which is a crucial difference between a true homogeneous (using just memcpy) and `list` in python that should iterate on that array of `PyObject`'s to copy every attribute (`type`, `size`, X`Object`) of every PyObject to the destination (another `list`), we can optimize that for `PyccelList` by using memcpy between `GenericOjbect`s since we don't have that casting as in Python (from `PyObject` to X`Object`, and intruducing parallelism 🤔.

In conclusion, `list` in Python can be either homogeneous or heterogeneous, and i don't think you can get homogeneity perks due to it's complex data structure.

`t_list` in the other hand, comes in between, the user can't use it for heavy computing because there are better alternatives and can't use it as lists in Python because it's homogeneous.

I think the user should expect Pyccel to handle lists **almost** exactly as Python handle them but faster ("slightly" is a win 😅 for me and why not using parallelism). and i tried to make PyccelList mimic Python list.

>Also, how would you handle the following code:
>
>```a = [1, 2.0, 'a']
>i = np.random.randint(2)
>b = 3+a[i]
>```
>?
>What is the type of b? I don't think there is any way to know at compile time
For this example, i don't think strings will be supported or even single characters so the list would be rather something like `[1,[2,3],[4[5]],]` in this case, if `a[i]` is a list the C code will fail at the compiling, C code would look like
```C
//code to initialise `a` PyccelList
int b = 3 + (PyccelList *)(a->items[i]->pointer);
```
or cleaner
```C
//code to initialise `a` PyccelList
PyccelList *c = (PyccelList *)a->items[i]-pointer;
int b = 3 + c;
```
For `a[i]` as a value (`a[i]->items->value`), i guess `b` would will take the biggest type whenever `PyccelList` is in an expression.

MostafaMamouni:
> Also, how would you handle the following code:
> 
> ```
> a = [1, 2.0, 'a']
> i = np.random.randint(2)
> b = 3+a[i]
> ```
> 
> ? What is the type of `b`? I don't think there is any way to know at compile time

😅 and there is no way to print `a[i]`, because it can be this `a->items[i]->pointer` or this `a->items[i]->value`, I'm thinking hard of using macros and pre-processors, but that is outside the execution.

EmilyBourne:
> are you sure `t_list` and `list` in Python are the same

No they're not the same, but neither is a Python `int` and a C `int`. Remember, Pyccel doesn't aim to reproduce everything in Python (Python already does that well itself), it just aims to translate a sub-set which can be expressed sufficiently well to get a large speed up. Types are very important for this.

> they have different data structures and thier manupilation is very different from eachother, `list` in Python is not strictly homogeneous, even if it some times can look homogeneous (eg [1,2,3]) you won't get the expected performance due to
> 
> * type checking, every element in the list is PyObject that will ba casted to PyIntObject (not sure about the correct the name) after the type checking and then you get the actual value.
> * copying data, which is a crucial difference between a true homogeneous (using just memcpy) and `list` in python that should iterate on that array of `PyObject`'s to copy every attribute (`type`, `size`, X`Object`) of every PyObject to the destination (another `list`), we can optimize that for `PyccelList` by using memcpy between `GenericOjbect`s since we don't have that casting as in Python (from `PyObject` to X`Object`, and intruducing parallelism thinking.
> 
> In conclusion, `list` in Python can be either homogeneous or heterogeneous, and i don't think you can get homogeneity perks due to it's complex data structure.

I agree. That is why the subset that should be acceptable for a program that can be translated by Pyccel only contains homogeneous lists where we can get these perks and therefore get a reasonable speed up

> `t_list` in the other hand, comes in between, the user can't use it for heavy computing because there are better alternatives and can't use it as lists in Python because it's homogeneous.

They can use it like a homogeneous list in Python, which is the main use case. I agree it shouldn't be used for heavy computing (just like `std::vector`). However some algorithms need variable array lengths.

> I think the user should expect Pyccel to handle lists **almost** exactly as Python handle them but faster ("slightly" is a win sweat_smile for me and why not using parallelism). and i tried to make PyccelList mimic Python list.

I may have been being generous when I said "slightly" :sweat_smile: . Python does a very good job at handling unknown types. I would also be surprised if it doesn't use parallelism under the hood for things like list comprehensions. You would have to prove to me that your implementation is faster than Python. As it is a very similar implementation I would expect it to have a very similar speed. Probably slower because it hasn't been optimised by the 10 or so extremely expert developers that optimise Python

PR #1395, opened 2023-05-17
Enforce checklist completion

EmilyBourne:
Multiple devs have forgotten to complete their checklists before marking the PRs as ready for review. As these checklists are especially important for new contributors and can sometimes be lost it is better to enforce their completion. This PR modifies the bot so that it refuses to mark the PR as ready if the list is not completed and leaves a message telling the user to check the list. This PR has been tested here : https://github.com/EmilyBourne/pyccel/pull/20

An additional improvement was made so that the reopen trigger acts the same as the open trigger.

_Future improvements_
In the future I would like to improve the bot in a few ways using the github api. This will allow me to interact with the PRs more. When this is in place it will be possible to remove all unchecked boxes and reprint them in the new message. This will make it easier to keep track. However currently it is hard to distinguish duplicate checklists and I can't delete old messages so a simple message asking the dev to find the checklist will suffice

### DIFF
--- a/ci_tools/bot_interaction.py
+++ b/ci_tools/bot_interaction.py
@@ -529,7 +529,7 @@ if __name__ == '__main__':
         else:
             leave_comment(pr_id, message_from_file('bot_commands.txt'))
 
-    elif event['action'] == 'opened':
+    elif event['action'] in ('opened', 'reopened'):
         # If new PR (opened trigger)
 
         # Collect id from a pull request event with an opened action
--- a/ci_tools/bot_interaction.py
+++ b/ci_tools/bot_interaction.py
@@ -337,9 +337,18 @@ def start_review_check(pr_id, event, outputs):
     if len(words) < 3:
         leave_comment(pr_id, message_from_file('set_draft_no_description.txt'))
         set_draft(pr_id)
-    else:
-        outputs['cleanup_trigger'] = 'request_review_status'
-        run_tests(pr_id, ['pr_tests'], outputs, event)
+        return
+
+    comments = get_previous_pr_comments(pr_id)
+    checklist = [c for c in comments if c.author == 'github-actions' and '- [ ]' in c.body]
+    print(checklist)
+    if checklist:
+        leave_comment(pr_id, message_from_file('set_draft_checklist_incomplete.txt'))
+        set_draft(pr_id)
+        return
+
+    outputs['cleanup_trigger'] = 'request_review_status'
+    run_tests(pr_id, ['pr_tests'], outputs, event)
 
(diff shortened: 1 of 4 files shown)


ci_tools/bot_interaction.py
```diff
@@ -520,7 +528,7 @@ def flagged_as_trusted(pr_id, user):
         else:
             leave_comment(pr_id, message_from_file('bot_commands.txt'))
 
-    elif event['action'] == 'opened':
+    elif event['action'] in ('opened', 'reopened'):
```
bauom, line 531: what does `reopened` do here
is it when you go from `draft` to `open` ?

EmilyBourne: No, it's if you close and then reopen the PR. Sometimes the bot gets stuck when you open it. It's useful to have this as a solution to get the checklist if it didn't trigger properly

bauom: ah okey I see :+1:

ci_tools/git_evaluation_tools.py
```diff
@@ -423,6 +424,16 @@ def check_previous_contributions(repo, author):
 
     with subprocess.Popen(cmds, stdout=subprocess.PIPE) as p:
         result, err = p.communicate()
+        returncode = p.returncode
     print(err)
+    print(returncode)
+    if returncode:
+        while returncode:
+            time.sleep(10)
```
yguclu, line 432: Hi @EmilyBourne. Why is it necessary to wait 10 seconds here?

EmilyBourne: It only fails when there are too many requests. The github message asks you to wait "a few seconds" then try again

EmilyBourne: This is documented here
https://docs.github.com/en/rest/guides/best-practices-for-integrators?apiVersion=2022-11-28#dealing-with-secondary-rate-limits
Ideally it should be improved to collect the right wait time

yguclu: Isn't there the danger of getting stuck here if the system is down? Maybe we can return an error after a certain number of tries.

EmilyBourne: I agree. I have capped it at 10 tries ([3781a6d](https://github.com/pyccel/pyccel/pull/1395/commits/3781a6d62b9cd19a9214e09a057bfa2a96b788cf))

I have found an example of this failure:
https://github.com/EmilyBourne/pyccel/actions/runs/5001434467/jobs/8960109824

I don't seem to have access to the `Retry-after` value or the rate limit. I imagine 10 seconds should be sufficient

PR #1412, opened 2023-06-01
Fix Map, Enumerate and Zip in comprehension list

smhah:
-  Modify line 194 {(if rank == 0)...} to not assert error if shape == (), to consider it as None because of a runtime error.

- Fix Enumerate zip map in comprehension list using Iterate class used in normal for loops.

- Work with nested variable in indices variable to prevent conflict with tuple variables and variables used in nested loops.
    - For example if we have expr.indices = [(a,b), c] or expr.indices = [a,(b,c)].
      To fix this conflict I changed indices to a nested list.
      expr.indices = [(a,b), c] => indices = [[a,b], c]].
      But the problem is an error gets raised at the returning of the codeblock in FunctionalFor
      "Basic child cannot be a tuple of tuples"
       So at the end of FunctionalFor I changed the nested list indices variable to a normal list variable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ All notable changes to this project will be documented in this file.
 -   #1302 : Raise error message in case of empty class
 -   #929 : Allow optional variables when compiling with intel or nvidia.
 -   #1117 : Allow non-contiguous arrays to be passed to Fortran code.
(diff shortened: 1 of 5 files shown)


EmilyBourne:
Superceded by #1977

PR #1416, opened 2023-06-12
Fix AST user tree bug causing problem with AugAssign

### LINKED ISSUE #1415: Incorrect translation of iadd in combination with other function call
### Describe the bug
When using the iadd "+=" in combination with a function call, i.e. `x += fun(y)`, it is incorrectly translated with two assignments in one line. This causes the compiler to throw an `Unclassifiable statement` error.

### To Reproduce
```python
def fun1(x: 'float') -> 'float':
    return x + 1

def fun2():
    x = 0.5
    y = 0.3
    x += y
    x += fun1(y)
```

### Error details
The translated code reads
```fortran
module bug


        C_INT64_T
  implicit none

  contains

  function fun1(x) result(Out_0001)

    implicit none

    real(f64) :: Out_0001
    real(f64), value :: x

    Out_0001 = x + 1_i64
    return

  end function fun1

  subroutine fun2() 

    implicit none

    real(f64) :: x
    real(f64) :: y

    x = 0.5_f64
    y = 0.3_f64
    x = x + y
    x = x + x = fun1(y)


  end subroutine fun2

end module bug
```
Upon running the command `pyccel bug.py`, the translation succeed but, as mentioned above, the compiler throws an `Unclassifiable statement` error.

### Expected behavior
Expected translation would either be
```fortran
    x = x + fun1(y)
```
(I am not sure if this is possible in `fortran`) or by introducing a generated dummy variable:
```fortran
    dummy = fun1(y)
    x = x + dummy
```

### Language
Python to Fortran.

EmilyBourne:
When handling `AugAssign` objects temporary `Assign` objects are sometimes created. These nodes were not invalidated once they were deemed to be unnecessary. As a result during the printing of a `FunctionCall` the printer saw this parent here:
https://github.com/pyccel/pyccel/blob/d6fea05ce9e2e45d9dc2496f0a599e40f0b406e2/pyccel/codegen/printing/fcode.py#L2956
This leads to the function `f1(y)` being printed as `x = f1(y)`.

This PR corrects the AST user tree by reordering expressions and invalidating unused nodes. This fixes #1415 .

Tests are added to ensure that the AST user tree is correct and that the issue is fixed

### DIFF
--- a/pyccel/codegen/printing/fcode.py
+++ b/pyccel/codegen/printing/fcode.py
@@ -2066,6 +2066,7 @@ class FCodePrinter(CodePrinter):
 
     def _print_AugAssign(self, expr):
         new_expr = expr.to_basic_assign()
+        expr.invalidate_node()
         return self._print(new_expr)
 
     def _print_PythonRange(self, expr):
--- a/tests/ast/test_basic.py
+++ b/tests/ast/test_basic.py
@@ -20,13 +20,13 @@ def get_functions(filename):
     ast = pyccel.parse()
 
     # Assert syntactic success
-    assert(not errors.has_errors())
+    assert not errors.has_errors()
 
     settings = {}
     ast = pyccel.annotate(**settings)
 
     # Assert semantic success
-    assert(not errors.has_errors())
+    assert not errors.has_errors()
 
     return list(ast.scope.functions.values())
 
@@ -36,7 +36,7 @@ def test_get_attribute_nodes():
     fst = get_functions(filename)[0]
     atts = fst.get_attribute_nodes(Variable)
 
-    assert(all(isinstance(a, Variable) for a in atts))
+    assert all(isinstance(a, Variable) for a in atts)
 
     expected = [Variable('int', 'a'),
                 Variable('int', 'b'),
@@ -46,7 +46,7 @@ def test_get_attribute_nodes():
                 Variable('int', 'g')]
 
     for e in expected:
-        assert(e in atts)
+        assert e in atts
 
 def test_get_attribute_nodes_exclude():
     filename = os.path.join(path_dir, "math.py")
@@ -54,16 +54,16 @@ def test_get_attribute_nodes_exclude():
     fst = get_functions(filename)[0]
     atts = fst.get_attribute_nodes(PyccelOperator, excluded_nodes=(PyccelAdd,))
 
-    assert(all(isinstance(a, PyccelOperator) for a in atts))
-    assert(all(not isinstance(a, PyccelAdd) for a in atts))
-    assert(len(atts)==1)
+    assert all(isinstance(a, PyccelOperator) for a in atts)
+    assert all(not isinstance(a, PyccelAdd) for a in atts)
+    assert len(atts)==1
     minus = atts[0]
-    assert(isinstance(minus, PyccelMinus)==1)
+    assert isinstance(minus, PyccelMinus)==1
 
     a = Variable('int', 'a')
     b = Variable('int', 'b')
-    assert(minus.args[0] == a)
-    assert(minus.args[1] == b)
+    assert minus.args[0] == a
+    assert minus.args[1] == b
 
 def test_get_user_nodes():
     filename = os.path.join(path_dir, "math.py")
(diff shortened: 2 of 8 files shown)


mkaddani (approved):
Looks good!

tests/ast/test_basic.py
```diff
@@ -96,6 +96,23 @@ def test_get_user_nodes_excluded():
...
+    atts = set(fst.get_attribute_nodes(Variable))
+    atts = [v for v in atts  if v == interesting_var]
+
+    a_var = atts[0]
+
+    users = set(a_var.get_all_user_nodes())
+
+    assert(all(isinstance(s, (PyccelMul, PyccelMinus, AugAssign, FunctionDefArgument)) for s in users))
+
+    assert(len(users) == 4)
```
yguclu, line 114: Is there any hope that we could remove the parentheses from the `assert` statements in this file before it gets too long?

tests/epyccel/modules/augassign.py
```diff
@@ -153,3 +153,9 @@ def augassign_div_2d_complex(a):
     b = a
     b /= (4.0 + 2.0j)
     return b[0][0]
+
+def augassign_func(x : float, y : float):
+    def fun1(x: 'float') -> 'float':
+        return x + 1
+    x += fun1(y)
+    return x
```
yguclu, line 161: Instead of testing a function in function which does not work in C, could we not use a simple `numpy` or `math` function? I would also add a test with arrays if possible.

pyccel/codegen/printing/fcode.py
```diff
@@ -2066,6 +2066,7 @@ def _print_Continue(self, expr):
 
     def _print_AugAssign(self, expr):
         new_expr = expr.to_basic_assign()
+        expr.invalidate_node()
         return self._print(new_expr)
```
yguclu, line 2070: Should we not apply a similar fix to the C printer in the case where `op == '%'`?

tests/epyccel/modules/augassign.py
EmilyBourne: No. NumPy and math functions are translated to objects from the ast not `FunctionCall` objects so they won't illustrate this bug

yguclu: Ah, that's true. But still we should test the C printer somehow...

EmilyBourne: I'm not sure we can for this specific problem. But in any case the problem does not exist in C as there we can directly print `x += fun1(y)`. So as long as the semantic stage doesn't mess up the tree everything is ok. The semantic stage is tested in `tests/ast`. Beyond that we can test `AugAssign` operators, but that is already done in this file

tests/ast/test_basic.py
EmilyBourne: Done 4676de74

pyccel/codegen/printing/fcode.py
EmilyBourne: Done [ccfaba4](https://github.com/pyccel/pyccel/pull/1416/commits/ccfaba42ba09d6dd87ef0bfc30b8a364a93c330e)

tests/epyccel/modules/augassign.py
EmilyBourne: I have added an array test: 5529f911

EmilyBourne: If you think it is important to test an expression of the type `x += fun1(y)` in C. We could translate a module, but we either need to create a module for this in the module tests, or translate https://github.com/pyccel/pyccel/blob/devel/tests/epyccel/modules/augassign.py
but in that case this test will fail whenever any augassign test fails.

Personally I don't think it is worth it given that function calls are not a special case in C

EmilyBourne:
Untested code is related to symbolic assignment. The issue for this can be found here: https://github.com/pyccel/pyccel/issues/330

tests/ast/test_basic.py
yguclu: Thanks! 😃

tests/epyccel/modules/augassign.py
yguclu: I think you are probably right, testing that expression in C is not worth the effort.

yguclu (approved):

PR #1419, opened 2023-06-16
Fix Product iterable

### LINKED ISSUE #1418: Product doesn't work properly for non range targets
### Describe the bug
`itertools.product` doesn't work properly. It is implemented as a cross-product of ranges not a cross-product of `Variable` iterators.

### To Reproduce
```python
import numpy as np
from itertools import product
if __name__ == "__main__":
    x = np.ones(20)
    y = np.ones(20)
    for i, j in product( x, y ):
        print(i)
        print(j)
```

### Error details
The generated code is:
```fortran
program prog_prog_prod

  use prod

        C_INT64_T
  implicit none

  real(f64), allocatable :: x(:)
  real(f64), allocatable :: y(:)
  real(f64) :: i
  real(f64) :: j

  allocate(x(0:19_i64))
  x = 1.0_f64
  allocate(y(0:19_i64))
  y = 1.0_f64
  do j = 0_i64, size(y, kind=i64) - 1_i64, 1_i64
    do i = 0_i64, size(x, kind=i64) - 1_i64, 1_i64
      write(stdout, '(F0.12)', advance="yes") i
      write(stdout, '(F0.12)', advance="yes") j
    end do
  end do
  if (allocated(x)) then
    deallocate(x)
  end if
  if (allocated(y)) then
    deallocate(y)
  end if

end program prog_prog_prod
```

Which fails to compile with:
```
/home/emily/Code/pyccel/pyccel/codegen/compiling/compilers.py:477: UserWarning: /home/emily/Code/test/__pyccel__/prog_prod.f90:20:7:

   20 |     do i = 0_i64, size(x, kind=i64) - 1_i64, 1_i64
      |       1
Warning: Deleted feature: Loop variable at (1) must be integer
/home/emily/Code/test/__pyccel__/prog_prod.f90:19:5:

   19 |   do j = 0_i64, size(y, kind=i64) - 1_i64, 1_i64
      |     1
Warning: Deleted feature: Loop variable at (1) must be integer
```

### Expected behaviour
A dummy variable should be created for the loop and `i` and `j` and assigned to `x[dummy]` and `y[dummy_1]`

### Language
All

EmilyBourne:
The `Product` iterable created all the necessary `Variable`s. However the nested `For`s were created incorrectly. As a result the  implementation did not work (except for cases such as the one used in the tests where the arguments are all equivalent to the ranges which are generated). This PR fixes the creation of the nested `For`s necessary for `itertools.product` and adds a more complex test. Fixes #1418.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,16 @@ All notable changes to this project will be documented in this file.
 ## \[Unreleased\]
 
 ### Added
+
+### Fixed
+
+### Changed
+
+## \[1.8.0\] - 2023-06-20
+
+### Added
 -   #1400 : Added flags to Pyccel for managing conda PATH warnings.
+
 ### Fixed
 
 -   #1404 : Bug preventing printing of some functions in a `print()` call.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ All notable changes to this project will be documented in this file.
 -   #929 : Allow optional variables when compiling with intel or nvidia.
 -   #1117 : Allow non-contiguous arrays to be passed to Fortran code.
 -   #1415 : Fix incorrect handling of assignments augmented by function calls.
+-   #1418 : Fix `itertools.product` implementation.
 
 ### Changed
 
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -2742,7 +2742,6 @@ class SemanticParser(BasicParser):
                    bounding_box=(self._current_fst_node.lineno, self._current_fst_node.col_offset),
                    severity='error')
 
-
         body = self._visit(expr.body)
 
         self.exit_loop_scope()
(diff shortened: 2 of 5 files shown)


smhah:
You said this PR fixes the creation of a nested For, there is no test containing the creation of a nested For.

EmilyBourne:
> You said this PR fixes the creation of a nested For, there is no test containing the creation of a nested For.

The nested `for` is created here:
https://github.com/pyccel/pyccel/blob/9000a8bac964864d93c01a084bded6a0d28be212/pyccel/parser/semantic.py#L2750-L2762

An `itertools.product` *is* a nested loop. It is tested here:
https://github.com/pyccel/pyccel/blob/9000a8bac964864d93c01a084bded6a0d28be212/tests/epyccel/test_loops.py#L103-L113

The generated code for:
https://github.com/pyccel/pyccel/blob/9000a8bac964864d93c01a084bded6a0d28be212/tests/epyccel/modules/loops.py#L106-L109
when translating to Python is:
```python
    k = 0
    for i in x:
        for j in y:
            z[k] = i - j
            k += 1

```
where you can clearly see the nested loops.

smhah (approved):

yguclu (approved):
Excellent!

I think we can update the package (minor) version. What do you think?

EmilyBourne:
@yguclu I agree. The update to the wrapping definitely warrants a minor version and the changelog shows that we have quite a few fixes in the `devel` branch. I have updated the version to 1.8.0 and put the release date in the changelog.

yguclu (approved):

PR #1431, opened 2023-07-06
Fix generic methods in classes

### LINKED ISSUE #1429: Pyccelisation of `numpy.conjugate`
### Describe the bug
When we use the `numpy.conjugate()` on a real value we get an error because the fortran version of `conjugate` don't accept real values as entries.

### To Reproduce
```python
from numpy import conjugate as np_conj
@template(name='T', types=['float', 'complex'])
def fun(val:'T'):
    return np_conj(val)
```

### Error details
Generated Code :
```bash
module main


        C_DOUBLE , c64 => C_DOUBLE_COMPLEX
  implicit none

  interface fun
    module procedure fun_00
    module procedure fun_01
  end interface

  contains

  function fun_00(val) result(Out_0001)

    implicit none

    complex(c64) :: Out_0001
    real(f64), value :: val

    Out_0001 = conjg( val )
    return

  end function fun_00

  function fun_01(val) result(Out_0001)

    implicit none

    complex(c64) :: Out_0001
    complex(c64), value :: val

    Out_0001 = conjg( val )
    return

  end function fun_01

end module main

```
Error:
```bash
/home/cato/Code/temp/__pyccel__/main.f90:23:22:

   23 |     Out_0001 = conjg( val )
      |                      1
Error: ‘z’ argument of ‘conjg’ intrinsic at (1) must be COMPLEX


```

### Expected behavior
If the `conjugate` function is applied to a real value, it should maybe be removed or at least allow real values in entries to behave like the numpy function.

### LINKED ISSUE #1430: Pyccelisation of python scalar method `conjugate()`
### Describe the bug
The `float` python type have a method `.conjugate()`.
When trying to pyccelise a file containing it, pyccel have a fatal error.

### To Reproduce
Try pyccelise this function
```python
def fun(val:'float'):
    return val.conjugate()
```
### Error details
```bash/fortran/c
pyccel:
 |fatal [semantic]: main.py [3,11]| Attribute conjugate not found
```
### Expected behavior
The conjugate method should work on a `float` or pyccel should remove it when it is applied to a `float`.

EmilyBourne:
Tidy up the handling of class methods to ensure inherited functions are correctly handled (e.g. all boolean functions inherited from integers). In order to do this cleanly the assignation and retrieval of class data types are cleaned up.

Urgent as blocking #1441 

**Commit summary**
- Fix string base class selection
- Add decorators to `PyccelFunctionDef` to allow them to be indicated as class properties
- Use `PyccelFunctionDef` to define class methods for int, complex, etc. This allows the methods to be used like normal functions via `_handle_function`
- Add `get_method` to `ClassDef` to return function/interface with requested name
- Clean up `CustomDataType` and `DataTypeFactory` to ensure created classes have clear type names
- Reorder `FCodePrinter._print_Module` to ensure classes are defined before module variables
- Remove `CustomDataType` options `is_iterable`, `is_polymorphic` and `is_with_construct` which are not currently helpful and should be handled via magic methods
- Add `self` as a `FunctionCallArgument`. This will help with printing the calls in c and ensures that the `FunctionCall` arguments match the `FunctionDef` arguments
- Simplify `_visit_DottedName` to use `_handle_function` for all `FunctionCall`s. This fixes `Interface`s in classes as they are no longer a special case.
- Ensure `superclass` contains `ClassDef` objects
- Ignore `object` in inheritance
- Add tests for boolean `conjugate`
- Move `generic_methods.py` from `known_bugs` to `tests/pyccel`
- Move working `concatenation.py` from `tests/codegen` to `tests/pyccel`
- Move `context.py` from `tests/codegen` to `known_bugs`
- Fix Python printing of expressions such as `p.f() + 3`
- Add printing of `Interface`s to `PyCodePrinter._print_ClassDef`
- Remove `PyCodePrinter._print_DottedFunctionCall` and use `PyCodePrinter._print_FunctionCall` instead to avoid code duplication


tests/epyccel/test_class_expressions.py
```diff
@@ -90,6 +90,81 @@ def f(a : 'complex64', b : 'complex64'):
...
+
+    r = f(a,b)
+    epyc_r = epyc_f(a,b)
+
+    assert r == epyc_r
+    assert isinstance(r, type(epyc_r))
+
+def test_bool_conjugate(language):
+    def f(a : 'bool', b : 'bool'):
+        return (a+b).conjugate()
```
yguclu, line 155: This test is not checking `bool` , because `a+b` is an `int`.

EmilyBourne: Yes I know, I have fixed it but the bool case needed special treatment to handle integers. It is especially tricky because bool is a sub-class of Int and sub-classes haven't been properly dealt with up until now

EmilyBourne: I have fixed it I am just cleaning up and testing as my fix ended up fixing a few other problems. Can I force push my changes?

yguclu: Sorry  I have just read your messages now. I have the browser open and it does not refresh automatically :sweat_smile:

yguclu: I stop touching your branch now and let you work :wink:

EmilyBourne:
The coverage issues are due to error testing for class inheritance which is currently only partially implemented

https://github.com/pyccel/pyccel/blob/5ae996e6ed9dccd8db5a8432ad309bfc6337a1a8/pyccel/parser/semantic.py#L3414-L3418

https://github.com/pyccel/pyccel/blob/5ae996e6ed9dccd8db5a8432ad309bfc6337a1a8/pyccel/parser/semantic.py#L3499-L3504

sboof911 (approved):
Everything appears to be in order and looks good.

tests/pyccel/scripts/concatenation.py
yguclu, line 1: What is the connection between this PR and list concatenation?

pyccel/parser/semantic.py
```diff
@@ -3498,16 +3491,24 @@ def _visit_ClassDef(self, expr):
...
         name = expr.name
+        # remove quotes for str representation
         name = name.replace("'", '')
+
+        parent = [self.scope.find(s, 'classes') for s in expr.superclass]
+        if any(s is None for s in parent):
+            for s in parent:
+                if s is None:
+                    errors.report("Couldn't find class {s} in scope", symbol=expr,
+                            severity='error')
```
yguclu, line 3502: Are you not missing an `f` in front of the string?

Actually, even if you add it this will print `Couldn't find class None in scope`, which is not very useful.

yguclu (changes requested):
This is a lot of work! Do the new tests showcase all the new capabilities introduced by this PR?

tests/pyccel/scripts/concatenation.py
EmilyBourne: I don't think there is one. I just noticed it was working so I moved it to be a proper test

pyccel/parser/scope.py
```diff
@@ -340,8 +343,35 @@ def insert_class(self, cls):
         if self.is_loop:
             self.parent_scope.insert_class(cls)
         else:
-            #if name in self._locals['classes']:
-            #    raise RuntimeError('New class already exists in scope')
+            if name in self._locals['classes']:
+                raise RuntimeError('New class already exists in scope')
```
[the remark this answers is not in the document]
EmilyBourne: Fixed 580c39d3f6ee93ea6a553564e988c11aa4ca559d

pyccel/ast/datatypes.py
```diff
@@ -263,10 +285,7 @@ def __init__(self, **kwargs):
                     {"__init__":          __init__,
                      "_name":             name,
                      "prefix":            prefix,
-                     "alias":             alias,
-                     "is_iterable":       is_iterable,
-                     "is_with_construct": is_with_construct,
-                     "is_polymorphic":    is_polymorphic})
+                     "alias":             name})
```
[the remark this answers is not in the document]
EmilyBourne: Fixed 6c4f5947

pyccel/parser/semantic.py
EmilyBourne: Fixed 5cb4c09b

tests/pyccel/scripts/concatenation.py
yguclu: All right. Please add a line to the PR description?

pyccel/codegen/printing/fcode.py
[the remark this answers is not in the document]
EmilyBourne: - `_print_Declare` : class objects should not be marked with `bind(c)` as they will be exposed via pointers and may contain `allocatables` which cannot be marked as `bind(c)` without using Fortran 2018

- `_print_ClassDef` : The `\n` character was missing. This didn't cause problems before as module variables were defined before classes (meaning they couldn't be classes) and there was additional space between the declarations and the definitions

- `_print_FunctionCall` : Previously to this PR the arguments for a function call `p.f(a,b)` were `[a,b]` but the arguments for the equivalent function def were `[self, a, b]`. This causes problems when doing type checking etc. I therefore added `self` to the function call arguments as removing it from the function def would require the addition of new classes to handle class methods. Currently they are handled perfectly well with a simple `FunctionDef`

pyccel/parser/semantic.py
```diff
@@ -3498,16 +3491,24 @@ def _visit_ClassDef(self, expr):
         #      - wouldn't be better if it is done inside ClassDef?
 
         name = expr.name
+        # remove quotes for str representation
         name = name.replace("'", '')
+
+        parent = {s: self.scope.find(s, 'classes') for s in expr.superclasses}
+        if any(s is None for s in parent.values()):
```
yguclu, line 3498: For consistency here I would use the variable name `c` instead of `s`

pyccel/parser/semantic.py
```diff
@@ -3498,16 +3491,24 @@ def _visit_ClassDef(self, expr):
...
+            for s,c in parent:
```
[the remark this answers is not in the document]
EmilyBourne: Oops. Fixed a156ea82

pyccel/parser/semantic.py
EmilyBourne: Done a156ea82

pyccel/parser/semantic.py
```diff
@@ -3423,8 +3409,15 @@ def _visit_FunctionDef(self, expr):
...
 
-                self.scope.insert_class(ClassDef(cls_name, cls.attributes,
-                        methods, superclass=cls.superclass))
+                superclasses = [self.scope.find(s, 'classes') for s in cls.superclasses]
+                if any(s is None for s in superclasses):
+                    for s in superclasses:
+                        if s is None:
+                            errors.report("Couldn't find class {s} in scope", symbol=expr,
+                                    severity='error')
+                    superclasses = [s for s in superclasses if s is not None]
```
yguclu, line 3418: This block of code also needs to be fixed like the one in `_visit_ClassDef`

pyccel/codegen/printing/fcode.py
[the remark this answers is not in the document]
yguclu: Thanks! Should any of this be added to the PR description?

pyccel/parser/semantic.py
EmilyBourne: Fixed duplication : 5c49581e

pyccel/parser/semantic.py
```diff
@@ -3498,16 +3491,24 @@ def _visit_ClassDef(self, expr):
...
+        # remove quotes for str representation
         name = name.replace("'", '')
+
+        parent = {s: self.scope.find(s, 'classes') for s in expr.superclasses}
+        if any(s is None for s in parent.values()):
+            for s,c in parent:
+                if c is None:
+                    errors.report(f"Couldn't find class {s} in scope", symbol=expr,
+                            severity='error')
+            parent = {s:c for s,c in parent if c is not None}
```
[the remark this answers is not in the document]
EmilyBourne: Fixed : ced6a2a3 3389fd5a

yguclu (approved):

yguclu (approved):

PR #1432, opened 2023-07-07
Add conjugate methods for ints and floats

EmilyBourne:
Add a `__new__` method to allow the use of `conjugate` and `conj` with integers and floats. Fixes https://github.com/pyccel/pyccel/issues/1430 and fixes https://github.com/pyccel/pyccel/issues/1429. Bools are not fixed as this requires corrections to the class inheritance handling. Update library version to 1.8.1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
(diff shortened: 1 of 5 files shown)


EmilyBourne:
Boolean case is not tested due to inheritance issues which will be fixed in #1431

yguclu (approved):

PR #1445, opened 2023-07-13
Fix race condition

EmilyBourne:
The race condition is occurring on Macosx because Pyccel occasionally thinks that the source files in the stdlib folder have been modified. This then creates the following series of concurrent events:

| Thread 1 | Thread 2 |
|--|---|
|Starts compilation process | Still translating |
| Acquires `ndarrays` folder lock | Still translating |
| Checks `ndarrays` folder | Starts compilation process |
| All ok  | Fails to acquire `ndarrays` folder lock. Waiting
| Releases `ndarrays` folder lock | Waiting |
| Acquires locks to compile `ndarrays.o` (`ndarrays.o.lock`) | Acquires `ndarrays` folder lock |
| Compilation in progress | Reads contents of `ndarrays` folder (may see `ndarrays.o.tmp` file created by compilation) |
| Finished compilation. Released locks | Checks access dates for `__pyccel__/ndarrays` folder and `stdlib/ndarrays` folder. **Incorrectly determines folder needs updating** |
| Prints wrapper file | Acquires locks to copy `ndarrays` folder  (`ndarrays.h.lock`, `ndarrays.c.lock`, `ndarrays.o.lock`) |
| Attempts to acquire locks to compile wrapper (including dependencies: `ndarrays.o`, (`ndarrays.h`, `ndarrays.c`) | deletes all files in `ndarrays` folder  |
| Waiting | Copies `ndarrays` folder contents back (only `ndarrays.h`, `ndarrays.c`)  |
| Waiting | Releases `ndarrays` locks |
|Acquires locks | Starts to check if compilation of `ndarrays` is necessary |
| Tries to compile generated file but `ndarrays.o` is missing | Tries to acquire `ndarrys.c.lock` |

The error is therefore not due to a race condition where 2 threads try to do the same thing, but rather due to a bad update condition. I think this is because we are checking when the files were accessed rather than when they were modified.

Additionally the compilation of `ndarrays.o` does not lock `ndarrays.c`. This is corrected in this PR but is not strictly necessary as the `ndarrays.o.lock` is always acquired if `ndarrays.c.lock` is acquired.

**Commit Summary**
- Make `CompileObj` a context manager (with methods `__enter__` and `__exit__`)
- Use `with compile_obj:` constructs to handle locks more compactly
- Ensure the source file is locked during compilation
- Check the modification time not the access time to decide whether a file should be updated
- Don't raise an error if temporary files are found in a folder being updated
- Add detailed flags which are activated when the tests are rerun with debug logging
image
- Improve the definition of the min/max bounds on a test that failed. The logic now only needs touching in a single place if the bounds need narrowing

**Testing**
It is quite difficult to test this kind of error, however I have run the full suite of tests with `assert False` in the code which updates the std libraries and macosx no longer seems to go into this section of the code. (Linux triggers the assertion in the [test designed to test this mechanism](https://github.com/pyccel/pyccel/blob/devel/.github/actions/pytest_run/action.yml#L20-L25))

### DIFF
--- a/pyccel/codegen/compiling/basic.py
+++ b/pyccel/codegen/compiling/basic.py
@@ -233,6 +233,7 @@ class CompileObj:
         Lock the file and its dependencies to prevent race conditions
         """
         self.acquire_simple_lock()
+        self._lock_source.acquire()
         for d in self.dependencies:
             d.acquire_simple_lock()
 
@@ -240,7 +241,6 @@ class CompileObj:
         """
         Lock the file to prevent race conditions but not its dependencies
         """
-        self._lock_source.acquire()
         if self.has_target_file:
             self._lock_target.acquire()
 
@@ -253,6 +253,7 @@ class CompileObj:
         """
         Unlock the file and its dependencies
         """
+        self._lock_source.release()
         self.release_simple_lock()
         for d in self.dependencies:
             d.release_simple_lock()
@@ -261,7 +262,6 @@ class CompileObj:
         """
         Unlock the file
         """
-        self._lock_source.release()
         if self.has_target_file:
             self._lock_target.release()
 
--- a/pyccel/codegen/compiling/basic.py
+++ b/pyccel/codegen/compiling/basic.py
@@ -221,6 +221,11 @@ class CompileObj:
             raise TypeError("Dependencies require necessary compile information")
         self._dependencies.update({a.module_target:a for a in args})
 
+    def __enter__(self):
+        self.acquire_simple_lock()
+        for d in self.dependencies:
+            d.acquire_simple_lock()
+
     def acquire_lock(self):
         """
         Lock the file and its dependencies to prevent race conditions
@@ -235,6 +240,11 @@ class CompileObj:
         """
         if self.has_target_file:
             self._lock.acquire()
+
+    def __exit__(self):
+        self.release_simple_lock()
+        for d in self.dependencies:
+            d.release_simple_lock()
 
     def release_lock(self):
         """
--- a/pyccel/codegen/compiling/basic.py
+++ b/pyccel/codegen/compiling/basic.py
@@ -224,9 +224,7 @@ class CompileObj:
         self._dependencies.update({a.module_target:a for a in args})
 
     def __enter__(self):
-        self.acquire_simple_lock()
-        for d in self.dependencies:
-            d.acquire_simple_lock()
+        self.acquire_lock()
 
     def acquire_lock(self):
         """
@@ -245,9 +243,7 @@ class CompileObj:
             self._lock_target.acquire()
 
     def __exit__(self, exc_type, value, traceback):
-        self.release_simple_lock()
-        for d in self.dependencies:
-            d.release_simple_lock()
+        self.release_lock()
 
     def release_lock(self):
         """
--- a/pyccel/codegen/compiling/basic.py
+++ b/pyccel/codegen/compiling/basic.py
@@ -54,6 +54,7 @@ class CompileObj:
         program. If no name is provided then the module name deduced from the file
         name is used.
     """
+    compilation_in_progress = FileLock('.lock_acquisition.lock')
     __slots__ = ('_file','_folder','_module_name','_module_target','_prog_target',
                  '_lock_target','_lock_source','_flags','_includes','_libs',
                  '_libdirs','_accelerators','_dependencies','_has_target_file')
@@ -224,14 +225,15 @@ class CompileObj:
         self._dependencies.update({a.module_target:a for a in args})
 
     def __enter__(self):
+        self.compilation_in_progress.acquire()
         self.acquire_lock()
 
     def acquire_lock(self):
         """
         Lock the file and its dependencies to prevent race conditions
         """
-        self.acquire_simple_lock()
         self._lock_source.acquire()
+        self.acquire_simple_lock()
         for d in self.dependencies:
             d.acquire_simple_lock()
 
@@ -244,15 +246,16 @@ class CompileObj:
 
     def __exit__(self, exc_type, value, traceback):
         self.release_lock()
+        self.compilation_in_progress.release()
 
     def release_lock(self):
         """
         Unlock the file and its dependencies
         """
-        self._lock_source.release()
-        self.release_simple_lock()
         for d in self.dependencies:
             d.release_simple_lock()
+        self._lock_source.release()
+        self.release_simple_lock()
 
     def release_simple_lock(self):
         """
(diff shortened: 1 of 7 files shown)


EmilyBourne:
Coverage problems are:
- https://github.com/pyccel/pyccel/blob/03e00f90858a6ec2aee4ba0b0542bf0bac9fdc18/pyccel/codegen/compiling/compilers.py#L25-L25
https://github.com/pyccel/pyccel/blob/03e00f90858a6ec2aee4ba0b0542bf0bac9fdc18/pyccel/codegen/compiling/compilers.py#L27-L27 minor changes to code only run on MacOS

- https://github.com/pyccel/pyccel/blob/03e00f90858a6ec2aee4ba0b0542bf0bac9fdc18/pyccel/codegen/utilities.py#L148-L152 easy error handling for an unwanted case (if it is unused, this is good news).

pyccel/codegen/compiling/basic.py
```diff
@@ -101,13 +103,26 @@ def __init__(self,
 
     def reset_folder(self, folder):
         """
-        Change the folder in which the source file is saved (useful for stdlib)
+        Change the folder in which the source file is saved.
+
+        Change the folder in which the source file is saved. Normally the location
+        of the source file should not change during the execution, however when
+        working with the stdlib, the `CompileObj` is created with the folder set
+        to the file's location in the Pyccel instal directory. When the file is
```
yguclu, line 111: Typo: instal -> install

pyccel/codegen/compiling/compilers.py
```diff
@@ -270,27 +269,30 @@ def _insert_prefix_to_list(lst, prefix):
 
     def _get_compile_components(self, compile_obj, accelerators = ()):
         """
-        Provide all components required for compiling
+        Provide all components required for compiling.
+
+        Provide all the different componenets (include directories, libraries, etc
+        which are needed in order to compile any file.
```
yguclu, line 275: Missing closing parenthesis

pyccel/codegen/compiling/compilers.py
```diff
@@ -270,27 +269,30 @@ def _insert_prefix_to_list(lst, prefix):
...
+        str
+            The command required to run the executable.
+        iterable of strs
+            The include directories required to compile.
+        iterable of strs
+            The libraries required to compile.
+        iterable of strs
+            The directories containing libraries required to compile.
+        iterable of strs
+            The objects required to compile.
```
yguclu, line 295: Given that we have 5 return variables, I find it much easier to read this docstring (and navigate the code) if we keep the variable names. Did Codacy complain about it?

pyccel/codegen/utilities.py
```diff
@@ -42,40 +42,64 @@
...
 def copy_internal_library(lib_folder, pyccel_dirpath, extra_files = None):
     """
+    Copy an internal library to the specified pyccel directory.
+
     Copy an internal library from its specified stdlib folder to the pyccel
     directory. The copy is only done if the files are not already present or
     if the files have changed since they were last copied. Extra files can be
     added to the folder if and when the copy occurs (e.g. for specifying
-    the numpy version compatibility)
+    the numpy version compatibility).
```
yguclu, line 85: In theory we should capitalize Pyccel and Numpy...

pyccel/codegen/utilities.py
```diff
@@ -140,30 +169,32 @@ def recompile_object(compile_obj,
...
-                  The compiler used
-    verbose     : bool
-                  Indicates whethere additional information should be printed
+        The object to compile.
+
+    compiler : str
+        The compiler used.
+
+    verbose : bool
+        Indicates whethere additional information should be printed.
```
yguclu, line 186: Typo: whethere -> whether

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5277,22 +5277,42 @@ def get_matmul(arr):
...
+        min_test = -np.sqrt(abs(min_for_type) / size[0])
+        max_test = np.sqrt(abs(max_for_type) / size[0])
+        print(min_test, max_test, cast(min_test), cast(max_test))
+        return cast(min_test), cast(max_test)
+
+    integer8 = randint(*calculate_max_values(min_int8, max_int8), size=size, dtype=np.int8)
+    integer16 = randint(*calculate_max_values(min_int16, max_int16), size=size, dtype=np.int16)
+    integer = randint(*calculate_max_values(min_int, max_int), size=size, dtype=int)
+    integer32 = randint(*calculate_max_values(min_int32, max_int32), size=size, dtype=np.int32)
+    integer64 = randint(*calculate_max_values(min_int64, max_int64), size=size, dtype=np.int64)
```
yguclu, line 5291: Can you please explain these changes in the PR description?

yguclu (changes requested):
I am so glad that you finally found the cause of the macOS failures!

The fix looks good and elegant to me. Great job!

I have a few minor comments.

pyccel/codegen/compiling/compilers.py
EmilyBourne: No, numpydoc does

tests/epyccel/recognised_functions/test_numpy_funcs.py
EmilyBourne: Oops. I didn't realise I'd committed that file ^^'

EmilyBourne: Done

pyccel/codegen/utilities.py
EmilyBourne: Fixed 45e7c0b9494cc4ca8d8b6a98c5c93c5a70d9cefd

pyccel/codegen/utilities.py
EmilyBourne: Fixed 45e7c0b9494cc4ca8d8b6a98c5c93c5a70d9cefd

pyccel/codegen/compiling/compilers.py
EmilyBourne: Never mind. I misremembered. It seems fine.
Fixed 892a6ad

pyccel/codegen/compiling/basic.py
EmilyBourne: Fixed 8104442

pyccel/codegen/compiling/compilers.py
EmilyBourne: Fixed 80d3415

yguclu (approved):

PR #1446, opened 2023-07-13
Adding Python support for a simple class

### LINKED ISSUE #1440: Add Python support for the simplest class possible
### Describe the feature
Add support for classes as simple objects containing scalars

### Test Code
```python
# pylint: disable=missing-class-docstring,  missing-function-docstring, missing-module-docstring
#$ header class Point(public)
#$ header method __init__(Point, double, double)

class Point:
    def __init__(self, x, y):
        self.x = x
        self.y = y

if __name__ == '__main__':
    p = Point(0.0, 0.0)
    x=p.x
    p.x=4
    a = p.x
    a = p.x - 2
    a = 2 * p.x - 2
    a = 2 * (p.x + 6) - 2

    p.y = a + 5
    p.y = p.x + 5

    print(p.x, p.y)
    print(x)
    print(a)
```

### Proposed Solution
The code should be printed as above.

### Target Language
Python

### Acceptance Criteria
To complete this issue the following should be done:
- [ ] Add test(s) similar to the one above in `tests/pyccel/scripts/classes`
    - [ ] Ensure that the tests are run and work for Fortran in `tests/pyccel/test_pyccel.py`
    - [ ] Ensure that the test can also be run for Python (should not pass yet)
- [ ] Implement `_print_ClassDef` to handle the subset of classes described above

sboof911:
Add new methods to the Python printer:

* `_print_ClassDef` prints a subset of classes. It ensures that class definitions are formatted correctly.
* `_print_DottedVariable` prints dotted variables.
* `_print_DottedFunctionCall` prints dotted function calls.
* `_print_ConstructorCall` prints object instantiation. It ensures that when objects are created using a constructor call (e.g., `ClassName(args)`), the instantiation is correctly represented.
* `_print_Del` prints object deletion. It ensures that when objects are deleted using the `del` keyword (e.g., `del ClassName`), the deletion is correctly displayed.

Activate some basic unit tests. Fix #1440.


EmilyBourne:
Don't forget to update your branch before running tests

sboof911:
/bot run linux coverage docs pylint lint

EmilyBourne:
/bot run linux coverage docs pylint lint

EmilyBourne:
I have cancelled the tests, updated the branch and retriggered the tests. If the branch is not up to date then the comparison between the source and the target branch doesn't work properly. This will be fixed in the new bot

mkaddani (approved):
looks good!

pyccel/codegen/printing/pycode.py
```diff
@@ -175,6 +175,14 @@ def _print_NativeComplex(self, expr):
     def _print_Variable(self, expr):
         return self._print(expr.name)
 
+    def _print_DottedVariable(self, expr):
+        rhs_code = self._print_Variable(expr)
+        lhs_code = self._print(expr.lhs)
+        return f"{lhs_code}.{rhs_code}"
+
+    def _print_DottedFunctionCall(self, expr):
+        return (f"{str(expr.func_name)}({', '.join([str(arg) for arg in expr.args])})\n")
```
EmilyBourne, line 184: `join` can take an iterable argument which avoids unnecessary memory allocation

```suggestion
        return (f"{str(expr.func_name)}({', '.join(str(arg) for arg in expr.args)})\n")
```

pyccel/codegen/printing/pycode.py
```diff
@@ -996,6 +1004,21 @@ def _print_PyccelSymbol(self, expr):
 
     def _print_PythonType(self, expr):
         return 'type({})'.format(self._print(expr.arg))
+    
+    #-----------------Class Printer---------------------------------
+
+    def _print_ClassDef(self, expr):
+        classDefName = 'class {}({}):'.format(expr.name,', '.join(str(arg) for arg in  expr.superclass))
+        methods = ''.join([self._print(method) for method in expr.methods])
```
EmilyBourne, line 1012: See above. `join` can take an iterable argument

pyccel/codegen/printing/pycode.py
EmilyBourne: You should use `self._print` to print objects, not `str()`

pyccel/codegen/printing/pycode.py
```diff
@@ -996,6 +1004,21 @@ def _print_PyccelSymbol(self, expr):
...
+        methods = self._indent_codestring(methods)
+        classDef = '\n'.join([classDefName, methods]) + '\n'
+        return classDef
+
+    def _print_ConstructorCall(self, expr):
+        return f"{expr.func.cls_name}({', '.join(str(arg) for arg in expr.arguments)})"
```
EmilyBourne, line 1018: See above. `str` should not be used

pyccel/codegen/printing/pycode.py
```diff
@@ -996,6 +1004,21 @@ def _print_PyccelSymbol(self, expr):
...
+
+    def _print_Del(self, expr):
+        return ''.join([f'del {var}\n' for var in expr.variables])
```
EmilyBourne, line 1021: See above. `join` takes iterable argument

EmilyBourne (changes requested):
This looks good. Great to finally improve the class support! I just have a couple of comments

EmilyBourne:
As per the [developer docs](https://github.com/pyccel/pyccel/blob/devel/developer_docs/review_process.md), please don't resolve conversations that you didn't open

pyccel/codegen/printing/pycode.py
```diff
@@ -175,6 +175,14 @@ def _print_NativeComplex(self, expr):
     def _print_Variable(self, expr):
         return self._print(expr.name)
 
+    def _print_DottedVariable(self, expr):
+        rhs_code = self._print_Variable(expr)
+        lhs_code = self._print(expr.lhs)
+        return f"{lhs_code}.{rhs_code}"
+
+    def _print_DottedFunctionCall(self, expr):
+        return (f"{self._print(expr.func_name)}({', '.join(self._print(arg) for arg in expr.args)})\n")
```
EmilyBourne, line 184: I think the readability of this line could be improved by introducing intermediate variables. Ie:
```suggestion
        func = self._print(expr.func_name)
        args = ', '.join(self._print(arg) for arg in expr.args)
        return (f"{func}({args})\n")
```

pyccel/codegen/printing/pycode.py
```diff
@@ -996,6 +1004,21 @@ def _print_PyccelSymbol(self, expr):
...
+
+    def _print_ClassDef(self, expr):
+        classDefName = 'class {}({}):'.format(expr.name,', '.join(self._print(arg) for arg in  expr.superclass))
+        methods = ''.join(self._print(method) for method in expr.methods)
+        methods = self._indent_codestring(methods)
+        classDef = '\n'.join([classDefName, methods]) + '\n'
+        return classDef
+
+    def _print_ConstructorCall(self, expr):
+        return f"{expr.func.cls_name}({', '.join(self._print(arg) for arg in expr.arguments)})"
```
EmilyBourne, line 1018: See above, it is not easy to see at a glance what the output string will look like

EmilyBourne:
```suggestion
        args = ', '.join(self._print(arg) for arg in expr.arguments)
        return f"{expr.func.cls_name}({args})"
```

EmilyBourne (approved):
Looks good. Well done

yguclu (approved):

PR #1447, opened 2023-07-17
Minor developer doc improvement

EmilyBourne:
Allow each stage of the docs to function more easily as an independent entity by adding a link to the relevant section in the overview.

**Additional Commit**
- Fix the `PYTHONPATH` to avoid errors when checking documentation when new functions are added to `ci_tools`


sboof911 (commented):
This looks perfect.

sboof911 (approved):

yguclu (approved):

PR #1454, opened 2023-07-21
Fix non-fatal developer-mode output

### LINKED ISSUE #1453: No output when using `--developer-mode`
### Describe the bug
When a non-fatal `NotImplementedError` is raised with the `--developer-mode` flag

### To Reproduce
To reproduce the behaviour make `ConstructorCall` inherit from `FunctionCall`. The semantic error:
```
|error [semantic]: classes_1.py| Cannot assign result of a function without a return
Pyccel has encountered syntax that has not been implemented yet. Please create an issue at https://github.com/pyccel/pyccel/issues and provide a small example of your problem. Do not forget to specify your target language
```
will appear. However when running in developer mode no error is shown.

### Expected behavior
The error should be shown with a traceback. Ideally the traceback should show where the error was raised not the line in `codegen/pipeline.py` where `errors.report` is called.

cc: @sboof911

EmilyBourne:
Improve traceback for non-fatal developer-mode errors and fix #1453 

**Bot fixes**
- Return a more precise line for doc errors.
- Fix a bug if a docstring was written in one line


EniddeallA (approved):
Looks good!

ci_tools/process_results.py
```diff
@@ -25,6 +25,7 @@
 
     pyccel_folder = os.path.abspath('compare')
 
+
```
yguclu, line 28: I suppose this empty line was not intentional

EmilyBourne: No it wasn't. Fixed : b008e889

ci_tools/process_results.py
```diff
@@ -55,14 +55,22 @@
...
                     with open(os.path.join(pyccel_folder, file), 'r', encoding='utf-8') as code_file:
                         lines = code_file.readlines()[start-1:end+2]
                     lines = [l.strip() for l in lines]
-                    doc_openings = [i for i,l in enumerate(lines) if l.startswith('"""') or l.endswith('"""')]
+                    doc_openings = []
+                    for i,l in enumerate(lines):
+                        if l.startswith('"""'):
+                            doc_openings.append(i)
+                        if l != '"""' and l.endswith('"""'):
+                            doc_openings.append(i)
```
yguclu, line 69: The new code looks **very** similar to the original one. The only case where it behaves differently is when the docstring is on a single line, in which case you append the line twice now. I do not quite understand why you want this.

EmilyBourne: I need to have a pair of indices for each docstring. One for the start line and one for the end line.
Occasionally people write docstrings like:
```
def f():
    """ Tiny docstring """
```
With the old code this causes problems as the start is identified, but the end is either the next docstring or isn't found. You then get an `IndexError` when you try to access it

sboof911 (approved):
This looks good

yguclu (approved):

PR #1464, opened 2023-07-26
[BOT] Always post run duration

EmilyBourne:
When a workflow is not retriggered (e.g. linux tests are not retriggered when only docs have changed) the duration of the original run is not shown. This information can be easily retrieved from the previous check run and this PR does so.

It was tested on my fork.


yguclu (approved):
Looks good!

PR #1457, opened 2023-07-24
Improve class `ConstructorCall`

sboof911:
Fix #1455:

- Change in the semantic stage the type of `p = Point(0.1, 0.2)` from an `Assign(Variable, ConstructorCall)` to a `ConstructorCall`.
- Make `ConstructorCall` class inherit from `DottedFunctionCall`.
- Add the `cls_variable` as first argument of the `ConstructorCall.__init__()`.
- Fix any issues triggered in the method `_print_ConstructorCall` in codegen after the changes above.

PR #1466, opened 2023-07-27
Bot fixes

EmilyBourne:
Try to fix additional bot problems:

1. Sort reviews before filtering them by author (hopefully will fix the stage determination failing in #1280 and #1457). This should result in only the most recent review being preserved.
2. Set flags if tests are already passing. If the tests were run before marking the PR as draft the bot doesn't check the status and update the flags as this was only done in the clean up. (Bug seen in #1457)
3. Fix a bug in the documentation test where it failed when a function is removed from a class. (Bug seen in #1457)


yguclu (approved):
Looks good to me..

PR #1470, opened 2023-07-31
Reduce use of types decorator

EmilyBourne:


EniddeallA (approved):
Looks good!

docs/templates.md
```diff
@@ -1,6 +1,36 @@
...
+def f(a : 'T', b : 'T'):
+	pass
+```
+Arguments:
+-   name: the name of the template
+-   types: the types the template represents.
+---
+*Note:*
+The arguments **name** and **types** could also be passed of the form
+`@template(T, types=['int', 'real'])` without specifying the keywords.and the order in which the decorators are called is not important.
```
yguclu, line 18: Should `T` be a string here?

Maybe the second sentence should be a separate point in a list, and not start with "and".

docs/templates.md
```diff
@@ -1,6 +1,36 @@
...
+
+---
+When  a function is decorated with the template decorator:
+-   The templates are only available to the decorated function.
+-   The templates overrides any existing templates with the same name (declared as header comment).
```
yguclu, line 23: Typo: overrides -> override.

But maybe a better sentence is "A template overrides any existing template with the same name [...]."

docs/templates.md
```diff
@@ -22,6 +52,10 @@ def f(a,b):
 When a function is declared using a header comment as above:
 -   The function can access the templates declared in its parent's scopes.
 -   A template declared in the same scope as the function overrides any template with the same name in the parent's scopes.
+
+*Note:*
+When both a decorator and a header comment are used, the decorator takes precedent.
```
yguclu, line 57: Typo: precedent -> precedence.

docs/templates.md
EmilyBourne: Fixed 9934fdf8

docs/templates.md
EmilyBourne: Fixed 9934fdf8

docs/templates.md
EmilyBourne: Fixed 9934fdf8

I think the second point was about mixing `@types` and `@template`. I have removed it

yguclu (approved):
Thanks for this PR! Everything looks good to me.

PR #1452, opened 2023-07-21
Add C support for a class containing only functions

sboof911:
Adding the printing of a struct in the `/codegen/printing/ccode.py` file in the `_print_ModuleHeader` function.
Adding the printing of the class methods in the `_print_ClassDef` and passing the struct as a pointer as arguments of the methods.
The `ConstructorCall`  is handeled by the `FunctionCall` and the `init` function is already renamed in `_print_ModuleHeader`.

Issue #1443, opened 2023-07-12
Add destructors for classes


EmilyBourne:
### Relevant Discussion

https://github.com/pyccel/pyccel/discussions/1119

### Describe the feature
Add destructors for classes. These should be implemented similarly to other class methods (see #1441). Additionally it is important to ensure that the destructor calls the destructors of any destructable objects stored in the class (other class objects or arrays). Finally the class destructor should automatically be called when the class goes out of scope (see `pyccel.parser.semantic.SemanticParser._garbage_collector`.

### Test Code
Provide code which does not currently work but which should do when this issue is fixed:
```python
#$ header class Point(public)
#$ header method __init__(Point, double[:])
#$ header method __del__(Point)
#$ header method translate(Point, double[:])
import numpy as np
class Point(object):
    def __init__(self, x):
        self.x = x

    def __del__(self):
        pass

    def translate(self, a):
        self.x[:]   =  self.x + a

if __name__ == '__main__':
    x = np.array([0.,0.,0.])
    p = Point(x)

    a = np.array([1.,1.,1.])

    p.translate(a)
    print(p.x)
```
This should already work after issues #1439 and #1441 but it will contain a memory leak. This issue should fix that leak.

Additionally the following code, where the destructor is not explicit, should also work correctly
```python
#$ header class Point(public)
#$ header method __init__(Point, double[:])
#$ header method translate(Point, double[:])
import numpy as np
class Point(object):
    def __init__(self, x):
        self.x = x

    def translate(self, a):
        self.x[:]   =  self.x + a

if __name__ == '__main__':
    x = np.array([0.,0.,0.])
    p = Point(x)

    a = np.array([1.,1.,1.])

    p.translate(a)
    print(p.x)
```

After this issue all the tests in `tests/pyccel/scripts/classes` should be translatable to all target languages.

### Proposed Solution
Modify the semantic stage to augment any existing `__del__` function, or create one if it is missing.
The deallocation function should be called on any allocatable members.

The printer should print a del function (`<className>__del` for C) which should be called automatically when an instance of the class goes out of scope

PR #1472, opened 2023-08-01
Add C support for a class containing only scalar data

sboof911:
Add visiting the lhs of a `DottedVariable`.
Add `is_argument = True` for the `self` variable when it's in a method argument of a class.
Add the declaration of the attributes of the class in the `struct def`.


mkaddani (approved):
Looks good!

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ All notable changes to this project will be documented in this file.
 -   Python support for a simple class.
 -   #1430 : Added conjugate support to booleans.
 -   #1452 : Added C printing support for a class containing only functions.
+-   #1472 : Added C printing support for a class containing only scalar data.
```
EmilyBourne, line 12: Does the "only" mean that we have an either/or situation? Or is it possible to have a class with both scalar data and functions after this PR and #1452

pyccel/codegen/printing/ccode.py
```diff
@@ -346,7 +345,7 @@ def is_c_pointer(self, a):
             return True
         if isinstance(a, FunctionCall):
             a = a.funcdef.results[0].var
-        if isinstance(getattr(a, 'dtype', None), CustomDataType) and a.is_argument:
+        if isinstance(getattr(a, 'dtype', None), CustomDataType) and (a.is_argument or a.get_user_nodes(FunctionDef)):
```
EmilyBourne, line 348: Please can you explain this change

pyccel/codegen/printing/ccode.py
```diff
@@ -685,6 +684,7 @@ def _print_ModuleHeader(self, expr):
         funcs = ""
         for classDef in expr.module.classes:
             classes += f"struct {classDef.name} {{\n"
+            classes += ''.join(self._print_Declare(Declare(var.dtype,var)) for var in classDef.attributes)
```
EmilyBourne, line 687:
```suggestion
            classes += ''.join(self._print(Declare(var.dtype,var)) for var in classDef.attributes)
```

tests/pyccel/scripts/classes/classes_3_C.py
```diff
@@ -0,0 +1,21 @@
+# pylint: disable=missing-class-docstring,  missing-function-docstring, missing-module-docstring
+#$ header class Point(public)
+#$ header method __init__(Point, double[:])
+#$ header method translate(Point, double[:])
```
EmilyBourne, line 4: These are not scalars. Please stick to scalars until we have looked at #1443 , otherwise we are introducing further memory leaks

EmilyBourne (changes requested):
This looks nice and simple, but I have a couple of questions

pyccel/codegen/printing/ccode.py
sboof911: The `a.get_user_nodes(FunctionDef)` checks if the variable passed in the function `is_c_pointer` is in a `Method Def`. All the attributes of a class in a `Method Def` should be pointers.

pyccel/codegen/printing/ccode.py
sboof911: Fixed

tests/pyccel/scripts/classes/classes_3_C.py
sboof911: Fixed

pyccel/codegen/printing/ccode.py
EmilyBourne: > All the attributes of a class in a `Method Def` should be pointers.

I disagree, but that is not what this does. It ensures that all instances of a class in a `Method Def` should be pointers. E.g:
```
class Point1:
    def __init__(self, x):
        self.x = x

class Point2:
    def __init__(self, y):
        self.y = y

    def test_func(self):
        p = Point1(self.y)
        print(p.x)
```
You say that `p` has to be a pointer?

sboof911: Blocked by https://github.com/pyccel/pyccel/pull/1479

yguclu:
Hi @sboof911, what is blocking this PR?

sboof911:
> Hi @sboof911, what is blocking this PR?

I need to finish PR: https://github.com/pyccel/pyccel/pull/1484

pyccel/codegen/printing/ccode.py
sboof911: Fixed in https://github.com/pyccel/pyccel/pull/1472/commits/4f440b59370f1170d0f5e188b881ea0b41c9bcf8.
But the `self.y` in `test_func` isn't working cause  his `get_user_nodes(ClassDef)` returns `None` but I think normally it has to return the `ClassDef(Point2)`.

EmilyBourne: It can't return `ClassDef(Point2)` until the function `__init__` has been added to the class (which only happens once it's built). I think you'll have to get the class def from the self variable or via the dtype

EmilyBourne: Why do you need to know the current class?

EmilyBourne: `self` is already an argument

sboof911: > 

So I can compare if it s the same class I ll return True in the `is_c_pointer`

EmilyBourne: But it's already an argument. Doesn't it already return true?

sboof911: I m talking about `self.y` in `p = Point1(self.y)`

EmilyBourne: Yes. `self.y` shouldn't be a pointer because `y` isn't a pointer, `self` should be a pointer because it is an argument

EmilyBourne: So in c we will print `self->y` (a variable accessed via a pointer) not `&self->y` (a pointer accessed via a pointer)

sboof911: The `self` return `False` when I print the `is_argument`.

sboof911: I found a simple way and it s to add an attribute to the `CCodePrinter` and it s `self._current_class` and by using it I ll compare it with the cls_base of the variable passed in the `is_c_pointer`.

EmilyBourne: > The `self` return `False` when I print the `is_argument` in all the `DottedVariable` in the exemple.

Ah. That's either means that there are multiple `self` variables or that there's a bug in the initialisation (`self` didn't used to be treated as an argument, but I changed it because it looks like one in most languages and has the same properties). Probably this should be fixed

sboof911: Can you take a look for what I changed. I think the `self._current_class` will really help me in the future.

pyccel/codegen/printing/ccode.py
```diff
@@ -347,6 +348,8 @@ def is_c_pointer(self, a):
             a = a.funcdef.results[0].var
         if isinstance(getattr(a, 'dtype', None), CustomDataType) and a.is_argument:
             return True
+        if a.get_user_nodes(DottedVariable) and a.cls_base is self._current_class:
+            return True
```
EmilyBourne, line 352: I don't understand this. If the object is used by a DottedVariable and the class base is the current class. Why is it necessary for one of the elements to be accessed? What if `self` is passed to another function?

pyccel/codegen/printing/ccode.py
```diff
@@ -2251,7 +2255,9 @@ def _print_CustomDataType(self, expr):
         return "struct " + expr.name
 
     def _print_ClassDef(self, expr):
+        self._current_class = expr
```
EmilyBourne, line 2258: If you want to future-proof this, you may want to consider using a list and using the last element. This will then also handle classes in classes

tests/pyccel/scripts/classes/classes_2_C.py
```diff
@@ -2,13 +2,14 @@
 import numpy as np
 
 #$ header class Point(public)
-#$ header method __init__(Point)
+#$ header method __init__(Point, float, float)
```
EmilyBourne, line 5: Why not create a new test rather than modifying this one to try and cover everything at once? Having multiple incremental tests will make it easier to locate a bug if something breaks

EmilyBourne (changes requested):

pyccel/codegen/printing/ccode.py
EmilyBourne: I have had a look. I imagine this works and I think `_current_class` will be useful, but I still think it could be useful to fix the `is_argument` property.

Or maybe we should just open an issue to remove it and use something like `expr.has_user(FunctionDefArgument)` instead

tests/pyccel/scripts/classes/classes_2_C.py
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1472/commits/2f3a91f4cec9be0463fa0f1621cc76fa8c46b5b2

pyccel/codegen/printing/ccode.py
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1472/commits/b9b353587bd848a608c5e2150cef6434decb65f0.

pyccel/codegen/printing/ccode.py
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1472/commits/b9b353587bd848a608c5e2150cef6434decb65f0.

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
+-   #1472 : Add C printing support for a class contains `scalar data`.
```
EmilyBourne, line 8:
```suggestion
-   #1472 : Add C printing support for a class containing scalar data.
```

EmilyBourne (changes requested):

EmilyBourne:
This seems really clean. I just have a grammar comment then I'm happy

yguclu (approved):

yguclu:
@sboof911 Is the PR description up to date? I do not see a clear correspondence with the changes made.

sboof911:
> @sboof911 Is the PR description up to date? I do not see a clear correspondence with the changes made.

Fixed.

PR #1473, opened 2023-08-02
Fix review status

EmilyBourne:
By default the GitHub API only returns the first 30 reviews. This PR changes that so it collects 100 reviews per page and keeps collecting pages until there are none left. With all the reviews the bot can make a correct decision on whether a review should be re-requested and what is the review state of a PR.
Additionally ci_tools/bot_clean_up.py is modified to avoid duplicating code.

This PR was tested in my fork,


sboof911 (approved):
Looks perfect

yguclu (approved):

PR #1475, opened 2023-08-03
[BOT] Don't cancel dependent jobs

EmilyBourne:
Currently when linux fails or is cancelled, the coverage test is also cancelled. This is problematic when the linux failure was a false positive or the job was cancelled accidentally. In this case the coverage must be triggered manually by providing a large number of parameters. This PR changes that so the coverage tests remains queued permanently. This means it will run whenever the linux test passes


sboof911 (approved):
Looks good.

yguclu (approved):

PR #1478, opened 2023-08-07
Fix bot clean up

EmilyBourne:
The bot is currently failing in the clean-up due to a stupid error. This problem prevents coverage tests being run automatically. This PR fixes it.

Additionally packages were installed on the wrong Python version for anaconda resulting in windows results never being reported. This is also fixed

### DIFF
--- a/ci_tools/bot_comment_react.py
+++ b/ci_tools/bot_comment_react.py
@@ -29,11 +29,11 @@ def get_unique_test_list(keys):
     tests.discard('pr_tests')
     result = list(tests)
     if 'coverage' in tests:
-        tests.append('linux')
+        result.append('linux')
         # Ensure coverage is last in case dependencies are ready
-        tests.remove('coverage')
-        tests.append('coverage')
-    return tests
+        result.remove('coverage')
+        result.append('coverage')
+    return result
 
 if __name__ == '__main__':
     # Parse event payload from $GITHUB_EVENT_PATH variable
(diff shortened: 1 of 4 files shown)


sboof911 (approved):
Looks Good

ci_tools/bot_comment_react.py
```diff
@@ -29,6 +29,9 @@ def get_unique_test_list(keys):
     tests.discard('pr_tests')
     if 'coverage' in tests:
         tests.add('linux')
+        # Ensure coverage is last in case dependencies are ready
+        tests.discard('coverage')
+        tests.add('coverage')
```
yguclu, line 34: I do not understand this change, as Python sets are not ordered AFAIK. Maybe you want to use a list instead?

yguclu (changes requested):

ci_tools/bot_comment_react.py
EmilyBourne: Ah. You are right, dictionaries have confused me. Fixed: a62afc4989a35e521d082320c04c3b808b65f466

ci_tools/bot_comment_react.py
```diff
@@ -27,9 +27,13 @@ def get_unique_test_list(keys):
     if 'pr_tests' in tests:
         tests.update(pr_test_keys)
     tests.discard('pr_tests')
+    result = list(tests)
     if 'coverage' in tests:
-        tests.add('linux')
-    return tests
+        result.append('linux')
```
yguclu, line 32: We should be careful here when using a list. If `linux` is already in the list you will end up with two identical entries.

yguclu (changes requested):

ci_tools/bot_comment_react.py
EmilyBourne: Oops

EmilyBourne: Fixed b6d3319b

yguclu (approved):

PR #1483, opened 2023-08-08
[BOT] Fix docs for ci_tools

EmilyBourne:
Coverage should not raise an issue if an `except` statement is not covered (we do not enforce coverage for error statements such as `raise` and `errors.report`, this is similar and the body of the `except` clause will still raise an issue if anything more complex is used). This PR adds `except` to the exceptions.

Correcting this leads to an error with the documentation. The error arises because `numpydoc` imports all files that it tests so all possible dependencies must be installed. This is corrected in this PR. This issue has previously been misunderstood and fixed by trying to import from different locations. The unnecessary code is therefore removed.

The tests were triggering manually on this branch to ensure it can pass and to test the fix.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -66,7 +66,6 @@ jobs:
         uses: ./compare/.github/actions/linux_install
       - name: Setup imports
         run: |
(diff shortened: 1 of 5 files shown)


sboof911 (approved):
This looks good

yguclu (approved):

PR #1484, opened 2023-08-08
Improve handling of `DottedName` in `_assign_lhs_variable`

sboof911:
- Add methods `add_new_attribute`, `add_new_interface`, `add_new_method` to `ClassDef` class.
- Check class attributes of the variable passed to `check_for_variable` if it is a `DottedName`.
- Change `visit_ClassDef` to instating one `ClassDef`, and use new methods to capture all class attributes without duplicates.
- Add `Datatype` of new class to `dtype_registry`.
- Improve use of class scope to handle name collisions.
- Raise error if the `rhs` of an `Assign` is not a `PythonMap` or `PythonZip` or `PythonEnumerate` or `PythonRange`.


sboof911:

sboof911:

pyccel/ast/core.py
```diff
@@ -3291,8 +3291,25 @@ def name(self):
...
+        """
+
         return self._attributes
 
+    @attributes.setter
+    def attributes(self, attr):
+        if not isinstance(attr, tuple):
+            raise TypeError('attributes must be a tuple.')
+        self._attributes = attr
+        self.set_current_user_node(self)
```
EmilyBourne, line 3311:
```suggestion
    @attributes.setter
    def attributes(self, attr):
        if not isinstance(attr, tuple):
            raise TypeError('attributes must be a tuple.')
        for a in self._attributes:
            self.remove_user_node(a)
        self._attributes = attr
        for a in attr:
             a.set_current_user_node(self)
```

pyccel/parser/semantic.py
```diff
@@ -356,6 +368,34 @@ def check_for_variable(self, name):
...
+            The object describing the attribute.
+
+        Returns
+        -------
+        Variable
+            Returns the attribute if found or None.
+        """
+
+        if self.current_class:
+            attributes_list = list(self._current_class.attributes)
```
EmilyBourne, line 390: Why do you need to convert to a list?

pyccel/parser/semantic.py
```diff
@@ -356,6 +368,34 @@ def check_for_variable(self, name):
...
+            for key, value in enumerate(attributes_list):
+                if value.name == name.name[-1]:
```
EmilyBourne, line 392: Does this basic check (`value.name == name.name[-1]:`) not cause problems in the case of name collisions? Or are these not handled for classes currently anyway?
(See https://github.com/pyccel/pyccel/blob/devel/developer_docs/scope.md)

E.g:
```python
class Point:
    def __init__(self, a, A);
        self.a = a
        self.A = A
```

pyccel/parser/semantic.py
```diff
@@ -356,6 +368,34 @@ def check_for_variable(self, name):
...
+                    attributes_list[key] = value
+                    self._current_class.attributes = tuple(attributes_list)
```
EmilyBourne, line 394: What are you doing here?
The attributes exist, e.g. they are `('a', 'b', 'c')` and we are looking for `'c'`
You get a list: `['a', 'b', 'c']`, find `'c'`, insert the `'c'` from the list into the same location in the list (list unchanged) then update the attributes with their original values (attributes unchanged)

pyccel/parser/semantic.py
```diff
@@ -356,6 +368,34 @@ def check_for_variable(self, name):
...
+                    return value
+        return None
```
EmilyBourne, line 396: I'm not sure if `self._current_class` is the most useful object here. As I understand it in the following example:
```python
class Line:
    def __init__(self, l):
        self.l = l

class Point:
    def func(self, l : Line):
         print(l.l)
```
in the function `func`, `self._current_class` is `Point`, but you want to use this function to find out about `Line`.

I suggest something like:
```suggestion
        assert isinstance(name, DottedName)
        prefix = self._visit(DottedName(*name.name[:-1]))
        class_def = prefix.cls_base
        attr_name = name.name[-1]
        try:
             value = next(a for a in class_def.attributes if a.name == attr_name)
        except StopIteration:
             value = None
        return value
```

Although if we are avoiding name collision problems, maybe it is simply:
```suggestion
        assert isinstance(name, DottedName)
        prefix = self._visit(DottedName(*name.name[:-1]))
        class_def = prefix.cls_base
        attr_name = name.name[-1]
        return class_def.scope.find(attr_name, 'variables')
```

pyccel/parser/semantic.py
```diff
@@ -1108,11 +1148,36 @@ def _assign_lhs_variable(self, lhs, d_var, rhs, new_expressions, is_augassign,ar
...
+            if isinstance(lhs, DottedName):
+                var = self.check_for_attribute(lhs)
+            else:
+                var = self.check_for_variable(name)
             # Variable not yet declared (hence array not yet allocated)
             if var is None:
-
+                if isinstance(lhs, DottedName):
+                    if (self._current_function == '__init__' or self.current_class) and lhs.name[0] == 'self':
+                        cls      = self.get_variable('self')
```
EmilyBourne, line 1159:
```suggestion
                        var      = self.get_variable('self')
```
This is recollected below

pyccel/parser/semantic.py
```diff
@@ -1108,11 +1148,36 @@ def _assign_lhs_variable(self, lhs, d_var, rhs, new_expressions, is_augassign,ar
...
+                        var      = self.get_variable('self')
+
+                        member = self._create_variable(name, dtype, rhs, d_lhs)
+                        lhs    = member.clone(member.name, new_class = DottedVariable, lhs = var)
+
+                        # update the attributes of the class and push it to the scope
+                        attributes += [member]
+                        new_cls = ClassDef(cls_name, attributes, [], superclasses=parent)
+                        self._current_class = new_cls
+                        self.scope.parent_scope.update_class(new_cls)
```
EmilyBourne, line 1177: Do you think we can update the existing class or do we need to create a new one each time? Shouldn't your setter let us simplify this?

pyccel/ast/core.py
EmilyBourne: I wonder if it would make more sense to have a simpler `add_new_attribute`:
```python
def add_new_attribute(self, attr):
     if not isinstance(attr, Variable):
          raise TypeError("Attributes must be Variables")
     attr.set_current_user_node(self)
     self._attributes += (attr,)
```

EmilyBourne: If we want to handle name collisions properly we should probably also add new attributes to the class scope:
```python
self.scope.insert_variable(attr)
```

EmilyBourne (commented):
Let me know if you want a chat about the `Scope` class. I don't think you've worked with it before. It is a bit cumbersome but I think you will need it in several places for this PR

pyccel/ast/core.py
sboof911: Fixed.

sboof911: we don't really need the `add_new_attribute` the adding of the new attributes is already handled in : https://github.com/pyccel/pyccel/blob/58a355e5f498b271a4c456478a397c40c0d4f9fa/pyccel/parser/semantic.py#L1165-L1169

pyccel/parser/semantic.py
sboof911: Fixed.

pyccel/parser/semantic.py
sboof911: Fixed.

pyccel/parser/semantic.py
EmilyBourne: > Their is no problems in the case of name collisions

Is that because name collisions aren't handled? Or because you are only testing C? Or is this handled in some unexpected way?

I would be very surprised if this works. Maybe I should give a test case which has a C name collision, rather than a fortran name collision. E.g:
```python
class Point:
    def __init__(self, a);
        self.double = a*2
```
`double` is a keyword in C so the attribute should rename itself to something like `double_0001` (I'm not sure this collision avoidance is implemented though, it is possible that names are not always verified with the Scope).
If this si the case then I think `name.name[-1]` will still be `double` but `value.name` will be `double_0001`. Is this not the case? If not then how did `name.name[-1]` become `double_0001`?

pyccel/parser/semantic.py
EmilyBourne: This is still collected twice

pyccel/ast/core.py
sboof911: I used it here so we can avoid the creating of the class each time
https://github.com/pyccel/pyccel/blob/58a355e5f498b271a4c456478a397c40c0d4f9fa/pyccel/parser/semantic.py#L1170-L1174

pyccel/parser/semantic.py
```diff
@@ -1108,11 +1143,37 @@ def _assign_lhs_variable(self, lhs, d_var, rhs, new_expressions, is_augassign,ar
...
+                        name     = str(lhs.name[-1])
+
+                        member = self._create_variable(name, dtype, rhs, d_lhs)
+                        lhs    = member.clone(member.name, new_class = DottedVariable, lhs = var)
+
+                        # update the attributes of the class and push it to the scope
+                        attributes += [member]
+                        if not self._current_class:
+                            new_cls = ClassDef(cls_name, attributes, [], superclasses=parent)
+                            self._current_class = new_cls
```
EmilyBourne, line 1171: Is this possible? The coverage bot will make you test it, but I am not clear on what case it covers

pyccel/parser/semantic.py
```diff
@@ -1108,11 +1143,37 @@ def _assign_lhs_variable(self, lhs, d_var, rhs, new_expressions, is_augassign,ar
...
+                        else:
+                            self._current_class.attributes = tuple(attributes)
```
EmilyBourne, line 1173: I would prefer:
```python
self._current_class.add_new_attribute(member)
```
to be sure the tree is not invalidated

pyccel/parser/semantic.py
sboof911: It will be tested each time their is a new creating of a class (precisely in `__init__` method).

pyccel/parser/semantic.py
EmilyBourne: Where?

pyccel/parser/semantic.py
sboof911: it is, but the `self.current_class` remain empty.

pyccel/parser/semantic.py
sboof911: Fixed.

pyccel/ast/core.py
EmilyBourne: Then I think this is a problem. For example:
```python
1 class MyClass:
2     def __init__(self, a : int, b : float, c : int):
3          self.t = (a,b)
4          self.c = c
```
At line 3 we create an `InhomogeneousTupleVariable` (`t`). This variable uses 2 sub-variables `t_0001` and `t_0002` which represent respectively the `int` and `float` parts of the tuple.
We create a class and save the attribute in this class.

At line 4 we create a `Variable`. We need to add this to the attributes of the class so we call the attribute setter.
The setter is designed to completely replace all the attributes with a new set of unrelated attributes so it begins by
telling the attributes that it currently uses that they are no longer needed.
The `InhomogeneousTupleVariable` `t` sees that it is no longer used by anyone. It assumes that this means it will never be used. It therefore no longer needs to use anything else and should be removed from the `users` of anything it is using so they do not have the wrong tree (this makes more sense in the case of something like a `PyccelAdd` which you may decide is unnecessary. You want it to be garbage collected, but it can't be as long as it says it is using existing Variables as they store it as an attribute). It therefore removes itself from the users of `t_0001` and `t_0002`.
Now we add the new attributes.
We therefore tell them that they have new users. `t` is also in the new users, so it is not garbage collected, however it is no longer aware that it has the attributes `t_0001` and `t_0002` which could cause big problems later.

There are therefore 2 solutions:
1. Add an `invalidate` argument to the setter (which will therefore become a function) so you don't invalidate the arguments if you are planning to reuse them.
2. Stop doing so many unnecessary steps and just have a `add_new_attribute` function

pyccel/parser/semantic.py
EmilyBourne: That's stupid. We should fix that

EmilyBourne: Or get rid of `self._current_class` completely. Is it needed?

EmilyBourne: The use of `self._current_class` here:
https://github.com/pyccel/pyccel/blob/275b85717e8bfc2f42eab6646ec3007f3753aa09/pyccel/parser/semantic.py#L2102-L2104
seems especially dangerous. If we are going to rely on this variable for anything then we need to make sure it can't change at random. I would feel safer if we used the `cls_base` of the lhs of a `DottedName` to find the class as this information is not global

pyccel/ast/core.py
sboof911: Fixed.

pyccel/parser/semantic.py
sboof911: Fixed now.

pyccel/parser/semantic.py
sboof911: Fixed.

pyccel/parser/semantic.py
sboof911: Fixed.

pyccel/parser/semantic.py
```diff
@@ -344,11 +344,22 @@ def change_to_module_scope(self):
...
+
+        Returns
+        -------
+        Variable
+            Returns the varibale if found or None.
+        """
+        if self.current_class and not name == 'self':
             for i in self._current_class.attributes:
                 if i.name == name:
                     var = i
```
EmilyBourne, line 365: Should this bit call `check_for_attribute`?
Do we need both `check_for_variable` and `check_for_attribute`?
Or could we do:
```python
def check_for_variable(self, name):
    if isinstance(name, DottedName):
        # body of check_for_attribute
    else:
        # body of check_for_variable
```
?

pyccel/parser/semantic.py
```diff
@@ -357,6 +368,32 @@ def check_for_variable(self, name):
...
+        name : DottedName
+            The object describing the attribute.
+
+        Returns
+        -------
+        Variable
+            Returns the attribute if found or None.
+        """
+
+        prefix = self._visit(DottedName(*name.name[:-1]))
```
EmilyBourne, line 389: Does the class `DottedName` return a PyccelSymbol if you only pass one element?

pyccel/parser/semantic.py
EmilyBourne: Where?

pyccel/parser/semantic.py
EmilyBourne: This looks better. But the new code may raise the name collision problems that I mentioned above

pyccel/parser/semantic.py
```diff
@@ -1109,11 +1146,34 @@ def _assign_lhs_variable(self, lhs, d_var, rhs, new_expressions, is_augassign,ar
...
+                        cls_name = str(var.cls_base.name)
+                        cls      = self.scope.find(cls_name, 'classes')
+
+                        name     = str(lhs.name[-1])
+                        name = self.scope.get_expected_name(name)
+                        member = self._create_variable(name, dtype, rhs, d_lhs)
+
+                        lhs    = member.clone(member.name, new_class = DottedVariable, lhs = var)
+                        if not str(lhs.name[-1]) == name:
+                            self.scope.insert_variable(lhs, name)
```
EmilyBourne, line 1169: Why does this variable only need inserting if the name has changed? Doesn't it always need inserting?

pyccel/parser/semantic.py
EmilyBourne: When possible ([as mentioned in the docs](https://github.com/pyccel/pyccel/blob/devel/developer_docs/review_process.md?plain=1#L25C116-L25C192)), please leave references to commits or code when you say something is fixed so it is easy to check.

pyccel/parser/semantic.py
sboof911: It s already insterted (I insert just it just if the `get_expected_name` returns a new name)

pyccel/parser/semantic.py
sboof911: Fixed.

pyccel/parser/semantic.py
sboof911: I added the `self.scope.remove_variable` to delete the previous variable from the scope.

pyccel/parser/semantic.py
EmilyBourne: When possible ([as mentioned in the docs](https://github.com/pyccel/pyccel/blob/devel/developer_docs/review_process.md?plain=1#L25C116-L25C192)), please leave references to commits or code when you say something is fixed so it is easy to check.

pyccel/parser/semantic.py
EmilyBourne: > It s already insterted

Can you point me at the code where this happens please

sboof911: the `id(var)` in here equals the `id(lhs)` in the `__assign_lhs_variable()` 
https://github.com/pyccel/pyccel/blob/275b85717e8bfc2f42eab6646ec3007f3753aa09/pyccel/parser/semantic.py#L3270-L3271

EmilyBourne: Huh, how does that happen?
In:
```python
class Point:
    def __init__(self : Point, x : float):
        self._x = x
```
Don't we have `var`, which is `self`, and `lhs` which is `self._x`?

pyccel/parser/semantic.py
sboof911: Fixed. [2fcd3f2](https://github.com/pyccel/pyccel/pull/1484/commits/2fcd3f251abe4f9584e5de7060ab6e20f05509c5)

EmilyBourne: > Fixed. [2fcd3f2](https://github.com/pyccel/pyccel/pull/1484/commits/2fcd3f251abe4f9584e5de7060ab6e20f05509c5)

This does not fix the collisions it is just a different way of writing the same thing

pyccel/parser/semantic.py
```diff
@@ -344,18 +344,38 @@ def change_to_module_scope(self):
...
+        name : str
+            The object describing the variable.
+
+        Returns
+        -------
+        Variable
+            Returns the varibale if found or None.
+        """
+        if isinstance(name, DottedName):
+            prefix = self._visit(name)
```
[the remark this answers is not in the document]
sboof911: Fixed. [ec8d7f6](https://github.com/pyccel/pyccel/pull/1484/commits/ec8d7f675e709bef973e70d45bf55ec633e2aae0)

pyccel/parser/semantic.py
EmilyBourne: See #1503

pyccel/parser/semantic.py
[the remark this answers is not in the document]
EmilyBourne: Have you tested this fix? I am concerned that it returns a tuple and therefore `prefix` will be `PyccelTuple(LiteralString('self'),)`.

Shouldn't you have:
```python
prefix_parts = name.name[:-1]
syntactic_prefix = PyccelSymbol(prefix_parts[0]) if len(prefix_parts) == 1 else DottedName(*prefix_parts)
prefix = self._visit(syntactic_prefix)
```
(I'm not sure if the `PyccelSymbol` is necessary, `prefix_parts[0]` may already be a `PyccelSymbol` )

Or alternatively keep the code you had originally, but add a `__new__` method to `DottedName`:
```python
def __new__(cls, *args):
    if len(args) == 1:
        return PyccelSymbol(args[0])
    else:
        return super().__new__(cls)
```

pyccel/parser/semantic.py
```diff
@@ -1114,6 +1135,26 @@ def _assign_lhs_variable(self, lhs, d_var, rhs, new_expressions, is_augassign,ar
...
+                        member = self._create_variable(name, dtype, rhs, d_lhs)
+
+                        lhs    = member.clone(member.name, new_class = DottedVariable, lhs = var)
+                        if not str(lhs.name[-1]) == name:
+                            self.scope.remove_variable(lhs, lhs.name[-1])
+                            self.scope.insert_variable(lhs, name)
+
+                        # update the attributes of the class and push it to the scope
+                        class_def.add_new_attribute(member)
+                        self.scope.parent_scope.update_class(class_def)
```
EmilyBourne, line 1154:
```suggestion
                    if isinstance(class_def, ClassDef) and lhs.name[0] == 'self':
                        var      = self.get_variable('self')
                        
                        # Get the scope for the class
                        class_scope = class_def.scope

                        # Collect the name that should be used in the generated code
                        attribute_name = lhs.name[-1]
                        new_name = class_scope.get_expected_name(attribute_name)
                        # Create the attribute
                        member = self._create_variable(name, dtype, rhs, d_lhs)
                        
                        # Insert the attribute to the class scope
                        # Passing the original name ensures that the attribute can be found under this name 
                        class_def.scope.insert_variable(member, attribute_name)

                        # Create the local DottedVariable
                        lhs    = member.clone(member.name, new_class = DottedVariable, lhs = var)

                        # update the attributes of the class and push it to the scope
                        class_def.add_new_attribute(member)
```

EmilyBourne:
You will need to change something somewhere so we don't create `Declare` objects for `DottedVariable`s. Either remove them at the end of `_print_FunctionDef` and `_print_Module` or filter out these objects when the `Declare`s are created

pyccel/parser/semantic.py
```diff
@@ -344,18 +344,39 @@ def change_to_module_scope(self):
...
         else:
-            return self.scope.find(name, 'variables')
+            if self.current_class:
+                for i in self._current_class.attributes:
+                    if i.name == name:
+                        var = i
+                        return var
+                return None
+            else:
+                return self.scope.find(name, 'variables')
```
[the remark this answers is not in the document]
sboof911: Fixed in https://github.com/pyccel/pyccel/pull/1484/commits/975113feafc9e9351e2780c515b649aaf9c02030.
But I left : https://github.com/pyccel/pyccel/blob/975113feafc9e9351e2780c515b649aaf9c02030/pyccel/parser/semantic.py#L372-L380
Because it s used in https://github.com/pyccel/pyccel/blob/975113feafc9e9351e2780c515b649aaf9c02030/pyccel/parser/semantic.py#L2122-L2124

pyccel/parser/semantic.py
sboof911: Fixed in https://github.com/pyccel/pyccel/pull/1484/commits/975113feafc9e9351e2780c515b649aaf9c02030.
But I added a conditions cause the `insert_variable` returns an Error `Variable already exists in scope` and I m investigating the source of the Error.

pyccel/parser/semantic.py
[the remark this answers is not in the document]
sboof911: Fixed in https://github.com/pyccel/pyccel/pull/1484/commits/1409206f5f7e43db2ba26bdbf971a9d81c9dd74c.
The `prefix_parts[0]` is alredy a `PyccelSymbol`.

pyccel/parser/semantic.py
sboof911: and I just investigated and the problem comes from:
https://github.com/pyccel/pyccel/blob/1409206f5f7e43db2ba26bdbf971a9d81c9dd74c/pyccel/parser/semantic.py#L3342-L3343

sboof911: Fixed https://github.com/pyccel/pyccel/pull/1484/commits/802d795332b4c5c772830c7bd197c4b16347f873.

pyccel/parser/semantic.py
sboof911: Fixed.https://github.com/pyccel/pyccel/blob/802d795332b4c5c772830c7bd197c4b16347f873/pyccel/parser/semantic.py#L1149-L1149

pyccel/parser/semantic.py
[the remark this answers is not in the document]
EmilyBourne: Does it fail there if you remove it? It seems like you should be able to simply do:
```
self.check_for_variable(expr)
```
Instead?

pyccel/parser/semantic.py
```diff
@@ -1109,11 +1133,36 @@ def _assign_lhs_variable(self, lhs, d_var, rhs, new_expressions, is_augassign,ar
...
                 self._ensure_target(rhs, d_lhs)
 
-            var = self.check_for_variable(name)
+            var = self.check_for_variable(lhs)
 
             # Variable not yet declared (hence array not yet allocated)
             if var is None:
 
+                if isinstance(lhs, DottedName):
+                    prefix = self._visit(DottedName(*lhs.name[:-1]))
```
EmilyBourne, line 1142: See above? Symbol if len == 1

pyccel/parser/semantic.py
```diff
@@ -1109,11 +1133,36 @@ def _assign_lhs_variable(self, lhs, d_var, rhs, new_expressions, is_augassign,ar
...
+                    class_def = prefix.cls_base
+                    if isinstance(class_def, ClassDef) and lhs.name[0] == 'self':
+                        var      = self.get_variable('self')
+
+                        # Collect the name that should be used in the generated code
+                        attribute_name = lhs.name[-1]
+                        new_name = self.scope.get_expected_name(attribute_name)
```
EmilyBourne, line 1149: Is this the correct scope? Or should it be the class scope? Are class variables correctly inserted in the symbolic stage?

pyccel/parser/semantic.py
```diff
@@ -3562,32 +3581,19 @@ def _visit_ClassDef(self, expr):
                    bounding_box=(self._current_fst_node.lineno, self._current_fst_node.col_offset),
                    severity='error')
 
-        ms = []
         for i in methods:
             self._visit_FunctionDef(i)
             m_name = i.name.replace("'", '')
```
EmilyBourne, line 3586: I suspect this is no longer needed. Do you mind testing?

EmilyBourne (changes requested):
Getting cleaner and cleaner :) but please add the test in #1503

pyccel/parser/semantic.py
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1484/commits/05b9fdaecb57509ec9f43a70a9acd791f4ffdee8.

pyccel/parser/semantic.py
```diff
@@ -1217,42 +1266,10 @@ def _assign_lhs_variable(self, lhs, d_var, rhs, new_expressions, is_augassign,ar
...
-                lhs    = member.clone(member.name, new_class = DottedVariable, lhs = var)
-
-                # update the attributes of the class and push it to the scope
-                attributes += [member]
-                new_cls = ClassDef(cls_name, attributes, [], superclasses=parent)
-                self.scope.parent_scope.update_class(new_cls)
-            else:
-                lhs = self._visit(lhs)
+                if isinstance(lhs, DottedName):
+                    lhs = var.clone(var.name, new_class = DottedVariable, lhs = lhs.name[0])
```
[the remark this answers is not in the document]
sboof911: `var` here is a `Variable`.

pyccel/parser/semantic.py
sboof911: It s still used, it s here where the `__init__` method is visited and raises an error if no one found https://github.com/pyccel/pyccel/blob/0bb6b91d22218d114e61e5497db52ddc0a090167/pyccel/parser/semantic.py#L3572-L3585
and here where the rest of the methods if visited https://github.com/pyccel/pyccel/blob/0bb6b91d22218d114e61e5497db52ddc0a090167/pyccel/parser/semantic.py#L3587-L3590

EmilyBourne: I just mean the last line which removes quotes

sboof911: Fixed https://github.com/pyccel/pyccel/pull/1484/commits/a93c88ddfb418120ef1227bb6c42c4171a2cfb49.

pyccel/parser/semantic.py
EmilyBourne: Oops I mean syntactic

EmilyBourne: If the test is passing this is probably not critical for now

sboof911: Fixed https://github.com/pyccel/pyccel/pull/1484/commits/b9019b1faba522c9db5185e165a5a9cbaa619b90.

pyccel/parser/semantic.py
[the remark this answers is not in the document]
EmilyBourne: I'm not sure why you would want to change the `if`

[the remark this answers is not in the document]
EmilyBourne: I meant you should be able to use
`self.check_for_variable(expr)` here:
https://github.com/pyccel/pyccel/blob/975113feafc9e9351e2780c515b649aaf9c02030/pyccel/parser/semantic.py

EmilyBourne:
It looks like variable names are chosen correctly with the scope but not function names. It seems like each of the `FunctionDef` objects in the `Interface` have the same name at the moment

sboof911:
The commit https://github.com/pyccel/pyccel/pull/1484/commits/0b14d17fe420c74be8bc984cc2ca41e0089e76ab is necessary because the expected error is solved by this PR and new raised error should be fixed in the issue https://github.com/pyccel/pyccel/issues/1505

EmilyBourne:
> The commit https://github.com/pyccel/pyccel/pull/1484/commits/0b14d17fe420c74be8bc984cc2ca41e0089e76ab is necessary because the expected error is solved by this PR and new raised error should be fixed in the issue https://github.com/pyccel/pyccel/issues/1505

The files in `known_bugs` are not expected to raise any errors in particular. The folder simply contains files that have been successfully translated in the past (eg maybe 5 years ago in this case) but no longer work. Files in this folder shouldn't be deleted, only moved if they start working

pyccel/parser/semantic.py
sboof911: I found that this code : 
```python
class Line:
    def __init__(self, l):
        self.l = l

class Point:
    def __init__(self, l : Line):
        print(l.l)
```
I found that the problem comes from the `dtype` of the `lhs` of the variable `l.l` cause this `dtype` is saved as `line` but the classe name saved in the scope is `Line` so it returns an Error cause it s not the same string. I need just a hint if this change in the naming comes from the `semantic part` or the `syntactic part`.

EmilyBourne: Huh. That's odd. It can't be the syntactic part because there are no types there.  It could be in ast.datatypes but I thought I'd got rid of the decapitalisation

EmilyBourne: How are you printing the dtype? I think there is a difference between `str(dtype)` and `dtype.name`. I think you should be using `dtype.name` to look things up in the class not `dtype` which is a `CustomDataType` instance

sboof911: Fixed https://github.com/pyccel/pyccel/pull/1484/commits/da659e8c20e668003a7ee8480d3724d51b9179b0.

pyccel/ast/core.py
```diff
@@ -3323,6 +3323,16 @@ def name(self):
...
+        """
+        The attributes of a class.
+
+        A tuple of the attributes of a ClassDef.
+
+        Returns
+        -------
+        None
+        """
+
```
mkaddani, line 3335:
```suggestion
        """
        The attributes of a class.
        
        Returns a tuple of the attributes of a ClassDef.
        
        Args:
            self (ClassDef): The class whose attributes are being retrieved.
        
        Returns:
            tuple: A tuple of the attributes of the class.
        """

```

pyccel/ast/core.py
```diff
@@ -3371,8 +3381,68 @@ def attributes_as_dict(self):
...
+        if not isinstance(method, FunctionDef):
+            raise TypeError("Method must be FunctionDef")
+        method.set_current_user_node(self)
+        self._methods += (method,)
+
+    def add_new_interface(self, interface):
+        """
+        Adding a new interface to the current class.
```
mkaddani, line 3428:
```suggestion
        Add a new interface to the current class.
```

mkaddani (changes requested):
Looks good to me, i did request some change on docstring

mkaddani (approved):
Looks good

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
+-   #1484 : Improving handling of `DottedName` in `_assign_lhs_variable`.
```
EmilyBourne, line 20: I think you can add
```
-    #1484 : Using scope for classes to avoid name clashes 
```

This PR fixes your test file

pyccel/ast/core.py
```diff
@@ -3371,8 +3380,68 @@ def attributes_as_dict(self):
...
+        A new attributes of the current ClassDef.
+
+        Parameters
+        ----------
+        attr : Varibale
+            The Variable that will be added.
+
+        Returns
+        -------
+        None
```
EmilyBourne, line 3396: You don't need to provide a `Returns` section if nothing is returned

pyccel/ast/core.py
```diff
@@ -3371,8 +3380,68 @@ def attributes_as_dict(self):
...
+        A new method of the current ClassDef.
+
+        Parameters
+        ----------
+        attr : FunctionDef
+            The Method that will be added.
+
+        Returns
+        -------
+        None
```
EmilyBourne, line 3417: See above

pyccel/ast/core.py
```diff
@@ -3371,8 +3380,68 @@ def attributes_as_dict(self):
...
+        A new interface of the current ClassDef.
+
+        Parameters
+        ----------
+        attr : FunctionDef
+            The interface that will be added.
+
+        Returns
+        -------
+        None
```
EmilyBourne, line 3438: See above

pyccel/parser/semantic.py
```diff
@@ -344,17 +344,39 @@ def change_to_module_scope(self):
...
+
+        Parameters
+        ----------
+        name : str
+            The object describing the variable.
+
+        Returns
+        -------
+        Variable
+            Returns the varibale if found or None.
```
EmilyBourne, line 360: Typo

```suggestion
            Returns the variable if found or None.
```

pyccel/parser/semantic.py
```diff
@@ -344,17 +344,39 @@ def change_to_module_scope(self):
...
+            if attribute:
+                return attribute
+            else:
+                return None
         else:
+            if self.current_class:
+                for i in self._current_class.attributes:
+                    if i.name == name:
+                        var = i
+                        return var
```
EmilyBourne, line 379: Can you remind me why this is necessary please. Is it because `_visit_DottedName` strips off the prefix? Can `_visit_DottedName` pass the full `DottedName` instead? If not, why not?

pyccel/parser/semantic.py
```diff
@@ -1116,11 +1138,37 @@ def _assign_lhs_variable(self, lhs, d_var, rhs, new_expressions, is_augassign,ar
...
 
             # Variable not yet declared (hence array not yet allocated)
             if var is None:
 
+                if isinstance(lhs, DottedName):
+                    prefix_parts = lhs.name[:-1]
+                    syntactic_prefix = prefix_parts[0] if len(prefix_parts) == 1 else DottedName(*prefix_parts)
+                    prefix = self._visit(syntactic_prefix)
+                    class_def = prefix.cls_base
+                    if isinstance(class_def, ClassDef) and lhs.name[0] == 'self':
```
EmilyBourne, line 1151: Shouldn't this be:
```suggestion
                    if isinstance(class_def, ClassDef) and prefix.name == 'self':
```
?

Otherwise this would be called if the class contained an instance of another class. E.g:
```python
def __init__(self, p : Point, x : float):
     self._p = p
     self._p.x = x
```

EmilyBourne, line 1151: Is it possible for `class_def` to be a type other than `ClassDef`? If so then the associated error doesn't seem to be handled.

I suspect that `class_def` can only be a `ClassDef`. It may be `None` if there is a bug in our semantics but this probably doesn't need to be caught.
If you agree then I think we can remove this part of the condition

pyccel/parser/semantic.py
```diff
@@ -1148,13 +1196,13 @@ def _assign_lhs_variable(self, lhs, d_var, rhs, new_expressions, is_augassign,ar
                             bounding_box=(self._current_fst_node.lineno,
                                 self._current_fst_node.col_offset))
 
-                new_name = self.scope.get_expected_name(name)
-
-                # Create new variable
-                lhs = self._create_variable(new_name, dtype, rhs, d_lhs, arr_in_multirets=arr_in_multirets)
+                if not (isinstance(lhs, DottedVariable)):
```
EmilyBourne, line 1199:
```suggestion
                if not isinstance(lhs, DottedVariable):
```

pyccel/parser/semantic.py
```diff
@@ -3274,6 +3293,7 @@ def _visit_FunctionDef(self, expr):
                 arguments = arguments[1:]
                 dt        = self.get_class_construct(cls_name)()
                 cls_base  = self.scope.find(cls_name, 'classes')
+                cls_base.scope.insert_symbols(expr.scope.local_used_symbols.copy())
```
EmilyBourne, line 3296: Shouldn't this be in `_visit_ClassDef`?

pyccel/ast/core.py
```diff
@@ -3323,6 +3323,15 @@ def name(self):
...
     @property
     def attributes(self):
+        """
+        The attributes of a class.
+
+        Returns a tuple of the attributes of a ClassDef.
+
+        Returns
+        -------
+        tuple: A tuple of the attributes of the class.
```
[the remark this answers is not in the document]
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1484/commits/be10551c86b0a69ce02281606259134a26ed64a7.

pyccel/ast/core.py
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1484/commits/be10551c86b0a69ce02281606259134a26ed64a7.

pyccel/ast/core.py
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1484/commits/be10551c86b0a69ce02281606259134a26ed64a7.

pyccel/ast/core.py
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1484/commits/be10551c86b0a69ce02281606259134a26ed64a7.

pyccel/parser/semantic.py
sboof911: It will not do a change cause the `check_for_varibale` returns the variable in the exemple of `self._p.x` so it doesn't access the condition `if var == None` but it's fixed in https://github.com/pyccel/pyccel/pull/1484/commits/e7b19d20720e2c2800a7858162928c0aae6ea639 cause it's more readable.

sboof911: Fixed https://github.com/pyccel/pyccel/pull/1484/commits/e7b19d20720e2c2800a7858162928c0aae6ea639.

pyccel/parser/semantic.py
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1484/commits/e7b19d20720e2c2800a7858162928c0aae6ea639.

pyccel/parser/semantic.py
sboof911: It's the same things if it's in the `_visit_ClassDef` and I think it's better to let it in the `_visit_FunctionDef`.

CHANGELOG.md
EmilyBourne: Beware when you merge. The version was just updated to 1.9.1 so you will need to move your comments to a newer version

pyccel/parser/semantic.py
EmilyBourne: Ah you're right :+1:

pyccel/parser/semantic.py
EmilyBourne: Why do you think it is better here? I think in `_visit_ClassDef` it can be done at the creation of the scope. It will only need to be done once and an `if` check is not needed

EmilyBourne: Ah! Wait I think I see. Here you are not updating with the class scope variables but rather with the local variables of the function.
Can you explain why this is needed please

pyccel/parser/semantic.py
sboof911: This code is used for : https://github.com/pyccel/pyccel/blob/e7b19d20720e2c2800a7858162928c0aae6ea639/pyccel/parser/semantic.py#L2125-L2128
You suggested to change it to a simple `second = self.check_for_variable(rhs_new_name)`.
I couldn't find a way to pass it from : https://github.com/pyccel/pyccel/blob/e7b19d20720e2c2800a7858162928c0aae6ea639/pyccel/parser/semantic.py#L363-L369

pyccel/parser/semantic.py
sboof911: I need it for : https://github.com/pyccel/pyccel/blob/e7b19d20720e2c2800a7858162928c0aae6ea639/pyccel/parser/semantic.py#L1156-L1156

pyccel/parser/semantic.py
EmilyBourne: No. I suggest to change this:
https://github.com/pyccel/pyccel/blob/e7b19d20720e2c2800a7858162928c0aae6ea639/pyccel/parser/semantic.py#L2123-L2129
to:
```python
            # standard class attribute 
             second = self.check_for_variable(expr)
             if second: 
                 return second
```

pyccel/parser/semantic.py
EmilyBourne: Hmm, these shouldn't be the same thing but I'm not sure how to fix it. Let's leave it for now

pyccel/parser/semantic.py
sboof911: We can't do that cause for exemple if we have : `get_Point_Class().x`; `get__Point_Class()` is a `FunctionCall` and it doesn't have a `cls_base`attributes so it will return an Error in `check_for_variable()`.

pyccel/parser/semantic.py
sboof911: The `get_expected_name()` check's if the `attribute_name` is in the `class_def.scope._used_symbols` That's why I insert them into the `classdef.scope`
https://github.com/pyccel/pyccel/blob/2fb0e0da37382e0ed8c144865547a70bcbb14b8c/pyccel/parser/scope.py#L560-L567

pyccel/parser/semantic.py
EmilyBourne: Ah ok makes sense. I wonder if `cls_base` should be added to `PyccelAstNode` at some point in the future. You could also use the `dtype` and the scope to get the class def

sboof911: Fixed https://github.com/pyccel/pyccel/pull/1484/commits/d0556e56b54da503b742820df96af7ad95b0c520. 
I left the `second.clone()` because we need it to be a `DottedVariable`.

pyccel/parser/semantic.py
EmilyBourne:
```
class Point:
     def __init__(self):
         self.x = 1
         self.X = 1

    def get_x(self):
        return self.x

    def get_X(self):
        return self.X
```

pyccel/parser/semantic.py
EmilyBourne: Do you think `check_for_variable` should return a DottedVariable? That makes more sense to me

pyccel/parser/semantic.py
```diff
@@ -2105,10 +2115,9 @@ def _visit_DottedName(self, expr):
         # look for a class attribute / property
         elif isinstance(rhs, PyccelSymbol) and cls_base:
             # standard class attribute
-            if rhs in attr_name:
-                self._current_class = cls_base
-                second = self._visit(rhs)
-                self._current_class = None
+            rhs_new_name = cls_base.scope.get_expected_name(rhs) if cls_base.scope else rhs
+            if rhs_new_name in attr_name:
```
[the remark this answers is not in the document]
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1484/commits/f88f28431dfd69ed4666bba5378025beb7c455ea.

pyccel/parser/semantic.py
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1484/commits/f88f28431dfd69ed4666bba5378025beb7c455ea.

sboof911:
/bot run docs pylint pyccel_lint spelling linux

EmilyBourne (approved):
Well done for working through this. Sorry for nitpicking so much but hopefully this should make your next few PRs easier

pyccel/parser/semantic.py
EmilyBourne: Can you test this example and open an issue containing the generated code if it fails please. I suspect it will generate code that won't compile. I don't have a computer to run it myself

sboof911:
/bot run docs pylint pyccel_lint spelling linux

pyccel/parser/semantic.py
sboof911: It returns an Error, I'll open an issue with what should be fixed.

pyccel/ast/core.py
```diff
@@ -3371,8 +3376,56 @@ def attributes_as_dict(self):
...
-    # TODO add other attributes?
+    def add_new_attribute(self, attr):
+        """
+        Add new attributes to the current class.
+
+        A new attributes of the current ClassDef.
+
+        Parameters
+        ----------
+        attr : Varibale
```
yguclu, line 3387: Typo:
```suggestion
        attr : Variable
```

pyccel/ast/core.py
```diff
@@ -3371,8 +3376,56 @@ def attributes_as_dict(self):
             d_attributes[i.name] = i
         return d_attributes
 
-    # TODO add other attributes?
+    def add_new_attribute(self, attr):
+        """
+        Add new attributes to the current class.
+
+        A new attributes of the current ClassDef.
```
yguclu, line 3383: Typos:
```suggestion
        Add a new attribute to the current class.

        Add a new attribute to the current ClassDef.
```

pyccel/ast/core.py
```diff
@@ -3371,8 +3376,56 @@ def attributes_as_dict(self):
...
+        if not isinstance(attr, Variable):
+            raise TypeError("Attributes must be Variables")
+        attr.set_current_user_node(self)
+        self._attributes += (attr,)
+
+    def add_new_method(self, method):
+        """
+        Add a new method to the current class.
+
+        A new method of the current ClassDef.
```
yguclu, line 3400: Typos:
```suggestion
        Add a new method to the current ClassDef.
```

pyccel/ast/core.py
```diff
@@ -3371,8 +3376,56 @@ def attributes_as_dict(self):
...
+        if not isinstance(method, FunctionDef):
+            raise TypeError("Method must be FunctionDef")
+        method.set_current_user_node(self)
+        self._methods += (method,)
+
+    def add_new_interface(self, interface):
+        """
+        Add a new interface to the current class.
+
+        A new interface of the current ClassDef.
```
yguclu, line 3417: Typo:
```suggestion
        Add a new interface to the current ClassDef.
```

pyccel/ast/core.py
```diff
@@ -3371,8 +3376,56 @@ def attributes_as_dict(self):
...
+
+        Parameters
+        ----------
+        interface : FunctionDef
+            The interface that will be added.
+        """
 
+        if not isinstance(interface, Interface):
+            raise TypeError("Interface must be Interface")
```
yguclu, line 3426: The error message is a bit cryptic. I suggest
```suggestion
            raise TypeError("Argument "interface" must be of type Interface")
```

pyccel/parser/semantic.py
```diff
@@ -1116,11 +1132,37 @@ def _assign_lhs_variable(self, lhs, d_var, rhs, new_expressions, is_augassign,ar
...
             # Variable not yet declared (hence array not yet allocated)
             if var is None:
 
+                if isinstance(lhs, DottedName):
+                    prefix_parts = lhs.name[:-1]
+                    syntactic_prefix = prefix_parts[0] if len(prefix_parts) == 1 else DottedName(*prefix_parts)
+                    prefix = self._visit(syntactic_prefix)
+                    class_def = prefix.cls_base
+                    if prefix.name == 'self':
+                        var      = self.get_variable('self')
```
yguclu, line 1146: Why the empty spaces?
```suggestion
                        var = self.get_variable('self')
```

pyccel/parser/semantic.py
```diff
@@ -1116,11 +1132,37 @@ def _assign_lhs_variable(self, lhs, d_var, rhs, new_expressions, is_augassign,ar
...
+                        new_name = class_def.scope.get_expected_name(attribute_name)
+                        # Create the attribute
+                        member = self._create_variable(new_name, dtype, rhs, d_lhs)
+
+                        # Insert the attribute to the class scope
+                        # Passing the original name ensures that the attribute can be found under this name
+                        class_def.scope.insert_variable(member, attribute_name)
+
+                        # Create the local DottedVariable
+                        lhs    = member.clone(member.name, new_class = DottedVariable, lhs = var)
```
yguclu, line 1159: Why the empty spaces?
```suggestion
                        lhs = member.clone(member.name, new_class = DottedVariable, lhs = var)
```

CHANGELOG.md
```diff
@@ -7,8 +7,12 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 
+-   #1484 : Using scope for classes to avoid name clashes.
+
 ### Changed
 
+-   #1484 : Improving handling of `DottedName` in `_assign_lhs_variable`.
```
yguclu, line 14: You should use the base form of the verb here:
```suggestion
-   #1484 : Use scope for classes to avoid name clashes.

### Changed

-   #1484 : Improve handling of `DottedName` in `_assign_lhs_variable`.
```

yguclu (changes requested):
Good job!

pyccel/ast/core.py
```diff
@@ -3323,6 +3323,11 @@ def name(self):
 
     @property
     def attributes(self):
+        """
+        The attributes of a class.
+
+        Returns a tuple of the attributes of a ClassDef.
+        """
```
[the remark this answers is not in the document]
sboof911: It's a `tuple` I deleted the Returns section cause I think it's already clear on the extended Summary.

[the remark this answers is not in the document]
yguclu: If you do not want a Returns section you need to write explicitly in the extended summary that the attributes are of type `Variable`.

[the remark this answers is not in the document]
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1484/commits/a064500ef52d3a1dfb5f010d9644e21acccb5b46.

sboof911:
/bot run pylint pyccel_lint spelling linux docs

yguclu (approved):
Good job!

Issue #1487, opened 2023-08-10
Type specification


EmilyBourne:
The current implementation of the type specification is a bit messy.

There are 3 possible methods.
1. Headers in comments:
    ```python
    #$ header function f2(int, int) results(int)
    ```

2. Types decorator
    ```python
    @types(int, int, results = int)
    def f2(x, y):
    ```
3. Type hinting
    ```python
    def f2(x : int, y : int) -> int:
    ```

The first method is the oldest, but the least familiar for the current devs. Most of the code hasn't been touched since it was written in 2018. A lot of the code for this section contains "TODO" comments.

The methods are also handled in different ways.
Method 1 is handled in the SemanticParser : 
https://github.com/pyccel/pyccel/blob/2cce6b0270aa312b0764ff7d7909a17cde992b53/pyccel/parser/semantic.py#L3197-L3199

Method 2 is converted in the SyntacticParser from a decorator to a header:
https://github.com/pyccel/pyccel/blob/2cce6b0270aa312b0764ff7d7909a17cde992b53/pyccel/parser/syntactic.py#L749-L756

Method 3 is handled in the SyntacticParser by creating a `@types` decorator:
https://github.com/pyccel/pyccel/blob/2cce6b0270aa312b0764ff7d7909a17cde992b53/pyccel/parser/syntactic.py#L644-L653
which is then converted to a header

While it is good that all three methods rely on the same underlying code it is not ideal that the underlying method is the least well known.  Further it seems that this is a missed opportunity to simplify the code as the type hints are associated with the relevant symbols. In contrast the headers must be broken up into parts in order to associate them with the arguments.

This complexity makes it difficult to handle issues such as #1336 .

We would like to move towards having one simple way of specifying types. Type hinting is preferred as it is already part of the Python library.

**Phase 1**
Prepare for breaking changes
- [x] Add Deprecation warnings when headers or types are used (#1488)
- [x] Remove the use of the `@types` decorator in our tests except for testing the decorator (#1470)

**Phase 2**
Ensure changes can be implemented
- [x] Ensure all supported structures can use type hinting (this is not yet the case for classes) (#1492)
- [x] Remove remaining `@types` decorators in tests
- [x] Remove remaining `@template` decorators in tests
- [x] Change default behaviour (#1517)

**Phase 3**
Breaking changes.
This should be begun once the deprecation warning has been present for **at least** one version.
- [ ] Remove explicit support for headers and types
[...]

PR #1488, opened 2023-08-10
Add warnings to Type specification deprecations

EmilyBourne:
Phase 1 of issue #1487 . `FutureWarning`s are added to the code to warn users to adapt their code before future breaking changes.

Should be merged after #1470

### DIFF
--- a/docs/templates.md
+++ b/docs/templates.md
@@ -1,76 +1,32 @@
 # Headers and Decorators
 
 ## Template
-### Templates using header comments
 A **template** in Pyccel, is used to allow the same function to take arguments of different types from a selection of types the user specifies.
 #### The usage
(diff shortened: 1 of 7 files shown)


docs/templates.md
```diff
@@ -2,26 +2,36 @@
...
 #### The usage
+In this example the argument **a**, could either be an integer or float, and the same for the argument **b**:
 ```python
-@template(name='T', types=['int','real'])
+@template(name='T', types=['int','float'])
+@template(name='Z', types=['int','float'])
 def f(a : 'T', b : 'T'):
 	pass
 ```
```
yguclu, line 14: If I understand this example correctly, the argument `b` should be of type `'Z'`.

yguclu (changes requested):
Looks good to me. I have just one comment.

yguclu (approved):

PR #1499, opened 2023-08-15
Fix passing temporary arrays to functions

### LINKED ISSUE #1126: Can't pass an unnamed array to a function in C

EmilyBourne:
Fix problems when passing temporary arrays to functions by widening the applicability of the `if` block which results in temporary variables being created for arrays. Fixes #1126

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ All notable changes to this project will be documented in this file.
 
 -   #682 : Wrong data layout when copying a slice of an array.
 -   #1453 : Fix error-level developer mode output.
+-   #1499 : Fix passing temporary arrays to functions.
 -   \[INTERNALS\] Fix string base class selection.
 
 ### Changed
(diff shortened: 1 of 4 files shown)


EniddeallA (approved):
Looks good!

yguclu (approved):
Looks good to me 👍

PR #1506, opened 2023-08-30
Fix none_dealloc bug

EmilyBourne:
Increment the counter on `Py_None` before returning `None` from functions which return nothing. This avoids unexpected deallocation, which would result in a fatal Python error.

cc: @pyccel/psydac-dev

### DIFF
--- a/.dict_custom.txt
+++ b/.dict_custom.txt
@@ -13,6 +13,7 @@ pyccelised
 pyccelising
 allocatable
 deallocate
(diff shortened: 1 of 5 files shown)


yguclu:
Psydac's PR [#326](https://github.com/pyccel/psydac/pull/326) shows the bug and proves that this PR fixes it.

yguclu (approved):

Issue #1493, opened 2023-08-10
Add import of a class type in the C-Python wrapper


EmilyBourne:
### Describe the feature
The most basic support should be added for classes allowing just the importation of the class type object.

### Test Code
Provide code which does not currently work but which should do when this issue is fixed:
```python
from module import class
```

### Proposed Solution
A `PyClassDef` class should be created in `pyccel/ast/cwrapper.py`. This object will be built in the class `CToPythonWrapper` in a new function `_visit_ClassDef`.

In the `CWrapperPrinter` the function `self._print_PyClassDef` should be added printing the following code from the information in the `PyClassDef` instance:
```c
typedef struct {
    PyObject_HEAD
    <class_name>* instance;
} Py<class_name>Object;

static PyTypeObject Py<class_name>Type = {
    PyVarObject_HEAD_INIT(NULL, 0)
    .tp_name = "<module_name>.<class_name>",
    .tp_doc = PyDoc_STR("<class_docstring>"),
    .tp_basicsize = sizeof(Py<class_name>Object),
    .tp_itemsize = 0,
    .tp_flags = Py_TPFLAGS_DEFAULT,
    .tp_new = PyType_GenericNew,
};
```

The function `CToPythonWrapper._build_module_exec_function` should be modified to add the new object `PyTypeObject Py<class_name>Type` to the module.

### Additional Context

This step is equivalent to the first step of this tutorial : https://docs.python.org/3/extending/newtypes_tutorial.html#the-basics

PR #1508, opened 2023-09-06
Add C support for a class destructor

sboof911:
Add a Setter for the `body` attribute in `FunctionDef`.
Add `__del__` Method to the Class even if not inserted.
Add a new boolean attribute called `is_freed` and initialize it to `False` in `__init__` method of the class.
In the `__del__` method, add a condition that checks if `is_freed` is `True` before freeing resources. This helps prevent double-freeing of resources.
Add the `Class Varibale constructor` to the `self._allocs` list.
Change `_visit_Del` to return a list of `Deallocate` instances.
Implement changes in `_print_Del` on C, Fortran, Python to accommodate the new changes.
In the `_print_Deallocate` method for C and Fortran, add logic to check whether the variable passed as an argument is of type `CustomDataType` and handle it accordingly.


pyccel/parser/semantic.py
```diff
@@ -3361,6 +3365,11 @@ def _visit_FunctionDef(self, expr):
             self._allocs.append([])
 
             # we annotate the body
+            if cls_name and expr.name == '__Pyccel__del__':
+                attributes = self.scope.find(cls_name, 'classes').attributes
+                attribute = []
+                attribute.extend(attr for attr in attributes if not attr.on_stack)
```
EmilyBourne, line 3371:
```suggestion
                attribute = [attr for attr in attributes if not attr.on_stack]
```

pyccel/parser/semantic.py
```diff
@@ -3361,6 +3365,11 @@ def _visit_FunctionDef(self, expr):
...
+                self._allocs.append(attribute) if len(attribute) > 0 else None
```
EmilyBourne, line 3372: Append? Not:

```suggestion
                self._allocs[-1].extend(attribute) 
```
?

pyccel/codegen/printing/ccode.py
```diff
@@ -2272,6 +2274,9 @@ def _print_Program(self, expr):
     def _print_CustomDataType(self, expr):
         return "struct " + expr.name
 
+    def _print_Del(self, expr):
+        return ""
```
EmilyBourne, line 2278: This should call `__del__` to free memory (although maybe an issue should be opened for this as `del` probably ought to be translated to `Deallocate` in the semantic stage

pyccel/parser/semantic.py
```diff
@@ -487,10 +487,17 @@ def _garbage_collector(self, expr):
...
-        else:
-            deallocs = []
+            for i in self._allocs[-1]:
+                if isinstance(i, DottedVariable):
+                    if isinstance(i.lhs.dtype, CustomDataType) and not self._current_function == '__Pyccel__del__':
+                        continue
+                if isinstance(i, ConstructorCall):
+                    deallocs.append(Deallocate(i.cls_variable))
+                    continue
+                deallocs.append(Deallocate(i))
```
EmilyBourne, line 500: I find this more readable:
```suggestion
                else:
                    deallocs.append(Deallocate(i))
```

pyccel/parser/semantic.py
```diff
@@ -2389,6 +2396,7 @@ def _visit_FunctionCall(self, expr):
             # TODO treat parametrized arguments.
 
             expr = ConstructorCall(method, args, cls_variable)
+            self._allocs.append([expr])
```
EmilyBourne, line 2399: Why not append `cls_variable` directly? Why does this need to be a list?

pyccel/parser/semantic.py
```diff
@@ -3388,6 +3396,10 @@ def _visit_FunctionDef(self, expr):
             self._allocs.append([])
 
             # we annotate the body
+            if cls_name and expr.name == '__Pyccel__del__':
```
EmilyBourne, line 3399: Where does the `__Pyccel` come from? Do we have 2 freeing functions?

pyccel/codegen/printing/ccode.py
```diff
@@ -1444,6 +1444,9 @@ def _print_Deallocate(self, expr):
         variable_address = self._print(ObjectAddress(expr.variable))
         if expr.variable.is_alias:
             return f'free_pointer({variable_address});\n'
+        if isinstance(expr.variable.dtype, CustomDataType):
+            Pyccel__del = next((method.name for method in expr.variable.cls_base.methods if method.name.endswith('__Pyccel__del__')), None)
+            return(f"{Pyccel__del}({variable_address});\n")
```
EmilyBourne, line 1449: Why is the IterationError  suppressed? Won't the code fail if `None` is used anyway? Maybe the destructor should be saved explicitly in the ClassDef similarly to what is done in `Module` to make it easy to find?

pyccel/parser/semantic.py
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1508/commits/4464522edd12b45c32f6d7ee46ababb878edcd2f.

pyccel/parser/semantic.py
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1508/commits/4464522edd12b45c32f6d7ee46ababb878edcd2f.

pyccel/parser/semantic.py
sboof911: Fixed f31858d2e5402008754c6cf25de3381f8c0bca94.

pyccel/codegen/printing/ccode.py
sboof911: Fixed f31858d2e5402008754c6cf25de3381f8c0bca94.

pyccel/codegen/printing/ccode.py
sboof911: Fixed f31858d2e5402008754c6cf25de3381f8c0bca94.

pyccel/ast/core.py
```diff
@@ -3331,6 +3332,22 @@ def attributes(self):
...
+        Returns True if the class instance has been deallocated, False otherwise.
+        """
+
+        return self._is_deallocated
+
+    @is_deallocated.setter
+    def is_deallocated(self, value):
+        if not isinstance(value, bool):
+            raise TypeError("is_deallocated must be a boolean (Bool)")
+        self._is_deallocated = value
```
EmilyBourne, line 3349: I expected this variable to appear in the generated code not the Pyccel code. How do you handle if blocks?

pyccel/codegen/printing/ccode.py
```diff
@@ -2272,6 +2275,15 @@ def _print_Program(self, expr):
     def _print_CustomDataType(self, expr):
         return "struct " + expr.name
 
+    def _print_Del(self, expr):
```
EmilyBourne, line 2278: Can you explain the difference between `Del` and `Deallocate`? For me they are conceptually the same. I think it would make more sense to convert `Del` nodes to `Deallocate` nodes in the semantic stage and not have a `_print_Del` here. This will avoid duplication/confusion

pyccel/codegen/printing/ccode.py
```diff
@@ -2272,6 +2275,15 @@ def _print_Program(self, expr):
...
+        for variable in expr.variables:
+            if isinstance(variable.dtype, CustomDataType) and not variable.cls_base.is_deallocated:
+                variable_address = self._print(ObjectAddress(variable))
+                Pyccel__del = next(method.name for method in variable.cls_base.methods if method.name.endswith('__del__'))
+                variable.cls_base.is_deallocated = True
+                return(f"{Pyccel__del}({variable_address});\n")
```
EmilyBourne, line 2284: The return is too early. What if there are multiple class objects being deleted?

EmilyBourne, line 2284: Del can be called on anything, the free function should be called (via `Deallocate`) if `del` is called on an array.

pyccel/parser/semantic.py
```diff
@@ -3185,7 +3190,8 @@ def _visit_Return(self, expr):
 
         # add the Deallocate node before the Return node and eliminating the Deallocate nodes
         # the arrays that will be returned.
-        code = assigns + [Deallocate(i) for i in self._allocs[-1] if i not in results]
+        code = assigns + [Deallocate(i if not isinstance(i, ConstructorCall) else i.cls_variable) for i in self._allocs[-1] \
+                            if (i if not isinstance(i, ConstructorCall) else i.cls_variable) not in results]
```
EmilyBourne, line 3194: Is this change still necessary?

pyccel/parser/syntactic.py
```diff
@@ -853,6 +853,13 @@ def _visit_ClassDef(self, stmt):
...
             elif isinstance(i, ast.Pass):
                 return errors.report(UNSUPPORTED_FEATURE_OOP_EMPTY_CLASS, symbol = stmt, severity='error')
+        if not any(method.name == '__del__' for method in methods):
+            header = hdr_parse(stmts=f"#$ header method __del__({name})")
+            self.scope.parent_scope.insert_header(header)
+            arguments = ast.arguments(args=[ast.arg(arg='self')], kwonlyargs=[], defaults=[])
+            body = [Pass()]
+            del_functiondef = ast.FunctionDef(name="__del__", args=arguments, body=body, decorator_list=[])
+            methods.append(self._visit(del_functiondef))
         for i in methods:
```
EmilyBourne, line 863: I don't suppose you can think of a way to write this in the semantic stage? Or avoid the header? I am working to deprecate these headers (see #1487 , #1523) so I don't want to introduce new ones in parallel.

tests/pyccel/test_pyccel.py
```diff
@@ -882,6 +882,8 @@ def test_classes_f_only( test_file , language):
 @pytest.mark.parametrize( "test_file", ["scripts/classes/classes_2_C.py",
                                         "scripts/classes/classes_5.py",
                                         "scripts/classes/classes_3.py",
+                                        "scripts/classes/classes_1.py",
+                                        "scripts/classes/classes.py",
```
EmilyBourne, line 886: If these are handled, do you still need the function `test_classes_f_only`?

EmilyBourne (changes requested):

pyccel/parser/semantic.py
sboof911: Fixed 0dbb55d7043fb5c0772d201935163c592e7097eb.

pyccel/codegen/printing/ccode.py
sboof911: Fixed 0dbb55d7043fb5c0772d201935163c592e7097eb.

pyccel/parser/syntactic.py
sboof911: Fixed 13152019b5509fd4eb35fde7c5f9dff374192d56.

pyccel/ast/core.py
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1508/commits/23157f9477e3bb5c71b413760b2855a1d08bcb5d.

EmilyBourne:
Beware, I have updated your branch. Some of the keys for dictionaries such as `dtype_registry` have changed. I don't think this should affect this PR, but maybe check the tests still pass just in case.

tests/pyccel/test_pyccel.py
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1508/commits/e0c2cdfe2d24bbbd64c6207157efaa365f2cdb50.

pyccel/ast/core.py
```diff
@@ -2569,6 +2569,14 @@ def body(self):
         """ CodeBlock containing all the statements in the function """
         return self._body
 
+    @body.setter
+    def body(self, body):
+        if iterable(body):
+            body = CodeBlock(body)
+        elif not isinstance(body, CodeBlock):
+            raise TypeError('body must be an iterable or a CodeBlock')
+        self._body = body
```
EmilyBourne, line 2578:
```suggestion
        self._body.invalidate_node()
        self._body = body
        self._body.set_current_user_node(self)
```

pyccel/codegen/printing/ccode.py
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1508/commits/da631a38dd2c3cc6cc44aa5c9962d3d747ccaa36.

pyccel/parser/semantic.py
```diff
@@ -3608,13 +3621,35 @@ def _visit_ClassDef(self, expr):
...
+            self.insert_function(del_method)
+            cls.add_new_method(del_method)
+
+        for method in cls.methods:
+            if method.name == '__del__':
+                self._current_function = method.name
+                attribute = [attr for attr in cls.attributes if not attr.on_stack]
+                if attribute:
+                    self._allocs[-1].extend(attribute)
+                    is_freed = next(is_freed for is_freed in cls.attributes if is_freed.name.startswith("is_freed"))
```
EmilyBourne, line 3638: You should still have access to `deallocater_rhs` here. You can use it directly instead of searching for it. This will avoid any potential problems due to name collisions

pyccel/parser/semantic.py
```diff
@@ -3608,13 +3621,35 @@ def _visit_ClassDef(self, expr):
...
+
         self.exit_class_scope()
 
         return EmptyNode()
 
     def _visit_Del(self, expr):
 
-        ls = [self._visit(i) for i in expr.variables]
+        ls = [Deallocate(self._visit(i)) for i in expr.variables]
         return Del(ls)
```
EmilyBourne, line 3653: Why not return a `CodeBlock`? That way you can completely avoid the need for `_print_Del`

pyccel/parser/semantic.py
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1508/commits/fe0041d898f4fc1f33be35c56f9532bf83f884f7.
The condition was in the beggining just if their is some attributes of the class that should be deleted.
But now the check is on all the condition even if the `__del__` function is empty.

pyccel/ast/core.py
EmilyBourne: @sboof911 Sorry I just noticed that I had a line wrong here

pyccel/parser/semantic.py
EmilyBourne: In the Python printer it is important to keep `Del` objects as these should be printed while `Deallocate` objects shouldn't be

pyccel/ast/core.py
```diff
@@ -2569,6 +2569,17 @@ def body(self):
         """ CodeBlock containing all the statements in the function """
         return self._body
 
+    @body.setter
+    def body(self, body):
+        if iterable(body):
+            body = CodeBlock(body)
+        elif not isinstance(body, CodeBlock):
+            raise TypeError('body must be an iterable or a CodeBlock')
+        self._body.invalidate_node()
```
EmilyBourne, line 2578:
```suggestion
        self._body.remove_user_node(self)
```

@sboof911 This is the line I wrote wrong

pyccel/ast/core.py
sboof911: Fixed.

CHANGELOG.md
```diff
@@ -9,7 +9,7 @@ All notable changes to this project will be documented in this file.
 -   #1472 : Add C printing support for a class containing scalar data.
 -   #1492 : Types of arguments for class methods can be declared like function arguments.
 -   #1511 : Add support for the `cmath` library.
--   #1508 : Add C support for a class destructor.
+-   #1508 : Add C support for a class `Destructor`.
```
EmilyBourne, line 12: This word is valid. It is not code. Please change this back and add it to the file `.dict_custom.txt`

sboof911: Fixed https://github.com/pyccel/pyccel/pull/1508/commits/f0373f446e6206f5fcbc5676ce6f0d28d9cbff05.

sboof911:
/bot run spelling pyccel_lint pylint docs linux

pyccel/ast/core.py
```diff
@@ -2566,9 +2566,23 @@
...
+
+        Returns a CodeBlock contains all the statements in the function.
+        """
         return self._body
 
+    @body.setter
+    def body(self, body):
+        if iterable(body):
+            body = CodeBlock(body)
+        elif not isinstance(body, CodeBlock):
```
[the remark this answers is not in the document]
sboof911: /bot accept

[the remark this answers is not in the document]
sboof911: Body protection.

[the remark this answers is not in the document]
EmilyBourne: I think the line it is complaining about is 2579.

pyccel/ast/core.py
```diff
@@ -2566,9 +2566,23 @@ def results(self):
...
```
EmilyBourne, line 2580: I don't think this should be an if/elif block. Shouldn't it be:
```suggestion
        if iterable(body):
            body = CodeBlock(body)
        if not isinstance(body, CodeBlock):
```
? As is you don't save the result if it is an iterable

pyccel/ast/core.py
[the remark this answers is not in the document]
EmilyBourne: I am not sure that this error should be ignored. I don't think anything is saved if you pass an iterable. This should have been noticed in the tests

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ All notable changes to this project will be documented in this file.
 -   #1472 : Add C printing support for a class containing scalar data.
 -   #1492 : Types of arguments for class methods can be declared like function arguments.
 -   #1511 : Add support for the `cmath` library.
+-   #1508 : Add C support for a class destructor.
```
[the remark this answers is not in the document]
EmilyBourne:
```suggestion
-   #1508 : Add C support for a class destructor.
-   #1508 : Add support for array data in classes.
```
and in the fixed section:
```
-  #1508 : Generate deallocations for classes and their attributes.
```

pyccel/parser/semantic.py
```diff
@@ -3598,6 +3619,14 @@ def _visit_ClassDef(self, expr):
                 self._visit_FunctionDef(method)
                 methods.pop(i)
                 init_func = self.scope.functions.pop(m_name)
+
+                # creat a new attribute to check allocation
```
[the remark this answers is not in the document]
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1508/commits/94a707d2b7f51287350a7d4b949f7f9c3385be4e.

CHANGELOG.md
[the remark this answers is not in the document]
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1508/commits/94a707d2b7f51287350a7d4b949f7f9c3385be4e.

pyccel/ast/core.py
sboof911: The bot is complaining about the line 2580.
Cause in : https://github.com/pyccel/pyccel/blob/8c3ba1f6b90f7d3d5b45a70ff3d7c76037dfa28c/pyccel/parser/semantic.py#L3731-L3731
the body passed is an iterable.
The condition should stay an if/elif cause it access the first condition, so the second condition is always True in this case.

EmilyBourne:
> > Well done. This looks really good. And it is great to finally get the memory leaks fixed. I just have a few comments.
> > Also could you add a leak test. Similar to here: https://github.com/pyccel/pyccel/blob/devel/.github/actions/valgrind_run/action.yml https://github.com/pyccel/pyccel/blob/devel/tests/ndarrays/leaks_check.py
> > to show that we can now put arrays in classes and they will be correctly deleted
> 
> I don't think we can check leaks until we implemente the wrapper for classes

Ah, you may be right. That is a shame. Once the first step of the wrapper (#1493) is completed this should be possible. #1493 should be a reasonably small PR so I think we have 3 options:
1. I do #1493 quickly tonight
2. You start working on #1493 and we come back to this ASAP
3. You put this in a separate file, run pyccel with the `-t` option to avoid the wrapping, and write the compilation commands to compile the file manually rather than using pyccel.

pyccel/ast/core.py
EmilyBourne: Ah, sorry I misread this. For some reason I thought we had:
```python
        if iterable(body):
            body = CodeBlock(body)
        elif not isinstance(body, CodeBlock):
            self._body.remove_user_node(self)
            self._body = body
            self._body.set_current_user_node(self)
```
In that case I agree. This is fine :+1:

pyccel/parser/semantic.py
```diff
@@ -3682,13 +3709,36 @@ def _visit_ClassDef(self, expr):
...
+        if not any(method.name == '__del__' for method in methods):
+            argument = FunctionDefArgument(Variable(cls.name, 'self', cls_base = cls))
+            scope = self.create_new_function_scope('__del__')
+            del_method = FunctionDef('__del__', [argument], (), [Pass()], cls_name=cls.name, scope=scope)
+            self.exit_function_scope()
+            self.insert_function(del_method)
+            cls.add_new_method(del_method)
+
+        if not self._allocs:
+            self._allocs.append([])
```
EmilyBourne, line 3721: Can you explain this line? Why do we not systematically add a scope for the `ClassDef` variables? Won't this implementation potentially result in the enclosing scope's `_allocs` being popped in `self._garbage_collector`?

E.g. The following will fail:
```python
from numpy import ones

class Point(object):
    def __init__(self : 'Point', n : 'int'):
        self.y = ones(n)

    def __del__(self : 'Point'):
        pass

x = ones(4)
```
please add this as a test and fix this bug

sboof911:
> > > Well done. This looks really good. And it is great to finally get the memory leaks fixed. I just have a few comments.
> > > Also could you add a leak test. Similar to here: https://github.com/pyccel/pyccel/blob/devel/.github/actions/valgrind_run/action.yml https://github.com/pyccel/pyccel/blob/devel/tests/ndarrays/leaks_check.py
> > > to show that we can now put arrays in classes and they will be correctly deleted
> > 
> > 
> > I don't think we can check leaks until we implemente the wrapper for classes
> 
> Ah, you may be right. That is a shame. Once the first step of the wrapper (#1493) is completed this should be possible. #1493 should be a reasonably small PR so I think we have 3 options:
> 
> 1. I do [Add import of a class type in the C-Python wrapper #1493](https://github.com/pyccel/pyccel/issues/1493) quickly tonight
> 2. You start working on [Add import of a class type in the C-Python wrapper #1493](https://github.com/pyccel/pyccel/issues/1493) and we come back to this ASAP
> 3. You put this in a separate file, run pyccel with the `-t` option to avoid the wrapping, and write the compilation commands to compile the file manually rather than using pyccel.

I think the first option is the best one so I also have an idea how does things work so I could finish the other issues more easily.

pyccel/parser/semantic.py
```diff
@@ -3682,13 +3709,36 @@ def _visit_ClassDef(self, expr):
...
+
+        for method in cls.methods:
+            if method.name == '__del__':
+                self._current_function = method.name
+                attribute = [attr for attr in cls.attributes if not attr.on_stack]
+                if attribute:
+                    self._allocs[-1].extend(attribute)
+                    method.body.insert2body(*self._garbage_collector(method.body))
+                    if not self._allocs:
+                        self._allocs.append([])
```
EmilyBourne, line 3728: I am surprised that this fixes the issue. Can you explain why you think this `if` is necessary.

Or why you made this change:
https://github.com/pyccel/pyccel/pull/1508/commits/94a707d2b7f51287350a7d4b949f7f9c3385be4e

EmilyBourne (approved):
Did you check if this fixes any of the cases in tests/codegen/fcode/scripts

sboof911:
/bot run spelling pyccel_lint pylint docs linux

sboof911:
/bot run spelling pyccel_lint pylint docs linux

EmilyBourne:
Maybe it is simpler to add the leak test later. I think the code should be safe. Can you open an issue for it so we don't forget

EmilyBourne:
@sboof911 can you update the version on this branch please (to 1.9.2)

pyccel/ast/core.py
```diff
@@ -2566,9 +2566,23 @@ def results(self):
 
     @property
     def body(self):
-        """ CodeBlock containing all the statements in the function """
+        """
+        CodeBlock containing all the statements in the function.
+
+        Returns a CodeBlock contains all the statements in the function.
```
yguclu, line 2572: Typo: contains -> containing
I would also prefer "Return" to "Returns"

tests/pyccel/scripts/classes/classes_4.py
```diff
@@ -0,0 +1,17 @@
...
+import numpy as np
+
+y = np.ones(4)
+
+class A:
+    def __init__(self : 'A', n : int):
+        self.x = np.ones(n)
+
+    def __del__(self : 'A'):
+        del self.x
```
yguclu, line 12: This `__del__` method is essentially identical to the default one. Maybe we can have it do something special, like printing a message?

yguclu (changes requested):

pyccel/ast/core.py
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1508/commits/63d09e3bcdf9f18da4477da007cc16875c9cb92a.

tests/pyccel/scripts/classes/classes_4.py
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1508/commits/f414979b3676813938fa7478ad469f071b8c8942.

yguclu (approved):

PR #1510, opened 2023-09-07
Add support for type annotations to class methods

### LINKED ISSUE #1492: Add support for type hinting to classes
### Describe the bug
When classes were implemented the only way to specify types in Pyccel was via headers. Before writing the class headers must be used to declare that the class will exist and to declare its methods. E.g:

```python
#$ header class Point(public)
#$ header method __init__(Point, float)
#$ header method __del__(Point)
#$ header method translate(Point, float)
```

since then Pyccel has added support for type hinting (and the `@types` decorator). However the class support was too minimal for it to be given the same attention. As a result headers are still the only way of specifying types in classes. This should be fixed before adding documentation.

The `@types` decorator will be deprecated in the future so this can be ignored.

### Test Code
I would like to be able to write the following code:
```python
class Point:
    def __init__(self : 'Point', x : float):
        self.x = x

    def __del__(self : 'Point'):
        pass

    def translate(self : 'Point', a : float):
        self.x[:]   =  self.x + a
```
or (given that the type of the first variable cannot be modified):
```python
class Point:
        self.x = x

    def __del__(self):
        pass

    def translate(self, a : float):
        self.x[:]   =  self.x + a
```

The start of the class should be sufficient to replace `#$ header class Point(public)`. Currently types hints are handled by converting the hints to functions as described in #1487 . Changing this will be handled in a different issue. For this issue it is sufficient to build and insert the expected header into the `FunctionDef`

### Language
All languages, this problem is in the syntactic/semantic level

### Additional context
See #1487

EmilyBourne:
Remove the need for a class header declaration by creating the appropriate datatype at the class declaration in the Python code. Change the generated header syntax for class methods. This allows the use of the `@types` decorator or type annotations for class methods (fixes #1492).
Clean up tests to use type annotations. Delete duplicate tests and context test which doesn't use contexts correctly


pyccel/parser/syntactic.py
```diff
@@ -750,11 +750,14 @@ def fill_types(ls):
...
-                txt  = '#$ header ' + name
-                txt += '(' + ','.join(types) + ')'
+                txt  = '#$ header '
+                if len(self._context) > 1 and isinstance(self._context[-2], ast.ClassDef):
+                    txt += 'method '
+                txt += name + '(' + ','.join(types) + ')'
 
                 if results:
                     txt += ' results(' + ','.join(results) + ')'
+                print(txt)
```
sboof911, line 760: Is this necessary ?

sboof911 (changes requested):
This looks perfect.

pyccel/parser/syntactic.py
EmilyBourne: Oops. Good spot I was trying (unsuccessfully) to debug and forgot about it

EmilyBourne, line 760:
```suggestion
```

sboof911 (approved):
Perfect.

yguclu (approved):
Nice job!

Issue #1182, opened 2022-08-16
inhomogeneous tuples breaks pyccel


azouiten:
**Describe the bug**
If you give to the tuple arguments that only differ in rank it considers it a homogeneous tuple but then Pyccel breaks when it tries to treat it as such.

**To Reproduce**
Code to reproduce the behavior:
```python
if __name__ == "__main__":
    a = (1, [1, 2, 3])
```

The error message:
```bash
Traceback (most recent call last):
  File "/Users/azouiten/Desktop/internship/pyccel-env/bin/pyccel", line 33, in <module>
    sys.exit(load_entry_point('pyccel', 'console_scripts', 'pyccel')())
  File "/Users/azouiten/Desktop/internship/pyccel/pyccel/commands/console.py", line 239, in pyccel
    execute_pyccel(filename,
  File "/Users/azouiten/Desktop/internship/pyccel/pyccel/codegen/pipeline.py", line 238, in execute_pyccel
    parser.annotate(**settings)
  File "/Users/azouiten/Desktop/internship/pyccel/pyccel/parser/parser.py", line 160, in annotate
    parser = SemanticParser(self._syntax_parser,
  File "/Users/azouiten/Desktop/internship/pyccel/pyccel/parser/semantic.py", line 216, in __init__
    self.annotate()
  File "/Users/azouiten/Desktop/internship/pyccel/pyccel/parser/semantic.py", line 264, in annotate
    ast = self._visit(ast, **settings)
  File "/Users/azouiten/Desktop/internship/pyccel/pyccel/parser/semantic.py", line 1453, in _visit
    obj = getattr(self, annotation_method)(expr, **settings)
  File "/Users/azouiten/Desktop/internship/pyccel/pyccel/parser/semantic.py", line 1465, in _visit_Module
    body = self._visit(expr.program).body
  File "/Users/azouiten/Desktop/internship/pyccel/pyccel/parser/semantic.py", line 1453, in _visit
    obj = getattr(self, annotation_method)(expr, **settings)
  File "/Users/azouiten/Desktop/internship/pyccel/pyccel/parser/semantic.py", line 1677, in _visit_CodeBlock
    line = self._visit(b, **settings)
  File "/Users/azouiten/Desktop/internship/pyccel/pyccel/parser/semantic.py", line 1453, in _visit
    obj = getattr(self, annotation_method)(expr, **settings)
[...]

PR #1535, opened 2023-10-09
Clean up backtrace error

EmilyBourne:
The error reported in #1182 leads to a nasty backlog. This PR does not fix the reported issue, but it make things cleaner so a `NotImplementedError` is raised rather than a backlog


sboof911 (approved):
Looks Good.

pyccel/ast/builtins.py
```diff
@@ -468,11 +478,14 @@ def __init__(self, *args):
...
+            max_rank = max(a.rank for a in args)
+            self._rank      = max_rank + 1
             self._dtype     = NativeGeneric()
             self._precision = 0
             if self._rank == 1:
                 self._shape     = (LiteralInteger(len(args)), )
+            elif any(a.rank != max_rank for a in args):
+                self._shape     = (LiteralInteger(len(args)), ) + (None,)*(self._rank-1)
             else:
                 self._shape     = (LiteralInteger(len(args)), ) + args[0].shape
```
yguclu, line 490: It is not clear to me how these changes solve issue #1182. Can you please explain?

yguclu: Sorry I wrote too quickly. My question was about the neater error.

yguclu: So the problem was that we were trying to concatenate a tuple with `None`, correct?

yguclu, line 490: Would this not work then?
```suggestion
            else:
                extra_dims  = args[0].shape or (None,) * (self._rank - 1)
                self._shape = (LiteralInteger(len(args)), ) + extra_dims
```

EmilyBourne: Yes exactly

EmilyBourne: I don't think so. If we have:
```python
a = ((1,2), ((3,4),(5,6))
```
Then your solution will give:
```
rank = 3
shape = (2,2)
```
whereas mine will give:
```
rank = 3
shape = (2,None,None)
```

Maybe the mismatch between the rank and the shape would be enough to trigger a neat error but I am not sure. This would need to be tested

EmilyBourne: I've just checked. Your solution raises an `AssertionError` with the example I wrote, while the implemented error raises a neat `NotImplementedError`:
```
pyccel:
 |fatal [semantic]: TUPLE_WITH_INCOHERENT_RANKS.py [3,4]| Pyccel has encountered syntax that has not been implemented yet. Please create an issue at https://github.com/pyccel/pyccel/issues and provide a small example of your problem. Do not forget to specify your target language
Cannot infer shape of right-hand side for expression a = ((1, 2), ((1, 2), (3, 4)))
```

yguclu: There is a missing parenthesis in your example, or an extra one.

EmilyBourne: Oops. Fixed

yguclu: All right, I see the problem now!

yguclu (approved):

Issue #1536, opened 2023-10-10
`DottedVariable`s allocated to the wrong size.


EmilyBourne:
### Describe the bug
`Allocate` nodes which allocate `DottedVariable`s contain the wrong allocation size.

### To Reproduce
Provide code to reproduce the behavior:
```python
import numpy as np

class A:
    def __init__(self : 'A', n : int):
        self.x = np.ones(n)

    def __del__(self : 'A'):
        del self.x
```

### Error details
Provide the generated code, or the error message:
```fortran
module init_unknown_size


  use, intrinsic :: ISO_C_Binding, only : f64 => C_DOUBLE , i64 => &
        C_INT64_T
  implicit none

  type, public :: A
    real(f64), allocatable :: x(:)

    contains
    procedure :: create => A_create
    procedure :: free => A_free

  end type A

  contains

  !........................................
  subroutine A_create(self, n)

    implicit none

    class(A), intent(inout) :: self
    integer(i64), value :: n

    allocate(self%x(0:size(x, kind=i64) - 1_i64))
    self%x = 1.0_f64
    if (allocated(self%x)) then
      deallocate(self%x)
    end if

  end subroutine A_create
  !........................................

  !........................................
  subroutine A_free(self)

    implicit none

    class(A), intent(inout) :: self

    deallocate(self%x)

  end subroutine A_free
  !........................................

end module init_unknown_size
```

### Expected behaviour
The array should be allocated with `allocate(self%x(0:n - 1_i64))`

### Language
Fortran (Problem in semantic stage so it will also affect C)

### Additional context
This is due to the clone call here:
https://github.com/pyccel/pyccel/blob/125563cc6f25ed2ae9512618e1f2007621d9ade8/pyccel/parser/semantic.py#L1161

The function `Variable.clone` does not preserve the `alloc_shape`.

PR #1537, opened 2023-10-10
Fix Variable.clone function

EmilyBourne:
Fix the `Variable.clone` function. This fixes the problem described in #1536. This problem is recent so it does not need to go into the CHANGELOG. The changes will be covered by the descriptions in #1508.

A test cannot be added due to #1538 .


sboof911 (approved):
Look Good.

PR #1543, opened 2023-10-12
Fix devel branch tests

EmilyBourne:
Version strings shouldn't be quoted


EmilyBourne:
Tests failing on Windows with anaconda due to textx. I have opened an issue here: https://github.com/textX/textX/issues/411

yguclu (approved):

PR #1546, opened 2023-10-13
Bot bug fixes for collaborators on forks

EmilyBourne:
Don't try to examine commits from forks to ensure that tests are not rerun as the bot doesn't pull these commits (and shouldn't do for security reasons). Correct use of trust_givers


yguclu (approved):
Looks good

Issue #1540, opened 2023-10-11
OpenMP bug with Intel compiler


EmilyBourne:
### Describe the bug
The test `omp_range_sum_critical` fails when using an Intel compiler

### To Reproduce
Provide code to reproduce the behavior:
```python
def omp_range_sum_critical(x : 'int'):
    func_result = 0
    #$ omp parallel for num_threads(4) shared(func_result)
    for i in range(0, x):
        #$ omp critical
        func_result += i
        #$ omp end critical
    return func_result
```

### Error details
```
>           assert f1(x) == openmp.omp_range_sum_critical(x)
E           assert 44100 == 100128
E            +  where 44100 = <built-in function omp_range_sum_critical>(448)
E            +  and   100128 = <function omp_range_sum_critical at 0x7f338e9a4d30>(448)
E            +    where <function omp_range_sum_critical at 0x7f338e9a4d30> = openmp.omp_range_sum_critical
```

### Expected behavior
It appears that the critical section is not correctly managed. This test should pass

### Language
C

### Additional context
Compiler : Intel OneAPI.
OS : Linux, Ubuntu Jammy

EmilyBourne:
This appears to be fixed now. It must have been a compiler bug

Issue #542, opened 2020-11-13
Test different compilers on travis


EmilyBourne:
**Description**
Currently we only compile with gnu in our tests however we claim to support other compilers (e.g. intel, pgi) we should have tests to confirm that we use the correct flags/etc for these compilers

EmilyBourne:
See [here](https://github.com/CFMIP/COSPv2.0/actions/runs/289092756/workflow) for an example

EmilyBourne:
@yguclu Is this something we still want to do? If so where? I imagine this should be run on the devel branch only?

This should be able to be added pretty quickly as the work has already been done for intel in the benchmarks repo https://github.com/pyccel/pyccel-benchmarks/pull/23
and for `nvhpc` in the cuda repo

PR #1547, opened 2023-10-13
Add Intel tests to devel branch

EmilyBourne:
Add the configuration files to run tests with an Intel compiler. To this end add an environment variable `PYCCEL_DEFAULT_COMPILER` which specifies the default compiler.

There are 3 failing tests due to:
- passing NaN as an argument
- #1540 

Additionally some array tests were modified to use `isclose` for comparisons instead of `np.all_equal` these tests all involve divisions and should have been using `isclose` anyway.

This is related to #542 but does not fully fix the issue as it only adds Intel.


EmilyBourne:
Intel installation tested here:
https://github.com/EmilyBourne/pyccel/actions/workflows/intel.yml

PR #1563, opened 2023-10-16
Fix coverage trigger

### LINKED ISSUE #1560: Coverage report won't auto-trigger on forks
### Describe the bug
The coverage test must be triggered after the linux test. The `clean_up` workflow determines the PR using the commit. As commits from forks are not available to the bot this causes the workflow to fail. As a result the coverage is never triggered. A workaround is to trigger the coverage manually **after** the linux test has completed

EmilyBourne:
Simplify get_pr_ids with a method that handles forks. Instead of investigating available branches, get all open PRs and compare the SHAs directly. Fixes #1560 

Tested on https://github.com/EmilyBourne/pyccel


sboof911 (approved):
Looks Good.

PR #1574, opened 2023-10-19
Add type annotations for variable

### LINKED ISSUE #1509: Class attribute PEP annotation
Hey, very useful project! It would be great if you could consider [PEP526](https://peps.python.org/pep-0526/#class-and-instance-variable-annotations) syntax for this. E.g.
```python
class Point:
    x: int
    y: int
```

_Originally posted by @adam-urbanczyk in https://github.com/pyccel/pyccel/issues/1492#issuecomment-1697887006_

### LINKED ISSUE #1528: Support type annotations for variables
### Relevant Discussions

https://github.com/pyccel/pyccel/discussions/1462
https://github.com/pyccel/pyccel/issues/1487#issuecomment-1697899203
https://github.com/pyccel/pyccel/discussions/1489

### Describe the feature
As we move towards better support for Python's typing module. It is important to be able to have type annotations for variables. This will be important for issue #297 especially as we phase out header comments.

The valid type annotations should be the same as those already used for function argument annotations.

### Test Code
```python
a : int = 10
```

### Proposed Solution
The implementation is inspired by #1524 . In the syntactic stage we should save annotated variables into the new `AnnotatedPyccelSymbol` class. The functions added in #1524 should take care of the rest, but special handling may be required for annotated variables in assignments.

Ideally there should be a way to keep a const Variable grouped with its initialiser such that it can be printed this way in the future headers:
```python
a : 'const int' = 10
```

EmilyBourne:
Add type annotations for variables. Fixes #1528 . Also handle type annotations appearing in classes. Errors are raised for class variables, and class instances declared with PEP526 are handled. Fixes #1509 .

**Commit Summary**
- Move `AnnotatedPyccelSymbol` from `ast.internals` to `ast.variables` to avoid circular imports (`AnnotatedPyccelSymbol` depends on the same modules as a `Variable`: datatypes, internals, etc and is needed in the same places)
- Allow `DottedName` as `name` of a `AnnotatedPyccelSymbol`.
- Add `is_const` to `IndexedElement` to allow simpler constness check
- Improve error message for missing type annotations.
- Allow `_visit_AnnotatedPyccelSymbol` to create a `DottedVariable`
- Handle `AnnotatedPyccelSymbol` in class constructor call and assignment
- Move const check from `_visit_FunctionDef` to `_visit_Assign`
- Make const check non-blocking
- Add support for PEP526 annotations in classes (fixes #1509 )
- Raise errors for unsupported syntax in classes
- Add tests


tests/pyccel/scripts/classes/class_variables.py
```diff
@@ -0,0 +1,15 @@
+# pylint: disable=missing-function-docstring, missing-module-docstring, missing-class-docstring
+
+class A:
+    x : int
+    def __init__(self : 'A'):
+        self.x = 3
+        self.z : float = 10.0
+
+    def get_4(self : 'A'):
+        return 4
```
yguclu, line 10: Is the annotation of `self` mandatory or optional?

pyccel/ast/variable.py
```diff
@@ -883,48 +901,40 @@ def __init__(
...
-            self._shape = None if self._rank == 0 else tuple(new_shape)
-        else:
-            new_rank = rank
-            for i in range(rank):
-                if not isinstance(args[i], Slice):
-                    new_rank -= 1
-            self._rank = new_rank
+        new_shape = []
+        from .mathext import MathCeil
+        for a,s in zip(indices, shape):
```
yguclu, line 917: How do we know that `shape is not None`?

tests/pyccel/scripts/classes/class_variables.py
EmilyBourne: Currently it is mandatory. See #1578

pyccel/ast/variable.py
EmilyBourne: I spotted this thanks to the coverage results. The shape is set here:
https://github.com/pyccel/pyccel/blob/267be30007685fb23fa799be48c782fb144e46eb/pyccel/ast/variable.py#L198

And in the semantic stage an `IndexedElement` can only be called on an object with rank>1 (otherwise the Python code wouldn't be valid)

yguclu (approved):
Looks good to me!

PR #1576, opened 2023-10-23
Ensure support for functions which pass and return classes

### LINKED ISSUE #1444: Test functions which pass and return classes
### Describe the feature
After #1439 , #1440 , #1441 , #1442 , and #1443 classes are implemented. In order to use them effectively, they should also work in functions. Multiple tests should be added with functions that have classes as arguments, and functions that return classes

### LINKED ISSUE #1577: Missing type check for class constructor arguments
### Describe the bug
The constructor of a class does not check if the arguments provided to it have the expected type

### To Reproduce
```python
class A:
    def __init__(self : 'A', x : int):
        self.x = x

if __name__ == '__main__':
    my_a = A(3.5)
```

### Error details
```fortran
/home/emily/Code/test/tmp/__pyccel__/prog_tmp3.f90:11:29:

   11 |   call my_a % create(3.5_f64)
      |                             1
Error: Type mismatch in argument ‘x’ at (1); passed REAL(8) to INTEGER(8)
```

### Expected behavior
A neat Pyccel error should be raised similarly to what would happen if this was a function.

### Language
Semantic stage issue, concerns all languages.

### Additional context
The constructor should use the function `_check_argument_compatibility` in the `SemanticParser` to check that the arguments are valid.

sboof911:
Fixes #1444. Thanks to the new unit tests, the following fixes were made:

- Fix destructor call for C and Fortran.
- Remove header definition for inline class methods.
- Add check compatibility of class attributes.
- Add error call for multiple type definition for each of the arguments of the `__init__` method (fixes #1577).


pyccel/parser/semantic.py
```diff
@@ -2638,8 +2638,9 @@ def _visit_FunctionCall(self, expr):
             cls_variable = self._assign_lhs_variable(expr.get_user_nodes(Assign)[0].lhs, d_var, expr, new_expression, False)
             self._additional_exprs[-1].extend(new_expression)
             args = (FunctionCallArgument(cls_variable), *args)
-            # TODO check compatibility
             # TODO treat parametrized arguments.
```
EmilyBourne, line 2641: > TODO treat parametrized arguments.

Do you know if this works?

```python
class A:
    def __init__(self, x : int, y : bool):
        self.x = x
        self.y = y

if __name__ == '__main__':
    a = (3, True)
    my_class_var = A(*a)
```

sboof911: It works and I added a test for it in https://github.com/pyccel/pyccel/pull/1576/commits/5d93f602a8f9a77b8420dff5ea3ed42f49bdfac6
I already deleted the TODO comment in https://github.com/pyccel/pyccel/pull/1576/commits/c22d618b3bef773f237fe7c51c99ecc10426f3df

EmilyBourne: Ah cool :)

CHANGELOG.md
```diff
@@ -5,8 +5,12 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
+-   ##1576 : Add test functions which pass and return classes.
```
EmilyBourne, line 8: You don't need to document the addition of tests in the CHANGELOG. Remember, the changelog is there for users to work out which version of the code they should be using, and for devs to find developments

CHANGELOG.md
```diff
@@ -5,8 +5,12 @@ All notable changes to this project will be documented in this file.
...
+
 ### Fixed
 
+-   ##1576 : Fix Destructor Call for C and Fortran.
```
EmilyBourne, line 12: What is the problem that you are fixing?

pyccel/codegen/printing/ccode.py
```diff
@@ -761,12 +761,14 @@ def _print_ModuleHeader(self, expr):
...
+                if not method.is_inline:
+                    method.rename(classDef.name + ("__" + method.name if not method.name.startswith("__") else method.name))
+                    funcs += f"{self.function_signature(method)};\n"
             for interface in classDef.interfaces:
                 for func in interface.functions:
-                    func.rename(classDef.name + ("__" + func.name if not func.name.startswith("__") else func.name))
-                    funcs += f"{self.function_signature(func)};\n"
+                    if not func.is_inline:
+                        func.rename(classDef.name + ("__" + func.name if not func.name.startswith("__") else func.name))
+                        funcs += f"{self.function_signature(func)};\n"
```
EmilyBourne, line 771: Does this fix inline methods? Should this be mentioned in the CHANGELOG?

pyccel/codegen/printing/fcode.py
```diff
@@ -307,15 +307,40 @@ def print_kind(self, expr):
...
         scope = self.scope
         func = expr.funcdef
 
         # Print any arguments using the same inline function
         # As the function definition is modified directly this function
         # cannot be called recursively with the same FunctionDef
         args = []
+        if isinstance(expr, DottedFunctionCall):
+            if self.scope.find(expr.funcdef.cls_name, 'classes'):
+                args.append(expr.args[0])
```
EmilyBourne, line 343: Do you know why `self` is not in the args? Can this be fixed in the call to `_handle_inline_func_call`?

pyccel/parser/semantic.py
```diff
@@ -3814,7 +3823,7 @@ def _visit_ClassDef(self, expr):
         for method in cls.methods:
             if method.name == '__del__':
                 self._current_function = method.name
-                attribute = [attr for attr in cls.attributes if not attr.on_stack]
+                attribute = [attr for attr in cls.attributes if not attr.on_stack or isinstance(attr.dtype, CustomDataType)]
```
EmilyBourne, line 3826: Can you add parentheses to clarify this statement. Is it:
```suggestion
                attribute = [attr for attr in cls.attributes if not (attr.on_stack or isinstance(attr.dtype, CustomDataType))]
```
in which case why do we not deallocate classes?
Or is it:
```suggestion
                attribute = [attr for attr in cls.attributes if (not attr.on_stack) or isinstance(attr.dtype, CustomDataType)]
```
Is this to deallocate classes on the stack?
If so then please add a comment to clarify

tests/codegen/fcode/scripts/imports.py
```diff
@@ -4,11 +4,12 @@
...
+from numpy import ones
 xi = 2
 
 from decorators_types import decr
 yi = decr(xi)
 
 from classes_2 import Point
 
-p = Point(0.0)
+p = Point(ones(4))
```
EmilyBourne, line 15: This is related to #1586 . I think it would be better to avoid this complication. Can `classes_2.py` be modified to fix this?

tests/errors/semantic/blocking/classes.py
```diff
@@ -0,0 +1,5 @@
+# pylint: disable=missing-function-docstring, missing-module-docstring, missing-class-docstring
+
+class MyClass:
+    def __init__(self : 'MyClass', param1 : 'int|float'):
+        print(param1)
```
EmilyBourne, line 5: Please can you use names which explain what is being tested. E.g
```
git mv classes.py INTERFACE_CLASS__INIT__.py
```

tests/errors/semantic/non_blocking/classes_interface.py
```diff
@@ -0,0 +1,9 @@
+# pylint: disable=missing-function-docstring, missing-module-docstring, missing-class-docstring
+
+class MyClass:
+    def __init__(self : 'MyClass', param1 : 'int'):
+        self.param1 = param1
+        print("MyClass Object created!")
+
+    def help(self : 'MyClass', param1 : 'int|float'):
+        self.param1 = param1
```
EmilyBourne, line 9: See above

tests/pyccel/scripts/classes/classes_6.py
```diff
@@ -12,8 +14,13 @@ def translate(self : 'Point', a : float, b : float):
         self.x = self.x + a
         self.y = self.y + b
 
+    @inline
+    def get_attributes(self : 'Point', a : 'int | float'):
+        print(self.x, self.y, a)
+
```
EmilyBourne, line 20: Great! :)

EmilyBourne (changes requested):

tests/codegen/fcode/scripts/imports.py
EmilyBourne: Never mind, the scope of the Dummy variable that would be created is the same as the scope of `p` so you should be ok here.

CHANGELOG.md
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1576/commits/b6f1c7054ac8904ecf769c09923be7befd883a19.

CHANGELOG.md
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1576/commits/b6f1c7054ac8904ecf769c09923be7befd883a19.

pyccel/codegen/printing/fcode.py
sboof911: the `self` is removed here : https://github.com/pyccel/pyccel/blob/a332075ee1f8b7d3bd0f4390a01bb7af7343ac69/pyccel/codegen/printing/fcode.py#L2907-L2908

CHANGELOG.md
EmilyBourne: I still don't know what this means :sweat_smile: Can you explain in more words here

pyccel/codegen/printing/fcode.py
EmilyBourne: Maybe we should use `expr.args` in the arguments to `_handle_inline_func_call` instead of removing `self` then using costly functions to rediscover it

tests/errors/semantic/blocking/classes.py
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1576/commits/70412aeb3da4844bce30dbdec815e675a5794da2.

tests/errors/semantic/non_blocking/classes_interface.py
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1576/commits/70412aeb3da4844bce30dbdec815e675a5794da2.

tests/pyccel/scripts/classes/classes_6.py
sboof911: ☺️​

CHANGELOG.md
EmilyBourne: Here as in on the thread, not in the changelog

sboof911: Before the automatique call of the Distructor was not perfect, sometimes it was called even if the class is still used.
Now I fixed that problem and it's called at the right place.

EmilyBourne: Do you know what conditions caused this?

pyccel/parser/semantic.py
sboof911: It's the first case that is correct cause if an attribute has another class as value, it shoudn't be added to the garbage collector of the current scope.

pyccel/codegen/printing/fcode.py
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1576/commits/c163e184f5d393e8fb5289b35159e4a8c413dced.

pyccel/parser/semantic.py
EmilyBourne: So then when is this class attribute deallocated if not in the destructor?

CHANGELOG.md
sboof911: it's in here : https://github.com/pyccel/pyccel/blob/c163e184f5d393e8fb5289b35159e4a8c413dced/pyccel/parser/semantic.py#L3826

EmilyBourne: I mean what user code causes the destructor to be called in the wrong place

EmilyBourne (changes requested):
Just a couple of minor comments. Otherwise it looks good :+1:

pyccel/parser/semantic.py
```diff
@@ -511,6 +511,8 @@ def _garbage_collector(self, expr):
                 if isinstance(i, DottedVariable):
                     if isinstance(i.lhs.dtype, CustomDataType) and self._current_function != '__del__':
                         continue
+                if isinstance(i.dtype, CustomDataType) and i.is_alias:
+                    continue
```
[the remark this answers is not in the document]
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1576/commits/c4490f1d70bfe240eea923c0533bf7e01f21bec0.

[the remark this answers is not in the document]
sboof911: We should not delete the aliases of class or we gonna get a seg fault if we wanna use the original variabale.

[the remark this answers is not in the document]
EmilyBourne: I agree, that is why you don't add the aliases to `_allocs` here:
https://github.com/pyccel/pyccel/pull/1576/files#diff-9ba0c438140113a543a41e16000b939989e841d4fc6a6a4112b5c8e5d8807524R1279

Are there some aliases being added to `_allocs` elsewhere?

EmilyBourne (approved):
Looks good to me

tests/pyccel/test_pyccel.py
```diff
@@ -796,8 +796,7 @@ def test_basic_header():
 
 #------------------------------------------------------------------------------
 @pytest.mark.xdist_incompatible
-@pytest.mark.parametrize( "test_file", ["scripts/classes/generic_methods.py",
-                                        "scripts/classes/classes.py",
+@pytest.mark.parametrize( "test_file", ["scripts/classes/classes.py",
```
yguclu, line 799: Can you please explain this change?

tests/errors/semantic/blocking/INTERFACE_CLASS__INIT__.py
```diff
@@ -0,0 +1,5 @@
+# pylint: disable=missing-function-docstring, missing-module-docstring, missing-class-docstring
+
+class MyClass:
+    def __init__(self : 'MyClass', param1 : 'int|float'):
+        print(param1)
```
yguclu, line 5: Why should this raise an error? The input parameter is only printed out and not stored in the object, so potentially we could have multiple constructor functions which are not in contradiction with each other.

pyccel/parser/semantic.py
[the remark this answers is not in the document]
yguclu: Is this conversation resolved?

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ All notable changes to this project will be documented in this file.
 ### Fixed
 
 -   #1587 : Fix unnecessarily long file names generated by `epyccel`.
+-   #1576 : Corrected Destructor Invocation for Proper Cleanup.
```
yguclu, line 14:
```suggestion
-   #1576 : Correct destructor invocation for proper cleanup.
```

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ All notable changes to this project will be documented in this file.
...
+-   #1576 : Remove inline class method definition.
```
yguclu, line 15: How about issue #1444?

yguclu (changes requested):
Good job!

I have a few comments and questions.

tests/pyccel/test_pyccel.py
EmilyBourne: This change is my fault. There is some kind of undefined behaviour in `generic_methods.py`. See #1595.
The test often passes but not always. I moved the test to `test_interfaces_in_classes` in my PR. But I forgot to remove it from this list

CHANGELOG.md
EmilyBourne: I don't think issue #1444 explains anything about these problems. In my opinion it is more useful to link to this PR than that issue

tests/errors/semantic/blocking/INTERFACE_CLASS__INIT__.py
EmilyBourne: This is true, I hadn't thought about the case where an argument is not used when @sboof911 asked me about this. I don't know how easy/difficult it is to manage Interface constructors. If it is easy then it should probably be fixed in this PR, but if not then we should probably open a dedicated issue

CHANGELOG.md
yguclu: The title of this PR is "Test functions which pass and return classes" but we do not mention that...

yguclu (changes requested):

tests/errors/semantic/blocking/INTERFACE_CLASS__INIT__.py
yguclu: @sboof911 What do you think?

CHANGELOG.md
EmilyBourne: I usually ask people to not mention tests in the CHANGELOG. The CHANGELOG is designed for users. If we include tests in this then we will have a line in every PR about tests added

EmilyBourne: We could rename the PR:
> Ensure support for functions which pass and return classes

tests/errors/semantic/blocking/INTERFACE_CLASS__INIT__.py
yguclu: I suggest we open a dedicated issue and merge this PR now.

yguclu (approved):

pyccel/parser/semantic.py
```diff
@@ -3859,6 +3881,10 @@ def _visit_ClassDef(self, expr):
             m_name = method.name
             if m_name == '__init__':
                 self._visit(method)
+                if cls.interfaces:
+                    errors.report("Pyccel does not support interface constructor", symbol=m_name,
+                        bounding_box=(self._current_fst_node.lineno, self._current_fst_node.col_offset),
+                        severity='fatal')
```
EmilyBourne, line 3887: `_current_fst_node` was renamed in a recent PR. But why not avoid this problem by providing the method directly:

```suggestion
                    errors.report("Pyccel does not support interface constructor",
                        symbol=method, severity='fatal')
```

EmilyBourne (changes requested):

EmilyBourne (approved):

Issue #1491, opened 2023-08-10
Add documentation for classes


EmilyBourne:
In the [`docs/`](https://github.com/pyccel/pyccel/tree/devel/docs) folder add a file called `classes.md`. Describe everything that is implemented with examples (similarly to what is done in the [quickstart](https://github.com/pyccel/pyccel/blob/devel/docs/quickstart.md)). Be clear that not everything is supported.

Put a message at the top which makes it clear that there is no wrapper support. So the classes cannot be used from Python. Once phase 3 of Epic #453 is completed this message can simply be removed

PR #1479, opened 2023-08-07
Flexible Instance Naming in Class Methods.

sboof911:

PR #1580, opened 2023-10-24
Add documentation for classes.

sboof911:
Fix https://github.com/pyccel/pyccel/issues/1491.
Add documentation about approved and disallowed practices within class structures in Pyccel.

### DIFF
--- a/.dict_custom.txt
+++ b/.dict_custom.txt
@@ -13,6 +13,7 @@ pyccelised
 pyccelising
 allocatable
 deallocate
+deallocated
 deallocation
 deallocations
 conda
@@ -99,3 +100,7 @@ unhandled
 unallocated
 mypy
 destructor
+destructor's
+VALGRIND
+initialize
+initialized
--- a/.dict_custom.txt
+++ b/.dict_custom.txt
@@ -101,3 +101,4 @@ unallocated
 mypy
 destructor
 destructor's
+Valgrind
--- a/.dict_custom.txt
+++ b/.dict_custom.txt
@@ -101,6 +101,3 @@ unallocated
 mypy
 destructor
 destructor's
-VALGRIND
-initialize
-initialized
--- a/docs/classes.md
+++ b/docs/classes.md
@@ -261,4 +261,4 @@ MyClass Object created!
 
 ## Limitations
 
-It's important to note that Pyccel does not support class wrappers, inheritance, properties, magic methods or class variables. For our first implementation, the focus of Pyccel is primarily on core class functionality and memory management.
+It's important to note that Pyccel does not support class inheritance, properties, magic methods or class variables. For our first implementation, the focus of Pyccel is primarily on core class functionality and memory management.
--- a/docs/classes.md
+++ b/docs/classes.md
@@ -14,8 +14,9 @@ Pyccel strives to provide robust support for object-oriented programming concept
 
 -   The Destructor Method, `__del__`, is used to perform cleanup actions when an object is destroyed.
 -   Pyccel automatically takes care of garbage collection for classes.
-    - Attributes are released during the destructor's execution.
-    - The class destructor is called automatically once the class goes out of scope.
+
+    -  Attributes are released during the destructor's execution.
+    -  The class destructor is called automatically once the class goes out of scope.
 
 ### - Python Example
 
--- a/docs/classes.md
+++ b/docs/classes.md
@@ -13,10 +13,10 @@ Pyccel strives to provide robust support for object-oriented programming concept
 ## Destructor Method
 
 -   The Destructor Method, `__del__`, is used to perform cleanup actions when an object is destroyed.
--   Pyccel automatically takes care of garbage collection for classes.
 
-    -  Attributes are released during the destructor's execution.
-    -  The class destructor is called automatically once the class goes out of scope.
+-   Pyccel automatically takes care of garbage collection for classes.
+    -   Attributes are released during the destructor's execution.
+    -   The class destructor is called automatically once the class goes out of scope.
 
 ### - Python Example
 
@@ -261,4 +261,4 @@ MyClass Object created!
 
 ## Limitations
 
-It's important to note that Pyccel does not support class wrappers, inheritance, properties, magic methods, class variables, or inline class methods. For our first implementation, the focus of Pyccel is primarily on core class functionality and memory management.
+It's important to note that Pyccel does not support class wrappers, inheritance, properties, magic methods or class variables. For our first implementation, the focus of Pyccel is primarily on core class functionality and memory management.
--- a/docs/classes.md
+++ b/docs/classes.md
@@ -12,8 +12,9 @@ Pyccel strives to provide robust support for object-oriented programming concept
 ## Destructor Method
 
 -   The Destructor Method, `__del__`, is used to perform cleanup actions when an object is destroyed.
--   Pyccel automatically adds attributes that need to be freed to the garbage collector. These attributes are subsequently released during the destructor's execution. This mechanism helps manage memory efficiently, ensuring that resources are appropriately deallocated.
--   One notable feature of Pyccel is its automatic management of destructor calls, even when constructors are invoked in different scopes. This ensures that resources are properly cleaned up when class instances are no longer in use.
+-   Pyccel automatically takes care of garbage collection for classes.
+    - Attributes are released during the destructor's execution.
+    - The class destructor is called automatically once the class goes out of scope.
 -   The first parameter of the `Destructor Method` should always be named `self`.
 
 ### - Python Example
--- a/docs/classes.md
+++ b/docs/classes.md
@@ -263,4 +263,4 @@ MyClass Object created!
 
 ## Limitations
 
-It's important to note that Pyccel does not support class wrappers, inheritance, properties, magic methods, class variables, or inline class methods. The focus of Pyccel is primarily on core class functionality and memory management.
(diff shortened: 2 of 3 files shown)


.dict_custom.txt
```diff
@@ -99,3 +100,7 @@ unhandled
 unallocated
 mypy
 destructor
+destructor's
+VALGRIND
```
EmilyBourne, line 104: According to their website this should simply be `Valgrind`

.dict_custom.txt
```diff
@@ -99,3 +100,7 @@ unhandled
...
+initialize
+initialized
```
EmilyBourne, line 106: These words probably exist in the dictionary already but we have a UK dictionary not a US dictionary so they will be spelled:
- initialise
- initialised

docs/classes.md
```diff
@@ -0,0 +1,263 @@
+# Supported Class by Pyccel
+
+Pyccel strives to provide robust support for object-oriented programming concepts commonly used by developers. In Pyccel, classes are a fundamental building block for creating structured and reusable code. This documentation outlines key features and considerations when working with classes in Pyccel.
+
+## Constructor Method
+
+-   The Constructor Method, `__init__`, is used to initialize the object's attributes.
+-   It must be named `__init__`, or the Pyccel compiler will generate an error.
```
EmilyBourne, line 8: What error are you talking about here? Normally it is not possible to initialise an object's attributes with a function with a different name. Is your point that you have to have an `__init__` method or that you cannot assign to attributes in other methods if they have not first been created in the `__init__` method?

docs/classes.md
```diff
@@ -0,0 +1,263 @@
...
+-   The first parameter of the `Constructor Method` should always be named `self`.
+-   Pyccel doesn't support Interface Constructor.
```
EmilyBourne, line 10: The term `Interface` comes from Fortran. I'm not sure it is widely understood by our users. I suggest:
```suggestion
-   Pyccel only permits one type definition for each of the arguments of the `__init__` method. Union types or templates cannot be used.
```

docs/classes.md
```diff
@@ -0,0 +1,263 @@
...
+
+## Destructor Method
+
+-   The Destructor Method, `__del__`, is used to perform cleanup actions when an object is destroyed.
+-   Pyccel automatically adds attributes that need to be freed to the garbage collector. These attributes are subsequently released during the destructor's execution. This mechanism helps manage memory efficiently, ensuring that resources are appropriately deallocated.
+-   One notable feature of Pyccel is its automatic management of destructor calls, even when constructors are invoked in different scopes. This ensures that resources are properly cleaned up when class instances are no longer in use.
```
EmilyBourne, line 16: I'm not sure this is notable. It is critical to avoid seg faults though.
```suggestion
-   Pyccel automatically takes care of garbage collection for classes.
    - Attributes are released during the destructor's execution.
    - The class destructor is called automatically once the class goes out of scope.
```

docs/classes.md
```diff
@@ -0,0 +1,263 @@
...
+-   The first parameter of the `Destructor Method` should always be named `self`.
```
EmilyBourne, line 17: Isn't this true for any class method not just for `__init__` and `__del__`?

EmilyBourne: I've just seen this is mentioned below. I don't know if it's worth repeating 3 times. Especially as you are already working on removing this restriction (#1479) so you will need to delete this in 3 places. If you don't mind deleting it 3 times though then why not.

docs/classes.md
```diff
@@ -0,0 +1,263 @@
...
+==151606== All heap blocks were freed -- no leaks are possible
+==151606==
+==151606== For lists of detected and suppressed errors, rerun with: -s
+==151606== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0
+```
+
+## Class Methods
+
+-   Pyccel now supports Class Methods and Interfaces.
+-   In Pyccel, class attributes can be initialized within any method of the class.
```
EmilyBourne, line 124: Is this true? Doesn't it contradict what you say about the constructor?

docs/classes.md
```diff
@@ -0,0 +1,263 @@
...
+
+```Shell
+MyClass1 Object created!
+MyClass Object created!
+1
+```
+
+## Limitations
+
+It's important to note that Pyccel does not support class wrappers, inheritance, properties, class variables, or inline class methods. The focus of Pyccel is primarily on core class functionality and memory management.
```
EmilyBourne, line 263: Have you tested inline class methods? I plan to use this soon if it doesn't work then I'll need to fix it first :sweat_smile:

EmilyBourne, line 263: I think we should mention the wrappers at the top, especially while we're developing the wrapper. We can then delete the line once we're done.

EmilyBourne, line 263:
```suggestion
It's important to note that Pyccel does not support class wrappers, inheritance, properties, magic methods, class variables, or inline class methods. The focus of Pyccel is primarily on core class functionality and memory management.
```

docs/classes.md
```diff
@@ -0,0 +1,263 @@
+# Supported Class by Pyccel
+
+Pyccel strives to provide robust support for object-oriented programming concepts commonly used by developers. In Pyccel, classes are a fundamental building block for creating structured and reusable code. This documentation outlines key features and considerations when working with classes in Pyccel.
```
EmilyBourne, line 3:
```suggestion
Pyccel strives to provide robust support for object-oriented programming concepts commonly used by developers. In Pyccel, classes are a fundamental building block for creating structured and reusable code. This documentation outlines key features and considerations when working with classes in Pyccel.

:warning: Pyccel's class support is currently limited to translations only. The implementation of the wrapper which will make the class accessible from Python is in progress.
```

docs/classes.md
```diff
@@ -0,0 +1,263 @@
...
+==151606== For lists of detected and suppressed errors, rerun with: -s
+==151606== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0
+```
+
+## Class Methods
+
+-   Pyccel now supports Class Methods and Interfaces.
+-   In Pyccel, class attributes can be initialized within any method of the class.
+-   Pyccel enables classes to be passed as arguments to methods and functions and returned with modified data.
+-   The first parameter of the Class Method should always be named `self`.
```
EmilyBourne, line 126: I think it may be worth moving this further up the file and grouping the examples. As they are quite detailed it could be easy to miss this section in the middle

EmilyBourne (changes requested):
These docs look great, thanks for taking care of this. I just have a few comments but I don't see anything critical missing.

Generally in the documentation I assume that I am talking to a Python developer so I try to use Python terminology where possible. This may be worth bearing in mind for the sections on the constructors and destructors

docs/classes.md
EmilyBourne, line 263:
```suggestion
It's important to note that Pyccel does not support class wrappers, inheritance, properties, class variables, or inline class methods. For our first implementation, the focus of Pyccel is primarily on core class functionality and memory management.
```

.dict_custom.txt
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1580/commits/2760eebc29beb59adec461e85472f699b981ffff.

.dict_custom.txt
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1580/commits/a603c43dc6faf2844942ca4541fab1d52b177432.

docs/classes.md
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1580/commits/2467c0d7adf8ac3547ccf5ce4edc7ad2c6891083.

docs/classes.md
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1580/commits/f92c3e1883b231711e3cbad1f475439610d22f96.

docs/classes.md
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1580/commits/f92c3e1883b231711e3cbad1f475439610d22f96.

docs/classes.md
sboof911: Yeah normally `__init__` where we have to initilise them but we can create other attributes of a class in any other method of the same class.

docs/classes.md
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1580/commits/3e0def115492fb89aae4d30826365da820416f87.
I need also to add tests in https://github.com/pyccel/pyccel/pull/1576.

sboof911: Fixed https://github.com/pyccel/pyccel/pull/1580/commits/4f63dd4707a49ee3652c0eb212dad688eed1e600.

docs/classes.md
sboof911: Fixed https://github.com/pyccel/pyccel/pull/1580/commits/65baee9d82decc0a4f2026777e35685a2ded628a.
Fixed https://github.com/pyccel/pyccel/pull/1580/commits/2cbad56fd8f824b2e827ac6512a991e661baa2a9.

EmilyBourne:
Ok for me if the example is changed to avoid the pointer problem

EmilyBourne (changes requested):
This looks great. Thanks

EmilyBourne (approved):
This looks great. Thanks!

yguclu (approved):

PR #1596, opened 2023-10-31
Add preliminary wrapper for Python class

### LINKED ISSUE #1493: Add import of a class type in the C-Python wrapper

EmilyBourne:
Add preliminary support for importing classes. Classes are added to modules but no class methods or variables are added in this PR. This allows us to import the class but not to interact with it. Fixes #1493


EmilyBourne:
/bot run spelling docs pylint pyccel_lint coverage

sboof911 (approved):
Looks Good.

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -240,6 +240,8 @@ def _print_PyArgKeywords(self, expr):
     def _print_PyModule_AddObject(self, expr):
         name = self._print(expr.name)
         var  = self._print(expr.variable)
+        if not expr.variable.dtype is PyccelPyObject():
```
yguclu, line 243: More Pythonic is writing
```suggestion
        if expr.variable.dtype is not PyccelPyObject():
```

tests/epyccel/test_epyccel_classes.py
```diff
@@ -0,0 +1,11 @@
...
+from pyccel.epyccel import epyccel
+
+def test_class_import(language):
+    class A:
+        def __init__(self : 'A'):
+            pass
+
+    epyc_A = epyccel(A)
+
+    assert isinstance(epyc_A, type)
```
yguclu, line 11: Shouldn't `epyc_A` be of type `A` just like in Python? If not, code that checks `isinstance(obj, A)` will fail to work after pyccelization.

yguclu (changes requested):
Looks good to me

pyccel/codegen/printing/cwrappercode.py
EmilyBourne: Huh, pylint used to raise an error about this

tests/epyccel/test_epyccel_classes.py
EmilyBourne: No. We are returning the Pyccelised class A not an instance of that class so the result must be a type. If we want the class methods to call Pyccelised functions I don't think the resulting class can be equivalent to A

yguclu: Ah sorry, this is not an instance of A but A itself!

yguclu (approved):

Issue #1410, opened 2023-05-30
Conditional statement not working inside of a list comprehension


smhah:
### Describe the bug
The conditional statement inside of a list comprehension is ignored at the pyccelization

### To Reproduce
Provide code to reproduce the behavior:
```python

if __name__ == "__main__":
    fy = [3, 1, 4]
    ar = [y for y in fy if y == 99]
    print(ar)

```

### Error details, (Wrong Output)
Provide the generated code, or the error message:
```bash/fortran/c
[3, 1, 4]
```

### Expected behavior
```
[]
```
Empty list because no number is matching 99
### Language
Translated to Fortran/C

EmilyBourne:
This issue cannot be fixed currently as we do not have lists of unknown size, however as a minimum an error should be raised here so as not to mislead users

PR #1598, opened 2023-11-01
Raise an error for ifs in comprehension statements

EmilyBourne:
Raise an error when if conditions are used in comprehension statements.
E.g:
```python
a = [i for i in range(10) if i%2==0]
```
See #1410


pyccel/parser/syntactic.py
```diff
@@ -1119,6 +1119,10 @@ def _visit_comprehension(self, stmt):
 
         self.exit_loop_scope()
 
+        if stmt.ifs:
+            errors.report("Can't calculate list length when comprehension contains if statements.\n" + PYCCEL_RESTRICTION_TODO,
```
bauom, line 1123: I found this error a bit confusing, the error assumes that i know the issue.
(i did understand it after I read the issue)

EmilyBourne: Hmm, I was hoping that the line number would help make it obvious. Do you have another suggestion?

EmilyBourne:
```suggestion
            errors.report("Cannot handle if statements in list comprehensions. List length cannot be calculated.\n" + PYCCEL_RESTRICTION_TODO,
```
?

bauom: I think This describe the Problem clearly. `Cannot handle if statements in list comprehensions.`
which for me was hard to know from the start with the previous message.

EmilyBourne: Fixed

bauom (approved):

PR #1604, opened 2023-11-04
 Add wrapping of a FunctionDefResult containing a class

### LINKED ISSUE #1494: Add wrapping of a FunctionDefResult containing a class
### Describe the feature
I want to be able to call functions which return class instances.

### Test Code

For the following module handled by Pyccel:
```python
class Point:
    def __init__(self, x : int):
         self._x = x

def get_point(x : int):
    return Point(x)
```
we should be able to do:
```python
from mod import get_point

p = get_point(3)
```

### Proposed Solution
The function `CToPythonWrapper._visit_FunctionDefResult` should be improved to handle classes. The resulting code will be:
```c
struct PyPointObject* get_point_wrapper(PyObject* self, PyObject* args, PyObject* kwargs)
{
    PyObject* x_obj;
    int64_t x;
    struct PyPointObject* Out_0001_obj;
    static char *kwlist[] = {
        "x",
        NULL
    };
    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "O", kwlist, &x_obj))
    {
        return NULL;
    }
    Out_0001_obj = (struct PyPointObject*)PyPointType.tp_alloc(&PyPointType, 0);
    if (PyIs_NativeInt(x_obj))
    {
        x = PyInt64_to_Int64(x_obj);
    }
    else
    {
        PyErr_SetString(PyExc_TypeError, "Expected an argument of type int for argument x");
        return NULL;
    }
    Out_0001_obj->instance = malloc(sizeof(struct Point));
    (*(struct Point*)(Out_0001_obj->instance)) = get_point(x);
    return Out_0001_obj;
}
```

### Issues which must be tackled first
- #1493 : To have a class instance
- #1444 : Functions which return classes

EmilyBourne:
Add wrapping of a `FunctionDefResult` containing an instance of a user-defined class. Fixes #1494.

**Commit Summary**
- Add new `PointerCast` function to avoid compiler warnings in C
- Put back `Assign.like` parameter to allow us to `malloc` a void pointer in the wrapper
- Add docstrings
- Add `WrapperCustomDataType` to describe the DataType of the Python-compatible class object in the wrapper
- Ensure scopes are available as necessary 
- Allow the allocation of non-array objects (e.g. class objects in a `bindcpointer`)
- Handle Python-compatible classes in the wrapper (i.e. objects which inherit from `PyObject`)
- Allow renaming classes in the scope. This allows us to find the Python-compatible class object (which has an auto-generated name) using the name of the original class (so it can be found even in case of name collisions)
- Add tests


EmilyBourne:
Blocked by #1576

pyccel/codegen/wrapper/c_to_python_wrapper.py
```diff
@@ -71,6 +71,11 @@ def get_new_PyObject(self, name, is_temp = False):
         name : str
             The desired name.
 
+        dtype : DataType, optional
+            The datatype of the object which will be represented by this PyObject.
+            This is not necessary unless a variable sis required which will describe
```
EmilyBourne, line 76:
```suggestion
            This is not necessary unless a variable is required which will describe
```

pyccel/codegen/printing/ccode.py
```diff
@@ -1486,22 +1486,32 @@ def _print_PyccelArrayShapeElement(self, expr):
...
+            return '{}{}'.format(free_code, alloc_code)
+        elif variable.is_alias:
+            var_code = self._print(ObjectAddress(variable))
+            if expr.like:
+                declaration_type = self.get_declare_type(expr.like)
+                return f'{var_code} = malloc(sizeof({declaration_type}));\n'
+            else:
+                return f'{var_code} = malloc(3);\n'
+        else:
+            raise NotImplementedError(f"Allocate not implemented for {variable}")
```
EmilyBourne, line 1514:
```suggestion
            variable.inspect()
            raise NotImplementedError(f"Allocate not implemented for {variable}")
```

pyccel/codegen/printing/ccode.py
```diff
@@ -1486,22 +1486,33 @@ def _print_PyccelArrayShapeElement(self, expr):
...
+            return '{}{}'.format(free_code, alloc_code)
+        elif variable.is_alias:
+            var_code = self._print(ObjectAddress(variable))
+            if expr.like:
+                declaration_type = self.get_declare_type(expr.like)
+                return f'{var_code} = malloc(sizeof({declaration_type}));\n'
+            else:
+                return f'{var_code} = malloc(3);\n'
+        else:
+            variable.inspect()
```
EmilyBourne, line 1514:
```suggestion
            print(variable.dtype, variable.rank, variable.is_ndarray, type(variable.dtype))
            variable.inspect()
```

pyccel/ast/c_concepts.py
```diff
@@ -57,6 +58,68 @@ def obj(self):
         """
         return self._obj
 
+
+class PointerCast(TypedAstNode):
+    """
+    A class which represents the casting of one pointer to another.
+
+    A class which represents the casting of one pointer to another in C code.
+    This is useful for storing addresses in a void pointer or for storing.
```
yguclu, line 67: Was this sentence truncated by accident?

pyccel/ast/c_concepts.py
```diff
@@ -57,6 +58,68 @@ def obj(self):
...
+    This class is not necessary but avoids warnings being raised in the
+    translated code.
```
yguclu, line 69: I suggest to expand the explanation:
```suggestion
    Using this class is not strictly necessary to produce correct C code,
    but avoids compiler warnings about the implicit conversion of pointers.
```

pyccel/codegen/wrapper/c_to_python_wrapper.py
yguclu: Was this typo fixed?

pyccel/codegen/wrapper/c_to_python_wrapper.py
```diff
@@ -80,10 +85,20 @@ def get_new_PyObject(self, name, is_temp = False):
...
-                        is_temp=is_temp)
+        if isinstance(dtype, CustomDataType):
+            try:
+                var = Variable(dtype=self._python_object_map[dtype],
+                                name=self.scope.get_new_name(name),
+                                memory_handling='alias',
+                                cls_base = self.scope.find(dtype.name, 'classes'),
+                                is_temp=is_temp)
+            except KeyError as e:
+                raise NotImplementedError("Can't return an object whose type was imported") from e
```
yguclu, line 96: What exactly does this message mean?

pyccel/ast/c_concepts.py
EmilyBourne: I think I just wrote the truncated version first for the "short summary" then wrote the longer version. The "in C code" isn't strictly necessary as it is implicit from the location of the class, however it is still useful information so I put it in the extended description only

pyccel/codegen/wrapper/c_to_python_wrapper.py
EmilyBourne: It means that currently we cannot do:
File1.py
```python
class A:
    def __init__(self, x : int):
        self.x = x
```
File2.py
```python
from File1 import A

def f():
    return A(3)
```

The wrapper needs splitting into a .h and a .c file so that the class description is accessible from other modules, however this seemed beyond the scope of this PR so I left it as is for now

pyccel/ast/c_concepts.py
yguclu: I was referring to the "or for storing" part. Storing what?

EmilyBourne: Ah, oops. I don't remember what I was thinking of here. I have removed the snippet. 23885db6241bf170363effd1aa9e8dc7404bfe43

pyccel/codegen/wrapper/c_to_python_wrapper.py
yguclu: Do we have an issue about this? If so, it could be referenced in the error message.

EmilyBourne: I have created an issue and referenced it: 0f82d8d4b

PR #1606, opened 2023-11-06
Add wrapping of a FunctionDefArgument containing a class 

### LINKED ISSUE #1495: Add wrapping of a FunctionDefArgument containing a class
### Describe the feature
I want to be able to call functions and pass class arguments.

### Test Code
```python
class Point:
    def __init__(self, x : int):
         self._x = x

def get_point(x : int):
    return Point(x)

def print_point(p : Point):
    print(p.x)
```

### Proposed Solution
The function `CToPythonWrapper._visit_FunctionDefArgument` should be improved to handle classes. The resulting code should look something like:
```c
typedef struct {
    PyObject_HEAD
    Point* instance;
} PyPointObject;

static PyTypeObject PyPointType = {
    PyVarObject_HEAD_INIT(NULL, 0)
    .tp_name = "mod.Point",
    .tp_doc = PyDoc_STR("<class_docstring>"),
    .tp_basicsize = sizeof(PyPointObject),
    .tp_itemsize = 0,
    .tp_flags = Py_TPFLAGS_DEFAULT,
    .tp_new = PyType_GenericNew,
};

PyObject* print_point_wrapper(PyObject* self, PyObject* args, PyObject* kwargs)
{
    PyObject* p_obj;
    Point* p;
    static char *kwlist[] = {
        "p",
        NULL
    };
    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "O", kwlist, &p_obj))
    {
        return NULL;
    }
    if (PyObject_IsInstance(p_obj, (PyObject*)&PyPointType))
    {
        p = p_obj.instance;
    }
    else
    {
        PyErr_SetString(PyExc_TypeError, "Expected an argument of type Point for argument p");
        return NULL;
    }
    print_point(p);
    return Py_None;
}
```

### Issues which must be tackled first
- #1493 : To have a class instance
- #1494 : To get access to a class instance in a Module for testing
- #1444 : Functions which take classes as arguments

EmilyBourne:
Add the code necessary to wrap class instances as function arguments. Add a test and let Pyccel take care of translating the existing tests in the `test/pyccel/scripts/classes` folder as they no longer fail during the wrapping stage. Fixes #1495


EmilyBourne:
Blocked by #1651

tests/pyccel/test_pyccel.py
```diff
@@ -811,15 +813,18 @@ def test_classes( test_file , language):
...
-                                        "scripts/classes/classes_7.py",
-                                        "scripts/classes/classes_8.py",
+@pytest.mark.parametrize( "test_file", ["scripts/classes/classes_7.py",
                                         ] )
+@pytest.mark.parametrize( 'language', (
+        pytest.param("python", marks = pytest.mark.python),
+        pytest.param("c", marks = pytest.mark.c),
+        pytest.param("fortran", marks = [
+            pytest.mark.xfail(reason="Issue #1651"),
+            pytest.mark.fortran])
```
yguclu, line 823: Issue #1651 has just been fixed.

pyccel/codegen/printing/ccode.py
```diff
@@ -1856,7 +1856,8 @@ def _print_FunctionCall(self, expr):
          # Ensure the correct syntax is used for pointers
         args = []
         for a, f in zip(expr.args, func.arguments):
-            a = a.value if a else Nil()
+            a = a.value
+            a = a if a else Nil()
```
yguclu, line 1860: Can you not write this?
```suggestion
            a = a.value or Nil()
```
By the way, we could use a different variable instead of redefining `a`.

yguclu (changes requested):

tests/pyccel/test_pyccel.py
EmilyBourne: Fixed f26cf34fb

yguclu (approved):
Looks good!

Issue #1586, opened 2023-10-27
Missing target on pointers in classes


EmilyBourne:
### Describe the bug
It is important to identify any objects which are targets. This allows errors to be raised  (e.g when the target is redefined). This information is missing when a pointer is created by a class.

### To Reproduce
Provide code to reproduce the behaviour:
```python
import numpy as np

class MyClass:
    def __init__(self : 'MyClass', param1 : 'int', param2 : 'float[:]'):
        self.param1 = param1
        self.param2 = param2
        print("MyClass Object created!")


if __name__ == '__main__':
    p = np.ones(4)
    obj = MyClass(1, p)
    p = np.array([1., 2., 3., 4.])
    print(obj.param1, obj.param2)
```

### Error details
Provide the generated code:
```fortran
  allocate(p(0:3_i64))
  p = 1.0_f64
  call obj % create(1_i64, p)
  if (any(size(p) /= [5_i64])) then
    deallocate(p)                                     ! Causes pointer in MyClass to point at deallocated memory
    allocate(p(0:4_i64))
  end if
  p = [1.0_f64, 2.0_f64, 3.0_f64, 4.0_f64, 5.0_f64]
  write(stdout, '(I0, A)', advance="no") obj%param1 , ' '
  write(stdout, '(A)', advance="no") '['
  do i = 0_i64, size(obj%param2, kind=i64) - 1_i64 - 1_i64, 1_i64
    write(stdout, '(F0.12, A)', advance="no") obj%param2(i) , ' '
  end do
  write(stdout, '(F0.12, A)', advance="no") obj%param2(size(obj%param2, &
        kind=i64) - 1_i64) , ']'
  write(stdout, '()', advance="yes")
  if (allocated(p)) then
    deallocate(p)
  end if
  call obj % free()
```

### Expected behavior
In an example without classes:
```python
import numpy as np

if __name__ == '__main__':
    p = np.ones(4)
    q = p
    p = np.array([1., 2., 3., 4.])
    print(p)
    print(q)
```

we obtain:
```
ERROR at annotation (semantic) stage
pyccel:
 |error [semantic]: tmp7.py [6,4]| Attempt to reallocate an array which is being used by another variable (p)
```

The same error should be raised here. To manage this we either need to disallow pointers in classes or add errors to handle all possible problems. There are lots of potential pitfalls so for now I think it is safer to disallow pointers in classes. The message could reference this issue to explain potential pitfalls.

### Language
All, semantic stage

[...]

Issue #1554, opened 2023-10-16
Cannot specify precision in stop statement


EmilyBourne:
### Describe the bug
The Intel compiler does not allow type specifiers on the integers passed to the stop intrinsic. 

### To Reproduce
Provide code to reproduce the behaviour:
```python
import sys

if __name__ == "__main__":
    sys.exit()
```

### Error details
Provide the generated code, or the error message:
```bash/fortran/c
2023-10-13T18:06:37.9907314Z /home/runner/work/pyccel/pyccel/tests/pyccel/scripts/exits/__pyccel__/__pyccel__gw0/prog_empty_exit.f90(8): error #5082: Syntax error, found IDENTIFIER '_I32' when expecting one of: :: ) , : * <END-OF-STATEMENT> ; . (/ + - ] /) ' ** / // > PRIVATE ...
2023-10-13T18:06:37.9907517Z   stop 0_i32
2023-10-13T18:06:37.9907740Z ---------^
```
https://github.com/pyccel/pyccel/actions/runs/6511414122/job/17687068773

### Expected behaviour
The code should compile. We should not print type specifiers at this location in the code.

### Language
Fortran

harsha-mangena:
Hi @EmilyBourne, Can you assign me this?

EmilyBourne:
Hi @harsha-mangena , thanks for your help! I have assigned the issue to you. Let me know if you need any help with this. The changes should be in the [codegen stage](https://github.com/pyccel/pyccel/blob/devel/developer_docs/codegen_stage.md). You will need to modify `FCodePrinter._print_SysExit` to handle `LiteralInteger`s as a special case.

harsha-mangena:
@EmilyBourne 
I added a check, If it is LiteralInteger and ensured no type specifier is added when printing integers for the stop intrinsic.
Please find changes in this [PR ](https://github.com/pyccel/pyccel/pull/1624)

EmilyBourne:
@harsha-mangena It loooks good, but I think the changes should be restricted to the `LiteralInteger` case. Don't forget to check with the bot to make sure nothing is missing from your PR

PR #1607, opened 2023-11-06
Allow calling class methods from Python

EmilyBourne:
Adds all code necessary to wrap class functions. `__init__`, `__del__` and `__new__` are special cases of this with their own prototypes in the cwrapper. These are also handled and tests are added in `tests/epyccel` (inspired by the tests in `tests/pyccel/scripts/classes`).


EmilyBourne:
/bot run docs coverage spelling pylint pyccel_lint

EmilyBourne:
/bot run pylint pyccel_lint docs spelling coverage

EmilyBourne:
/bot run pylint pyccel_lint docs spelling coverage

pyccel/codegen/printing/fcode.py
```diff
@@ -449,17 +480,19 @@
             name = self._print(expr.name)
         name = name.replace('.', '_')
         if not name.startswith('mod_') and self.prefix_module:
-            name = '{prefix}_{name}'.format(prefix=self.prefix_module,
-                                            name=name)
+            name = f'{self.prefix_module}_{name}'
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
#1677

pyccel/codegen/printing/fcode.py
```diff
@@ -437,6 +438,36 @@ def _get_external_declarations(self):
...
+                if m_name in _default_methods:
+                    suggested_name = _default_methods[m_name]
+                    scope.rename_function(method, suggested_name)
+                method.cls_name = scope.get_new_name(f'{name}_{method.name}')
+        for i in expr.interfaces:
+            for f in i.functions:
+                if not f.is_inline:
+                    i_name = f.name
+                    if i_name in _default_methods:
+                        suggested_name = _default_methods[i_name]
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
Will become useful when magic methods are supported

EmilyBourne:
/bot run pylint pyccel_lint docs spelling coverage

tests/epyccel/test_epyccel_classes.py
```diff
@@ -40,3 +44,167 @@ def test_class_argument(modnew):
...
+    c = modnew.C()
+    assert c.get_3() == 3
+
+@pytest.mark.parametrize( 'language', (
+        pytest.param("fortran", marks = [
+            pytest.mark.skip(reason="Argument reference count not increased. See #1586"),
+            pytest.mark.fortran]
+        ),
+        pytest.param("c", marks = [
+            pytest.mark.skip(reason="Argument reference count not increased. See #1586"),
```
yguclu, line 58: Did you mean #1686?

EmilyBourne: #1586 is the issue. #1686 is the associated PR

EmilyBourne: Currently #1686 doesn't actually fix this issue. The problem arises when a class stores a pointer to an object passed as argument. #1586 describes this problem. #1686 allows us to detect this however it then additionally needs to be taken into account in the wrapper to increase the reference count on the object and decrease it when the class is freed. I plan to fix this either in this PR or #1686 whichever is merged last, but I can create a separate PR if you think this will be clearer

yguclu: I think it is fine to fix it in #1686

pyccel/codegen/printing/ccode.py
```diff
@@ -260,6 +260,7 @@ class CCodePrinter(CodePrinter):
                       (NativeInteger(),8)     : 'int64_t',
                       (NativeInteger(),2)     : 'int16_t',
                       (NativeInteger(),1)     : 'int8_t',
+                      (NativeInteger(),-2)    : 'int',
```
yguclu, line 263: Why `-2`?

pyccel/codegen/wrapper/c_to_python_wrapper.py
```diff
@@ -500,6 +518,221 @@ def _build_module_exec_function(self, expr):
 
         return FunctionDef(func_name, [FunctionDefArgument(module_var)], [FunctionDefResult(result_var)], body,
                 scope = func_scope, is_static=True)
+
+    def _get_class_allocator(self, class_dtype, func = None):
+        """
+        Create the allocator for the class.
+
+        Create a function which will allocate the memory for the class. This
+        is equivalent to the `__new__` function.
```
yguclu, line 527: Maybe you meant "class object" of "class instance" rather than just "class"?

yguclu (approved):
Very nice!

pyccel/codegen/printing/ccode.py
EmilyBourne: Why not ? I needed a number for a key. Usually 0 is used for things that don't require a precision (e.g strings/classes). Positive numbers are used for actual precisions (as they would be printed in Fortran). -1 is used for the default Python precision. -2 seemed to be the next logical thing to use for the C default precision. It would probably be better with a name but I'm not sure how easy that would be to fit alongside what we currently have

If you have a better suggestion I'm so ears :)

pyccel/codegen/wrapper/c_to_python_wrapper.py
```diff
@@ -500,6 +518,221 @@ def _build_module_exec_function(self, expr):
 
         return FunctionDef(func_name, [FunctionDefArgument(module_var)], [FunctionDefResult(result_var)], body,
                 scope = func_scope, is_static=True)
+
+    def _get_class_allocator(self, class_dtype, func = None):
+        """
+        Create the allocator for the class.
+
+        Create a function which will allocate the memory for the class. This
```
EmilyBourne, line 526:
```suggestion
        Create a function which will allocate the memory for the class instance. This
```

pyccel/codegen/wrapper/c_to_python_wrapper.py
EmilyBourne: Fixed f23ff323a1faff158c75773e4bffb73e282b3cdb

pyccel/codegen/printing/ccode.py
yguclu: Why not indeed? 😄 But where do you need the C default integers exactly?

EmilyBourne: Python seems to be quite picky about the signature of the `__init__` function (https://docs.python.org/3/extending/newtypes_tutorial.html#adding-data-and-methods-to-the-basic-example). It expects 
```c
int
Custom_init(CustomObject *self, PyObject *args, PyObject *kwds)
```

On most computers I think we could get away with using precision=4 because it's usually mapped to the same datatype under the hood but this can't be guaranteed

EmilyBourne: There was a similar problem for `Exit` in Fortran : https://github.com/pyccel/pyccel/issues/1554

There we fixed it by handling this print as a special case. But this solution could also be used there.

PR #1616, opened 2023-11-13
Add documentation for AST nodes

EmilyBourne:
Add documentation for AST nodes. This PR assumes that #1590 has been merged and includes the descriptions of the new objects in that PR.


developer_docs/ast_nodes.md
```diff
@@ -0,0 +1,73 @@
...
+-   `dtype`
+-   `precision`
+-   `rank`
+-   `shape`
+-   `order`
+-   `class_type`
+
+The contents of these types are explained in more detail below.
+
+When examining the class `TypedAstNode` you may notice that there are two methods for getting each of these properties. Each time, one is a standard method, while the other is a static class method. In general in the code you will always use the normal method. This will return the instance variable if it is available, otherwise it will return the static class variable. The static class method is used for type deductions when [parsing type annotations](./type_inference.md). In type annotations we do not generally have an instance of a class, however we can get access to the class itself. For instance let us consider the following type annotation:
```
yguclu, line 21: I find it the use of the word "variable" a bit confusing here, and I suggest to replace it with "attribute":
- `instance variable` -> `instance attribute`
- `static class variable` -> `static class attribute`

developer_docs/ast_nodes.md
```diff
@@ -0,0 +1,73 @@
...
+```python
+a : int
+```
+When we visit `int` in the [semantic stage](./semantic_stage.md) the `SemanticParser` will return the class `PythonInt`. This is usually used as a function (e.g to cast a variable), however here we use it to deduce the type. Following the [development conventions](./development_conventions.md#Class-variables-vs.-Instance-variables) any variables which will remain constant over all instances of a class should be stored in static class variables. This means that they can be accessed via these static methods. Returning to our example, a call to the function `int` always returns a scalar object with an integer type and default precision. This means that all the properties of a `TypedAstNode` can be defined without having an instance of this class. These properties cannot be defined statically for all nodes (e.g. it would not be possible for `PyccelAdd`), however generally they can be defined for the nodes which can be used in type annotations.
```
yguclu, line 25: Same comment as above:
```suggestion
When we visit `int` in the [semantic stage](./semantic_stage.md) the `SemanticParser` will return the class `PythonInt`. This is usually used as a function (e.g to cast a variable), however here we use it to deduce the type. Following the [development conventions](./development_conventions.md#Class-variables-vs.-Instance-variables) any attributes which will remain constant over all instances of a class should be stored in static class attributes. This means that they can be accessed via these static methods. Returning to our example, a call to the function `int` always returns a scalar object with an integer type and default precision. This means that all the properties of a `TypedAstNode` can be defined without having an instance of this class. These properties cannot be defined statically for all nodes (e.g. it would not be possible for `PyccelAdd`), however generally they can be defined for the nodes which can be used in type annotations.
```

developer_docs/ast_nodes.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Pyccel Internal Function
+
+The class `pyccel.ast.internals.PyccelInternalFunction` is a super class. This class should never be used directly but provides functionalities which are common to certain AST objects. These AST nodes are those which describe functions which are supported by Pyccel. For example it is used for functions from the `math` library, the `cmath` library, the `numpy` library, etc. `PyccelInternalFunction` inherits from `TypedAstNode`. The type information for the sub-class describes the type of the result of the function.
+
+Objects inheriting from `PyccelInternalFunction` are created from a `FunctionCall` object resulting from the syntactic stage. The `PyccelInternalFunction` objects are initialised by passing the contents of the `FunctionCallArgument`s used in the `FunctionCall` to the constructor. It is therefore important that the names of the arguments and keyword arguments of any classes inherited from `PyccelInternalFunction` match the names of the arguments and keyword arguments of the function being described.
```
yguclu, line 57: Not sure what you mean with "Objects inheriting from `PyccelInternalFunction`". Did you mean "Instances of the subclasses of `PyccelInternalFunction`".?

EmilyBourne: Yes. Fixed 6a13e18fc

developer_docs/ast_nodes.md
EmilyBourne: Fixed 1863e097f

developer_docs/ast_nodes.md
EmilyBourne: Fixed 1863e097f

developer_docs/ast_nodes.md
yguclu: It seems to me that the word "therefore" does not provide the necessary explanation. Did you want to say that the actual arguments to the `FunctionCall` constructor could be either positional or keyword-based, while the `__init__` of `PyccelInternalFunction` only accepts positional arguments and hence cannot process keyword arguments correctly?

EmilyBourne: Better: ecaaa4fd2 ?

yguclu: Better! It could be a bit more explicit though.

EmilyBourne: 18c27f4f8 ?

developer_docs/ast_nodes.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Pyccel Internal Function
+
+The class `pyccel.ast.internals.PyccelInternalFunction` is a super class. This class should never be used directly but provides functionalities which are common to certain AST objects. These AST nodes are those which describe functions which are supported by Pyccel. For example it is used for functions from the `math` library, the `cmath` library, the `numpy` library, etc. `PyccelInternalFunction` inherits from `TypedAstNode`. The type information for the sub-class describes the type of the result of the function.
+
+Instances of the subclasses of `PyccelInternalFunction` are created from a `FunctionCall` object resulting from the syntactic stage. The `PyccelInternalFunction` objects are initialised by passing the contents of the `FunctionCallArgument`s used in the `FunctionCall` to the constructor. These arguments may be either positional or keyword arguments. It is therefore important that the names of the arguments and keyword arguments of any classes inherited from `PyccelInternalFunction` match the names of the arguments and keyword arguments of the function being described.
```
yguclu:
```suggestion
Instances of the subclasses of `PyccelInternalFunction` are created from a `FunctionCall` object resulting from the syntactic stage. The `PyccelInternalFunction` objects are initialised by passing the contents of the `FunctionCallArgument`s used in the `FunctionCall` to the constructor. The actual arguments may be either positional or keyword arguments, but the constructor of `PyccelInternalFunction` only accepts positional arguments through the variadic `*args`. It is therefore important that any subclasses of `PyccelInternalFunction` provide their own `__init__` with the correct function signature.
```

yguclu: Your change was quicker than my suggestion 😅

developer_docs/ast_nodes.md
yguclu: Either that, or what I suggested below. Or maybe a mix of them. I let you choose.

EmilyBourne: I've mixed them: [ac4fc59](https://github.com/pyccel/pyccel/pull/1616/commits/ac4fc59a22a3fa5d4b6735637dfc9f54a7e4195e)

yguclu: Excellent!

yguclu (approved):

PR #1624, opened 2023-11-18
Fix exit statement in Fortran with Intel compiler

harsha-mangena:
The Intel compiler does not allow specifying the precision in a `stop` statement. To fix this, a condition was added in `_print_SysExit` to print `LiteralInteger` objects with default precision. Fixes https://github.com/pyccel/pyccel/issues/1554.


pyccel/codegen/printing/fcode.py
```diff
@@ -2640,15 +2640,17 @@ def _print_Header(self, expr):
...
             print_arg = FunctionCallArgument(expr.status)
             code = self._print(PythonPrint((print_arg, ), file="stderr"))
             arg = "1"
         else:
-            arg = expr.status
-            if arg.precision != 4:
-                arg = NumpyInt32(arg)
-            arg = self._print(arg)
+            # Ensure no type specifier is added for NativeInteger
+            arg = self._print(expr.status)
```
EmilyBourne, line 2651: I think this is one of the rare exceptions where it is not worth calling `self._print` as you then have to correct what it did. Why not just convert the integer to a string directly?

harsha-mangena: @EmilyBourne , I have added the  change for `int` to `str`.

pyccel/codegen/printing/fcode.py
```diff
@@ -2640,15 +2640,14 @@ def _print_Header(self, expr):
...
+        elif expr.status.dtype is not NativeInteger() or expr.status.rank > 0:
             print_arg = FunctionCallArgument(expr.status)
             code = self._print(PythonPrint((print_arg, ), file="stderr"))
             arg = "1"
         else:
-            arg = expr.status
-            if arg.precision != 4:
-                arg = NumpyInt32(arg)
-            arg = self._print(arg)
+            arg = str(expr.status)
```
EmilyBourne, line 2650: I'm not sure that this change is a good idea. The `if` above will handle the literal integer case, but for non-literal integers (e.g. expressions/variables) we should keep the conversion and the call to `_print`. Does this pass the tests as written?

EmilyBourne:
Hi @harsha-mangena , are you still working on this PR? Do you need a hand or want me to complete it for you?

harsha-mangena:
Hi @EmilyBourne, I couldn't put time on this as its my end semester week, Will work on this on this weekend. 
It would be helpful for me If you can guide me through this.

EmilyBourne:
@harsha-mangena No problem. Great news if you will have time this weekend :)
I don't think there's too much left to do. You need to:
1. Interact with the bot to get the checklist of jobs to complete the PR
    As described by the bot the method for doing this is to leave a comment containing `/bot checklist`.
    The bot will then modify this comment so it contains a checklist where you can tick things off.
    It's mostly basic things which are easily forgotten (e.g. CHANGELOG)
2. Fix the bugs. As I hinted in my last comment on your code the current implementation doesn't quite work. For example:
    - the test `tests/pyccel/scripts/exits/positive_exit3.py` is failing with:
        ```
            /home/runner/work/pyccel/pyccel/tests/pyccel/scripts/exits/__pyccel__/__pyccel__gw0/prog_positive_exit3.f90:11:6:

           11 |   stop exit_code
              |      1
        Error: STOP code at (1) must be default integer KIND=4
        ```
        To fix this, non-literals need to be converted to `precision=4` (as was done in the original code)
    - The test `tests/pyccel/scripts/exits/negative_exit2.py` is failing with:
        ```
        home/runner/work/pyccel/pyccel/tests/pyccel/scripts/exits/__pyccel__/__pyccel__gw0/prog_negative_exit2.f90:8:6:
                8 |   stop -Literal(345)
                  |      1
        Error: STOP code at (1) must be either INTEGER or CHARACTER type
        ```
        To fix this you need to make sure you are using the `_print` function whenever you want to print an instance of a class. (For LiteralInteger you probably need `expr.status.python_value` to get the underlying `int`)
    - You can find the list of relevant tests here:
         https://github.com/pyccel/pyccel/blob/6e6f088edc909ae339d52a5b46b0c2f0553be0d9/tests/pyccel/test_pyccel.py#L955-L962
         If you have managed to install pyccel then you should be able to run the command line tool on these files to see the compiler errors.

harsha-mangena:
@EmilyBourne,
I regret the delayed response, I had tried this in several ways but I don't know why the code isn't working and test cases as well. So I kindly request if this issue is in priority, Please assign this to someone.

EmilyBourne:
@harsha-mangena no problem for the slow reply, especially over the holidays!
This is not a priority issue. As far as I am aware there are very few people using exit statements and even fewer using intel. So if you want to keep going then feel free. I'm not quite sure where you're stuck though. I think between your first version and this version you have at some point implemented working code for each section.
If you have more specific questions then you're welcome to contact me on discord. But if you'd rather leave it then I'm happy to take over. Whatever you prefer

EmilyBourne (approved):

yguclu (approved):

PR #1627, opened 2023-11-18
Remove GCC intrinsics from Fortran printing of sign function

### LINKED ISSUE #1553: Remove GCC Intrinsics
### Describe the bug
The code in this file: https://github.com/pyccel/pyccel/blob/devel/pyccel/stdlib/numpy/numpy_f90.f90
uses GCC intrinsics. As a result it does not compile with the intel compiler.

### To Reproduce
```python
def literal_complex_pos():
    import numpy as np
    b = np.sign(1+2j)
    return b
```

### Error details
```bash/fortran/c
E           RuntimeError: Failed to build module
E           /home/runner/work/pyccel/pyccel/__epyccel__gw0/__pyccel__gw0/numpy/numpy_f90.f90(126): error #7137: Any procedure referenced in a PURE procedure, including one referenced via a defined operation or assignment, must have an explicit interface and be declared PURE.   [REALPART]
E                 x_ne_zero = (REALPART(x) .ne. 0_f32) .or. (IMAGPART(x) .ne. 0_f32)
E           -------------------^
```
See : https://github.com/pyccel/pyccel/actions/runs/6511414122/job/17687068773

### Expected behaviour
The code should compile.

### Language
Fortran

### Additional context
There is a `SIGN` intrinsic in Fortran but it only handles integers and floating point numbers. It may make more sense to use this rather than our own functions where possible

### Implementation
For complexes a small benchmark test should be run to determine what is faster between:
1. `np.sign(c.real) + (1-np.abs(np.sign(c.real)))*np.sign(c.imag)`
2. `np.sign(x.real) if x.real != 0 else np.sign(x.imag)`

**Option 1**
Correct the code in the file to remove the GCC intrinsics

**Option 2**
Change `pyccel.codegen.fcode._print_NumpySign` to call the fortran [intrinsic](https://fortranwiki.org/fortran/show/sign) directly for integers and floats and correct the code in the file for complexes.

EmilyBourne:
Remove GCC intrinsics in sign function. Instead use Fortran intrinsics. Only use an external function for the complex case which requires intermediate variables for readability. Fixes #1553


pyccel/codegen/printing/fcode.py
```diff
@@ -2684,9 +2684,16 @@ def _print_NumpySign(self, expr):
...
-        return f'numpy_sign({self._print(expr.args[0])})'
+        arg = expr.args[0]
+        arg_code = self._print(arg)
+        if isinstance(expr.dtype, NativeComplex):
+            func = PyccelFunctionDef('numpy_sign', NumpySign)
+            self._additional_imports.add(Import('numpy_f90', AsName(func, 'numpy_sign')))
+            return f'numpy_sign({arg_code})'
+        else:
+            cast_func = DtypePrecisionToCastFunction[expr.dtype.name][expr.precision]
+            abs_result = self._print(cast_func(PythonBool(arg)))
```
bauom, line 2695: What does `PythonBool` achieve here ?

EmilyBourne: It's a cheat's way of checking if `arg!=0` it lets me avoid duplicating the associated code. I'll add a comment

EmilyBourne: ce2e6f439

bauom (approved):

yguclu (approved):

PR #1652, opened 2024-01-08
Fix scoping issue

### LINKED ISSUE #1651: Bad translation of optional class argument
### Describe the bug
The code generated currently for optional class arguments does not work. Presumably there is a Fortran semantic error. This may be related to #1595 

### To Reproduce
```python
class A:
    def __init__(self : 'A', x : int):
        self.x = x

def get_x_from_A(a : 'A' = None):
    if a is not None:
        return a.x
    else:
        return 5
```

### Error details
Provide the generated code:
```fortran
module classes_7


  use, intrinsic :: ISO_C_Binding, only : i64 => C_INT64_T , b1 => &
        C_BOOL
  implicit none

  type, public :: A
    integer(i64) :: x
    logical(b1) :: is_freed

    contains
    procedure :: create => A_create
    procedure :: free => A_free
  end type A

  contains


  subroutine A_create(self, x)

    implicit none

    class(A), intent(inout) :: self
    integer(i64), value :: x

    self%is_freed = .False._b1
    self%x = x

  end subroutine A_create

  subroutine A_free(self)

    implicit none

    class(A), intent(inout) :: self

    if (.not. self%is_freed) then
      ! pass
      self%is_freed = .True._b1
    end if

  end subroutine A_free

  function get_x_from_A(a) result(Out_0001)

    implicit none

    integer(i64) :: Out_0001
    class(A), optional, intent(in) :: a

    if (present(a)) then
      Out_0001 = a%x
      return
    else
      Out_0001 = 5_i64
      return
    end if

  end function get_x_from_A

end module classes_7
```
and the error message:
```
/home/emily/Code/pyccel/tests/pyccel/scripts/classes/__pyccel__/classes_7.f90:61:13:

   61 |     class(A), optional, intent(in) :: a
      |             1
Error: Derived type ‘a’ at (1) is being used before it is defined
/home/emily/Code/pyccel/tests/pyccel/scripts/classes/__pyccel__/classes_7.f90:64:20:

   64 |       Out_0001 = a%x
      |                    1
Error: Symbol ‘a’ at (1) has no IMPLICIT type
/home/emily/Code/pyccel/tests/pyccel/scripts/classes/__pyccel__/classes_7.f90:56:25:

   56 |   function get_x_from_A(a) result(Out_0001)
      |                         1
Error: Symbol ‘a’ at (1) has no IMPLICIT type
```

### Language
Fortran

EmilyBourne:
Ensure that the symbols in all enclosing scopes are checked when looking for collisions. This fixes #1651 where the symbol associated with the type `A` was not present in the function scope but was correctly defined in the global scope.

Fixing this brought to light a few places in the wrapper where the scoping was not correctly used. Notably not all variables inserted into the scope provided their original name, which lead to unnecessary calls to `get_expected_name`. The names of the arguments and results of the C call are defined in a loop to improve readability and make it simpler to match the logic used in `_wrap_BindCFunctionDefResult` where basic scalar types inherit their logic from `_wrap_FunctionDefResult`.


EmilyBourne:
The test already exists. It is not clear why it is passing on the CI

pyccel/codegen/wrapper/c_to_python_wrapper.py
```diff
@@ -782,7 +792,7 @@ def _wrap_FunctionDef(self, expr):
         # is an ndarray.
         for a in original_c_args:
             orig_var = getattr(a, 'original_function_argument_variable', a.var)
-            v = self.scope.find(self.scope.get_expected_name(orig_var.name), category='variables')
+            v = self.scope.find(orig_var.name, category='variables')
```
bauom, line 795: why we are not using `self.scope.get_expected_name` anymore

EmilyBourne: The `find` function of `Scope` is used to get a `Variable` object from the `PyccelSymbol` containing its name. The `PyccelSymbol` contains the name used in Python, while the `Variable` contains the name used in the target language. The 2 objects are mapped to one another with `Scope._locals` which is populated in `insert_variable`. The second argument of `insert_variable` ensures that the correct symbol is used for the key. As mentioned in the PR this argument was missing in several places so this look-up wasn't working correctly. I seem to have got muddled up when originally writing this as instead of adding the missing arguments I added calls to `get_expected_name`.

The function `get_expected_name` is used to get the name of an object in the target language from its Python name. This is usually used in the semantic stage to create a `Variable` it should not be used to call `find`.

Does this make sense? I guess the current docs for the Scope in [developer_docs](https://github.com/pyccel/pyccel/blob/devel/developer_docs/scope.md) and the docstrings are not sufficient?

bauom: now i got it i haven't interacted with the `find` much
in the times i did i always used `get_expected_name`
it is more clear now.

developer_docs/scope.md
```diff
@@ -21,6 +21,12 @@ The `Scope` object keeps track of all names used in the described scope. This me
...
 
-3.  All newly created variables must be inserted into the scope using the `insert_variable` function.
+3.  All newly created variables must be inserted into the scope using the `insert_variable` function. In order to be able to locate the variable later the second argument of the `insert_variable` function (the name of the variable in the Python code) should be provided.
 
 Temporary variables can also be created using the `get_temporary_variable` function. In this case it is not necessary to use the `insert_variable` function.
+
+## Locating objects in the scope
+
+The `find` function allows objects to be collected from the scope after they have been inserted. The function uses the `PyccelSymbol` (originating in the syntactic stage and therefore matching the Python name of the object) to locate the associated variable/class/etc (originating in the semantic stage and therefore matching the new name of the object). This difference in naming is why it is important to provide the Python name to the `insert_variable` function. The find function can be accelerated a little by only looking in the category of object that we are expecting (e.g. variables not classes).
```
yguclu, line 30:
```suggestion
The `find` function allows objects to be collected from the scope after they have been inserted. The function uses the `PyccelSymbol` (originating in the syntactic stage and therefore matching the Python name of the object) to locate the associated variable/class/etc (originating in the semantic stage and therefore matching the new name of the object). This difference in naming is why it is important to provide the Python name to the `insert_variable` function. The `find` function can be accelerated a little by only looking in the category of object that we are expecting (e.g. variables not classes).
```

pyccel/codegen/wrapper/c_to_python_wrapper.py
yguclu: I think the docs are very clear now. Can we resolve this conversation, @bauom?

pyccel/parser/scope.py
```diff
@@ -476,7 +476,7 @@ def insert_symbol(self, symbol):
                 self.parent_scope.insert_symbol(symbol)
             elif symbol not in self._used_symbols:
                 collisionless_symbol = self.name_clash_checker.get_collisionless_name(symbol,
-                        self._used_symbols.values())
+                        self.all_used_symbols)
```
yguclu, line 479: The properties `local_used_symbols` and `all_used_symbols` are very useful and should be mentioned in the developers docs, if they are not already!

By the way, it appears that `local_used_symbols` returns a `dict` while `all_used_symbols` returns a `set`. What is the reason for this choice?

yguclu (changes requested):
Looks good to me! I have a few comments.

pyccel/parser/scope.py
EmilyBourne: `local_used_symbols` gets the `_used_symbols` object which is a dictionary. It is used for initialising new scopes, usually for initialising the semantic equivalent of a syntactic scope.

`all_used_symbols` gets all the symbols which are currently used in the scope (including enclosing scopes). It is almost exclusively used for obtaining new names while avoiding collisions with existing names.

These are not currently mentioned in the developers docs as I didn't have a place in mind where developers frequently need these objects. These docs are supposed to supplement the docstrings not replace them entirely so I would hope that devs would go looking for functions if they need them. I can add something though

EmilyBourne: [4f10089](https://github.com/pyccel/pyccel/pull/1652/commits/4f100894c7ff46b4a5a7c75a670c2f3c8322a798)

yguclu (approved):

PR #1653, opened 2024-01-10
Manage Pylint errors via bot

### LINKED ISSUE #1648: [CODACY] Pylint errors not raised
Codacy has recently added a new restriction to their free version. Now only the first 50 errors are reported. This means that in files with lots of errors the pylint check is no longer complaining. It is therefore important that we stop using Codacy ASAP. Up till now we have been working towards this by fixing errors in files before adding them to our own tests however this a slow process. Without these checks the code will swiftly deteriorate.

In order to fix this without solving all the pylint issues we therefore need to add a test which compares the pylint output on a branch to the pylint output on the master branch and reports and new issues

EmilyBourne:
Modify the pylint test. Pylint is now run on all files with a Python extension. The git diff is used to filter the results to avoid an extreme number of errors. Errors are only reported on lines that were changed. This means new unused-imports will not be reported for now. This mostly removes the necessity for Codacy for Python files (although it may still be useful for some corner cases) and fixes #1648 . Errors are not reported for cyclic-imports as they seem to be reported on lines unrelated to the error.

Additionally some pylint errors are fixed and the ci_tools folder is added to the PYTHONPATH for the docs test to prevent import issues causing failing tests

### DIFF
--- a/ci_tools/parse_pylint_output.py
+++ b/ci_tools/parse_pylint_output.py
@@ -7,15 +7,25 @@ import sys
 
 PylintMessage = namedtuple('PylintMessage', ['file','line', 'position', 'message'])
 
-if __name__ == '__main__':
-    parser = argparse.ArgumentParser(description='Parse pylint output and format the output for neat bot results.')
-    parser.add_argument('pylint', type=str,
-                            help='The file containing the pylint output')
-    parser.add_argument('output', metavar='output', type=str,
-                            help='File where the markdown output will be printed')
-    args = parser.parse_args()
+def get_pylint_results(filename):
+    """
+    Extract pylint results from a file.
 
-    with open(args.pylint, 'r', encoding='utf-8') as p_file:
+    Parse a file containing pylint results and save the results
+    to a dictionary whose keys are modules and whose values are
+    PylintMessage objects.
+
+    Parameters
+    ----------
+    filename : str
+        The name of the file being parsed.
+
+    Returns
+    -------
+    dict
+        The dictionary containg the pylint results.
+    """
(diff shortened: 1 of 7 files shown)


ci_tools/parse_pylint_output.py
```diff
@@ -59,6 +114,7 @@
         json.dump(json_data, json_file)
     with open(args.output, mode='a', encoding="utf-8") as md_file:
         md_file.write(output)
+    print(output)
```
yguclu, line 117: Do you really want to print the output to terminal? Maybe we can add an option for this?

EmilyBourne: I can remove it, but I tend to think it is a little odd if there is no information in the terminal output of a failing test. This output is human readable.

As an example here:
https://github.com/pyccel/pyccel/actions/runs/7476864741/job/20348250262
The contents of `output` were:
```
# Pylint errors found
## Errors found in module ci_tools/parse_pylint_output.py
-  On line [41](https://github.com/pyccel/pyccel/actions/runs/7476864741/job/20348250262#step:7:42) : Unused variable 'code' (unused-variable)
```
This is what would be removed if the print didn't exist

EmilyBourne: These scripts are only ever really run from the CI as they require a certain amount of effort to set up (e.g. getting the diff with the target branch) so I don't think the output is really critical

yguclu: All right!

yguclu (approved):
Looks good to me

PR #1672, opened 2024-01-11
Update Intel environment

EmilyBourne:
Update the Intel environment to avoid incompatibilities due to different versions for the C/C++ compiler and the Fortran compiler. Create issues and mark all failing tests for Intel. Change the default Intel compilers to the new versions from Intel-OneApi. Intel are deprecating `icc` and `ifort` this year.

### DIFF
--- a/tests/epyccel/test_epyccel_mod.py
+++ b/tests/epyccel/test_epyccel_mod.py
@@ -12,8 +12,8 @@ if sys.platform == 'win32':
     RTOL = 1e-13
     ATOL = 1e-14
 elif os.environ.get('PYCCEL_DEFAULT_COMPILER', 'GNU') == 'intel':
-    RTOL = 1e-11
-    ATOL = 1e-14
+    RTOL = 1e-10
+    ATOL = 1e-10
 else:
     RTOL = 2e-14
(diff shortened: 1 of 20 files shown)


tests/epyccel/test_epyccel_mod.py
```diff
@@ -12,8 +12,8 @@
     RTOL = 1e-13
     ATOL = 1e-14
 elif os.environ.get('PYCCEL_DEFAULT_COMPILER', 'GNU') == 'intel':
-    RTOL = 1e-11
-    ATOL = 1e-14
+    RTOL = 1e-10
+    ATOL = 1e-10
```
yguclu, line 16: These tolerances are huge compared to the default ones for GNU. Are we using excessive optimizations with the Intel compilers?

EmilyBourne: Maybe. I'm not sure. I suspect they have a different algorithm as it is only the modulo operator that gives such different values

yguclu:
Is this PR ready for review? 😃

yguclu (approved):
Looks good!

yguclu (approved):

PR #1673, opened 2024-01-11
Use Python 3.10 with new Intel compiler

yguclu:
When running tests with the Intel compiler on Linux in the `devel` branch, use Python 3.10 to avoid the workflow to hang. 

With Python 3.9 the workflow was hanging as shown here: https://github.com/pyccel/pyccel/actions/runs/7494351871.

See PR #1672, where the same modification was applied to the `run intel` bot command.

Fix a minor bug with the pylint test by handling the case where there are no Python files modified in a PR.

### DIFF
--- a/.github/workflows/intel.yml
+++ b/.github/workflows/intel.yml
@@ -35,7 +35,7 @@ jobs:
     steps:
     - if: github.event_name == 'push'
       run: |
-        echo "version=3.9" >> $GITHUB_ENV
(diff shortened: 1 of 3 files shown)


EmilyBourne (approved):

yguclu:
@EmilyBourne Can you please update the PR description with your changes?

EmilyBourne:
Done

Issue #1645, opened 2024-01-03
DeprecationWarning with Python3.12


EmilyBourne:
### Describe the bug
Python 3.12 is raising a deprecation warning in the syntactic stage due to our use of certain objects/parameters in the `ast` module which will be deprecated.

See the [docs](https://docs.python.org/3/library/ast.html#literals)

![image](https://github.com/pyccel/pyccel/assets/25363473/cb0b5f86-35c2-47fa-9b44-8742c876ec67)

According to the warning these warnings will become errors in version 3.14. Currently 3.8 is the oldest version of Python which has not reached end of life so we should be able to solve half of these issues now

EmilyBourne:
Fixed by #1675

PR #1675, opened 2024-01-14
Use new class `ast.Constant` to avoid deprecation warnings

PriyabrataMo:
Fix issue #1645: Use the new Python `ast` class `Constant` (introduced with Python 3.8) instead of the obsolete `NamedConstant`, `Num` and `Str` (deprecated since Python 3.12, to be removed in Python 3.14).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
--   #1645 : Handle deprecated ast classes.
+-   #1645 : Handle deprecated `ast` classes.
 -   #1649 : Add support for `np.min` in C code.
 -   #1621 : Add support for `np.max` in C code.
 -   #1571 : Add support for the function `tuple`.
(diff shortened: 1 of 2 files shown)


pyccel/parser/syntactic.py
```diff
@@ -538,25 +520,13 @@ def _visit_Constant(self, stmt):
             return LiteralComplex(stmt.value.real, stmt.value.imag)
 
         elif isinstance(stmt.value, str):
-            return self._visit_Str(stmt)
+            if isinstance(self._context[-2], ast.Expr):
+                return CommentBlock(stmt.s)
+            return LiteralString(stmt.s)
```
EmilyBourne, line 525: I think `stmt.s` is part of the `Str` interface but not the `Constant` interface, so it raises the deprecation warning.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
+-   #1645 : Handle deprecated ast classes.
```
EmilyBourne, line 8:
```suggestion
-   #1645 : Handle deprecated `ast` classes.
```

EmilyBourne (approved):

yguclu (approved):

PR #1682, opened 2024-01-19
Fix name collision issue for class methods

### LINKED ISSUE #1507: Method not correctly added in the ClassDef when functions has name collision.
### Describe the bug
The error comes from: https://github.com/pyccel/pyccel/blob/2fb0e0da37382e0ed8c144865547a70bcbb14b8c/pyccel/parser/semantic.py#L3576-L3576
This line should change to : `m = self.scope.functions.pop(cls.scope.get_expected_name(m_name))`.
And check if the method is correctly added to the `ClassDef.methods`.

### To Reproduce
```python
class Point:
     def __init__(self):
         self.x = 1
         self.X = 1

    def get_x(self):
        return self.x

    def get_X(self):
        return self.X
```

### Error details
```bash/fortran/c
File "/home/pyccel/pyccel/parser/semantic.py", line 1633, in _visit
    obj = getattr(self, annotation_method)(expr)
  File "/home/pyccel/pyccel/parser/semantic.py", line 3576, in _visit_ClassDef
    m = self.scope.functions.pop(m_name)
```

### Expected behavior
The Code should work and the methods should be correctly added to the `ClassDef.methods`.

EmilyBourne:
Fix the name collision issue for class methods by using `Scope.get_expected_name` in `ClassDef.get_method` to get the name which should be used for searching. Ensure renamed functions are also added to classes. Add a test. Fixes #1507

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ All notable changes to this project will be documented in this file.
 -   #1259 : Fix bug causing problems with user editable installation.
 -   #1651 : Fix name collision resolution to include parent scopes.
 -   #1156 : Raise an error for variable name collisions with non-variable objects.
+-   #1507 : Fix problems with name collisions in class functions.
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


yguclu:
@EmilyBourne Please resolve the conflict in the semantic parser.

EmilyBourne:
@yguclu Fixed

yguclu (approved):
Looks good to me!

PR #1683, opened 2024-01-22
Add typing.Final and simplify type inference

### LINKED ISSUE #1581: Stop using separate treatment for textx type annotations
### Describe the feature
Textx is used to parse type annotations in strings. This is necessary as the type annotations are not always valid Python code. However the objects generated when parsing an in-language type annotation and a string type annotation are different. This leads to a complex implementation of `SyntacticTypeAnnotation` as it must juggle the storage of the output of textx and the output of non-string annotations. A way to unify the two should be found.

### LINKED ISSUE #1680: Add support for `Final` from the `typing` module
### Describe the feature
[PEP 591](https://peps.python.org/pep-0591/) describes `Final` which is equivalent to `const` in other languages. Introducing support for this object in Pyccel will force us to introduce a class for this which will make type handling easier. Notably it will make it simpler to use the same code in the `SyntacticParser` to treat in-language type annotations and string type annotations.

### Test Code
```python
from typing import Final

a : Final[int] = 1
```

### Proposed Solution
In order to implement this the following steps should be taken:

1. Create a file pyccel/ast/typingext.py
2. In the file create an object which describes the object `Final`
3. Create a module description for the typing module (see pyccel.ast.mathext for an example). For now this module can contain just the Final type.
4. Add the module to the pyccel.ast.utilities.builtin_import_registry
5. In the syntactic stage, make the necessary changes to use this type
6. Add tests

### Related issues
#1544

EmilyBourne:
Add support for `typing.Final`. Fixes #1680 . This means that the AST can describe everything that textx parses in type string annotations. Therefore the type annotation methods are simplified and unified to reduce code duplication. Fixes #1581. This should make it simpler to add support for tuple/list/set type annotations

**Commit Summary**
- Fix pickling of `Variable` and `AnnotatedPyccelSymbol` (lambda expressions can't be pickled) by moving `ast.core.apply`->`ast.core.apply_pickle`
- Remove unused redundant class `UnionType`
- Simplify `SyntacticTypeAnnotation` to only handle one type (instead of all textx output).
- Add support for `typing.Final` (represents `is_const` concept)
- Remove `python_builtin_datatypes` and simply use  `_visit_PyccelSymbol` to collect type.
  It is not useful to differentiate between datatypes and functions. The two dictionaries point at the same objects and Python itself doesn't differentiate between the two. Anything that was missing was moved into buitin_functions.
- Fix unnecessary import of deprecated types from NumPy in generated Python code.
- Stop duplicating textx logic.
- Add `_get_indexed_type` function to handle `IndexedElement` objects which describe a type.
- Use textx class descriptors to generate Pyccel AST elements that can be parsed in the semantic stage.
- Fix bug in test where `float`s were passed instead of NumPy floats.
- Stop raising a warning when the order is provided unnecessarily.
- Fix minor bot bug when whole file is new code.
- Add `is_const` setter to type annotations.
- Fix pyccel-init script which was not calling the related function.

### DIFF
--- a/ci_tools/coverage_analysis_tools.py
+++ b/ci_tools/coverage_analysis_tools.py
@@ -257,7 +257,7 @@ def get_json_summary(untested, content_lines, existing_comments, diff):
             if j < n_code_lines-1:
                 end_line = line_indices[j]-1
             else:
-                end_line = line_indices[j]
+                end_line = line_indices[n_code_lines-1]
             last_valid_line = start_line
             last_valid_line_idx = next(i for i in diff[f]['addition'] if i == last_valid_line)
             for i in diff[f]['addition'][last_valid_line_idx:]:
--- a/ci_tools/coverage_analysis_tools.py
+++ b/ci_tools/coverage_analysis_tools.py
@@ -257,7 +257,7 @@ def get_json_summary(untested, content_lines, existing_comments, diff):
             if j < n_code_lines-1:
                 end_line = line_indices[j]-1
             else:
-                end_line = line_indices[n_code_lines-1]
+                end_line = line_indices[-1]
             last_valid_line = start_line
             last_valid_line_idx = next(i for i in diff[f]['addition'] if i == last_valid_line)
             for i in diff[f]['addition'][last_valid_line_idx:]:
--- a/pyccel/ast/builtins.py
+++ b/pyccel/ast/builtins.py
@@ -54,7 +54,6 @@ __all__ = (
     'PythonType',
     'PythonZip',
     'builtin_functions_dict',
-    'python_builtin_datatype',
 )
 
 #==============================================================================
@@ -1219,32 +1218,10 @@ class PythonType(PyccelAstNode):
             return LiteralString(f"<class 'numpy.{dtype}{precision}'>")
 
 #==============================================================================
-python_builtin_datatypes_dict = {
-    'bool'   : PythonBool,
-    'float'  : PythonFloat,
-    'int'    : PythonInt,
-    'complex': PythonComplex,
-    'str'    : LiteralString
-}
-
-def python_builtin_datatype(name):
-    """
-    Given a symbol name, return the corresponding datatype.
-
-    name: str
-        Datatype as written in Python.
-
-    """
-    if not isinstance(name, str):
-        raise TypeError('name must be a string')
-
-    if name in python_builtin_datatypes_dict:
-        return python_builtin_datatypes_dict[name]
-
-    return None
 
 builtin_functions_dict = {
     'abs'      : PythonAbs,
+    'bool'     : PythonBool,
     'range'    : PythonRange,
     'zip'      : PythonZip,
     'enumerate': PythonEnumerate,
@@ -1258,6 +1235,7 @@ builtin_functions_dict = {
     'min'      : PythonMin,
     'not'      : PyccelNot,
     'map'      : PythonMap,
+    'str'      : LiteralString,
     'type'     : PythonType,
     'tuple'    : PythonTupleFunction,
 }
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -4060,13 +4060,15 @@ class EmptyNode(PyccelAstNode):
 
 
 class Comment(PyccelAstNode):
+    """
+    Represents a Comment in the code.
 
-    """Represents a Comment in the code.
+    Represents a Comment in the code.
 
     Parameters
     ----------
     text : str
-       the comment line
+       The comment line.
 
     Examples
     --------
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -22,7 +22,7 @@ from .datatypes import (datatype, DataType, NativeSymbol, NativeHomogeneousTuple
                         NativeBool, NativeTuple, str_dtype, NativeInhomogeneousTuple,
                         NativeVoid)
 
-from .internals import PyccelSymbol, PyccelInternalFunction, get_final_precision
+from .internals import PyccelSymbol, PyccelInternalFunction, get_final_precision, apply_pickle
 
 from .literals  import Nil, LiteralFalse, LiteralInteger
 from .literals  import NilArgument, LiteralTrue
@@ -108,7 +108,6 @@ __all__ = (
 #      - add a new Idx that uses Variable instead of Symbol
 
 #==============================================================================
-def apply(func, args, kwargs):return func(*args, **kwargs)
 
 #==============================================================================
 def create_variable(forbidden_names, prefix = None, counter = 1):
@@ -698,7 +697,7 @@ class CodeBlock(PyccelAstNode):
            and its arguments.
         """
         kwargs = dict(body = self.body)
-        return (apply, (self.__class__, (), kwargs))
+        return (apply_pickle, (self.__class__, (), kwargs))
 
     def set_current_ast(self, ast_node):
         """
@@ -2828,7 +2827,7 @@ class FunctionDef(ScopedAstNode):
            and its arguments.
         """
         args, kwargs = self.__getnewargs__()
-        out = (apply, (self.__class__, args, kwargs))
+        out = (apply_pickle, (self.__class__, args, kwargs))
         return out
 
 
@@ -4108,7 +4107,7 @@ class Comment(PyccelAstNode):
            and its arguments.
         """
         kwargs = dict(text = self.text)
-        return (apply, (self.__class__, (), kwargs))
+        return (apply_pickle, (self.__class__, (), kwargs))
 
 
 class SeparatorComment(Comment):
--- a/pyccel/ast/type_annotations.py
+++ b/pyccel/ast/type_annotations.py
@@ -118,6 +118,12 @@ class VariableTypeAnnotation(PyccelAstNode):
         """
         return self._order
 
+    @order.setter
+    def order(self, order):
+        if order not in ('C', 'F', None):
+            raise ValueError("Order must be C, F, or None")
+        self._order = order
+
     @property
     def is_const(self):
         """
@@ -131,10 +137,6 @@ class VariableTypeAnnotation(PyccelAstNode):
     @is_const.setter
     def is_const(self, val):
         self._is_const = val
-
-    @order.setter
-    def order(self, order):
-        self._order = order
 
     def __hash__(self):
         return hash((self.datatype, self.class_type, self.precision, self.rank, self.order))
--- a/pyccel/ast/type_annotations.py
+++ b/pyccel/ast/type_annotations.py
@@ -240,7 +240,6 @@ class UnionTypeAnnotation(PyccelAstNode):
     _attribute_nodes = ('_type_annotations',)
 
     def __init__(self, *type_annotations):
-        assert len(type_annotations) > 0
         annots = [ti for t in type_annotations for ti in (t.type_list if isinstance(t, UnionTypeAnnotation) else [t])]
         self._type_annotations = tuple(set(annots))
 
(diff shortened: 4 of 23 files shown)


pyccel/ast/core.py
```diff
@@ -2828,7 +2827,7 @@ def __reduce_ex__(self, i):
            and its arguments.
         """
         args, kwargs = self.__getnewargs__()
-        out = (apply, (self.__class__, args, kwargs))
+        out = (apply_pickle, (self.__class__, args, kwargs))
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept

Does not need testing here.

pyccel/ast/core.py
```diff
@@ -4108,7 +4109,7 @@ def __reduce_ex__(self, i):
            and its arguments.
         """
         kwargs = dict(text = self.text)
-        return (apply, (self.__class__, (), kwargs))
+        return (apply_pickle, (self.__class__, (), kwargs))
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept

Does not need testing here.

pyccel/ast/type_annotations.py
```diff
@@ -205,6 +218,20 @@ def results(self):
...
 
+    @property
+    def is_const(self):
+        """
+        Indicates whether the object will remain constant.
+
+        Returns a boolean which is false if the value of the object can be
+        modified, and true otherwise.
+        """
+        return self._is_const
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept

Simple getter doesn't need a test

pyccel/ast/variable.py
```diff
@@ -580,8 +581,7 @@ def __reduce_ex__(self, i):
             'cls_base':self.cls_base,
             }
 
-        apply = lambda func, args, kwargs: func(*args, **kwargs)
-        out =  (apply, (Variable, args, kwargs))
+        out =  (apply_pickle, (self.__class__, args, kwargs))
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept

Variables may not need pickling but this code is more correct than the old code.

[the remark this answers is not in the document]
EmilyBourne: This will be used when #297 is fixed

pyccel/commands/pyccel_init.py
```diff
@@ -22,8 +22,11 @@ def pyccel_init():
...
-    """ Wrapper around the pyccel_init function removing the need
-    for ArgumentParser
+    """
+    Wrapper around the pyccel_init function removing the need for ArgumentParser.
+
+    Wrapper around the pyccel_init function removing the need for ArgumentParser.
     """
     parser = ArgumentParser(description='Pickle internal pyccel files')
     parser.parse_args()
+    pyccel_init()
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept

Not testable here

pyccel/ast/type_annotations.py
```diff
@@ -205,6 +218,20 @@ def results(self):
...
+
+    @is_const.setter
+    def is_const(self, val):
+        self._is_const = val
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept

This doesn't really make sense at the moment but it is necessary to describe everything that is possible according to our type grammar.

pyccel/ast/variable.py
```diff
@@ -1082,3 +1082,6 @@ def annotation(self):
 
     def __str__(self):
         return f'{self.name} : {self.annotation}'
+
+    def __reduce_ex__(self, i):
+        return (self.__class__, (self.name, self.annotation))
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
Tested in installation (tests fail without this code)

EmilyBourne:
Coverage errors are reported on pickling elements. Often these are tested during the installation however the coverage cannot see this

pyccel/ast/builtins.py
```diff
@@ -1219,32 +1218,10 @@ def print_string(self):
...
-        Datatype as written in Python.
-
-    """
-    if not isinstance(name, str):
-        raise TypeError('name must be a string')
-
-    if name in python_builtin_datatypes_dict:
-        return python_builtin_datatypes_dict[name]
-
-    return None
```
Farouk-Echaref, line 1244: Can you please explain what happened here? I was just recently reading this part.

EmilyBourne: I don't think it is useful for us to differentiate between datatypes and functions. The two dictionaries point at the same objects and Python itself doesn't differentiate between the two. So I moved anything that was missing into `buitin_functions` and removed this duplication.

Good job you were checking the PRs to notice it changed ;)

Farouk-Echaref (commented):
Needed clarification on some details in the code that I haven't used yet, but overall, everything seems good.

pyccel/ast/core.py
```diff
@@ -4061,13 +4060,15 @@ def __str__(self):
 
 
 class Comment(PyccelAstNode):
+    """
+    Represents a Comment in the code.
 
-    """Represents a Comment in the code.
+    Represents a Comment in the code.
```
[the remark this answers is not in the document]
EmilyBourne: Yes :sweat_smile: . The documentation checker asks for a short description and an extended description to provide more details, however in a lot of cases there aren't really any extra details to provide.

pyccel/ast/type_annotations.py
```diff
@@ -269,141 +296,49 @@ class SyntacticTypeAnnotation(PyccelAstNode):
...
-        if any(not isinstance(r, int) for r in ranks):
-            raise ValueError("Ranks should have integer values")
-        if not all(o is None or isinstance(o, str) for o in orders):
-            raise ValueError("Orders should be strings")
-        if not (isinstance(is_const, bool) or is_const is None):
-            raise ValueError("Is const should be a boolean")
-        self._dtypes = tuple(dtypes)
-        self._ranks = tuple(ranks)
-        self._orders = tuple(o if o != '' else None for o in orders)
-        self._is_const = is_const
```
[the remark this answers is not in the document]
EmilyBourne: There is no need to check the `_ranks` attribute as it no longer exists in the class. Instead the rank is included in the dtype information. For example `'float[:,:]'` was previously stored as `SyntacticTypeAnnotation(['float'], [2], [None], None)`, but when adding support for `tuple[int,..]` I noticed that this meant I needed to store it as: `SyntacticTypeAnnotation([IndexedElement(PythonTuple, int)], [1], [None], None)` which duplicates the information.

Now `'float[:,:]'` is stored as `SyntacticTypeAnnotation(IndexedElement(PythonFloat, Slice(None, None), Slice(None, None), None, None)`. This will allow us to easily support syntax such as:
```
from pyccel.types import float64
a : float64[:,:]
```
(as suggested in #1487) as the string syntax and the non-string syntax will be equivalent.

pyccel/ast/core.py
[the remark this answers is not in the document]
Farouk-Echaref: thanks for explaining.

Farouk-Echaref (approved):
Features I haven't used yet have helped me gain more insights for the project.

ci_tools/coverage_analysis_tools.py
```diff
@@ -257,7 +257,7 @@ def get_json_summary(untested, content_lines, existing_comments, diff):
             if j < n_code_lines-1:
                 end_line = line_indices[j]-1
             else:
-                end_line = line_indices[j]
+                end_line = line_indices[n_code_lines-1]
```
yguclu, line 260: Why not just using `-1` as index?

EmilyBourne: No good reason. Fixed: 0c0a6cca

pyccel/ast/builtins.py
yguclu: @EmilyBourne Can you please add a point to the PR description about these changes?

yguclu: I mean, there is a point but it does not elaborate...

EmilyBourne: @yguclu Done

pyccel/ast/type_annotations.py
```diff
@@ -128,6 +128,14 @@ def is_const(self):
         """
         return self._is_const
 
+    @is_const.setter
+    def is_const(self, val):
+        self._is_const = val
+
+    @order.setter
+    def order(self, order):
+        self._order = order
```
yguclu, line 137: If you do not mind, I would move this setter to just after the corresponding getter.

By the way, it could be useful to check that `order in ['C', 'F']`.

pyccel/ast/type_annotations.py
```diff
@@ -128,6 +128,14 @@ def is_const(self):
         """
         return self._is_const
 
+    @is_const.setter
+    def is_const(self, val):
+        self._is_const = val
```
yguclu, line 133: Maybe check that `isinstance(val, bool)`?

pyccel/ast/type_annotations.py
EmilyBourne: Done 15c24c13

pyccel/ast/type_annotations.py
yguclu, line 235: Same comment as above about checking the type

pyccel/ast/type_annotations.py
EmilyBourne: Fixed 8f0dd0d6

pyccel/ast/type_annotations.py
EmilyBourne: Fixed f6c7ee3e

pyccel/codegen/printing/pycode.py
```diff
@@ -250,7 +251,7 @@ def _print_FunctionDefArgument(self, expr):
         default = ''
 
         if expr.annotation:
-            type_annotation = self._print(expr.annotation)
+            type_annotation = f"'{self._print(expr.annotation)}'"
```
yguclu, line 254: Sorry, I do not understand this change. `self._print` should always return a string, right?

EmilyBourne: Yes but the string is not necessarily quoted. Note the `"'` where `"` denotes a string and `'` will appear in the string `type_annotation`

EmilyBourne: Previously I was letting `_print_SyntacticTypeAnnotation` but this is unhelpful if you need to combine it with `_print_UnionTypeAnnotation`, `_print_FunctionTypeAnnotation` or `_print_TypingFinal`

pyccel/commands/pyccel_init.py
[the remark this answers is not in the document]
yguclu: Was the `pyccel-init` command broken?

pyccel/codegen/printing/pycode.py
yguclu: I looked at it many times but my brain could not see the quotes! 😅

pyccel/commands/pyccel_init.py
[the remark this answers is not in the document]
EmilyBourne: It seems so. It wasn't critical as the `pyccel_init` function was called directly from `setup.py` and there is usually no reason to call it again later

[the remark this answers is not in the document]
EmilyBourne: It was mentioned in the CHANGELOG. I have added it to the PR description

pyccel/ast/type_annotations.py
```diff
@@ -205,6 +222,22 @@
...
+        Indicates whether the object will remain constant.
+
+        Returns a boolean which is false if the value of the object can be
+        modified, and true otherwise.
+        """
+        return self._is_const
+
+    @is_const.setter
+    def is_const(self, val):
+        if not isinstance(val, bool):
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept

This doesn't really make sense at the moment but it is necessary to describe everything that is possible according to our type grammar.

PR #1704, opened 2024-01-29
Make `PythonComplex` a subclass of `PyccelInternalFunction`

### LINKED ISSUE #1703: Fixing class inheritance of PythonComplex
## Description
`PythonComplex` represents a call to Python's native `complex()` function. This class accepts arguments to construct complex number objects. To accurately reflect that this is a function call and will be displayed as a function in C/Fortran, it should be represented as a `PyccelInternalFunction`. The previous code doesn't result in any errors but i believe it is a good enhancement.

### Class Prototype Before:
```python
class PythonComplex(TypedAstNode):
```

### Class Prototype After:
```python
class PythonComplex(PyccelInternalFunction):
```

Farouk-Echaref:
Change the base class of `PythonComplex` from `TypedAstNode` to `PyccelInternalFunction`. This fixes #1703.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -28,3 +28,4 @@ Contributors
 * Soukaina Hikma
 * Mudit Loya
 * Priyabrata Mondal
+* Farouk ECH-CHAREF
--- a/AUTHORS
+++ b/AUTHORS
@@ -28,4 +28,4 @@ Contributors
 * Soukaina Hikma
 * Mudit Loya
 * Priyabrata Mondal
-* Farouk ECH-CHAREF
+* Farouk Ech-Charef
--- a/pyccel/ast/builtins.py
+++ b/pyccel/ast/builtins.py
@@ -236,7 +236,7 @@ class PythonBool(PyccelInternalFunction):
         return f'Bool({self.arg})'
 
 #==============================================================================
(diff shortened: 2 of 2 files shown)


EmilyBourne (approved):

AUTHORS
```diff
@@ -28,3 +28,4 @@ Contributors
 * Soukaina Hikma
 * Mudit Loya
 * Priyabrata Mondal
+* Farouk ECH-CHAREF
```
yguclu, line 31: We do not capitalize last names.

yguclu (changes requested):

AUTHORS
yguclu: Sorry, I meant that we do not use ALL CAPS for names.

yguclu: Would this work for you?
```suggestion
* Farouk Ech-Charef
```

Farouk-Echaref: Yes it's fine, thanks for the clarification

Farouk-Echaref: Should I commit the suggested change, or is it for you to decide?

yguclu: You can commit the suggested change

yguclu (approved):

PR #1707, opened 2024-01-31
Allow class usage in a project

### LINKED ISSUE #1650: Wrapper class requires a header file
### Describe the bug
Currently the C-Python wrapper is a C file. It was not considered necessary to add a .h file as wrappers do not need to be called from other wrappers. However with the introduction of classes this is no longer true. In order for other modules to use types declared in a module they must be able to access the structures describing those types.

### Example
File1:
```python
class A:
    def __init__(self, x : int):
        self.x = xclass A:
    def __init__(self, x : int):
        self.x = x
```
File2:
```python
from File1 import A

def f():
    return A(3)
```
The wrapper for File2 will need the description of `A` from the wrapper of File1 in order to return an object with the same type.

### Additional context
This issue describes the state of the code after #1604 is merged.

EmilyBourne:
Add a header file to the wrapper. The header file has different behaviour depending on where it is included from. The implementation follows the tutorial here: https://docs.python.org/3/extending/extending.html#using-capsules.
When the header is imported from the associated wrapper, it contains the declarations of all types and functions used in the module. When it is imported from another file it contains a definition of a Capsule and macros which expose the type declarations. Fixes #1650 .

**Commit Summary**
- Modify the pylint flags from `-rXx` to `-rx`. Pylint was updated and reports now show backtraces making it hard to find errors.
- Add the datatype `CStackArray` and associated class definition `StackArrayClass` to print C-style arrays (e.g. `float a[3]`)
- Change inheritance of `PyBuildValueNode` and `PyModule_AddObject` to inherit from `PyccelInternalFunction` and indicate that they represent functions
- Remove `Errors.parser_scope` (superceded by `pyccel.utilities.stage.PyccelStage`)
- Allow adding extra library dependencies to a `CompileObj`
- Allow a compiler to link libraries using absolute paths
- Expose the Python shared library suffix stored in a compiler
- Ensure that the names of functions in imported classes match those that were printed by renaming (when stub files are introduced this should become possible via annotations)
- Print a `ModuleHeader` for a C-Python wrapper
- Add translated modules as library dependencies for wrappers
- Wrap imports to use classes from other modules in wrappers
- Add a test involving classes in a project

### DIFF
--- a/pyccel/ast/c_concepts.py
+++ b/pyccel/ast/c_concepts.py
@@ -7,14 +7,29 @@
 Module representing object address.
 """
 
-from .basic import TypedAstNode, PyccelAstNode
+from .basic     import TypedAstNode, PyccelAstNode
+from .datatypes import DataType
 from .literals  import LiteralString
+from pyccel.utilities.metaclasses import Singleton
 
 __all__ = ('CMacro',
            'CStringExpression',
            'ObjectAddress',
            'PointerCast')
 
+#------------------------------------------------------------------------------
+
+class CStackArray(DataType, metaclass=Singleton):
+    """
+    A data type representing an array allocated on the stack.
+
+    A data type representing an array allocated on the stack.
+    E.g. `float a[4];`
+    """
+    __slots__ = ()
+    _name = 'c_stackarray'
+
+#------------------------------------------------------------------------------
 class ObjectAddress(TypedAstNode):
     """
     Class representing the address of an object.
@@ -58,7 +73,11 @@ class ObjectAddress(TypedAstNode):
         """
         return self._obj
 
+    @property
+    def is_alias(self):
+        return True
 
+#------------------------------------------------------------------------------
 class PointerCast(TypedAstNode):
     """
     A class which represents the casting of one pointer to another.
--- a/pyccel/ast/c_concepts.py
+++ b/pyccel/ast/c_concepts.py
@@ -7,10 +7,10 @@
 Module representing object address.
 """
 
+from pyccel.utilities.metaclasses import Singleton
 from .basic     import TypedAstNode, PyccelAstNode
 from .datatypes import DataType
 from .literals  import LiteralString
-from pyccel.utilities.metaclasses import Singleton
 
 __all__ = ('CMacro',
            'CStringExpression',
@@ -75,6 +75,11 @@ class ObjectAddress(TypedAstNode):
 
     @property
     def is_alias(self):
+        """
+        Indicate that an ObjectAddress uses alias memory handling.
+
+        Indicate that an ObjectAddress uses alias memory handling.
+        """
         return True
 
 #------------------------------------------------------------------------------
--- a/pyccel/ast/c_concepts.py
+++ b/pyccel/ast/c_concepts.py
@@ -13,6 +13,7 @@ from .datatypes import DataType
 from .literals  import LiteralString
 
 __all__ = ('CMacro',
+           'CStackArray',
            'CStringExpression',
            'ObjectAddress',
            'PointerCast')
--- a/pyccel/ast/numpy_wrapper.py
+++ b/pyccel/ast/numpy_wrapper.py
@@ -229,7 +229,7 @@ def find_in_numpy_dtype_registry(var):
         return numpy_dtype_registry[(dtype, prec)]
     except KeyError:
         return errors.report(PYCCEL_RESTRICTION_TODO,
-                symbol = "{}[kind = {}]".format(dtype, prec),
+                symbol = f"{dtype}[kind = {prec}]",
                 severity='fatal')
 
 def array_type_check(py_variable, c_variable, raise_error):
--- a/pyccel/ast/numpy_wrapper.py
+++ b/pyccel/ast/numpy_wrapper.py
@@ -125,6 +125,8 @@ array_get_data  = FunctionDef(name   = 'nd_data',
                            arguments = [FunctionDefArgument(Variable(dtype=NativeVoid(), name = 'o', is_optional=True))],
                            results   = [FunctionDefResult(Variable(dtype=NativeVoid(), name = 'v', memory_handling='alias', rank = 1, class_type = NativeVoid()))])
 
+import_array = FunctionDef('import_array', (), (), ())
+
 # Basic Array Flags
 # https://numpy.org/doc/stable/reference/c-api/array.html#c.NPY_ARRAY_OWNDATA
 numpy_flag_own_data     = Variable(dtype=NativeInteger(),  name = 'NPY_ARRAY_OWNDATA')
--- a/pyccel/ast/numpy_wrapper.py
+++ b/pyccel/ast/numpy_wrapper.py
@@ -229,7 +229,7 @@ def find_in_numpy_dtype_registry(var):
         return numpy_dtype_registry[(dtype, prec)]
     except KeyError:
         return errors.report(PYCCEL_RESTRICTION_TODO,
-                symbol = f"{dtype}[kind = {prec}]",
+                symbol = "{}[kind = {}]".format(dtype, prec),
                 severity='fatal')
 
 def array_type_check(py_variable, c_variable, raise_error):
--- a/pyccel/codegen/printing/cwrappercode.py
+++ b/pyccel/codegen/printing/cwrappercode.py
@@ -16,7 +16,6 @@ from pyccel.ast.cwrapper   import Py_None, WrapperCustomDataType
 from pyccel.ast.cwrapper   import PyccelPyObject, PyccelPyTypeObject
 from pyccel.ast.literals   import LiteralString, Nil, LiteralInteger
 from pyccel.ast.c_concepts import ObjectAddress
-from pyccel.ast.variable   import Variable
 
 from pyccel.errors.errors  import Errors
 
@@ -25,8 +24,8 @@ __all__ = ("CWrapperCodePrinter", "cwrappercode")
 errors = Errors()
 
 module_imports = [Import('numpy_version', Module('numpy_version',(),())),
-                  Import('numpy/arrayobject', Module('numpy/arrayobject',(),())),
-                  Import('cwrapper', Module('cwrapper',(),()))]
+              Import('numpy/arrayobject', Module('numpy/arrayobject',(),())),
+              Import('cwrapper', Module('cwrapper',(),()))]
 
 
 class CWrapperCodePrinter(CCodePrinter):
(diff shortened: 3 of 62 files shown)


EmilyBourne:
/bot run docs coverage spelling pylint pyccel_lint

EmilyBourne:
/bot run docs coverage spelling pylint pyccel_lint

pyccel/ast/numpy_wrapper.py
```diff
@@ -227,7 +229,7 @@ def find_in_numpy_dtype_registry(var):
         return numpy_dtype_registry[(dtype, prec)]
     except KeyError:
         return errors.report(PYCCEL_RESTRICTION_TODO,
-                symbol = "{}[kind = {}]".format(dtype, prec),
+                symbol = f"{dtype}[kind = {prec}]",
```
EmilyBourne, line 232:
```suggestion
                symbol = "{}[kind = {}]".format(dtype, prec),
```

pyccel/codegen/wrapper/c_to_python_wrapper.py
```diff
@@ -460,64 +502,158 @@ def _build_module_exec_function(self, expr):
...
+        -------
+        API_var : Variable
+            The variable which contains the data extracted from the capsule.
+
+        import_func : FunctionDef
+            The import function.
+
+        See Also
+        --------
+        https://docs.python.org/3/extending/extending.html
```
EmilyBourne, line 620:
```suggestion
        reducing the size of lists).
        See <https://docs.python.org/3/extending/extending.html>
        for more details.

        Parameters
        ----------
        expr : Module
            The module of interest.

        Returns
        -------
        API_var : Variable
            The variable which contains the data extracted from the capsule.

        import_func : FunctionDef
            The import function.
```

pyccel/parser/semantic.py
```diff
@@ -302,7 +302,6 @@ def annotate(self):
         errors = Errors()
         if self.filename:
             errors.set_target(self.filename, 'file')
-        errors.set_parser_stage('semantic')
```
Farouk-Echaref, line 305: what was the need to unset the parser stage in errors for the semantic stage

Farouk-Echaref (approved):
An insightful PR that introduces a new concept for me—Capsules! I hope there will be documentation for their usage in the classes supported by Pyccel

pyccel/codegen/codegen.py
```diff
@@ -147,7 +147,6 @@ def set_printer(self, **settings):
         # instantiate codePrinter
         code_printer = printer_registry[language]
         errors = Errors()
-        errors.set_parser_stage('codegen')
```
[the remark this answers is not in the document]
EmilyBourne: As mentioned in the PR description `Errors.parser_scope` was superceded a while ago by `pyccel.utilities.stage.PyccelStage`.

`pyccel.utilities.stage.PyccelStage` stores the current stage of processing and is used in many places in the code including in the AST. `Errors.parser_scope` stores the Pyccel stage internally to the `Errors` class so that it can be used to indicate the stage in the error message. This is unnecessary duplication that was missed when `pyccel.utilities.stage.PyccelStage` was first implemented. Furthermore the stages for the errors were syntax/semantic/codegen. There was not wrapper stage. It seemed to make more sense to remove the duplication rather than add the wrapper stage to `Errors.parser_scope`

pyccel/ast/c_concepts.py
```diff
@@ -7,14 +7,30 @@
...
+    """
+    A data type representing an array allocated on the stack.
+
+    A data type representing an array allocated on the stack.
+    E.g. `float a[4];`
+    """
+    __slots__ = ()
+    _name = 'c_stackarray'
+
+#------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
EmilyBourne: Written like this, I'm not sure. I added it because I needed to create a `void API_pointers[n_classes];` and this seemed to be the cleanest way of doing it.

Conceptually though it is already possible to allocate on the stack. See : https://github.com/pyccel/pyccel/blob/devel/docs/decorators.md#stack-array

(See also https://fortran-lang.discourse.group/t/why-stack-is-faster-than-heap-and-what-exactly-is-stack/2130/6 for more details about why you might want to use stack instead of heap). I think I allocate tuples on the stack but I'm not sure (I think there was some disagreement about the right way to handle this). In my opinion, tuples should be on the stack as they are usually small objects (I think the counter argument was that this can't be guaranteed). Lists tend to be bigger so they should be allocated on the heap

EmilyBourne:
> An insightful PR that introduces a new concept for me—Capsules! I hope there will be documentation for their usage in the classes supported by Pyccel

@Farouk-Echaref , Capsules were also a new concept for me. Where were you hoping for documentation? I could add something to the user docs but I suspect that this would be irrelevant unless we add a mechanism allowing functions to be exported in capsules (thus allowing users to create packages which can be used by other codes on a low level). I could also add a line to the developer docs to say that class importing is supported via Capsules, but I'm not sure I could add much more that doesn't already appear in Python's documentation.

pyccel/ast/c_concepts.py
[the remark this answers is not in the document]
Farouk-Echaref: I guess both arguments are valid, in a perfect world, there would be a "check first and do" for the available stack memory and the size of the tuple coming from the input,  or at least "try and recover", but i doubt both approaches would be efficient.

Farouk-Echaref:
**I could also add a line to the developer docs to say that class importing is supported via Capsules,**
I guess this would be enough.

pyccel/ast/c_concepts.py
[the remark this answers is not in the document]
EmilyBourne: I agree, it is also tricky to implement. I'm not aware of a (ideally multi-platform) command that tells you the available stack memory. And there is no "try and recover" mechanism for this. Heap is therefore the safest and `@stack_array` can always be used to allocate on the stack if the user thinks there is an advantage

EmilyBourne:
> **I could also add a line to the developer docs to say that class importing is supported via Capsules,** I guess this would be enough.

Done : cdebe01e
Is this clear enough?

yguclu:
@EmilyBourne Please resolve the conflicts with `devel` and update this branch.

EmilyBourne:
@yguclu done

tests/pyccel/project_class_imports/project/shapes/Square_mod.py
```diff
@@ -0,0 +1,25 @@
+# pylint: disable=missing-function-docstring, missing-module-docstring, missing-class-docstring, relative-beyond-top-level
+from ..basics.Point_mod import Point
+
+class Square:
+    def __init__(self, a : Point, b : Point, c : Point, d : Point):
+        self.a = a
+        self.b = b
+        self.c = c
+        self.d = d
```
yguclu, line 9: How can you be sure that this shape is a square? 😜

EmilyBourne: :sweat_smile:

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -245,11 +244,70 @@ def _print_PyModule_AddObject(self, expr):
...
+        mod = expr.module
+        name = mod.name
+
+        # Print imports last to be sure that all additional_imports have been collected
+        imports  = [*module_imports, *mod.imports]
+        for i in imports:
+            self.add_import(i)
+        imports  = ''.join(self._print(i) for i in imports)
+
+        function_signatures = ''.join(self.function_signature(f, print_arg_names = False) + ';\n' for f in mod.external_funcs)
```
yguclu, line 269: If I am not mistaken one-line statements in the printer should end with a newline character, preceded by a semicolon in C. Do we not follow the same convention in this module?

EmilyBourne: The convention is respected for one-line statements, however `self.function_signature` is used in 2 places:
1. Declaration (one-line statement) `void my_func();`
2. Definition (not one-line statement) `void my_func() { ... }`

So in this case the addition of `;\n` is left to the calling function

EmilyBourne: `self.function_signature` mostly relies on the definition coming from ccode

pyccel/codegen/printing/cwrappercode.py
```diff
@@ -400,7 +433,15 @@ def _print_PyClassDef(self, expr):
...
-        return class_code + '\n' + function_sigs + '\n' + method_def + '\n' + type_code + '\n' + functions
+        return method_def + '\n' + type_code + '\n' + functions
+
+    def _print_PyModInitFunc(self, expr):
+        decs = ''.join(self._print(d) for d in expr.declarations)
+        body = self._print(expr.body)
+        return ''.join([f'PyMODINIT_FUNC {expr.name}(void)\n{{\n',
+                *decs,
+                *body,
+                '}\n'])
```
yguclu, line 444: I am confused by the unpacking of `decs` and `body`, which are strings. Can you please explain what is happening here?

EmilyBourne: The result is the same as it splits it into characters and then joins then together again. It's just slower

pyccel/codegen/wrapper/c_to_python_wrapper.py
```diff
@@ -489,7 +496,42 @@ def _save_referenced_objects(self, func, func_args):
                                           [Return([self._error_exit_code])]))])
         return body
 
-    def _build_module_exec_function(self, expr):
+    def _add_object_to_mod(self, module_var, obj, name, initialised):
+        """
+        Get code for adding an object to the module.
+
+        This function creates the Ast nodes necessary to add an object to
```
yguclu, line 503: Suggestion: Ast -> AST

pyccel/codegen/wrapper/c_to_python_wrapper.py
```diff
@@ -502,64 +544,156 @@ def _build_module_exec_function(self, expr):
...
+        for i,c in enumerate(expr.classes):
+            wrapped_class = self._python_object_map[c]
+            type_object = wrapped_class.type_object
+
+            API_elem = IndexedElement(API_var, i)
+            body.append(AliasAssign(API_elem, PointerCast(ObjectAddress(type_object), API_elem)))
+
+        ok_code = LiteralInteger(0)
+
+        # Save Capsule describing types (needed for dependant modules)
```
yguclu, line 591: Typo? dependant -> dependent

pyccel/codegen/wrapper/c_to_python_wrapper.py
EmilyBourne: Fixed 13f4093f

yguclu (approved):
Looks good!

PR #1709, opened 2024-02-02
List support: append method

### LINKED ISSUE #1689: append: used for adding elements to the end of the list.
### Feature Description
Add support for append() method for lists in the Python translation.

### Test Code
```python
a = [1, 2]
a.append(1)
```

### Proposed Solution
See description in #1529

Farouk-Echaref:
Add support for the append() method of Python lists to the semantic stage. Add handling of that method to the Python printer. This fixes #1689.

**Commit Summary**
- Add a class representation for the `append()` method, inheriting from `PyccelInternalFunction`. The constructor ensures compatibility between the datatypes of the `append()` argument and the list for homogenous lists.
- Create `ClassDef` representing a `NativeHomogeneousList`. It includes a method `append()` implemented by the `ListAppend` class.`
- Update condition in `get_cls_base()` function to properly check on the classtype of the containers.
- Add the python printer `_print_ListAppend()` to generate the appropriate python representation of the `append()` method.
- Add check on the semantic to ensure that the correct class definition(`PythonList`) is assigned to variables created via expressions such as `a = [1]`
- Add tests for the `append()` method involving many scenarios where the function could be used.


mustapha-belbiad (approved):
everything looks good to me

pyccel/ast/class_defs.py
```diff
@@ -5,11 +5,15 @@
 """
 This module contains all types which define a python class which is automatically recognised by pyccel
 """
+
+from pyccel.ast.builtin_objects.list_functions import (ListAppend)
```
EmilyBourne, line 9:
```suggestion
from pyccel.ast.builtin_objects.list_functions import ListAppend
```
These parentheses do nothing

pyccel/codegen/printing/pycode.py
```diff
@@ -845,6 +845,13 @@ def _print_NumpyCountNonZero(self, expr):
 
         return "{}({})".format(name, arg)
 
+    def _print_ListAppend(self, expr):
+        method_name = expr.name
+        dotted_var = self._print(expr.args[0])
+        method_args = self._print(expr.args[1])
```
EmilyBourne, line 851: It would be nice to give `expr.args[0]` and `expr.args[1]` more explicit names. This can be done using the `@property` decorator with class methods

pyccel/parser/semantic.py
```diff
@@ -679,7 +682,6 @@ def _infer_type(self, expr):
             return d_var
 
         elif isinstance(expr, TypedAstNode):
-
```
EmilyBourne, line 682: Unnecessary change

EmilyBourne (changes requested):
This looks like a good start but I see a few major flaws. Please also add some tests in the `tests/errors` folder to ensure that neat errors are correctly raised when `append` is misused

pyccel/codegen/printing/pycode.py
```diff
@@ -845,6 +845,13 @@ def _print_NumpyCountNonZero(self, expr):
...
+
+        return f"{dotted_var}.{method_name}({method_args})\n"
```
EmilyBourne, line 853: Are you sure that the `\n` is necessary? Have you checked your code to ensure that this is not being added after? Generally functions don't usually add `\n` themselves but I don't remember if functions which don't return results are an exception to this?

.gitignore
```diff
@@ -67,4 +67,4 @@ tests/examples/openmp/*.f90
 tests/examples/openmp/*.mod
 tests/examples/openmp/*.o
 
-benchmarks/tmp/*
+benchmarks/tmp/*
```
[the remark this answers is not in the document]
Farouk-Echaref: Sorry i may have deleted an additional space when i modified the file, will solve it in the next commit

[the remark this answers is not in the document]
EmilyBourne: :+1: You can revert changes to a file with `git checkout devel -- .gitignore`

pyccel/ast/builtin_objects/list_functions.py
```diff
@@ -0,0 +1,47 @@
...
+    Parameters
+    ----------
+    *args : iterable
+        The arguments passed to the function call.
+    """
+    __slots__ = ()
+    _dtype = NativeVoid()
+    name = 'append'
+
+    def __init__(self, *args) -> None:
```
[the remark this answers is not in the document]
Farouk-Echaref: When I checked the arguments, I found that, in addition to the `X` that gets passed to `list.append`, the dotted variable (in this case, `a.append(2)`, is `a`) is also returned with the args, so I figured if I can have access to that variable, I can retrieve its type and use it to run the check
args

pyccel/ast/builtin_objects/list_functions.py
```diff
@@ -0,0 +1,47 @@
+# coding: utf-8
+#------------------------------------------------------------------------------------------#
+# This file is part of Pyccel which is released under MIT License. See the LICENSE file or #
+# go to https://github.com/pyccel/pyccel/blob/master/LICENSE for full license details.     #
+#------------------------------------------------------------------------------------------#
+"""
+The List container has a number of built-in methods that are 
+always available.
+
+In this module we implement List methods.
```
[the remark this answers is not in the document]
Farouk-Echaref: will update the doc asap

.gitignore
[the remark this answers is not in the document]
Farouk-Echaref: yes i did exactly that, will push that commit when i update the rest of the changes

pyccel/ast/builtin_objects/list_functions.py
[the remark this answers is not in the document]
EmilyBourne: I've just had another look and it seems that the `x` is positional only so its name is not important, however your prototype says that any of the following statements is valid:
```python
a = [1,2,3]
a.append()
a.append(4)
a.append(5,6)
a.append(7,8,9,10,11,12)
```
This is not true. It is general easier to understand the code if it is as explicit as possible. In this case what is the advantage of writing:
```python
def __init__(self, *args) -> None:
```
over:
```python
def __init__(self, lst_bound_arg, new_elem) -> None:
```
?

pyccel/ast/class_defs.py
Farouk-Echaref: For now, yes, but importing additional classes will require putting them in parentheses for convenience

EmilyBourne: Then they can be added when they are necessary. Personally I prefer using multiple lines over grouping inside a tuple. I have never looked into which is faster though. As this is left to personal preference inside the code we should leave it to the person who needs more lines to decide. Besides, as written there is nothing stopping the next person from doing:
```python
from pyccel.ast.builtin_objects.list_functions import (ListAppend), ListPop
```

pyccel/ast/builtin_objects/list_functions.py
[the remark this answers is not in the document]
Farouk-Echaref: you're right, i run those tests in the debugger and i got the clearer idea on how the prototype should be. should i check on the `len()` instead?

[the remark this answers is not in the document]
EmilyBourne: Why would:
```python
def __init__(self, *args) -> None:
    assert len(args) == 2
```
be better than:
```python
def __init__(self, lst_bound_arg, new_elem) -> None:
```
?

pyccel/ast/builtin_objects/list_functions.py
```diff
@@ -36,12 +36,27 @@ class ListAppend(PyccelInternalFunction):
...
+            lst_bound_arg.dtype == new_elem.dtype and
+            list_precision == get_final_precision(new_elem) and
+            lst_bound_arg.rank - 1 == new_elem.rank and
+            (getattr(lst_bound_arg, 'order', None) == getattr(new_elem, 'order',None))
+        ]
+        is_homogeneous = lst_bound_arg.dtype is not NativeGeneric() and all(conditions)
+        if is_homogeneous:
+            inner_shape = [() if new_elem.rank == 0 else new_elem.shape]
+            new_list_shape = (lst_bound_arg.shape,) + inner_shape[0]
+            lst_bound_arg.set_init_shape(new_list_shape)
```
EmilyBourne, line 60: The "init shape" is the shape at the initialisation. Any append statement happens after the initialisation so it doesn't make sense to modify this value

pyccel/ast/builtin_objects/list_functions.py
```diff
@@ -36,12 +36,27 @@ class ListAppend(PyccelInternalFunction):
...
+    def __init__(self, lst_bound_arg, new_elem) -> None:
+        super().__init__(new_elem)
+        list_precision = get_final_precision(lst_bound_arg)
+        conditions = [
+            new_elem.dtype is not NativeGeneric() and
+            lst_bound_arg.dtype == new_elem.dtype and
+            list_precision == get_final_precision(new_elem) and
+            lst_bound_arg.rank - 1 == new_elem.rank and
+            (getattr(lst_bound_arg, 'order', None) == getattr(new_elem, 'order',None))
+        ]
```
EmilyBourne, line 55: This is a list containing 1 boolean. If you use () then you can avoid the use of `all` in the line below.
```suggestion
        conditions = (
            new_elem.dtype is not NativeGeneric() and
            lst_bound_arg.dtype == new_elem.dtype and
            list_precision == get_final_precision(new_elem) and
            lst_bound_arg.rank - 1 == new_elem.rank and
            (getattr(lst_bound_arg, 'order', None) == getattr(new_elem, 'order',None))
        )
```

pyccel/codegen/printing/pycode.py
Farouk-Echaref: yes i did check and somehow `\n` was not added after, in case of multiple `append()` calls i was getting this:
`a.append()a.append()`
ill check again tho after i fix the class

EmilyBourne: Ok. No need to check again. I actually saw this code yesterday and `\n` is added automatically if the result is not assigned so it should be here for this function :+1:

pyccel/parser/semantic.py
```diff
@@ -1108,10 +1111,13 @@ def _handle_function(self, expr, func, args, is_method = False):
...
             try:
                 new_expr = func(*args, **kwargs)
-            except TypeError:
-                errors.report(UNRECOGNISED_FUNCTION_CALL,
-                        symbol = expr,
-                        severity = 'fatal')
+            except TypeError as e:
+                message = str(e)
+                if not message:
+                    message = UNRECOGNISED_FUNCTION_CALL
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
The final line is the only one without coverage. It was there before and is a fallback for strange function calls.

pyccel/ast/builtin_objects/list_functions.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def __init__(self, lst_bound_arg, new_elem) -> None:
+        list_precision = get_final_precision(lst_bound_arg)
+        check = (getattr(lst_bound_arg, 'order', None) == getattr(new_elem, 'order', None))
+        conditions = (
+            new_elem.dtype is not NativeGeneric() and
+            lst_bound_arg.dtype == new_elem.dtype and
+            list_precision == get_final_precision(new_elem) and
+            lst_bound_arg.rank - 1 == new_elem.rank and
+            (check if check else (lst_bound_arg.class_type == new_elem.class_type))
```
EmilyBourne, line 55: Why do you only check if the class types match if the order is different?

pyccel/ast/builtin_objects/list_functions.py
```diff
@@ -0,0 +1,80 @@
...
+    _class_type = NativeHomogeneousList()
+    name = 'append'
+
+    def __init__(self, lst_bound_arg, new_elem) -> None:
+        list_precision = get_final_precision(lst_bound_arg)
+        check = (getattr(lst_bound_arg, 'order', None) == getattr(new_elem, 'order', None))
+        conditions = (
+            new_elem.dtype is not NativeGeneric() and
+            lst_bound_arg.dtype == new_elem.dtype and
+            list_precision == get_final_precision(new_elem) and
```
EmilyBourne, line 53: I imagine that you are using `get_final_precision` to match what is done in `builtins.py`. There we have been a little generous. Probably overly generous I suspect it would be better removed.

By using `get_final_precision` we allow the insertion of objects which would be homogeneous in C/Fortran. So you can then do:
```python
a = [1,2,3]
a.append(np.int64(4))
```
(or in Windows where the default precision is smaller
```python
a = [1,2,3]
a.append(np.int32(4))
```
)

pyccel/ast/builtin_objects/list_functions.py
Farouk-Echaref: it was probably because of this case :
```python
a = [[1, 2, 3]]
a.append([1, 2])
a.append((2, 3))
```
in both append methods, the conditions for the homogeneity without the class_type checking would be true(both have `order: None`), so there was no other way to differentiate between the tuple as argument or the list as argument

pyccel/ast/builtin_objects/list_functions.py
Farouk-Echaref: i see, what can i replace it with to check for the precision ?

EmilyBourne: `lst_bound_arg.precision == new_elem.precision`

pyccel/ast/builtin_objects/list_functions.py
EmilyBourne: But isn't `check` true here? `None == None`. I'm not sure why the order and class type should be connected in any way

Farouk-Echaref: The order of `a` a 2D list is 'C' ,the order of a 1D list/tuple is None, and they all have the same `dtype`, `rank`, precision, that's why i needed to check the class type,if not, appending a tuple to a list of lists would be possible

pyccel/ast/builtin_objects/list_functions.py
```diff
@@ -0,0 +1,79 @@
...
+    def __init__(self, lst_bound_arg, new_elem) -> None:
+        check = (getattr(lst_bound_arg, 'order', None) == getattr(new_elem, 'order', None))
+        conditions = (
+            new_elem.dtype is not NativeGeneric() and
+            lst_bound_arg.dtype == new_elem.dtype and
+            lst_bound_arg.precision == new_elem.precision and
+            lst_bound_arg.rank - 1 == new_elem.rank and
+            (check if check else (lst_bound_arg.class_type == new_elem.class_type))
+        )
+        is_homogeneous = lst_bound_arg.dtype is not NativeGeneric() and conditions
```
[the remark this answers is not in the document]
Farouk-Echaref: /bot accept
those are essential lines i don't get why they are flagged like they weren't tested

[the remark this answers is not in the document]
EmilyBourne: "I don't believe you" is not a good reason to ignore the bot. It is better to work out why this is not called or what is going wrong with the bot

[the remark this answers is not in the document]
Farouk-Echaref: when the bot flags code with "isn't tested" does it mean the unit tests don't execute that part of the code?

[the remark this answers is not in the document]
EmilyBourne: Yes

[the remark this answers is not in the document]
EmilyBourne: I'm not sure how that happened here. Maybe it somehow got the wrong artifact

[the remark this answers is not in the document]
Farouk-Echaref: Im pretty sure every `append()` test executes this code, and it's not like the tests aren't executed themselves cause i checked the log of the Windows tests and they ran just fine

pyccel/ast/builtin_objects/list_functions.py
EmilyBourne: My question isn't so much why you check the class type but why you **only** do this under certain circumstances? For example how do you think this code would behave if I try to append a cuda array to a list of numpy arrays?

pyccel/ast/builtin_objects/list_functions.py
```diff
@@ -0,0 +1,79 @@
...
+        self._lst_bound_arg = lst_bound_arg.name
+        self._append_arg = new_elem
+        super().__init__()
+
+    @property
+    def list_variable(self):
+        """
+        Get the variable name representing the list.
+
+        Get the variable name representing the list.
```
EmilyBourne, line 68: According to this docstring the method returns a name (a string). Is this correct? I would expect it to return a `Variable`.

pyccel/parser/semantic.py
```diff
@@ -739,7 +739,10 @@ def _infer_type(self, expr):
             d_var['cls_base'       ] = TupleClass
             d_var['memory_handling'] = 'heap'
             return d_var
-
+        elif isinstance(expr, PythonList):
+            d_var['cls_base'       ] = ListClass
+            d_var['memory_handling'] = 'heap'
+            return d_var
```
EmilyBourne, line 745: Is this condition necessary? Can you explain why this is not handled by the condition below `elif isinstance(expr, TypedAstNode):`? (Maybe `PythonTuple` caused confusion? The condition `elif isinstance(expr, PythonList):` is removed in one of my unfinished PRs as it is no longer necessary)

EmilyBourne (changes requested):
Looks good, just a few final comments

pyccel/ast/builtin_objects/list_functions.py
Farouk-Echaref: I don't really know the expected behavior with cuda arrays, but my understanding, going into implementing this method, is to try as much as possible to notice any pattern for homogeneity. In my case, it's related to the information that is accessible in my `ListAppend`: dtype, rank, precision, and order. However, sometimes, this information is not enough to check for homogeneity in the case of embedded containers. Although I realize this is a bit superficial, it's necessary to implement a mechanism to check the class types of the incoming container and compare them with the class types of the list.

pyccel/ast/builtin_objects/list_functions.py
Farouk-Echaref: yes that's correct, it should return the name of the list which is a string, i'll update the method name

pyccel/ast/builtin_objects/list_functions.py
EmilyBourne: The future cuda arrays will be very similar to numpy arrays. The only property in `TypedAstNode` which will be different is `class_type`

EmilyBourne: Ahhhh, ok I understand. Does that mean that your implementation doesn't handle this:
```python
a = [(1, 2), (3,4)]
a.append((2, 3))
```
?

pyccel/parser/semantic.py
Farouk-Echaref: i guess you're right, it's not needed

pyccel/ast/builtin_objects/list_functions.py
Farouk-Echaref: If it's going to be homogeneous in C/Fortran, then that check would no longer be needed, just unnecessary complexity

EmilyBourne: I agree, that check would in fact prevent us from going the homogeneous in C/Fortran route. But I think we are better checking like Python. This matches what you have written in your tests and it will make life easier for @smazouz42 and @bauom as I suspect they will need to use `class_type` to differentiate between memory on CPU and memory on GPU

Farouk-Echaref: I will have to check with them when they need such a thing; until then, the current implementation will be okay. We can always change the checks based on the future features

EmilyBourne: I disagree. It is much easier to change this when designing the implementation than to go back through and fix it.

Also I still don't understand the logic of this code.

Also
> until then, the current implementation will be okay.

So you can handle this code:
```python
a = [(1, 2), (3,4)]
a.append((2, 3))
```
?

pyccel/ast/builtin_objects/list_functions.py
```diff
@@ -0,0 +1,77 @@
...
+        conditions = (
+            new_elem.dtype is not NativeGeneric() and
+            lst_bound_arg.dtype == new_elem.dtype and
+            lst_bound_arg.precision == new_elem.precision and
+            lst_bound_arg.rank - 1 == new_elem.rank
+        )
+        is_homogeneous = lst_bound_arg.dtype is not NativeGeneric() and conditions
+        if not is_homogeneous:
+            raise TypeError("Expecting an argument of the same type as the elements of the list")
+        self._lst_bound_arg = lst_bound_arg.name
```
EmilyBourne, line 57: There seems to be a problem here. Your docstring says that you take a `TypedAstNode` but that type doesn't have a `name` property. How can you be sure that you can access this property?

pyccel/ast/builtin_objects/list_functions.py
EmilyBourne: That sounds like a problem. That means you can't append to a list stored in a class right?

pyccel/ast/builtin_objects/list_functions.py
Farouk-Echaref: With the deleted check, yes, I can. I also discovered that this logic doesn't allow for the following
```python
import numpy as np
a = [np.array([1, 2, 3], order='C'), np.array([1, 2, 4], order='C')]
a.append(np.array([1, 2,], order='C'))
```
As I have no access to the order, so that logic is not right

EmilyBourne: Ok. It looks like you aren't checking the order anymore though. The following is inhomogeneous but not handled by your code:
```python
import numpy as np
a = [np.array([1, 2, 3], order='C'), np.array([1, 2, 4], order='C')]
a.append(np.array([1, 2,], order='F'))
```

pyccel/ast/builtin_objects/list_functions.py
Farouk-Echaref: there's two cases that i have checked:
```python
a = [4]
b = 3
a.append(b)
a.append(5)
```
`name` property in the first statement is `b`,while in the second statement it would be something like `dummy_000`
in which case do you think the passed variable wouldn't have a `name` attribute?

EmilyBourne: Interesting, so `lst_bound_arg` doesn't represent the list on which we are operating? I expected `name` to be `a` not `b`.
If the name is `dummy_000` then the semantic stage must be creating a temporary variable to store the argument. I didn't think it did this for literals. However if it does then that is fine, but you are not handling `TypedAstNode`, you are handling `Variable` so your docstring needs correcting

pyccel/ast/builtin_objects/list_functions.py
Farouk-Echaref: yes that's my issue that i described in the previous comment:
i don't have access to the order of the passed argument
order

pyccel/ast/builtin_objects/list_functions.py
Farouk-Echaref: i didn't quite understand what you meant by "**That means you can't append to a list stored in a class**", can you elaborate more please ?

EmilyBourne: The name of a class property depends on the language. For example:
```python
class A:
    def __init__(self, elem : int):
        self.my_list : list[int] = []
        self.my_list.append(elem)
```
Inside Pyccel `lst_bound_arg.name` will call `DottedVariable.name` this returns `my_list`. But when printing we want `self.my_list`. You can get this with `str(lst_bound_arg)` but this will still cause problems. The C translation would be something like:
```C
void A__init__(A self, int64_t elem):
    create(self.my_list)
    append(self.my_list, elem)
```
But the Fortran translation would be something like:
```fortran
subroutine A__create(self, elem)
    type(A), intent(inout) :: self
    integer(I64), intent(inout) :: elem

    create(self % my_list)
    append(self % my_list, elem)
end subroutine A__create
```
This is why the AST should never take care of deciding how something should be printed

pyccel/ast/builtin_objects/list_functions.py
Farouk-Echaref: this case as well:
```python
a = [[1, 2, 3]]
a.append([4, 5])
```
`a` has an `order` of `C`, and the passed argument has an `order` of `None`

pyccel/ast/builtin_objects/list_functions.py
EmilyBourne: I have managed to think of an example that should break:
```python
class A:
    def __init__(self):
         self.my_list : list[int] = []

    def get_list(self):
        return self.my_list
        
if __name__ == '__main__':
    my_a = A()
    my_a.get_list().append(4)
```

pyccel/ast/builtin_objects/list_functions.py
EmilyBourne: Ah crap, I think this may be why I ended up not implementing `np.ndarray[int]` as a datatype instead of `np.ndarray`. Do you think you have enough infrastructure to handle everything you need for now? If you can't work out the order of an element of a list then I think we need to disallow saving `np.ndarray` with rank>1 in lists.

My impression is that it is possible to get a working implementation with certain restrictions at the moment. But if we want the most flexible method possible then the datatypes need an overhaul. There should be 2 types of datatype:
- base types : these should hold enough information to describe the datatype in C, so
    - datatype
    - precision
- container types : these should hold enough information to describe the datatype in C, so
    - rank
    - order
    - base type

The `TypedAstNode` would then just save a container type. I can create an issue for this. I agree that it is out of scope for your PR

pyccel/ast/builtin_objects/list_functions.py
Farouk-Echaref: yes i see your point, do you have any suggestions to fix this?

pyccel/ast/builtin_objects/list_functions.py
Farouk-Echaref: Can't we just base the `order` on the targeted language? Meaning, even if the user specified the `order` as 'C' and wanted to convert to Fortran, the `order` would still be 'F'. Similarly, if someone wanted to convert to 'C' and specified the `order` as 'F', the order would still be 'C'.Or will there be a difference, memory wise, in C and Fortran if the user specified otherwise?

EmilyBourne: > Can't we just base the order on the targeted language? 

No. The order is unrelated to the target language. See : https://github.com/pyccel/pyccel/blob/devel/developer_docs/order_docs.md

> Or will there be a difference, memory wise, in C and Fortran if the user specified otherwise?

Yes there will be a memory difference

pyccel/ast/builtin_objects/list_functions.py
EmilyBourne: Don't save the name, just save a `TypedAstNode` and let the printer decide how to print it

pyccel/parser/semantic.py
EmilyBourne: Fixed ad6fca5eb78d573c514430038008ad187d865733

pyccel/ast/builtin_objects/list_functions.py
Farouk-Echaref: fixed

pyccel/ast/builtin_objects/list_functions.py
EmilyBourne: I think we should open an issue for this and leave it as it is for now.

EmilyBourne: Nothing. I think raising the error is out of scope for this issue and rewriting the TypedAstNode definitely is. I will open 2 issues. One easy one that should be done sooner rather than later to raise an error and a much larger one to clean up the typing. This will be done later if we think it is useful

EmilyBourne: See #1721 and #1722

Farouk-Echaref: okey, so this PR should wait until the issue related to raising an error is fixed?

EmilyBourne: No. It just shouldn't be put in a release until the error is raised. It can be marked as ready

pyccel/parser/semantic.py
```diff
@@ -58,7 +58,7 @@
 from pyccel.ast.core import PyccelFunctionDef
 from pyccel.ast.core import Assert
 
-from pyccel.ast.class_defs import NumpyArrayClass, TupleClass, get_cls_base
+from pyccel.ast.class_defs import NumpyArrayClass, TupleClass,get_cls_base
```
EmilyBourne, line 61: Useless change

EmilyBourne (changes requested):
Looks good to me. Just one minor comment

pyccel/ast/class_defs.py
```diff
@@ -130,6 +135,14 @@
 
 #=======================================================================================
 
+ListClass = ClassDef('list', class_type = NativeHomogeneousList(),
+        methods=[
+            PyccelFunctionDef('append', func_class = ListAppend,
+                decorators = {'property': 'property'}),
```
EmilyBourne, line 141: Sorry, I was just talking about the property decorator and remembered this was here. Why is this needed? Doesn't this cause a TypeError if you try to translate
```python
a = [1,2,3]
a.append
```

EmilyBourne (changes requested):

pyccel/ast/class_defs.py
Farouk-Echaref: I specified it cs the `ListAppend` class uses it 
I didn't quite understand what you meant with this example ?

EmilyBourne: Can you point me at where the `ListAppend` class uses the `property` decorator please. 

The property decorator (https://www.geeksforgeeks.org/python-property-decorator-property/) is used to allow a function to be called without passing brackets. So normally indicating that the `PyccelFunctionDef` has the `property` decorator should be the same as saying that you are allowed to call the function with the following syntax:
```
a = [1,2,3]
a.append
```
Without the decorator the code above will raise a neat error saying that the property is unknown. With the decorator it will try to call `ListAppend` passing only the list as an argument

Farouk-Echaref: Ohh my bad, i thought it was about the decorator that's used for attributes, Python's beginners' mistake.

pyccel/ast/builtin_objects/list_functions.py
```diff
@@ -0,0 +1,77 @@
...
+
+    Parameters
+    ----------
+    list_variable : Variable
+        The variable representing the list.
+    
+    new_elem : Variable
+        The argument passed to append() method.
+    """
+    __slots__ = ("_list_variable", "_append_arg")
```
EmilyBourne, line 38: See the [docs](https://github.com/pyccel/pyccel/blob/devel/developer_docs/ast_nodes.md#pyccel-internal-function):

> if the arguments are saved inside the sub-class then the `_attribute_nodes` static class attribute must be correctly set to ensure the tree is correctly constructed. As `PyccelInternalFunction` implements `_attribute_nodes` these classes are some of the only ones which will not raise an error during the pull request tests if this information is missing.

This PR does not set `_attribute_nodes`

Farouk-Echaref: Missed that part in the docs, fixed it

EmilyBourne (approved):
Looks good to me

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ All notable changes to this project will be documented in this file.
 -   #1605 : Add support for methods and interfaces in classes (including `__init__` and `__del__`).
 -   #1618 : Add support for class instance attributes.
 -   #1680 : Add support for `typing.Final`.
+-   #1529 : Add Python support for list method `append()`
```
EmilyBourne, line 28: Please move this to the new UNRELEASED section

Farouk-Echaref: Done

pyccel/ast/builtin_objects/list_functions.py
yguclu, line 1: @EmilyBourne I feel that the name of this folder is not fully adequate. Why not calling it just 
`builtins`?

yguclu: Sorry, we already have a `builtins.py` file so that would cause problems. Anyway the "object" part of the name sounds confusing to me.

EmilyBourne: As you say we already have `builtins.py` so I was looking for another name. This folder holds all the classes describing methods associated with a builtin datatype/object (e.g. lists, tuples, sets, etc). Do you have a better suggestion?

yguclu: If this folder should contain the methods of the builtin classes, how about `builtin_methods`?

EmilyBourne: That makes sense :+1:

pyccel/ast/builtin_objects/list_functions.py
```diff
@@ -0,0 +1,78 @@
...
+    _precision = -1
+    _class_type = NativeHomogeneousList()
+    name = 'append'
+
+    def __init__(self, list_variable, new_elem) -> None:
+        conditions = (
+            new_elem.dtype is not NativeGeneric() and
+            list_variable.dtype == new_elem.dtype and
+            list_variable.precision == new_elem.precision and
+            list_variable.rank - 1 == new_elem.rank
```
yguclu, line 53: How is the `rank` going to work for a list of objects of a user-defined class? Is it going to be 1 for `list_variable` and 0 for `new_elem`?

pyccel/ast/builtin_objects/list_functions.py
```diff
@@ -0,0 +1,78 @@
...
+        )
+        is_homogeneous = list_variable.dtype is not NativeGeneric() and conditions
```
yguclu, line 55: Why do we need the `conditions` boolean variable? Could we not define `is_homogeneous` directly?
```suggestion
        is_homogeneous = (
            new_elem.dtype is not NativeGeneric() and
            list_variable.dtype is not NativeGeneric() and
            list_variable.dtype == new_elem.dtype and
            list_variable.precision == new_elem.precision and
            list_variable.rank - 1 == new_elem.rank
        )
```

pyccel/ast/builtin_objects/list_functions.py
EmilyBourne: Yes. In the future I imagine it could be interesting to allow class with a rank (maybe this would be calculated using `__getitem__`) however that would require several improvements to Pyccel and is out of scope for this PR. Currently in the code all class instances have rank=0.

pyccel/ast/builtin_objects/list_functions.py
yguclu: If you agree I would also change the module name accordingly: list_functions -> list_methods

pyccel/parser/semantic.py
```diff
@@ -1094,7 +1094,7 @@ def _handle_function(self, expr, func, args, is_method = False):
         func : FunctionDef instance, Interface instance or PyccelInternalFunction type
                The function being called.
 
-        args : tuple
+        args : list
                The arguments passed to the function.
```
yguclu, line 1098: Do we care if `args` is a list or a tuple? I suppose it just needs to be iterable, as we unpack it below.

pyccel/ast/builtin_objects/list_functions.py
Farouk-Echaref: Yes that is the case

tests/errors/semantic/blocking/APPEND_LIST_INCOMPATIBLE_TYPES_3.py
yguclu, line 1: This file is identical to `APPEND_LIST_INCOMPATIBLE_TYPES_1.py` (excluding the whitespaces).

pyccel/ast/builtin_objects/list_functions.py
Farouk-Echaref: should I rename it and commit it in my PR?

tests/epyccel/test_epyccel_lists.py
```diff
@@ -0,0 +1,127 @@
...
+@pytest.mark.parametrize( 'language', [
+        pytest.param("c", marks = [
+            pytest.mark.skip(reason="append() not implemented in c"),
+            pytest.mark.c]),
+        pytest.param("fortran", marks = [
+            pytest.mark.skip(reason="append() not implemented in fortran"),
+            pytest.mark.fortran]),
+        pytest.param("python", marks = pytest.mark.python)
+    ]
+)
```
yguclu, line 18: Could we not define a `language` fixture in this file, rather than copying this long decorator many times?

yguclu (changes requested):

pyccel/ast/builtin_objects/list_functions.py
yguclu: Yes please

pyccel/ast/builtin_objects/list_functions.py
yguclu: Sounds reasonable. Thanks

tests/epyccel/test_epyccel_lists.py
EmilyBourne: Is it possible to put xfail marks in a fixture? The xfail marks should be removed in a few weeks or months once the method is supported

pyccel/ast/builtin_objects/list_functions.py
Farouk-Echaref: Yes it's possible, I'll refactor it in my next commit

pyccel/parser/semantic.py
Farouk-Echaref: when i was executing the code inside this method, i checked the type of `args` and it was `list` so i thought I'd update the doc, but iterable makes as you mentioned the unpacking

tests/epyccel/test_epyccel_lists.py
Farouk-Echaref: could you please explain how that would be possible? I was unable to do so since I'm not that familiar with writing test files using pytest so I did in the most basic way

yguclu: > Is it possible to put xfail marks in a fixture? The xfail marks should be removed in a few weeks or months once the method is supported

It should be possible:
```python
@pytest.fixture( params=[
            pytest.mark.skip(reason="list methods not implemented in fortran"),
            pytest.mark.fortran]),
        pytest.param("c", marks = [
            pytest.mark.skip(reason="list methods not implemented in c"),
            pytest.mark.c]),
    ],
    scope = "module"
)
def language(request):
    return request.param
```

Farouk-Echaref: can you check the refactored version please

yguclu: Looks good! Have you checked that it actually works? 😉

pyccel/ast/builtin_methods/list_methods.py
```diff
@@ -0,0 +1,78 @@
...
+    Parameters
+    ----------
+    list_variable : Variable
+        The variable representing the list.
+    
+    new_elem : Variable
+        The argument passed to append() method.
+    """
+    __slots__ = ("_list_variable", "_append_arg")
+    _attribute_nodes = ("_list_variable", "_append_arg")
```
yguclu, line 39: Do we really need this additional storage? The base class already stores `_args = (list_variable, new_elem)`

yguclu (changes requested):

tests/epyccel/test_epyccel_lists.py
Farouk-Echaref: It did work in my linux machine 🤞

pyccel/ast/builtin_methods/list_methods.py
Farouk-Echaref: I added `_attribute_nodes` based on this part of the doc:
> if the arguments are saved inside the sub-class then the `_attribute_nodes` static class attribute must be correctly set to ensure the tree is correctly constructed. As `PyccelInternalFunction` implements `_attribute_nodes` these classes are some of the only ones which will not raise an error during the pull request tests if this information is missing.
and used `__slots__` based on the **development_conventions** docs.
Can you please elaborate on how the base class stores `_args` if we are not passing anything to `super().__init()`

tests/epyccel/test_epyccel_lists.py
EmilyBourne: Looks good but I wonder if we should name this fixture something like `language_python_ok`. At some point we will have some tests in this file which can use the existing `language` fixture (without the xfails) and some which need this one (and maybe some that need one where only C or only Fortran fails)

pyccel/ast/builtin_methods/list_methods.py
EmilyBourne: The base class only stores these objects if they are passed to the constructor. Generally this is not super useful. It can be a good idea if the superclass is used to help with printing

tests/epyccel/test_epyccel_lists.py
yguclu: Good point. How about using multiple test files and move the tests from one to the other as they start passing?

pyccel/ast/builtin_methods/list_methods.py
yguclu: All right!

tests/errors/semantic/blocking/APPEND_LIST_INCOMPATIBLE_TYPES_3.py
yguclu: @Farouk-Echaref please mention if and where my comment was addressed.

pyccel/parser/semantic.py
yguclu: @Farouk-Echaref please mention if and where my comment was addressed.

pyccel/ast/builtin_objects/list_functions.py
yguclu: @Farouk-Echaref please mention if and where my comment was addressed.

pyccel/ast/builtin_objects/list_functions.py
yguclu: @Farouk-Echaref please mention if and where my comment was addressed.

Farouk-Echaref: @yguclu Sorry, should i resolve this conversation ?

pyccel/ast/builtin_objects/list_functions.py
Farouk-Echaref: @yguclu fixed

pyccel/parser/semantic.py
Farouk-Echaref: @yguclu fixed

tests/errors/semantic/blocking/APPEND_LIST_INCOMPATIBLE_TYPES_3.py
Farouk-Echaref: @yguclu fixed

pyccel/ast/builtin_objects/list_functions.py
yguclu: Typically the PR author should write something like "solved in 0772a46", and let the reviewer resolve the conversation.

pyccel/ast/builtin_objects/list_functions.py
yguclu: Where?

Farouk-Echaref: commit `ce7ec14dd05f8adbe1f680325a175a173733837d`
commit

tests/epyccel/test_epyccel_lists.py
EmilyBourne: This is also possible if you have a suggestion for a good name? We may need up to 4 files:
- Passing in Python only
- Passing in C and Python
- Passing in Fortran and Python
- Passing in C, Fortran and Python

pyccel/ast/builtin_objects/list_functions.py
yguclu: Thanks! no need for the screenshot 😉

pyccel/ast/builtin_objects/list_functions.py
EmilyBourne: See the [docs](https://github.com/pyccel/pyccel/blob/devel/developer_docs/review_process.md):

:wink:

Farouk-Echaref: Understood, thanks for clarifying
Concerning the naming change, this was solved in [ce7ec14](https://github.com/pyccel/pyccel/pull/1727/commits/ce7ec14dd05f8adbe1f680325a175a173733837d)

pyccel/parser/semantic.py
Farouk-Echaref: solved in [ed6e77c](https://github.com/pyccel/pyccel/pull/1727/commits/ed6e77c4d083b01b3de93bbb8ce87f3a379e6f6d)

tests/errors/semantic/blocking/APPEND_LIST_INCOMPATIBLE_TYPES_3.py
Farouk-Echaref: solved in [c349a6e](https://github.com/pyccel/pyccel/pull/1727/commits/c349a6e288c26fa0a376e7f556d3c10a37db8007)

tests/epyccel/test_epyccel_lists.py
yguclu: Well, using four files is a bit cumbersome... We could call them
- `test_epyccel_lists.py`
- `test_epyccel_lists__no_c.py`
- `test_epyccel_lists__no_fortran.py`
- `test_epyccel_lists__python_only.py`

Another solution is using a single file, but providing a utility function which generates the correct decorators with minimum user input.

In any case I suggest that we tackle this in a future PR.

yguclu (approved):
Good job @Farouk-Echaref!

PR #1710, opened 2024-02-02
List support: pop method

### LINKED ISSUE #1690: pop: removes and returns the last value from the list or the given index value.
### Feature Description
Add support for pop() method for lists in the Python translation.

### Test Code
```python
>> a = [1, 3]
>> b = a.pop(3)
>> print(b)
>> 3
```

### Proposed Solution
See description in https://github.com/pyccel/pyccel/issues/1529

mustapha-belbiad:
Add support for the pop() method of Python lists to the semantic stage. Add handling of that method to the Python printer. This fixes #1690.


EmilyBourne:
Don't forget to update the AUTHORS file

mustapha-belbiad:
> Don't forget to update the AUTHORS file
alright

Farouk-Echaref (approved):
All looks good

.gitignore
```diff
@@ -68,3 +68,5 @@ tests/examples/openmp/*.mod
 tests/examples/openmp/*.o
 
 benchmarks/tmp/*
+
+my_main.py
```
EmilyBourne, line 72: I don't think it is a good idea to add files to this folder following the convention that each individual developer uses for their test files. I recommend running tests in a separate folder if you are worried about accidentally committing your tests.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ All notable changes to this project will be documented in this file.
 -   #1685 : Add support for `type()` function with class instance argument.
 -   #1605 : Add support for class methods and interfaces (including `__init__` and `__del__`).
 -   #1680 : Add support for `typing.Final`.
+-   #1690 : Add support for pop() method for
```
EmilyBourne, line 22: This sentence is not complete.

pyccel/ast/basic.py
```diff
@@ -66,7 +66,6 @@ def __init__(self):
         self._recursion_in_progress = False
         for c_name in self._my_attribute_nodes: #pylint: disable=not-an-iterable
             c = getattr(self, c_name)
-
```
EmilyBourne, line 69: This change does not seem relevant to this PR. Please try to remove irrelevant changes from PRs so that they are as small as possible. This makes them easier to review (this will be especially important in the future when your PRs are larger)

pyccel/ast/builtin_objects/list_functions.py
```diff
@@ -0,0 +1,39 @@
+# coding: utf-8
+#------------------------------------------------------------------------------------------#
+# This file is part of Pyccel which is released under MIT License. See the LICENSE file or #
+# go to https://github.com/pyccel/pyccel/blob/master/LICENSE for full license details.     #
+#------------------------------------------------------------------------------------------#
+"""
+The List container has a number of built-in methods that are 
+always available.
+
+In this module we implement List methods.
```
EmilyBourne, line 10: I don't think the implementation of list methods is found in this file. Rather it contains objects which describe these methods within Pyccel's AST

pyccel/ast/builtin_objects/list_functions.py
```diff
@@ -0,0 +1,39 @@
...
+"""
+
+from pyccel.ast.internals import PyccelInternalFunction
+from pyccel.ast.datatypes import NativeHomogeneousList
+
+class ListPop(PyccelInternalFunction) :
+    """
+    Represents a call to the .pop() method.
+    
+    Represents a call to the .pop() method wich
```
EmilyBourne, line 20:
```suggestion
    Represents a call to the .pop() method which 
```

pyccel/ast/builtin_objects/list_functions.py
```diff
@@ -0,0 +1,39 @@
...
+    removes the item at the specified index. 
+    The method also returns the removed item.
+    
+    Parameters
+    ----------
+    *args : iterable
+        The arguments passed to the function call.
+    """
+    __slots__ = ('_dtype',)
+    _precision = -1
```
EmilyBourne, line 30: How do you know the precision of the result?

pyccel/ast/builtin_objects/list_functions.py
```diff
@@ -0,0 +1,39 @@
...
+    _rank  = 0
+    _shape = None
+    _order = None
+    _class_type = NativeHomogeneousList()
+    name = 'pop'
+    def __init__(self, *args):
+        super().__init__(*args)
+        self._dtype = args[0].dtype
```
EmilyBourne, line 38: As described in the second paragraph [here](https://github.com/pyccel/pyccel/blob/devel/developer_docs/ast_nodes.md#pyccel-internal-function), objects inheriting from PyccelInternalFunction must provide a constructor whose positional and keyword arguments match the names of the positional and keyword arguments of the Python function being described.

According to [Python's docs](https://docs.python.org/3/tutorial/datastructures.html#more-on-lists) list.append only accepts 1 argument whose name may be `i`. Your function takes any number of arguments. What are these additional arguments supposed to represent? Further the argument to `pop` is optional however this case does not seem to be handled by your implementation. This isn't a problem in the Python implementation but it will be much simpler for the C/Fortran implementations if it is already defined here.

pyccel/ast/core.py
```diff
@@ -3383,9 +3381,7 @@ def __init__(
 
          #  methods = list(methods) + [free]
          # TODO move this somewhere else
-
         methods = tuple(methods)
-
```
EmilyBourne, line 3388: Lots of unnecessary changes in this file

pyccel/codegen/pipeline.py
```diff
@@ -169,7 +169,6 @@ def handle_error(stage):
 
     if conda_warnings not in ('off', 'basic', 'verbose'):
         raise ValueError("conda warnings accept {off, basic,verbose}")
-
```
EmilyBourne, line 172: Unnecessary change

pyccel/codegen/printing/ccode.py
```diff
@@ -514,7 +514,6 @@ def _init_stack_array(self, expr):
         tot_shape = self._print(functools.reduce(
             lambda x,y: PyccelMul(x,y,simplify=True), var.alloc_shape))
         declare_dtype = self.find_in_dtype_registry(NativeInteger(), 8)
-
```
EmilyBourne, line 517: Unnecessary change

pyccel/codegen/printing/pycode.py
```diff
@@ -249,7 +249,6 @@ def _print_DottedVariable(self, expr):
     def _print_FunctionDefArgument(self, expr):
         name = self._print(expr.name)
         default = ''
-
```
EmilyBourne, line 252: Unnecessary change. Why do you hate vertical whitespace ^^

pyccel/codegen/printing/pycode.py
```diff
@@ -1136,6 +1135,12 @@ def _print_TypingFinal(self, expr):
         annotation = self._print(expr.arg)
         return f'const {annotation}'
 
+    def _print_ListPop(self, expr):
+        args = ','.join(self._print(a) for a in expr.args[1:])
+        name = self._print(expr.args[0])
```
EmilyBourne, line 1140: It would be nice to give expr.args[0] a more explicit names. This can be done using the @property decorator with class methods. ` expr.args[1:]` could stay as is if you can tell me what all these arguments are for

pyccel/codegen/printing/pycode.py
```diff
@@ -1136,6 +1135,12 @@ def _print_TypingFinal(self, expr):
...
+        method = self._print(expr.name)
+        return f"{name}.{method}({args})"
```
EmilyBourne, line 1142: Do you foresee a case where this would be something other than:
```suggestion
        return f"{name}.pop({args})"
```
?
For example an object which inherits from `pop`? If not then I don't think that there is anything to be gained from using `self._print` which incurs a cost

pyccel/parser/semantic.py
```diff
@@ -298,7 +298,6 @@ def annotate(self):
 
         # TODO - add settings to Errors
         #      - filename
-
```
EmilyBourne, line 301: Lots of unnecessary changes in this file

pyccel/parser/syntactic.py
```diff
@@ -323,7 +323,6 @@ def _visit(self, stmt):
         # TODO - add settings to Errors
         #      - line and column
         #      - blocking errors
-
```
EmilyBourne, line 326: Unnecessary change

tests/epyccel/test_epyccel_lists.py
```diff
@@ -0,0 +1,80 @@
...
+    )
+)
+
+def test_pop_2(language) :
+    def po():
+        a = [1.7,2.7,45.0]
+        a.pop()
+        return a.pop(-1)
+    re = epyccel(po, language = language)
+    assert po() == re()
```
EmilyBourne, line 80: It would be good to add some tests to this file which check that the return type is as expected.

tests/pyccel/scripts/hope_benchmarks/test.json
```diff
@@ -0,0 +1,67 @@
...
+        "includes": [
+            "/usr/include/python3.11",
+            "/usr/local/lib/python3.11/dist-packages/numpy/core/include"
+        ],
+        "shared_suffix": ".cpython-311-x86_64-linux-gnu.so",
+        "dependencies": [
+            "/usr/lib/x86_64-linux-gnu/libpython3.11.so"
+        ]
+    }
+}
```
EmilyBourne, line 67: This file is generated by the tests. It should not be committed to the repo

tests/pyccel/scripts/test2.json
```diff
@@ -0,0 +1,67 @@
...
+        "includes": [
+            "/usr/include/python3.11",
+            "/usr/local/lib/python3.11/dist-packages/numpy/core/include"
+        ],
+        "shared_suffix": ".cpython-311-x86_64-linux-gnu.so",
+        "dependencies": [
+            "/usr/lib/x86_64-linux-gnu/libpython3.11.so"
+        ]
+    }
+}
```
EmilyBourne, line 67: This file is generated by the tests. It should not be committed to the repo

tests/semantic/test_semantic.py
EmilyBourne, line 1: What did you modify in this file?

tests/syntax/test_syntax.py
EmilyBourne, line 1: What did you modify in this file?

EmilyBourne (changes requested):
This looks like a good start but I see a few major flaws. There are also lots and lots of whitespace changes. Everyone had different opinions on what whitespace is more readable. We don't impose any conventions in the code, you shouldn't modify lots of unrelated code to impose your preferred conventions.

CHANGELOG.md
mustapha-belbiad: My apologies, I'm going to fix it

pyccel/codegen/printing/pycode.py
mustapha-belbiad: **expr.args[1:]** The purpose of this is to print multiple arguments.

EmilyBourne: Please give me an example where `pop` needs to handle multiple arguments

pyccel/ast/builtin_objects/list_functions.py
```diff
@@ -0,0 +1,62 @@
...
+    _order = None
+    _shape = None
+    _class_type = NativeHomogeneousList()
+    name = 'pop'
+    def __init__(self, name, index_elemnt=None):
+        if index_elemnt:
+            super().__init__(name, index_elemnt)
+        else:
+            super().__init__(name)
+        self._index = index_elemnt
```
EmilyBourne, line 43: Please fix the typo:
```suggestion
    def __init__(self, name, index_elemnt=None):
        if index_elemnt:
            super().__init__(name, index_element)
        else:
            super().__init__(name)
        self._index = index_element
```

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ All notable changes to this project will be documented in this file.
 -   #1685 : Add support for `type()` function with class instance argument.
 -   #1605 : Add support for class methods and interfaces (including `__init__` and `__del__`).
 -   #1680 : Add support for `typing.Final`.
+-   #1690 : Add support for pop() method of list class
```
EmilyBourne, line 22: Code should be quoted
```suggestion
-   #1690 : Add support for `pop()` method of list class
```

pyccel/ast/builtin_objects/list_functions.py
EmilyBourne: Not sure why this was resolved. The typo is still there

pyccel/ast/class_defs.py
```diff
@@ -5,12 +5,14 @@
 """
 This module contains all types which define a python class which is automatically recognised by pyccel
 """
+
+from pyccel.ast.builtin_objects.list_functions import (ListPop)
```
EmilyBourne, line 9:
```suggestion
from pyccel.ast.builtin_objects.list_functions import ListPop
```
These parentheses do nothing

pyccel/ast/builtin_objects/list_functions.py
```diff
@@ -0,0 +1,65 @@
...
+        self._dtype = name.dtype
+        self._precision = name.precision
+
+    @property
+    def args(self):
+        """
+        The current index value for the element to be popped.
+
+        The current index value for the element to be popped.
+        """
```
EmilyBourne, line 55: Can you explain why "args" is the best name to describe "The current index value for the element to be popped.". Why not something like `pop_index`?

The only information that I can get from seeing "args" in the code is that it contains more than 1 object (as args is plural), but this isn't true :/

pyccel/parser/semantic.py
```diff
@@ -740,6 +740,10 @@ def _infer_type(self, expr):
             d_var['memory_handling'] = 'heap'
             return d_var
 
+        elif isinstance(expr, PythonList):
+            d_var['cls_base'       ] = ListClass
+            d_var['memory_handling'] = 'heap'
+            return d_var
```
EmilyBourne, line 746: Is this condition necessary? Can you explain why this is not handled by the condition below `elif isinstance(expr, TypedAstNode):`? (Maybe `PythonTuple` caused confusion? The condition `elif isinstance(expr, PythonTuple):` is removed in one of my unfinished PRs as it is no longer necessary)

tests/epyccel/test_epyccel_lists.py
```diff
@@ -0,0 +1,84 @@
...
+)
+
+def test_pop_2(language) :
+    def po():
+        a = [1.7,2.7,45.0]
+        a.pop()
+        return a.pop(-1)
+    re = epyccel(po, language = language)
+    assert isinstance(po(), type(re()))
+    assert po() == re()
```
EmilyBourne, line 84: This file contains lots of repeat calls. This could be problematic if the function or its wrapper is costly. Why not save the results?

Additionally the names are not very explicit. What does `re` stand for?

```suggestion
def test_pop_2(language) :
    def pop_2():
        a = [1.7,2.7,45.0]
        a.pop()
        return a.pop(-1)
    pop_2_epyc = epyccel(po, language = language)
    python_result = pop_2()
    pyccel_result = pop_2_epyc()
    assert isinstance(python_result, type(pyccel_result))
    assert python_result == pyccel_result
```

EmilyBourne (changes requested):
This looks good but I still have a few comments

CHANGELOG.md
EmilyBourne: Fixed 540283d66e640f3f7dd9644ea25427981d31cfbe

pyccel/ast/class_defs.py
EmilyBourne: Fixed 8d649bfdb127979ea3223f0bfc80ba57ca5efacc

pyccel/ast/builtin_objects/list_functions.py
EmilyBourne: Fixed 61433ec7f31d1c86e7449b3b91482fae3adfc110

pyccel/parser/semantic.py
EmilyBourne: Fixed 7bd019398f4fa6e6b0f7050c9bbc38724fdabd2e

tests/epyccel/test_epyccel_lists.py
EmilyBourne: Fixed 8e921a7c9a042604974a2e1d53a18cfc8b3e8d3f

pyccel/ast/builtin_objects/list_functions.py
```diff
@@ -0,0 +1,65 @@
...
+    _shape = None
+    _class_type = NativeHomogeneousList()
+    name = 'pop'
+    def __init__(self, name, index_element=None):
+        if index_element:
+            super().__init__(name, index_element)
+        else:
+            super().__init__(name)
+        self._index = index_element
+        self._list_variable = name
```
EmilyBourne, line 45: You are saving `name` in an attribute called `_list_variable`. These two names seem to imply different things. Why the difference? In my opinion `list_variable` makes more sense than `name` as you are storing a `TypedAstNode` describing a list not the name of a Variable (see similar discussion here: https://github.com/pyccel/pyccel/pull/1709#discussion_r1485854955)

EmilyBourne (changes requested):
One final comment

pyccel/ast/builtin_objects/list_functions.py
```diff
@@ -0,0 +1,65 @@
...
+    index_element : TypedAstNode
+        The current index value for the element to be popped.
+    """
+    __slots__ = ('_dtype','_precision', '_index','_list_variable')
+    _rank = 0
+    _order = None
+    _shape = None
+    _class_type = NativeHomogeneousList()
+    name = 'pop'
+    def __init__(self, last_variable, index_element=None):
```
EmilyBourne, line 39: Typo `last_variable`->`list_variable`

pyccel/ast/builtin_objects/list_functions.py
```diff
@@ -0,0 +1,65 @@
...
+    _shape = None
+    _class_type = NativeHomogeneousList()
+    name = 'pop'
+    def __init__(self, list_variable, index_element=None):
+        if index_element:
+            super().__init__(list_variable, index_element)
+        else:
+            super().__init__(list_variable)
+        self._index = index_element
+        self._list_variable = list_variable
```
EmilyBourne, line 45: See the [docs](https://github.com/pyccel/pyccel/blob/devel/developer_docs/ast_nodes.md#pyccel-internal-function)

> The arguments can be passed through in the constructor of the sub-class to let PyccelInternalFunction take care of saving the arguments, or the arguments can be saved with more useful names inside the class. However care must be taken to ensure that the argument is not saved inside the class twice.

Here the argument is saved twice

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ All notable changes to this project will be documented in this file.
 -   #1605 : Add support for methods and interfaces in classes (including `__init__` and `__del__`).
 -   #1618 : Add support for class instance attributes.
 -   #1680 : Add support for `typing.Final`.
+-   #1690 : Add support for `pop()` method of list class
```
EmilyBourne, line 28: Please move this to the new UNRELEASED section

yguclu:
@mustapha-belbiad Please resolve the conflicts with the `devel` branch and update your branch.

pyccel/ast/builtin_objects/list_functions.py
yguclu, line 1: Please add the `__all__` tuple to this file.

pyccel/codegen/printing/pycode.py
```diff
@@ -1136,6 +1136,13 @@ def _print_TypingFinal(self, expr):
         annotation = self._print(expr.arg)
         return f'const {annotation}'
 
+    def _print_ListPop(self, expr):
+        args = ""
+        if expr.pop_index:
+            args = self._print(expr.pop_index)
```
yguclu, line 1142: Small suggestion. Instead of redefining the variable `args`, it is more Pythonic to use the single-liner
```suggestion
        args = self._print(expr.pop_index) if expr.pop_index else ""
```

yguclu (changes requested):
Good job! I have a few minor comments.

Do not forget to resolve the conflicts with `devel`!

pyccel/ast/builtin_objects/list_functions.py
EmilyBourne: Beware, the name of this file changed in #1709 you should also move your functions when you merge devel into your branch

EmilyBourne:
@mustapha-belbiad Don't forget to react to the comments on your PR. See the [docs](https://github.com/pyccel/pyccel/blob/devel/developer_docs/review_process.md):

pyccel/codegen/printing/pycode.py
EmilyBourne: Pyccel only works properly if your code runs properly in Python. `.pop(1,3,4)` will raise an error when you run your code in Python so we can safely ignore this case in Pyccel.

yguclu (approved):

PR #1726, opened 2024-02-13
Fix Pyccel checks

EmilyBourne:
The PR test which checks that all AST nodes are following the expected conventions (`__slots__` and `_attribute_nodes` defined) is not running on subfolders. As new PRs are introducing subfolders this needs fixing ASAP. This PR fixes that

### DIFF
--- a/ci_tools/check_slots.py
+++ b/ci_tools/check_slots.py
@@ -93,7 +93,7 @@ if __name__ == '__main__':
 
     # Get ast modules
     ast_folder = os.path.dirname(ast.__file__)
-    ast_modules = [mod[:-3] for mod in os.listdir(ast_folder) if mod != '__init__.py' and mod.endswith('.py')]
+    ast_modules = ['.'.join(f.parts)[:-3] for f in folder.rglob('*.py') if f.parts[-1] != '__init__.py']
 
     # Prepare error collection
     missing_all = []
--- a/ci_tools/check_slots.py
+++ b/ci_tools/check_slots.py
@@ -4,6 +4,7 @@ import argparse
 import importlib
 import inspect
 import os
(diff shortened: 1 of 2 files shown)


EmilyBourne:
/bot run coverage macosx windows pylint docs

Farouk-Echaref (approved):
Looks good to me

ci_tools/check_slots.py
```diff
@@ -92,8 +92,8 @@ def fill_dictionary(title, message, file, start_line, end_line, annotation_level
     args = parser.parse_args()
 
     # Get ast modules
-    ast_folder = os.path.dirname(ast.__file__)
-    ast_modules = [mod[:-3] for mod in os.listdir(ast_folder) if mod != '__init__.py' and mod.endswith('.py')]
+    ast_folder = pathlib.Path(ast.__file__).parent
+    ast_modules = ['.'.join(f.relative_to(ast_folder).parts)[:-3] for f in ast_folder.rglob('*.py') if f.parts[-1] != '__init__.py']
```
yguclu, line 96: Why are you taking the last three characters of a string? Should it not be this instead?
```suggestion
    ast_modules = ['.'.join(f.relative_to(ast_folder).parts[:-3]) for f in ast_folder.rglob('*.py') if f.parts[-1] != '__init__.py']
```

EmilyBourne: I'm not taking the last three characters. I am stripping the last three characters. In other words removing the extension `.py` so I can create a module name that can be imported

yguclu: True! I read it in the opposite way! 😅

yguclu (approved):

PR #1727, opened 2024-02-13
List support/insert method

### LINKED ISSUE #1692: insert: inserts a given element at a given index in a list.
### Feature Description

Add support for insert() method for lists in the Python translation.

### Test Code

This Code Should Work 

```python
digits = [3, 4, 6]

digits.insert(1, 2)

print(digits)
>> [3, 2, 4, 6]
```

### Proposed Solution
See description in #1529

Farouk-Echaref:
Add support for the insert() method of Python lists to the semantic stage.
Add handling of that method to the Python printer. This fixes #1692.

**Commit Summary**

- Add a class representation for the `insert()` method, inheriting from `PyccelInternalFunction`. The constructor ensures compatibility between the datatypes of the `insert()` argument and the list for homogenous lists.
- Update the `_class_type` attribute for methods that return `None` 
- Add the python printer `_print_ListInsert()` to generate the appropriate Python representation of the `insert()` method.
- Add success and error tests for the `insert()` method involving many scenarios where the function could be used.


mustapha-belbiad (approved):
To me, everything looks good

pyccel/ast/builtin_methods/list_methods.py
```diff
@@ -14,7 +14,7 @@
 from pyccel.ast.internals import PyccelInternalFunction
 
 
-__all__ = ('ListAppend',)
+__all__ = ('ListAppend','ListInsert',)
```
EmilyBourne, line 17: `__all__` is usually written with one object per line. This ensures it stays readable as it gets longer:
```suggestion
__all__ = ('ListAppend',
           'ListInsert',)
```

pyccel/ast/builtin_methods/list_methods.py
```diff
@@ -79,3 +79,79 @@ def append_argument(self):
...
+    Parameters
+    ----------
+    list_variable : Variable
+        The variable representing the list.
+
+    index : TypedAstNode
+        The index value for the element to be added.
+    
+    new_elem : Variable
+        The argument passed to insert() method.
```
EmilyBourne, line 105: Why does the `new_elem` need to be a `Variable`? Is there a problem with:
```python
a : list[int] = []
a.insert(0, 1+2)
```
? Or does the semantic stage create a temporary variable here?

pyccel/codegen/printing/pycode.py
```diff
@@ -852,6 +852,14 @@ def _print_ListAppend(self, expr):
 
         return f"{list_var}.{method_name}({append_arg})\n"
 
+    def _print_ListInsert(self, expr):
+        method_name = expr.name
+        index = self._print(expr.index)
+        list_var = self._print(expr.list_variable)
+        insert_arg = self._print(expr.insert_argument)
+
+        return f"{list_var}.{method_name}({index},{insert_arg})\n"
```
EmilyBourne, line 861: For readability and as a convention in most Python codes spaces are usually used after commas
```suggestion
        return f"{list_var}.{method_name}({index}, {insert_arg})\n"
```

tests/epyccel/test_epyccel_lists.py
```diff
@@ -79,3 +79,62 @@ def f():
...
+
+def test_insert_range_tuple(language):
+    def f():
+        a = [[1, 2, 3]]
+        for i in range(4, 1000):
+            a.insert(i, (i, i + 1))
+        return a
+
+    epyc_f = epyccel(f, language=language)
+    assert f() == epyc_f()
```
EmilyBourne, line 140: All these tests insert in the last position. Please add a test which inserts to a different location. `insert` is usually used for prepending data so it would be good to at least have a test for this. With this many tests you could also test inserting in the middle of the list in one of them.

EmilyBourne (changes requested):
Looks good. I just have a few comments

pyccel/ast/builtin_methods/list_methods.py
Farouk-Echaref: That was my previous understanding (the semantic creates a temporary variable) but I was wrong, now that i've tested it with different values, `new_elem` could be a `Variable`, `Literal`, `PyccelAdd` ... in general, a `TypedAstNode`

EmilyBourne: From what I remember the semantic creates a temporary variable when memory needs to be allocated. For simple objects like `1+2` this is not necessary so it just passes the object

pyccel/ast/builtin_methods/list_methods.py
Farouk-Echaref: updated in 9ed608928a7697654d55948ee1bbe70304500d9e

pyccel/codegen/printing/pycode.py
Farouk-Echaref: updated in cd5067e8f92adfe31a66bee6e69028e6d7ef56a0

tests/epyccel/test_epyccel_lists.py
Farouk-Echaref: updated in 7a0a586792d87284a8953cc5e8fb268043871096

EmilyBourne:
@Farouk-Echaref Please resolve the conflicts in this PR

tests/epyccel/test_epyccel_lists.py
```diff
@@ -117,3 +117,79 @@ def f():
...
+            pass
+        a = A()
+        b = A()
+        c = A()
+        d = A()
+        e = A()
+        lst = [a, b, c]
+        lst.insert(0, d)
+        lst.insert(1, e)
+        return a
```
EmilyBourne, line 192: This is a really nice idea but it seems that it is crashing your tests. I had a quick look and Pyccel doesn't currently handle classes inside functions (but unfortunately it doesn't raise a neat error about this). If you want to write a test like this then you will need to create a test module:
```python
class A:
    pass
def f():
    a = A()
    b = A()
    c = A()
    d = A()
    e = A()
    lst = [a, b, c]
    lst.insert(0, d)
    lst.insert(1, e)
    return a
```
and translate the whole module

EmilyBourne: The problem is that `mod.A` and `modnew.A` are not the same type even if they look like it. For this you can check the length of the list and you can check the properties of the class to make sure the list is correctly ordered. For example:
```python
class A:
    def __init__(self, x : int):
        self.x = x

def f():
    a = A(1)
    b = A(2)
    c = A(3)
    d = A(4)
    e = A(5)
    lst = [a, b, c]
    lst.insert(0, d)
    lst.insert(1, e)
    return a
```

```
def test_insert_user_defined_objects(language):
    import modules.list_user_defined_objs as mod

    modnew = epyccel(mod, language=language)
    python_list = mod.fn()
    accelerated_list = modnew.fn()
    assert len(python_list) == len(accelerated_list)
    for python_elem, accelerated_elem in zip(python_list, accelerated_list):
         assert python_elem.x == accelerated_elem.x
```

Farouk-Echaref: tried this:
```python
class A:
    def __eq__(self, other):
        return isinstance(other, A)

def fn():
    a = A()
    b = A()
    c = A()
    d = A()
    e = A()
    lst = [a, b, c]
    lst.insert(0, d)
    lst.insert(1, e)
    return lst
```
but got a semantic error, i don't really know the extent of the classes support in Pyccel

EmilyBourne: Classes support:
- variables
- basic functions

They don't support:
- magic methods
- decorators

I'm not sure we have support for `isinstance` either actually

Farouk-Echaref: fixed in 34d82c90fff7be918ee42328813295c2bee65ca9

pyccel/ast/builtin_methods/list_methods.py
```diff
@@ -15,8 +15,9 @@
 
 
 __all__ = ('ListAppend',
+           'ListInsert',
+           'ListPop',
            'ListClear',
-           'ListPop'
```
EmilyBourne, line 19: Please fix the pylint warning

pyccel/ast/builtin_methods/list_methods.py
```diff
@@ -83,7 +84,82 @@ def append_argument(self):
...
+    Represents a call to the .insert() method.
+
+    Represents a call to the .insert() method of an object with a list type,
+    which inserts a given element at a given index in a list. This method returns `None`.
+    The insert method is called as follows:
+
+    >>> a = [2, 3, 4]
+    >>> a.insert(1, 1)
+    >>> print(a)
+    [1, 2, 3, 4]
```
EmilyBourne, line 98: Did you run this test or just type it out? Python is zero indexed so shouldn't the result be:

```suggestion
    >>> a = [2, 3, 4]
    >>> a.insert(1, 1)
    >>> print(a)
    [2, 1, 3, 4]
```
?

pyccel/ast/builtin_methods/list_methods.py
```diff
@@ -83,7 +84,82 @@ def append_argument(self):
...
+        The argument passed to insert() method.
+    """
+    __slots__ = ("_index", "_list_variable", "_insert_arg")
+    _attribute_nodes = ("_index", "_list_variable", "_insert_arg")
+    _dtype = NativeVoid()
+    _shape = None
+    _order = None
+    _rank = 0
+    _precision = -1
+    _class_type = NativeHomogeneousList()
```
EmilyBourne, line 118: Insert returns a homogenous list of void objects?

pyccel/ast/builtin_methods/list_methods.py
```diff
@@ -133,7 +209,6 @@ def list_variable(self):
         """
         return self._list_variable
 
-#==============================================================================
```
EmilyBourne, line 136: These separators help with the readability. Why is this removed?

tests/epyccel/test_epyccel_lists.py
```diff
@@ -1,4 +1,4 @@
-# pylint: disable=missing-function-docstring, missing-module-docstring
+# pylint: disable=missing-function-docstring, missing-module-docstring, missing-class-docstring
```
EmilyBourne, line 1: This change is no longer needed. There are no classes in this file

tests/epyccel/test_epyccel_lists.py
```diff
@@ -118,6 +118,74 @@ def f():
...
+        a.insert(4, 4)
+        return a
+
+    epyc_f = epyccel(f, language=language)
+    assert f() == epyc_f()
+
+def test_insert_multiple(language):
+    def f():
+        a = [1, 2, 3]
+        a.insert(4, 4)
```
EmilyBourne, line 133: @yguclu this is apparently valid code in Python:
```
a = [1,2,3]
a.insert(100, 4)
```
We can usea an insert index much larger than the size of the list. The element is then simply appended. Is this something we want to support? From a performance perspective it seems similar to negative indexing

EmilyBourne (changes requested):

pyccel/ast/builtin_methods/list_methods.py
Farouk-Echaref: my bad, an oversight since i was thinking of inserting in the "first" position and got mixed up with index 1, will fix in the next commit

pyccel/ast/builtin_methods/list_methods.py
Farouk-Echaref: Certainly! Here's the corrected version:

Not really. Initially, it wasn't clear to me what the `_class_type` attribute was meant to represent. I thought it was related to the class that the method was associated with, but that was a stupid mistake. It should be `NativeVoid` as well.

Farouk-Echaref: Not really. Initially, it wasn't clear to me what the `_class_type` attribute was meant to represent. I thought it was related to the class that the method was associated with, but that was a stupid mistake. It should be `NativeVoid` as well.

pyccel/ast/builtin_methods/list_methods.py
Farouk-Echaref: Must've removed them while resolving conflicts somehow, i'll put them back

EmilyBourne:
@Farouk-Echaref the description seems to refer to changes which were merged in other PRs. Don't forget to update the commit summary

pyccel/ast/builtin_methods/list_methods.py
Farouk-Echaref: fixed in ecb58740e3f5b11a07555ef325f3933982e58160

pyccel/ast/builtin_methods/list_methods.py
Farouk-Echaref: fixed in ecb58740e3f5b11a07555ef325f3933982e58160

pyccel/ast/builtin_methods/list_methods.py
Farouk-Echaref: fixed in 5351255e5d429627252ffe3fa7d18707a789db5f

pyccel/ast/builtin_methods/list_methods.py
Farouk-Echaref: fixed in b57fc5722e45d99a550c34b5d7db72559ce97fcd

tests/epyccel/test_epyccel_lists.py
Farouk-Echaref: fixed in b9d140fcf30aa56f70f2a360875aa18856b97bc5

pyccel/ast/builtin_methods/list_methods.py
EmilyBourne: Thanks. FYI the order in `__all__` doesn't need to match the order of the classes in the file

Farouk-Echaref: Noted, thanks.

pyccel/ast/builtin_methods/list_methods.py
```diff
@@ -83,89 +85,169 @@ def append_argument(self):
         """
         return self._append_arg
 
-#==============================================================================
-class ListPop(PyccelInternalFunction) :
+#=======================================================================================
+
+class ListClear(PyccelInternalFunction) :
     """
```
EmilyBourne, line 91: Please can you put these classes back in the order that they appear in the devel branch. By reordering everything is showing up as changed. But I am not sure that you have actually changed the implementation of either class?

pyccel/ast/builtin_methods/list_methods.py
```diff
@@ -83,89 +85,169 @@ def append_argument(self):
...
+        The current index value for the element to be popped.
+        """
+        return self._index
+
     @property
     def list_variable(self):
         """
         Provide the name of the list as the return value.
-
+
```
EmilyBourne, line 250: Unnecessary change that usually triggers pylint. Trailing whitespace

EmilyBourne (approved):
Looks good. Well done

tests/epyccel/test_epyccel_lists.py
yguclu: Yes, I think we should support it. Lists cannot compete with arrays performance-wise, and list insertion is definitely not a fast operation. Most likely, comparing the given index with the size of `a` has a totally negligible overhead.

yguclu: @Farouk-Echaref I suppose that the Python printer has no issues supporting this syntax. So I suggest to add a test for it.

EmilyBourne: The example above does test this. You can insert at positions `0, 1, 2, 3` or out of bounds (4)

pyccel/ast/builtin_methods/list_methods.py
```diff
@@ -16,10 +16,11 @@
 
 __all__ = ('ListAppend',
            'ListClear',
-           'ListPop'
+           'ListInsert',
+           'ListPop',
            )
 
-
+#=======================================================================================
```
yguclu, line 23: Please try to keep the lines within 80 characters (including the terminal newline characters).

pyccel/ast/builtin_methods/list_methods.py
```diff
@@ -83,7 +84,7 @@ def append_argument(self):
         """
         return self._append_arg
 
-#==============================================================================
+#=======================================================================================
```
yguclu, line 87: Same comment

pyccel/ast/builtin_methods/list_methods.py
```diff
@@ -133,7 +134,7 @@ def list_variable(self):
         """
         return self._list_variable
 
-#==============================================================================
+#=======================================================================================
```
yguclu, line 137: Same comment

pyccel/ast/builtin_methods/list_methods.py
```diff
@@ -169,3 +170,80 @@ def list_variable(self):
...
         return self._list_variable
+
+#=======================================================================================
+class ListInsert(PyccelInternalFunction):
+    """
+    Represents a call to the .insert() method.
+
+    Represents a call to the .insert() method of an object with a list type,
+    which inserts a given element at a given index in a list. This method returns `None`.
+    The insert method is called as follows:
```
yguclu, line 181: Same comment. Please stay within 80 characters per line, unless this makes the code unreadable.

pyccel/ast/class_defs.py
```diff
@@ -139,6 +139,8 @@
         methods=[
             PyccelFunctionDef('append', func_class = ListAppend,
                 decorators = {}),
+            PyccelFunctionDef('insert', func_class = ListInsert,
+                decorators = {}),
             PyccelFunctionDef('pop', func_class = ListPop,
                 decorators = {}),
             PyccelFunctionDef('clear', func_class = ListClear,
```
yguclu, line 146: If you want to list the methods alphabetically, than maybe move `clear` to before `insert`?

tests/errors/semantic/blocking/INSERT_LIST_INCOMPATIBLE_TYPES_2.py
```diff
@@ -0,0 +1,6 @@
+# pylint: disable=missing-function-docstring, missing-module-docstring
+
+a = [1,2,3]
+b = (4,5)
+a.append(3, b)
```
yguclu, line 5: This is a call to `append`, not `insert`.

tests/errors/semantic/blocking/INSERT_LIST_INCOMPATIBLE_TYPES_0.py
```diff
@@ -0,0 +1,7 @@
+# pylint: disable=missing-function-docstring, missing-module-docstring
+
+import numpy as np
+a = [1,2,3]
```
yguclu, line 4: Please leave (at least) one blank line after the import statements.

yguclu (changes requested):
Looks good! I have some minor comments

pyccel/ast/builtin_methods/list_methods.py
Farouk-Echaref: fixed in 12cd0b2a08f4f7d59e7c61e94769a7a9d9092039

pyccel/ast/builtin_methods/list_methods.py
Farouk-Echaref: fixed in 12cd0b2a08f4f7d59e7c61e94769a7a9d9092039

pyccel/ast/builtin_methods/list_methods.py
Farouk-Echaref: fixed in 12cd0b2a08f4f7d59e7c61e94769a7a9d9092039

pyccel/ast/builtin_methods/list_methods.py
Farouk-Echaref: fixed in 12cd0b2a08f4f7d59e7c61e94769a7a9d9092039

pyccel/ast/class_defs.py
Farouk-Echaref: fixed in 650a33f87857e6ba07ceeb7a3837a2df10272e9d

tests/errors/semantic/blocking/INSERT_LIST_INCOMPATIBLE_TYPES_2.py
Farouk-Echaref: fixed in cfa5a2a986d745a91d730022bb3f166c06a32311

tests/errors/semantic/blocking/INSERT_LIST_INCOMPATIBLE_TYPES_0.py
Farouk-Echaref: fixed in cfa5a2a986d745a91d730022bb3f166c06a32311

tests/epyccel/test_epyccel_lists.py
Farouk-Echaref: I think inserting at the index 4 is already is out of bounds, supporting this in the python printer is no issue, my concerns were the equivalent generated code for C and Fortran

yguclu (approved):

PR #1760, opened 2024-02-23
Improve argument singletons

EmilyBourne:


mustapha-belbiad (approved):
Looks good to me

pyccel/utilities/metaclasses.py
```diff
@@ -4,23 +4,47 @@
...
-        if index not in cls._instances:
-            cls._instances[index] = super().__call__(*args, **kwargs)
-        return cls._instances[index]
+        existing_instance = cls._instances.get(index, None)
+        if existing_instance:
+            return existing_instance
+        else:
+            new_instance = super().__call__(*args, **kwargs)
+            cls._instances[index] = new_instance
+            return new_instance
```
EmilyBourne, line 47:
```suggestion
        existing_instance = cls._instances.get(index, None)
        if existing_instance is None:
            new_instance = super().__call__(*args, **kwargs)
            cls._instances[index] = new_instance
            return new_instance
        else:
            return existing_instance
```

pyccel/utilities/metaclasses.py
```diff
@@ -4,23 +4,47 @@
...
+    Parameters
+    ----------
+    name : str
+        The name of the class.
+    bases : tuple[class,...]
+        A tuple of the superclasses of the class.
+    dct : dict
+        A dictionary of the class attributes.
     """
     _instances = {}
```
yguclu, line 31: I find it a bit confusing to give an `_instances` attribute to the metaclass. This will mean that the dictionary is in the metaclass rather than in each of the actual argument-singleton types. Why don't we create the dictionary in the `__init__` instead?

EmilyBourne: I agree. I just kept the old implementation without thinking about it. Fixed: 4c0369d6

yguclu: If the dictionary is local to each argument-singleton type, it sounds like there is no need to include `cls` in the dictionary key. What do you think?

yguclu: If my last comment is correct, then we have no need to use a dictionary in the "simple" singleton case. And in fact it would be beneficial not to have the metaclass `Singleton` subclass `ArgumentSingleton`, in order to simplify the code.

EmilyBourne: I agree. I have removed it: b273ef98

The original reason for making `Singleton` subclass `ArgumentSingleton` was to allow a class (`DataType`) to have the general metaclass (`ArgumentSingleton`), and the subclasses to redefine it (which is only allowed if the new metaclass is a subclass). But most of the time we want `Singleton` anyway and I am not sure that this is really helpful. I have split them back up again. [3405e0f](https://github.com/pyccel/pyccel/pull/1760/commits/3405e0f80e036b27426e95718aa3c2b1208428b4)

PR #1761, opened 2024-02-23
Remove dead code

EmilyBourne:
The `len` function is handled through the `__new__` function for all examples except homogeneous tuple literals. This does not need to be a special case. The unnecessary code is removed and tests are added.

### DIFF
--- a/tests/epyccel/test_builtins.py
+++ b/tests/epyccel/test_builtins.py
@@ -357,3 +357,15 @@ def test_len_tuple(language):
     epyc_f = epyccel(f, language=language)
 
     assert epyc_f() == f()
+
+
+def test_len_inhomog_tuple(language):
+    def f():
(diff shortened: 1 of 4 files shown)


Farouk-Echaref (approved):
Looks good!

tests/epyccel/test_builtins.py
```diff
@@ -333,3 +333,27 @@ def f(x : 'T', y : 'T'):
...
+
+
+def test_len_tuple(language):
+    def f():
+        a = (3,4)
+        b = (4,3,5)
+        c = b
+        return len(a), len(b), len(c), len((1,2))
+
+    epyc_f = epyccel(f, language=language)
```
yguclu, line 357: Should we test inhomogeneous tuples as well?

yguclu (approved):

PR #1762, opened 2024-02-26
Support array copy between different datatypes (#1641)

### LINKED ISSUE #1641: Can't copy array and change type in C
### Describe the bug
It is possible to create an array from another. This functions as a copy operator. This works well for objects of the same type but it fails in a C translation when the type is different

### To Reproduce
```python
import numpy as np

def create_float_copy(a : 'int[:,:]'):
    c = np.array(a, dtype=float)
    return c

if __name__ == '__main__':
    arr = np.ones((3,4), dtype=int)
    print(arr)
    arr2 = create_float_copy(arr)
    print(arr2)
```

### Error details
The output of the C translation is:
```
[[1 1 1 1]
[1 1 1 1]
[1 1 1 1]]
[[0.000000000000 0.000000000000 0.000000000000 0.000000000000]
[0.000000000000 0.000000000000 0.000000000000 0.000000000000]
[0.000000000000 0.000000000000 0.000000000000 0.000000000000]]
```

### Expected behavior
The output of the Python code is:
```
[[1 1 1 1]
 [1 1 1 1]
 [1 1 1 1]]
[[1. 1. 1. 1.]
 [1. 1. 1. 1.]
 [1. 1. 1. 1.]]
```

### Language
C

### Additional context
This occurs due to the implementation of `array_copy_data` here:
https://github.com/pyccel/pyccel/blob/6e6f088edc909ae339d52a5b46b0c2f0553be0d9/pyccel/stdlib/ndarrays/ndarrays.c#L491-L510
It uses `memcpy` which is very efficient for objects of the same datatype but leads to the wrong result in this case

VaradRengaraj:
Fixes #1641:

- Add functions to support array copy between different data types.
- Add test cases covering all possible scenarios of data type conversions.
- Add separate test cases with source and destination arrays of non-matching sizes.
- Add separate test cases with Fortran ordering.


pyccel/stdlib/ndarrays/ndarrays.c
```diff
@@ -490,6 +490,699 @@ bool is_same_shape(t_ndarray a, t_ndarray b)
...
+            {
+                *(int64_t *)(d + (offset * dest->type_size) + j) = *(bool *)(s + i);
+                j += dest->type_size;
+            }
+            break;
+
+        case nd_float:
+            for (int64_t i = 0; i < src.buffer_size; i += src.type_size)
+            {
+                *(float *)(d + (offset * dest->type_size) + j) = *(bool *)(s + i);
```
EmilyBourne, line 529: @VaradRengaraj Is this equivalent to:
```suggestion
                dest->raw_data[offset * dest->type_size + j] = (void)src.nd_float[s+i];
```
?
Is there a reason to prefer complicated (and therefore error prone) pointer arithmetic?

Regarding your question on #1641 you need to ensure that this warning is not raised so the PR can be merged. This can usually be done with things like `(float*)` . I'm not sure why the error here is:
> Casting between unsigned char * and float * which have an incompatible binary data representation.

I would have expected
> Casting between float and bool which have an incompatible binary data representation.

EmilyBourne:
@VaradRengaraj do you think there is a way to reduce the code duplication? Perhaps similarly to what is done for `sum`:
https://github.com/pyccel/pyccel/blob/46d0f51fa7cef391bb7568b9b22d3b718a4cb773/pyccel/stdlib/ndarrays/ndarrays.c#L558-L586

VaradRengaraj:
@EmilyBourne thanks for the comments. Let me look into your suggestions.

VaradRengaraj:
@EmilyBourne Hi, should I need to do something to run these tests?

EmilyBourne:
@VaradRengaraj I have unchecked the box:
>  Add tests to check your code works as expected

As I don't see any tests added in this PR. We need to add examples that were failing in the devel branch to show that your fix works correctly.

VaradRengaraj:
@EmilyBourne sorry about checking that box, i was not sure whether to check or uncheck that box. Can you please let me know where we add these tests?

EmilyBourne:
> @EmilyBourne sorry about checking that box, i was not sure whether to check or uncheck that box. Can you please let me know where we add these tests? 

The tests should be added in one of these two files:
https://github.com/pyccel/pyccel/blob/devel/tests%2Fepyccel%2Ftest_arrays.py
https://github.com/pyccel/pyccel/blob/devel/tests%2Fepyccel%2Frecognised_functions%2Ftest_numpy_funcs.py

EmilyBourne:
@VaradRengaraj you should be able to do this yourself once the checklist is complete. The missing item is the AUTHORS file. This is so that you can get credit for your work. Did you not want to fill this in for some reason? You can use your GitHub username if you want to remain anonymous but still get credit.

EmilyBourne:
The bot does not check if the checked item was done, just that it has been ticked so if you don't want credit for some reason you can tick this box and then mark the PR as ready for review

VaradRengaraj:
Changed the sentence in the test_arrays.py file as template is not the root cause of the problem. Template based code was only inducing this problem, but the compilation error happens because of the type conversion missing between non-compatible datatypes in Fortran.

tests/epyccel/test_arrays.py
```diff
@@ -3975,6 +3975,808 @@ def test_array_ndmin_2_order(language):
...
+    bl = randint(0, 2, size = size, dtype= bool)
+
+    epyccel_func = epyccel(arrays.dtype_convert_to_int32, language=language)
+    assert epyccel_func(bl) == arrays.dtype_convert_to_int32(bl)
+
+    epyccel_func = epyccel(arrays.dtype_convert_to_int64, language=language)
+    assert epyccel_func(bl) == arrays.dtype_convert_to_int64(bl)
+
+    epyccel_func = epyccel(arrays.dtype_convert_to_float32, language=language)
+    assert epyccel_func(bl) == arrays.dtype_convert_to_float32(bl)
```
EmilyBourne, line 4018: Please can you explain the logic for how these functions are split up? Did you just want to avoid having a large test?

tests/epyccel/test_arrays.py
```diff
@@ -3975,6 +3975,808 @@ def test_array_ndmin_2_order(language):
...
+    size = (2, 2)
+    integer8 = randint(low = iinfo('int8').min, high = iinfo('int8').max , size = size, dtype=np.int8)
+
+    epyccel_func = epyccel(arrays.dtype_convert_to_int8, language=language)
+    assert epyccel_func(integer8) == arrays.dtype_convert_to_int8(integer8)
+
+    epyccel_func = epyccel(arrays.dtype_convert_to_bool, language=language)
+    assert epyccel_func(integer8) == arrays.dtype_convert_to_bool(integer8)
+
+    epyccel_func = epyccel(arrays.dtype_convert_to_int16, language=language)
```
EmilyBourne, line 4077: These functions seem to be translated many times over these tests. Running the compiler is quite slow. It would be better to organise the tests in a way that means that each function only needs to be translated and compiled once.

tests/epyccel/test_arrays.py
```diff
@@ -3975,6 +3975,808 @@ def test_array_ndmin_2_order(language):
...
+    fl32 = np.float32(fl32)
+
+    epyccel_func = epyccel(arrays.dtype_convert_to_int32, language=language)
+    assert epyccel_func(fl32) == arrays.dtype_convert_to_int32(fl32)
+
+    epyccel_func = epyccel(arrays.dtype_convert_to_int64, language=language)
+    assert epyccel_func(fl32) == arrays.dtype_convert_to_int64(fl32)
+
+    epyccel_func = epyccel(arrays.dtype_convert_to_float32, language=language)
+    assert epyccel_func(fl32) == arrays.dtype_convert_to_float32(fl32)
```
EmilyBourne, line 4425: The error that you reported is about casting to a logical (boolean) type. Not all of these tests require such a conversion. Are you sure that they all fail?

tests/epyccel/test_arrays.py
VaradRengaraj: Putting everything under one function was making that function to execute for a long time. It was clearly visible when I ran the run_tests_py3.sh script. So i have split them them into 4 functions for each datatype.

EmilyBourne: This is probably linked to my comment below. Translating/compiling is slow so if you group by argument type instead of result type then yes the tests will be significantly slower

VaradRengaraj:
1. Addressed the review comments. Also there was a copy paste error in ndarrays.c file last time when I addressed the review comment there, because of which many test cases were failing. This was because I tested with old pyccel binaries mistakenly. Addressed that as well. 
2. Added many new test cases. This covers all possible code paths.

tests/epyccel/test_arrays.py
VaradRengaraj: Skipping the test cases for fortran language only for cases it will result in compilation error. Other cases, fortran tests are run now.

EmilyBourne (approved):
This looks great! Thanks for putting so much effort into the tests :)

tests/epyccel/test_arrays.py
```diff
@@ -3975,6 +3975,2070 @@ def test_array_ndmin_2_order(language):
...
+    fl_3 = uniform(finfo('float').min / 2, finfo('float').max / 2, size = size)
+
+    fl32_1 = uniform(finfo('float32').min / 2, finfo('float32').max / 2, size = size)
+    fl32_2 = uniform(finfo('float32').min / 2, finfo('float32').max / 2, size = size)
+    fl32_3 = uniform(finfo('float32').min / 2, finfo('float32').max / 2, size = size)
+    fl32_1 = np.float32(fl32_1)
+    fl32_2 = np.float32(fl32_2)
+    fl32_3 = np.float32(fl32_3)
+ 
+    fl64_1 = uniform(finfo('float64').min / 2, finfo('float64').max / 2, size = size)
```
EmilyBourne, line 5347:
```suggestion
    fl32_3 = np.float32(fl32_3)

    fl64_1 = uniform(finfo('float64').min / 2, finfo('float64').max / 2, size = size)
```

AUTHORS
```diff
@@ -30,3 +30,4 @@ Contributors
 * Priyabrata Mondal
 * Farouk Ech-Charef
 * Mustapha Belbiad
+* VaradRengaraj
```
yguclu, line 33: The authors' names should be listed as 
`First Last`
with a whitespace in between. Is there a reason for not using the whitespace?

EmilyBourne: Possibly relevant: https://github.com/pyccel/pyccel/pull/1762#issuecomment-1988299977

tests/epyccel/modules/arrays.py
```diff
@@ -1753,3 +1753,304 @@ def multiple_np_linspace():
...
+    c = array(arr, dtype='bool')
+    s = shape(c)
+    return len(s), c[0,0], c[0,1], c[1,0], c[1,1]
+
+@template('T1', ['bool[:,:]', 'int[:,:]', 'int8[:,:]', 'int16[:,:]', 'int32[:,:]', 'int64[:,:]', 'float[:,:]', 'float32[:,:]', 'float64[:,:]', 'complex64[:,:]', 'complex128[:,:]'])
+def dtype_convert_to_int8(arr : 'T1'):
+    from numpy import shape, array
+    c = array(arr, dtype='int8')
+    s = shape(c)
+    return len(s), c[0,0], c[0,1], c[1,0], c[1,1]
```
yguclu, line 1773: @EmilyBourne There is a lot of code duplication in this file. Clearly the problem is that Pyccel must be able to infer the type of the `c` array (and `d` array further down). Does this suggest that we need templated variables in addition to templated types?

AUTHORS
yguclu: I do not agree with using the GitHub username. The AUTHORS file is there to clearly identify who has the copyright over the code.

yguclu: Although one could argue that first and last name are not necessarily sufficient for that... 😅

tests/epyccel/modules/arrays.py
EmilyBourne: I'm not sure what you mean by "templated variables". The problem here is mostly on the Python side. I think Pyccel would happily translate:
```python
@template('T1', ['bool[:,:]', 'int[:,:]', 'int8[:,:]', 'int16[:,:]', 'int32[:,:]', 'int64[:,:]', 'float[:,:]', 'float32[:,:]', 'float64[:,:]', 'complex64[:,:]', 'complex128[:,:]'])
@template('T2', ['bool', 'int', 'int8', 'int16', 'int32', 'int64', 'float', 'float32', 'float64', 'complex64', 'complex128'])
def dtype_convert_to_int8(arr : 'T1'):
    from numpy import shape, array
    c = array(arr, dtype='T2')
    s = shape(c)
    return len(s), c[0,0], c[0,1], c[1,0], c[1,1]
```
But the Python code won't run. And we use the Python code for comparison. Another option is to add an extra parameter with the target type. E.g.:
```python
@template('T1', ['bool[:,:]', 'int[:,:]', 'int8[:,:]', 'int16[:,:]', 'int32[:,:]', 'int64[:,:]', 'float[:,:]', 'float32[:,:]', 'float64[:,:]', 'complex64[:,:]', 'complex128[:,:]'])
@template('T2', ['bool', 'int', 'int8', 'int16', 'int32', 'int64', 'float', 'float32', 'float64', 'complex64', 'complex128'])
def dtype_convert_to_int8(arr : 'T1[:,:]', type_var : 'T2'):
    from numpy import shape, array
    c = array(arr, dtype=type(type_var))
    s = shape(c)
    return len(s), c[0,0], c[0,1], c[1,0], c[1,1]
```

yguclu: The Python interpreter should be happy with
```python
@template('T1', ['bool[:,:]', 'int[:,:]', 'int8[:,:]', 'int16[:,:]', 'int32[:,:]', 'int64[:,:]', 'float[:,:]', 'float32[:,:]', 'float64[:,:]', 'complex64[:,:]', 'complex128[:,:]'])
def dtype_convert_to_other_dtype(arr : 'T1', dtype : type):
    from numpy import shape, array
    c = array(arr, dtype=dtype)
    s = shape(c)
    return len(s), c[0,0], c[0,1], c[1,0], c[1,1]
```
Unfortunately we cannot handle this code in Pyccel. As a workaround we could say that `dtype` is a "templated variable" which can only assume one of the values that we provide through a template. Maybe a proper MyPy annotation exists which provides the necessary information?

EmilyBourne: @VaradRengaraj Feel free to ignore this discussion. This is for us for future developments. It does not affect your PR

VaradRengaraj: Hi @EmilyBourne  i know this is not the right place to ask , but can you suggest me any bugs/ backlogs in Pyccel which is not time bound, so that I can work on my own pace. Can you suggest me any such issues?

EmilyBourne: > can you suggest me any bugs/ backlogs in Pyccel which is not time bound

I have replied on discord

EmilyBourne:
@VaradRengaraj Please fix the conflicts

AUTHORS
VaradRengaraj: @EmilyBourne No, please do it.

tests/epyccel/modules/arrays.py
VaradRengaraj: @EmilyBourne thank you

AUTHORS
EmilyBourne: Would you mind doing it. I assume that your name is "Varad Regaraj" but I don't actually know if this is your name or just your GitHub username

VaradRengaraj: ok I will do it.

tests/epyccel/test_arrays.py
```diff
@@ -6033,7 +6033,35 @@ def test_src_dest_array_diff_sizes_dtype_conversion_to_pyfloat_orderF(language):
     assert epyccel_func(fl32_1, fl32_2, fl32_3) == arrays.src_dest_diff_sizes_dtype_convert_to_pyfloat_orderF(fl32_1, fl32_2, fl32_3)
     assert epyccel_func(fl64_1, fl64_2, fl64_3) == arrays.src_dest_diff_sizes_dtype_convert_to_pyfloat_orderF(fl64_1, fl64_2, fl64_3)
     assert epyccel_func(cmplx64_1, cmplx64_2, cmplx64_3) == arrays.src_dest_diff_sizes_dtype_convert_to_pyfloat_orderF(cmplx64_1, cmplx64_2, cmplx64_3)
-    assert epyccel_func(cmplx128_1, cmplx128_2, cmplx128_3) == arrays.src_dest_diff_sizes_dtype_convert_to_pyfloat_orderF(cmplx128_1, cmplx128_2, cmplx128_3)
+    assert epyccel_func(cmplx128_1, cmplx128_2, cmplx128_3) == arrays.src_dest_diff_sizes_dtype_convert_to_pyfloat_orderF(cmplx128_1, cmplx128_2, cmplx128_3
```
EmilyBourne, line 6036: Bad merge:
```suggestion
    assert epyccel_func(cmplx128_1, cmplx128_2, cmplx128_3) == arrays.src_dest_diff_sizes_dtype_convert_to_pyfloat_orderF(cmplx128_1, cmplx128_2, cmplx128_3)
```

VaradRengaraj: yes there are some issues with merge. I will merge them manually and push those files again.

yguclu (approved):

tests/epyccel/test_arrays.py
yguclu: Fixed with 5a4f24e

PR #1765, opened 2024-02-27
Set support : copy method

mustapha-belbiad:
Add support for the `clear()` method of Python sets to the semantic stage. 
Add handling of that method to the Python printer. This fixes  #1740

PR #1756, opened 2024-02-20
Improve basic datatypes

EmilyBourne:
This PR rewrites the datatyping system. Fixes #1729.
As described in the issue, the existing superclass `Datatype` is replaced with 2 superclasses representing slightly different concepts:
- `PrimitiveType` : Representing the category of datatype (integer/floating point/etc)
- `PyccelType` : Representing the actual type of the object in Python

Subclasses of `PyccelType` generally fall into one of two categories:
- `FixedSizeType`
- `ContainerType`

These types are described in `developer_docs/type_inference.md` (it is recommended to read this file first before reviewing this pull request.

The `precision` is removed from `ast.basic.TypedAstNode` and is stored in the `FixedSizeType`s where it is relevant. This allows us to remove the necessity for negative precision values as defaults, as the types can now be easily differentiated using the different `FixedSizeType`s.
The attribute `_dtype` is removed from `ast.basic.TypedAstNode` instances. It is still accessible via the `dtype` property but is extracted from the class type. This ensures that the two do not deviate from one another.

The NumPy type deduction is significantly improved due to two major improvements. Firstly an assertion is added to `NumpyNDArrayType` to ensure that the internal datatypes are NumPy types. This uncovered a few bugs where the native Python types were used in our current code. Thus the wrong type would be returned if an element of one of these objects was returned.
Secondly `np.result_type` is now used to deduce the result type of NumPy functions. This function is implemented by NumPy which guarantees that it handles all corner cases and matches the expected behaviour of NumPy. Fixes #1763.

The `CustomDatatype` is simplified slightly to remove the unused parameters.

The printing of floats/complexes in Fortran is improved to stop printing decimal digits beyond the precision of the number. Additionally printing the return value of functions returning multiple results of different types is fixed.

Unfortunately these changes have a huge effect on the code as datatypes appear almost everywhere. I have tried my best to find all the places where the code has changed however it is possible that some places were missed if coverage is missing.

PR #1772, opened 2024-02-29
Set support: `pop` method

### LINKED ISSUE #1749: [SETS] pop : Removes an element from the set.
### Feature Description

**Add support for pop() method for sets in the Python translation.**
`pop()` :  Removes and returns an arbitrary element from the set. Raises KeyError if the set is empty.

### Test Code

This Code Should Work 

```python
my_set = {1, 2, 3}
popped_element = my_set.pop()
print(popped_element) 
>>> 1
```

### Proposed Solution
See description in #1531

mustapha-belbiad:
Add support for the `pop()` method of Python sets to the semantic stage. 
Add handling of that method to the Python printer. This fixes  #1749


mustapha-belbiad:
/bot run coverage docs pylint spelling pyccel_lint

mustapha-belbiad:
/bot run coverage docs pylint spelling pyccel_lint

EmilyBourne:
@mustapha-belbiad I don't think that this PR needs merging after #1765 . These 2 PRs should be independent. Please rebase this PR onto the devel branch so it can be reviewed simultaneously. The command should be something like:
```
git fetch
git rebase --onto origin/devel Set_Support-Remove_Method Set_Support-Pop_Method
```

pyccel/ast/builtin_methods/set_methods.py
```diff
@@ -114,12 +114,8 @@ class SetCopy(SetMethod):
...
 
     The copy() method in set class creates a shallow 
-    copy of a set object and returns it. 
-
-    Parameters
-    ----------
-    set_variable : TypedAstNode
-        The set on which the method will operate.
+    copy of a set object and returns it
+    The set on which the method will operate.
```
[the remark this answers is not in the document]
mustapha-belbiad: Fixed : 4830fd0

pyccel/ast/builtin_methods/set_methods.py
```diff
@@ -132,3 +132,30 @@ def __init__(self, set_variable):
...
         self._class_type = set_variable._class_type
         super().__init__(set_variable)
+
+
+class SetPop(SetMethod):
+    """
+    Represents a call to the .pop() method.
+
+    The pop() method pops an element from the set. 
+    It does not take any argument but returns the popped
```
EmilyBourne, line 142:
```suggestion
    It does not take any arguments but returns the popped 
```

pyccel/ast/builtin_methods/set_methods.py
```diff
@@ -132,3 +132,30 @@ def __init__(self, set_variable):
...
+    __slots__ = ('_dtype', '_precision', '_class_type')
+    _rank = 0
+    _order = None
+    _shape = None
+    name = 'pop'
+
+    def __init__(self, set_variable):
+        self._dtype = set_variable.dtype
+        self._precision = set_variable.precision
+        self._class_type = set_variable.class_type
```
EmilyBourne, line 160: Are you sure about the `class_type` here?

pyccel/ast/builtin_methods/set_methods.py
```diff
@@ -132,3 +132,30 @@ def __init__(self, set_variable):
...
+
+
+class SetPop(SetMethod):
+    """
+    Represents a call to the .pop() method.
+
+    The pop() method pops an element from the set. 
+    It does not take any argument but returns the popped 
+    element. It raises an error if the element is not 
+    present in the set.
```
EmilyBourne, line 144: This sentence looks like it has been copy/pasted from the `SetRemove` PR. I don't think it is relevant here.

EmilyBourne, line 144: This docstring contains a lot more about the `pop()` method than the `SetPop`class that it is documenting. It could be good to relate the two. E.g. by also mentioning how/if the `SetPop`class handles the mentioned error

EmilyBourne (changes requested):
I'm not convinced

pyccel/ast/builtin_methods/set_methods.py
mustapha-belbiad: I think it should be  `self._class_type = set_variable.dtype` because the set in Pyccel contains just hash scalar types (without tuples).

pyccel/ast/builtin_methods/set_methods.py
mustapha-belbiad: I will fix it in my next commit.

pyccel/ast/builtin_methods/set_methods.py
mustapha-belbiad: Yes, my bad. I will fix in the next commit. It should be `It raises an error if the set is empty.`

mustapha-belbiad: Yeah, you're right, but this class doesn't contain any error checks. I always have a question about why the docstring doesn't match what the class does?

mustapha-belbiad: Fixed : 35bef02

pyccel/ast/builtin_methods/set_methods.py
mustapha-belbiad: Fixed : 35bef02

pyccel/ast/builtin_methods/set_methods.py
mustapha-belbiad: Fixed : 21c64ad

pyccel/ast/builtin_methods/set_methods.py
EmilyBourne: > I always have a question about why the docstring doesn't match what the class does?

It is a little worrying if you think this is the case. The docstring _should_ describe what the class does. If you don't think this is the case then you may need to improve your docstrings.

In the AST, the class exists to describe a method so the two are linked. It does therefore make sense to describe the method being represented, but docs for these methods already exist. So you should only need to describe the method (e.g. `pop()`) briefly to give context to understand how the class works.

> Yeah, you're right, but this class doesn't contain any error checks.

That is why I think it is worth mentioning as you might expect it to given the docstring. For example you could write:
```
    Represents a call to the .pop() method.

    The pop() method pops an element from the set. 
    It does not take any arguments but returns the popped 
    element. It raises an error if the set is empty.
    The class does not raise an error as it assumes that the
    user code is valid.
```

mustapha-belbiad: Understood, I'll make the changes to the docstring according to your suggestion.

mustapha-belbiad: Fixed : d910add

pyccel/ast/builtin_methods/set_methods.py
```diff
@@ -123,3 +123,31 @@ def __init__(self, set_variable):
...
+    __slots__ = ('_dtype', '_precision', '_class_type')
+    _rank = 0
+    _order = None
+    _shape = None
+    name = 'pop'
+
+    def __init__(self, set_variable):
+        self._dtype = set_variable.dtype
+        self._precision = set_variable.precision
+        self._class_type = set_variable.dtype
```
EmilyBourne, line 152: Please update this section now that #1756 has been merged

mustapha-belbiad: Fixed : cfe3c6f

EmilyBourne (approved):
Looks good to me

yguclu (approved):

EmilyBourne:
Error due to slightly too tight tolerance. I am rerunning the test

PR #1764, opened 2024-02-27
Set support : clear method

mustapha-belbiad:
Add support for the `clear()` method of Python sets to the semantic stage.
I define an abstract class named `SetMethod` to handle calls to various set methods.
Add handling of that method to the Python printer. This fixes #1739.

PR #1777, opened 2024-03-04
Clarify developer docs

EmilyBourne:
As remarked in PR #1764 , the documentation is clear on when `_print` should be used, but not where it is not necessary. For example it is not necessary to call `_print` on an object which is known to be a string as this is not a `PyccelAstNode` so the 3 points used to justify the need to call `_print` do not apply:
1.  It ensures that the same conventions are used throughout the generated code (a string does not have enough context to apply conventions)
2.  It ensures that code details are not forgotten (there are no details available)
3.  It makes refactoring much simpler (a string is so fundamental that it will not be refactored via the `_print` implementation.


mustapha-belbiad (approved):
Looks good to me.

yguclu (approved):

PR #1780, opened 2024-03-05
Fix partial templates

### LINKED ISSUE #1779: Fix bug with partial templates
## Problem description
Pyccel fails during the semantic stage when trying to accelerate a function which uses a partial template `'T'`, if this not used alone for any of the arguments. That is, Pyccel works correctly if the arguments have types `('T', 'T[:]', 'T[:,:]')`, but fails if they have types `('T[:]', 'T[:,:]')`.

## To reproduce
In this code snippet
``` python

@template(name='T', types=[float, complex])
def mul_sum(a: 'T', v: 'T[:]'):
    return a * v.sum()

@template(name='T', types=[float, complex])
def just_sum(v: 'T[:]'):
    return v.sum()
```
the function `mul_sum` is pyccelized correctly, while the function `just_sum` leads to the error below.

## Error details
```bash/fortran/c
pyccel:
 |fatal [semantic]: test_template.py [8,13]| Undefined variable (T)
```

## Expected behavior
Pyccel should be able to process function `just_sum` with no issues. 

## Language
The error happens at the semantic stage, and it is independent of the language.

EmilyBourne:
Fix bug with partial templates (fixes #1779 ) where the partial template is incorrectly stripped away with unused template arguments. Add a test and update patch version.

### DIFF
--- a/tests/epyccel/test_epyccel_decorators.py
+++ b/tests/epyccel/test_epyccel_decorators.py
@@ -155,18 +155,16 @@ def test_indexed_template(language):
     pyccel_sum = epyccel(my_sum, language=language)
 
     x = np.ones(4, dtype=float)
-    y = np.ones(4, dtype=float)
 
     python_fl = my_sum(x)
-    pyccel_fl = pyccel_sum(y)
+    pyccel_fl = pyccel_sum(x)
 
     assert python_fl == pyccel_fl
     assert isinstance(python_fl, type(pyccel_fl))
 
-    x = np.ones(4, dtype=complex)
-    y = np.ones(4, dtype=complex)
+    y = np.ones(4, dtype=complex) * (1 + 3j)
 
-    python_cmplx = my_sum(x)
+    python_cmplx = my_sum(y)
     pyccel_cmplx = pyccel_sum(y)
 
     assert python_cmplx == pyccel_cmplx
--- a/tests/epyccel/test_epyccel_decorators.py
+++ b/tests/epyccel/test_epyccel_decorators.py
@@ -162,7 +162,7 @@ def test_indexed_template(language):
     assert python_fl == pyccel_fl
     assert isinstance(python_fl, type(pyccel_fl))
 
-    y = np.ones(4, dtype=complex) * (1 + 3j)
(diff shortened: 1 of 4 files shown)


tests/epyccel/test_epyccel_decorators.py
```diff
@@ -146,3 +146,28 @@ def get_val(x : int = None , y : int = None):
...
+    def my_sum(v: 'T[:]'):
+        return v.sum()
+
+    pyccel_sum = epyccel(my_sum, language=language)
+
+    x = np.ones(4, dtype=float)
+    y = np.ones(4, dtype=float)
+
+    python_fl = my_sum(x)
+    pyccel_fl = pyccel_sum(y)
```
yguclu, line 161: Why are two arrays needed?

tests/epyccel/test_epyccel_decorators.py
```diff
@@ -146,3 +146,28 @@ def get_val(x : int = None , y : int = None):
...
+
+    assert python_fl == pyccel_fl
+    assert isinstance(python_fl, type(pyccel_fl))
+
+    x = np.ones(4, dtype=complex)
+    y = np.ones(4, dtype=complex)
```
yguclu, line 167: Maybe we should have a proper complex test?
```suggestion
    x = np.ones(4, dtype=complex) * (1 + 3j)
    y = np.ones(4, dtype=complex)
```

tests/epyccel/test_epyccel_decorators.py
EmilyBourne: They're not :+1:

tests/epyccel/test_epyccel_decorators.py
EmilyBourne: Fixed f0cd8cc6

yguclu: Or better:
```suggestion
    x = np.full(4, 1 + 3j)
```

EmilyBourne: Fixed 201ec8b0

PR #1787, opened 2024-03-11
Ensure STC is installed with Pyccel

### LINKED ISSUE #1654: Ensure STC is installed with Pyccel
### Relevant Discussion

https://github.com/pyccel/pyccel/discussions/1053#discussioncomment-7290234

### Describe the feature
In order to add support for containers we need to include our new dependencies in Pyccel and ensure that they are correctly installed. In C we will use https://github.com/Stclib/STC/ . We need to make sure that the necessary files for compilation are packaged with Pyccel.

### Proposed Solution
Add https://github.com/Stclib/STC/ as a git submodule. Update the `pyproject.toml` file to ensure that the necessary files from STC are installed with Pyccel. We do not need to install the docs from the library.

There may need to be an investigation to find the most appropriate way to handle this. I think it would be a good idea to create a folder called `extensions` at the root and keep dependencies here. It is not clear how to map the include and src folders to the Pyccel folder from there. The solution will involve artifacts and may require symbolic links

mustapha-belbiad:
Fix #1654:

- Add https://github.com/Stclib/STC as a git submodule.
- Update file `pyproject.toml` to include essential files from STC.
- Implement an action in `check_for_stc` to verify STC installation.

### DIFF
--- a/.github/actions/check_for_stc/action.yml
+++ b/.github/actions/check_for_stc/action.yml
@@ -6,8 +6,11 @@ runs:
     - name: check stc installation
       shell: python
       run: |
-        import pyccel
+        import pyccel, os
         from os.path import exists, join
         pyccel_loc = pyccel.__path__[0]
         print(pyccel_loc)
+        files = os.listdir(directory)
+        for file in files:
+          print("---> ",file)
         assert(exists(join(pyccel_loc,'stc')))
--- a/.github/actions/check_for_stc/action.yml
+++ b/.github/actions/check_for_stc/action.yml
@@ -6,8 +6,12 @@ runs:
     - name: check stc installation
       shell: python
       run: |
-        import pyccel
+        import pyccel , os
         from os.path import exists, join
+        os.chdir('pyccel')
+        contents = os.listdir()
+        for item in contents:
+          print("===---------===> ", item)
         pyccel_loc = pyccel.__path__[0]
         print(pyccel_loc)
         assert(exists(join(pyccel_loc,'stc')))
--- a/.github/actions/check_for_stc/action.yml
+++ b/.github/actions/check_for_stc/action.yml
@@ -2,7 +2,7 @@ name: 'Ensure STC is installed with Pyccel'
 inputs:
   editable:
     description: 'editable installation'
-    default: 'true'
+    default: 'True'
 runs:
   using: "composite"
   steps:
--- a/.github/actions/check_for_stc/action.yml
+++ b/.github/actions/check_for_stc/action.yml
@@ -9,7 +9,8 @@ runs:
         import pyccel, os
         from os.path import exists, join
         pyccel_loc = pyccel.__path__[0]
+        input = os.getenv('editable')
         print(pyccel_loc)
         assert(exists(join(pyccel_loc,'extensions/STC/include/stc')))
-        if not os.access(pyccel_loc, os.W_OK):
+        if input is not None:
           assert(not exists(join(pyccel_loc,'extensions/STC/docs')))
--- a/.github/actions/check_for_stc/action.yml
+++ b/.github/actions/check_for_stc/action.yml
@@ -8,7 +8,7 @@ runs:
       run: |
         import pyccel , os
         from os.path import exists, join
-        os.chdir('pyccel')
+        os.chdir('pyccel/extensions')
         contents = os.listdir()
         for item in contents:
           print("===---------===> ", item)
--- a/.github/actions/check_for_stc/action.yml
+++ b/.github/actions/check_for_stc/action.yml
@@ -13,6 +13,7 @@ runs:
         from os.path import exists, join
         pyccel_loc = pyccel.__path__[0]
         print(pyccel_loc)
+        print("==========================================------------------> ${{ inputs.editable_string }}")
         assert(exists(join(pyccel_loc,'extensions/STC/include/stc')))
         if ${{ inputs.not_editable }}:
           assert(not exists(join(pyccel_loc,'extensions/STC/docs')))
--- a/.github/actions/check_for_stc/action.yml
+++ b/.github/actions/check_for_stc/action.yml
@@ -2,7 +2,6 @@ name: 'Ensure STC is installed with Pyccel'
 inputs:
   not_editable:
     description: 'not_editable installation'
-    default: 'true'
 runs:
   using: "composite"
   steps:
@@ -15,5 +14,5 @@ runs:
         print(pyccel_loc)
         print("${{ inputs.not_editable }}")
         assert(exists(join(pyccel_loc,'extensions/STC/include/stc')))
-        if ${{ inputs.not_editable }}:
+        if "${{ inputs.not_editable }}" != "-e":
           assert(not exists(join(pyccel_loc,'extensions/STC/docs')))
--- a/.github/actions/check_for_stc/action.yml
+++ b/.github/actions/check_for_stc/action.yml
@@ -8,7 +8,7 @@ runs:
       run: |
         import pyccel , os
         from os.path import exists, join
-        os.chdir('/pyccel')
+        os.chdir('pyccel')
         contents = os.listdir()
         for item in contents:
           print("===---------===> ", item)
--- a/.github/actions/check_for_stc/action.yml
+++ b/.github/actions/check_for_stc/action.yml
@@ -8,7 +8,7 @@ runs:
       run: |
         import pyccel , os
         from os.path import exists, join
-        os.chdir('pyccel/extensions/STC')
+        os.chdir('pyccel')
         contents = os.listdir()
         for item in contents:
           print("===---------===> ", item)
--- a/.github/actions/check_for_stc/action.yml
+++ b/.github/actions/check_for_stc/action.yml
@@ -6,16 +6,8 @@ runs:
     - name: check stc installation
       shell: python
       run: |
-        import pyccel , os
+        import pyccel
         from os.path import exists, join
-        os.chdir('pyccel')
-        contents = os.listdir()
-        for item in contents:
-          print("===---------===> ", item)
-        os.chdir('extensions/STC')
-        contents = os.listdir()
-        for item in contents:
-          print("===----2-----===> ", item)
         pyccel_loc = pyccel.__path__[0]
         print(pyccel_loc)
-        assert(exists(join(pyccel_loc,'stc')))
+        assert(exists(join(pyccel_loc,'extensions/STC/include/stc')))
--- a/.github/actions/check_for_stc/action.yml
+++ b/.github/actions/check_for_stc/action.yml
@@ -13,7 +13,7 @@ runs:
         from os.path import exists, join
         pyccel_loc = pyccel.__path__[0]
         print(pyccel_loc)
-        print("${{ inputs.editable_string }}")
+        print("${{ inputs.not_editable }}")
         assert(exists(join(pyccel_loc,'extensions/STC/include/stc')))
         if ${{ inputs.not_editable }}:
           assert(not exists(join(pyccel_loc,'extensions/STC/docs')))
--- a/.github/actions/check_for_stc/action.yml
+++ b/.github/actions/check_for_stc/action.yml
@@ -2,6 +2,7 @@ name: 'Ensure STC is installed with Pyccel'
 inputs:
   not_editable:
     description: 'not_editable installation'
+    default: 'True'
 runs:
   using: "composite"
   steps:
@@ -14,5 +15,5 @@ runs:
         print(pyccel_loc)
         print("${{ inputs.not_editable }}")
         assert(exists(join(pyccel_loc,'extensions/STC/include/stc')))
-        if "${{ inputs.not_editable }}" != "-e":
+        if ${{ inputs.not_editable }}:
           assert(not exists(join(pyccel_loc,'extensions/STC/docs')))
--- a/.github/actions/check_for_stc/action.yml
+++ b/.github/actions/check_for_stc/action.yml
@@ -8,7 +8,11 @@ runs:
       run: |
         import pyccel , os
         from os.path import exists, join
-        os.chdir('pyccel')
+        os.chdir('.')
+        contents = os.listdir()
+        for item in contents:
+          print("===---------===> ", item)
+        os.chdir('pyccel/extensions/STC')
         contents = os.listdir()
         for item in contents:
           print("===---------===> ", item)
--- a/.github/actions/check_for_stc/action.yml
+++ b/.github/actions/check_for_stc/action.yml
@@ -6,9 +6,10 @@ runs:
     - name: check stc installation
       shell: python
       run: |
-        import pyccel
+        import pyccel, os
         from os.path import exists, join
         pyccel_loc = pyccel.__path__[0]
         print(pyccel_loc)
         assert(exists(join(pyccel_loc,'extensions/STC/include/stc')))
-        assert(not exists(join(pyccel_loc,'extensions/STC/docs')))
+        if not os.access(pyccel_loc, os.W_OK):
+          assert(not exists(join(pyccel_loc,'extensions/STC/docs')))
--- a/.github/actions/check_for_stc/action.yml
+++ b/.github/actions/check_for_stc/action.yml
@@ -8,7 +8,7 @@ runs:
       run: |
         import pyccel , os
         from os.path import exists, join
-        os.chdir('pyccel/extensions')
+        os.chdir('pyccel/extensions/STC')
         contents = os.listdir()
         for item in contents:
           print("===---------===> ", item)
--- a/.github/actions/check_for_stc/action.yml
+++ b/.github/actions/check_for_stc/action.yml
@@ -1,8 +1,8 @@
 name: 'Ensure STC is installed with Pyccel'
 inputs:
-  non_editable:
-    description: 'non_editable installation'
-    default: 'non_editable'
+  editable:
+    description: 'editable installation'
+    default: 'true'
 runs:
   using: "composite"
   steps:
@@ -14,5 +14,5 @@ runs:
         pyccel_loc = pyccel.__path__[0]
         print(pyccel_loc)
         assert(exists(join(pyccel_loc,'extensions/STC/include/stc')))
-        if ${{ inputs.non_editable }}:
+        if ${{ inputs.editable }}:
           assert(not exists(join(pyccel_loc,'extensions/STC/docs')))
--- a/.github/actions/check_for_stc/action.yml
+++ b/.github/actions/check_for_stc/action.yml
@@ -6,11 +6,8 @@ runs:
     - name: check stc installation
       shell: python
       run: |
-        import pyccel, os
+        import pyccel
         from os.path import exists, join
         pyccel_loc = pyccel.__path__[0]
         print(pyccel_loc)
-        files = os.listdir(".")
-        for file in files:
-          print("---> ",file)
         assert(exists(join(pyccel_loc,'stc')))
--- a/.github/actions/check_for_stc/action.yml
+++ b/.github/actions/check_for_stc/action.yml
@@ -11,3 +11,4 @@ runs:
         pyccel_loc = pyccel.__path__[0]
         print(pyccel_loc)
         assert(exists(join(pyccel_loc,'extensions/STC/include/stc')))
+        assert(not exists(join(pyccel_loc,'extensions/STC/docs')))
--- a/.github/actions/check_for_stc/action.yml
+++ b/.github/actions/check_for_stc/action.yml
@@ -10,7 +10,7 @@ runs:
         from os.path import exists, join
         pyccel_loc = pyccel.__path__[0]
         print(pyccel_loc)
-        files = os.listdir(directory)
+        files = os.listdir(".")
         for file in files:
           print("---> ",file)
         assert(exists(join(pyccel_loc,'stc')))
--- a/.github/actions/check_for_stc/action.yml
+++ b/.github/actions/check_for_stc/action.yml
@@ -8,7 +8,7 @@ runs:
       run: |
         import pyccel , os
         from os.path import exists, join
-        os.chdir('.')
+        os.chdir('/pyccel')
         contents = os.listdir()
         for item in contents:
           print("===---------===> ", item)
--- a/.github/actions/check_for_stc/action.yml
+++ b/.github/actions/check_for_stc/action.yml
@@ -1,8 +1,8 @@
 name: 'Ensure STC is installed with Pyccel'
 inputs:
-  editable:
-    description: 'editable installation'
-    default: 'True'
+  not_editable:
+    description: 'not_editable installation'
+    default: 'true'
 runs:
   using: "composite"
   steps:
@@ -14,5 +14,5 @@ runs:
         pyccel_loc = pyccel.__path__[0]
         print(pyccel_loc)
         assert(exists(join(pyccel_loc,'extensions/STC/include/stc')))
-        if ${{ inputs.editable }}:
+        if ${{ inputs.not_editable }}:
           assert(not exists(join(pyccel_loc,'extensions/STC/docs')))
--- a/.github/actions/check_for_stc/action.yml
+++ b/.github/actions/check_for_stc/action.yml
@@ -13,7 +13,7 @@ runs:
         from os.path import exists, join
         pyccel_loc = pyccel.__path__[0]
         print(pyccel_loc)
-        print("==========================================------------------> ${{ inputs.editable_string }}")
+        print("${{ inputs.editable_string }}")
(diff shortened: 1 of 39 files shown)


.github/actions/check_for_stc/action.yml
```diff
@@ -0,0 +1,13 @@
...
+  using: "composite"
+  steps:
+    - name: check stc installation
+      shell: python
+      run: |
+        from extensions import STC
+        from os.path import exists, join
+        pyccel_loc = stc.__path__[0]
+        print(pyccel_loc)
+        assert(exists(join(pyccel_loc,'include')))
```
EmilyBourne, line 13: You need to ensure that this action is called from a workflow. I think it makes sense in the workflows which check the wheels

EmilyBourne: Sorry. I'll leave you to it and only look when you request a review ;)

.github/workflows/docs.yml
```diff
@@ -44,6 +44,10 @@ jobs:
           ref: ${{ env.COMMIT }}
           repository: ${{ env.PR_REPO }}
           fetch-depth: 0
+      - name: Checkout
+        uses: actions/checkout@v4.1.1
+        with:
+          submodules: true
```
EmilyBourne, line 50: Why did you add a new checkout action here?

.github/workflows/linux.yml
```diff
@@ -61,6 +61,10 @@ jobs:
         with:
           ref: ${{ env.COMMIT }}
           repository: ${{ env.PR_REPO }}
+      - name: Checkout
+        uses: actions/checkout@v4.1.1
+        with:
+          submodules: true
```
EmilyBourne, line 67: Does this workflow checkout Pyccel twice?

.github/workflows/windows.yml
```diff
@@ -57,6 +57,10 @@ jobs:
         with:
           ref: ${{ env.COMMIT }}
           repository: ${{ env.PR_REPO }}
+      - name: Checkout
+        uses: actions/checkout@v4.1.1
```
EmilyBourne, line 61: You don't need to specify the minor version of the checkout action

EmilyBourne (changes requested):
Don't forget the other actions that are only run on the devel branch

.github/workflows/docs.yml
EmilyBourne: That explains why you added:
```
          submodules: true
```
Not why you added a new:
```
      - name: Checkout
        uses: actions/checkout@v4.1.1
```

.github/workflows/linux.yml
mustapha-belbiad: I think each workflow needs to check out submodules

EmilyBourne: I agree. That is not my question

EmilyBourne: These lines check out more than just submodules

.github/workflows/docs.yml
```diff
@@ -44,6 +44,7 @@ jobs:
           ref: ${{ env.COMMIT }}
           repository: ${{ env.PR_REPO }}
           fetch-depth: 0
+          submodules: true
```
EmilyBourne, line 47: There are 2 checkout commands in this file

mustapha-belbiad: Fixed :  e721e3d

.github/workflows/coverage.yml
```diff
@@ -44,6 +44,7 @@ jobs:
         with:
           ref: ${{ env.COMMIT }}
           repository: ${{ env.PR_REPO }}
+          submodules: true
```
EmilyBourne, line 47: There are 2 checkout commands in this file

.github/workflows/docs.yml
```diff
@@ -38,12 +38,14 @@ jobs:
         with:
           ref: ${{ inputs.base }}
           path: base
+           submodules: true
```
EmilyBourne, line 41: Indentation error

mustapha-belbiad: fixed : f28f58c

.github/workflows/pickle_wheel.yml
```diff
@@ -78,8 +79,11 @@ jobs:
       - name: Check for existence of pickled files
         id: pickle_check
         uses: ./.github/actions/check_for_pickled
+      - name: Check for STC installation
+        id: stc_check
+        uses: ./.github/actions/check_for_stc
       - name: "Post completed"
```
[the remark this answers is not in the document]
mustapha-belbiad: You're right, i will add it in my next commit.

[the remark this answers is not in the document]
mustapha-belbiad: Fixed : b9e1d08

.github/workflows/comment_bot.yml
```diff
@@ -12,6 +12,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
+          submodules: true
```
EmilyBourne, line 15: The bot doesn't need access to the submodules

.github/workflows/clean_up.yml
```diff
@@ -13,6 +13,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
+          submodules: true
```
EmilyBourne, line 16: The bot doesn't need access to the submodules

.github/workflows/pickle.yml
```diff
@@ -87,6 +88,9 @@ jobs:
       - name: Check for existence of pickled files
         id: pickle_check
         uses: ./.github/actions/check_for_pickled
+      - name: Check for STC installation
+        id: stc_check
+        uses: ./.github/actions/check_for_stc
```
EmilyBourne, line 93: Should be added in status checks

.github/workflows/prepare_review_stage.yml
```diff
@@ -13,6 +13,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
+          submodules: true
```
EmilyBourne, line 16: The bot doesn't need access to the submodules

.github/workflows/pickle.yml
mustapha-belbiad: Fixed :[dc3fcce

.github/workflows/prepare_review_stage.yml
mustapha-belbiad: Fixed : f406515

.github/workflows/clean_up.yml
mustapha-belbiad: Fixed : f406515

.github/workflows/comment_bot.yml
mustapha-belbiad: Fixed : f406515

EmilyBourne (dismissed):
This looks good to me. I will trigger a complete test suite

pyproject.toml
```diff
@@ -52,3 +52,6 @@ artifacts = [
 [tool.hatch.build.targets.wheel.hooks.custom]
 path = "install_scripts/hook.py"
 require-runtime-dependencies = true
+
+[tool.hatch.build.targets.wheel.force-include]
+"./extensions/STC/include/stc" = "./pyccel/stc"
```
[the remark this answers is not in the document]
EmilyBourne: See https://hatch.pypa.io/1.9/plugins/builder/wheel/

> A [wheel](https://packaging.python.org/specifications/binary-distribution-format/) is a binary distribution of a Python package that can be installed directly into an environment.

I suspect what is needed is sdist: https://hatch.pypa.io/1.9/plugins/builder/sdist/

> A source distribution, or sdist, is an archive of Python "source code". Although largely unspecified, by convention it should include everything that is required to build a [wheel](https://hatch.pypa.io/1.9/plugins/builder/wheel/) without making network requests.

EmilyBourne:
@yguclu The implementation that we suggested where we install into `/home/emily/.local/lib/python3.10/site-packages/pyccel/stc` doesn't work in editable mode. In editable mode any files that can be modified aren't moved and the only symbolic links that is created is in `/home/emily/.local/lib/python3.10/site-packages/`.

I suggest a slightly different location to work round this. E.g. `/home/emily/.local/lib/python3.10/site-packages/pyccel_externals`

So far we have only worked out how to copy these additional files, not create an extra link.
This means that when we install in editable mode the external files are not editable. But they can be easily accessed.

So in an editable installation we would have at the start:
```
/home/emily/Code/pyccel/pyccel
/home/emily/Code/pyccel/externals/STC/include/stc
```
and after installation:
```
/home/emily/Code/pyccel/pyccel
/home/emily/.local/lib/python3.10/site-packages/pyccel_externals/stc
```
And in a non editable installation we would have at the start:
```
/home/emily/Code/pyccel/pyccel
/home/emily/Code/pyccel/externals/STC/include/stc
```
and after installation:
```
/home/emily/.local/lib/python3.10/site-packages/pyccel
/home/emily/.local/lib/python3.10/site-packages/pyccel_externals/stc
```

What do you think? In my opinion a fully editable installation would be better but the external files should rarely be modified so I think this solution would be sufficient for now

EmilyBourne:
@mustapha-belbiad What version of STC did you pin to? A specific release or just the current state of the master branch?

EmilyBourne:
> So far we have only worked out how to copy these additional files, not create an extra link.

After some more googling it seems that this is roughly equivalent to installing 2 projects from 1 `pyproject.toml` which is not advised or fully supported (at least by hatch). So either we need to accept that files that we should not be modifying (except to update the version) cannot be modified. Or we need to put the externals/ folder in pyccel/ and we cannot change the path (although we can avoid installing the README/docs/etc). Which is preferred? @yguclu @mustapha-belbiad

pyproject.toml
```diff
@@ -54,7 +54,7 @@ path = "install_scripts/hook.py"
 require-runtime-dependencies = true
 
 [tool.hatch.build.targets.wheel.force-include]
-"./pyccel/extensions/STC/include/stc" = "./pyccel/extensions/stc"
+"./pyccel/extensions/STC/include/stc" = "./pyccel/stc"
 
 [tool.hatch.build.targets.wheel.force_include_editable]
-"./pyccel/extensions/STC/include/stc" = "./pyccel/extensions/stc"
\ No newline at end of file
+"./pyccel/extensions/STC/include/stc" = "./pyccel/stc"
```
EmilyBourne: Have you tried with `sdist` instead of `wheel`? I don't think a wheel is created for an editable installation

mustapha-belbiad: Yes, unfortunately it didn't work. I'll need to look into it to identify the problem.

.github/actions/check_for_stc/action.yml
```diff
@@ -0,0 +1,13 @@
...
+  using: "composite"
+  steps:
+    - name: check stc installation
+      shell: python
+      run: |
+        import pyccel
+        from os.path import exists, join
+        pyccel_loc = pyccel.__path__[0]
+        print(pyccel_loc)
+        assert(exists(join(pyccel_loc,'extensions/STC/include/stc')))
```
EmilyBourne, line 13: Please can you add a few lines to this action to make sure that we don't install the docs. I don't want them creeping back in in the future.

docs/installation.md
```diff
@@ -203,6 +203,7 @@ for a system-wide installation.
     ```sh
     git clone git@github.com:pyccel/pyccel.git
     cd pyccel
+    git submodule update --init --recursive
```
EmilyBourne, line 206:
```suggestion
    git clone --recurse-submodules git@github.com:pyccel/pyccel.git
    cd pyccel
```

docs/installation.md
```diff
@@ -211,6 +212,7 @@ for a system-wide installation.
...
```
EmilyBourne, line 215: See above

docs/installation.md
mustapha-belbiad: fixed : f76c559

docs/installation.md
mustapha-belbiad: fixed : f76c559

.github/actions/check_for_stc/action.yml
mustapha-belbiad: Fixed : dd1abd8

EmilyBourne (approved):
This looks good to me now. Thanks for taking care of this

.github/actions/check_for_stc/action.yml
```diff
@@ -0,0 +1,14 @@
...
+        assert(not exists(join(pyccel_loc,'extensions/STC/docs')))
```
EmilyBourne, line 14: Don't you need an if condition here so it doesn't complain in editable mode?

EmilyBourne: Answer, yes:
https://github.com/pyccel/pyccel/actions/runs/8468392946/job/23201312501

mustapha-belbiad: you're right, I will add it in my next commit.

mustapha-belbiad: I added a condition her :  787aeb9. Let me know if it's the right way to solve this problem.

EmilyBourne: I'm not sure this is the right way to do this. A user installation (`--user`) is writable but is not necessarily editable. I think the simplest is to parametrise the test

EmilyBourne: The link tells you what to put in the action.yml file. Maybe you can look at our `post_completed` action for an example

.github/actions/check_for_stc/action.yml
```diff
@@ -0,0 +1,18 @@
+name: 'Ensure STC is installed with Pyccel'
+inputs:
+  editable:
+    description: 'editable installation'
```
EmilyBourne, line 4: It might be easier to add a default value for this

mustapha-belbiad: Okay, but what should be the default value? How about using `default: '-e'` or `editable`?

EmilyBourne: I think the default should be the non-editable case as this is used more often

.github/actions/check_for_stc/action.yml
```diff
@@ -0,0 +1,18 @@
...
+    - name: check stc installation
+      shell: python
+      run: |
+        import pyccel, os
+        from os.path import exists, join
+        pyccel_loc = pyccel.__path__[0]
+        print(pyccel_loc)
+        assert(exists(join(pyccel_loc,'extensions/STC/include/stc')))
+        tmp = "${{ inputs.editable }}"
+        if tmp is None:
```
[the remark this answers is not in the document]
mustapha-belbiad: Fixed : 8007dce

.github/actions/check_for_stc/action.yml
```diff
@@ -0,0 +1,18 @@
...
+    - name: check stc installation
+      shell: python
+      run: |
+        import pyccel, os
+        from os.path import exists, join
+        pyccel_loc = pyccel.__path__[0]
+        print(pyccel_loc)
+        assert(exists(join(pyccel_loc,'extensions/STC/include/stc')))
+        if "${{ inputs.editable }}" == "false":
+          assert(not exists(join(pyccel_loc,'extensions/STC/docs')))
```
EmilyBourne, line 18: Why use strings here? Instead of:
```
if "false" == "false":
    assert ...
```
You could simply do:
```
if true:
```
You may need to change the defaults/the argument for this

.github/workflows/pickle.yml
```diff
@@ -87,8 +88,13 @@ jobs:
...
+        with:
+          editable: 'False'
```
EmilyBourne, line 95: This value needs to depend on "editable_string". Currently it is False even if the installation is editable

mustapha-belbiad: Fixed : 12dfa29

.github/actions/check_for_stc/action.yml
```diff
@@ -0,0 +1,18 @@
...
+    - name: check stc installation
+      shell: python
+      run: |
+        import pyccel, os
+        from os.path import exists, join
+        pyccel_loc = pyccel.__path__[0]
+        print(pyccel_loc)
+        assert(exists(join(pyccel_loc,'extensions/STC/include/stc')))
+        if ${{ inputs.editable }}:
+          assert(not exists(join(pyccel_loc,'extensions/STC/docs')))
```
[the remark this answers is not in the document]
mustapha-belbiad: Fixed : 12dfa29

.github/workflows/pickle.yml
```diff
@@ -87,8 +88,13 @@ jobs:
       - name: Check for existence of pickled files
         id: pickle_check
         uses: ./.github/actions/check_for_pickled
+      - name: Check for STC installation
+        id: stc_check
+        uses: ./.github/actions/check_for_stc
+        with:
+          not_editable: 'false'
```
EmilyBourne, line 95: This workflow can be editable or non-editable depending on the contents of `${{ inputs.editable_string }}`. You will need to check this value. If you specify the type of the argument (boolean) in the `check_for_stc` action then you should be able to pass something like `${{ inputs.editable_string == '-e' }}` here

.github/actions/check_for_stc/action.yml
```diff
@@ -13,6 +13,7 @@ runs:
         from os.path import exists, join
         pyccel_loc = pyccel.__path__[0]
         print(pyccel_loc)
+        print("==========================================------------------> ${{ inputs.editable_string }}")
```
EmilyBourne: You don't have access to this here. It is defined in one (but not all) of the workflows that calls this action. It is not an input to this action

.github/workflows/pickle.yml
EmilyBourne: The fix for this should be done in this file

EmilyBourne: >  If you specify the type of the argument

Ah unfortunately this only seems possible with a workflow

EmilyBourne: This may also be useful: https://docs.github.com/en/actions/learn-github-actions/expressions#example

.github/actions/check_for_stc/action.yml
```diff
@@ -0,0 +1,18 @@
...
+    - name: check stc installation
+      shell: python
+      run: |
+        import pyccel, os
+        from os.path import exists, join
+        pyccel_loc = pyccel.__path__[0]
+        print(pyccel_loc)
+        print("${{ inputs.not_editable }}")
+        assert(exists(join(pyccel_loc,'extensions/STC/include/stc')))
+        if "${{ inputs.not_editable }}" != "-e":
```
EmilyBourne, line 17: Ideally I don't think you should mention `-e` in this file as there are probably other ways of handling editable installations so it would be good to use something more general like you had before

.github/workflows/pickle.yml
```diff
@@ -92,7 +92,7 @@ jobs:
         id: stc_check
         uses: ./.github/actions/check_for_stc
         with:
-          not_editable: "${{ matrix.editable_string }}"
+          not_editable: "${{ matrix.editable_string == "-e" && "False" || "True"}}"
```
EmilyBourne, line 95: https://docs.github.com/en/actions/learn-github-actions/expressions#literals
> You don't need to enclose strings in ${{ and }}. However, if you do, you must use single quotes (') around the string. To use a literal single quote, escape the literal single quote using an additional single quote (''). Wrapping with double quotes (") will throw an error.

.github/workflows/docs.yml
```diff
@@ -75,8 +77,8 @@ jobs:
         id: doc_coverage
         run: |
           echo "${{ github.event.pull_request.base.sha }}"
-          docstr-coverage --skip-magic --skip-init --fail-under=0 --docstr-ignore-file=base/.docstr_coverage base/pyccel base/ci_tools 2>&1 | tee base_cov
-          docstr-coverage --skip-magic --skip-init --fail-under=0 --docstr-ignore-file=compare/.docstr_coverage compare/pyccel compare/ci_tools 2>&1 | tee compare_cov
+          docstr-coverage --skip-magic --skip-init --fail-under=0  --exclude=base/pycce/extensions --docstr-ignore-file=base/.docstr_coverage base/pyccel base/ci_tools 2>&1 | tee base_cov
+          docstr-coverage --skip-magic --skip-init --fail-under=0 --exclude=compare/pycce/extensions --docstr-ignore-file=compare/.docstr_coverage compare/pyccel compare/ci_tools 2>&1 | tee compare_cov
```
EmilyBourne, line 81:
```suggestion
          docstr-coverage --skip-magic --skip-init --fail-under=0 --exclude=base/pyccel/extensions --docstr-ignore-file=base/.docstr_coverage base/pyccel base/ci_tools 2>&1 | tee base_cov
          docstr-coverage --skip-magic --skip-init --fail-under=0 --exclude=compare/pyccel/extensions --docstr-ignore-file=compare/.docstr_coverage compare/pyccel compare/ci_tools 2>&1 | tee compare_cov
```

mustapha-belbiad: Fixed : e0c4db5

.github/workflows/docs.yml
```diff
@@ -75,8 +77,8 @@ jobs:
         id: doc_coverage
         run: |
           echo "${{ github.event.pull_request.base.sha }}"
-          docstr-coverage --skip-magic --skip-init --fail-under=0 --docstr-ignore-file=base/.docstr_coverage base/pyccel base/ci_tools 2>&1 | tee base_cov
-          docstr-coverage --skip-magic --skip-init --fail-under=0 --docstr-ignore-file=compare/.docstr_coverage compare/pyccel compare/ci_tools 2>&1 | tee compare_cov
+          docstr-coverage --skip-magic --skip-init --fail-under=0 --exclude=base/pyccel/extensions --docstr-ignore-file=base/.docstr_coverage base/pyccel base/ci_tools 2>&1 | tee base_cov
+          docstr-coverage --skip-magic --skip-init --fail-under=0 --exclude=compare/pyccel/extensions --docstr-ignore-file=compare/.docstr_coverage compare/pyccel compare/ci_tools 2>&1 | tee compare_cov
```
EmilyBourne, line 81: Maybe try:
```suggestion
          docstr-coverage --skip-magic --skip-init --fail-under=0 --exclude=".*base/pyccel/extensions" --docstr-ignore-file=base/.docstr_coverage base/pyccel base/ci_tools 2>&1 | tee base_cov
          docstr-coverage --skip-magic --skip-init --fail-under=0 --exclude=".*compare/pyccel/extensions" --docstr-ignore-file=compare/.docstr_coverage compare/pyccel compare/ci_tools 2>&1 | tee compare_cov
```
so the regex matches the absolute path?

EmilyBourne (approved):

pyproject.toml
```diff
@@ -48,7 +48,23 @@ path = "pyccel/version.py"
...
+  "*.pyccel",
+  "pyccel/stdlib/*/*.h",
+  "pyccel/stdlib/*/*.f90",
+  "*.tx",
+  "*.c"
+]
+exclude = [
+    "misc",
+    "src"
+]
```
yguclu, line 64: Can you please explain how you have determined what to list under `include` and `exclude`?

This is not totally clear to me. For example, I expected that we should write "pyccel/extensions/STC/misc" instead of just "misc"...

yguclu (commented):
Good job! I just have a couple of comments.

pyproject.toml
yguclu: @EmilyBourne Do you happen to know the answer?

EmilyBourne: I'm not sure but I can take a guess. The `exclude` folder also doesn't list everything in STC (e.g. `docs` is not listed) but seems to be sufficient to exclude these folders.

I think that once an `include` section is specified it is used to define all files that should be included using this as a regex. Given that the regex for c files is simply `*.c` this is probably pulling in `misc` (pyccel/extensions/STC/misc/)  which contains `*.c` files and `src` (pyccel/extensions/STC/src/) which contains both `*.c` and  `*.py` files.

I agree that it would be better to use more explicit paths for this though as it should avoid several of these conflicts. Do you want me to fix it?

EmilyBourne: I have run a couple of tests and this explanation seems to line up with what I am seeing so the fix is easy.

EmilyBourne: 96221ceeb

yguclu: OK, yes. Please modify this file to use more explicit paths.

pyproject.toml
```diff
@@ -48,7 +48,23 @@ path = "pyccel/version.py"
...
 
 [tool.hatch.build.targets.wheel.hooks.custom]
 path = "install_scripts/hook.py"
 require-runtime-dependencies = true
+
+
```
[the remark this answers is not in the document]
EmilyBourne: 8ec5f1cc3

yguclu (approved):

Issue #1732, opened 2024-02-14
Pyccel's Code Translation Error: Converts List Index l[o] to l[o,::]


mustapha-belbiad:
### When using Pyccel to process the Python code:

```python
l = [[1, 2, 3], [57, 8, 9]]
l[0]
```
**Pyccel generates incorrect code that alters the intended behavior. 
Instead of accessing the single element at index 0 of the first sublist `l[0]`, the generated code attempts to access a slice of the entire first sublist `l[0,::]`**

 **the Pyccel-generated code:**
```python
l = [[1, 2, 3], [57, 8, 9]]
l[0,::]
```
**However, it's crucial to acknowledge that the syntax lis[0,::] is not valid for list indexing in Python. Typically, list indexing allows accessing elements or sublists using integers (e.g., l[0]) or slices (e.g., l[1:3])**

### Error details
During the execution of the Pyccel-generated code, I get the following error message:
```bash/fortran/c
Traceback (most recent call last):
  File "/home/mbelbiad/Desktop/clear_method/zz/test.py", line 3, in <module>
    Dummy_0000 = lis[0,::]
                 ~~~^^^^^^
TypeError: list indices must be integers or slices, not tuple
```

EmilyBourne:
Does this work for tuples?

mustapha-belbiad:
> Does this work for tuples?

the Pyccel-generated code for tuple:
```python
lis = ([1, 2, 3, 4], [3, 6, 788, 8], [66, 8, 0, 2])

Dummy_0000 = lis[1][::]
```

EmilyBourne:
Interesting. We must be doing something different for tuples then as the generated code is valid for tuples. I wonder if this is related to the rank? If so then this issue is related to #1722

EmilyBourne:
The fix for tuples is here:
https://github.com/pyccel/pyccel/blob/666aa640c8cf4e6253a92ff5f9a2f51534ac7bdf/pyccel/codegen/printing/pycode.py#L295-L296

If we need this fix for a test then we could also add lists as a temporary fix but ideally this should be fixed via the improvements in #1722

PR #1797, opened 2024-03-18
List support: `copy` method

### LINKED ISSUE #1696: copy: It returns a shallow copy of a list.
### Feature Description
Add support for copy() method for lists in the Python translation.

### Test Code
```python
lst = [2, 3, 5]

new_lst = lst.copy()

print(new_lst)
>> [2, 3, 5]
```

### Proposed Solution
See description in https://github.com/pyccel/pyccel/issues/1529

Farouk-Echaref:
Add support for the `copy()` method of Python lists to the semantic stage and Python codegen. This fixes #1696.

**Commit Summary**

- Add a class representation for the `copy()` method, inheriting from `ListMethod`. The constructor copies the attributes of the list object, to the current object that represents the resulting new copy of the list.
- Add check for `HomogeneousListType` in `_print_IndexedElement` to support indexing a multidimensional list.
- Add success and error tests for the `copy()` method involving many scenarios where the function could be used.


mustapha-belbiad (approved):
Looks good

pyccel/ast/builtin_methods/list_methods.py
```diff
@@ -229,3 +230,39 @@ class ListExtend(ListMethod):
...
+    >>> b = a.copy()
+    >>> print(a, b)
+    [[1, 2], [3, 4]]
+    [[1, 2], [3, 4]]
+    >>> a[0][0] = 0
+    >>> a[0][1] = 0
+    >>> print(a, b)
+    [[0, 0], [3, 4]]
+    [[0, 0], [3, 4]]
+
```
EmilyBourne, line 254: Please add a test like this which ensures that we are using pointers

EmilyBourne (changes requested):
Please complete tests and fix conflicts

pyccel/ast/builtin_methods/list_methods.py
Farouk-Echaref: sorry for the late reply, i've added a simple test for this and updated the docstring since we don't have support for accessing an element of a two-dimensional list`a[0][1]`.
6d93c8b16536f51836089b9a1b29313c6c996845

EmilyBourne: Ah, there is a dirty fix and a clean fix. The clean fix is coming in my next monster PR. The quick and dirty fix is to add lists here:
https://github.com/pyccel/pyccel/blob/8c77c1dce1f6c11ee57cd824afed019ca1d17e46/pyccel/codegen/printing/pycode.py#L287-L290

EmilyBourne: Please either use the dirty fix (this will only cause a very minor conflict), or write the test and mark it as xfail for now. It can then be activated in my PR

EmilyBourne (changes requested):
I still think the test should be added. Even if it is marked as `xfail` for now

pyccel/ast/builtin_methods/list_methods.py
Farouk-Echaref: I think I'd go for the dirty fix since it will make it work for now, and I know it will be handled after your PR as well

Farouk-Echaref: updated f532d127a9673862bad5936a99dda106101f5b28

EmilyBourne (approved):
Looks good to me

pyccel/ast/builtin_methods/list_methods.py
EmilyBourne: Please mention the dirty fix in the PR description.

EmilyBourne: I think if the issue is fixed then we should also link https://github.com/pyccel/pyccel/issues/1732

EmilyBourne: I'm not sure about the CHANGELOG though as the fix will only work for lists of lists

yguclu (approved):

pyccel/codegen/printing/pycode.py
```diff
@@ -284,7 +284,7 @@ def _print_IndexedElement(self, expr):
                 indices = indices[0]
 
             indices = [self._print(i) for i in indices]
-            if expr.pyccel_staging != 'syntactic' and isinstance(expr.base.class_type, HomogeneousTupleType):
+            if expr.pyccel_staging != 'syntactic' and isinstance(expr.base.class_type, (HomogeneousTupleType, HomogeneousListType)):
```
EmilyBourne, line 287: Please update the CHANGELOG to mention this change

EmilyBourne (changes requested):

pyccel/codegen/printing/pycode.py
Farouk-Echaref: can you check if the fix description needs more improvements ? 
0df73736f8062f68a843e3565f13bd4558469b9b

pyccel/ast/builtin_methods/list_methods.py
Farouk-Echaref: 0df73736f8062f68a843e3565f13bd4558469b9b
I've updated the PR as well, i can change the descriptions in CHANGELOG and PR if they need improvements

CHANGELOG.md
```diff
@@ -22,6 +23,7 @@ All notable changes to this project will be documented in this file.
 -   Allow printing the result of a function returning multiple objects of different types.
 -   #1792 : Fix array unpacking.
 -   #1795 : Fix bug when returning slices in C.
+-   #1732 : Fix multidimensional list indexing.
```
EmilyBourne, line 26:
```suggestion
-   #1732 : Fix multidimensional list indexing in Python.
```
This is not entirely true, but will be soon

Farouk-Echaref: updated 074745b3043674dbb60b6b6bf5eaa03fa6036df7

EmilyBourne (approved):

PR #1798, opened 2024-03-18
List support: `sort` method

### LINKED ISSUE #1700: sort: sort a list in ascending, descending, or user-defined order.
### Feature Description

Add support for sort() method for lists in the Python translation.

### Test Code

This Code Should Work 

```python
prime_numbers = [11, 3, 7, 5, 2]

prime_numbers.sort()

print(prime_numbers)

>> [2, 3, 5, 7, 11]
```

### Proposed Solution
See description in #1529

Farouk-Echaref:
Add support for the `sort()` method of Python lists to the semantic stage and Python codegen. This fixes #1700 .

**Commit Summary**

- Add a class representation for the `sort()` method, inheriting from `ListMethod`. The constructor checks the passing arguments, and raises an error if the method has an an optional argument.
- Add success and error tests for the `sort()` method involving scenarios where the function could be used.


mustapha-belbiad (approved):
Looks Good.

pyccel/ast/builtin_methods/list_methods.py
```diff
@@ -306,3 +307,41 @@ def __init__(self, list_obj) -> None:
...
+    >>> print(a)
+    [1, 1, 2, 3, 3, 4, 5, 5, 5, 6, 9]
+
+    Parameters
+    ----------
+    list_obj : TypedAstNode
+        The list object which the method is called from.
+
+    reverse : TypedAstNode
+        Optional, reverse=True will sort the list descending. Default is reverse=False.
```
EmilyBourne, line 332:
```suggestion
    reverse : TypedAstNode, default=LiteralFalse
        If True the list is sorted in descending order.
```

pyccel/ast/builtin_methods/list_methods.py
```diff
@@ -306,3 +307,41 @@ def __init__(self, list_obj) -> None:
...
+
+    key : TypedAstNode
+        Optional, a function to specify the sorting criteria(s).
```
EmilyBourne, line 335:
```suggestion
    key : TypedAstNode, optional
        A function to specify the sorting criteria(s).
```

pyccel/ast/builtin_methods/list_methods.py
```diff
@@ -306,3 +307,41 @@ def __init__(self, list_obj) -> None:
...
+    __slots__ = ()
+    _rank = 0
+    _order = None
+    _shape = None
+    _class_type = VoidType()
+    name = 'sort'
+
+    def __init__(self, list_obj, reverse=None, key=None) -> None:
+        if reverse is not None or key is not None:
+            raise TypeError("Optional Parameters are not supported for sort() method.")
```
EmilyBourne, line 346: Please mention this restriction somewhere in the docstring

pyccel/ast/builtin_methods/list_methods.py
EmilyBourne: I suggested a change to the docstring but now that I've seen the body of the `__init__` method I don't think this is correct. The documented default is not the default value for the constructor of this class. Remember you need to document the class not Python's `sort` method.

```suggestion
    reverse : TypedAstNode, optional
        Argument mimicking sort's reverse parameter. This argument is unsupported so
        it should not be provided.
```

tests/errors/semantic/blocking/SORT_LIST_OPTIONAL_PARAM.py
```diff
@@ -0,0 +1,6 @@
+# pylint: disable=missing-function-docstring, missing-module-docstring
+
+
+a = [1, 2]
+a.sort(reverse=True, key=len(a))
```
EmilyBourne, line 5: It would be better to use an example that runs in Python. This code raises:
```
<ipython-input-2-923795b47a06> in <module>
----> 1 a.sort(reverse=True, key=len(a))

TypeError: 'int' object is not callable

```

EmilyBourne (changes requested):

pyccel/ast/builtin_methods/list_methods.py
Farouk-Echaref: updated 35a6df89c347a11eba1dbd61acd0317e55666844

pyccel/ast/builtin_methods/list_methods.py
Farouk-Echaref: updated 35a6df89c347a11eba1dbd61acd0317e55666844

pyccel/ast/builtin_methods/list_methods.py
Farouk-Echaref: updated 35a6df89c347a11eba1dbd61acd0317e55666844

tests/errors/semantic/blocking/SORT_LIST_OPTIONAL_PARAM.py
Farouk-Echaref: fixed c7b4cae7f696ddc73b85fa805cbdebb02a362c4e

pyccel/ast/builtin_methods/list_methods.py
EmilyBourne: Does the second sentence not also apply to the `key` argument?

pyccel/ast/builtin_methods/list_methods.py
```diff
@@ -306,3 +307,43 @@ def __init__(self, list_obj) -> None:
...
+
+#==============================================================================
+class ListSort(ListMethod) :
+    """
+    Represents a call to the .sort() method.
+
+    Represents a call to the `.sort()` method, which sorts the elements of the
+    list in ascending order and modifies the original list in place. This means
+    that the elements of the original list are rearranged to be in sorted order.
+    Optional Parameters are not supported, therefore they should not be provided.
```
EmilyBourne, line 319:
```suggestion
    Optional parameters are not supported, therefore they should not be provided. 
```

pyccel/ast/builtin_methods/list_methods.py
Farouk-Echaref: yes i'll updated in the next commit

Farouk-Echaref: 88ab14fff2fe3ef6756a1cff5887173ba887536f

pyccel/ast/builtin_methods/list_methods.py
```diff
@@ -306,3 +307,44 @@ def __init__(self, list_obj) -> None:
...
+    list_obj : TypedAstNode
+        The list object which the method is called from.
+
+    reverse : TypedAstNode, optional
+        Argument mimicking sort's reverse parameter. This argument is 
+        unsupported so it should not be provided.
+
+    key : TypedAstNode, optional
+        A function to specify the sorting criteria(s). This argument is 
+        unsupported so it should not be provided.
```
EmilyBourne, line 338: If this argument were supported then I imagine `key` would be a `FunctionDef`. Do you agree? Does `FunctionDef` inherit from `TypedAstNode`?

EmilyBourne (approved):

EmilyBourne:
I have approved but I am not entirely convinced that the number of tests is sufficient. Do you think a test with booleans and a test with floats would be useful?

pyccel/ast/builtin_methods/list_methods.py
```diff
@@ -306,3 +307,44 @@ def __init__(self, list_obj) -> None:
...
+    list_obj : TypedAstNode
+        The list object which the method is called from.
+
+    reverse : TypedAstNode, optional
+        Argument mimicking sort's reverse parameter. This argument is 
+        unsupported so it should not be provided.
+
+    key : FunctionDef, optional
+        A function to specify the sorting criteria(s). This argument is 
+        unsupported so it should not be provided.
```
yguclu, line 338: I understand that it is technically complicated to support the `key` argument, but why not supporting `reverse`?

EmilyBourne: There are 2 ways of looking at this:
1. Sort in reverse order
2. Sort, then reverse

In case (1), it is a somewhat similar problem from a technical perspective. @Farouk-Echaref spent some time looking at the STC implementation. I believe he only added what could be supported simply out of the box with STC. We can always increase support later.

In case (2) it should be possible but I don't think that this is the most readable (or the fastest) way to write this.

yguclu: Are we suggesting the user to replace a call to `l.sort(reverse=True)` with these two lines then?
```python
l.sort()
l.reverse()
```
If yes, we should probably write it explicitly in our error message.

EmilyBourne: We did not decide. I was waiting for you to voice an opinion as I don't know if what I wrote:
> I believe he only added what could be supported simply out of the box with STC. We can always increase support later.

is true.

I also don't remember if you have already implemented `reverse`

pyccel/ast/builtin_methods/list_methods.py
```diff
@@ -306,3 +307,44 @@ def __init__(self, list_obj) -> None:
...
+    """
+    __slots__ = ()
+    _rank = 0
+    _order = None
+    _shape = None
+    _class_type = VoidType()
+    name = 'sort'
```
EmilyBourne, line 345: Beware. This code needs changing now that #1808 is merged

pyccel/ast/builtin_methods/list_methods.py
Farouk-Echaref: > I believe he only added what could be supported simply out of the box with STC. We can always increase support later.
 
Yes that is the case, as for reverse it's not yet implemented since it doesn't have a direct equivalent in STC or gFTL, we can add support for `sort()` later as you stated when `reverse()` is implemented.

EmilyBourne: In that case I think that the reply to this question:
> Are we suggesting the user to replace a call to l.sort(reverse=True) with these two lines then?

is no. This is simply something that is not supported yet.

Farouk-Echaref: okey, should I stick with the same error message ?

Farouk-Echaref: tests for sort updated
af66fca9da2dafca479fad70f0e22565ccee9dbf

pyccel/ast/builtin_methods/list_methods.py
Farouk-Echaref: updated 
792d7f449b75eedbcf0170c195307badc732de8f

pyccel/ast/builtin_methods/list_methods.py
Farouk-Echaref: tests for bool and float added da97dfd9d659967daed2651614dfd7deb2e0c8d5

EmilyBourne (approved):
Looks good to me.

tests/epyccel/test_epyccel_lists.py
```diff
@@ -476,6 +476,33 @@ def g():
...
+        a = [True, False, False, True, False]
+        a.sort()
+        return a
+
+    epyc_f = epyccel(f, language=language)
+    assert f() == epyc_f()
+
+def test_sort_float(language):
+    def f():
+        a = [3.4, 1.0, -4,5, 0.0, 2.1]
```
EmilyBourne, line 499:
```suggestion
        a = [3.4, 1.0, -4.5, 0.0, 2.1]
```

Farouk-Echaref: fixed f7ea489c1e010f5432838eed072d9a8a40d3bc2e

pyccel/ast/builtin_methods/list_methods.py
yguclu: All right.

yguclu (approved):
Looks good to me.

PR #1806, opened 2024-03-20
Add support for char as a type annotation

AjaySDwivedi1:
Add the char annotation and created a test. Not sure if I created the test correctly.

### DIFF
--- a/tests/pyccel/test_pyccel.py
+++ b/tests/pyccel/test_pyccel.py
@@ -1189,3 +1189,12 @@ def test_time_execution_flag():
     assert 'Total' in result_lines[-2]
     for l in result_lines[1:-1]:
         assert ' : ' in l
+
+#------------------------------------------------------------------------------
+def test_char:
+    if __name__ == '__main__':
+        string = np.empty((2, 5), dtype='c')
+        string[0] = "HELLO"
+        string[1] = "WORLD"
+        for i in range(5):
+            print(string[0,i], end='')
--- a/pyccel/ast/type_annotations.py
+++ b/pyccel/ast/type_annotations.py
@@ -348,4 +348,5 @@ typenames_to_dtypes = { 'float'   : PythonNativeFloat(),
                         'void'    : VoidType(),
                         '*'       : GenericType(),
                         'str'     : StringType(),
+                        'char'    : CharType(),
                         }


tests/pyccel/test_pyccel.py
```diff
@@ -1189,3 +1189,12 @@ def test_time_execution_flag():
...
         assert ' : ' in l
+
+#------------------------------------------------------------------------------
+def test_char:
+    if __name__ == '__main__':
+        string = np.empty((2, 5), dtype='c')
+        string[0] = "HELLO"
+        string[1] = "WORLD"
+        for i in range(5):
+            print(string[0,i], end='')
```
EmilyBourne, line 1200: Please use the other tests in this file as an example. The workflow is to create a test script to be translated in the scripts folder and then call it from here. The output of the translated program is compared to the output of the Python program. You should be able to use the `compile_pyccel` function for this

tests/pyccel/test_pyccel.py
```diff
@@ -1189,3 +1189,12 @@ def test_time_execution_flag():
     assert 'Total' in result_lines[-2]
     for l in result_lines[1:-1]:
         assert ' : ' in l
+
+#------------------------------------------------------------------------------
+def test_char:
+    if __name__ == '__main__':
+        string = np.empty((2, 5), dtype='c')
```
EmilyBourne, line 1196: You will need to make changes in `pyccel.ast.numpyext` or `pyccel.ast.numpytypes` so that 'c' is recognised.

AjaySDwivedi1:
Okay I think I did the right test. I used the test_bool function as a reference.

tests/pyccel/test_pyccel.py
```diff
@@ -1189,3 +1189,8 @@ def test_time_execution_flag():
     assert 'Total' in result_lines[-2]
     for l in result_lines[1:-1]:
         assert ' : ' in l
+
+#------------------------------------------------------------------------------
+def test_char(language):
+    types = str
+    pyccel_test("scripts/c_arrays.py", language = language, output_dtype = types)
```
EmilyBourne, line 1196: Did you forget to commit the file `scripts/c_arrays.py` ?

AjaySDwivedi1: I did not because it already existed and I didn’t make any chances to it.

EmilyBourne: Ah you are right sorry. It is already tested here:
https://github.com/pyccel/pyccel/blob/b39fe9a411736a833d980674b7f6380278febf0f/tests/pyccel/test_pyccel.py#L751-L755

In that case it seems that you have not finished adding a test as nothing new is tested.

AjaySDwivedi1: Ok, sounds good! Do I need to do anything else for you to approve my PR?

EmilyBourne: You need to:
- add a test/multiple tests that tests the things that the PR is supposed to implement (so a test which translates a file with a char type annotation)
- Ensure all the tests pass (e.g. the Codacy static analysis is currently failing)
- Interact with the bot to get and complete your checklist
- Get the example to work. Your fix is not sufficient to make the following code work:
```python
if __name__ == '__main__':
        string = np.empty((2, 5), dtype='c')
        string[0] = "HELLO"
        string[1] = "WORLD"
        for i in range(5):
            print(string[0,i], end='')
```


The issue you are trying to fix specifically mentions this test: [pyccel/tests/internal/scripts/lapack/ex1.py](https://github.com/pyccel/pyccel/blob/e248f8a67adc06c079629924cec54279d6e7a709/tests/internal/scripts/lapack/ex1.py#L46) and asks that the header file describing the LAPACK library be updated to use `char` instead of `str` where appropriate.

AjaySDwivedi1: Oh sorry I misread the previous message.

Issue #1768, opened 2024-02-29
Remove the `FunctionDef`, `Imports` and `Classes` from the `Module` body in the syntactic stage


said-hadjout:
Instances of `FunctionDef`, `Import`, and `ClassDef` are usually saved in specific attributes of the `Module`, `FunctionDef` and `ClassDef` classes. However at the end of the syntactic stage these objects are not saved into the separate attributes. Rather they are indiscriminately saved into a `CodeBlock`. The nodes are then retrieved and removed from the `CodeBlock` in the semantic stage. This is impractical in certain cases (see #1728), and can cause us to do visit functions twice in the semantics stage to collect these objects and analyse them. We should instead collect them directly in the syntactic stage.

EmilyBourne:
If this is done after #1768 then it should be used to remove the `annotate` argument from `_visit_FunctionDef`

PR #1812, opened 2024-03-29
Remove the FunctionDef, Imports and Classes from CodeBlocks

EmilyBourne:
Remove instances of `FunctionDef`, `Import` and `ClassDef` from `CodeBlock`s. They were previously collected here in the syntactic stage. Fixes https://github.com/pyccel/pyccel/issues/1768 . Use this change to remove the argument `annotate` from `_visit_FunctionDef`.

In order to correctly group the imports, move the recognition of a `if __name__ == '__main__'` statement to the syntactic stage. This allows the code to be simplified. An error is now raised if multiple blocks like this are found and the program is then handled in `_visit_Module` in the semantic stage.

### DIFF
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -1291,11 +1291,11 @@ class Program(ScopedAstNode):
     name : str
         The name used to identify the program (this is used for printing in Fortran).
 
-    variables : list[Variable]
-        A list of the variables that appear in the program.
+    variables : Iterable[Variable]
+        An iterable object containing the variables that appear in the program.
 
-    body : Iterable[PyccelAstNode]
-        An iterable object containing the statements in the body of the program.
+    body : CodeBlock
+        An CodeBlock containing the statements in the body of the program.
 
     imports : Iterable[Import]
         An iterable object containing the imports used by the program.
@@ -1330,11 +1330,11 @@ class Program(ScopedAstNode):
         if not iterable(imports):
             raise TypeError('imports must be an iterable')
 
-        imports = set(imports)  # for unicity
-        imports = tuple(imports)
+        imports = {i : None for i in imports}  # for unicity and ordering
+        imports = tuple(imports.keys())
 
         self._name = name
-        self._variables = variables
+        self._variables = tuple(variables)
         self._body = body
         self._imports = imports
         super().__init__(scope)
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -1279,20 +1279,29 @@ class ModuleHeader(PyccelAstNode):
         return self._module
 
 class Program(ScopedAstNode):
+    """
+    Represents a Program in the code.
 
-    """Represents a Program in the code. A block consists of the following inputs
+    A class representing a program in the code. A program is a set of statements
+    that are executed when the module is run directly. In Python these statements
+    are located in an `if __name__ == '__main__':` block.
 
     Parameters
     ----------
-    variables: list
-        list of the variables that appear in the block.
+    name : str
+        The name used to identify the program (this is used for printing in Fortran).
 
-    body: list
-        a list of statements
+    variables : list[Variable]
+        A list of the variables that appear in the program.
 
-    imports: list, tuple
-        list of needed imports
+    body : Iterable[PyccelAstNode]
+        An iterable object containing the statements in the body of the program.
 
+    imports : Iterable[Import]
+        An iterable object containing the imports used by the program.
+
+    scope : Scope
+        The scope of the program.
     """
     __slots__ = ('_name', '_variables', '_body', '_imports')
     _attribute_nodes = ('_variables', '_body', '_imports')
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -10,7 +10,6 @@ from pyccel.errors.errors     import Errors
 from pyccel.errors.messages   import RECURSIVE_RESULTS_REQUIRED
 
 from pyccel.utilities.stage   import PyccelStage
-from pyccel.utilities.strings import create_incremented_string
 
 from .basic     import PyccelAstNode, TypedAstNode, iterable, ScopedAstNode
 from .builtins  import (PythonEnumerate, PythonLen, PythonMap, PythonTuple,
@@ -107,9 +106,18 @@ __all__ = (
 
 
 class AsName(PyccelAstNode):
-
     """
-    Represents a renaming of a variable, used with Import.
+    Represents a renaming of an object, used with Import.
+
+    A class representing the renaming of an object such as a function or a
+    variable. This usually occurs during an Import.
+
+    Parameters
+    ----------
+    obj : PyccelAstNode or PyccelAstNodeType
+        The variable, function, or module being renamed.
+    target : str
+        Name of variable or function in this context.
 
     Examples
     --------
@@ -120,13 +128,6 @@ class AsName(PyccelAstNode):
     old as new
     >>> AsName(NumpyFull, 'fill_func')
     full as fill_func
-
-    Parameters
-    ==========
-    obj    : PyccelAstNode or PyccelAstNodeType
-             The variable, function, or module being renamed
-    target : str
-             name of variable or function in this context
     """
     __slots__ = ('_obj', '_target')
     _attribute_nodes = ()
@@ -163,10 +164,7 @@ class AsName(PyccelAstNode):
         return self._obj
 
     def __repr__(self):
-        return '{0} as {1}'.format(str(self.name), str(self.target))
-
-    def __str__(self):
-        return '{0} as {1}'.format(str(self.name), str(self.target))
+        return f'{self.name} as {self.target}'
 
     def __eq__(self, string):
         if isinstance(string, str):
@@ -222,10 +220,10 @@ class Duplicate(TypedAstNode):
         return self._length
 
     def __str__(self):
-        return '{} * {}'.format(str(self.val), str(self.length))
+        return f'{self.val} * {self.length}'
 
     def __repr__(self):
-        return '{} * {}'.format(repr(self.val), repr(self.length))
+        return f'{repr(self.val)} * {repr(self.length)}'
 
 class Concatenate(TypedAstNode):
     """
@@ -282,13 +280,6 @@ class Assign(PyccelAstNode):
         In the semantic stage :
           TypedAstNode with the same shape as the lhs.
 
-    status : str, optional
-        If lhs is not allocatable, then status is None.
-        otherwise, status is {'allocated', 'unallocated'}.
-
-    like : Variable, optional
-        Contains the name of the variable from which the lhs will be cloned.
-
     python_ast : ast.Ast
         The ast object parsed by Python's ast module.
 
@@ -309,15 +300,13 @@ class Assign(PyccelAstNode):
     >>> Assign(A[0,1], x)
     IndexedElement(A, 0, 1) := x
     """
-    __slots__ = ('_lhs', '_rhs', '_status', '_like')
+    __slots__ = ('_lhs', '_rhs')
     _attribute_nodes = ('_lhs', '_rhs')
 
     def __init__(
         self,
         lhs,
         rhs,
-        status=None,
-        like=None,
         *,
         python_ast = None
         ):
@@ -325,17 +314,15 @@ class Assign(PyccelAstNode):
             lhs = PythonTuple(*lhs)
         self._lhs = lhs
         self._rhs = rhs
-        self._status = status
-        self._like = like
         super().__init__()
         if python_ast is not None:
             self.set_current_ast(python_ast)
 
     def __str__(self):
-        return '{0} := {1}'.format(str(self.lhs), str(self.rhs))
+        return f'{self.lhs} := {self.rhs}'
 
     def __repr__(self):
-        return '({0} := {1})'.format(repr(self.lhs), repr(self.rhs))
+        return f'{repr(self.lhs)} := {repr(self.rhs)}'
 
     @property
     def lhs(self):
@@ -344,20 +331,6 @@ class Assign(PyccelAstNode):
     @property
     def rhs(self):
         return self._rhs
-
-    # TODO : remove
-
-    @property
-    def expr(self):
-        return self.rhs
-
-    @property
-    def status(self):
-        return self._status
-
-    @property
-    def like(self):
-        return self._like
 
     @property
     def is_alias(self):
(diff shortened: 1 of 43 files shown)


pyccel/parser/semantic.py
```diff
@@ -2071,6 +2071,15 @@ def _visit(self, expr):
...
 
     def _visit_Module(self, expr):
+        for f in expr.funcs:
+            self.insert_function(f)
+
+        for i in expr.imports:
+            self._visit(i)
+
+        for c in expr.classes:
+            self._visit(c)
```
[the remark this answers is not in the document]
EmilyBourne: I agree but I suspect this is already the case. It is very similar to the problem with functions that you fixed. For the moment it seems classes in functions are not supported as they are not mentioned in the arguments
https://github.com/pyccel/pyccel/blob/b39fe9a411736a833d980674b7f6380278febf0f/pyccel/ast/core.py#L2228-L2298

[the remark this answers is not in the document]
EmilyBourne: Actually won't these problems only arise inside class methods? In which case it should also be handled by your fix?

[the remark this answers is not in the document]
EmilyBourne: Ah, no these are annotated directly:
https://github.com/pyccel/pyccel/blob/b39fe9a411736a833d980674b7f6380278febf0f/pyccel/parser/semantic.py#L4347-L4348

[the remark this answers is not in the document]
EmilyBourne: https://github.com/pyccel/pyccel/issues/1814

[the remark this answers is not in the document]
EmilyBourne: As this is already broken I would rather leave these changes for a second PR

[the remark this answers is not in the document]
said-hadjout: ah yes, that is another issue, but I was talking about something like
```python
a = 1
class A:
    x = a
```
This would have worked before but not anymore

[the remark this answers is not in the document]
said-hadjout: ah yes, that is another issue, but I was talking about something like
```python
a = 1
class A:
    x = a
```
This would have worked before but not anymore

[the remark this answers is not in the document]
EmilyBourne: I agree but given that the fix is the same for both I don't know if it is worth fixing here? I can't just visit the classes later otherwise the more common case of:
```python
class A:
    pass

my_a = A()
```
will break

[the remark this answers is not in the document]
EmilyBourne: Or we can put this PR on standby, fix #1814 and then come back to clean it up in this PR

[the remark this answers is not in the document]
said-hadjout: As you prefer, we just need to make sure that we handle those cases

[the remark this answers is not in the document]
EmilyBourne: The issue should ensure we don't forget

pyccel/parser/syntactic.py
```diff
@@ -1107,6 +1122,10 @@ def _visit_For(self, stmt):
         iterable = self._visit(stmt.iter)
         body = self._visit(stmt.body)
 
+        if any(isinstance(l, (FunctionDef, ClassDef, Import)) for l in body):
+            errors.report("Unexpected object in a while block (function, class or import)",
+                    symbol=stmt, severity='error')
+
```
said-hadjout, line 1128: It makes more sense to have the checks in the semantics stage, where we raise this error in
`_visit_FunctionDef`, `_visit_ClassDef`, and `_visit_Import` if the methods are called, and it is more efficient,
because here we go over the body every time to check for a special case, contrary to the other method which will not do the check unless those methods are called.

pyccel/parser/syntactic.py
```diff
@@ -1251,6 +1270,9 @@ def _visit_While(self, stmt):
 
         test = self._visit(stmt.test)
         body = self._visit(stmt.body)
+        if any(isinstance(l, (FunctionDef, ClassDef, Import)) for l in body):
+            errors.report("Unexpected object in a while block (function, class or import)",
+                    symbol=stmt, severity='error')
```
said-hadjout, line 1275: Similar to the other case above

pyccel/parser/syntactic.py
```diff
@@ -1298,6 +1320,11 @@ def _visit_With(self, stmt):
         if len(domain) == 1:
             domain = domain[0]
         body = self._visit(stmt.body)
+
+        if any(isinstance(l, (FunctionDef, ClassDef, Import)) for l in body):
+            errors.report("Unexpected object in a while block (function, class or import)",
+                    symbol=stmt, severity='error')
```
said-hadjout, line 1326: There is also the case of an If Block
```python
if cond1:
    import file1
else:
    import fil2
```
This should not be allowed

pyccel/parser/syntactic.py
EmilyBourne: But if we want to have these in the semantic stage then we either have to add a `functions` argument to `With`, `While` and `For` or we need to add an inheritance check in `_visit_FunctionDef`, `_visit_ClassDef`, and `_visit_Import`. I suspect such a check might be relatively complex.

The simplest solution I can think of is to check the `direct_user_nodes` and exclude acceptable enclosing nodes (`Module`/`FunctionDef`/`FunctionCall`). Given the speed of `any` I am not convinced that this would lead to faster runtimes. Do you see a simpler check?

EmilyBourne: Fixed : c40da083d

EmilyBourne: I'm not a huge fan of the idea of creating a `handle_functiondef` which completely replaces `_visit_FunctionDef` as it means we no longer use the function `_visit` which does a few extra commonly forgotten steps like updating `self.current_ast_node` and setting the ast of the resulting object

said-hadjout: what does `get_direct_user_nodes` do? does it give the parent nodes or what exactly?

EmilyBourne: Improved docs: eb76720f1

said-hadjout: Okay, it seems that the cost of this function is O(1) as most objects have one user node, so it should be fine

EmilyBourne: Most objects except the only ones we are applying it to. A `FunctionDef` usually has lots of user nodes as it is used whenever it is called

EmilyBourne: But I don't think that this loop should be worse than the one in `For`/`While`/`With`

said-hadjout: Okay, it should be fine

pyccel/parser/syntactic.py
EmilyBourne: Now I have caught ifs. But this means that I am also catching `if __name__ == '__main__'`. Do you think we should disallow putting imports here or do I need to use a cleverer check?

saidctb: Maybe we should create another class called `MainIf` or something similar to catch that special case

saidctb: I'm not sure what to do if the user does something like
```python
if __name__ == '__main__':
    import file1
else:
    import file2

Comment (EmilyBourne) [reply pyccel/parser/syntactic.py]: In the semantic stage it is caught via
https://github.com/pyccel/pyccel/blob/b39fe9a411736a833d980674b7f6380278febf0f/pyccel/ast/core.py#L4465-L4473 

and turned into an instance of `Program`. But here we are checking the syntactic `expr` so we don't have that information yet

Comment (EmilyBourne) [reply pyccel/parser/syntactic.py]: I would prefer to disallow the condition:
```
if __name__ != '__main__':
```
I don't see a use case that would make it worth handling this properly

Comment (EmilyBourne) [reply pyccel/parser/syntactic.py]: We should probably ensure an error is raised in this case though

Comment (EmilyBourne) [reply pyccel/parser/syntactic.py]: For me it raises:
```
pyccel:
 |fatal [semantic]: tmp_sum.py [2,11]| Undefined variable (np)
```
Which seems like a very valid complaint as `f()` won't work as an import.

Comment (EmilyBourne) [reply pyccel/parser/syntactic.py]: https://peps.python.org/pep-0008/#imports

> Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants.

Comment (EmilyBourne) [reply pyccel/parser/syntactic.py]: I'm not sure how I feel about that :sweat_smile: . At least this should also work for imports I guess

Comment (saidctb) [reply pyccel/parser/syntactic.py]: Yes, it is the same thing as
```python
def f():
    return x+1
x = 1
```
or 
```python
def f():
    return x+1
from file1 import x
``

Comment (saidctb) [reply pyccel/parser/syntactic.py]: Maybe we should stick with the best practice

Comment (EmilyBourne) [reply pyccel/parser/syntactic.py]: So keep the error here and just move the import in the failing tests

Comment (EmilyBourne) [reply pyccel/parser/syntactic.py]: Fixed a8c984d02

Comment (EmilyBourne) [reply pyccel/parser/syntactic.py]: Not sure how I pushed this commit on its own. Here is the commit on this branch

File: tests/pyccel/scripts/hope_benchmarks/hope_pairwise_python.py
Comment (yguclu) [inline tests/pyccel/scripts/hope_benchmarks/hope_pairwise_python.py:1]: I saw a long conversation with @saidctb about this, but I am not convinced of the solution. (Sorry! 😅 )

NumPy's `zeros` is only used by the main program, hence to me it makes sense that we should allow the import inside the if block with condition `__name__ == '__main__':`. Maybe we can just require that no additional code can follow this specific if block?

Comment (EmilyBourne) [reply tests/pyccel/scripts/hope_benchmarks/hope_pairwise_python.py:1]: The main problem is that this if block isn't identified as being the Program until the semantic stage. So in the syntactic stage either we disallow putting imports here (as recommended by PEP) or we allow imports in all if blocks

Comment (EmilyBourne) [reply tests/pyccel/scripts/hope_benchmarks/hope_pairwise_python.py:1]: Maybe it is possible to collect imports at the syntactic stage in a similar way to what is done in the semantic stage :thinking:

Comment (yguclu) [reply tests/pyccel/scripts/hope_benchmarks/hope_pairwise_python.py:1]: If the syntactic stage cannot identify the if block as being the Program, it should probably not complain about imports placement...

Comment (EmilyBourne) [reply tests/pyccel/scripts/hope_benchmarks/hope_pairwise_python.py:1]: Hmm, the trouble is, if I collect the imports in the way I was thinking they will end up in the module scope not the program scope anyway. Do you think we should make an exception for this if block and identify it in the syntactic stage to allow us to organise the code properly?

Comment (EmilyBourne) [reply tests/pyccel/scripts/hope_benchmarks/hope_pairwise_python.py:1]: While waiting for an answer I have implemented my suggestion as a PoC: [98a75c0](https://github.com/pyccel/pyccel/pull/1812/commits/98a75c0b43606e32f38139fc9a1497086df3c8e9) [3f33f8a](https://github.com/pyccel/pyccel/pull/1812/commits/3f33f8a5aba3ad6f47180a86c2ab1238b505f415)

If you think these changes seem reasonable then I can go ahead and clean up the tests

Comment (EmilyBourne) [reply tests/pyccel/scripts/hope_benchmarks/hope_pairwise_python.py:1]: Hmm, it seems I still need to clean up my PoC further, but it would be useful to know if this seems like a sensible direction to work in?

Comment (saidctb) [reply tests/pyccel/scripts/hope_benchmarks/hope_pairwise_python.py:1]: I think it is fine if the imports are in the module scope because this works in Python
```python
def f():
    print(x)
    
if __name__ == '__main__':
    x = 1
    f()
```
At the moment Pyccel doesn't allow this.
But you get an error if you do this in Python
```python
from file import f
f()
NameError                                 Traceback (most recent call last)
Cell In [2], line 1
----> 1 f()

File ~/file.py:2, in f()
      1 def f():
----> 2     print(x)

NameError: name 'x' is not defined
``

Comment (yguclu) [reply tests/pyccel/scripts/hope_benchmarks/hope_pairwise_python.py:1]: What exactly are you trying to do? I looked at your changes but I am not sure I get it...

Comment (EmilyBourne) [reply tests/pyccel/scripts/hope_benchmarks/hope_pairwise_python.py:1]: @saidctb I'm not sure I understand why your example without imports shows why imports should belong in the module scope? I do think it makes sense to keep the imports in the module to a minimum. You don't always need everything that you need for the program in the module. If you do want it in the module then the solution that we discussed should be sufficient.

@yguclu
> What exactly are you trying to do? 

I am trying to do what I suggested above:
> we should make an exception for this if block and identify it in the syntactic stage

So recognise the `Program` in the syntactic stage ([98a75c0](https://github.com/pyccel/pyccel/pull/1812/commits/98a75c0b43606e32f38139fc9a1497086df3c8e9)) . This then allows us to give it it's own scope and handle it properly in the semantic stage

Comment (yguclu) [reply tests/pyccel/scripts/hope_benchmarks/hope_pairwise_python.py:1]: @saidctb In your example the function `f` cannot be imported, and I agree with you that we should not support a situation like that. My point was just that the program usually performs some additional operations and for this reason it may need to import a module that is not needed otherwise. Accordingly, in my scripts I find it clearer to separate the "module" imports from the "program" imports.

Comment (saidctb) [reply tests/pyccel/scripts/hope_benchmarks/hope_pairwise_python.py:1]: > @saidctb I'm not sure I understand why your example without imports shows why imports should belong in the module scope? I do think it makes sense to keep the imports in the module to a minimum. You don't always need everything that you need for the program in the module. If you do want it in the module then the solution that we discussed should be sufficient.

I'm trying to show that everything defined in the Program is accessible to the Module in Python,
maybe this is clearer 
```python
def f():
    print(pi)
if __name__ == '__main__':
    from numpy import pi
    f()
```

yguclu: @EmilyBourne Yes, making an exception for this if block makes a lot of sense to me. It is not exactly a "syntactic" operation, but we must be pragmatic 😅

pyccel/ast/core.py
```diff
@@ -1278,20 +1279,29 @@ def module(self):
...
 
-    body: list
-        a list of statements
+    body : Iterable[PyccelAstNode]
+        An iterable object containing the statements in the body of the program.
 
-    imports: list, tuple
-        list of needed imports
+    imports : Iterable[Import]
+        An iterable object containing the imports used by the program.
```
yguclu, line 1301: This is not strictly related to this PR, but it is worth addressing here nonetheless:

- Do we need `variables` to be a list, given that we only check that it is an iterable below?
- In the `__init__` we store `_imports = tuple(set(imports))` as a tuple, but we set `_variables = variables` and `_body = body`, which is potentially dangerous. I suggest to convert them to the proper type (probably a list).

tests/pyccel/scripts/runtest_multiple_results.py
```diff
@@ -5,35 +5,51 @@
...
+    x = 0
+    y = 1
+    print_func(x,y)
+    return x,y
+
 if __name__ == '__main__':
+    a,b = const_int_float()
+    print(a)
+    print(b)
+
```
yguclu, line 42: PEP 8 spacing:

```suggestion

@pure
def expr_complex_int_bool(n : 'int'):
    return 0.5+n*1j, 2*n, n==3


def f3(x  : 'float' =  1.5, y  : 'float' =  2.5):
    return x+y, x-y


def print_multiple():
    print(f3())


def f4(x : 'float', y  : 'float' =  2.5):
    x = x + y
    return x+y, x-y


def print_func(x : int, y : int):
    print(x,y)


def test_issue910():
    x = 0
    y = 1
    print_func(x,y)
    return x,y


if __name__ == '__main__':
    a,b = const_int_float()
    print(a)
    print(b)

```

pyccel/ast/core.py
EmilyBourne: af892c8a96ed163fbfacbd5e2c16dfb315679e57

yguclu (approved):

PR #1817, opened 2024-04-02
Improve Windows install

EmilyBourne:
The windows installation uses a `sleep` command because we were unsure how to wait for the executable to finish and also prevent it from asking the user for input. This probably leads to the CI being slower than necessary and definitely occasionally leads to the CI crashing when the executable takes longer than we hoped. This PR is derived from #1180 where I found the necessary commands to wait for the execution.


smazouz42 (approved):
nicely refactored using Start-Process offers better control and readability. Approved

yguclu (approved):

PR #1824, opened 2024-04-09
Filter out cast warnings in cast tests

EmilyBourne:
Mac in particular is bad at handling warnings in pytest. #1762 introduced lots of warnings as it is testing casting between types. Such behaviour often raises warnings in NumPy. This PR deactivates those warnings for these specific tests thus speeding up the mac tests by ~x2


yguclu (approved):
Nice solution!

PR #1829, opened 2024-04-11
Add a `pyccel.lambdify` function with similar functionality to `sympy.lambdify`

### LINKED ISSUE #1827: Add a pyccel.lambdify function with similar functionality to sympy.lambdify
### Relevant Discussion

Discussed with @yguclu on zoom

### Describe the feature
We would like to provide a `lambdify` function similar to sympy's `lambdify` function.

### Test Code
```python
        domain   = ((0,1), (0,1))
        periodic = (False, False)
        mapping  = Mapping('F', dim=2).get_callable_mapping()

        from sympy import symbols, sin, cos, pi, lambdify
        x,y   = symbols('x y')
        phi_e = sin(mx * pi * x) * sin(my * pi * y)
        rho_e = -phi_e.diff(x, 2) - phi_e.diff(y, 2)

        templates = {'T', ['float', 'float[:]', 'float[:,:]']}
        phi = lambdify({x : 'T', y : 'T'}, phi_e, float, templates)
        rho = lambdify({x : 'T', y : 'T'}, rho_e, float, templates)

        return Poisson2D(domain, periodic, mapping, phi, rho)
```

### Proposed Solution
The lambdify function should have the following prototype:
```python
def lambdify(expr : sp.Expr, args : dict[sp.Symbol, str], result_type : str = None, templates : dict[str,list[str]] = None):
```

it can use sympy's `NumPyPrinter` to generate code which is decorated using the types specified in the dictionary(ies) passed as argument. This code can then by translated by calling `epyccel`

EmilyBourne:
Add a `pyccel.lambdify` function with similar functionality to `sympy.lambdify`. Given a SymPy expression `f` and type annotations, `lambdify` returns a "pyccelised" function `f_fast` that can be used in the same Python session. Fixes #1827.


EmilyBourne:
/bot run docs coverage pyccel_lint pylint spelling

pyccel/commands/lambdify.py
```diff
@@ -0,0 +1,81 @@
...
+from packaging.version import parse
+
+from pyccel.epyccel import epyccel
+from pyccel.utilities.strings import random_string
+
+if parse(sp.__version__) >= parse('1.8'):
+    from sympy.printing.numpy import NumPyPrinter
+else:
+    from sympy.printing.pycode import NumPyPrinter
+
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
Tested for the paper [10.1016/j.jcp.2023.112249](https://doi.org/10.1016/j.jcp.2023.112249). This ensures support for older versions of sympy.

pyccel/commands/lambdify.py
```diff
@@ -0,0 +1,81 @@
...
+        signature = f'def {func_name}({args}):'
+    if templates:
+        decorators = '\n'.join(f'@template({key}, ['+', '.join(f'"{annot}"' for annot in annotations)+']' \
+                for key, annotations in templates.items())
+    else:
+        decorators = ''
+    code = f'    return {expr}'
+    numpy_import = 'import numpy\n'
+    func = '\n'.join((numpy_import, decorators, signature, code))
+    print(func)
```
yguclu, line 78: Missing debug print?

It could actually be useful to point the user to the generated function if a `verbose` mode was activated. But this is probably a general improvement for Pyccel.

EmilyBourne: Yes indeed. Fixed : 2aca66949

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ All notable changes to this project will be documented in this file.
 -   #1750 : Add Python support for set method `remove()`.
 -   #1787 : Ensure `STC` is installed with Pyccel.
 -   #1743 : Add Python support for set method `discard()`.
+-   #1830 : Add a `pyccel.lambdify.lambdify` function to accelerate SymPy expressions.
```
yguclu, line 17: According to the quickstart file this function is just `pyccel.lambdify` without any repetition.

EmilyBourne: Ah, this should read `pyccel.commands.lambdify.lambdify` because I moved the file. I wanted to reference the code but I did https://github.com/pyccel/pyccel/blob/f9753b6e87a815dd8d34d7280d3e7db65e5ab350/pyccel/__init__.py#L2

to allow the users to simply do `from pyccel import lambdify`. By the way I think we should do this for `epyccel` too. It always seems strange to write `from pyccel.epyccel import epyccel` rather than just `from pyccel import epyccel`. What do you think? I can open an issue if you want

pyccel/epyccel.py
```diff
@@ -277,8 +284,10 @@ def epyccel( python_function_or_module, **kwargs ):
 
     Parameters
     ----------
-    python_function_or_module : function | module
+    python_function_or_module : function | module | str
         Python function or module to be accelerated.
+        If a string is passed then it is assumed to be the code from a module which
+        should be accelerated..
```
yguclu, line 290: Probably worth mentioning: this code must include any necessary `import` statements.

docs/quickstart.md
```diff
@@ -453,6 +453,36 @@ Out[9]: 210.99245283018868
...
+expr = x**2 + x*5
+f = lambdify(expr, {x : 'float'})
+print(f(3.0))
+
+expr2 = x-x
+f2 = lambdify(expr, {x : 'float'}, float)
+print(f2(3.0))
+
+expr = x**2 + x*5 + 4.5
+f3 = lambdify(expr, {x : 'T'}, float, {'T': ['float', 'float[:]', 'float[:,:]']})
```
yguclu, line 475: Here the return statement is ignored in the case of an input array. I feel that this is a bit confusing.

pyproject.toml
```diff
@@ -21,6 +21,7 @@ dependencies = [
     "termcolor >= 1.0.0",
     "textx >= 2.2",
     "astunparse >= 1.6.0", # astunparse is only needed for Python3.8, we should use ast.unparse when we drop Python3.8.
+    "packaging",
```
yguclu, line 24: Where is this module used?

docs/quickstart.md
EmilyBourne: Oops. I meant to put `T`

pyproject.toml
EmilyBourne: Here:
https://github.com/pyccel/pyccel/blob/f9753b6e87a815dd8d34d7280d3e7db65e5ab350/pyccel/commands/lambdify.py#L6-L14

pyccel/epyccel.py
EmilyBourne: 36894f12a29a880b655b65aff24affd404b7abb2

docs/quickstart.md
EmilyBourne: Actually `'T'` doesn't seem to be possible. I have opened an issue https://github.com/pyccel/pyccel/issues/1835 and corrected the example: [bc3d3f9](https://github.com/pyccel/pyccel/pull/1829/commits/bc3d3f94068cd93fc81532193196ea2843b1d960)

CHANGELOG.md
yguclu: Yes, I agree! The user will only import a few things from Pyccel, so we should make them easy to find.

pyproject.toml
yguclu: I see! Thanks

pyccel/commands/lambdify.py
```diff
@@ -0,0 +1,89 @@
...
+import sympy as sp
+from packaging.version import parse
+
+from pyccel.epyccel import epyccel
+from pyccel.utilities.strings import random_string
+
+if parse(sp.__version__) >= parse('1.8'):
+    from sympy.printing.numpy import NumPyPrinter
+else:
+    from sympy.printing.pycode import NumPyPrinter
```
yguclu, line 14: May I suggest to avoid the standalone "parse" function? We already have enough parsers in Pyccel 😉 
For example:

```suggestion
import sympy as sp
from packaging import version

from pyccel.epyccel import epyccel
from pyccel.utilities.strings import random_string

if version.parse(sp.__version__) >= version.parse('1.8'):
    from sympy.printing.numpy import NumPyPrinter
else:
    from sympy.printing.pycode import NumPyPrinter
```

docs/quickstart.md
```diff
@@ -462,20 +462,19 @@ import numpy as np
...
 
-x = sp.Symbols('x')
+x = sp.Symbol('x')
 expr = x**2 + x*5
 f = lambdify(expr, {x : 'float'})
 print(f(3.0))
 
 expr2 = x-x
-f2 = lambdify(expr, {x : 'float'}, float)
+f2 = lambdify(expr, {x : 'float'}, 'float')
```
yguclu: Do we accept a proper type as well as a string? e.g.
```python
f2 = lambdify(expr, {x : float}, float)
```

EmilyBourne: No. Theoretically it is possible to retrieve the type name with something like `result_type.__name__` but this doesn't give a complete type description for something like `list[float]` (which shouldn't be used for this function but gives us an idea of how `np.ndarray[int, 2]` will work when NumPy decide on the syntax). We could add a list of acceptable types, but I thought the simplest thing was to just avoid this problem for now

CHANGELOG.md
EmilyBourne: https://github.com/pyccel/pyccel/issues/1836

docs/quickstart.md
yguclu: All right!

pyccel/commands/lambdify.py
```diff
@@ -0,0 +1,89 @@
...
+from pyccel.epyccel import epyccel
+from pyccel.utilities.strings import random_string
+
+if version.parse(sp.__version__) >= version.parse('1.8'):
+    from sympy.printing.numpy import NumPyPrinter
+else:
+    from sympy.printing.pycode import NumPyPrinter
+
+def lambdify(expr : sp.Expr, args : 'dict[sp.Symbol, str]', result_type : str = None,
+             templates : 'dict[str,list[str]]' = None, **kwargs):
```
yguclu, line 17: I feel that we should make `result_type` and `templates` keyword-only arguments, as this will avoid many possible user errors.
```suggestion
def lambdify(expr : sp.Expr, args : 'dict[sp.Symbol, str]', *, result_type : str = None,
             templates : 'dict[str,list[str]]' = None, **kwargs):
```
What do you think?

yguclu (changes requested):
Excellent job @EmilyBourne! This is a dream feature for me 🤩

pyccel/commands/lambdify.py
EmilyBourne: I agree. This should make it clearer. I will update this and the docs

EmilyBourne: Done : 37a553b3b007690cf8831974ec155173ec8a61ba

yguclu (approved):

yguclu:
Maybe we could expand a bit the PR description and say two words about the new function?

EmilyBourne:
> Maybe we could expand a bit the PR description and say two words about the new function?

I'm surprised the bot even let me mark the PR as ready with such a short description. I have fixed it

yguclu:
> > Maybe we could expand a bit the PR description and say two words about the new function?
> 
> I'm surprised the bot even let me mark the PR as ready with such a short description. I have fixed it

I actually removed the first sentence as it was just a copy-paste of the title 😉

PR #1831, opened 2024-04-11
Fix missing allocation when returning an annotated array expression

EmilyBourne:
Fix missing allocation when returning an annotated array expression by ensuring a syntactic expression is passed to `_visit_Assign`

### DIFF
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -3855,6 +3855,9 @@ class SemanticParser(BasicParser):
         for o,r in zip(return_objs, results):
             v = o.var
             if not (isinstance(r, PyccelSymbol) and r == (v.name if isinstance(v, Variable) else v)):
+                # Create a syntactic obect to visit
+                if isinstance(v, Variable):
+                    v = PyccelSymbol(v.name)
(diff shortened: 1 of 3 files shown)


pyccel/parser/semantic.py
```diff
@@ -3855,6 +3855,9 @@ def _visit_Return(self, expr):
         for o,r in zip(return_objs, results):
             v = o.var
             if not (isinstance(r, PyccelSymbol) and r == (v.name if isinstance(v, Variable) else v)):
+                # Create a syntactic obect to visit
```
yguclu, line 3858: Typo: obect -> object

yguclu:
Do I understand correctly that the array allocation was missing only if the return type was specified?

EmilyBourne:
> Do I understand correctly that the array allocation was missing only if the return type was specified?

Yes. The allocation is missing only if:
1. The return type is specified
2. The object being returned is an expression

pyccel/parser/semantic.py
EmilyBourne, line 3858:
```suggestion
                # Create a syntactic object to visit
```

yguclu (approved):

PR #1837, opened 2024-04-16
Clean infer_type function

### LINKED ISSUE #1842: Tuple incorrectly identified as inhomogeneous
### Describe the bug
Tuples are occasionally misidentified as being inhomogeneous. This prevents them from being indexed.

### To Reproduce
```python
def homogeneous_tuple_containing_var():
    elem = 4
    ai = (1, elem, 5)
    i = 1
    return ai[0], ai[i], ai[2]
```

### Error details
```
pyccel:
 |fatal [semantic]: tmp_idx.py [5,18]| Inhomogeneous tuples must be indexed with constant integers for the type inference to work (ai)
```

### Additional context
Found while cleaning up tuples in #1838

EmilyBourne:
The function `infer_type` is critical to the `SemanticParser` but it has several issues which are fixed in this PR. Firstly the dictionary that is returned is modified to ensure that all the documented fields are defined. Fixes #1842 . This removes the need for calls to `dict.get` elsewhere in the code. Secondly the `is_target` property is removed. This property is either set to False or is incorrectly set. It is not used for calculations elsewhere.

A test is added for #1842. Further the tuple tests are updated so that all objects that are supposed to be homogeneous are indexed by a variable (this is not possible for inhomogeneous objects).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ All notable changes to this project will be documented in this file.
 -   #1218 : Fix bug when assigning an array to a slice in Fortran.
 -   #1830 : Fix missing allocation when returning an annotated array expression.
 -   #1821 : Ensure an error is raised when creating an ambiguous interface.
+-   #1842 : Fix homogeneous tuples incorrectly identified as inhomogeneous.
 
 ### Changed
 -   #1720 : functions with the `@inline` decorator are no longer exposed to Python in the shared library.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ All notable changes to this project will be documented in this file.
 -   \[INTERNALS\] Stop storing `FunctionDef`, `ClassDef`, and `Import` objects inside `CodeBlock`s.
 -   \[INTERNALS\] Remove the `order` argument from the `pyccel.ast.core.Allocate` constructor.
 -   \[INTERNALS\] Remove `rank` and `order` arguments from `pyccel.ast.variable.Variable` constructor.
+-   \[INTERNALS\] Ensure `SemanticParser.infer_type` returns all documented information.
 
 ### Deprecated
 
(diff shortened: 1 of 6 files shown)


yguclu (approved):
Looks good!

Could you elaborate on the bug that is fixed? Is it something that the user can notice?

EmilyBourne:
> Looks good!
> 
> Could you elaborate on the bug that is fixed? Is it something that the user can notice?

In this code:
```
def tuples_as_indexing_var():
    from numpy import ones
    x = ones((2,3,2))
    for z in range(2):
        for y in range(3):
            for w in range(2):
                x[z,y,w] = w+y*2+z*6
    idx_0 = 1
    idx = (1,idx_0,0)
    return x[idx]
```
`idx` is correctly identified as a `tuple[int,...]` but it is saved into an `InhomogeneousTupleVariable`. This happens because `infer_type` populates `is_target` for the variable but not for the literals.

I haven't checked but this probably means that the user cannot do `idx[i]` they should see an error claiming that `idx` is inhomogeneous.

yguclu:
> > Looks good!
> > Could you elaborate on the bug that is fixed? Is it something that the user can notice?
> 
> In this code:
> 
> ```
> def tuples_as_indexing_var():
>     from numpy import ones
>     x = ones((2,3,2))
>     for z in range(2):
>         for y in range(3):
>             for w in range(2):
>                 x[z,y,w] = w+y*2+z*6
>     idx_0 = 1
>     idx = (1,idx_0,0)
>     return x[idx]
> ```
> 
> `idx` is correctly identified as a `tuple[int,...]` but it is saved into an `InhomogeneousTupleVariable`. This happens because `infer_type` populates `is_target` for the variable but not for the literals.
> 
> I haven't checked but this probably means that the user cannot do `idx[i]` they should see an error claiming that `idx` is inhomogeneous.

I have just checked and you are totally right. If I add an `idx[i]` expression where `i` is an integer variable, then I get the error
```sh
pyccel:
 |fatal [semantic]: infer_type.py [12,14]| Inhomogeneous tuples must be indexed with constant integers for the type inference to work (idx)
```
Could you please add a unit test of this sort?

EmilyBourne:
@yguclu I have created an issue that I can reference in the CHANGELOG. I have also added a test for this bug and ensured that all objects that are supposed to be homogeneous are indexed by a variable (this is not possible for inhomogeneous objects).

yguclu (approved):

PR #1838, opened 2024-04-16
Remove InhomogeneousTupleVariable

### LINKED ISSUE #1392: COVERAGE : tuples in _visit_Assign Function
### Untested code

The untested code is:
https://github.com/pyccel/pyccel/blob/28aa4a5570ff0948ab354307b062ddc65b9a5118/pyccel/parser/semantic.py#L2573-L2609

The problem was discovered in PR : #1385

This problem needs to be investigated. Either a test should be added which uses this code, or the code should be removed.

### LINKED ISSUE #1583: Deprecate InhomogeneousTupleVariable
A tuple does not need to have its own type of Variable. This distinction makes it impossible to create tuples in classes. Inhomogeneous tuples can be handled using inline class definitions.

EmilyBourne:
Remove the `InhomogeneousTupleVariable` class. Fixes #1583  There is now enough information in the types to make this class unnecessary. Removing it allows inhomogeneous tuples to be used in classes (as it was previously not possible to combine `DottedVariable` and `InhomogeneousTupleVariable`).

**Commit Summary**
- Use the `symbolic_alias` category in a `Scope` to map an indexed element of an inhomogeneous tuple to the equivalent variable.
- Create `_build_X` functions for functions which accept inhomogeneous tuple arguments (created `SemanticParser._build_PythonTupleFunction` and `SemanticParser._build_NumpyArray`). This allows the variables to be collected from the scope during the building process
- Allow a `PyccelUnarySub` to be used as an index (to facilitate indexing with literals)
- Add a `is_literal_integer` function to determine if an expression is a literal integer
- Recognise inhomogeneous tuples via their class type
- Remove `InhomogeneousTupleVariable`
- Ensure `IndexedElement` can handle inhomogeneous tuple objects.
- Don't declare inhomogeneous tuples (their elements are now already present in the scope)
- Clarify use of dtype/class type in `FCodePrinter._print_Declare`
- Correct type annotation printing in Python (quotes are no longer present to be removed)
- Raise a warning if trying to wrap tuples
- Clean up tuple handling in `_visit_Assign`. Fixes #1392 
- Extract printed inhomogenous tuple variables into `PythonTuple` at the semantic stage
- Increase coverage
- Add a test for inhomogeneous tuples in classes


EmilyBourne:
/bot run coverage pylint pyccel_lint spelling docs

EmilyBourne:
/bot run coverage pylint pyccel_lint spelling docs

Farouk-Echaref (approved):
Looks good and the new docs are well written as well.

jalalium (changes requested):
It seems like it goes into an infinite loop in cases like this:
```
if __name__ == '__main__':
    x = (1,(1,(1,1)))
    print(x)
```

EmilyBourne:
> It seems like it goes into an infinite loop in cases like this:
> 
> ```
> if __name__ == '__main__':
>     x = (1,(1,(1,1)))
>     print(x)
> ```

Fixed : [ad0c057](https://github.com/pyccel/pyccel/pull/1838/commits/ad0c057df55bbe5334ceccf60753880c5392b6bc)
I also added a test: 
68b46d3

jalalium (approved):
Excellent Work!

PR #1839, opened 2024-04-16
Use `PyccelAstNode.pyccel_staging` in `SemanticParser`

EmilyBourne:
`PyccelAstNode`s have a `pyccel_staging` property. This is not used which makes it hard to detect when it is set incorrectly. Separately we often want to create a partially syntactic object (e.g. when unravelling an assignment where the `rhs` has been visited and the `lhs` is an inhomogeneous tuple). This PR uses the `pyccel_staging` property to facilitate the implementation of the latter situation. This is done by exiting `SemanticParser._visit` immediately if the object passed to the function is already a semantic object. As the property is now used it is obvious when it has not been correctly set. These cases are therefore fixed

### DIFF
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -793,9 +793,11 @@ class SemanticParser(BasicParser):
                 else:
                     return self._visit(var[indices[0]][indices[1:]])
             else:
+                pyccel_stage.set_stage('syntactic')
                 tmp_var = PyccelSymbol(self.scope.get_new_name())
                 assign = Assign(tmp_var, var)
                 assign.set_current_ast(expr.python_ast)
+                pyccel_stage.set_stage('semantic')
                 self._additional_exprs[-1].append(self._visit(assign))
                 var = self._visit(tmp_var)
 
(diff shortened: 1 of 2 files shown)


Farouk-Echaref (approved):
looks good, i just want to know in which case exactly this pyccel_staging managed to break the code.

EmilyBourne:
> looks good, i just want to know in which case exactly this pyccel_staging managed to break the code.

@Farouk-Echaref After this PR, if `pyccel_staging` is incorrectly set to semantic the node is not visited. This means that it doesn't contain any type information. Errors quickly follow when you try to do anything with the node

pyccel/parser/semantic.py
```diff
@@ -1933,6 +1948,8 @@ def _visit(self, expr):
         pyccel.ast.basic.PyccelAstNode
             AST object which is the semantic equivalent of expr.
         """
+        if getattr(expr, 'pyccel_staging', 'syntactic') == 'semantic':
+            return expr
```
yguclu, line 1952: Why using `getattr` with a default value if `expr` is of type `PyccelAstNode`, which has the `pyccel_staging` property?

yguclu (changes requested):

pyccel/parser/semantic.py
EmilyBourne: Because sometimes `expr` is a `PyccelSymbol` which isn't a `PyccelAstNode`

yguclu: The docstring says otherwise...

EmilyBourne: Fixed 18c473a6f54633e9e3d52ec60e5a576fee28ca2d

yguclu (approved):

PR #1841, opened 2024-04-16
Remove `pyccel.ast.utilities.builtin_function`

EmilyBourne:
The function `pyccel.ast.utilities.builtin_function` does the same thing as the code here:
https://github.com/pyccel/pyccel/blob/78d56dec8303d728bc9cf43c55464fc42786645f/pyccel/parser/semantic.py#L1064-L1085

however it was not wrapped in a try/except and its implementation leaves no room for specialised `_visit` functions if the class needs more context information. This PR removes the function and uses the existing implementation to retrieve the same behaviour. In order to be able to wrap all builtin classes in a `PyccelFunctionDef` the inheritance is changed so `PythonRange`, `PythonMap`, `PythonEnumerate` and `PythonType` all inherit from `PyccelFunction`. As these objects do not represent objects in memory in the low level code the class type `SymbolicType()` is used.

`PyccelInternalFunction` is renamed to `PyccelFunction`


Farouk-Echaref (approved):
looks good and makes sense to use `SymbolicType()` class type

pyccel/ast/builtins.py
```diff
@@ -1045,7 +1062,7 @@ def __init__(self, *x):
 
         if not x.is_homogeneous:
             types = ', '.join(str(xi.dtype) for xi in x)
-            raise PyccelError("Cannot determine final dtype of 'max' call with arguments of different "
+            raise TypeError("Cannot determine final dtype of 'max' call with arguments of different "
```
yguclu, line 1065: Why have you decided not to use `PyccelError` any more?

pyccel/parser/semantic.py
```diff
@@ -2925,7 +2924,12 @@ def _visit_FunctionCall(self, expr):
         except RuntimeError:
             pass
 
-        func     = self.scope.find(name, 'functions')
+        func = self.scope.find(name, 'functions')
+
+        if func is None:
+            name = str(expr.funcdef)
+            if name in builtin_functions_dict:
+                func = PyccelFunctionDef(name, builtin_functions_dict[name])
```
yguclu, line 2932: Could you explain this change to me? How is this related to the next modification in the same function?

CHANGELOG.md
```diff
@@ -52,6 +52,7 @@ All notable changes to this project will be documented in this file.
 -   \[INTERNALS\] Stop storing `FunctionDef`, `ClassDef`, and `Import` objects inside `CodeBlock`s.
 -   \[INTERNALS\] Remove the `order` argument from the `pyccel.ast.core.Allocate` constructor.
 -   \[INTERNALS\] Remove `rank` and `order` arguments from `pyccel.ast.variable.Variable` constructor.
+-   \[INTERNALS\] All internal classes which can be generated from `FunctionCall`s must inherit from `TypedAstNode`.
```
yguclu, line 55: In the PR description you mention that all built-in classes must inherit from `PyccelInternalFunction`, not just `TypedAstNode`.

pyccel/ast/builtins.py
```diff
@@ -355,6 +355,8 @@ class PythonEnumerate(PyccelAstNode):
     __slots__ = ('_element','_start')
     _attribute_nodes = ('_element','_start')
     name = 'enumerate'
+    _class_type = SymbolicType()
```
yguclu, line 358: I feel that it should be useful to mention in the CHANGELOG that `_class_type = SymbolicType` for the classes `PythonEnumerate`, `PythonMap`, `PythonRange`, and `PythonType`.

yguclu (changes requested):
Looks good! I have a couple of questions and comments.

pyccel/ast/builtins.py
EmilyBourne: In Pyccel we have 3 types of error:
- `PyccelSyntaxError`
- `PyccelSemanticError`
- `PyccelCodegenError`

each of these inherits from `PyccelError`. They are raised by `errors.report` and caught in `codegen/pipeline.py`. Ideally we want to avoid doing `raise PyccelError` directly because it means that the error is caught in pipeline.py without any context (no symbol, no line number, etc).

Before this PR we were using `PyccelError` in builtin functions because there was no try/except call around these calls. After this PR these functions are handled in the same way as imported functions (e.g. `MathCos`). for these functions we catch `TypeErrors` here:
https://github.com/pyccel/pyccel/blob/046e93f1ca42839698de08de6ab08beeb2f0943d/pyccel/parser/semantic.py#L1058-L1065

this allows the correct subclass of error to be raised with the necessary context.

I've just checked and there is now only 1 `PyccelError` which is raised directly. I have removed it too: 65e9698c1

Now that that is done I think it may be worth changing our approach in the future. Instead of raising `TypeError`s we should probably be raising `PyccelError`s

pyccel/parser/semantic.py
EmilyBourne: When we translate a `FunctionCall` such as `math.cos(4.0)` we first search for a function definition in the scope. We find a `PyccelFunctionDef` and handle this call in `_handle_function`. `_handle_function` does the same thing as `builtin_function`. It splits the `FunctionCallArguments` in to `args` and `kwargs` and calls the class constructor. Unlike `builtin_function` it additionally wraps the constructor call in a try/except so that context clues can be added to the error message. By packing the builtin function found in `builtin_functions_dict` into a `PyccelFunctionDef` we can use the same code to handle built-in Python functions and imported functions

CHANGELOG.md
EmilyBourne: Fixed f8163f2b3

pyccel/ast/builtins.py
EmilyBourne: Done 02c0511bf

pyccel/ast/builtins.py
yguclu: Thanks for the detailed explanation!

I agree that raising `TypeError`s all the time is probably not ideal. This calls for a separate discussion.

CHANGELOG.md
yguclu: Just wondering. Do we still have a reason to call it `PyccelInternalFunction` rather than just `PyccelFunction`?

EmilyBourne: No I don't think so

yguclu: How about making a search & replace in this PR?

EmilyBourne: Done 3c478b38a
and CHANGELOG 7c67d46

yguclu: Great! The PR description also needs updating.

EmilyBourne: Done

yguclu (approved):

pyccel/ast/itertoolsext.py
```diff
@@ -7,18 +7,23 @@
...
+class Product(PyccelFunction):
     """
     Represents a call to itertools.product for code generation.
 
-    arg : list, tuple
+    Represents a call to itertools.product for code generation.
+
+    Parameters
+    ----------
+    args : PyccelAstType
```
yguclu, line 25:
```suggestion
    *args : PyccelAstType
```

yguclu (approved):

PR #1844, opened 2024-04-17
Group and rename_visit functions which receive a FunctionCall 

### LINKED ISSUE #1840: Group _visit functions which receive a FunctionCall
In `SemanticParser._visit_FunctionCall` the following lines are very useful for handling functions which need more context to be correctly called:
https://github.com/pyccel/pyccel/blob/78d56dec8303d728bc9cf43c55464fc42786645f/pyccel/parser/semantic.py#L2930-L2934

However the functions that are called this way work slightly differently to the rest. Rather than receiving an `expr` of the type described in the function name (`_visit_X`) they sometimes receive a `FunctionCall`. These methods should be grouped together in the file to help with documenting this phenomenon. Further they should receive an argument called `func_call` to make this distinction clear.

EmilyBourne:
The functions mentioned in #1840 are grouped together at the end of `semantic.py` (fixes #1840). As these functions work differently to all other `_visit_X` functions they are renamed to `_build_X`. This has multiple benefits:
1. It allows them to be clearly referred to in the docs
2. It allows separate functions for things like `NumpyMatmul` which can be called either via a `FunctionCall` or directly.
3. It ensures that errors are raised if these functions are not documented.

Documentation is added explaining why these functions are necessary.

These functions were previously called from multiple places in the code. This was always done before calling `_handle_function`. It is easy to forget this extra code so I have moved it into `_handle_function` so it cannot be forgotten. Following #1839 the visited `args` are reused to avoid revisiting objects.

**Commit Summary**
- Rename `_visit` functions called from a `FunctionCall` which don't match the documented naming pattern to `_build` functions.
- Update developer documentation to explain the role of `_build` functions.
- Avoid visiting an object twice
- Add `use_build_functions` argument to `_handle_function` to prevent infinite recursion.
- Move all calls to `_build` functions inside `_handle_function` (this ensures that all errors from `FunctionCall`s are correctly caught and handled)
- Correct import type for `sqrt` in `** 0.5`
- Group and document `_build` functions
- Increase coverage


EmilyBourne:
/bot run coverage docs spelling pylint pyccel_lint

EmilyBourne:

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ All notable changes to this project will be documented in this file.
 -   #1787 : Ensure `STC` is installed with Pyccel.
 -   #1656 : Ensure `gFTL` is installed with Pyccel.
 -   #1830 : Add a `pyccel.lambdify.lambdify` function to accelerate SymPy expressions.
+-   #1844 : Added line numbers and code to errors from built-in function calls.
```
yguclu, line 20:
```suggestion
-   #1844 : Add line numbers and code to errors from built-in function calls.
```

pyccel/parser/semantic.py
```diff
@@ -948,12 +948,14 @@ def _handle_function_args(self, arguments):
...
             a = self._visit(arg)
-            if isinstance(a.value, StarredArguments):
-                args.extend([FunctionCallArgument(av) for av in a.value.args_var])
+            val = a.value
+            if isinstance(val, FunctionDef) and not isinstance(val, PyccelFunctionDef) and not val.is_semantic:
+                semantic_func = self._annotate_the_called_function_def(val)
+                a = FunctionCallArgument(semantic_func, keyword = a.keyword, python_ast = a.python_ast)
+
+            if isinstance(val, StarredArguments):
+                args.extend([FunctionCallArgument(av) for av in val.args_var])
```
yguclu, line 958: Could you please explain this change?

yguclu (changes requested):
Looks really good!

pyccel/parser/semantic.py
```diff
@@ -1035,15 +1037,33 @@ def _handle_function(self, expr, func, args, is_method = False):
...
+                if isinstance(expr, DottedName):
+                    pyccel_stage.set_stage('syntactic')
+                    if is_method:
+                        new_expr = DottedName(args[0].value, FunctionCall(func, args[1:]))
+                    else:
+                        new_expr = FunctionCall(func, args)
+                    new_expr.set_current_ast(expr.python_ast)
+                    pyccel_stage.set_stage('semantic')
+                    expr = new_expr
+                return getattr(self, annotation_method)(expr)
```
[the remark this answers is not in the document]
EmilyBourne: Done : 0d29061fd

pyccel/parser/semantic.py
EmilyBourne: Before this change we were visiting the syntactic node twice. The behaviour on the second pass can be slightly different and this *may* cause problems. Further if the argument is complex (e.g. an arithmetic expression) then it could be costly to (unnecessarily) visit twice.

While I was developing the double visitation caused problems however if I revert the change I see that they have now been resolved so this change is no longer necessary. The problem that I was seeing was the following:

If we have:
```python
def f() -> 'float[:]':
    ....

def g(a : 'float[:]', b : int):
    ...

if __name__ == '__main__':
    g(f(), 4)
```
Pyccel translates this to:
```python
def f() -> 'float[:]':
    ....

def g(a : 'float[:]', b : float):
    ...

if __name__ == '__main__':
    Dummy_0000 = f()
    g(Dummy_0000, 4)
```
This is done in `_visit_FunctionCallArgument` via the visit of the value. When we visit a `FunctionCall` (`f()`), if it does not have an `Assign` in its user nodes then we create the assignment and return the `Dummy` variable. This means that `_visit_FunctionCallArgument(f())` returns `Dummy_0000`. Somehow on the second pass it was collecting the semantic `f()` which already had an `Assign` in its user nodes. As a result it was not creating the assign in the call where the result is saved. I don't remember if this meant that the dummy was not allocated or if the function call was simply not printed.

As I say I can no longer reproduce this problem so if you prefer the changes can be reverted. However I think it is better to avoid visiting an object of unknown complexity multiple times. This could be costly in the case of something like:
```python
g(1+b**2+a*(4+c*d))
```

yguclu: Thanks for the explanation. I agree, we should avoid visiting the node twice.

yguclu (approved):

Issue #1659, opened 2024-01-10
[SETS] Support variable declarations in C


EmilyBourne:
### Describe the feature
Minimal printing support should be implemented for lists. It should be possible to declare a list using a type annotation.

### Test Code
Provide code which does not currently work but which should do when this issue is fixed:
```python
def declare_list():
    a : list[int]
```

### Proposed Solution
This feature should be implemented with the https://github.com/stclib/STC/ library. The vector type should be used.

### Target Language
C

### Pre-requisites
- #1654

EmilyBourne:
The syntax `set[int]` is not supported in Python 3.8 (https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html). As a first step Python 3.8 tests should be marked as xfail. An issue should be opened to request support for `List` from the typing module. This is related to #1544

PR #1847, opened 2024-04-22
Support variable declarations in c

### LINKED ISSUE #1655: [LISTS] Support variable declarations in C
### Describe the feature

### Test Code
```python
def declare_list():
    a : list[int]
```

### Proposed Solution
This feature should be implemented with the https://github.com/stclib/STC/ library. The vector type should be used.

### Target Language
C

### Pre-requisites
- #1654

mustapha-belbiad:
This feature pull request adds the appropriate C language equivalent for declaring `list` and `set` containers in Python, as well as freeing their memory, with the help of the STC library. This fixes #1655 and #1659.

**Commit Summary**
- Copy STC files to the destination path using `copy_internal_library()`.
- Change the order of printing the module and the module header in `codegen` to determine the appropriate STC macros.
- Perform the appropriate class method renaming in `_print_Module()` instead of `_print_Module_Header()`.
- Determine the right STC container type and construct its macros, as well as the container variable name (`i_type` macro).
- Generate the deallocation node of a container using the STC method `container_X_drop(&container_var)`.
- Generate the initialization of an STC container in C using the method `c_init()`.
- Add tests to check the initialization of `list` and `set` using different data types.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,9 @@ All notable changes to this project will be documented in this file.
 -   #1844 : Add line numbers and code to errors from built-in function calls.
 -   \[INTERNALS\] Added `container_rank` property to `ast.datatypes.PyccelType` objects.
 -   \[DEVELOPER\] Added an improved traceback to the developer-mode errors for errors in function calls.
--   #1847 : Support `list()` and `set()` declarations in c.
+-   #1655 : \[LISTS\] Support variable declarations in C
+-   #1659 : \[SETS\] Support variable declarations in C
+
 ### Fixed
 
 -   #1720 : Fix Undefined Variable error when the function definition is after the variable declaration.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,10 +19,10 @@ All notable changes to this project will be documented in this file.
 -   #1787 : Ensure `STC` is installed with Pyccel.
 -   #1656 : Ensure `gFTL` is installed with Pyccel.
 -   #1844 : Add line numbers and code to errors from built-in function calls.
+-   #1655 : Add the appropriate C language equivalent for declaring a Python `list` container using the `STC` library.
+-   #1659 : Add the appropriate C language equivalent for declaring a Python `set` container using the `STC` library. 
 -   \[INTERNALS\] Added `container_rank` property to `ast.datatypes.PyccelType` objects.
 -   \[DEVELOPER\] Added an improved traceback to the developer-mode errors for errors in function calls.
--   #1655 : \[LISTS\] Support variable declarations in C
--   #1659 : \[SETS\] Support variable declarations in C
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ All notable changes to this project will be documented in this file.
 -   #1844 : Add line numbers and code to errors from built-in function calls.
 -   \[INTERNALS\] Added `container_rank` property to `ast.datatypes.PyccelType` objects.
 -   \[DEVELOPER\] Added an improved traceback to the developer-mode errors for errors in function calls.
-
+-   #1847 : Support `list()` and `set()` declarations in c.
 ### Fixed
 
 -   #1720 : Fix Undefined Variable error when the function definition is after the variable declaration.
--- a/pyccel/ast/builtins.py
+++ b/pyccel/ast/builtins.py
@@ -1262,7 +1262,6 @@ builtin_functions_dict = {
     'min'      : PythonMin,
     'not'      : PyccelNot,
     'map'      : PythonMap,
-    'set'      : PythonSet,
     'str'      : LiteralString,
     'type'     : PythonType,
     'tuple'    : PythonTupleFunction,
--- a/pyccel/ast/datatypes.py
+++ b/pyccel/ast/datatypes.py
@@ -760,10 +760,10 @@ class HomogeneousListType(HomogeneousContainerType, metaclass = ArgumentSingleto
         self._element_type = element_type
         self._order = 'C' if (element_type.order == 'C' or element_type.rank == 1) else None
         super().__init__()
-    
+
     def __eq__(self, other):
         return isinstance(other, self.__class__) and self._element_type == other._element_type \
-                and self._order == other._order  
+                and self._order == other._order
 
     def __hash__(self):
         return hash((self.__class__, self._element_type, self._order))
--- a/pyccel/ast/datatypes.py
+++ b/pyccel/ast/datatypes.py
@@ -442,6 +442,12 @@ class GenericType(FixedSizeType):
     @lru_cache
     def __add__(self, other):
         return other
+    
+    def __eq__(self, other):
+        return True
+
+    def __hash__(self):
+        return hash(self.__class__)
 
 class SymbolicType(FixedSizeType):
     """
@@ -754,6 +760,13 @@ class HomogeneousListType(HomogeneousContainerType, metaclass = ArgumentSingleto
         self._element_type = element_type
         self._order = 'C' if (element_type.order == 'C' or element_type.rank == 1) else None
         super().__init__()
+    
+    def __eq__(self, other):
+        return isinstance(other, self.__class__) and self._element_type == other._element_type \
+              and self._order == other._order  
+
+    def __hash__(self):
+        return hash((self.__class__, self._element_type, self._order))
 
 class HomogeneousSetType(HomogeneousContainerType, metaclass = ArgumentSingleton):
     """
@@ -776,6 +789,12 @@ class HomogeneousSetType(HomogeneousContainerType, metaclass = ArgumentSingleton
         assert isinstance(element_type, PyccelType)
         self._element_type = element_type
         super().__init__()
+    
+    def __eq__(self, other):
+        return isinstance(other, self.__class__) and self._element_type == other._element_type
+
+    def __hash__(self):
+        return hash((self.__class__, self._element_type))
 
 #==============================================================================
 
--- a/pyccel/ast/datatypes.py
+++ b/pyccel/ast/datatypes.py
@@ -149,6 +149,16 @@ class PyccelType:
     is expected when calling a bitwise comparison operator on objects of these types.
     """
     __slots__ = ()
+    _name = None
+
+    @property
+    def name(self):
+        """
+        Get the name of the pyccel type.
+        
+        Get the name of the pyccel type.
+        """
+        return self._name
 
     def __init__(self): #pylint: disable=useless-parent-delegation
         # This __init__ function is required so the ArgumentSingleton can
@@ -488,16 +498,6 @@ class HomogeneousContainerType(ContainerType):
     This is the case for objects such as arrays, lists, etc.
     """
     __slots__ = ()
-    _name = None
-
-    @property
-    def name(self):
-        """
-        Get the name of the container.
-        
-        Get the name of the container.
-        """
-        return self._name
 
     @property
     def datatype(self):
--- a/pyccel/ast/datatypes.py
+++ b/pyccel/ast/datatypes.py
@@ -442,7 +442,7 @@ class GenericType(FixedSizeType):
     @lru_cache
     def __add__(self, other):
         return other
-    
+
     def __eq__(self, other):
         return True
 
--- a/pyccel/ast/datatypes.py
+++ b/pyccel/ast/datatypes.py
@@ -760,10 +760,10 @@ class HomogeneousListType(HomogeneousContainerType, metaclass = ArgumentSingleto
         self._element_type = element_type
         self._order = 'C' if (element_type.order == 'C' or element_type.rank == 1) else None
         super().__init__()
-
+    
     def __eq__(self, other):
         return isinstance(other, self.__class__) and self._element_type == other._element_type \
-              and self._order == other._order  
+                and self._order == other._order  
 
     def __hash__(self):
         return hash((self.__class__, self._element_type, self._order))
@@ -789,7 +789,7 @@ class HomogeneousSetType(HomogeneousContainerType, metaclass = ArgumentSingleton
         assert isinstance(element_type, PyccelType)
         self._element_type = element_type
         super().__init__()
-    
+
     def __eq__(self, other):
         return isinstance(other, self.__class__) and self._element_type == other._element_type
 
--- a/pyccel/ast/datatypes.py
+++ b/pyccel/ast/datatypes.py
@@ -760,7 +760,7 @@ class HomogeneousListType(HomogeneousContainerType, metaclass = ArgumentSingleto
         self._element_type = element_type
         self._order = 'C' if (element_type.order == 'C' or element_type.rank == 1) else None
         super().__init__()
-    
+
     def __eq__(self, other):
         return isinstance(other, self.__class__) and self._element_type == other._element_type \
               and self._order == other._order  
--- a/pyccel/codegen/pipeline.py
+++ b/pyccel/codegen/pipeline.py
@@ -21,7 +21,6 @@ from pyccel.codegen.codegen        import Codegen
 from pyccel.codegen.utilities      import recompile_object
 from pyccel.codegen.utilities      import copy_internal_library
 from pyccel.codegen.utilities      import internal_libs
-from pyccel.codegen.utilities      import external_libs
 from pyccel.codegen.python_wrapper import create_shared_library
 from pyccel.naming                 import name_clash_checkers
 from pyccel.utilities.stage        import PyccelStage
@@ -338,10 +337,10 @@ def execute_pyccel(fname, *,
 
     # Iterate over the external_libs list and determine if the printer
     # requires an external lib to be included.
-    for ext_lib_name, ext_src_folder in external_libs.items():
-        if ext_lib_name in codegen.get_printer_imports():
-
-            lib_dest_path = copy_internal_library(ext_src_folder, pyccel_dirpath)
+    
+    result = [key for key in codegen.get_printer_imports() if key.startswith('stc')]
+    if result[0]:
+        lib_dest_path = copy_internal_library('stc', pyccel_dirpath)
 
     if convert_only:
         # Change working directory back to starting point
--- a/pyccel/codegen/pipeline.py
+++ b/pyccel/codegen/pipeline.py
@@ -21,6 +21,7 @@ from pyccel.codegen.codegen        import Codegen
 from pyccel.codegen.utilities      import recompile_object
 from pyccel.codegen.utilities      import copy_internal_library
 from pyccel.codegen.utilities      import internal_libs
+from pyccel.codegen.utilities      import external_libs
 from pyccel.codegen.python_wrapper import create_shared_library
 from pyccel.naming                 import name_clash_checkers
 from pyccel.utilities.stage        import PyccelStage
@@ -337,8 +338,10 @@ def execute_pyccel(fname, *,
 
     # Iterate over the external_libs list and determine if the printer
     # requires an external lib to be included.
-    if [key for key in codegen.get_printer_imports() if key.startswith('stc')] :
-        lib_dest_path = copy_internal_library('stc', pyccel_dirpath)
+    for key in codegen.get_printer_imports():
+        lib_name = key.split("/", 1)[0]
+        if lib_name in external_libs:
+            lib_dest_path = copy_internal_library(lib_name, pyccel_dirpath)
 
     if convert_only:
         # Change working directory back to starting point
--- a/pyccel/codegen/pipeline.py
+++ b/pyccel/codegen/pipeline.py
@@ -337,7 +337,6 @@ def execute_pyccel(fname, *,
 
     # Iterate over the external_libs list and determine if the printer
     # requires an external lib to be included.
-    
     if [key for key in codegen.get_printer_imports() if key.startswith('stc')] :
         lib_dest_path = copy_internal_library('stc', pyccel_dirpath)
 
--- a/pyccel/codegen/pipeline.py
+++ b/pyccel/codegen/pipeline.py
@@ -338,8 +338,7 @@ def execute_pyccel(fname, *,
     # Iterate over the external_libs list and determine if the printer
     # requires an external lib to be included.
     
-    result = [key for key in codegen.get_printer_imports() if key.startswith('stc')]
-    if result[0]:
+    if [key for key in codegen.get_printer_imports() if key.startswith('stc')] :
         lib_dest_path = copy_internal_library('stc', pyccel_dirpath)
 
     if convert_only:
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -830,7 +830,7 @@ class CCodePrinter(CodePrinter):
         self._current_module = expr
         for item in expr.imports:
             self.rename_imported_methods(item)
-        for classDef in  expr.classes: 
+        for classDef in  expr.classes:
             class_scope = classDef.scope
             for method in classDef.methods:
                 if not method.is_inline:
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -965,7 +965,7 @@ class CCodePrinter(CodePrinter):
                                   f'#define _{container_type.upper()}',
                                   f'#define i_type {container_type}',
                                   f'#define i_key {container_type[index:]}',
-                                  f'#include "{_ + "/" +container[0]}.h"',
+                                  f'#include "{"extensions/STC/include/"+ _ + "/" +container[0]}.h"',
                                   '#endif\n'))
 
         # Get with a default value is not used here as it is
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -25,7 +25,7 @@ from pyccel.ast.operators import PyccelAdd, PyccelMul, PyccelMinus, PyccelLt, Py
 from pyccel.ast.operators import PyccelAssociativeParenthesis, PyccelMod
 from pyccel.ast.operators import PyccelUnarySub, IfTernaryOperator
 
-from pyccel.ast.datatypes import PythonNativeInt, PythonNativeBool, VoidType
+from pyccel.ast.datatypes import PythonNativeInt, PythonNativeBool, VoidType, GenericType
 from pyccel.ast.datatypes import TupleType, FixedSizeNumericType
 from pyccel.ast.datatypes import CustomDataType, StringType, HomogeneousTupleType, HomogeneousListType, HomogeneousSetType
 from pyccel.ast.datatypes import PrimitiveBooleanType, PrimitiveIntegerType, PrimitiveFloatingPointType, PrimitiveComplexType
@@ -958,15 +958,14 @@ class CCodePrinter(CodePrinter):
         if expr.target:
             dtype = expr.target.pop().name
             if source.startswith('stc/'):
-                dtype_macro = dtype.upper().replace(" ", "_")
                 _,container_type = source.split("/")
-                class_type_macro = import_stc.get(container_type)
-                i_type_arg = f"{container_type}_{dtype.replace(' ', '_')}"
-                return '\n'.join((f'#ifndef _{class_type_macro}_{dtype_macro}',
-                                  f'#define _{class_type_macro}_{dtype_macro}',
-                                  f'#define i_type {i_type_arg}',
-                                  f'#define i_key {dtype}',
-                                  f'#include "{source}.h"',
+                container = container_type.split("_")
+                index = len(container[0]) + 1
+                return '\n'.join((f'#ifndef _{container_type.upper()}',
+                                  f'#define _{container_type.upper()}',
+                                  f'#define i_type {container_type}',
+                                  f'#define i_key {container_type[index:]}',
+                                  f'#include "{_ + "/" +container[0]}.h"',
                                   '#endif\n'))
 
         # Get with a default value is not used here as it is
@@ -1186,10 +1185,10 @@ class CCodePrinter(CodePrinter):
             key = (primitive_type, dtype.precision)
         elif isinstance(dtype, (HomogeneousSetType, HomogeneousListType)):
             container_type = 'hset_' if dtype.name == 'set' else 'vec_'
-            vec_dtype = self.find_in_dtype_registry(dtype.element_type)
+            if not isinstance(dtype.element_type, GenericType):
+                vec_dtype = self.find_in_dtype_registry(dtype.element_type)
             i_type = container_type + vec_dtype.replace(' ', '_')
-            source = 'stc/'+ container_type[:-1]
-            self.add_import(Import(source, Module(vec_dtype, (), ())))
+            self.add_import(Import('stc/' + i_type, Module('stc/' + i_type, (), ())))
             return i_type
         else:
             key = dtype
@@ -1325,7 +1324,6 @@ class CCodePrinter(CodePrinter):
         static = 'static ' if expr.static else ''
 
         declaration = f'{static}{external}{declaration_type} {variable}{init};\n'
-
         return preface + declaration
 
     def function_signature(self, expr, print_arg_names = True):
@@ -2015,14 +2013,10 @@ class CCodePrinter(CodePrinter):
         str
             The generated C code for the container initialization.
         """
-        dtype = self.find_in_dtype_registry(expr.current_user_node.lhs.dtype).replace(" ", "_")
-        container_type = "hset_"if isinstance(expr.class_type, HomogeneousSetType) else "vec_"
-        if (len(expr.args) == 0):
-            return f'{container_var} = c_init({container_type}{dtype},{"{}"});\n'
 
+        dtype = self.find_in_dtype_registry(container_var.lhs.class_type)
         keyraw = '{' + ', '.join([self._print(a) for a in expr.args]) + '}'
-        init = f'{container_var} = c_init({container_type}{dtype}, {keyraw});\n'
-
+        init = f'{container_var.lhs.name} = c_init({dtype}, {keyraw});\n'
         return init
 
     def _print_Return(self, expr):
@@ -2158,6 +2152,7 @@ class CCodePrinter(CodePrinter):
         prefix_code = ''
         lhs = expr.lhs
         rhs = expr.rhs
+
         if isinstance(rhs, FunctionCall) and isinstance(rhs.class_type, TupleType):
             self._temporary_args = [ObjectAddress(a) for a in lhs]
             return prefix_code+'{};\n'.format(self._print(rhs))
@@ -2168,7 +2163,7 @@ class CCodePrinter(CodePrinter):
             return prefix_code+self.arrayFill(expr)
         lhs = self._print(expr.lhs)
         if isinstance(rhs, (PythonList, PythonSet)):
-            return prefix_code+self.init_lists_or_sets(rhs, lhs)
+            return prefix_code+self.init_lists_or_sets(rhs, expr)
         rhs = self._print(expr.rhs)
         return prefix_code+'{} = {};\n'.format(lhs, rhs)
 
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -1187,9 +1187,9 @@ class CCodePrinter(CodePrinter):
             key = (primitive_type, dtype.precision)
         elif isinstance(dtype, (HomogeneousSetType, HomogeneousListType)):
             container_type = 'hset_' if dtype.name == 'set' else 'vec_'
-            vec_dtype = self.find_in_dtype_registry(dtype.element_type)
-            i_type = container_type + vec_dtype.replace(' ', '_')
-            self.add_import(Import('stc/' + i_type + "/" + vec_dtype, Module('stc/' + i_type, (), ())))
+            element_type = self.find_in_dtype_registry(dtype.element_type)
+            i_type = container_type + element_type.replace(' ', '_')
+            self.add_import(Import('stc/' + i_type + "/" + element_type, Module('stc/' + i_type, (), ())))
             return i_type
         else:
             key = dtype
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -1302,14 +1302,14 @@ class CCodePrinter(CodePrinter):
         rank  = expr.rank
 
         if rank > 0:
-            if isinstance(class_type, NumpyNDArrayType):
-                if rank > 15:
+            if isinstance(expr.class_type, (HomogeneousSetType, HomogeneousListType)):
+                    dtype = self.find_in_type_registry(expr.class_type)
+                    return dtype
+            if expr.is_ndarray or isinstance(expr.class_type, HomogeneousContainerType):
+                if expr.rank > 15:
                     errors.report(UNSUPPORTED_ARRAY_RANK, symbol=expr, severity='fatal')
                 self.add_import(c_imports['ndarrays'])
                 dtype = 't_ndarray'
-            elif isinstance(class_type, (HomogeneousSetType, HomogeneousListType)):
-                dtype = self.find_in_type_registry(expr.class_type)
-                return dtype
             else:
                 errors.report(PYCCEL_RESTRICTION_TODO+' (rank>0)', symbol=expr, severity='fatal')
         elif not isinstance(class_type, CustomDataType):
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -802,7 +802,19 @@ class CCodePrinter(CodePrinter):
                 #endif // {name}_H\n")
 
     def rename_imported_methods(self, expr):
-         if expr.source_module and expr.source_module is not self._current_module:
+        """
+        Rename Imported Methods.
+
+        This function is responsible for renaming methods of classes from
+        the imported modules, ensuring that the names are correct 
+        by prefixing them with their class names.
+
+        Parameters
+        ----------
+        expr : Import
+            The import found in the module being renamed.
+        """
+        if expr.source_module and expr.source_module is not self._current_module:
             for classDef in expr.source_module.classes:
                 class_scope = classDef.scope
                 for method in classDef.methods:
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -1305,7 +1305,7 @@ class CCodePrinter(CodePrinter):
             if isinstance(expr.class_type, (HomogeneousSetType, HomogeneousListType)):
                     dtype = self.find_in_type_registry(expr.class_type)
                     return dtype
-            if expr.is_ndarray or isinstance(expr.class_type, HomogeneousContainerType):
+            if isinstance(expr.class_type,(HomogeneousTupleType, NumpyNDArrayType)):
                 if expr.rank > 15:
                     errors.report(UNSUPPORTED_ARRAY_RANK, symbol=expr, severity='fatal')
                 self.add_import(c_imports['ndarrays'])
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -1207,7 +1207,7 @@ class CCodePrinter(CodePrinter):
 
         Raises
         ------
-        PYCCEL_RESTRICTION_TODO
+        PyccelCodegenError
             If the dtype is not found in the dtype_registry.
         """
         if isinstance(dtype, FixedSizeNumericType):
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -225,10 +225,6 @@ c_library_headers = (
 
 import_dict = {'omp_lib' : 'omp' }
 
-import_stc = {'hset' : 'SET',
-              'vec'  : 'VEC'
-              }
-
 c_imports = {n : Import(n, Module(n, (), ())) for n in
                 ['stdlib',
                  'math',
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -403,7 +403,7 @@ class CCodePrinter(CodePrinter):
 
         order = lhs.order
         lhs_dtype = lhs.dtype
-        declare_dtype = self.find_in_type_registry(lhs_dtype)
+        declare_dtype = self.get_c_type(lhs_dtype)
         if isinstance(lhs.class_type, NumpyNDArrayType):
             #set dtype to the C struct types
             dtype = self.find_in_ndarray_type_registry(lhs_dtype)
@@ -496,7 +496,7 @@ class CCodePrinter(CodePrinter):
         rhs = expr.rhs
         lhs = expr.lhs
         code_init = ''
-        declare_dtype = self.find_in_type_registry(rhs.dtype)
+        declare_dtype = self.get_c_type(rhs.dtype)
 
         if rhs.fill_value is not None:
             if isinstance(rhs.fill_value, Literal):
@@ -524,7 +524,7 @@ class CCodePrinter(CodePrinter):
             String containing the rhs of the initialization of a stack array.
         """
         var = expr
-        dtype = self.find_in_type_registry(var.dtype)
+        dtype = self.get_c_type(var.dtype)
         if isinstance(var.class_type, NumpyNDArrayType):
             np_dtype = self.find_in_ndarray_type_registry(var.dtype)
         elif isinstance(var.class_type, HomogeneousContainerType):
@@ -534,7 +534,7 @@ class CCodePrinter(CodePrinter):
         shape = ", ".join(self._print(i) for i in var.alloc_shape)
         tot_shape = self._print(functools.reduce(
             lambda x,y: PyccelMul(x,y,simplify=True), var.alloc_shape))
-        declare_dtype = self.find_in_type_registry(NumpyInt64Type())
+        declare_dtype = self.get_c_type(NumpyInt64Type())
 
         dummy_array_name = self.scope.get_new_name('array_dummy')
         buffer_array = "{dtype} {name}[{size}];\n".format(
@@ -645,19 +645,20 @@ class CCodePrinter(CodePrinter):
 
         return code
 
-    def init_stc_container(self, expr, container_var):
+    def init_stc_container(self, expr, assignment_var):
         """
         Generate the initialization of an STC container in C.
 
-        This method generates and prints the C code for initializing a container using the STC `init()` method.
+        This method generates and prints the C code for initializing a container using the STC `c_init()` method.
 
         Parameters
         ----------
         expr : TypedAstNode
             The object representing the container being printed (e.g., PythonList, PythonSet).
     
-        container_var : str
-            The variable name to which the container is being assigned.
+        assignment_var : Assign
+            The assignment node where the Python container (rhs) is being initialized 
+            and saved into a variable (lhs).
         
         Returns
         -------
@@ -665,9 +666,9 @@ class CCodePrinter(CodePrinter):
             The generated C code for the container initialization.
         """
 
-        dtype = self.find_in_type_registry(container_var.lhs.class_type)
+        dtype = self.get_c_type(assignment_var.lhs.class_type)
         keyraw = '{' + ', '.join(self._print(a) for a in expr.args) + '}'
-        init = f'{container_var.lhs.name} = c_init({dtype}, {keyraw});\n'
+        init = f'{assignment_var.lhs.name} = c_init({dtype}, {keyraw});\n'
         return init
 
     def rename_imported_methods(self, expr):
@@ -680,7 +681,7 @@ class CCodePrinter(CodePrinter):
 
         Parameters
         ----------
-        expr : Tuple of ClassDef objects
+        expr : iterable[ClassDef]
             The ClassDef objects found in the module being renamed.
         """
         for classDef in expr:
@@ -755,13 +756,13 @@ class CCodePrinter(CodePrinter):
 
     def _print_PythonFloat(self, expr):
         value = self._print(expr.arg)
-        type_name = self.find_in_type_registry(expr.dtype)
+        type_name = self.get_c_type(expr.dtype)
         return '({0})({1})'.format(type_name, value)
 
     def _print_PythonInt(self, expr):
         self.add_import(c_imports['stdint'])
         value = self._print(expr.arg)
-        type_name = self.find_in_type_registry(expr.dtype)
+        type_name = self.get_c_type(expr.dtype)
         return '({0})({1})'.format(type_name, value)
 
     def _print_PythonBool(self, expr):
@@ -795,7 +796,7 @@ class CCodePrinter(CodePrinter):
         else:
             value = self._print(PyccelAssociativeParenthesis(PyccelAdd(expr.real,
                             PyccelMul(expr.imag, LiteralImaginaryUnit()))))
-        type_name = self.find_in_type_registry(expr.dtype)
+        type_name = self.get_c_type(expr.dtype)
         return '({0})({1})'.format(type_name, value)
 
     def _print_LiteralImaginaryUnit(self, expr):
@@ -1181,19 +1182,20 @@ class CCodePrinter(CodePrinter):
             code += formatted_args_to_printf(args_format, args, end)
         return code
 
-    def find_in_type_registry(self, dtype):
+    def get_c_type(self, dtype):
         """
-        Find the corresponding C dtype in the dtype_registry or the STC container type.
+        Find the corresponding C type of the PyccelType.
 
-        This function searches for the corresponding C data type in the `dtype_registry`.
-        If the provided dtype is a container (like `HomogeneousSetType` or
-        `HomogeneousListType`), it identifies the scalar type of the container and returns
-        the appropriate type for the `STC` container.
-        It raises a `PYCCEL_RESTRICTION_TODO` error if the dtype is not found in the registry.
+        For scalar types, this function searches for the corresponding C data type
+        in the `dtype_registry`.  If the provided type is a container (like
+        `HomogeneousSetType` or `HomogeneousListType`),  it recursively identifies
+        the type of an element of the container and uses it to calculate the
+        appropriate type for the `STC` container.
+        A `PYCCEL_RESTRICTION_TODO` error is raised if the dtype is not found in the registry.
 
         Parameters
         ----------
-        dtype : DataType or ContainerType
+        dtype : PyccelType
             The data type of the expression. This can be a fixed-size numeric type,
             a primitive type, or a container type.
 
@@ -1212,7 +1214,7 @@ class CCodePrinter(CodePrinter):
             primitive_type = dtype.primitive_type
             if isinstance(primitive_type, PrimitiveComplexType):
                 self.add_import(c_imports['complex'])
-                return f'{self.find_in_type_registry(dtype.element_type)} complex'
+                return f'{self.get_c_type(dtype.element_type)} complex'
             elif isinstance(primitive_type, PrimitiveIntegerType):
                 self.add_import(c_imports['stdint'])
             elif isinstance(dtype, PythonNativeBool):
@@ -1222,7 +1224,7 @@ class CCodePrinter(CodePrinter):
             key = (primitive_type, dtype.precision)
         elif isinstance(dtype, (HomogeneousSetType, HomogeneousListType)):
             container_type = 'hset_' if dtype.name == 'set' else 'vec_'
-            element_type = self.find_in_type_registry(dtype.element_type)
+            element_type = self.get_c_type(dtype.element_type)
             i_type = container_type + element_type.replace(' ', '_')
             self.add_import(Import(f'stc/{i_type}/{element_type}', Module(f'stc/{i_type}', (), ())))
             return i_type
@@ -1235,7 +1237,6 @@ class CCodePrinter(CodePrinter):
             raise errors.report(PYCCEL_RESTRICTION_TODO, #pylint: disable=raise-missing-from
                     symbol = dtype,
                     severity='fatal')
-
 
     def find_in_ndarray_type_registry(self, dtype):
         """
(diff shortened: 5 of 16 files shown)


pyccel/parser/semantic.py
```diff
@@ -2289,6 +2289,7 @@ def _visit_CodeBlock(self, expr):
                 if len(line) != 1:
                     errors.report(f"Variable {line[0]} cannot have multiple types",
                             severity='error', symbol=line[0])
+                ls.extend(line)
```
EmilyBourne, line 2292: Please explain this change. I don't think it is necessary and am worried that it might break something

pyccel/parser/semantic.py
```diff
@@ -1896,7 +1896,7 @@ def _get_indexed_type(self, base, args, expr):
                 raise errors.report(f"Unknown annotation base {base}\n"+PYCCEL_RESTRICTION_TODO,
                         severity='fatal', symbol=expr)
             rank = 1
-            if len(args) == 2 and args[1] is LiteralEllipsis():
+            if (len(args) == 2 and args[1] is LiteralEllipsis()) or (len(args) > 0):
```
EmilyBourne, line 1899: if `len(args) == 2 and args[1] is LiteralEllipsis()` is true, then `len(args) > 0` is also true. Are you sure you need both conditions?

pyccel/parser/semantic.py
mustapha-belbiad: In the case of `int : list[int]`, the `args` holds `int`, so we need to check for it.

pyccel/parser/semantic.py
mustapha-belbiad: In the case of `int : list[int]`, the `args` holds `int`, so we need to check for it.

EmilyBourne: I agree. But that doesn't mean that you need both conditions

pyccel/parser/semantic.py
mustapha-belbiad: Without this line, the `codeBlock` is returns  empty object because `ls` is still empty if it enters this condition.

pyccel/parser/semantic.py
mustapha-belbiad: If it's okay with you, I will remove the first condition. However, I think it has a value in the code, and honestly, I don't know its benefits.

pyccel/parser/semantic.py
EmilyBourne: Hmm, this seems like a problem with the printing in Python not a general translation problem

> because ls is still empty if it enters this condition.

I disagree that this is necessarily the case

pyccel/parser/semantic.py
EmilyBourne: The reason for specifying `len(args) == 2 and args[1] is LiteralEllipsis()` was because the code below was only designed to handle expressions like `tuple[int, ...]` now that it is going to be more general I think it makes sense to change the condition. There is certainly no value in keeping slower code that evaluates more things but gives the same result. I am fine to remove the condition :+1:

pyccel/parser/semantic.py
mustapha-belbiad: if we tried to translate this code :
```python
def func():
    a : int
```
python translation :
```python
def func():
    pass
```
The only reason I found for this is that the `Codeblock` object is empty

EmilyBourne: > python translation :

Is this the translation on this branch or the devel branch? If it is broken on the devel branch then you should open an issue about that

mustapha-belbiad: > I agree. What is your point?

In the case of `a: int`, the `ls` variable doesn't take the value of `line`. This is why I added this line, to prevent the loss of the value of line and avoid returning an empty object.

EmilyBourne: > to prevent the loss of the value of line

The information is not lost it is saved elsewhere.

You do indeed arrive here if your code only contains `a : int` but that is not the only case and this doesn't guarantee that `ls` is empty.

> and avoid returning an empty object.

Why does this need to be avoided in all languages? Specifically can you explain why this should be avoided for Fortran and C (as these are the languages for which the translation needs to be optimised)

mustapha-belbiad: > > python translation :
> 
> Is this the translation on this branch or the devel branch? If it is broken on the devel branch then you should open an issue about that
 
The devel branch.

EmilyBourne: > The devel branch.

The devel branch generates:
```python
def func():
    pass
```
?
Then I don't understand what is the problem here

pyccel/parser/semantic.py
mustapha-belbiad: Fixed : d0bd13c

pyccel/codegen/printing/ccode.py
```diff
@@ -944,6 +946,22 @@ def _print_Import(self, expr):
             source = source.name[-1]
         else:
             source = self._print(source)
+        if expr.target:
+           dtype = expr.target.pop().name
+           if source.startswith('stc'):
```
EmilyBourne, line 951: I think you should keep the `/` so that you don't have problems if a user imports from a file that they created called something like `stc_pythonfile`

pyccel/codegen/printing/ccode.py
```diff
@@ -944,6 +946,22 @@ def _print_Import(self, expr):
...
+             dtype_macro = dtype.upper()
+             _,container_type = source.split("/")
+             class_type_macro = "SET" if container_type.startswith("hset") else "VEC"
```
EmilyBourne, line 954: I think a dictionary would be appropriate for this as there will be at least 2 more cases added later (dicts and arrays)

pyccel/codegen/printing/ccode.py
```diff
@@ -944,6 +946,22 @@ def _print_Import(self, expr):
...
+             import_file,_ = source.split('_')
+             if class_type_macro == 'SET':
+                  additional_defines = ''
+             additional_defines = '' if class_type_macro == 'SET' else "#define i_use_cmp\n"
```
EmilyBourne, line 958: I don't think you need both of these

EmilyBourne, line 958: `"#define i_use_cmp\n"` is specific to lists, it would be better to make this explicit instead of making it the default value. Otherwise when new containers are added this will break

pyccel/codegen/printing/ccode.py
```diff
@@ -944,6 +946,22 @@ def _print_Import(self, expr):
...
+             return '\n'.join((f'#ifndef _{class_type_macro}_{dtype_macro}',
+             f'#define _{class_type_macro}_{dtype_macro}',
+             f'#define i_key {dtype}',
+             additional_defines,
+             f'#include "{import_file}.h"',
+             f'#endif\n'))
```
EmilyBourne, line 964:
```suggestion
             return '\n'.join((f'#ifndef _{class_type_macro}_{dtype_macro}',
                               f'#define _{class_type_macro}_{dtype_macro}',
                               f'#define i_key {dtype}',
                               additional_defines,
                               f'#include "{import_file}.h"',
                               f'#endif\n'))
```

pyccel/codegen/printing/ccode.py
mustapha-belbiad: Fixed : 81411c4

mustapha-belbiad: Fixed : 81411c4

pyccel/codegen/printing/ccode.py
mustapha-belbiad: Fixed : 62e7694

EmilyBourne: Yes, but I think `import_file,_ = source.split('_')` needs to come first

mustapha-belbiad: Like this : cad722d ?

EmilyBourne: Almost, if you test the code you should see what I mean

mustapha-belbiad: > Yes, but I think `import_file,_ = source.split('_')` needs to come first

Could you please explain to me why we need this change?

mustapha-belbiad: Fixed : 01150c5

pyccel/parser/semantic.py
mustapha-belbiad: Fixed : 01150c5

pyccel/codegen/printing/ccode.py
EmilyBourne: @Farouk-Echaref did you put this back?

Farouk-Echaref: @EmilyBourne  no I don't think so, is there a version issue ? Ive only merged with the devel branch and added two more lines from my previous branch

EmilyBourne: It seems you put it back in this commit: 0f58e7cde2689f8f4587b8cfdcd3129077bcc2a1
You should be careful when merging changes. I guess this was flagged as a merge conflict and was wrongly corrected. I don't see why it would reappear otherwise.
Ideally you should pull all changes before committing anything to be sure that you are adding to the newest version but if you have forgotten to do this you can get these changes. Personally I use `git pull --rebase` to do this as this preserves the order of the commits which often makes it easier to find the code that you added but I think git merges by default.

pyccel/codegen/printing/ccode.py
```diff
@@ -952,7 +952,8 @@ def _print_Import(self, expr):
         if expr.target:
            dtype = expr.target.pop().name
            if source.startswith('stc/'):
-             import_file,_ = source.split('_')
+             import_file, *_ = source.split('_')
+             _ = '_'.join(_)
```
EmilyBourne, line 956: Another option is to use the `maxsplit` argument of `split`. This avoids unnecessary calculations

tests/epyccel/test_epyccel_variable_annotations.py
```diff
@@ -166,3 +166,20 @@ def homogeneous_tuple_annotation():
 
     assert epyc_homogeneous_tuple_annotation() == homogeneous_tuple_annotation()
     assert isinstance(epyc_homogeneous_tuple_annotation(), type(homogeneous_tuple_annotation()))
+
+@pytest.mark.parametrize( 'language', [
+        pytest.param("c", marks = pytest.mark.c),
+        pytest.param("fortran", marks = [
+            pytest.mark.skip(reason="append() not implemented in fortran"),
```
EmilyBourne, line 173: What is the relevance of `append` here?

tests/epyccel/test_epyccel_variable_annotations.py
```diff
@@ -166,3 +166,20 @@ def homogeneous_tuple_annotation():
...
+            pytest.mark.fortran]),
+        pytest.param("python", marks = pytest.mark.python)
+    ]
+)
+
+def test_homogeneous_set_annotation_int(language):
+    def homogeneous_set_annotation ():
+        a : set[int]
```
EmilyBourne, line 181: You will have to temporarily tell pylint to ignore this error:
```suggestion
        a : set[int] #pylint: disable=unused-variable
```

Farouk-Echaref: fixed e87dbf113f01d210f72bcc5679bddff399985988

tests/epyccel/test_epyccel_variable_annotations.py
```diff
@@ -166,3 +174,82 @@ def homogeneous_tuple_annotation():
...
+        pytest.param("c", marks = pytest.mark.c),
+        pytest.param("fortran", marks = [
+            pytest.mark.skip(reason="Variable declaration not implemented in fortran"),
+            pytest.mark.fortran]),
+        pytest.param("python", marks = pytest.mark.python)
+    ]
+)
+def test_homogeneous_set_annotation_int(language):
+    def homogeneous_set_annotation ():
+        a : set[int] #pylint: disable=unused-variable
```
EmilyBourne, line 188: This is only valid in Python > 3.8.
3.8 only reaches end of life in November so we still support it. You can use:
```suggestion
        a : 'set[int]' #pylint: disable=unused-variable
```
to get round this

EmilyBourne:
@Farouk-Echaref 64249cacc7bf7f038fa831099d2a47ba751b9af0 looks like a rather opaque solution to this problem and one that could cause problems in some situations. E.g:
```python
class A:
    def A__method():
        pass
```

The only reason the renaming is done in `ModuleHeader` is because `_print_ModuleHeader` is called before `_print_Module`. If you change the order then you could simply change where the renaming is done

EmilyBourne:
The tests that are failing currently are all to do with more complex list functionalities (list comprehension). I suggest that you mark these tests as `xfail` for now and open issues about the problems. It is not surprising that these don't work out of the box when we don't even have #1876 yet

tests/epyccel/test_functionals.py
```diff
@@ -1,10 +1,24 @@
...
+import pytest
+
 
 from pyccel import epyccel
 from modules import functionals
 
+@pytest.fixture( params=[
+        pytest.param("fortran", marks = pytest.mark.fortran),
+        pytest.param("c", marks = [
+            pytest.mark.xfail(reason="C does not support list indexing yet."),
```
EmilyBourne, line 13: Please reference the relevant issue

pyccel/codegen/printing/ccode.py
```diff
@@ -1241,6 +1282,9 @@ def get_declare_type(self, expr):
             if expr.is_ndarray or isinstance(expr.class_type, HomogeneousContainerType):
                 if expr.rank > 15:
```
jalalium, line 1283: Won't this conflict with the rank of `HomogeneousContainerType`? is this relicate from ndarrays still relevant?

pyccel/codegen/printing/ccode.py
```diff
@@ -1963,6 +2014,31 @@ def _print_FunctionCall(self, expr):
         else:
             return call_code
 
+    def init_lists_or_sets(self, expr, container_var):
```
jalalium, line 2017: I think this function should be broken down into two functions, one for lists and another one for sets. Either this or make it initialize all containers

pyccel/codegen/wrapper/fortran_to_c_wrapper.py
```diff
@@ -483,6 +484,8 @@ def _wrap_Variable(self, expr):
                           original_function = expr)
             return expr.clone(expr.name, new_class = BindCArrayVariable, wrapper_function = func,
                                 original_variable = expr)
+        else:
+            raise NotImplementedError("Lists cannot be wrapped yet")
```
jalalium, line 488: What about sets?

tests/epyccel/modules/list_comprehension.py
```diff
@@ -0,0 +1,4 @@
+# pylint: disable=missing-function-docstring, missing-module-docstring
+
+F = [False for _ in range(5)]
```
jalalium, line 3: F was the name of the variable in the old file, Is it better to change it?

pyccel/codegen/pipeline.py
```diff
@@ -334,6 +334,12 @@ def handle_error(stage):
 
             mod_obj.add_dependencies(stdlib)
 
+
+    # Iterate over the external_libs list and determine if the printer
+    # requires an external lib to be included.
```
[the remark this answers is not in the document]
Farouk-Echaref: Since STC only depends on header files and doesn't require a compiled object, we can use a map with the library name as the key and its path as the value, this way we can avoid static lookups to retrieve the library name and its path. The same method can be applied to future external libraries.
Please check this update in commit `98967292141cfa853b1b8ea5c4dec35334210dc9` and share your opinion.

pyccel/codegen/printing/ccode.py
```diff
@@ -225,6 +225,10 @@
 
 import_dict = {'omp_lib' : 'omp' }
 
+import_stc = {'hset' : 'SET',
```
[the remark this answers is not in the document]
Farouk-Echaref: this is now deprecated e4c4f3ad7d567b8beb0c147c28f2a04b7f3ba34f

pyccel/codegen/printing/ccode.py
Farouk-Echaref: @EmilyBourne I'm not familiar with this case, could you provide more details?

pyccel/codegen/printing/ccode.py
Farouk-Echaref: Would `init_stc_container()` be a more fitting function name?

pyccel/codegen/wrapper/fortran_to_c_wrapper.py
Farouk-Echaref: updated 296ca4e2ad0d07effc7402f327d7087e964b1224

tests/epyccel/modules/list_comprehension.py
Farouk-Echaref: updated 378ade635b72a58c302931e7c76ddc2f16d4f3c6

pyccel/codegen/printing/ccode.py
```diff
@@ -1163,6 +1197,13 @@ def find_in_dtype_registry(self, dtype):
                 return 'bool'
 
             key = (primitive_type, dtype.precision)
+        elif isinstance(dtype, (HomogeneousSetType, HomogeneousListType)):
+            container_type = 'hset_' if dtype.name == 'set' else 'vec_'
+            if not isinstance(dtype.element_type, GenericType):
```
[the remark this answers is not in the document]
Farouk-Echaref: this is an edge case that will require special handling(keep track of the variable until it gets assigned a known type), we can discuss this over our meeting @EmilyBourne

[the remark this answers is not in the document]
EmilyBourne: A `GenericType` is equivalent to "AnyType". It is used to describe the type of arguments passed to MPI or objects whose type cannot yet be identified (e.g. an empty list). If you need to print a `GenericType` in the codegen stage then something has gone wrong and an error must be raised.

An example where a `GenericType` may be used is:
```python
a = []
```
However an error should be raised here as the type of `a` is not properly defined. Instead you should have:
```python
a : list[int] = []
```

If I translate this example to Fortran I get:
```
pyccel:
 |fatal [codegen]: tmp_lst.py| _print_GenericType is not yet implemented for language : Fortran
Pyccel has encountered syntax that has not been implemented yet. Please create an issue at https://github.com/pyccel/pyccel/issues and provide a small example of your problem. Do not forget to specify your target language. (Generic)
```
This is a poor error as it doesn't even tell us the line.
If I translate the example to C on your branch, the result is much worse. I get:
```
  File "/home/emily/.local/bin/pyccel", line 8, in <module>
    sys.exit(pyccel())
  File "/home/emily/Code/pyccel/pyccel/commands/console.py", line 266, in pyccel
    execute_pyccel(filename,
  File "/home/emily/Code/pyccel/pyccel/codegen/pipeline.py", line 266, in execute_pyccel
    fname, prog_name = codegen.export(fname)
  File "/home/emily/Code/pyccel/pyccel/codegen/codegen.py", line 220, in export
    code = self._printer.doprint(self.ast)
  File "/home/emily/Code/pyccel/pyccel/codegen/printing/codeprinter.py", line 52, in doprint
    lines = self._print(expr).splitlines(True)
  File "/home/emily/Code/pyccel/pyccel/codegen/printing/codeprinter.py", line 88, in _print
    obj = getattr(self, print_method)(expr)
  File "/home/emily/Code/pyccel/pyccel/codegen/printing/ccode.py", line 838, in _print_Module
    body    = ''.join(self._print(i) for i in expr.body)
  File "/home/emily/Code/pyccel/pyccel/codegen/printing/ccode.py", line 838, in <genexpr>
    body    = ''.join(self._print(i) for i in expr.body)
  File "/home/emily/Code/pyccel/pyccel/codegen/printing/codeprinter.py", line 88, in _print
    obj = getattr(self, print_method)(expr)
  File "/home/emily/Code/pyccel/pyccel/codegen/printing/ccode.py", line 1949, in _print_FunctionDef
  File "/home/emily/Code/pyccel/pyccel/codegen/printing/codeprinter.py", line 88, in _print
    obj = getattr(self, print_method)(expr)
  File "/home/emily/Code/pyccel/pyccel/codegen/printing/ccode.py", line 2268, in _print_CodeBlock
    code = self._print(b)
  File "/home/emily/Code/pyccel/pyccel/codegen/printing/codeprinter.py", line 88, in _print
    obj = getattr(self, print_method)(expr)
  File "/home/emily/Code/pyccel/pyccel/codegen/printing/ccode.py", line 2181, in _print_Assign
    return prefix_code+self.init_lists_or_sets(rhs, expr)
  File "/home/emily/Code/pyccel/pyccel/codegen/printing/ccode.py", line 2033, in init_lists_or_sets
    dtype = self.find_in_dtype_registry(container_var.lhs.class_type)
  File "/home/emily/Code/pyccel/pyccel/codegen/printing/ccode.py", line 1200, in find_in_dtype_registry
    i_type = container_type + vec_dtype.replace(' ', '_')
UnboundLocalError: local variable 'vec_dtype' referenced before assignment
```

pyccel/codegen/printing/ccode.py
EmilyBourne: I'm not sure what you mean by "conflict with the rank". Ndarrays are restricted to maximum rank 15 due to restrictions in our implementation. This is not the case for lists but I also would not advise defining a `[[[[[[[[[[[[[[[[]]]]]]]]]]]]]]]]`. That said the code is currently:
```python
if expr.is_ndarray or isinstance(expr.class_type, HomogeneousContainerType):
                if expr.rank > 15:
                    errors.report(UNSUPPORTED_ARRAY_RANK, symbol=expr, severity='fatal')
                if isinstance(expr.class_type, (HomogeneousSetType, HomogeneousListType)):
                    dtype = self.find_in_dtype_registry(expr.class_type)
                    return dtype
                self.add_import(c_imports['ndarrays'])
                dtype = 't_ndarray'
```
So there does not seem to be any gain from grouping NDarrays with other HomogeneousContainerTypes. I think it would make more sense to have:
```python
class_type = expr.class_type
if isinstance(class_type, NDArrayType):
                if expr.rank > 15:
                    errors.report(UNSUPPORTED_ARRAY_RANK, symbol=expr, severity='fatal')
                    return dtype
                self.add_import(c_imports['ndarrays'])
                dtype = 't_ndarray'
elif isinstance(class_type, (HomogeneousSetType, HomogeneousListType)):
                dtype = self.find_in_dtype_registry(expr.class_type)
                return dtype
elif isinstance(class_type, FixedSizeNumericType):
     ....
else:
    errrors.report(...)
```

pyccel/codegen/printing/ccode.py
EmilyBourne: I don't see any benefit to splitting this given that the code is identical. I also don't think it should initialise arrays as the code would be completely different. I like the name `init_stc_container()`

pyccel/codegen/wrapper/fortran_to_c_wrapper.py
EmilyBourne: NDArray is a container that can be wrapped. Why not use the variable to clarify the error message:
```python
            raise NotImplementedError("Objects of type {expr.class_type} cannot be wrapped yet")
```

pyccel/codegen/printing/ccode.py
Farouk-Echaref: updated 749d5e524ccfb0319f0372cbe3d9cd7404eae994

pyccel/codegen/printing/ccode.py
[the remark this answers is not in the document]
Farouk-Echaref: fixed 4196bcdf994a6bb756c8c27ba40788155b41f96f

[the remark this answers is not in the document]
jalalium: It still does not reflect the actual error as it does in Fortran

C:
```
pyccel:
 |fatal [codegen]: odo.py| Pyccel has encountered syntax that has not been implemented yet. Please create an issue at https://github.com/pyccel/pyccel/issues and provide a small example of your problem. Do not forget to specify your target language. (Generic)
```
Fortran:
```
pyccel:
 |fatal [codegen]: odo.py| _print_GenericType is not yet implemented for language : Fortran
Pyccel has encountered syntax that has not been implemented yet. Please create an issue at https://github.com/pyccel/pyccel/issues and provide a small example of your problem. Do not forget to specify your target language. (Generic)
```

is this enough?

pyccel/codegen/printing/ccode.py
Farouk-Echaref: can you check the update in this commit 941703584992ce6d99ca196ee146ef23b29e7065

pyccel/codegen/wrapper/fortran_to_c_wrapper.py
Farouk-Echaref: updated c7d984cc6fc3ffbb26a030a3e7562eb212872497

pyccel/codegen/printing/ccode.py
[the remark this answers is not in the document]
Farouk-Echaref: An issue has been created for this:
 https://github.com/pyccel/pyccel/issues/1910

pyccel/codegen/printing/ccode.py
EmilyBourne: Looks good to me

jalalium (approved):

pyccel/codegen/printing/ccode.py
```diff
@@ -1963,6 +2010,31 @@ def _print_FunctionCall(self, expr):
...
+            The variable name to which the container is being assigned.
+        
+        Returns
+        -------
+        str
+            The generated C code for the container initialization.
+        """
+
+        dtype = self.find_in_dtype_registry(container_var.lhs.class_type)
+        keyraw = '{' + ', '.join([self._print(a) for a in expr.args]) + '}'
```
EmilyBourne, line 2034:
```suggestion
        keyraw = '{' + ', '.join(self._print(a) for a in expr.args) + '}'
```
`join` takes an iterator. This is usually better as it avoids an unnecessary memory allocation

pyccel/parser/semantic.py
```diff
@@ -1615,7 +1615,7 @@ def _ensure_inferred_type_matches_existing(self, class_type, d_var, var, is_auga
             # to remove memory leaks
             new_expressions.append(Deallocate(var))
 
-        elif class_type != var.class_type:
+        elif class_type != var.class_type and not isinstance(class_type.datatype, GenericType):
```
EmilyBourne, line 1618: Please can you explain this change. I don't see it mentioned in the PR description

pyccel/parser/semantic.py
```diff
@@ -1919,8 +1919,7 @@ def _get_indexed_type(self, base, args, expr):
             else:
                 raise errors.report(f"Unknown annotation base {base}\n"+PYCCEL_RESTRICTION_TODO,
                         severity='fatal', symbol=expr)
-            rank = 1
-            if len(args) == 2 and args[1] is LiteralEllipsis():
+            if len(args) > 0:
```
EmilyBourne, line 1922: This condition is significantly more permissive than the old condition. It implies that you now correctly handle (or correctly raise an error for expressions such as `tuple[int, bool, float]`. Is this correct?

pyccel/parser/semantic.py
```diff
@@ -1932,6 +1931,8 @@ def _get_indexed_type(self, base, args, expr):
                     class_type = HomogeneousTupleType
                 elif dtype_cls is PythonList:
                     class_type = HomogeneousListType
+                elif dtype_cls is PythonSet:
+                    class_type = HomogeneousSetType
```
EmilyBourne, line 1935: This is becoming quite a long list of if/elif blocks. Maybe a dict could be used here (this has already been mentioned in #1901 so it should probably be fixed there as that PR must be merged first anyway)

tests/epyccel/test_arrays.py
```diff
@@ -4027,7 +4027,15 @@ def test_array_float_nested_C_array_initialization_3(language):
 #==============================================================================
 # NUMPY SUM
 #==============================================================================
-
+@pytest.mark.parametrize( 'language', (
+        pytest.param("fortran", marks = pytest.mark.fortran),
+        pytest.param("c", marks = [
+            pytest.mark.xfail(reason="List indexing is not yet supported in C"),
```
EmilyBourne, line 4033: Please reference the issue that will fix this (this makes it easier to find all xfails that should be removed when you fix the issue as you just need to search for the number)

tests/epyccel/test_epyccel_modules.py
```diff
@@ -171,6 +170,25 @@ def test_module_7(language):
     mod.reset_c()
     mod.reset_e()
 
+@pytest.mark.parametrize( 'language', (
+        pytest.param("fortran", marks = [
+            pytest.mark.xfail(reason="List wrapper is not implemented yet."),
+            pytest.mark.fortran]
+            ),
+        pytest.param("c", marks = [
+            pytest.mark.xfail(reason="List indexing is not yet supported in C"),
```
EmilyBourne, line 179: Please reference the issues that will fix this (this makes it easier to find all xfails that should be removed when you fix the issue as you just need to search for the number)

tests/epyccel/test_epyccel_variable_annotations.py
```diff
@@ -10,6 +10,14 @@
 from pyccel.errors.errors import PyccelSemanticError, Errors
 from pyccel.decorators import allow_negative_index, stack_array
 
+def parametrize_languages():
+    return pytest.mark.parametrize('language', [
+        pytest.param("c", marks=pytest.mark.c),
+        pytest.param("fortran", marks=[
+            pytest.mark.skip(reason="Variable declaration not implemented in fortran"),
```
EmilyBourne, line 17: Please reference the issue that will fix this (this makes it easier to find all xfails that should be removed when you fix the issue as you just need to search for the number)

tests/epyccel/test_epyccel_variable_annotations.py
```diff
@@ -166,3 +174,107 @@ def homogeneous_tuple_annotation():
 
     assert epyc_homogeneous_tuple_annotation() == homogeneous_tuple_annotation()
     assert isinstance(epyc_homogeneous_tuple_annotation(), type(homogeneous_tuple_annotation()))
+
+@pytest.mark.parametrize( 'language', [
+        pytest.param("c", marks = pytest.mark.c),
+        pytest.param("fortran", marks = [
+            pytest.mark.skip(reason="Variable declaration not implemented in fortran"),
```
EmilyBourne, line 181: Please reference the issue that will fix this (this makes it easier to find all xfails that should be removed when you fix the issue as you just need to search for the number)

tests/epyccel/test_epyccel_variable_annotations.py
```diff
@@ -10,6 +10,14 @@
...
 from pyccel.decorators import allow_negative_index, stack_array
 
+def parametrize_languages():
+    return pytest.mark.parametrize('language', [
+        pytest.param("c", marks=pytest.mark.c),
+        pytest.param("fortran", marks=[
+            pytest.mark.skip(reason="Variable declaration not implemented in fortran"),
+            pytest.mark.fortran]),
+        pytest.param("python", marks=pytest.mark.python)
+    ])
```
EmilyBourne, line 20: An alternative to this (and a more standard way of handling it) would be to create a [`fixture`](https://docs.pytest.org/en/stable/explanation/fixtures.html) with a different name that is used until Fortran can be included.

EmilyBourne: This is what you have done for functionals

CHANGELOG.md
```diff
@@ -21,7 +21,7 @@ All notable changes to this project will be documented in this file.
 -   #1844 : Add line numbers and code to errors from built-in function calls.
 -   \[INTERNALS\] Added `container_rank` property to `ast.datatypes.PyccelType` objects.
 -   \[DEVELOPER\] Added an improved traceback to the developer-mode errors for errors in function calls.
-
+-   #1847 : Support `list()` and `set()` declarations in c.
 ### Fixed
```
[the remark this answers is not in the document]
Farouk-Echaref: updated 674da248f4a75a0847b88d71a5a8513304ef1c57

pyccel/ast/datatypes.py
```diff
@@ -488,6 +488,16 @@ class HomogeneousContainerType(ContainerType):
...
+    _name = None
+
+    @property
+    def name(self):
+        """
+        Get the name of the container.
+        
+        Get the name of the container.
+        """
+        return self._name
```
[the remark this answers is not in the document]
Farouk-Echaref: If I remember correctly, it was because of this line (line 1197 in `ccode.py`) :
```python
container_type = 'hset_' if dtype.name == 'set' else 'vec_'
```
we were accessing the private attribute directly (`dtype._name`) so I created a getter property for it.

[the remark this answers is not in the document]
Farouk-Echaref: I understand we can just check on the class type without needing this

pyccel/codegen/printing/ccode.py
```diff
@@ -802,16 +797,51 @@ def _print_ModuleHeader(self, expr):
                 {funcs}\n \
                 #endif // {name}_H\n")
 
+    def rename_imported_methods(self, expr):
+        """
+        Rename Imported Methods.
+
```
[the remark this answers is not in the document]
Farouk-Echaref: updated c865d8effdfb8026e976709a10d9249c60120848

pyccel/codegen/printing/ccode.py
```diff
@@ -802,16 +797,51 @@ def _print_ModuleHeader(self, expr):
...
+        This function is responsible for renaming methods of classes from
+        the imported modules, ensuring that the names are correct 
+        by prefixing them with their class names.
```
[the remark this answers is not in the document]
Farouk-Echaref: updated c865d8effdfb8026e976709a10d9249c60120848

pyccel/codegen/printing/ccode.py
```diff
@@ -802,16 +797,51 @@ def _print_ModuleHeader(self, expr):
...
+            self.rename_imported_methods(item)
+        for classDef in  expr.classes:
+            class_scope = classDef.scope
+            for method in classDef.methods:
+                if not method.is_inline:
+                    class_scope.rename_function(method, f"{classDef.name}__{method.name.lstrip('__')}")
+            for interface in classDef.interfaces:
+                for func in interface.functions:
+                    if not func.is_inline:
+                        class_scope.rename_function(func, f"{classDef.name}__{func.name.lstrip('__')}")
```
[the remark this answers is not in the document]
Farouk-Echaref: tell what you think of this update c865d8effdfb8026e976709a10d9249c60120848

pyccel/ast/datatypes.py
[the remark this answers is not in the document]
EmilyBourne: Ah ok. That makes sense :+1: I just wanted to check it wasn't being misused
Please can you move this getter to `PyccelType` so all types can use this

pyccel/codegen/printing/ccode.py
```diff
@@ -645,6 +645,29 @@ def _handle_inline_func_call(self, expr):
...
+        """
+        for classDef in expr:
+            class_scope = classDef.scope
+        for method in classDef.methods:
+            if not method.is_inline:
+                class_scope.rename_function(method, f"{classDef.name}__{method.name.lstrip('__')}")
+        for interface in classDef.interfaces:
+            for func in interface.functions:
+                if not func.is_inline:
+                    class_scope.rename_function(func, f"{classDef.name}__{func.name.lstrip('__')}")
```
EmilyBourne, line 669: Bad indentation

CHANGELOG.md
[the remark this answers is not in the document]
EmilyBourne: > Messages for users are usually placed above messages for developers so this line should be 2 rows higher.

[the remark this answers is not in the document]
EmilyBourne: I don't think the `[LISTS]` and `[SETS]` tags are useful in the changelog. The aim is to have a reference (i.e. the number) to an issue with more details and 1 sentence describing what was implemented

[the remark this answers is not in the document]
Farouk-Echaref: > > Messages for users are usually placed above messages for developers so this line should be 2 rows higher.

sorry forgot about this one, will be updated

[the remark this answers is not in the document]
Farouk-Echaref: > I don't think the `[LISTS]` and `[SETS]` tags are useful in the changelog. The aim is to have a reference (i.e. the number) to an issue with more details and 1 sentence describing what was implemented

what do you think about this description:
#1655 : Add the appropriate C language equivalent for declaring a Python `list` container using the `STC` library.

pyccel/ast/datatypes.py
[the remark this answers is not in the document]
Farouk-Echaref: updated d0c218b344206f673c1a4977a25b9b16d806d227

CHANGELOG.md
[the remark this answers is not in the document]
EmilyBourne: Looks great :+1:

pyccel/codegen/printing/ccode.py
Farouk-Echaref: fixed 1e988678f8492fd1644c474e3e00752bed29dd39

CHANGELOG.md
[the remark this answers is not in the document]
Farouk-Echaref: updated 3765e6bfd2f42030cc39869a0ee343afc8913578

pyccel/codegen/printing/ccode.py
```diff
@@ -1163,6 +1193,12 @@ def find_in_dtype_registry(self, dtype):
                 return 'bool'
 
             key = (primitive_type, dtype.precision)
+        elif isinstance(dtype, (HomogeneousSetType, HomogeneousListType)):
+            container_type = 'hset_' if dtype.name == 'set' else 'vec_'
+            vec_dtype = self.find_in_dtype_registry(dtype.element_type)
+            i_type = container_type + vec_dtype.replace(' ', '_')
```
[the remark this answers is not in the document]
Farouk-Echaref: this was from my last PR I think, I forgot to change it when I merged with Mustapha
updated 262e3036e7bd48596c9f291dd0982c6d31615a0c

pyccel/codegen/printing/ccode.py
```diff
@@ -1163,6 +1193,12 @@ def find_in_dtype_registry(self, dtype):
...
+            self.add_import(Import('stc/' + i_type + "/" + vec_dtype, Module('stc/' + i_type, (), ())))
```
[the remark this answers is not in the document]
Farouk-Echaref: understandable, updated here cc73b94c306a4196c448739af85245460364d550

pyccel/codegen/printing/ccode.py
```diff
@@ -1238,11 +1274,15 @@ def get_declare_type(self, expr):
         rank  = expr.rank
 
         if rank > 0:
-            if expr.is_ndarray or isinstance(expr.class_type, HomogeneousContainerType):
-                if expr.rank > 15:
+            if isinstance(class_type, NumpyNDArrayType):
+                if rank > 15:
                     errors.report(UNSUPPORTED_ARRAY_RANK, symbol=expr, severity='fatal')
+                    return dtype
```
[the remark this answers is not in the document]
Farouk-Echaref: new issue for the coverage bot apparently 
fixed 8defc5825cca506717b7774377d960fb8067e661

pyccel/codegen/printing/ccode.py
```diff
@@ -1238,11 +1274,15 @@ def get_declare_type(self, expr):
...
                 self.add_import(c_imports['ndarrays'])
                 dtype = 't_ndarray'
+            elif isinstance(class_type, (HomogeneousSetType, HomogeneousListType)):
+                dtype = self.find_in_dtype_registry(expr.class_type)
+                return dtype
```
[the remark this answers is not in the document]
Farouk-Echaref: Should the docstring be updated as well?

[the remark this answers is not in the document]
EmilyBourne: Yes probably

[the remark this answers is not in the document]
Farouk-Echaref: what do you think about this docstring:
```python
"""
        Find the corresponding C dtype in the dtype_registry or the STC container type.

        This function searches for the corresponding C data type in the `dtype_registry`.
        If the provided dtype is a container (like `HomogeneousSetType` or
        `HomogeneousListType`), it identifies the scalar type of the container and returns
        the appropriate type for the `STC` container.
        It raises a `PYCCEL_RESTRICTION_TODO` error if the dtype is not found in the registry.

        Parameters
        ----------
        dtype : DataType or ContainerType
            The data type of the expression. This can be a fixed-size numeric type,
            a primitive type, or a container type.

        Returns
        -------
        str
            The code which declares the data type in C or the corresponding `STC` container
            type.

        Raises
        ------
        PYCCEL_RESTRICTION_TODO
            If the dtype is not found in the dtype_registry.
        """
```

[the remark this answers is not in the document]
Farouk-Echaref: you can check the updated version here e620bc4552e3560084f6115b353e6919a959d655

pyccel/codegen/printing/ccode.py
```diff
@@ -1559,6 +1601,11 @@ def _print_Allocate(self, expr):
             raise NotImplementedError(f"Allocate not implemented for {variable}")
 
     def _print_Deallocate(self, expr):
+        if isinstance(expr.variable.class_type, (HomogeneousListType, HomogeneousSetType)):
+            variable_address = self._print(ObjectAddress(expr.variable))
+            dtype = self.find_in_dtype_registry(expr.variable.class_type.element_type).replace(" ", "_")
+            container_type = "hset_"if isinstance(expr.variable.class_type, HomogeneousSetType) else "vec_"
```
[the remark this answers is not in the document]
Farouk-Echaref: Yes, this is repetitive
updated 6d07beb104be5119e1efbe0bd6988fc3a2e1c423

pyccel/codegen/printing/ccode.py
```diff
@@ -1963,6 +2010,31 @@ def _print_FunctionCall(self, expr):
         else:
             return call_code
 
+    def init_stc_container(self, expr, container_var):
+        """
+        Generate the initialization of an STC container in C.
+
```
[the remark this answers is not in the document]
Farouk-Echaref: updated f9a7fb0c614fbe942b702149910043a43676e1f9

pyccel/parser/semantic.py
Farouk-Echaref: this is now deprecated after I merged the `devel` branch into my branch

pyccel/parser/semantic.py
Farouk-Echaref: this is now deprecated after I merged the `devel` branch into my branch

pyccel/codegen/printing/ccode.py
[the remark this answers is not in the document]
EmilyBourne: Looks good. It seems like it should talk about returning the type but this is probably more accurate as it doesn't work for arrays :+1:

pyccel/ast/builtins.py
```diff
@@ -1231,6 +1231,7 @@ def print_string(self):
     'min'      : PythonMin,
     'not'      : PyccelNot,
     'map'      : PythonMap,
+    'set'      : PythonSet,
```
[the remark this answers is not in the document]
Farouk-Echaref: it was still in Mustapha's pr:
```python
'set'      : PythonSet,
'set'      : PythonSetFunction
```
should I remove it in my PR?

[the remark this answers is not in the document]
EmilyBourne: > it was still in Mustapha's pr:

No it wasn't. This was a new line in this PR. Mustapha's PR adds:
```python
'set'      : PythonSetFunction
```
which is the function that we want to get when we call `set()`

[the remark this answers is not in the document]
Farouk-Echaref: > To be removed once #1901 is fixed

what did you mean by this?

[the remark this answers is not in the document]
Farouk-Echaref: my bad
updated 16b721cd1158ba4f64b8050ce3a49d6d3cfb03ef

pyccel/parser/semantic.py
Farouk-Echaref: this is now deprecated. I've used `__eq__()` of the `GenericType` class from your PR, and added new ones for `HomogeneousListType` and `HomogeneousSetType` 
38d214c4ea66a9a06b29bb66979c9f8ac116a853

Farouk-Echaref: the trailing new lines error were fixed

pyccel/codegen/printing/ccode.py
Farouk-Echaref: Since `tuples` are implemented using `ndarrays`, this change caused the `tuple` tests to fail. By simply checking `lists` and `sets` in the first part and then handling the rest of the homogeneous containers (such as `ndarrays` and `tuples`) in the second part, the restriction of `rank ≤ 15` is applied only to `ndarrays` and `tuples`.
f3b368b983e387834e4863df0597199befb38f83

tests/epyccel/test_epyccel_variable_annotations.py
Farouk-Echaref: what do you think about this:
721dfd8f5fde86dfb3602051f83d9156da89cce9

pyccel/codegen/printing/ccode.py
EmilyBourne: Ah, good catch! I forgot about tuples. I would rather check `if isinstance(class_type, (NDArrayType, HomogeneousTupleType)):` though. Dicts and cuda arrays will be homogeneous containers and it will be easier to find the places where they are missing support if we have clean errors instead of automatically treating them like ndarrays

tests/epyccel/test_epyccel_variable_annotations.py
EmilyBourne: I'm not sure why you created a new file? Presumably this is because you kept the same name for the fixture? It is hard enough to locate tests at the moment I would rather not make this work. I think it would be better to rename the fixture. E.g. `stc_language`
This can still easily be changed when it's fixed using replace all

pyccel/codegen/printing/ccode.py
```diff
@@ -1134,35 +1181,51 @@ def formatted_args_to_printf(args_format, args, end):
...
             elif isinstance(dtype, PythonNativeBool):
                 self.add_import(c_imports['stdbool'])
                 return 'bool'
 
             key = (primitive_type, dtype.precision)
+        elif isinstance(dtype, (HomogeneousSetType, HomogeneousListType)):
+            container_type = 'hset_' if dtype.name == 'set' else 'vec_'
+            element_type = self.find_in_type_registry(dtype.element_type)
+            i_type = container_type + element_type.replace(' ', '_')
+            self.add_import(Import(f'stc/{i_type}/{element_type}', Module(f'stc/{i_type}', (), ())))
```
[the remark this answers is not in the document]
Farouk-Echaref: When initializing an object of class `Import`, it runs an assertion that expects a `Module` object.
```python
assert isinstance(i, (AsName, Module))
```

pyccel/codegen/printing/ccode.py
Farouk-Echaref: updated 1134b8e52e17fcc75c1c4123d5820f685ff9c167

tests/epyccel/test_epyccel_variable_annotations.py
Farouk-Echaref: tell me what you think about this version:
cd22d9552fa7c174faa97aaba068fe8ff2acd247

tests/epyccel/test_epyccel_variable_annotations.py
Farouk-Echaref: updated cd22d9552fa7c174faa97aaba068fe8ff2acd247

tests/epyccel/test_epyccel_variable_annotations.py
Farouk-Echaref: updated cd22d9552fa7c174faa97aaba068fe8ff2acd247

tests/epyccel/test_epyccel_modules.py
Farouk-Echaref: updated cd22d9552fa7c174faa97aaba068fe8ff2acd247

pyccel/codegen/printing/ccode.py
EmilyBourne: @jalalium Can we close this conversation? As per the [docs](https://github.com/pyccel/pyccel/blob/devel/developer_docs/review_process.md)
> Please only resolve conversations that you opened. You may think you fixed the problem, but the reviewer may disagree and leaving the discussion open makes it easier for them to verify that they agree with you. If you are reviewing then please close all conversations that you open once the problem is resolved. If you don't this can block the merge.

pyccel/codegen/printing/ccode.py
EmilyBourne: @jalalium Can we close this conversation?

pyccel/codegen/wrapper/fortran_to_c_wrapper.py
EmilyBourne: @jalalium Can we close this conversation?

tests/epyccel/test_epyccel_variable_annotations.py
EmilyBourne: Looks good

pyccel/codegen/printing/ccode.py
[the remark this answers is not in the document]
EmilyBourne: Ah I didn't realise this. I was actually hoping that we could use the target parameter to avoid fiddling around with strings.

I have had to make this change in order to support STC arrays via cspan (the import is the same in this case). I am tempted to incorporate the change in this PR but I think it might be better to leave it as an improvement in a later PR (at least so I don't review too much of my own code).

In case you're interested this is what I had in mind:
https://github.com/pyccel/pyccel/blob/8c04f46c13ec01074d60f48413ae1137a5518d90/pyccel/codegen/printing/ccode.py#L1228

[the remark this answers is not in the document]
Farouk-Echaref: Looks better and more readable,I wish we had taken this approach earlier.
which PR are you using this?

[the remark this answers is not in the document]
EmilyBourne: I haven't opened the PR yet. I think I will open one just for this once this PR is merged. The branch is `ebourne_stc_arrays`

pyccel/parser/semantic.py
```diff
@@ -61,7 +61,7 @@
 from pyccel.ast.class_defs import NumpyArrayClass, TupleClass, get_cls_base
 
 from pyccel.ast.datatypes import CustomDataType, PyccelType, TupleType, VoidType, GenericType
-from pyccel.ast.datatypes import PrimitiveIntegerType, HomogeneousListType, StringType, SymbolicType
+from pyccel.ast.datatypes import PrimitiveIntegerType, HomogeneousListType, StringType, SymbolicType, HomogeneousSetType
 from pyccel.ast.datatypes import PythonNativeBool, PythonNativeInt, PythonNativeFloat
 from pyccel.ast.datatypes import DataTypeFactory, PrimitiveFloatingPointType
 from pyccel.ast.datatypes import InhomogeneousTupleType, HomogeneousTupleType, HomogeneousSetType
```
EmilyBourne, line 67: Duplicate import

Farouk-Echaref: updated 59e39e8bcaf07e66951693ba9cb17d668be0a071

pyccel/codegen/printing/ccode.py
```diff
@@ -645,6 +645,54 @@ def _handle_inline_func_call(self, expr):
 
         return code
 
+    def init_stc_container(self, expr, container_var):
+        """
+        Generate the initialization of an STC container in C.
+
+        This method generates and prints the C code for initializing a container using the STC `init()` method.
```
EmilyBourne, line 652:
```suggestion
        This method generates and prints the C code for initializing a container using the STC `c_init()` method.
```

pyccel/codegen/printing/ccode.py
```diff
@@ -645,6 +645,54 @@ def _handle_inline_func_call(self, expr):
...
+        
+        Returns
+        -------
+        str
+            The generated C code for the container initialization.
+        """
+
+        dtype = self.find_in_type_registry(container_var.lhs.class_type)
+        keyraw = '{' + ', '.join(self._print(a) for a in expr.args) + '}'
+        init = f'{container_var.lhs.name} = c_init({dtype}, {keyraw});\n'
```
EmilyBourne, line 670: `str` doesn't have a `lhs` method. Are you sure that you describe the correct type here?

EmilyBourne, line 670: See the [docs](https://github.com/pyccel/pyccel/blob/devel/developer_docs/codegen_stage.md):
> It can be tempting to skip some of these `_print` calls, especially for basic types such as `Literal`s.
> However it is very important to use these functions as much as possible, for several reasons:
> 1.  It ensures that the same conventions are used throughout the generated code
> 2.  It ensures that code details are not forgotten
> 3.  It makes refactoring much simpler
>
> For example, it is sometimes necessary to print a `0`.
> It is very tempting to just use `'0'` directly, however this would mean that the precision of the integer would not be printed.
> In some languages (e.g. Fortran) not specifying the precision in literals can have unintended consequences on the final result.

For example, your code will break with both of these examples:
```python
class A:
    def __init__(self, a):
        self.a = [1,2,3]
```

```python
def f():
    a = [1,2,3]
    b = [3,4,5]
    return a,b
```

pyccel/codegen/printing/ccode.py
```diff
@@ -645,6 +645,54 @@ def _handle_inline_func_call(self, expr):
...
+        """
+        Rename class methods from user-defined imports.
+
+        This function is responsible for renaming methods of classes from
+        the imported modules, ensuring that the names are correct 
+        by prefixing them with their class names.
+
+        Parameters
+        ----------
+        expr : Tuple of ClassDef objects
```
EmilyBourne, line 683:
```suggestion
        expr : tuple[ClassDef]
```
Do you require a tuple or would a list also be ok?
Maybe:
```suggestion
        expr : iterable[ClassDef]
```
?

pyccel/codegen/printing/ccode.py
```diff
@@ -947,23 +994,23 @@ def _print_Import(self, expr):
...
+        if expr.target:
+            if source.startswith('stc/'):
+                stc_name, container_type, container_key = source.split("/")
+                container = container_type.split("_")
+                return '\n'.join((f'#ifndef _{container_type.upper()}',
+                                  f'#define _{container_type.upper()}',
+                                  f'#define i_type {container_type}',
+                                  f'#define i_key {container_key}',
+                                  f'#include "{stc_name + "/" + container[0]}.h"',
+                                  '#endif\n'))
```
EmilyBourne, line 1006: Given that you don't use the target is `if expr.target:` useful here?

pyccel/codegen/printing/ccode.py
```diff
@@ -1134,35 +1181,51 @@ def formatted_args_to_printf(args_format, args, end):
...
-        Find the corresponding C dtype in the dtype_registry.
+        Find the corresponding C dtype in the dtype_registry or the STC container type.
 
-        Find the corresponding C dtype in the dtype_registry.
-        Raise PYCCEL_RESTRICTION_TODO if not found.
+        This function searches for the corresponding C data type in the `dtype_registry`.
+        If the provided dtype is a container (like `HomogeneousSetType` or
+        `HomogeneousListType`), it identifies the scalar type of the container and returns
+        the appropriate type for the `STC` container.
+        It raises a `PYCCEL_RESTRICTION_TODO` error if the dtype is not found in the registry.
```
EmilyBourne, line 1192: Sorry, but I'm not sure this name and docstring make sense. A "dtype" as per the docstring definition:
https://github.com/pyccel/pyccel/blob/20aab161e912483e9e8a85fca771e21b2d432bbb/pyccel/ast/basic.py#L543-L552
cannot be a container.


I suggest:

```python
    def get_c_type(self, dtype):
        """
        Find the corresponding C type of the PyccelType.

        For scalar types, this function searches for the corresponding C data type
        in the `dtype_registry`.  If the provided type is a container (like
        `HomogeneousSetType` or `HomogeneousListType`),  it recursively identifies
        the type of an element of the container and uses it to calculate the
        appropriate type for the `STC` container.
        A `PYCCEL_RESTRICTION_TODO` error is raised if the dtype is not found in the registry.
```

pyccel/codegen/printing/ccode.py
```diff
@@ -1173,6 +1236,7 @@ def find_in_dtype_registry(self, dtype):
                     symbol = dtype,
                     severity='fatal')
 
+
     def find_in_ndarray_type_registry(self, dtype):
```
EmilyBourne, line 1240: Unnecessary extra space

pyccel/codegen/printing/ccode.py
```diff
@@ -1134,35 +1181,51 @@ def formatted_args_to_printf(args_format, args, end):
...
+        If the provided dtype is a container (like `HomogeneousSetType` or
+        `HomogeneousListType`), it identifies the scalar type of the container and returns
+        the appropriate type for the `STC` container.
+        It raises a `PYCCEL_RESTRICTION_TODO` error if the dtype is not found in the registry.
 
         Parameters
         ----------
-        dtype : DataType
-            The data type of the expression.
+        dtype : DataType or ContainerType
```
EmilyBourne, line 1196: `DataType` was replaced/renamed in #1756. It is now called `PyccelType`. Types like `HomogeneousListType` also inherit from `PyccelType` so there is no need to add `ContainerType` here

EmilyBourne (changes requested):
This looks very good but I still have a few final comments

pyccel/codegen/printing/ccode.py
Farouk-Echaref: When I was inspecting the `expr` object while working on the `Point_mod.py` file, I found that `expr` is a tuple of `ClassDef objects`, but that's only one case, is there a way it could take another iterable  ?

EmilyBourne: With the calls that you are currently using it will always be a tuple. But here you shouldn't be documenting how it is currently used, you should be describing what would be a valid way to use it.

In the future maybe I would like to only rename certain classes:
```python
renamed_classes = [c for c in expr.classes if not c.is_inline_class]
```
according to your docstring I would then need to do:
```python
self.rename_imported_methods(tuple(renamed_classes))
```
because I can't use a list directly right?

pyccel/codegen/printing/ccode.py
Farouk-Echaref: Ah, my bad. I think at first we were only passing the name attribute to this function, but after we updated it, we were passing the whole variable. The docstring didn't get updated afterward

Farouk-Echaref: it's actually the whole assignment variable

pyccel/codegen/printing/ccode.py
Farouk-Echaref: makes sense 
will be updated in my next commit

pyccel/codegen/printing/ccode.py
Farouk-Echaref: updated 8b4fa11b2c6dce02e30c77cef4d60da0d58e9315

pyccel/codegen/printing/ccode.py
Farouk-Echaref: updated 8b4fa11b2c6dce02e30c77cef4d60da0d58e9315

pyccel/codegen/printing/ccode.py
Farouk-Echaref: updated 8b4fa11b2c6dce02e30c77cef4d60da0d58e9315

pyccel/codegen/printing/ccode.py
Farouk-Echaref: updated 8b4fa11b2c6dce02e30c77cef4d60da0d58e9315

pyccel/codegen/printing/ccode.py
Farouk-Echaref: updated 8b4fa11b2c6dce02e30c77cef4d60da0d58e9315

pyccel/codegen/printing/ccode.py
Farouk-Echaref: updated 8b4fa11b2c6dce02e30c77cef4d60da0d58e9315

pyccel/codegen/printing/ccode.py
Farouk-Echaref: actually it's not, it's removed now 
4992779c9c11524fdb9495a0db9fc8d5e3287bfa

pyccel/codegen/printing/ccode.py
Farouk-Echaref: fixed 5c72144c6a07a1f8cd40885d15e1da677993abc0

pyccel/codegen/printing/ccode.py
```diff
@@ -1134,35 +1182,52 @@ def formatted_args_to_printf(args_format, args, end):
...
         -------
         str
-            The code which declares the datatype in C.
+            The code which declares the data type in C or the corresponding `STC` container
+            type.
+
+        Raises
+        ------
+        PYCCEL_RESTRICTION_TODO
+            If the dtype is not found in the dtype_registry.
```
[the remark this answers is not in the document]
Farouk-Echaref: so `PyccelCodegenError` instead of `PYCCEL_RESTRICTION_TODO`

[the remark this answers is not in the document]
Farouk-Echaref: updated c381e60cad0b9cc554ac36c1a170f2b796f011e5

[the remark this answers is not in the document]
EmilyBourne: Yes sorry. This type is useful because it tells you that you can do:
```python
try:
    ctype = self.get_c_type(my_type)
except PyccelCodegenError:
    ctype = 'void*'
```

EmilyBourne (approved):
Thank you very much for this crucial PR. Really good work!

PR #1848, opened 2024-04-23
list declare in c: prototype

### LINKED ISSUE #1655: [LISTS] Support variable declarations in C

Farouk-Echaref:
draft for list declaration in C
this PR fixes #1655


pyccel/codegen/printing/ccode.py
```diff
@@ -821,6 +821,10 @@ def _print_Module(self, expr):
         self.exit_scope()
         return code
 
+    def _print_PythonList(self, expr):
+        vec_declare = 'vec_' + expr.dtype._name
```
EmilyBourne, line 825: Beware I don't think you have the right function name here

EmilyBourne: It depends on what exactly you're trying to print, but it looks like it should be: `_print_HomogeneousListType`

But I don't know if `_print_Declare` uses `_print` to print the types anyway

EmilyBourne: The class `PythonList` represents an object such as `[1,2,3]` not a type

PR #1855, opened 2024-04-29
Simplify `epyccel` import

### LINKED ISSUE #1836: Allow the user to do `from pyccel import epyccel`
### Describe the feature
I would like to be able to import `epyccel` directly from Pyccel.

### Test Code
```python
from pyccel import epyccel

def func():
     return 1

new_func = epyccel(func)
```

### Proposed Solution
The file `epyccel.py` should be moved to `pyccel/commands/epyccel.py` and `pyccel/__init__.py` should be modified to allow importing `epyccel` directly

### Pitfalls
This change will make the old code `from pyccel.epyccel import epyccel` invalid. Unfortunately it is not possible for the 2 syntaxes to exist. This change should therefore be made in the new major version (which is in preparation).

EmilyBourne:
Move `epyccel.py` to `pyccel/commands/` folder. This allows a shortcut to be introduced for the import of the most commonly used `epyccel` function. It is now possible to do `from pyccel import epyccel`. Fixes #1836.


EmilyBourne:
Blocked by #1856

tests/epyccel/test_loops.py
```diff
@@ -2,7 +2,7 @@
 import pytest
 import numpy as np
 
-from pyccel.epyccel import epyccel
+from pyccel import epyccel
 from modules        import loops
```
yguclu, line 6: Alignment is broken, so we can just remove the blank spaces
```suggestion
from pyccel import epyccel
from modules import loops
```

tests/epyccel/test_mpi_point_to_point.py
```diff
@@ -3,7 +3,7 @@
 import numpy as np
 import pytest
 
-from pyccel.epyccel import epyccel
+from pyccel import epyccel
 from modules        import mpi_point_to_point as pmod
```
yguclu, line 7:
```suggestion
from pyccel import epyccel
from modules import mpi_point_to_point as pmod
```

yguclu (changes requested):
Nice!

tests/epyccel/test_loops.py
```diff
@@ -2,8 +2,8 @@
 import pytest
 import numpy as np
 
-from pyccel.epyccel import epyccel
-from modules        import loops
+from pyccel import epyccel
+from modules import loops
```
yguclu: Python linting:
```suggestion
from modules import loops

from pyccel import epyccel
```

tests/epyccel/test_mpi_point_to_point.py
```diff
@@ -3,8 +3,8 @@
 import numpy as np
 import pytest
 
-from pyccel.epyccel import epyccel
-from modules        import mpi_point_to_point as pmod
+from pyccel import epyccel
+from modules import mpi_point_to_point as pmod
```
yguclu: Python linting:
```suggestion
from modules import mpi_point_to_point as pmod

from pyccel import epyccel
```

yguclu (approved):

PR #1862, opened 2024-05-07
Create deploy workflow

EmilyBourne:


ci_tools/bot_tools/github_api_interactions.py
```diff
@@ -799,6 +814,67 @@ def request_reviewers(self, pr_id, request_team = False, reviewers = ()):
...
+
+        request_result = self._post_request("GET", url, params={'head_sha': commit_sha}).json()
+        workflow_runs = [j for j in request_result['workflow_runs'] if j['name'] in run_names]
+        completed = all(j['status'] == 'completed' for j in workflow_runs)
+
+        print(commit_sha)
+        print(workflow_runs)
+        print([j['name'] for j in workflow_runs])
+        print([j['status'] for j in workflow_runs])
+        print([j['conclusion'] for j in workflow_runs])
```
yguclu, line 860: Do we need the prints?

ci_tools/bot_tools/github_api_interactions.py
```diff
@@ -799,6 +814,67 @@ def request_reviewers(self, pr_id, request_team = False, reviewers = ()):
...
+
+        while not completed and time.time() < timeout:
+            time.sleep(15)
+            request_result = self._post_request("GET", url, params={'head_sha': commit_sha}).json()
+            workflow_runs = [j for j in request_result['workflow_runs'] if j['name'] in run_names]
+            completed = all(j['status'] == 'completed' for j in workflow_runs)
+            print(workflow_runs)
+            print([j['name'] for j in workflow_runs])
+            print([j['status'] for j in workflow_runs])
+            print([j['conclusion'] for j in workflow_runs])
```
yguclu, line 870: And these?

Maybe you can add a `debug` flag to this function, which will add the print statements.

ci_tools/bot_tools/github_api_interactions.py
EmilyBourne: Technically no, they were added for debugging purposes. But without them if there is a problem with this code we will have no clues as to what may be causing it. And as it only runs on the `main` branch it will be hard to put them back later. In theory we can test on a fork but if there is a repo-specific problem it will be impossible to debug without these prints.

Given that we will never open the job where these will be printed unless there is a problem I don't see any harm in leaving them

ci_tools/bot_tools/github_api_interactions.py
EmilyBourne: Same answer as above. I can add a `debug` flag but it will be difficult to activate the flag on the main branch in case of problems. The debug flag could always be added later when the code has been run a few times on this branch if the code is needed for elsewhere where the prints could be annoying

yguclu (approved):

PR #1870, opened 2024-05-10
Fix Wrapper Documentation.

mustapha-belbiad:
Correct a small issue in the wrapper documentation : 
From : 
```python
def _print_ClassName(self, stmt):
    ...
    return Y
```
To : 
```python
def _wrap_ClassName(self, stmt):
    ...
    return Y
```


smazouz42 (approved):
approved good job Mustapha"

EmilyBourne (approved):

PR #1871, opened 2024-05-13
Fix minor doc bot issue (6 lines)

EmilyBourne:
While reviewing https://github.com/pyccel/pyccel-cuda/pull/49 I noticed that the bot does not check `_wrap_X` docstring formats. We decided that wrap functions should be documented, so they should be documented correctly. This PR stops ignoring these functions and makes minor improvements to the associated file by using f-strings instead of join.


smazouz42 (approved):
looks good to me

jalalium (approved):

PR #1899, opened 2024-05-22
Use Python version to make artifact name unique (1 line)

EmilyBourne:
The updated version of `upload-artifact` means that you cannot upload multiple artifacts with the same name. This is problematic on the `devel` branch as the same job is run for multiple Python versions. The easy fix for this is to add the version to the artifact name. Local tests show that the artifacts can still be retrieved with a partial name.


EmilyBourne:
This method has been successfully tested in [pyccel/pyccel-cuda](https://github.com/pyccel/pyccel-cuda)

PR #1901, opened 2024-05-24
Add Python Support for set initialisation with `set()`

### LINKED ISSUE #1893: Add Python support for set() function
### Describe the feature
I would like to be able to use the `set()` function. This is important so that empty sets can be created. The `set()` function is different to the `{,}` function as it takes an iterable.

### Test Code
```python
if __name__ == '__main__':
    a = set([1,2,3])
    b = [4,5,6]
    c = set(b)
    d = set(c)
```

### Proposed Solution
This feature should be implemented similarly to what was done for tuples. The equivalent issues for tuples was resolved in #1589.
The pertinent code is:
https://github.com/pyccel/pyccel/blob/d64be7ff7072faa07ce19c9e70fa02ed45a25c9c/pyccel/ast/builtins.py#L596-L620

### Additional context
It is important to handle the `set` function so that this function can be used for type annotations. If the `{,}` function is used bugs such as #1892 can arise.

mustapha-belbiad:
Add support for the `set()` function in Python. This fixes #1893. 
In this PR, the `PythonSetFunction` class is introduced to handle calls to the `set()` function, along with the addition of related tests.

### DIFF
--- a/pyccel/ast/builtins.py
+++ b/pyccel/ast/builtins.py
@@ -803,9 +803,9 @@ class PythonSetFunction(PyccelFunction):
             return super().__new__(cls)
         else:
             raise TypeError(f"Can't unpack {arg} into a Set")
-    
+
     def __init__(self, copied_obj):
-        self._class_type = copied_obj.class_type 
+        self._class_type = copied_obj.class_type
         self._shape = copied_obj.shape
         super().__init__(copied_obj)
 
--- a/pyccel/ast/builtins.py
+++ b/pyccel/ast/builtins.py
@@ -776,7 +776,7 @@ class PythonSet(TypedAstNode):
         return True
 
 
-class PythonSetFunction(TypedAstNode):
+class PythonSetFunction(PyccelFunction):
     """
     Class representing a call to the `set` function.
 
@@ -789,9 +789,9 @@ class PythonSetFunction(TypedAstNode):
     arg : TypedAstNode
         The argument passed to the function call.
     """
-    __slots__ = ()
-    _attribute_nodes = ()
 
+    __slots__ = ('_shape', '_class_type')
+    name = 'set'
     def __new__(cls, arg):
         if isinstance(arg.class_type, HomogeneousSetType):
             return arg
@@ -799,8 +799,15 @@ class PythonSetFunction(TypedAstNode):
             return PythonSet(*arg)
         elif isinstance(arg.shape[0], LiteralInteger):
             return PythonSet(*[arg[i] for i in range(arg.shape[0])])
+        elif isinstance(arg.shape[0], PyccelArrayShapeElement):
+            return super().__new__(cls)
         else:
             raise TypeError(f"Can't unpack {arg} into a Set")
+    
+    def __init__(self, copied_obj):
+        self._class_type = copied_obj.class_type 
+        self._shape = copied_obj.shape
+        super().__init__(copied_obj)
 
 #==============================================================================
 class PythonMap(PyccelFunction):
--- a/pyccel/ast/builtins.py
+++ b/pyccel/ast/builtins.py
@@ -795,10 +795,8 @@ class PythonSetFunction(PyccelFunction):
     def __new__(cls, arg):
         if isinstance(arg.class_type, HomogeneousSetType):
             return arg
-        elif isinstance(arg, PythonList):
+        elif isinstance(arg, (PythonList, PythonSet, PythonTuple)):
             return PythonSet(*arg)
-        elif isinstance(arg.shape[0], LiteralInteger):
-            return PythonSet(*[arg[i] for i in range(arg.shape[0])])
         elif isinstance(arg.shape[0], PyccelArrayShapeElement):
             return super().__new__(cls)
         else:
--- a/pyccel/ast/builtins.py
+++ b/pyccel/ast/builtins.py
@@ -797,10 +797,8 @@ class PythonSetFunction(PyccelFunction):
             return arg
         elif isinstance(arg, (PythonList, PythonSet, PythonTuple)):
             return PythonSet(*arg)
-        elif isinstance(arg.shape[0], PyccelArrayShapeElement):
-            return super().__new__(cls)
         else:
-            raise TypeError(f"Can't unpack {arg} into a Set")
+            return super().__new__(cls)
 
     def __init__(self, copied_obj):
         self._class_type = copied_obj.class_type
--- a/pyccel/ast/builtins.py
+++ b/pyccel/ast/builtins.py
@@ -50,6 +50,7 @@ __all__ = (
     'PythonRange',
     'PythonReal',
     'PythonSet',
+    'PythonSetFunction',
     'PythonSum',
     'PythonTuple',
     'PythonTupleFunction',
@@ -774,6 +775,33 @@ class PythonSet(TypedAstNode):
         """
         return True
 
+
+class PythonSetFunction(TypedAstNode):
+    """
+    Class representing a call to the `set` function.
+
+    Class representing a call to the `set` function. This is
+    different to the `{,}` syntax as it only takes one argument
+    and unpacks any variables.
+
+    Parameters
+    ----------
+    arg : TypedAstNode
+        The argument passed to the function call.
+    """
+    __slots__ = ()
+    _attribute_nodes = ()
+
+    def __new__(cls, arg):
+        if isinstance(arg.class_type, HomogeneousSetType):
+            return arg
+        elif isinstance(arg, PythonList):
+            return PythonSet(*arg)
+        elif isinstance(arg.shape[0], LiteralInteger):
+            return PythonSet(*[arg[i] for i in range(arg.shape[0])])
+        else:
+            raise TypeError(f"Can't unpack {arg} into a Set")
+
 #==============================================================================
 class PythonMap(PyccelFunction):
     """
@@ -1234,4 +1262,5 @@ builtin_functions_dict = {
     'str'      : LiteralString,
     'type'     : PythonType,
     'tuple'    : PythonTupleFunction,
+    'set'      : PythonSetFunction
 }
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -146,6 +146,12 @@ import pyccel.decorators as def_decorators
 errors = Errors()
 pyccel_stage = PyccelStage()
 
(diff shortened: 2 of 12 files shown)


tests/epyccel/test_epyccel_sets.py
```diff
@@ -283,3 +283,50 @@ def update_tuple_as_arg():
     pyccel_result = epyccel_update()
     python_result =  update_tuple_as_arg()
     assert python_result == pyccel_result
+
+def test_set_with_list(language):
+    def set_With_list():
+        a = [1.6, 6.3, 7.2]
+        b = set(a)
```
EmilyBourne, line 290: With your current code, copies like this won't work if `a` is an argument (unknown length). Is this intentional?

mustapha-belbiad: Can you give an example, please?

EmilyBourne:
```python
def f(a : list[float]):
    b = set(a)
```

EmilyBourne: If I try to translate this function I get:
```

pyccel:
 |fatal [semantic]: tmp_set_copy.py [1,10]| Cannot handle non-homogenous type index
Pyccel has encountered syntax that has not been implemented yet. Please create an issue at https://github.com/pyccel/pyccel/issues and provide a small example of your problem. Do not forget to specify your target language. (list[float])
```

mustapha-belbiad: Fixed : 39e344c

Farouk-Echaref (approved):
Looks good and thoroughly tested!

pyccel/ast/builtins.py
```diff
@@ -774,6 +775,40 @@ def is_homogeneous(self):
...
+
+    __slots__ = ('_shape', '_class_type')
+    name = 'set'
+    def __new__(cls, arg):
+        if isinstance(arg.class_type, HomogeneousSetType):
+            return arg
+        elif isinstance(arg, PythonList):
+            return PythonSet(*arg)
+        elif isinstance(arg.shape[0], LiteralInteger):
+            return PythonSet(*[arg[i] for i in range(arg.shape[0])])
```
EmilyBourne, line 801: A `PythonList` has a shape which is a `LiteralInteger`. Why do you feel that it is important to treat the case of a `PythonList` individually but not a `PythonTuple` or a `PythonSet`?

pyccel/ast/builtins.py
```diff
@@ -774,6 +775,40 @@ def is_homogeneous(self):
...
+        elif isinstance(arg.shape[0], PyccelArrayShapeElement):
+            return super().__new__(cls)
+        else:
+            raise TypeError(f"Can't unpack {arg} into a Set")
```
EmilyBourne, line 805: Why do you require the shape to be a `PyccelArrayShapeElement`?
I am concerned that this may exclude the following code (or may if we modify slightly how shapes are changed):
```python
def f():
    n : 'const int' = 5
    a : 'const float[:]' = np.ones(n+1)
```
And I don't see a justification for this

pyccel/parser/semantic.py
```diff
@@ -1932,6 +1932,8 @@ def _get_indexed_type(self, base, args, expr):
                     class_type = HomogeneousTupleType
                 elif dtype_cls is PythonList:
                     class_type = HomogeneousListType
+                elif dtype_cls is PythonSetFunction:
+                    class_type = HomogeneousSetType
```
EmilyBourne, line 1936: What do you think about using a dictionary here? I guess it should be more compact and faster than a series of if conditions

tests/epyccel/test_epyccel_sets.py
```diff
@@ -283,3 +283,84 @@ def update_tuple_as_arg():
...
+
+    epyc_set_With_set = epyccel(set_With_set, language = language)
+    pyccel_result = epyc_set_With_set()
+    python_result = set_With_set()
+    assert isinstance(python_result, type(pyccel_result))
+    assert python_result == pyccel_result
+
+def test_set_kwarg_init(language):
+    def kwarg_init():
+        b = set({4.6, 7.9, 2.5})
```
EmilyBourne, line 325: What is the relevance of keyword arguments here?

tests/epyccel/test_epyccel_sets.py
```diff
@@ -283,3 +283,84 @@ def update_tuple_as_arg():
...
+
+    epyc_kwarg_init = epyccel(kwarg_init, language = language)
+    pyccel_result = epyc_kwarg_init()
+    python_result = kwarg_init()
+    assert isinstance(python_result, type(pyccel_result))
+    assert python_result == pyccel_result
+
+def test_set_kwarg_init_list(language):
+    def kwarg_init():
+        b = set([4.6, 7.9, 2.5])
```
EmilyBourne, line 336: What is the relevance of keyword arguments here?

tests/epyccel/test_epyccel_sets.py
```diff
@@ -283,3 +283,84 @@ def update_tuple_as_arg():
...
+
+    epyc_kwarg_init = epyccel(kwarg_init, language = language)
+    pyccel_result = epyc_kwarg_init()
+    python_result = kwarg_init()
+    assert isinstance(python_result, type(pyccel_result))
+    assert python_result == pyccel_result
+
+
+def test_set_copie_from_arg1(language):
+    def copie_from_arg1(a : 'list[float]'):
```
EmilyBourne, line 347: Typo "copy" in English

tests/epyccel/test_epyccel_sets.py
```diff
@@ -283,3 +283,84 @@ def update_tuple_as_arg():
...
+        return b
+    a = [2.5, 1.4, 9.2]
+    epyc_copie_from_arg = epyccel(copie_from_arg1, language = language)
+    pyccel_result = epyc_copie_from_arg(a)
+    python_result = copie_from_arg1(a)
+    assert isinstance(python_result, type(pyccel_result))
+    assert python_result == pyccel_result
+
+def test_set_copie_from_arg2(language):
+    def copie_from_arg2(a : 'set[float]'):
```
EmilyBourne, line 358: Typo "copy" in English

EmilyBourne (changes requested):

pyccel/ast/builtins.py
mustapha-belbiad: Do you think it should be like this: 
```suggestion
        elif isinstance(arg, (PythonList, PythonSet, PythonTuple)):
            return PythonSet(*arg))
```

pyccel/parser/semantic.py
mustapha-belbiad: That sounds like a good idea. I will implement it.

pyccel/ast/builtins.py
mustapha-belbiad: I used `PyccelArrayShapeElement` to handle `an argument (unknown length)`.

tests/epyccel/test_epyccel_sets.py
mustapha-belbiad: fixed : 2c3e4ff

tests/epyccel/test_epyccel_sets.py
mustapha-belbiad: fixed : 2c3e4ff

tests/epyccel/test_epyccel_sets.py
mustapha-belbiad: fixed : 2c3e4ff

pyccel/parser/semantic.py
mustapha-belbiad: Fixed : 7fa1e49

pyccel/ast/builtins.py
EmilyBourne: Actually I was thinking of:
```suggestion
        elif isinstance(arg.shape[0], LiteralInteger):
            return PythonSet(*[arg[i] for i in range(arg.shape[0])])
```
which should be the most general and keep working for future types, but it also depends on what translation you want for `Variable`s of known length

pyccel/ast/builtins.py
EmilyBourne: What do you mean by "unknown length"? In the example above the length of `a` is `n+1`. Is this a known length?

pyccel/ast/builtins.py
EmilyBourne: I would be fine with either solution

mustapha-belbiad: I got an error when i tried this :
```python 
k = set([1,2,3,4])
```
This is the error: `|fatal [semantic]: test.py [2,4]| 'PythonList' object is not subscriptable (set((1, 2, 3, 4)))`

pyccel/ast/builtins.py
mustapha-belbiad: I was talking about the reason for using `PyccelArrayShapeElement`, Like this example : 
```python
def f(a : list[int]):
    d = set(a)
```
Can you please explain to me the logic behind the set and the example that you provided.

pyccel/ast/builtins.py
mustapha-belbiad: Fixed : 4576d39

pyccel/ast/builtins.py
EmilyBourne: The more complete example is:
```python
def f(n : 'const int'):
    a : 'const float[:]' = np.ones(n+1)
    b = set(a)
```
historically in this case the shape of `a` would have been `(PyccelAdd(Variable(n), LiteralInteger(1)),)`. I suspect that currently the shape would simply be `(PyccelArrayShapeElement('a',0),)` but at some point I would like to go back to the more explicit shape.

You said that "I used PyccelArrayShapeElement to handle an argument (unknown length).". This implies that a `PyccelArrayShapeElement` is the only way to describe an "unknown length". Hence my question
> What do you mean by "unknown length"? In the example above the length of a is n+1. Is this a known length?

mustapha-belbiad: Like I mentioned before, the reason I used `PyccelArrayShapeElement` is for the example that I provided, which explains the `unknown length`. If you have a better solution, I would appreciate it.

mustapha-belbiad: Fixed : 28479a7

EmilyBourne (approved):

PR #1904, opened 2024-05-28
Store annotated local variables to avoid memory leaks

### LINKED ISSUE #1903: Memory leak when using type annotations on local variables
### Describe the bug
When a local variable is annotated and defined in the same line the associated deallocation is missing.

### To Reproduce
```python
import numpy as np

def foo():
    a : 'float[:]' = np.ones(5)
```

### Error details
Provide the generated code:
```c
#include "tmp_new_iss.h"
#include <stdlib.h>
#include "ndarrays.h"
#include <stdint.h>


void foo(void)
{
    a = array_create(1, (int64_t[]){INT64_C(5)}, nd_double, false, order_c);
    array_fill((double)1.0, a);
}
```

### Expected behavior
The code should be:
```c
#include "tmp_new_iss.h"
#include <stdlib.h>
#include "ndarrays.h"
#include <stdint.h>


void foo(void)
{
    a = array_create(1, (int64_t[]){INT64_C(5)}, nd_double, false, order_c);
    array_fill((double)1.0, a);
    free_array(&a);
}
```

### Language
The deallocation node is missing in all languages but it only causes a memory leak in C.

### Additional context

Deallocation nodes are added in the function:
https://github.com/pyccel/pyccel/blob/5a75f34563bd1ea3287afc29949bca98400be6fe/pyccel/parser/semantic.py#L535-L552
Nodes are created for variables that appear in the `_allocs` variable:
https://github.com/pyccel/pyccel/blob/5a75f34563bd1ea3287afc29949bca98400be6fe/pyccel/parser/semantic.py#L227-L228

Normally this list of sets is filled in the function:
https://github.com/pyccel/pyccel/blob/5a75f34563bd1ea3287afc29949bca98400be6fe/pyccel/parser/semantic.py#L1342-L1380

This is done here:
https://github.com/pyccel/pyccel/blob/5a75f34563bd1ea3287afc29949bca98400be6fe/pyccel/parser/semantic.py#L1505-L1518
when the first allocation is created. We know that we are looking at the first allocation because the variable doesn't yet exist:
https://github.com/pyccel/pyccel/blob/5a75f34563bd1ea3287afc29949bca98400be6fe/pyccel/parser/semantic.py#L1408-L1409

However when a local variable is annotated and defined in the same line the variable already exists. We therefore don't reach this code. Instead this function is called:
https://github.com/pyccel/pyccel/blob/5a75f34563bd1ea3287afc29949bca98400be6fe/pyccel/parser/semantic.py#L1559-L1584

As part of #1524 an extra condition was added to ensure that these variables are correctly allocated. This is done by checking if there are any existing allocation nodes for the variable:
https://github.com/pyccel/pyccel/blob/5a75f34563bd1ea3287afc29949bca98400be6fe/pyccel/parser/semantic.py#L1638-L1639
If none exist then the variable is identified as unallocated:
https://github.com/pyccel/pyccel/blob/5a75f34563bd1ea3287afc29949bca98400be6fe/pyccel/parser/semantic.py#L1682
At this point an `Allocate` node is created, however the variable is still missing from the `_allocs` variable. It should be added here

Tests for memory leaks can be found here:
https://github.com/pyccel/pyccel/blob/devel/tests/ndarrays/leaks_check.py

Farouk-Echaref:
Add annotated local variable to the `_allocs` variable so its deallocation node can be added in `_garbage_collector()`. This fixes #1903

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ All notable changes to this project will be documented in this file.
 -   #1853 : Fix translation of a file whose name conflicts with Fortran keywords.
 -   Link and mention `devel` branch, not `master`.
 -   #1047 : Print the value of an unrecognised constant.
+-   #1903 : Fix memory leak when using type annotations on local variables
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,7 @@ All notable changes to this project will be documented in this file.
 -   #1853 : Fix translation of a file whose name conflicts with Fortran keywords.
 -   Link and mention `devel` branch, not `master`.
 -   #1047 : Print the value of an unrecognised constant.
--   #1903 : Fix memory leak when using type annotations on local variables
+-   #1903 : Fix memory leak when using type annotations on local variables.
 
 ### Changed
 
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -1682,9 +1682,10 @@ class SemanticParser(BasicParser):
                         status = 'unallocated'
 
                     new_expressions.append(Allocate(var, shape=d_var['shape'], status=status))
-                    self._allocs[-1].add(var)
 
-                    if status != 'unallocated':
+                    if status == 'unallocated':
+                        self._allocs[-1].add(var)
+                    else:
                         errors.report(ARRAY_REALLOCATION, symbol=var.name,
                             severity='warning',
                             bounding_box=(self.current_ast_node.lineno,
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -1690,8 +1690,6 @@ class SemanticParser(BasicParser):
                             severity='warning',
                             bounding_box=(self.current_ast_node.lineno,
                                 self.current_ast_node.col_offset))
-                    elif status == 'unallocated':
-                        self._allocs[-1].add(var)
             elif previous_allocations and previous_allocations[-1].get_user_nodes(IfSection) \
                         and not previous_allocations[-1].get_user_nodes((If)):
                 # If previously allocated in If still under construction
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -1682,6 +1682,7 @@ class SemanticParser(BasicParser):
                         status = 'unallocated'
 
                     new_expressions.append(Allocate(var, shape=d_var['shape'], status=status))
+                    self._allocs[-1].add(var)
 
                     if status != 'unallocated':
                         errors.report(ARRAY_REALLOCATION, symbol=var.name,
--- a/tests/ndarrays/leaks_check.py
+++ b/tests/ndarrays/leaks_check.py
@@ -1,5 +1,5 @@
 # pylint: disable=missing-function-docstring, missing-module-docstring
-from numpy import array, zeros
+from numpy import array, zeros, ones
 
 def create_array():
     a = array([[1, 2, 3, 4, 5], [6, 7, 8, 9, 10]])# pylint: disable=unused-variable
@@ -45,6 +45,9 @@ def arrays_in_multi_returns():
     b = zeros(4)
     return a, b, 4
 
+def type_annotated_numpy_array():
+    a : 'float[:]' = ones(5) # pylint: disable=unused-variable
+
 # testing garbage collecting in a Function
 
 if __name__ == '__main__':
@@ -56,6 +59,7 @@ if __name__ == '__main__':
     conditional_alloc(True,True)
     conditional_alloc(True,False)
     conditional_alloc(False,False)
+    type_annotated_numpy_array()
 
     # testing garbage collecting in a Program
 


CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ All notable changes to this project will be documented in this file.
 -   #1853 : Fix translation of a file whose name conflicts with Fortran keywords.
 -   Link and mention `devel` branch, not `master`.
 -   #1047 : Print the value of an unrecognised constant.
+-   #1903 : Fix memory leak when using type annotations on local variables
```
EmilyBourne, line 39:
```suggestion
-   #1903 : Fix memory leak when using type annotations on local variables.
```

pyccel/parser/semantic.py
```diff
@@ -1682,6 +1682,7 @@ def _ensure_inferred_type_matches_existing(self, class_type, d_var, var, is_auga
                         status = 'unallocated'
 
                     new_expressions.append(Allocate(var, shape=d_var['shape'], status=status))
+                    self._allocs[-1].add(var)
```
EmilyBourne, line 1685: This is the wrong place for this line. The variable should only be added to `_allocs` if it's new (i.e. if the status is `unallocated`). The set should take care of this but it's better to be on the safe side

EmilyBourne (changes requested):

pyccel/parser/semantic.py
Farouk-Echaref: so the `else` block where `status` is assigned `unallocated`.
may I know what does this line do and what's `new_expression` used for:
```python
new_expressions.append(Allocate(var, shape=d_var['shape'], status=status))
```

CHANGELOG.md
Farouk-Echaref: fixed 
f06a5ac7

pyccel/parser/semantic.py
EmilyBourne: Sometimes while creating nodes you need to create more nodes than expected. For example when translating `a = np.ones(4)` you need 2 nodes not one. The nodes are `[Allocate(a, 4), Assign(a, 1)`. Or when calling a function with a vector expression e.g. `f(a+b)` you may need to create a temporary to store the argument in. The list `new_expression` is designed to hold these unexpected nodes that are created in the functions `_ensure_inferred_type_matches_existing` or `_assign_lhs_variable` neither of which are supposed to create nodes usually.

This line creates the necessary `Allocate` node to allocate the data. The purpose of the `Allocate` node is described in its docstring:
https://github.com/pyccel/pyccel/blob/5a75f34563bd1ea3287afc29949bca98400be6fe/pyccel/ast/core.py#L371-L399

Farouk-Echaref: check bacd8b3a and let me know if it's alright

Farouk-Echaref: Thanks for the explanation!

EmilyBourne: This looks fine but could be better. For execution speed it is always better to avoid `if` conditions. Here your new `elif` is unnecessary. You could just use `else`  or put the line in a slightly different location where the value has already been checked

Farouk-Echaref: I first wrote: 
```python
 else:
                        status = 'unallocated'
                        self._allocs[-1].add(var)
```
but didn't like how it looked so for more clarity i added the `elif` condition, but I understand the execution speed concerns

EmilyBourne: > but didn't like how it looked so for more clarity

That's fine. But why:
```python
                    if status != 'unallocated':
                        errors.report(ARRAY_REALLOCATION, symbol=var.name, ...)
                    elif status == 'unallocated':
                        self._allocs[-1].add(var)
```
instead of:
```python
                    if status != 'unallocated':
                        errors.report(ARRAY_REALLOCATION, symbol=var.name, ...)
                    else:
                        self._allocs[-1].add(var)
```

EmilyBourne: Or:
```python
                    if status == 'unallocated':
                        self._allocs[-1].add(var)
                    else:
                        errors.report(ARRAY_REALLOCATION, symbol=var.name, ...)
```

EmilyBourne: Unknown is for situations like:
```python
def f(b : bool):
   a : 'int[:]'
   if b:
       a = np.ones(4)

    a = np.zeros(4) # Does "a" need allocating or does it already exist with the correct size?
```

EmilyBourne: But as you say `'unknown' != 'unallocated'` so this shouldn't affect whether an `if` is needed here or not

Farouk-Echaref: > Or:
> 
> ```python
>                     if status == 'unallocated':
>                         self._allocs[-1].add(var)
>                     else:
>                         errors.report(ARRAY_REALLOCATION, symbol=var.name, ...)
> ```

this one better.

Farouk-Echaref: updated 4e91d41b

EmilyBourne (approved):
Looks great. Thanks

PR #1908, opened 2024-06-04
Add C and Fortran support for `len()` function

### LINKED ISSUE #1875: [SETS] Add C support for len function
### Describe the feature
In Python the len function can be called on many functions however the implementation of this function is different in different languages. Support should be added to test the length of a list in C (and Python if it does not work out of the box).

### Test Code
```python
def f():
    a = {1,2,1}
    b = len(a)
    return b
```

### Proposed Solution

Farouk-Echaref:
Add C and Fortran support for the `len()` function for `list` and `set` using the equivalent `container_X_size()` from the `STC` library and ` % size()` from the gFTL library. This PR fixes issues #1874 and #1875.

**Commit Summary**
- Reorganise CHANGELOG to group related points
- Update the `PythonLen` docstring. 
- Update `FortranCodePrinter._print_PyccelArrayShapeElement` method to print ` % size()` for containers.
- Update `CCodePrinter._print_PyccelArrayShapeElement` method to check whether the element passed is a `list` or `set`, and then print their equivalent using the `STC`.
- Update `shape_can_change` method to return `True` only for `HomogeneousListType` and `HomogeneousSetType` objects as their shape can change.
- Add tests for the `len()` function using `list` and `set` with different data types.


tests/epyccel/test_builtins.py
```diff
@@ -369,3 +369,72 @@ def f():
...
+
+def test_len_set_complex(stc_language):
+    def f():
+        a = {1j, 2 + 1j, 3 + 1j}
+        b = len(a)
+        return b
+
+    epyc_f = epyccel(f, language=stc_language)
+
+    assert epyc_f() == f()
```
jalalium, line 440: should we add more tests? especially for nested lists?

jalalium (changes requested):

pyccel/ast/builtins.py
```diff
@@ -629,6 +629,12 @@ class PythonLen(PyccelFunction):
     (aka the first element of the shape) of an object. This can usually
     be calculated in the generated code, but in an inhomogeneous object
     the integer value of the shape must be returned.
+    However, when dealing with variables that are passed to `len()` and
+    contain lists or sets, the shape is unknown and cannot be determined
```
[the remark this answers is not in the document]
Farouk-Echaref: updated 22495f63032a9ec44b1d68715cbaade0e3b55f20

tests/codegen/fcode/test_fcode_codegen.py
```diff
@@ -13,14 +13,16 @@
...
-files = [os.path.join(path_dir,f) \
-         #if f not in failing_files \
-         #else pytest.param(os.path.join(path_dir,f), marks = pytest.mark.xfail(reason=failing_files[f])) \
-         for f in files \
-         if f.endswith(".py") \
-        ]
+files = [
+    pytest.param(
+        os.path.join(path_dir, f),
+        marks=pytest.mark.xfail(reason="Pyccel crash due to list Initialization with the use of `len()`, related issue #1924.")
```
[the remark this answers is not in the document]
Farouk-Echaref: updated 29683c8dcd531847099cea585a85ce969e046797

tests/epyccel/test_builtins.py
```diff
@@ -369,3 +369,72 @@ def f():
     epyc_f = epyccel(f, language=language)
 
     assert epyc_f() == f()
+
+@pytest.fixture( params=[
+        pytest.param("c", marks = pytest.mark.c),
+        pytest.param("fortran", marks = [
+            pytest.mark.xfail(reason="Set declaration in Fortran is not yet implemented. related issue #1658"),
```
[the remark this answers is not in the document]
Farouk-Echaref: you are right
updated 4e0c11365addf0a91f488cf9747f0b9cf55a338a

tests/epyccel/test_builtins.py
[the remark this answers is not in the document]
Farouk-Echaref: It was based on this previous suggestion:
https://github.com/pyccel/pyccel/pull/1847/files/9872a5b3c8f24ec251ca34db38b8355bef29af4a#r1630750905

tests/epyccel/test_builtins.py
Farouk-Echaref: Im really not sure about the handling of nested lists, if they weren't supported by default in the previous PR, we can open an issue for them asap

EmilyBourne: Nested lists should be supported. I believe they are working for Python. Nested sets won't be supported (at least for now)

EmilyBourne: I am working on nested lists. They are working for Python but there are some problems in C so it is probably not a good idea to add lots of extra tests for this immediately. In any case the way the `len` function works does not change depending on the element type so I don't think that lots of nested list tests would necessarily add much

EmilyBourne:
@Farouk-Echaref Do you plan to complete this PR or should I pick it up and complete it?

Farouk-Echaref:
> @Farouk-Echaref Do you plan to complete this PR or should I pick it up and complete it?

Hello Emily, I'm planning to continue working on my Pyccel issues starting next month hopefully.

jalalium (changes requested):
Shall we add tests for the `len` of `stc` containers?

EmilyBourne:
> Shall we add tests for the `len` of `stc` containers?

What do you mean? Is that not what the tests that Farouk wrote do?

jalalium:
> > Shall we add tests for the `len` of `stc` containers?
> 
> What do you mean? Is that not what the tests that Farouk wrote do?

Nevermind. Sorry!

jalalium (dismissed):
looks good to me!

tests/epyccel/test_builtins.py
```diff
@@ -369,3 +369,71 @@ def f():
...
+        return len(a)
+
+    epyc_f = epyccel(f, language=language)
+
+    assert epyc_f() == f()
+
+def test_len_list_float(language):
+    def f():
+        a = [1.4, 2.6, 3.5]
+        b = len(a)
```
jalalium, line 385: why  use the extra  variable?

tests/epyccel/test_builtins.py
```diff
@@ -369,3 +369,71 @@ def f():
...
+        return b
+
+    epyc_f = epyccel(f, language=language)
+
+    assert epyc_f() == f()
+
+def test_len_list_complex(language):
+    def f():
+        a = [1j, 2 + 1j, 3 + 1j]
+        b = len(a)
```
jalalium, line 395: why use the extra variable?

tests/epyccel/test_builtins.py
```diff
@@ -369,3 +369,71 @@ def f():
...
+        return len(a)
+
+    epyc_f = epyccel(f, language=stc_language)
+
+    assert epyc_f() == f()
+
+def test_len_set_float(stc_language):
+    def f():
+        a = {1.4, 2.6, 3.5}
+        b = len(a)
```
jalalium, line 414: why use the extra variable?

tests/epyccel/test_builtins.py
```diff
@@ -369,3 +369,71 @@ def f():
...
+        return b
+
+    epyc_f = epyccel(f, language=stc_language)
+
+    assert epyc_f() == f()
+
+def test_len_set_complex(stc_language):
+    def f():
+        a = {1j, 2 + 1j, 3 + 1j}
+        b = len(a)
```
jalalium, line 424: why use the extra variable?

tests/epyccel/test_builtins.py
```diff
@@ -369,3 +369,71 @@ def f():
...
+        return b
+
+    epyc_f = epyccel(f, language=stc_language)
+
+    assert epyc_f() == f()
+
+def test_len_dict_int_float(stc_language):
+    def f():
+        a = {1:1.0, 2:2.0, 3:3.0, 4:4.0}
+        b = len(a)
```
jalalium, line 434: why use the extra variable?

jalalium (approved):
Very good job!

tests/epyccel/test_builtins.py
EmilyBourne: No particular reason. This just ensures that I test `len` in multiple different settings. There are functions above which use `return len(a)` so this case is also tested.

Issue #1877, opened 2024-05-16
[SETS] Add C support for pop()


EmilyBourne:
### Test Code
Provide code which does not currently work but which should do when this issue is fixed:
```python
def foo():
    a = {1,2,1}
    b = a.pop()
    c = a.pop()
    return b,c
```

### Proposed Solution
This feature should be implemented using the functionalities provided by STC

EmilyBourne:
Fixed by #1912

PR #1912, opened 2024-06-12
[SETS] Add C support for `pop()`.

mustapha-belbiad:
This Pull Request adds support for the pop() method in C, addressing issue #1877.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ All notable changes to this project will be documented in this file.
 -   \[INTERNALS\] Added `container_rank` property to `ast.datatypes.PyccelType` objects.
 -   \[DEVELOPER\] Added an improved traceback to the developer-mode errors for errors in function calls.
 -   #1893 : Add Python support for set initialisation with `set()`.
+-   #1877 : App C Support for set method `pop()`.
 
 ### Fixed
 
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -1005,7 +1005,9 @@ class CCodePrinter(CodePrinter):
                               f'#define i_key {container_key}',
                               f'#include "{stc_name + "/" + container[0]}.h"',
                               '#endif\n'))
-
+        elif source.startswith('Set_pop'):
+            name , i_type, i_key = source.split('/')
+            return f'\n\n{name}({i_type}, {i_key});\n\n'
         # Get with a default value is not used here as it is
         # slower and on most occasions the import will not be in the
         # dictionary
@@ -2183,6 +2185,12 @@ class CCodePrinter(CodePrinter):
             return prefix_code+self.init_stc_container(rhs, expr)
         rhs = self._print(expr.rhs)
         return prefix_code+'{} = {};\n'.format(lhs, rhs)
+
+    def _print_SetPop(self, expr):
+        self.add_import(Import(f'STC_Extensions/Set_extensions', Module(f'STC_Extensions/Set_extensions', (), ())))
+        self.add_import(Import(f'Set_pop_macro/hset_int64_t/int64_t', Module(f'Set_pop_macro/hset_int64_t/int64_t', (), ())))
+        set_var = expr.set_variable.name
+        return f'Set_pop(&{set_var})'
 
     def _print_AliasAssign(self, expr):
         lhs_var = expr.lhs
--- a/pyccel/codegen/utilities.py
+++ b/pyccel/codegen/utilities.py
@@ -29,7 +29,8 @@ language_extension = {'fortran':'f90', 'c':'c', 'python':'py'}
 
 #==============================================================================
 # map external libraries inside pyccel/extensions with their path
-external_libs = {"stc"  : "STC/include"}
+external_libs = {"stc"  : "STC/include",
+                 "STC_Extensions" : "."}
 
 #==============================================================================
 # map internal libraries to their folders inside pyccel/stdlib and their compile objects
--- a/pyccel/extensions/STC_Extensions/Set_extensions.h
+++ b/pyccel/extensions/STC_Extensions/Set_extensions.h
@@ -1,0 +1,12 @@
+#ifndef EXTENSIONS_H
+#define EXTENSIONS_H
+
+
+#define Set_pop_macro(type, key)\
+key Set_pop(type *l){\
+    type##_iter itr = type##_begin(*&l);\
+    type##_erase_at(*&l, itr);\
+    return(*itr.ref);\
+}
+
+#endif


CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ All notable changes to this project will be documented in this file.
 -   \[INTERNALS\] Added `container_rank` property to `ast.datatypes.PyccelType` objects.
 -   \[DEVELOPER\] Added an improved traceback to the developer-mode errors for errors in function calls.
 -   #1893 : Add Python support for set initialisation with `set()`.
+-   #1877 : Add C Support for set method `pop()`.
```
EmilyBourne, line 27: Please move these to be grouped with the other set issues in the section for users (before INTERNALS/DEVELOPERS)

pyccel/codegen/printing/ccode.py
```diff
@@ -1005,7 +1005,10 @@ def _print_Import(self, expr):
                               f'#define i_key {container_key}',
                               f'#include "{stc_name + "/" + container[0]}.h"',
                               '#endif\n'))
-
+        elif source.startswith('Set_pop'):
+            _ , i_type, i_key = source.split('/')
+            return '\n'.join((f'#define i_type {i_type}',
+                   f'#define i_key {i_key}\n',))
```
EmilyBourne, line 1011: Is this sufficient? You don't actually need to import anything? Please leave a comment on this PR with an example code and its translation

pyccel/codegen/printing/ccode.py
```diff
@@ -2184,6 +2187,14 @@ def _print_Assign(self, expr):
         rhs = self._print(expr.rhs)
         return prefix_code+'{} = {};\n'.format(lhs, rhs)
 
+    def _print_SetPop(self, expr):
+        var_type = self.get_declare_type(expr.set_variable)
+        element_type = self.get_c_type(expr.set_variable.class_type.element_type)
+        self.add_import(Import(f'Set_pop_macro/{var_type}/{element_type}', Module(f'Set_pop_macro/{var_type}/{element_type}', (), ())))
+        self.add_import(Import('STC_Extensions/Set_extensions', Module('STC_Extensions/Set_extensions', (), ())))
```
EmilyBourne, line 2194: Does this require the imports to be printed in order?

pyccel/codegen/printing/ccode.py
```diff
@@ -2184,6 +2187,14 @@ def _print_Assign(self, expr):
...
+        set_var = expr.set_variable.name
+        return f'{var_type}_pop(&{set_var})'
```
EmilyBourne, line 2196: Please use `self._print` to print objects as [documented](https://github.com/pyccel/pyccel/blob/devel/developer_docs/codegen_stage.md#_print)

This code will break for sets in classes. It will also be unnecessarily verbose for a set which is an optional argument (e.g. `set_int_pop(&(*set_var))`). You should use `ObjectAddress` to help with this

pyccel/extensions/STC_Extensions/Set_extensions.h
```diff
@@ -0,0 +1,18 @@
...
+    _c_MEMB(_iter) itr = _c_MEMB(_begin)(self);
+    if (itr.ref)
+    {
+        i_key value = *(itr.ref);
+        _c_MEMB(_erase_at)(self, itr);
+        return value;
+    }
+    return *(itr.ref);
+}
+
```
EmilyBourne, line 17: This is not an external extension. Please move it to `internals/` with the other methods that only exist inside Pyccel code

tests/epyccel/test_epyccel_sets.py
```diff
@@ -114,36 +114,6 @@ def copy_complex():
...
-
-def test_Pop_complex(language):
-    def Pop_complex():
-        se = {1j, 3j, 6j}
-        se.pop()
-        return se
-    epyccel_remove = epyccel(Pop_complex, language = language)
-    pyccel_result = epyccel_remove()
-    python_result = Pop_complex()
-    assert python_result == pyccel_result
```
EmilyBourne, line 145: Is there any reason that these methods were moved? It makes it harder to see what changes you made

tests/epyccel/test_epyccel_sets.py
```diff
@@ -364,3 +334,50 @@ def copy_from_arg2(a : 'set[float]'):
...
+
+
+@pytest.mark.parametrize( 'language', (
+        pytest.param("fortran",marks = [
+            pytest.mark.skip(reason="pop method not implemented in fortran"),
+            pytest.mark.fortran]),
+        pytest.param("c", marks = pytest.mark.c),
+        pytest.param("python", marks = pytest.mark.python)
+    )
+)
```
EmilyBourne, line 346: Why is this only needed for ints?

EmilyBourne, line 346: You may want to create a new module parameter with a similar name to avoid repetition here. E.g `language` and `python_language` or `c_language` `language_without_fortran`, etc

EmilyBourne (changes requested):

CHANGELOG.md
mustapha-belbiad: Fixed : a5efe57

pyccel/codegen/printing/ccode.py
mustapha-belbiad: If you're talking about order of defining the macros and include of `Set_extensions`, yes the file need the `i_type` and `i_key`.

pyccel/codegen/printing/ccode.py
mustapha-belbiad: Fixed : db9e6c0

pyccel/extensions/STC_Extensions/Set_extensions.h
mustapha-belbiad: Should i create a new folder named `internals`?

tests/epyccel/test_epyccel_sets.py
mustapha-belbiad: The reason is to separate pop tests from the other methods to run C and Python tests.

tests/epyccel/test_epyccel_sets.py
mustapha-belbiad: Can you please explain more about `creating a new module parameter with a similar name to avoid repetition`? If you have an example, I would be grateful.

pyccel/codegen/printing/ccode.py
EmilyBourne: `#include "STC_Extensions/Set_extensions`

This is the line that I was expecting to see here but don't

pyccel/codegen/printing/ccode.py
EmilyBourne: I mean what happens if the imports are stored in a set and they end up out of order.

pyccel/codegen/printing/ccode.py
EmilyBourne: Don't you need a scope guard for this?

pyccel/extensions/STC_Extensions/Set_extensions.h
EmilyBourne: Ah sorry the folder is called `stdlib`

tests/epyccel/test_epyccel_sets.py
EmilyBourne: The example is the current module fixture which can be found at the top of this file

pyccel/codegen/printing/ccode.py
mustapha-belbiad: I think I understand what you meant, so I tried this solution : 82aa8fd
 If you have a better idea, that would be good.

tests/epyccel/test_epyccel_sets.py
mustapha-belbiad: Fixed : 5828c07

mustapha-belbiad: Fixed : 5828c07

pyccel/codegen/printing/ccode.py
EmilyBourne: This looks safer :+1: don't you need a scope guard though?
```c
#ifdef ...
```

tests/epyccel/test_epyccel_sets.py
```diff
@@ -364,3 +335,54 @@ def copy_from_arg2(a : 'set[float]'):
     python_result = copy_from_arg2(a)
     assert isinstance(python_result, type(pyccel_result))
     assert python_result == pyccel_result
+
+@pytest.fixture( params=[
+        pytest.param("fortran", marks = [
+            pytest.mark.skip(reason="pop methods not implemented in fortran"),
```
EmilyBourne, line 341: Can we give this a more generic reason so it can be reused for future methods

```suggestion
            pytest.mark.skip(reason="set methods not implemented in fortran"),
```

pyccel/extensions/STC_Extensions/Set_extensions.h
mustapha-belbiad: Fixed : cade789

tests/epyccel/test_epyccel_sets.py
mustapha-belbiad: Fixed : 7ef1b00

pyccel/codegen/printing/ccode.py
mustapha-belbiad: Fixed : 8243ccb

pyccel/codegen/printing/ccode.py
```diff
@@ -1005,7 +1005,15 @@ def _print_Import(self, expr):
...
-
+        elif source.startswith('Set_pop'):
+            _ , i_type, i_key = source.split('/')
+            self.add_import(Import('STC_Extensions', Module('STC_Extensions', (), ())))
+            return '\n'.join((
+                   f'#ifndef TOOLS_SET_{str(i_key).upper()}\n'
+                   f'#define TOOLS_SET_{str(i_key).upper()}\n'           
+                   f'#define i_type {i_type}',
+                   f'#define i_key {i_key}\n',
+                   f'#include "STC_Extensions/Set_extensions.h"\n#endif\n'))
```
EmilyBourne, line 1016: STC extensions is not really a library so I don't think it is necessary to include it via the library folder

```suggestion
                   f'#include "Set_extensions.h"\n#endif\n'))
```

pyccel/codegen/utilities.py
```diff
@@ -111,7 +112,10 @@ def copy_internal_library(lib_folder, pyccel_dirpath, extra_files = None):
     """
     # get lib path (stdlib_path/lib_name or ext_path/lib_name)
     if lib_folder in external_libs:
-        lib_path = os.path.join(ext_path, external_libs[lib_folder], lib_folder)
+        if external_libs[lib_folder] == "STC_Extensions":
```
EmilyBourne, line 115: STC_Extensions is an internal library not an external library. Is it not possible to handle it in the same way as other external libraries?

pyproject.toml
```diff
@@ -58,7 +58,8 @@ include = [
   "pyccel/stdlib/**/*.c",
   "pyccel/stdlib/**/*.f90",
   "pyccel/extensions/STC/include",
-  "pyccel/extensions/gFTL/include/v2"
+  "pyccel/extensions/gFTL/include/v2",
+  "pyccel/extensions",
```
EmilyBourne, line 62: This change should not be done. It will lead to STC docs etc being included

tests/epyccel/test_epyccel_sets.py
```diff
@@ -13,6 +13,7 @@
     ],
     scope = "module"
 )
+
```
EmilyBourne, line 16: Useless change?

EmilyBourne (changes requested):

pyccel/codegen/printing/ccode.py
mustapha-belbiad: It's a good option if we have the `List_extensions.h` file in the future.

pyccel/codegen/utilities.py
mustapha-belbiad: The reason for this is to include the two folders, `STC` and `STC_extensions`, in the same way.

tests/epyccel/test_epyccel_sets.py
mustapha-belbiad: Fixed : c20c7ed

pyproject.toml
mustapha-belbiad: Fixed : c23fe61

pyccel/codegen/printing/ccode.py
EmilyBourne: What is a good option?
For me a library is something that can be downloaded on its own and used. So STC is a library, Pyccel is a library. If we put STC extensions in its own repo then it could be a library, but as is it is just a collection of functions provided by Pyccel to work round STC's limitations

pyccel/codegen/utilities.py
EmilyBourne: Yes that is my point. As they are conceptually different I don't see why they should be included in the same way. I think STC_extensions should be handled like every other folder in pyccel/stdlib

mustapha-belbiad: Fixed  : 96aec13

pyccel/codegen/printing/ccode.py
mustapha-belbiad: Fixed : 8000948

tests/epyccel/test_epyccel_sets.py
EmilyBourne: That is a different line

EmilyBourne: Decorators shouldn't have a space between them and the function that is being passed to them as an argument

tests/epyccel/test_epyccel_sets.py
```diff
@@ -364,3 +335,53 @@ def copy_from_arg2(a : 'set[float]'):
...
+            pytest.mark.skip(reason="set methods not implemented in fortran"),
+            pytest.mark.fortran]),
+        pytest.param("c", marks =  pytest.mark.c),
+        pytest.param("python", marks = pytest.mark.python)
+    ],
+    scope = "module"
+)
+
+def language_without_fortran(request):
+    return request.param
```
EmilyBourne, line 350:
```suggestion
@pytest.fixture( params=[
        pytest.param("fortran", marks = [
            pytest.mark.skip(reason="set methods not implemented in fortran"),
            pytest.mark.fortran]),
        pytest.param("c", marks =  pytest.mark.c),
        pytest.param("python", marks = pytest.mark.python)
    ],
    scope = "module"
)
def language_without_fortran(request):
    return request.param
```

pyccel/stdlib/STC_Extensions/Set_extensions.h
```diff
@@ -0,0 +1,18 @@
...
+static inline i_key _c_MEMB(_pop)(i_type* self) {
+    _c_MEMB(_iter) itr = _c_MEMB(_begin)(self);
+    if (itr.ref)
+    {
+        i_key value = *(itr.ref);
+        _c_MEMB(_erase_at)(self, itr);
+        return value;
+    }
+    return *(itr.ref);
+}
```
EmilyBourne, line 16: It would be great if you could add some comments here.

tests/epyccel/test_epyccel_sets.py
mustapha-belbiad: Fixed : dc1b329

tests/epyccel/test_epyccel_sets.py
mustapha-belbiad: Fixed : dc1b329

EmilyBourne (approved):
This looks like a nice clean fix :+1:

PR #1914, opened 2024-06-14
Fix renamed function lookup

### LINKED ISSUE #1913: Incorrect Handling of Function Name Overlap with Special Names in Target Language
### Describe the bug
Incorrect handling occurs when using a faction name that overlaps with a special name in the target language. When Pyccel renames the function, it should be found by the new name assigned by Pyccel

### To Reproduce
code to reproduce the behavior:
```python
def order_c(x:int, y:int):
    return x + y
x = order_c(1, 2)


```

### Error details
the error message:
```ERROR at annotation (semantic) stage
pyccel:
 |fatal [semantic]: test.py [3,4]| Undefined function (order_c)
```

### Expected behavior
the code should work just fine and gave this result 
```c
#include "test.h"
#include <stdlib.h>
#include <stdint.h>
#include <stdbool.h>

int64_t x;
bool initialised;

int64_t order_c(int64_t x, int64_t y)
{
    return x + y;
}

void test__init(void)
{
    if (!initialised)
    {
        x = order_c(INT64_C(1), INT64_C(2));
        initialised = 1;
    }
}

```

### Language
language the python code is translated to (c language)

EmilyBourne:
`FunctionDef` objects are annotated during the first `FunctionCall` where they are used. At this point they have not been renamed if there are name conflicts. Both the original and the collisionless name should therefore be used to locate the `FunctionDef`. The original name will retrieve the syntactic object while the collisionless name will retrieve the renamed function. Fixes #1913

### DIFF
--- a/pyccel/parser/base.py
+++ b/pyccel/parser/base.py
@@ -303,7 +303,7 @@ class BasicParser(object):
 
         Parameters
         ----------
-        func : FunctionDef | SympyFunction
+        func : FunctionDef | SympyFunction | Interface | FunctionAddress
             The function to be inserted into the scope.
         """
 
--- a/pyccel/parser/base.py
+++ b/pyccel/parser/base.py
@@ -301,7 +301,10 @@ class BasicParser(object):
             self.insert_symbolic_function(func)
         elif isinstance(func, (FunctionDef, Interface, FunctionAddress)):
             container = self.scope.functions
-            container[func.name] = func
+            if func.pyccel_staging == 'syntactic':
+                container[self.scope.get_expected_name(func.name)] = func
+            else:
+                container[func.name] = func
         else:
             raise TypeError('Expected a Function definition')
 
--- a/pyccel/parser/base.py
+++ b/pyccel/parser/base.py
@@ -295,7 +295,17 @@ class BasicParser(object):
         return self._blocking
 
     def insert_function(self, func):
-        """."""
+        """
+        Insert a function into the current scope.
+
+        Insert a function into the current scope under the final name by which it
+        will be known in the generated code.
+
+        Parameters
+        ----------
+        func : FunctionDef | SympyFunction
+            The function to be inserted into the scope.
+        """
 
         if isinstance(func, SympyFunction):
             self.insert_symbolic_function(func)
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -1593,7 +1593,14 @@ class SemanticParser(BasicParser):
         # TODO improve check type compatibility
         if not isinstance(var, Variable):
             name = var.name
-            errors.report(INCOMPATIBLE_TYPES_IN_ASSIGNMENT.format(type(var), class_type),
+            message = INCOMPATIBLE_TYPES_IN_ASSIGNMENT.format(type(var), class_type)
+            if var.pyccel_staging == "syntactic":
+                new_name = self.scope.get_expected_name(name)
+                if new_name != name:
+                    message += '\nThis error may be due to object renaming to avoid name clashes (language-specific or otherwise).'
+                    message += f'The conflict is with "{name}".'
+                    name = new_name
+            errors.report(message,
                     symbol=f'{name}={class_type}',
                     bounding_box=(self.current_ast_node.lineno, self.current_ast_node.col_offset),
                     severity='fatal')
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -2054,8 +2054,7 @@ class SemanticParser(BasicParser):
 
             self.scope = mod_scope
 
-        for f in self.scope.functions.copy():
-            f = self.scope.functions[f]
+        for f in self.scope.functions.copy().values():
             if not f.is_semantic and not isinstance(f, InlineFunctionDef):
                 assert isinstance(f, FunctionDef)
                 self._visit(f)
@@ -2817,10 +2816,6 @@ class SemanticParser(BasicParser):
             pass
 
         func = self.scope.find(name, 'functions')
-
-        if func is None:
-            # Look for a syntactic FunctionDef in case of a function renamed due to name collisions
-            func = self.scope.find(expr.funcdef, 'functions')
 
         if func is None:
             name = str(expr.funcdef)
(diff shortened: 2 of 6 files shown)


smazouz42 (approved):
Looks good to me

tests/epyccel/modules/function_call_before_definition_2.py
jalalium, line 1: there is an issue when the name of a function A is the new name of another function B. Calls to function B are replaced by calls to function A, and function B becomes unused.

This code
```
def f():
    return g() + do()

def g():
    return 2

def do():
    return 3

def do_0001():
    return 4

a = f()
```

returns

```

int64_t g(void)
{
    return INT64_C(2);
}
int64_t do_0001_0001(void)
{
    return INT64_C(4);
}
int64_t f(void)
{
    return g() + do_0001_0001();
}
int64_t do_0001(void)
{
    return INT64_C(3);
}
void test__init(void)
{
    if (!initialised)
    {
        a = f();
        initialised = 1;
    }
}
```

pyccel/parser/semantic.py
```diff
@@ -1214,8 +1214,9 @@ def _annotate_the_called_function_def(self, old_func, function_call_args=None):
         # Set the Scope to the FunctionDef's parent Scope and annotate the old_func
         self._scope = sc
         self._visit_FunctionDef(old_func, function_call_args=function_call_args)
+        new_name = self.scope.get_expected_name(old_func.name)
```
jalalium, line 1217: `get_expected_name` does  not seem to catch this kind of conflicts.

```
def f():
    do_0001 = 5
    return g() + do() + do_0001

def g():
    return 2

def do():
    return 4

a = f()
```

jalalium (changes requested):
I think these cases are within the scope of this PR.

tests/epyccel/modules/function_call_before_definition_2.py
EmilyBourne: You are right. This was happening because we were saving functions into the scope with `func.name` which may be either the syntactic or the semantic name depending on where you are in the code.
I have fixed this: [7ee0f15](https://github.com/pyccel/pyccel/pull/1914/commits/7ee0f158077a61c3812ed7754ce7b622a572cf3d)
and added your example as a test: [75928c2](https://github.com/pyccel/pyccel/pull/1914/commits/75928c28949c13f92e7e0ceb50eb8bc82f67de86)

pyccel/parser/semantic.py
EmilyBourne: This has improved after fixing the issue above. It is now flagged as a type redefinition. This occurs because the scope of `do_0001` is smaller than the scope of `do()` so `do_0001` doesn't get renamed in the syntactic stage.
There is a discussion to be had about whether we should be more enthusiastic about avoiding this kind of mix-up but it is so rare that it is probably not worth the additional code that would be needed to fix it. What do you think?

I have clarified the error that is raised though to help the user understand what the issue is:
image

[cbe8f67](https://github.com/pyccel/pyccel/pull/1914/commits/cbe8f67e03c497493c12c8ec97ae556f36d62af5)
[d98aab6](https://github.com/pyccel/pyccel/pull/1914/commits/d98aab65ff19827f2b7b1d637584fa6cd07a1a24)

jalalium: I think we should think of a systematic way to deal with scope and name collisions to avoid dealing with these corner cases. As for the fix, I wonder if you had in mind to prevent the following:
```
def f():
    y = 2.5
    return y

y = 5
a = f()
```
This is caught as an error, but I don't think it should.

EmilyBourne: > I think we should think of a systematic way to deal with scope and name collisions to avoid dealing with these corner cases.

We have one :sweat_smile: . It is `Scope` which is found [here](https://github.com/pyccel/pyccel/blob/devel/pyccel/parser/scope.py) and documented [here](https://github.com/pyccel/pyccel/blob/devel/developer_docs/scope.md). That's why this PR is so small. It's just that it's quite difficult to catch when people forget to use it.

This is why I asked @smazouz42 to add a test for this in his PR as it is more or less the only way to check if the Scope is being used correctly.



> I wonder if you had in mind to prevent the following

Yes exactly this must not be allowed in my opinion.

> This is caught as an error, but I don't think it should.

Then it seems that we don't agree here :sweat_smile: 

I have in mind your case, but also this case:
```python
y = 3.0

def f():
    print(y)
```
Here the `y` that is local to `f` **is** the global `y` and it is quite difficult to distinguish between the local y and the global y.

At some point I would also like to support:
```python
y = 3.0

def f():
    global y
    y = 4.0
    print(y)

print(y)
f()
print(y)
```

In this case we want to see the error displayed currently if we do:
```python
y = 3.0

def f():
    global y
    y = True
    print(y)

print(y)
f()
print(y)
```

For me the simplest way to implement `global` is to simply forbid having a local object which is in collision with a global object.

You may disagree but this is then beyond the scope of this PR as this is a choice that has been made for that case and is unrelated to the bug being fixed here.

jalalium: Got it!

jalalium (changes requested):

pyccel/parser/base.py
```diff
@@ -295,13 +295,26 @@ def blocking(self):
...
-        """."""
+        """
+        Insert a function into the current scope.
+
+        Insert a function into the current scope under the final name by which it
+        will be known in the generated code.
+
+        Parameters
+        ----------
+        func : FunctionDef | SympyFunction
```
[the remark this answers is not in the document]
EmilyBourne: Given the implementation I don't think so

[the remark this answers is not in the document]
EmilyBourne: Nevermind I read the code wrong :facepalm:

EmilyBourne, line 306:
```suggestion
        func : FunctionDef | SympyFunction | Interface | FunctionAddress
```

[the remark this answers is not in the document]
EmilyBourne: Fixed 7769a13db9c5a24d809a0939dd9bb8d1ab8bd811

tests/epyccel/modules/function_call_before_definition_2.py
EmilyBourne: @jalalium can we resolve this conversation?

jalalium (approved):

Issue #1915, opened 2024-06-24
change the implementation of sign in NumPy to make it compatible with the new version


smazouz42:
### Describe the bug
The root cause of the issue has been identified as the new release of NumPy 2.0, which is now being integrated into our Continuous Integration (CI) pipeline. The problem arises because the way the sign function works in NumPy 2.0 differs from its implementation in Pyccel, leading to compatibility issues in our project , but stell we nedd to change the aimplementation of sing for complex in the future

### To Reproduce
Provide code to reproduce the behavior:
```python
import numpy as np
print(np.sign(np.array([0.+0j, 0.j, 1.+2.j, -1.+2.j, 1.-2.j, -1.-2.j, 2.j, -2.j])))

```

### Array in new version
```py
[ 0.+0.j, 0.+0.j, 0.4472136+0.8944272j, -0.4472136+0.8944272j, 0.4472136-0.8944272j, -0.4472136-0.8944272j,0. +1.j , -0.-1.j]
```

### Array in old version
```py
[ 0.+0.j  0.+0.j  1.+0.j -1.+0.j  1.+0.j -1.+0.j  1.+0.j -1.+0.j]

```

EmilyBourne:
> Mark the complex sign test to xfail for old NumPy array versions. 

This seems to be the wrong solution to the described problem. The tests fail fro new NumPy array versions. Why would we mark the old version as failing?

EmilyBourne:
If this issue relates simply to marking tests as `xfail` for now, a second issue should be created so we remember to fix the problem in the future

PR #1920, opened 2024-06-26
Pin NumPy to < 2.0

EmilyBourne:
Pin NumPy to < 2.0 to avoid errors due to sign implementation change as seen in #1915


smazouz42 (approved):
Approved good job

jalalium (approved):
Approved

PR #1922, opened 2024-06-27
[SET] Add c support for `add`

### LINKED ISSUE #1917: [SETS] Add C support for `add()`
### Test Code
```python
def foo():
    a = {1,2,1}
    a.add(5)
```

### Proposed Solution

mustapha-belbiad:
Add C and Fortran support for the `set` method `add`. Fixes #1917


EmilyBourne:
@mustapha-belbiad this issue is finally unblocked! Do you plan on completing it or do you want me to pick it up?

jalalium (approved):
Well done!

PR #1926, opened 2024-06-28
feat: list indexing using STC methods

Farouk-Echaref:


docs/builtin-functions.md
```diff
@@ -74,6 +74,25 @@ Python contains a limited number of builtin functions defined [here](https://doc
...
+## List methods
+
+| Method | Supported |
+|----------|-----------|
+| `append` | Python-only |
+| `clear` | Python-only |
+| `copy` | Python-only |
+| `count` | No |
+| `extend` | Python-only |
+| `index` | No |
```
jalalium, line 86: shouldn't this be updated?

tests/epyccel/test_epyccel_variable_annotations.py
```diff
@@ -229,13 +229,15 @@ def test_homogeneous_list_annotation_int(stc_language):
...
         a = [1, 2, 3, 4]
+        return a[0], a[1], a[2], a[3]
     epyc_homogeneous_list_annotation = epyccel(homogeneous_list_annotation, language=stc_language)
     assert epyc_homogeneous_list_annotation() == homogeneous_list_annotation()
     assert isinstance(epyc_homogeneous_list_annotation(), type(homogeneous_list_annotation()))
 
 def test_homogeneous_list_without_annotation(stc_language):
     def homogeneous_list():
         a = [1, 2, 3, 4] # pylint: disable=unused-variable
+        return a[0], a[1], a[2], a[3]
```
jalalium, line 240: I think these are the only tests for list initialization, if it is the case, don't you think that having a test for both initialization and index retrieval goes against the separationf of concerns principle?

jalalium (changes requested):
As we discussed, handling negative indices are within the scope of this PR

tests/epyccel/test_epyccel_variable_annotations.py
EmilyBourne: No. I don't. In theory I agree that it is good to separate concerns. However until it is possible to return a list (which will come quite late in development) the only way to validate the values saved in the list is to retrieve them via indices. If the following code compiles:
```python
    def homogeneous_list():
        a = [1, 2, 3, 4] # pylint: disable=unused-variable
```
there is no way to tell if it is translated to:
```c
void homogeneous_list() {
}
```

EmilyBourne:
> As we discussed, handling negative indices are within the scope of this PR

Negative indices are now handled and tested

pyccel/codegen/printing/ccode.py
```diff
@@ -1478,7 +1471,26 @@ def _print_IndexedElement(self, expr):
...
 
+        if isinstance(base.class_type, HomogeneousListType):
+            assign = expr.get_user_nodes(Assign)
+            index = self._print(inds[0])
+            list_var = self._print(ObjectAddress(base))
+            container_type = self.get_c_type(base.class_type)
+            if assign:
+                assert len(assign) == 1
+                assign_node = assign[0]
+                if assign_node is expr or assign_node.lhs.is_user_of(expr):
```
jalalium, line 1482: This does not seem to catch when `_at_mut` should be used

jalalium (changes requested):

pyccel/codegen/printing/ccode.py
EmilyBourne: Oops. Fixed [5ee6bba](https://github.com/pyccel/pyccel/pull/1926/commits/5ee6bba522b615ca043bfc5bbca850fe3dd88952)

I will add a set of tests for this

tests/epyccel/test_epyccel_lists.py
```diff
@@ -658,3 +670,23 @@ def g():
     epyc_f = epyccel(f, language=language)
     assert f() == epyc_f()
 
+def test_mutable_indexing(stc_language):
```
jalalium, line 673: I don't know if you wanna handle the following case:
```

if __name__ == '__main__':
    x = [1,2,3]
    x[x[0]] = 5
    print(x)
```
which outputs:
```
...
 (*vec_int64_t_at_mut(&x,(*vec_int64_t_at_mut(&x,INT64_C(0))))) = INT64_C(5);
...
```
Not everyting in lhs should be at_mut

EmilyBourne: I agree that this output is not ideal, but I think the generated code should work. I suspect that it is quite difficult to recognise these cases. There is a similar issue with:
```python
x = [[1, 2, 3], [4, 5, 6]]
x[0][0] = 5
```
which should ideally give:
```c
 (*vec_int64_t_at_mut(vec_vec_int64_t_at(&x,INT64_C(0)), INT64_C(0))) = INT64_C(5);
```
but currently gives:
```c
 (*vec_int64_t_at_mut(vec_vec_int64_t_at_mut(&x,INT64_C(0)), INT64_C(0))) = INT64_C(5);
```

I think that as long as the code runs this is sufficient for a first attempt in order to unblock all other developments related to lists. I will open an issue so that we remember that this implementation is sub-optimal

jalalium: Excellent, thank you!

jalalium (approved):

tests/epyccel/test_epyccel_lists.py
EmilyBourne: See #1955

PR #1930, opened 2024-07-13
Clean STC imports

EmilyBourne:
Handling imports by separating concepts with limiters is potentially error prone. It also results in lots of special cases being written to handle each individual import. This PR unifies the strategy by using existing Pyccel concepts. An `Import` now takes the source file as the source argument (instead of the string containing limiters) and an `AsName` as the target. The object in the `AsName` is the PyccelType being handled by this import (described by a VariableTypeAnnotation) and the `target` (local name) of the `AsName` is the C name of this object.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,7 @@ All notable changes to this project will be documented in this file.
 -   \[INTERNALS\] `PyccelFunction` objects which do not represent objects in memory have the type `SymbolicType`.
 -   \[INTERNALS\] Rename `_visit` functions called from a `FunctionCall` which don't match the documented naming pattern to `_build` functions.
 -   \[INTERNALS\] Remove unnecessary argument `kind` to `Errors.set_target`.
+-   \[INTERNALS\] Handle STC imports with Pyccel objects.
 
 ### Deprecated
 
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -241,6 +241,8 @@ c_imports = {n : Import(n, Module(n, (), ())) for n in
                  'assert',
                  'numpy_c']}
 
+import_header_guard_prefix = {'Set_extensions' : '_TOOLS_SET'}
+
 class CCodePrinter(CodePrinter):
     """
     A printer for printing code in C.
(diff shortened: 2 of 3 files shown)


jalalium (approved):

pyccel/ast/core.py
```diff
@@ -164,7 +164,7 @@ def object(self):
         return self._obj
 
     def __repr__(self):
-        return f'{self.name} as {self.target}'
+        return f'{self.object} as {self.target}'
```
EmilyBourne, line 167: This change should be mentioned in the PR description

pyccel/ast/core.py
```diff
@@ -3729,14 +3729,14 @@ def __init__(self, source, target = None, ignore_at_print = False, mod = None):
...
         else:
             for i in target:
                 assert isinstance(i, (AsName, Module))
                 if isinstance(i, Module):
-                    self._target.add(AsName(i,source))
+                    self._target[AsName(i,source)] = None
                 else:
-                    self._target.add(i)
+                    self._target[i] = None
         super().__init__()
```
EmilyBourne, line 3740: Why is the call to `Import._format` only needed at one of these 3 lines?

pyccel/ast/core.py
```diff
@@ -3801,23 +3806,25 @@ def __str__(self):
...
+        ----------
+        new_target : str | AsName | iterable[str | AsName]
+            The new import target.
         """
         if iterable(new_target):
-            self._target.update(new_target)
+            self._target.update({t: None for t in new_target})
         else:
-            self._target.add(new_target)
+            self._target[new_target] = None
```
EmilyBourne, line 3827: See above. Is `_format` needed?

pyccel/ast/core.py
EmilyBourne: Done

pyccel/ast/core.py
EmilyBourne: Pyccel objects don't need to be formatted

pyccel/ast/core.py
EmilyBourne: `_format` is only needed at the syntactic stage. I have added an assert for this a22b6b34

jalalium (approved):

pyccel/ast/core.py
EmilyBourne: Assertion failed. Format needed!

PR #1931, opened 2024-07-13
Fix unreleased file lock

EmilyBourne:
The C tests occasionally fail in a way that causes a deadlock which only resolved when the code times out after an hour. This was happening due to 2 problems:
1. `copy_internal_library` was not designed to handle copying folders with subfolders leading to a `PermissionError` when calling `os.remove`
2. If an unhandled error is raised after the file lock is acquired the lock is never released.

This PR fixes this problem by handling subfolders using `os.walk` and ensuring that locks are always released using a context manager.

This change is not mentioned in the CHANGELOG as the bug was introduced after the last release.

### DIFF
--- a/pyccel/codegen/compiling/file_locks.py
+++ b/pyccel/codegen/compiling/file_locks.py
@@ -1,0 +1,48 @@
+# -*- coding: utf-8 -*-
+#------------------------------------------------------------------------------------------#
+# This file is part of Pyccel which is released under MIT License. See the LICENSE file or #
+# go to https://github.com/pyccel/pyccel/blob/devel/LICENSE for full license details.      #
+#------------------------------------------------------------------------------------------#
+"""
+Module handling classes which handle file locking to avoid deadlocks.
+"""
+from filelock import FileLock
+
+class FileLockSet:
+    """
+    Class for grouping file locks.
+
+    A class which groups file locks. By grouping these the locking can
+    be handled via a context manager which reduces the risk of the locks
+    not being correctly released.
+
+    Parameters
+    ----------
+    locks : iterable[FileLock], optional
+        The locks that should be stored in the FileLockSet.
+    """
+    def __init__(self, locks = ()):
+        assert all(isinstance(l, FileLock) for l in locks)
+        self._locks = list(locks)
+
+    def __enter__(self):
+        for l in self._locks:
+            l.acquire()
+
+    def __exit__(self, *args):
+        for l in self._locks:
+            l.acquire()
+
+    def append(self, new_lock):
+        """
+        Add a new lock to the FileLockSet.
+
+        Add a new lock to the FileLockSet.
+
+        Parameters
+        ----------
+        new_lock : FileLock
+            The new lock.
+        """
+        assert isinstance(new_lock, FileLock)
+        self._locks.append(new_lock)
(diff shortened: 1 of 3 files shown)


pyccel/codegen/utilities.py
```diff
@@ -145,36 +149,32 @@ def copy_internal_library(lib_folder, pyccel_dirpath, extra_files = None):
...
+                # Remove all files in destination directory
+                for d in dst_files:
+                    d_file = os.path.join(lib_dest_path, d)
+                    try:
+                        os.remove(d_file)
+                    except FileNotFoundError:
+                        # Don't call error in case of temporary compilation file that has disappeared
+                        # since reading the folder
+                        pass
+                # Copy all files from the source to the destination
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
Corner case which is handled by doing nothing doesn't need to be tested

pyccel/codegen/compiling/file_locks.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    def __init__(self, locks = ()):
+        assert all(isinstance(l, FileLock) for l in locks)
+        self._locks = list(locks)
+
+    def __enter__(self):
+        for l in self._locks:
+            l.acquire()
+
+    def __exit__(self, *args):
```
jalalium, line 32: Shouldn't the `__exit__` method release the locks instead of acquiring them? A context manager is also supposed to handle exceptions to ensure that the locks will be released.

pyccel/codegen/compiling/file_locks.py
```diff
@@ -0,0 +1,48 @@
...
+        for l in self._locks:
```
jalalium, line 33: It would be better to iterate like this instead:
`for l in reversed(self._locks):`

jalalium (changes requested):

pyccel/codegen/compiling/file_locks.py
EmilyBourne: I agree

pyccel/codegen/compiling/file_locks.py
EmilyBourne: :facepalm: I'm an idiot. I'm not sure how this didn't deadlock

EmilyBourne: > A context manager is also supposed to handle exceptions to ensure that the locks will be released.

What do you mean by this? It is my understanding that this code will always be run even if an exception is raised (this is why context managers are preferred over the current strategy).
https://docs.python.org/3/reference/datamodel.html#object.__exit__

The context manager **may** decide to intercept the exception if it wants to handle this. In my case I don't want to intercept the exception, I am happy for it to be raised I just need to release the locks first

jalalium: I was thinking  about changing it to `def __exit__(self, exc_type, exc_val, exc_tb):` to reflect that it intercepts errors. However, you are right there is no need for it.

jalalium (approved):

pyccel/codegen/compiling/file_locks.py
EmilyBourne: Ah ok. I think it is better to be explicit with arguments. I didn't check the official docs when I wrote this which is how I ended up with `*args`. Fixed 7cde31fb

PR #1934, opened 2024-07-17
Fix code exportation

### LINKED ISSUE #1933: Exporting code in Codegen is done a character at a time
### Relevant Discussion
the `export` method in the `Codegen` writes to the file a character at a time when the code is a string instead of a list of strings

### Describe the feature
When the code is a single string ` for line in code:` actually loops over the characters of the string instead of the lines. This makes it both ambiguous and slow.

### Test Code
```python
      code = self._printer.doprint(self.ast)
        with open(filename, 'w', encoding="utf-8") as f:
            for line in code:
                f.write(line)

        # print module header
        if header_ext is not None:
            code = self._printer.doprint(ModuleHeader(self.ast))
            with open(header_filename, 'w', encoding="utf-8") as f:
                for line in code:
                    f.write(line)

        # print program
        prog_filename = None
        if self.is_program and self.language != 'python':
            folder = os.path.dirname(filename)
            fname  = os.path.basename(filename)
            prog_filename = os.path.join(folder,"prog_"+fname)
            code = self._printer.doprint(self.ast.program)
            with open(prog_filename, 'w') as f:
                for line in code:
                    f.write(line)
```

### Proposed Solution

A solution would be to check for when the code is a single line, if it is not always the case at the time being.

jalalium:
This PR fixes #1933 . Code exportation to the file was previously done one character at a time. Now it is done all at once

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,7 +43,7 @@ All notable changes to this project will be documented in this file.
 -   #1903 : Fix memory leak when using type annotations on local variables.
 -   #1913 : Fix function calls to renamed functions.
 -   #1927 : Improve error Message for missing target language compiler in Pyccel
--   #1933 : Fix code exportation
+-   #1933 : Improve code printing speed.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ All notable changes to this project will be documented in this file.
 -   #1903 : Fix memory leak when using type annotations on local variables.
 -   #1913 : Fix function calls to renamed functions.
 -   #1927 : Improve error Message for missing target language compiler in Pyccel
+-   #1933 : Fix code exportation
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


smazouz42 (approved):
Good job Jalal  in this PR. Well done ,  good improvement in the complexity of the code generation

CHANGELOG.md
```diff
@@ -43,6 +43,7 @@ All notable changes to this project will be documented in this file.
 -   #1903 : Fix memory leak when using type annotations on local variables.
 -   #1913 : Fix function calls to renamed functions.
 -   #1927 : Improve error Message for missing target language compiler in Pyccel
+-   #1933 : Fix code exportation
```
EmilyBourne, line 46: This may have a positive effect for users so it would be nice to give it a more explicit description.
```suggestion
-   #1933 : Improve code printing speed.
```

jalalium: Done!

EmilyBourne (approved):

PR #1938, opened 2024-07-22
Optimize substitute method

jalalium:
### Description
This PR addresses an optimization for the substitute method in the PyccelAstNode class. The current implementation does not check if original and replacement are empty, leading to unnecessary recursion over the code. This update adds a condition to return immediately if  original is empty (and consequenty replacement), thereby improving the performance of the method.

### DIFF
--- a/.dict_custom.txt
+++ b/.dict_custom.txt
@@ -119,3 +119,4 @@ indexable
 traceback
 STC
 gFTL
+Optimize
--- a/.dict_custom.txt
+++ b/.dict_custom.txt
@@ -119,4 +119,3 @@ indexable
 traceback
 STC
 gFTL
-Optimize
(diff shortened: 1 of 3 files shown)


.dict_custom.txt
```diff
@@ -119,3 +119,4 @@ indexable
 traceback
 STC
 gFTL
+Optimize
```
EmilyBourne, line 122: optimize should be in the dictionary. Probably it is missing because we are using UK english

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ All notable changes to this project will be documented in this file.
 -   #1895 : Add Python support for dict initialisation with `dict()`.
 -   #1886 : Add Python support for dict method `pop()`.
 -   #1936 : Add missing C output for inline decorator example in documentation
+-   #1937 : Optimize substitute method
```
EmilyBourne, line 30:
```suggestion
-   #1937 : Optimise `pyccel.ast.basic.PyccelAstNode.substitute` method.
```

.dict_custom.txt
jalalium: Shall I remove it and use `optimise` instead?

EmilyBourne: Yes please

jalalium: done

EmilyBourne (approved):

PR #1943, opened 2024-07-22
 Support dict variable declarations in C

EmilyBourne:


smazouz42 (approved):
looks good to me , well done Emily.

pyccel/codegen/printing/ccode.py
```diff
@@ -1245,6 +1257,13 @@ def get_c_type(self, dtype):
             i_type = f'{container_type}_{element_type}'
             self.add_import(Import(f'stc/{container_type}', AsName(VariableTypeAnnotation(dtype), i_type)))
             return i_type
+        elif isinstance(dtype, DictType):
+            container_type = 'smap'
```
jalalium, line 1261: Shouldn't we be using `hmap` instead, even though the keys of a `dict` are sorted by insertion time, it is still a hash map. This might affect performance since lookup and insertions do not have the same complexity in both data structures. I think we should use `smap` for the `OrderedDict` from `collections` if we are planning on implementing it.

jalalium (changes requested):

pyccel/codegen/printing/ccode.py
EmilyBourne: Hmm good question. Since Python 3.7 made `dict` sorted by default I find myself relying on this regularly which is why I automatically reached for the sorted version, but you are correct, in any case they are not sorted using the same metric so maybe `hmap` would be better. It would have to be well documented but at least it should be quite easy to move back if we change our mind later.

I am not sure I understand why `OrderedDict` still exists. Given the (few) differences between `OrderedDict` and `dict` (mainly `move_to_end`) I don't see why `smap` would be suited for this case but not for `dict`?

EmilyBourne: Relevant link:
https://realpython.com/python-ordereddict/

EmilyBourne: Updated : [0bb4e62](https://github.com/pyccel/pyccel/pull/1943/commits/0bb4e62f2d79b8c03eec3a499b38e260c3c22f02)
and updated the docs: [6ef183c](https://github.com/pyccel/pyccel/pull/1943/commits/6ef183cdd506fbbbf0eccd30f4085f535e416f0c)

jalalium (approved):

PR #1974, opened 2024-08-29
Add C support for list.pop method

EmilyBourne:
Add C support for list.pop method in an assign or without an index element. In order to handle negative indices `_print_PyccelArrayShapeElement` is updated to handle STC objects and `Variable.shape_can_change` is corrected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,14 @@ All notable changes to this project will be documented in this file.
 ### Added
 
 -   #1720 : Add support for `Ellipsis` as the only index for an array.
+-   #1689 : Add Python support for list method `append()`.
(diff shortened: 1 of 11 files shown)


jalalium (commented):
for lists of lists, issues relative to `double free` can arise when the popped item is freed along with the original list.

EmilyBourne:
> for lists of lists, issues relative to `double free` can arise when the popped item is freed along with the original list.

This point is related to #1959. But I'm not sure I understand your point. If we pop then the popped item is no longer in the original list so it will not be freed by it.

However probably a free is missing if the result is not saved (and maybe also if it is as this object is not allocated).

Do you think we should fix this here or should we leave it until we have support for lists of lists?

jalalium (approved):

Issue #1657, opened 2024-01-10
[LISTS] Support variable declarations in Fortran


EmilyBourne:
### Describe the feature
Minimal printing support should be implemented for lists. It should be possible to declare a list using a type annotation.

### Test Code
Provide code which does not currently work but which should do when this issue is fixed:
```python
def declare_list():
    a : list[int]
```

### Proposed Solution
This feature should be implemented with the https://github.com/Goddard-Fortran-Ecosystem/gFTL library. The Vector type should be used.

### Target Language
Fortran

### Pre-requisites
- #1529 
- #1656

EmilyBourne:
The syntax `list[int]` is not supported in Python 3.8 (https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html). As a first step Python 3.8 tests should be marked as xfail. An issue should be opened to request support for `List` from the typing module. This is related to #1544

PR #1975, opened 2024-08-30
Unify import strategy

EmilyBourne:
Unify the strategy for handling additional imports in the printing stage for different languages. This allows `get_additional_imports` and `add_import` to be moved to the superclass `codegen.printing.CodePrinter` and will make it easier to handle different import strategies (e.g. #1657)

**Commit Summary**
- Move `get_additional_imports` to `codegen.printing.CodePrinter`
- Move `add_import` to `codegen.printing.CodePrinter`
- Use a dictionary for imports in `codegen.printing.FCodePrinter` to match what is done in `codegen.printing.CCodePrinter`
- Change the storage type of the `_additional_imports` dict in `codegen.printing.PythonCodePrinter` to match what is done in `codegen.printing.CCodePrinter`

### DIFF
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -314,28 +314,6 @@ class CCodePrinter(CodePrinter):
         self._current_module = None
         self._in_header = False
 
-    def get_additional_imports(self):
-        """return the additional imports collected in printing stage"""
-        return self._additional_imports.keys()
-
-    def add_import(self, import_obj):
-        """
-        Add a new import to the current context.
-
-        Add a new import to the current context. This allows the import to be recognised
-        at the compiling/linking stage. If the source of the import is not new then any
-        new targets are added to the Import object.
-
-        Parameters
-        ----------
-        import_obj : Import
-            The AST node describing the import.
-        """
-        if import_obj.source not in self._additional_imports:
-            self._additional_imports[import_obj.source] = import_obj
-        elif import_obj.target:
-            self._additional_imports[import_obj.source].define_target(import_obj.target)
-
     def _format_code(self, lines):
         return self.indent_code(lines)
 
--- a/pyccel/codegen/printing/pycode.py
+++ b/pyccel/codegen/printing/pycode.py
@@ -77,7 +77,6 @@ class PythonCodePrinter(CodePrinter):
     def __init__(self, filename):
         errors.set_target(filename)
         super().__init__()
-        self._additional_imports = {}
         self._aliases = {}
         self._ignore_funcs = []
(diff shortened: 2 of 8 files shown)


pyccel/codegen/printing/ccode.py
```diff
@@ -314,28 +314,6 @@ def __init__(self, filename, prefix_module = None):
         self._current_module = None
         self._in_header = False
 
-    def get_additional_imports(self):
```
jalalium, line 317: Did you leave `self._additional_imports` on purpose for the `CCodePrinter` class?

pyccel/codegen/printing/pycode.py
```diff
@@ -394,7 +373,7 @@ def _print_Program(self, expr):
         module = modules[0]
         imports = ''.join(self._print(i) for i in expr.imports if i.source_module is not module)
         body     = self._print(expr.body)
-        imports += ''.join(self._print(i) for i in self.get_additional_imports())
+        imports += ''.join(self._print(i) for i in self._additional_imports.values())
```
jalalium, line 376: does this private attribute need a getter?

pyccel/codegen/printing/pycode.py
```diff
@@ -983,7 +954,7 @@ def _print_Module(self, expr):
         if free_func:
             self._ignore_funcs.append(free_func)
 
-        imports += ''.join(self._print(i) for i in self.get_additional_imports())
+        imports += ''.join(self._print(i) for i in self._additional_imports.values())
```
jalalium, line 957: does this private attribute need a getter?

jalalium (changes requested):

pyccel/codegen/printing/ccode.py
EmilyBourne: Yes. It is initialised with a value so it is still needed. I can update the existing dict instead if you prefer?

pyccel/codegen/printing/pycode.py
EmilyBourne: Only if it is used outside the class

EmilyBourne: (pylint complains if you use private members. But you can always use private members of your own class otherwise there's no point saving anything privately)

jalalium: got it!

pyccel/codegen/printing/ccode.py
jalalium: this is fine too! just wanted to make sure!

jalalium (approved):
Very good job!

Issue #1297, opened 2023-01-08
Can't iterate over an enum, map or zip in a functional for


EmilyBourne:
### Describe the bug
Can't iterate over an enum, map or zip in a functional for

### To Reproduce
Provide code to reproduce the behavior:
```python
@types('int[:]')
def functional_for_1d_enum(y):
    a = [i*yi for i,yi in enumerate(y)]
    return len(a), a[0], a[1], a[2], a[3]
```

### Error details
Provide the generated code, or the error message:
```bash
RuntimeError: (i, yi) does not exist in scope
```

EmilyBourne:
Duplicate of #1399

PR #1977, opened 2024-09-02
Improve list comprehension

jalalium:
This PR aims to fix issue #1297
This PR refactors the FunctionalFor support in Pyccel by replacing direct array indexing with the use of Python lists and the list_append function in C.
It also aims at supporting list comprehension with conditions.


CHANGELOG.md
```diff
@@ -61,6 +61,7 @@ All notable changes to this project will be documented in this file.
 -   #1892 : Fix implementation of list function when an iterable is passed as parameter.
 -   #1924 : Fix internal error arising in Duplicate or list comprehensions.
 -   #1970 : Fix missing `TypeError` for wrong type passed as optional argument.
+-   #1297 : Fix iteration over an enum, map or zip in a functional for
```
EmilyBourne:
```suggestion
-   #1297 : Fix iteration over an enum, map or zip in a list comprehension.
```
Functional for is Pyccel terminology

pyccel/ast/builtins.py
```diff
@@ -1156,7 +1156,7 @@ def __init__(self, *args):
             if lengths:
                 self._length = min(lengths)
             else:
-                self._length = self.args[0].shape[0]
+                self._length = PythonMin([a.shape[0] for a in self.args])
```
EmilyBourne, line 1159: Does `PythonMin` have a `simplify` argument? It would be nice to get a literal if all the shapes are known

pyccel/codegen/printing/ccode.py
```diff
@@ -1387,6 +1387,9 @@ def _print_Declare(self, expr):
             assert init == ''
             preface = ''
             init    = ' = {.shape = NULL}'
+        elif declaration_type.startswith('vec'):
```
EmilyBourne, line 1390: Safer to check the type exactly

```suggestion
        elif isinstance(var.class_type, HomogeneousListType):
```

pyccel/codegen/printing/ccode.py
```diff
@@ -1649,7 +1652,13 @@ def _print_Allocate(self, expr):
         free_code = ''
         variable = expr.variable
         if isinstance(variable.class_type, (HomogeneousListType, HomogeneousSetType, DictType)):
-            return ''
+            if len(variable.alloc_shape) > 0 and isinstance(variable.class_type, HomogeneousListType):
```
EmilyBourne, line 1655: Is list the only type with a reserve function? I wouldn't be surprised if one was also available for set and dict

pyccel/codegen/printing/ccode.py
```diff
@@ -1649,7 +1652,13 @@ def _print_Allocate(self, expr):
...
+                size = self._print(variable.alloc_shape[0])
+                variable_address = self._print(ObjectAddress(expr.variable))
+                container_type = self.get_c_type(expr.variable.class_type)
+                return f'{container_type}_reserve({variable_address}, {size});\n'
+            else:
+                return ''
         if variable.rank > 0:
             #free the array if its already allocated and checking if its not null if the status is unknown
```
EmilyBourne, line 1663: Should there be a call to clear for lists if the status is unknown?

pyccel/codegen/printing/ccode.py
```diff
@@ -2353,6 +2362,13 @@ def _print_PythonImag(self, expr):
     def _print_PythonConjugate(self, expr):
         return 'conj({})'.format(self._print(expr.internal_var))
 
+    def _print_ListAppend(self, expr):
+        target = expr.list_obj
+        class_type = target.class_type
+        dtype = self.get_c_type(class_type)
```
EmilyBourne, line 2368: Dtype is usually for the underlying datatype. Eg `int` in `list[int]` better to avoid this name for containers

pyccel/codegen/printing/ccode.py
```diff
@@ -2353,6 +2362,13 @@ def _print_PythonImag(self, expr):
     def _print_PythonConjugate(self, expr):
         return 'conj({})'.format(self._print(expr.internal_var))
 
+    def _print_ListAppend(self, expr):
```
EmilyBourne, line 2365: Append should be mentioned in the changelog. This probably also fixes extend

tests/codegen/fcode/test_fcode_codegen.py
```diff
@@ -19,7 +19,7 @@
          #if f not in failing_files \
          #else pytest.param(os.path.join(path_dir,f), marks = pytest.mark.xfail(reason=failing_files[f])) \
          for f in files \
-         if f.endswith(".py") \
+         if f.endswith(".py") and f != 'ListComprehension.py' \
```
EmilyBourne, line 22: Better to uncomment code above regarding xfails?

pyccel/codegen/printing/ccode.py
EmilyBourne: Don't forget to activate the associated tests. Maybe this tiny function should be in its own PR ?

pyccel/parser/semantic.py
```diff
@@ -3455,53 +3455,84 @@ def _visit_For(self, expr):
...
             start = LiteralInteger(0)
-            step  = LiteralInteger(1)
-            var   = indices[i]
-            i += 1
-            a     = self._visit(body.iterable)
-            if isinstance(a, PythonRange):
+            step = LiteralInteger(1)
+            vars = []
+            a = Iterable(self._visit(body.iterable))
+            if isinstance(a.iterable, PythonRange):
```
EmilyBourne, line 3489: I don't remember the purpose of `var` here. For the rest is it not possible to use `a.to_range` to reduce the number of ifs?

pyccel/parser/semantic.py
```diff
@@ -3455,53 +3455,84 @@ def _visit_For(self, expr):
...
+                    indices.append(var)
+                    dvar = self._infer_type(arg)
+                    class_type = dvar.pop('class_type')
+                    if class_type.rank > 0:
+                        class_type = class_type.switch_rank(class_type.rank-1)
+                        dvar['shape'] = (dvar['shape'])[1:]
+                    if class_type.rank == 0:
+                        dvar['shape'] = None
+                        dvar['memory_handling'] = 'stack'
+                    var  = Variable(class_type, var, **dvar)
```
EmilyBourne, line 3528: Not ideal to create variables without using the vriable creation functions. This usually leads to duplicated type checks or forgotten tests (e.g setting target as necessary)

tests/pyccel/scripts/hope_benchmarks/test.json
```diff
@@ -0,0 +1,66 @@
+{
```
EmilyBourne, line 1: Accidental commit?

tests/pyccel/scripts/test2.json
```diff
@@ -0,0 +1,66 @@
+{
```
EmilyBourne, line 1: Accidental commit?

EmilyBourne (changes requested):
I'm a little early but I saw you were trying to mark as ready so I thought I'd review while I'm free

CHANGELOG.md
jalalium: Done [03b6f4a](https://github.com/pyccel/pyccel/pull/1977/commits/03b6f4ae39a811bf68b7f5a66e9f9a6b4524171a)

pyccel/ast/builtins.py
jalalium: this case deals with when the shape is not known, I think if all the shapes are known, then it would fall under the first case

pyccel/codegen/printing/ccode.py
jalalium: Definitely not, I didn't give much thought to this part as i didn't consider all the cases that would require a reserve. I will think more about it

pyccel/codegen/printing/ccode.py
jalalium: could you elaborate please?

pyccel/codegen/printing/ccode.py
EmilyBourne: I don't know what cases might need reserve either. I just don't see any particular reason to exclude set/dict if they have an equivalent method with a similar name. If they don't then this question becomes irrelevant

pyccel/codegen/printing/ccode.py
EmilyBourne: The comment at the bottom of the code section says that Deallocate is called for arrays if the status of the Allocate node is 'unknown'. The equivalent function for lists is probably clear. I was wondering if the Allocate status should be used here

pyccel/codegen/printing/ccode.py
jalalium: I excluded them because tests relative to variable annotations were failing.
```
if __name__ == '__main__':
    a : 'set[float]'
    a = {1.5, 2.5, 3.3, 4.3}
```
```
int main()
{
    hset_double a;
    hset_double_reserve(&a, INT64_C(4));
    a = c_init(hset_double, {1.5, 2.5, 3.3, 4.3});
    hset_double_drop(&a);
    return 0;
}
```
this would cause a double free because `c_init` doesn't work well with `reserve` in contrast to `append`. an `Allocate` might need to be treated differently depending on what will happen to the memory.

EmilyBourne: Is this not also true for:
```python
a = [1,2,3,4]
```
?
How do you avoid the reserve/init problem in this case?

jalalium: I think we should update `_assign_lhs_variable` to not add an `Allocate` node when the `rhs` is a `c_init` since it takes care of memory allocation.

pyccel/codegen/printing/ccode.py
jalalium: done #1982, however the tests require C wrapping for lists as returns of functions as well as the ability to print lists

pyccel/codegen/printing/ccode.py
EmilyBourne: I'm not sure how you manage this. The semantic stage is language agnostic.  What happens if the Allocate node is needed for Fortran? You put it back in `FCodePrinter._print_PythonList`?

It would be nice if there was a way to make the init value available. Maybe we could add a property to `Allocate`?

Additionally if you have two c_init statements in a row then you will definitely still need the Allocate to know that you need to call c_drop first. (I assume if you can't call c_init on an initialised list then you can't call it twice without some kind of deallocation either).

Eg:
```python
a = [1, 2, 3, 4]
...
a = [6, 7, 8, 9,  10]
```
If `a` was an array instead of a list then this would give:
```fortran
allocate(a(0:3))
a(:) = (/ 1, 2, 3, 4 /)
deallocate(a)
allocate(a(0:4))
a(:) = (/ 6, 7, 8, 9 /)
```
The deallocation comes from the Allocate node

pyccel/codegen/printing/ccode.py
EmilyBourne: Ah. So it's either blocked by the associated issue or the returns need changing to a len call and elements and then it's blocked by Farouk's issue

pyccel/ast/sympy_helper.py
```diff
@@ -104,6 +104,20 @@ def sympy_to_pyccel(expr, symbol_map):
...
+            result = PythonMin(result, a)
+        return result
+
+    elif isinstance(expr, sp.Max):
+        args = [sympy_to_pyccel(a, symbol_map) for a in expr.args]
+        result = args[0]
+        for a in args[1:]:
+            result = PythonMax(result, a)
+        return result
+
```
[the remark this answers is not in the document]
jalalium: no way to test, only for completeness

pyccel/ast/sympy_helper.py
```diff
@@ -179,6 +193,14 @@ def pyccel_to_sympy(expr, symbol_map, used_names):
...
         return sp.ceiling(pyccel_to_sympy(expr.args[0], symbol_map, used_names))
 
+    elif isinstance(expr, PythonMin):
+        args = [pyccel_to_sympy(ee, symbol_map, used_names) for e in expr.args for ee in e]
+        return sp.Min(*args)
+
+    elif isinstance(expr, PythonMax):
+        args = [pyccel_to_sympy(ee, symbol_map, used_names) for e in expr.args for ee in e]
+        return sp.Max(*args)
+
```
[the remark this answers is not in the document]
jalalium: no way to test, only for completeness

pyccel/ast/sympy_helper.py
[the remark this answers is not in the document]
EmilyBourne: @jalalium I agree with your initial decision to include this for completeness. Can you put it back and follow the bot's instructions to get it to ignore the lack of coverage.
The comment needs to address the bot. The comment that you left above is for me.

pyccel/parser/syntactic.py
```diff
@@ -1182,25 +1198,48 @@ def _visit_ListComp(self, stmt):
...
+        else:
+            operations.append(DottedName(lhs, FunctionCall('append', [FunctionCallArgument(result)])))
+
+        conditions = []
+        for generator in generators:
+            if generator.body.body:
+                assert isinstance(generator.body.body[0], If)
+                for ifSection in generator.body.body[0].blocks:
+                    conditions.append(ifSection.condition)
+                generator._body = CodeBlock([]) # pylint: disable=protected-access
```
EmilyBourne, line 1228: I agree with codacy. This is a code smell. Is there no other way to do this?

This will also cause bugs in the AST tree.

EmilyBourne: If the body is not used after this does it matter if it contains ifs?

pyccel/parser/syntactic.py
```diff
@@ -1182,25 +1198,48 @@ def _visit_ListComp(self, stmt):
...
+
+        if conditions:
+            if output_type != 'list':
+                errors.report("Cannot handle if statements in list comprehensions if lhs is a numpy array.\
+                              List length cannot be calculated.\n" + PYCCEL_RESTRICTION_TODO,
+                               symbol=stmt, severity='error')
```
EmilyBourne, line 1234: Errors related to the type should not appear in the syntactic stage which contains no information about typing

tests/epyccel/test_arrays.py
```diff
@@ -4049,6 +4049,7 @@ def test_array_float_nested_C_array_initialization_3(language):
         pytest.param("python", marks = pytest.mark.python)
     )
 )
+@pytest.mark.skipif(lambda lang: lang == 'fortran', reason="List comprehension is not supported in Fortran")
```
EmilyBourne, line 4052: The missing C support seemed very minimal. Is there any reason not to make the same minimal changes to fix Fortran?

tests/epyccel/test_functionals.py
```diff
@@ -16,6 +17,7 @@ def compare_epyccel(f, language, *args):
 def test_functional_for_1d_range(language):
     compare_epyccel(functionals.functional_for_1d_range, language)
 
+@pytest.mark.skipif(lambda lang: lang == 'c', reason="Skipping for C due to array redefinition issue. See #1979")
```
EmilyBourne, line 20: #1979 has been closed

tests/epyccel/test_loops.py
```diff
@@ -150,6 +151,7 @@ def test_enumerate_on_1d_array_with_start(language):
     assert np.array_equal( f1(z, 5), f2(z, 5) )
     assert np.array_equal( f1(z,-2), f2(z,-2) )
 
+
```
EmilyBourne, line 154: Unnecessary changes in this file.

pyccel/parser/semantic.py
```diff
@@ -3659,88 +3659,123 @@ def _visit_For(self, expr):
...
+            class_type = dvar.pop('class_type')
+            if class_type.rank > 0:
+                class_type = class_type.switch_rank(class_type.rank - 1)
+                dvar['shape'] = dvar['shape'][1:]
+            if class_type.rank == 0:
+                dvar['shape'] = None
+                dvar['memory_handling'] = 'stack'
+            var = self._create_variable(var_name, class_type, None, dvar)
+            dvar['class_type'] = class_type
+            variables.append((var, dvar))
```
EmilyBourne, line 3713: Why is this specific to `FunctionalFor`? Is there nothing from the `For` handling that would let you do something similar?

pyccel/parser/semantic.py
```diff
@@ -3659,88 +3659,123 @@ def _visit_For(self, expr):
...
-            existing_var = self.scope.find(var.name, 'variables')
-            if existing_var:
-                if self._infer_type(existing_var) != dvar:
-                    errors.report(f"Variable {var} already exists with different type",
+
+            for var, dvar in variables:
+                existing_var = self.scope.find(var.name, 'variables')
+                if var.name == expr.lhs:
+                    return errors.report(f"Variable {var} has the same name as the left hand side",
                             symbol = expr, severity='error')
```
EmilyBourne, line 3755: Do you have an example of where this can arise?

pyccel/parser/semantic.py
```diff
@@ -3857,31 +3886,37 @@ def _visit_FunctionalFor(self, expr):
...
         # ...
-        lhs_symbol = expr.lhs.base
+        lhs_symbol = expr.lhs
         ne = []
         lhs = self._assign_lhs_variable(lhs_symbol, d_var, rhs=expr, new_expressions=ne)
         lhs_alloc = ne[0]
 
         if isinstance(target, PythonTuple) and not target.is_homogeneous:
-            errors.report(LIST_OF_TUPLES, symbol=expr, severity='error')
+            errors.report(LIST_OF_TUPLES, symbol=expr, severity='fatal')
```
EmilyBourne, line 3895: Why does the `error` now cause crashes?

pyccel/parser/semantic.py
```diff
@@ -3857,31 +3886,37 @@ def _visit_FunctionalFor(self, expr):
...
             assert isinstance(l, For)
             # Sub in indices as defined here for coherent naming
             if idx.is_temp:
                 self.scope.remove_variable(l.target[0])
                 l.substitute(l.target[0], idx_subs[idx])
-            l = l.body.body[-1]
+            cnt += 1
+            if cnt == len(l.target):
+                l = l.body.body[-1]
+                cnt = 0
```
EmilyBourne, line 3914: I am surprised that this is not:
```python
            if idx.is_temp:
                self.scope.remove_variable(l.target[cnt])
                l.substitute(l.target[v], idx_subs[idx])
            cnt += 1
            if cnt == len(l.target):
                l = l.body.body[-1]
                cnt = 0
```
Why do you not care about the other targets?

pyccel/parser/semantic.py
jalalium: Yes.
```python
a = [a for a in range(10)]
```
Could you please point out where to add tests that should raise errors so I can add this kind of examples?

pyccel/codegen/printing/pycode.py
```diff
@@ -622,15 +626,23 @@ def _print_For(self, expr):
...
+        else:
+            condition = 'if ' + self._print(body.blocks[0].condition)
+            body = self._print(body.blocks[0].body.body[0].args[0])
         for_loops = ' '.join(['for {} in {}'.format(self._print(idx), self._print(iters))
                         for idx, iters in zip(expr.indices, iterators)])
 
-        name = self._aliases.get(type(expr),'array')
-        if name == 'array':
+        name = expr.target_type
+        if 'array' in str(name):
```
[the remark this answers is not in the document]
jalalium: You mean something like this:
```python
from numpy import array as chunk

if __name__ == '__main__':
    a = chunk([i for i in range(10)])
```

Yes, it works just fine:
```python
from numpy import array as chunk

if __name__ == "__main__":
    a = chunk([i for i in range(0, 10, 1) ])
```

pyccel/parser/semantic.py
```diff
@@ -3659,88 +3659,123 @@ def _visit_For(self, expr):
...
+        Returns
+        -------
+        pyccel.ast.basic.CodeBlock
+            A code block containing the equivalent loops and necessary variable allocations for the given `FunctionalFor` expression.
+        """
+        if expr.target_type != 'list':
+            old_index   = expr.index
+            new_index   = self.scope.get_new_name()
+            expr.substitute(old_index, new_index)
+            index = new_index
```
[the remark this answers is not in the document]
jalalium: Because list comprehensions are populated one of two ways, either through appending for lists, in which case this is irrelevant, or through indexedElement assignment for numpy arrays. This is necessary for the latter.

pyccel/parser/semantic.py
jalalium: Good catch! However, I don't know when this condition applies. I think this is vestigial code back from when `functionalFor` created loop scopes. Could you please come up with an example when idx is temp?

EmilyBourne: `idx` is temp for almost everything except `range` and `enumerate`. The difficulty is finding an example where there are multiple targets. Try this:
```python
import numpy as np

if __name__ == '__main__':
    a = np.array([1,2,3,4])
    b = np.array([5,6,7,8])
    c = [ai*bi for ai, bi in product(a, b)]
```

pyccel/parser/semantic.py
EmilyBourne: https://github.com/pyccel/pyccel/tree/devel/tests/errors

pyccel/codegen/printing/pycode.py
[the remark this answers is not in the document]
EmilyBourne: Nice! How does that work? Is `target_type` set to `array` instead of `chunk` in the semantic stage?

pyccel/codegen/printing/ccode.py
```diff
@@ -1760,7 +1760,9 @@ def _print_PyccelArrayShapeElement(self, expr):
     def _print_Allocate(self, expr):
         free_code = ''
         variable = expr.variable
-        if isinstance(variable.class_type, (HomogeneousListType, HomogeneousSetType, DictType)):
+        if isinstance(variable.class_type, (HomogeneousListType, HomogeneousSetType, DictType, StringType)):
+            if expr.status in ('allocated', 'unknown'):
+                free_code = f'{self._print(Deallocate(variable))}'
             if expr.shape[0] is None:
                 return free_code
             if expr.status in ('allocated', 'unknown'):
```
EmilyBourne, line 1768: Careful it looks like you have duplicated `if expr.status in ('allocated', 'unknown'):` in this merge

tests/epyccel/test_epyccel_generators.py
```diff
@@ -244,4 +244,12 @@ def f():
...
     f_epyc = epyccel(f, language = language)
+
+def test_sum_with_two_variables(language):
+    def f():
+        x = sum(i-j for i in range(10) for j in range(7))
+        return x
+
+    f_epyc = epyccel(f, language=language)
+
     assert f() == f_epyc()
```
EmilyBourne, line 255: You seem to have lost your assert
```suggestion
    f_epyc = epyccel(f, language = language)

    assert f() == f_epyc()

def test_sum_with_two_variables(language):
    def f():
        x = sum(i-j for i in range(10) for j in range(7))
        return x

    f_epyc = epyccel(f, language=language)

    assert f() == f_epyc()
```

pyccel/ast/sympy_helper.py
```diff
@@ -103,6 +104,13 @@ def sympy_to_pyccel(expr, symbol_map):
         else:
             return MathCeil(arg)
 
+    elif isinstance(expr, sp.Min):
+        args = [sympy_to_pyccel(a, symbol_map) for a in expr.args]
+        result = args[0]
+        for a in args[1:]:
+            result = PythonMin(result, a)
+        return result
```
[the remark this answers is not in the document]
jalalium: Good point!

pyccel/parser/semantic.py
jalalium: sorry, unintentional change.

tests/codegen/fcode/test_fcode_codegen.py
```diff
@@ -24,8 +24,7 @@
          if f not in failing_files \
          else pytest.param(os.path.join(path_dir,f), marks = pytest.mark.xfail(reason=failing_files[f])) \
          for f in files \
-         if f.endswith(".py") \
-        ]
+         if f.endswith(".py")]
```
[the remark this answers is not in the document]
jalalium: nope

tests/epyccel/test_functionals.py
jalalium: This test is not supported in C. I don't see what you are referring to.

pyccel/codegen/printing/pycode.py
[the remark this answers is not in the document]
jalalium: Now we wait till the semantic stage to deduce the target_type.

pyccel/parser/syntactic.py
```diff
@@ -1243,20 +1272,34 @@ def _visit_GeneratorExp(self, stmt):
...
-            body = generators.pop()
+
+        while(len(generators)) > 1:
+            indices.append(generators[-1].target)
+            outter_loop = generators.pop()
+            inserted_into = generators[-1]
+            if inserted_into.body.body:
+                inserted_into.body.body[0].blocks[0].body.insert2body(outter_loop)
+            else:
+                inserted_into.insert2body(outter_loop)
```
EmilyBourne, line 1292: Typo:
```suggestion
            outer_loop = generators.pop()
            inserted_into = generators[-1]
            if inserted_into.body.body:
                inserted_into.body.body[0].blocks[0].body.insert2body(outer_loop)
            else:
                inserted_into.insert2body(outer_loop)
```

EmilyBourne, line 1292: The if block before the while loop seems to be the same as the one inside the while loop. Can you not do something like:
```python
         outer_loop = body
         while generators:
             inserted_into = generators[-1]
￼            if inserted_into.body.body:
￼                inserted_into.body.body[0].blocks[0].body.insert2body(outer_loop)
￼            else:
￼                inserted_into.insert2body(outer_loop)
￼            indices.append(generators[-1].target)
￼            outer_loop = generators.pop()

        expr = FunctionalX(outer_loop, result, lhs, indices)
```
to reduce code duplication?

pyccel/parser/syntactic.py
```diff
@@ -1243,20 +1272,34 @@ def _visit_GeneratorExp(self, stmt):
...
-        generators = list(generators)
-        while len(generators) > 0:
-            indices.append(generators[-1].target)
+
+        for loop, condition in zip(generators, conditions):
+            if condition:
+                loop.insert2body(condition)
+
+        if generators[-1].body.body:
+            generators[-1].body.body[0].blocks[0].body.insert2body(body)
```
EmilyBourne, line 1281: Can we add some intermediate variables or comments to make this easier to follow?

tests/epyccel/test_functionals.py
EmilyBourne: The reason that you have added in this PR says "Skipping for C due to array redefinition issue. See #1979". #1979 has been closed as fixed so either this comment is the result of a bad merge or you need to update the issue that you refer to

tests/epyccel/test_epyccel_generators.py
```diff
@@ -199,6 +199,58 @@ def f(a0 : 'int[:]'):
...
+@pytest.mark.xfail(sys.platform == "darwin", reason="Constant('inf') printing")
+def test_min_with_condition(language):
+    def f():
+        v = min(i for i in range(20) if i % 2 == 1)
+        return v
+
+    f_epyc = epyccel(f, language = language)
+    assert f() == f_epyc()
+
+@pytest.mark.xfail(sys.platform == "darwin", reason="Constant('inf') printing")
```
EmilyBourne, line 245: Please reference an issue that describes this issue so it is easy to locate these xfails when the issue is fixed

tests/epyccel/test_arrays.py
EmilyBourne: This comment is still relevant

pyccel/ast/sympy_helper.py
```diff
@@ -106,6 +107,14 @@ def sympy_to_pyccel(expr, symbol_map):
...
             return MathCeil(arg)
 
+    elif isinstance(expr, sp.Min):
+        args = [sympy_to_pyccel(a, symbol_map) for a in expr.args]
+        return PythonMin(*args)
+
+    elif isinstance(expr, sp.Max):
+        args = [sympy_to_pyccel(a, symbol_map) for a in expr.args]
+        return PythonMax(*args)
+
```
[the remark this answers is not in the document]
EmilyBourne: @jalalium any way to add a short test for this or do we just need to tell the bot to ignore it?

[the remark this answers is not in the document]
EmilyBourne: Never mind you mentioned this above:
https://github.com/pyccel/pyccel/pull/1977#discussion_r1869791247

/bot accept
Included for completeness

pyccel/ast/sympy_helper.py
```diff
@@ -217,6 +226,14 @@ def pyccel_to_sympy(expr, symbol_map, used_names):
...
         return sp.ceiling(pyccel_to_sympy(expr.args[0], symbol_map, used_names))
 
+    elif isinstance(expr, PythonMin):
+        args = [pyccel_to_sympy(ee, symbol_map, used_names) for e in expr.args for ee in e]
+        return sp.Min(*args)
+
+    elif isinstance(expr, PythonMax):
+        args = [pyccel_to_sympy(ee, symbol_map, used_names) for e in expr.args for ee in e]
+        return sp.Max(*args)
+
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
Included for completeness

pyccel/ast/sympy_helper.py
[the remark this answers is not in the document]
EmilyBourne: /bot accept
Included for completeness

pyccel/ast/functionalexpr.py
```diff
@@ -56,9 +56,16 @@ class FunctionalFor(TypedAstNode):
     target_type : PyccelSymbol, optional
         The type of the result of the functional for. This is useful at
         the syntactic stage to pass along the final type of the lhs (list/set/array/etc).
+    conditions : list[If|None], optional
+        A list of filter conditions corresponding to each for-loop in the comprehension.
+        Each element of this list is either an `If` instance that describes the filtering
+        condition for that loop, or `None` if no condition is applied in that loop.
```
EmilyBourne, line 62: Does this fix #1410 ? It should be mentioned in the CHANGELOG

pyccel/codegen/printing/fcode.py
```diff
@@ -2311,7 +2311,7 @@ def _print_Allocate(self, expr):
             elif expr.alloc_type == 'reserve':
                 var_code = self._print(expr.variable)
                 size_code = self._print(expr.shape[0])
-                return f'call {var_code} % reserve({size_code})\n'
+                return f'call {var_code} % clear()\n' + f'call {var_code} % reserve({size_code})\n'
```
EmilyBourne, line 2314: The `+` seems unnecessary but I guess you introduced it for readability. What about:
```suggestion
                return (f'call {var_code} % clear()\n'
                        f'call {var_code} % reserve({size_code})\n')
```
?

pyccel/codegen/printing/pycode.py
```diff
@@ -660,15 +666,22 @@ def _print_For(self, expr):
...
-                        for idx, iters in zip(expr.indices, iterators)])
+        condition = ''
+        if isinstance(body, Assign):
+            body = self._print(body.rhs)
+        else:
+            assert isinstance(body, ListAppend)
+            body = self._print(body.args[0])
+
+        for_loops = ' '.join([f'for {self._print(idx)} in {self._print(iters)}{" if " + self._print(condition.blocks[0].condition) if condition else ""}'
+                             for idx, iters, condition in zip(expr.indices, iterators, expr.conditions)])
```
EmilyBourne, line 677:
```suggestion
        for_loops = ' '.join(f'for {self._print(idx)} in {self._print(iters)}{" if " + self._print(condition.blocks[0].condition) if condition else ""}'
                             for idx, iters, condition in zip(expr.indices, iterators, expr.conditions))
```

pyccel/codegen/printing/pycode.py
```diff
@@ -660,15 +666,22 @@ def _print_For(self, expr):
...
 
-        name = self._aliases.get(type(expr),'array')
-        if name == 'array':
+        name = expr.target_type
+        if 'array' in str(name):
```
EmilyBourne, line 680: Why:
`if 'array' in str(name):`
not:
`if 'array' == str(name):`
?

What additional case are you catching here?

pyccel/codegen/printing/pycode.py
```diff
@@ -684,8 +697,8 @@ def _print_GeneratorComprehension(self, expr):
                     rhs = type(body.rhs)(*args)
 
         body = self._print(rhs)
-        for_loops = ' '.join(['for {} in {}'.format(self._print(idx), self._print(iters))
-                        for idx, iters in zip(expr.indices, iterators)])
+        for_loops = ' '.join([f'for {self._print(idx)} in {self._print(iters)}{" if " + self._print(condition.blocks[0].condition) if condition else ""}'
+                             for idx, iters, condition in zip(expr.indices, iterators, expr.conditions)])
```
EmilyBourne, line 701:
```suggestion
        for_loops = ' '.join(f'for {self._print(idx)} in {self._print(iters)}{" if " + self._print(condition.blocks[0].condition) if condition else ""}'
                             for idx, iters, condition in zip(expr.indices, iterators, expr.conditions))
```

pyccel/parser/semantic.py
```diff
@@ -1966,9 +1969,13 @@ def _assign_GeneratorComprehension(self, lhs_name, expr):
         elif isinstance(expr, FunctionalMin):
             val = math_constants['inf']
             d_var = self._infer_type(result)
+            if d_var['class_type'] is PythonNativeInt():
+                val = LiteralInteger(2**63-1)
```
EmilyBourne, line 1973: Did you create an issue about these values?

pyccel/parser/semantic.py
```diff
@@ -4167,31 +4221,64 @@ def _visit_FunctionalFor(self, expr):
...
-            errors.report(LIST_OF_TUPLES, symbol=expr, severity='error')
+            errors.report(LIST_OF_TUPLES, symbol=expr, severity='fatal')
 
         target.invalidate_node()
+        operations = []
+        assign = None
+        if target_type_name != 'list':
+            old_index   = expr.index
+            new_index   = self.scope.get_new_name()
+            expr.substitute(old_index, new_index)
```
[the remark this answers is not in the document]
EmilyBourne: Why do you need to substitute in the syntactic object?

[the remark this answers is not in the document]
EmilyBourne: It seems that the old index is defined as : `index = PyccelSymbol('_', is_temp=True)`

Beware of the `is_equivalent` argument added in #2107

EmilyBourne (changes requested):
This looks much better than the last version but I still have several questions/comments.

pyccel/ast/functionalexpr.py
jalalium: Yes, it does. Added!

pyccel/ast/functionalexpr.py
```diff
@@ -56,9 +56,16 @@ class FunctionalFor(TypedAstNode):
...
+    operations : dict
+        A dictionary mapping each type of comprehension (e.g. list, array, etc.)
+        to the operation used for populating it.
```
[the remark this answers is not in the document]
jalalium: It depends on the stage. It needs to be passed from the syntactic stage to the semantic. but afterwards it is not required anymore

pyccel/parser/semantic.py
```diff
@@ -3970,90 +3977,140 @@ def _visit_For(self, expr):
...
+            class_type = dvar.pop('class_type')
+            if class_type.rank > 0:
+                class_type = class_type.switch_rank(class_type.rank - 1)
+                dvar['shape'] = dvar['shape'][1:]
+            if class_type.rank == 0:
+                dvar['shape'] = None
+                dvar['memory_handling'] = 'stack'
+            var = self._create_variable(var_name, class_type, None, dvar, insertion_scope=var_scope)
+            dvar['class_type'] = class_type
+            variables.append((var, dvar))
```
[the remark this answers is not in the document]
jalalium: I don't so, `For` handles iterable variables differently

pyccel/parser/semantic.py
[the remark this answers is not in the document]
jalalium: because it gives me this error:
```bash
 Variables named '_' are assumed to be throwaways so their value cannot be used
```

pyccel/parser/semantic.py
EmilyBourne: Hmm, you are right. The test is not passing anyway as a different error (`AttributeError`) is being raised. I guess if you had to change this it is because you somehow fixed that error. I would rather you move the test to the right folder instead of changing the error level. I will see if I can create a quick fix PR to correct the `AttributeError` explicitly

pyccel/ast/functionalexpr.py
```diff
@@ -56,9 +56,16 @@ class FunctionalFor(TypedAstNode):
     target_type : PyccelSymbol, optional
         The type of the result of the functional for. This is useful at
         the syntactic stage to pass along the final type of the lhs (list/set/array/etc).
+    conditions : list of If or None, optional
```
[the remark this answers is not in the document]
EmilyBourne: Just to confirm sometimes `conditions = [None]`? `list[If|None]` means a list containing either `None` or `If` instances. From you description it seems that the type should simply be `list[If], optional`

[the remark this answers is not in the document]
EmilyBourne: I guess this is answered by the discussion 100 lines below

pyccel/parser/syntactic.py
```diff
@@ -1148,26 +1148,70 @@ def _visit_For(self, stmt):
...
+            return index, operations
+
 
         result = self._visit(stmt.elt)
+        output_type = None
 
-        generators = list(self._visit(stmt.generators))
+        comprehensions = list(self._visit(stmt.generators))
+        generators, conditions = zip(*comprehensions)
+        generators, conditions = map(list, [generators, conditions])
```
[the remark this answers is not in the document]
EmilyBourne: > Your suggestion would work if we only had a 2 loops.

This is true. So what about:
```python
generators = [c[0] for c in comprehensions]
conditions = [c[1] for c in comprehensions]
```
?

From testing it seems that `zip(*comprehensions)` extracts splits a list of N tuples into a list of 2 tuples of size N. While `map(list, [generators, conditions])` simply changes the tuples to lists.

I find the above suggestion significantly more readable that the map/list/zip solution

pyccel/ast/functionalexpr.py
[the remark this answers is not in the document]
EmilyBourne: In that case I recommend mentioning this in the docstring and maybe adding an assertion. E.g:
```python
assert pyccel_stage != 'syntactic' || operations is not None
```

pyccel/parser/semantic.py
```diff
@@ -3970,90 +3977,140 @@ def _visit_For(self, expr):
...
+
+        def create_target_operations(target):
+            # this can be used to add opertations for set/dict comprehension
+            assert target in expr.operations
+            return expr.operations[target]
```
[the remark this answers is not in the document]
EmilyBourne: Maybe just document it in the `FunctionalFor` class

pyccel/parser/semantic.py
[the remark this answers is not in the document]
EmilyBourne: Yes see my second comment :+1:

Did you check #2107 ? 

It could cause problems if the user code contains `_` e.g. you should test:
```python
def functional_with_enumerate():
    a = [x + 1 for x in range(10)]
    b = [i for i,_ in enumerate(a)]
    return len(b), b[0], b[1], b[2]
```

pyccel/parser/semantic.py
EmilyBourne: https://github.com/pyccel/pyccel/pull/2245

pyccel/codegen/printing/pycode.py
jalalium: Fixed! I need to check if it is a list or a numpy array

pyccel/parser/semantic.py
```diff
@@ -3970,90 +3977,140 @@ def _visit_For(self, expr):
...
+
+        for loop, condition in zip(loops, expr.conditions):
+            if condition:
+                loop.insert2body(condition)
+
+        while len(loops) > 1:
+            outter_loop = loops.pop()
```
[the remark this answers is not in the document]
jalalium: fixed

pyccel/parser/semantic.py
```diff
@@ -3970,90 +3977,140 @@ def _visit_For(self, expr):
...
+                                symbol = expr, severity='error')
+                else:
+                    self.scope.insert_variable(var)
+
+            if isinstance(step, LiteralInteger):
+                step.invalidate_node()
             step  = pyccel_to_sympy(step , idx_subs, tmp_used_names)
-            start.invalidate_node()
+            if isinstance(start, LiteralInteger):
+                start.invalidate_node()
```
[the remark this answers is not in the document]
jalalium: You are right! I wasn't sure about these lines. they were here before

pyccel/parser/semantic.py
```diff
@@ -4167,31 +4221,64 @@ def _visit_FunctionalFor(self, expr):
...
+            assign = self._visit(array_ops[0])
+            array_ops = array_ops[1:]
+            for operation in array_ops:
+                operation.substitute(old_index, new_index)
+                operations.append(operation)
+            index = new_index
+            index = self._visit(index)
+        else:
+            index = None
+            operations.extend(create_target_operations('list'))
```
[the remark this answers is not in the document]
jalalium: This is because I dont have a way to check for arrays as in the following example:
```
from numpy import array as xx

if __name__ == '__main__':
    x = xx([i-3 for i in range(10)])
 ```
 We discussed this before

File: pyccel/parser/semantic.py (hunk 48 above)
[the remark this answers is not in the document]
Comment (EmilyBourne) [reply pyccel/parser/semantic.py]: I think I added them in order to ensure the AST tree was correct. At the time these objects were discarded here I think. If all the tests pass then I think you can remove them. I can always put it back later with a test in `tests/ast` if I notice an issue

File: pyccel/parser/semantic.py (hunk 49 above)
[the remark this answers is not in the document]
Comment (EmilyBourne) [reply pyccel/parser/semantic.py]: When we last discussed this you were doing this in the syntactic stage. Now you are in the semantic stage shouldn't it be possible with:
```python
target_conversion_func = self._visit(target_type_name)
if isinstance(target_conversion_func, PyccelFunctionDef) and target_conversion_func.cls_base is NumpyArray:
```

pyccel/ast/functionalexpr.py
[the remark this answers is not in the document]
jalalium: Actually, it does not depend only on the stage

pyccel/parser/syntactic.py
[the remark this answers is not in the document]
jalalium: fixed!

pyccel/parser/syntactic.py
```diff
@@ -1222,40 +1264,25 @@ def _visit_ListComp(self, stmt):
...
+        generators = []
+        indices = []
+        condition = None
 
         result = self._visit(stmt.elt)
 
-        generators = self._visit(stmt.generators)
+        comprehensions = list(self._visit(stmt.generators))
+        generators, conditions = zip(*comprehensions)
+        generators, conditions = map(list, [generators, conditions])
```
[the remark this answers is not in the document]
jalalium: fixed!

pyccel/parser/semantic.py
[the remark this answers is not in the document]
jalalium: I checked #2107 but i don't follow, sorry. I also read your second comment.
You are right! `b = [_ for _, i in enumerate(a)]` wouldn't work if that's what you meant.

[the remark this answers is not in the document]
EmilyBourne: > You are right! b = [_ for _, i in enumerate(a)] wouldn't work if that's what you meant.

No that is not what I meant. Maybe this is not the best example. I was trying to think of a case where a user has used `_` **and** you have `index = PyccelSymbol('_', is_temp=True)`. I think that this case would fail currently but would pass if you do:
```suggestion
            old_index   = expr.index
            new_index   = self.scope.get_new_name()
            expr.substitute(old_index, new_index, is_equivalent = True)
```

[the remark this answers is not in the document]
EmilyBourne: See the necessary change here:
https://github.com/pyccel/pyccel/pull/2107/files#diff-9ba0c438140113a543a41e16000b939989e841d4fc6a6a4112b5c8e5d8807524L3910-R3963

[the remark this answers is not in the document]
EmilyBourne: If you have `is_equivalent = False` then your substitution will substitute all PyccelSymbol('_'), if you have `is_equivalent = True` then your substitution will substitute only the symbols which have exactly the same id as `old_index` (and are therefore sure to represent the same thing)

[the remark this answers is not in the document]
EmilyBourne: Here is an example:
```python
import numpy as np

def f(b : 'float[:]'):
    a = np.array([i for i,_ in enumerate(b)])
```
On your branch this fails with:
```
|error [semantic]: tmp.py [4,17]| Object has already been defined with type 'int'. Type 'numpy.float64' in assignment is incompatible ('Dummy_0001=b[i]')
```
but with the suggested change it translates correctly

pyccel/parser/semantic.py
jalalium: fixed

pyccel/parser/semantic.py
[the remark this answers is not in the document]
jalalium: fixed

pyccel/parser/semantic.py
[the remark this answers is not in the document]
jalalium: fixed

pyccel/parser/semantic.py
```diff
@@ -3970,90 +3977,140 @@ def _visit_For(self, expr):
...
         i = 0
-        while isinstance(body, For):
+        loops = list(expr.loops)
+
+        # Inner function to handle PythonNativeInt variables
+        def handle_int_loop_variable(var_name, var_scope):
+            indices.append(var_name)
+            var = self._create_variable(var_name, PythonNativeInt(), None, {},insertion_scope=var_scope)
+            dvar = self._infer_type(var)
+            variables.append((var, dvar))
```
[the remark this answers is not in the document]
jalalium: fixed!

pyccel/parser/semantic.py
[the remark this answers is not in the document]
jalalium: fixed through the `FunctionalFor` class

pyccel/parser/semantic.py
```diff
@@ -3970,90 +3977,140 @@ def _visit_For(self, expr):
...
+
+            for var, dvar in variables:
+                existing_var = self.scope.find(var.name, 'variables')
+                if var.name == expr.lhs:
+                    errors.report(f"Variable {var} has the same name as the left hand side",
+                            symbol = expr, severity='fatal')
+                if existing_var or var.name ==  expr.lhs:
+                    if self._infer_type(existing_var)['class_type'] != dvar['class_type']:
+                        return errors.report(f"Variable {var} already exists with different type",
+                                symbol = expr, severity='error')
```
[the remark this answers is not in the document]
jalalium: Yes this is raised for examples like:
```Python
a = [a for a in range(10)]
```

pyccel/ast/datatypes.py
```diff
@@ -358,6 +358,39 @@ def __and__(self, other):
...
+        This property returns the smallest value that can be represented by
+        this numeric datatype. It is useful for validation, boundary checks,
+        and ensuring that generated code respects the limits of the type.
+
+        Returns
+        -------
+        int
+            The minimum value that can be represented.
+        """
+        return numpy.iinfo(int).min
```
EmilyBourne, line 375: This new property needs documenting. Given the urgency of this PR I think it is ok if you create an issue and add the documentation in a second PR

pyccel/ast/functionalexpr.py
EmilyBourne: What happens if `conditions` is not provided? Should the code convert this to `[None]`?

pyccel/ast/literals.py
```diff
@@ -474,6 +474,8 @@ def convert_to_literal(value, dtype = None):
     if isinstance(primitive_type, PrimitiveIntegerType):
         if value >= 0:
             literal_val = LiteralInteger(value, dtype)
+        elif value == -9223372036854775808:
+            literal_val =  PyccelMinus(PyccelUnarySub(LiteralInteger(-value-1, dtype)), LiteralInteger(1))
```
EmilyBourne, line 478: :vomiting_face:

EmilyBourne: Why is this necessary? Don't integers in Python adapt their precision to store any number?

pyccel/ast/numpytypes.py
```diff
@@ -118,7 +118,6 @@ class NumpyInt8Type(NumpyIntType):
     _name = 'numpy.int8'
     _precision = 1
 
-
 class NumpyInt16Type(NumpyIntType):
```
EmilyBourne, line 121: Bad change. This is unnecessary for this PR and breaks PEP recommendations : https://peps.python.org/pep-0008/#blank-lines

> Surround top-level function and class definitions with two blank lines.

pyccel/ast/numpytypes.py
```diff
@@ -176,6 +173,37 @@ class NumpyFloat64Type(NumpyNumericType):
...
+        This property returns the smallest value that can be represented by
+        this numeric datatype. It is useful for validation, boundary checks,
+        and ensuring that generated code respects the limits of the type.
+
+        Returns
+        -------
+        float64
+            The minimum value that can be represented.
+        """
+        return np.finfo(np.float64).min
```
EmilyBourne, line 190: Why is this only needed for Float64?

To be honest I think I would have preferred if these `min_value`, `max_value` methods were left for a second PR. I would have been happier with the hard coded values and an issue to say that this needs to be done cleanly

pyccel/codegen/printing/fcode.py
```diff
@@ -2311,7 +2311,8 @@ def _print_Allocate(self, expr):
             elif expr.alloc_type == 'reserve':
                 var_code = self._print(expr.variable)
                 size_code = self._print(expr.shape[0])
-                return f'call {var_code} % reserve({size_code})\n'
+                return (f'call {var_code} % clear()\n'
+                        f'call {var_code} % reserve({size_code})\n')
```
EmilyBourne, line 2315: Can the `clear` call be avoided if the status is `unallocated`?

pyccel/codegen/printing/pycode.py
```diff
@@ -660,15 +666,20 @@ def _print_For(self, expr):
...
-        name = self._aliases.get(type(expr),'array')
-        if name == 'array':
-            self.add_import(Import('numpy', [AsName(NumpyArray, 'array')]))
+        for_loops = ' '.join(f'for {self._print(idx)} in {self._print(iters)}{" if " + self._print(condition.blocks[0].condition) if condition else ""}'
+                             for idx, iters, condition in zip(expr.indices, iterators, expr.conditions))
 
-        return '{} = {}([{} {}])\n'.format(lhs, name, body, for_loops)
+        if isinstance(expr.class_type, NumpyNDArrayType):
+            self.add_import(Import('numpy', [AsName(NumpyArray, 'array')]))
+            return f'{lhs} = array([{body} {for_loops} {condition}])\n'
```
EmilyBourne, line 681: Can `_get_numpy_name` be used here to ensure the expected name?
```suggestion
        if isinstance(expr.class_type, NumpyNDArrayType):
            array = self._get_numpy_name(NumpyArray)
            return f'{lhs} = array([{body} {for_loops} {condition}])\n'
```

pyccel/parser/semantic.py
```diff
@@ -3970,90 +3973,129 @@ def _visit_For(self, expr):
...
+
+            for var, _ in variables:
+                existing_var = self.scope.find(var.name, 'variables')
+                if var.name == expr.lhs:
+                    errors.report(f"Variable {var} has the same name as the left hand side",
+                            symbol = expr, severity='fatal')
+                if existing_var or var.name ==  expr.lhs:
+                    if self._infer_type(existing_var)['class_type'] != var.class_type:
+                        return errors.report(f"Variable {var} already exists with different type",
+                                symbol = expr, severity='error')
```
EmilyBourne, line 4086: Strange to return an error. Should this be fatal or should it be:
```suggestion
                        errors.report(f"Variable {var} already exists with different type",
                                symbol = expr, severity='error')
```
?

pyccel/parser/semantic.py
```diff
@@ -4149,7 +4185,10 @@ def _visit_FunctionalFor(self, expr):
             errors.report("Unrecognised output type from functional for.\n"+PYCCEL_RESTRICTION_TODO,
                           symbol=expr,
                           severity='fatal')
-
+        if target_type_name != 'list' and any(cond is not None for cond in  expr.conditions):
```
EmilyBourne, line 4188: According to its docstring `conditions` can be `None`. This line will raise an error in this case. As mentioned above I think conditions should either be non-optional or `None` should be converted to `[None]*depth` in the constructor

pyccel/parser/semantic.py
```diff
@@ -4176,22 +4215,61 @@ def _visit_FunctionalFor(self, expr):
             errors.report(LIST_OF_TUPLES, symbol=expr, severity='error')
 
         target.invalidate_node()
+        operations = []
+        assign = None
+        target_conversion_func = self._visit(target_type_name)
+        if (isinstance(target_conversion_func, PyccelFunctionDef)
+            and target_conversion_func.cls_name is NumpyArray):
```
EmilyBourne, line 4222: For readability so the condition does not look like the start of the code
```suggestion
        if (isinstance(target_conversion_func, PyccelFunctionDef)
                and target_conversion_func.cls_name is NumpyArray):
```

pyccel/parser/semantic.py
```diff
@@ -4176,22 +4215,61 @@ def _visit_FunctionalFor(self, expr):
...
+            new_index   = self.scope.get_new_name()
+            expr.substitute(old_index, new_index, is_equivalent = lambda x,y: x is y)
+            array_ops = expr.operations['numpy_array']
+            assign = array_ops[0]
+            assign.substitute(old_index, new_index)
+            assign = self._visit(array_ops[0])
+            array_ops = array_ops[1:]
+            for operation in array_ops:
+                operation.substitute(old_index, new_index)
+                operations.append(operation)
```
EmilyBourne, line 4233:
```suggestion
            expr.substitute(old_index, new_index, is_equivalent = lambda x,y: x is y)
            array_ops = expr.operations['numpy_array']
            for operation in array_ops:
                operation.substitute(old_index, new_index, is_equivalent = lambda x,y: x is y)
            assign = array_ops[0]
            assign = self._visit(array_ops[0])
            operations.extend(array_ops[1:])
```

EmilyBourne (changes requested):

pyccel/ast/numpytypes.py
EmilyBourne: > Why is this only needed for Float64?

I.e. why is it not needed for Float32?

pyccel/ast/literals.py
jalalium: This removes the following warning:
```python
if __name__ == '__main__':
    x = -9223372036854775808
```
```bash
/home/jalal/Documents/pyccel_work/pyccel/codegen/compiling/compilers.py:505: UserWarning: /home/jalal/Documents/pyccel/__pyccel__/prog_boo.c: In function ‘main’:
/home/jalal/Documents/pyccel/__pyccel__/prog_boo.c:9:1: warning: integer constant is so large that it is unsigned
    9 |     x = -INT64_C(9223372036854775808);
      | ^   ~~~~~~~~~~~~~~~~

  warnings.warn(UserWarning(err))
  ```
  `9223372036854775808` should not be represented without a `-` sign

File: pyccel/parser/semantic.py (hunk 60 above)
Comment (jalalium) [reply pyccel/parser/semantic.py:4188]: You are right! I will update it!

File: pyccel/ast/literals.py (hunk 54 above)
Comment (EmilyBourne) [reply pyccel/ast/literals.py:478]: Ah interesting. I would expect this to be printed as:
```c
x = INT64_C(-9223372036854775808);
```
I will create an issue

pyccel/ast/numpytypes.py
jalalium: It is needed for all numpy types i e
```python
a : 'float32[:]' ...
x = min( i+1 for i in a) 
```
this would require `min` and `max` values for `Float32`.

I need to add them in a separate PR and add tests for coverage.

pyccel/ast/literals.py
EmilyBourne: Actually moving `INT64_C` doesn't help:
```
/home/emily/Code/test/tmp/__pyccel__/prog_tmp_val.c:7:1: warning: integer constant is so large that it is unsigned
    7 |     x = INT64_C(-9223372036854775808);
      | ^   ~~~~~~~~~~~~~~~~
```
Fortran also raises a similar error:
```
/home/emily/Code/test/tmp/__pyccel__/prog_tmp_val.f90:11:30:

   11 |   x = -9223372036854775808_i64
      |                              1
Error: Integer too big for its kind at (1). This check can be disabled with the option ‘-fno-range-check’
```

EmilyBourne: I guess it would make more sense to use : https://en.cppreference.com/w/c/types/limits
and `Huge` in fortran (see https://github.com/pyccel/pyccel/blob/b27a7a8eb3df2a65a1d0bb4004edbf3419884ac2/pyccel/codegen/printing/fcode.py#L2174-L2178)

Related to the comment below where I would rather have this in a separate PR

pyccel/codegen/printing/fcode.py
jalalium: good catch!

pyccel/parser/semantic.py
jalalium: You are right! it should be fatal! like in this case:
```python
    i = 5
    a = [3.4]
    b = [i-1 for i in a]
```

pyccel/codegen/printing/pycode.py
jalalium: fixed

pyccel/ast/numpytypes.py
jalalium: fixed

pyccel/ast/literals.py
jalalium: I will update this in a separate issue

pyccel/ast/functionalexpr.py
```diff
@@ -68,14 +75,19 @@ def __init__(
         lhs=None,
         indices=None,
         index=None,
-        target_type=None
+        target_type=None,
+        conditions=None,
+        operations=None
         ):
```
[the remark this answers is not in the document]
jalalium: You mean conditions right?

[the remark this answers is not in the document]
EmilyBourne: Ah possibly sorry. This wasn't a suggestion intended to be applied, it was just a suggestion to give you an idea of what I meant. Up to you to choose the appropriate location for `*` (the comment system wouldn't let me suggest it anywhere else) and remove `None` where there is no default

pyccel/codegen/printing/ccode.py
```diff
@@ -1876,7 +1876,7 @@ def _print_Allocate(self, expr):
             variable_address = self._print(ObjectAddress(expr.variable))
             container_type = self.get_c_type(expr.variable.class_type)
             if expr.alloc_type == 'reserve':
-                return free_code + f'{container_type}_reserve({variable_address}, {size});\n'
+                free_code = f'{container_type}_clear({variable_address});\n'
```
EmilyBourne, line 1879: What happened to the call to reserve?

pyccel/parser/semantic.py
```diff
@@ -3970,90 +3973,129 @@ def _visit_For(self, expr):
...
+
+        # Inner function to handle iterable variables
+        def handle_iterable_variable(var_name, element, var_scope):
+            indices.append(var_name)
+            dvar = self._infer_type(element)
+            class_type = dvar.pop('class_type')
+            if class_type.rank > 0:
+                class_type = class_type.switch_rank(class_type.rank - 1)
+                dvar['shape'] = dvar['shape'][1:]
+            if class_type.rank == 0:
```
EmilyBourne, line 4016:
```suggestion
            elif class_type.rank == 0:
```

tests/epyccel/modules/functionals.py
```diff
@@ -60,3 +60,39 @@ def functional_for_3d_range():
 def unknown_length_functional(x : 'int[:]'):
     a = [i*3 for i in range(len(x)*2)]
     return len(a), a[0], a[-1]
+
+def functional_with_enumerate():
+    a = [x + 1 for x in range(10)]
+    b = [i * j for i,j in enumerate(a)]
+    return len(b), b[0], b[1], b[2]
```
EmilyBourne, line 67: No need to change but just a remark. You realise that you could simply do:
```suggestion
    return b
```
now right?

jalalium: Thank you! I think I wrote this before it was possible

pyccel/parser/semantic.py
```diff
@@ -3970,90 +3973,129 @@ def _visit_For(self, expr):
...
+                dvar['shape'] = None
+                dvar['memory_handling'] = 'stack'
+            var = self._create_variable(var_name, class_type, None, dvar, insertion_scope=var_scope)
+            dvar['class_type'] = class_type
+            return var, dvar
```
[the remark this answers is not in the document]
jalalium: Sorry, I stopped using it. refactored!

pyccel/parser/semantic.py
jalalium: Actually no, the rank is reduced by one in the first condition. both conditions need to visited sometimes

pyccel/codegen/printing/ccode.py
jalalium: Sorry! fixed!

pyccel/codegen/printing/ccode.py
```diff
@@ -1876,7 +1876,10 @@ def _print_Allocate(self, expr):
             variable_address = self._print(ObjectAddress(expr.variable))
             container_type = self.get_c_type(expr.variable.class_type)
             if expr.alloc_type == 'reserve':
-                return free_code + f'{container_type}_reserve({variable_address}, {size});\n'
+                if expr.status == 'allocated':
+                    return (f'{container_type}_clear({variable_address});\n'
+                            f'{container_type}_reserve({variable_address}, {size});\n')
+                return f'{container_type}_reserve({variable_address}, {size});\n'
```
EmilyBourne, line 1882:
```suggestion
                if expr.status != 'unallocated':
                    return (f'{container_type}_clear({variable_address});\n'
                            f'{container_type}_reserve({variable_address}, {size});\n')
                return f'{container_type}_reserve({variable_address}, {size});\n'
```
`clear` should also be called if the status is unknown (e.g maybe allocated in an if, maybe not)

PR #1981, opened 2024-09-04
Fix gFTL installation 

### LINKED ISSUE #1978: Fix gFTL installation
### Describe the bug

gFTL has a CMake file to create and install the .inc files. Currently Pyccel includes the .m4 files in `gFTL/include/v2/parameters` which can be used to generate the .inc files but not the .inc files.

The .inc files should be generated during installation and packaged into Pyccel.

EmilyBourne:
The installation hook is updated to run cmake on gFTL during the installation process. This generates the .inc files which are used by the library. Fixes #1978

The CHANGELOG is not updated as the fix is not relevant compared to previous release version.

### DIFF
--- a/install_scripts/hook.py
+++ b/install_scripts/hook.py
@@ -21,7 +21,7 @@ class CustomBuildHook(BuildHookInterface):
     *args : tuple
         See hatch docs.
     **kwds : dict
-        See hatch docs
+        See hatch docs.
     """
 
     def initialize(self, version, build_data):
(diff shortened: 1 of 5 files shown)


install_scripts/hook.py
```diff
@@ -13,6 +15,13 @@ class CustomBuildHook(BuildHookInterface):
     A class inheriting from BuildHookInterface which allows code to be run to create artifacts during the build stage
     of the package installation using hatch.
     See <https://hatch.pypa.io/latest/plugins/build-hook/reference> for more details.
+
+    Parameters
+    ----------
+    *args : tuple
+        See hatch docs.
+    **kwds : dict
+        See hatch docs
```
EmilyBourne, line 24:
```suggestion
        See hatch docs.
```

jalalium (approved):
Very good job!

PR #1982, opened 2024-09-06
Add  support for list.append method 

jalalium:
Add support for list.append method


docs/builtin-functions.md
```diff
@@ -78,7 +78,7 @@ Python contains a limited number of builtin functions defined [here](https://doc
 
 | Method | Supported |
 |----------|-----------|
-| `append` | Python-only |
+| `append` | Python and C only |
```
EmilyBourne, line 81: Do you think we could add Fortran to our would you rather do that in a separate PR?

EmilyBourne (changes requested):

docs/builtin-functions.md
jalalium: Done!

EmilyBourne: Great! Thanks :)

tests/conftest.py
```diff
@@ -30,6 +30,16 @@ def language(request):
...
 
+@pytest.fixture( params=[
+        pytest.param("fortran", marks = pytest.mark.fortran),
+        pytest.param("c", marks = pytest.mark.c),
+        pytest.param("python", marks = pytest.mark.python),
+    ],
+    scope = "session"
+)
+def gftl_stc_language(request):
+    return request.param
```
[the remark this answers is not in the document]
jalalium: it is the same! but in the list module, there is already a language fixture that accounts for the limitations with a module scope.

```

@pytest.fixture( params=[
            pytest.mark.skip(reason="list methods not implemented in fortran"),
            pytest.mark.fortran]),
        pytest.param("c", marks = [
            pytest.mark.skip(reason="list methods not implemented in c"),
            pytest.mark.c]),
    ],
    scope = "module"
)
def language(request):
    return request.param
```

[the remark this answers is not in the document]
EmilyBourne: Ah that's a bit annoying :sweat_smile: Can we rename that one so we don't end up with fixtures all over the place?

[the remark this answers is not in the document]
jalalium: Sure!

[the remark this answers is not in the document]
jalalium: Done [61494f8](https://github.com/pyccel/pyccel/pull/1982/commits/61494f8e3e3c13820429615539aad5db33990641)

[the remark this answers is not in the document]
EmilyBourne: Perfect. Thanks :)

pyccel/codegen/printing/ccode.py
```diff
@@ -2580,6 +2580,13 @@ def _print_ClassDef(self, expr):
         return methods + interfaces
 
     #================== List methods ==================
+    def _print_ListAppend(self, expr):
+        target = expr.list_obj
+        class_type = target.class_type
+        c_type = self.get_c_type(class_type)
+        arg = self._print(expr.args[0])
+        return f'{c_type}_push(&{target.name}, {arg});\n'
```
EmilyBourne, line 2588: You shouldn't use `&{target.name}`. What happens in any of the following situations:
1. `target` is an optional argument (so stored in a pointer already)
2. `target` is a class variable
3. `target` is an element of a list
4. etc

EmilyBourne:
```suggestion
        target = expr.list_obj
        class_type = target.class_type
        c_type = self.get_c_type(class_type)
        target_code = self._print(ObjectAddress(target))
        arg = self._print(expr.args[0])
        return f'{c_type}_push({target_code}, {arg});\n'
```

EmilyBourne (changes requested):

pyccel/codegen/printing/ccode.py
jalalium: Yes, you are right!
I should use this instead:
```python
    def _print_ListAppend(self, expr):
        target = expr.list_obj
        class_type = target.class_type
        c_type = self.get_c_type(class_type)
        arg = self._print(expr.args[0])
        list_obj = self._print(ObjectAddress(expr.list_obj))
        return f'{c_type}_push({list_obj}, {arg});\n'
```
right?

EmilyBourne: Yes exactly

jalalium: oops i didn't see your other comment. Shall I go with yours or [this](https://github.com/pyccel/pyccel/pull/1982/commits/9331b105298b1d69a8273dfd246a6db1828f25e0)

EmilyBourne: As you prefer. Yours is already committed so let's go with that

EmilyBourne: I only gave a suggestion to be able to request changes ^^

EmilyBourne (approved):

PR #1988, opened 2024-09-23
Add support for NumPy v2

EmilyBourne:
Add support for NumPy v2 by fixing the return type of `np.floor` and updating the tests.

**Commit Summary**
- Remove NumPy version restriction
- Ensure `np.floor` returns an integer in NumPy v2 and a float in NumPy v1 when given an integer argument
- Fix type changes due to mixing of native and NumPy scalar types

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -77,7 +77,7 @@ jobs:
       - name: Install Pyccel with tests
         run: |
           python -m pip install --upgrade pip
-          python -m pip install .[test]
+          python -m pip install .[test] "numpy<2.0"
         shell: bash
       - name: Coverage install
         uses: ./.github/actions/coverage_install
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ All notable changes to this project will be documented in this file.
 -   #1544 : Add support for `typing.TypeAlias`.
 -   #1583 : Allow inhomogeneous tuples in classes.
 -   #1915 : Add support for NumPy v2 `sign` function.
+-   #1988 : Add support for NumPy v2.
 -   \[INTERNALS\] Added `container_rank` property to `ast.datatypes.PyccelType` objects.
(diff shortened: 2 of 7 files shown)


yguclu (commented):
Supporting both versions 1 and 2 of NumPy sounds great!

We should test them both in our CI workflows, though...

jalalium (approved):
looks good to me!

EmilyBourne:
> Supporting both versions 1 and 2 of NumPy sounds great!
> 
> We should test them both in our CI workflows, though...

NumPy v1 is tested by the Linux pipeline which uses Python 3.8. NumPy v2 is tested by the MacOSX pipeline which uses Python 3.12.

NumPy v2 only supports Python >= 3.9. However Python 3.8 reaches EoL in October. At that point we may have to change the pipeline to ensure NumPy v1 is tested. I guess it depends if we want to support both NumPy v1 and NumPy v2 in Pyccel v2?

yguclu:
> > Supporting both versions 1 and 2 of NumPy sounds great!
> > We should test them both in our CI workflows, though...
> 
> NumPy v1 is tested by the Linux pipeline which uses Python 3.8. NumPy v2 is tested by the MacOSX pipeline which uses Python 3.12.

Excellent!

> 
> NumPy v2 only supports Python >= 3.9. However Python 3.8 reaches EoL in October. At that point we may have to change the pipeline to ensure NumPy v1 is tested. I guess it depends if we want to support both NumPy v1 and NumPy v2 in Pyccel v2?

Good question. According to the guidelines of Scientific Python (see [SPEC 0](https://scientific-python.org/specs/spec-0000)), we should support NumPy v1 for another year.

EmilyBourne:
> Good question. According to the guidelines of Scientific Python (see [SPEC 0](https://scientific-python.org/specs/spec-0000)), we should support NumPy v1 for another year.

Ok. In that case I have ensured that the linux test uses NumPy v1 [a6f2e1f](https://github.com/pyccel/pyccel/pull/1988/commits/a6f2e1fb9a126b7b010be498c9767821b7270211) so this is not forgotten when Python 3.8 is deprecated

Issue #1956, opened 2024-07-30
Alias information should be in type description


EmilyBourne:
### Describe the bug
It is not possible to tell if the element of a list/tuple is an alias. This leads to confusion about who is responsible for the deallocation

### To Reproduce
Provide code to reproduce the behavior:
```python
def f():
    a = [1,2,3]
    b = [a]
    ...
    return a
```
When b is deallocated it deallocates its members which leads to a being deallocated. Here it shouldn't be as it is returned. If we don't deallocate the members then we have problems in the following case:
```python
def g():
    b = [[1,2,3]]
```

### Expected behavior
We could create a `AliasType` class which takes a data type as argument and is itself a data type. This way we can tell from the data type what is an alias/pointer and what is a memory allocation

EmilyBourne:
As a temporary solution expressions such as:
```python
b = [[1,2,3]]
```
could be translated by creating intermediate variables:
```python
Dummy_0000 = [1,2,3]
b = [Dummy_0000]
```

EmilyBourne:
> As a temporary solution expressions such as:
> 
> ```python
> b = [[1,2,3]]
> ```
> 
> could be translated by creating intermediate variables:
> 
> ```python
> Dummy_0000 = [1,2,3]
> b = [Dummy_0000]
> ```

This solution works but prevents returning lists of lists

EmilyBourne:
This is not the right strategy. This issue is replaced by #2210

PR #1989, opened 2024-09-23
Add a temporary warning for nested list

EmilyBourne:
As mentioned in #1956 nested lists do not yet function correctly. This can be fixed after #1959 but in the meantime a warning is added to discourage the use of this in user code.


jalalium (approved):
looks good to me!

PR #1994, opened 2024-09-24
Fix type bugs with NumPy v2

EmilyBourne:
After NumPy v2 the NumPy precision is preferred over the precision provided by any native Python types involved in calculations. This causes the result types of arithmetic operators to change. This was not a problem in the devel branch due to our use of `np.result_type` to calculate the types using information about both the type and the precision.

This PR fixes the problem to prepare the patch release. It does this by ignoring native types. The fix handles most cases but is not perfect. It is very difficult to do better without making the changes in #1756.

### DIFF
--- a/pyccel/ast/internals.py
+++ b/pyccel/ast/internals.py
@@ -414,7 +414,7 @@ def max_precision(objs : list, allow_native : bool = True):
         if ndarray_list:
             return get_final_precision(max(ndarray_list, key=attrgetter('precision')))
         if numpy_v1:
-            return max((get_final_precision(o) for o in objs), default = -1)
+            return max((get_final_precision(o) for o in objs))
         else:
             return get_final_precision(max(objs, key = lambda o : o.precision))
 
--- a/pyccel/ast/internals.py
+++ b/pyccel/ast/internals.py
@@ -414,7 +414,7 @@ def max_precision(objs : list, allow_native : bool = True):
         if ndarray_list:
             return get_final_precision(max(ndarray_list, key=attrgetter('precision')))
         if numpy_v1:
-            return max(get_final_precision(o) for o in objs, default = -1)
+            return max((get_final_precision(o) for o in objs), default = -1)
         else:
             return get_final_precision(max(objs, key = lambda o : o.precision))
 
--- a/pyccel/ast/internals.py
+++ b/pyccel/ast/internals.py
@@ -414,7 +414,7 @@ def max_precision(objs : list, allow_native : bool = True):
         if ndarray_list:
             return get_final_precision(max(ndarray_list, key=attrgetter('precision')))
         if numpy_v1:
-            return max((get_final_precision(o) for o in objs))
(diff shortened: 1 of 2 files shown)


pyccel/ast/internals.py
```diff
@@ -409,7 +413,10 @@ def max_precision(objs : list, allow_native : bool = True):
         ndarray_list = [o for o in objs if getattr(o, 'is_ndarray', False)]
         if ndarray_list:
             return get_final_precision(max(ndarray_list, key=attrgetter('precision')))
-        return max(get_final_precision(o) for o in objs)
+        if numpy_v1:
+            return max((get_final_precision(o) for o in objs), default = -1)
+        else:
+            return get_final_precision(max(objs, key = lambda o : o.precision))
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
This line is tested by macosx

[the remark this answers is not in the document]
yguclu: In the `True` case of this `if/else` block you handle the eventuality of an empty list `objs` by passing `default = -1` to the call to `max()`. In the second case you do not seem to worry about it, why is that?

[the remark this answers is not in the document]
yguclu: As another question, why not using the `key` argument also in the `True` case?
```python
return max(objs, key = get_final_precision, default = -1)
```

pyccel/ast/internals.py
```diff
@@ -409,7 +413,10 @@ def max_precision(objs : list, allow_native : bool = True):
         ndarray_list = [o for o in objs if getattr(o, 'is_ndarray', False)]
         if ndarray_list:
             return get_final_precision(max(ndarray_list, key=attrgetter('precision')))
-        return max(get_final_precision(o) for o in objs)
+        if numpy_v1:
+            return max((get_final_precision(o) for o in objs), default = -1)
```
EmilyBourne, line 417:
```suggestion
            return max((get_final_precision(o) for o in objs))
```

pyccel/ast/internals.py
```diff
@@ -414,7 +414,7 @@ def max_precision(objs : list, allow_native : bool = True):
         if ndarray_list:
             return get_final_precision(max(ndarray_list, key=attrgetter('precision')))
         if numpy_v1:
-            return max((get_final_precision(o) for o in objs), default = -1)
+            return max((get_final_precision(o) for o in objs))
```
yguclu, line 417: Do you still need the double parentheses?
```python
return max(get_final_precision(o) for o in objs)
```
should work.

EmilyBourne:
```suggestion
            return max(get_final_precision(o) for o in objs)
```

EmilyBourne: I agree. Fixed [46ee59f](https://github.com/pyccel/pyccel/pull/1994/commits/46ee59ff99baa6274ca57b1b0421a9e4e7c93915)

yguclu (approved):

PR #1999, opened 2024-09-27
Fix test failures with Intel compiler

### LINKED ISSUE #1932: Failing Intel tests mostly due to rounding errors
### Describe the bug
The intel runner reports failing tests for complexes. In most cases this is due to the equality being checked (instead of using a tolerance. E.g:
```
E         Full diff:
E           (
E               2,
E               0j,
E         -     0j,
E         -     (1+0j),
E         -     (1+0j),
E         +     (6.9252660709975e-310+1.5e-323j),
E         +     (4.64308889938656e-310+4.6430888990348e-310j),
E         +     (4.64308889932925e-310+0j),
E           )
```

However some cases such as `test_c_array_return` seem to show a trickier error:
```
array([[1.+0.j, 0.+2.j, 0.+2.125j ],\n    [0.+2.25j, 0.+2.3125j, 0.+2.375j ]], dtype=complex64)
array([[1.+0.j, 2.+0.j, 3.+0.j    ],\n    [4.+0.j  , 5.+0.j    , 6.+0.j    ]], dtype=complex64)
```

### To Reproduce
See the intel runner

### Error details
https://github.com/pyccel/pyccel/actions/runs/9938070532/job/27449745333

EmilyBourne:
Fix the intel tests in the release branch. In this branch the only errors are due to a compiler bug. The bug has been reported [here](https://community.intel.com/t5/Intel-Fortran-Compiler/Compiler-Bug-cmplx-does-not-handle-Y-argument-correctly/m-p/1634388) but a simple workaround has been implemented. This therefore fixes #1932 

While fixing this I noticed Windows tests which should have been reactivated 2 years ago (#735 was fixed in January 2022). These tests are now reactivated

### DIFF
--- a/pyccel/ast/builtins.py
+++ b/pyccel/ast/builtins.py
@@ -296,8 +296,7 @@ class PythonComplex(PyccelInternalFunction):
         return super().__new__(cls)
 
     def __init__(self, arg0, arg1 = LiteralFloat(0)):
-        self._is_cast = arg0.dtype is NativeComplex() or \
-                        isinstance(arg1, Literal) and arg1.python_value == 0
+        self._is_cast = isinstance(arg1, Literal) and arg1.python_value == 0
 
         if self._is_cast:
             self._real_part = self._real_cast(arg0)
--- a/pyccel/ast/builtins.py
+++ b/pyccel/ast/builtins.py
@@ -296,7 +296,7 @@ class PythonComplex(PyccelInternalFunction):
         return super().__new__(cls)
 
     def __init__(self, arg0, arg1 = LiteralFloat(0)):
-        self._is_cast = arg0.dtype is NativeComplex() and \
+        self._is_cast = arg0.dtype is NativeComplex() or \
                         isinstance(arg1, Literal) and arg1.python_value == 0
 
         if self._is_cast:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Change Log
 All notable changes to this project will be documented in this file.
 
-## \[1.12.1\] - 2024-09-27
+## \[1.12.1\] - 2024-10-01
 
 ### Added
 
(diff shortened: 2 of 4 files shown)


pyccel/ast/builtins.py
```diff
@@ -296,8 +296,7 @@ def __new__(cls, arg0, arg1=LiteralFloat(0)):
         return super().__new__(cls)
 
     def __init__(self, arg0, arg1 = LiteralFloat(0)):
-        self._is_cast = arg0.dtype is NativeComplex() and \
-                        isinstance(arg1, Literal) and arg1.python_value == 0
+        self._is_cast = isinstance(arg1, Literal) and arg1.python_value == 0
```
yguclu, line 299: What is the exact meaning of the attribute `_is_cast`?

It will be `False` if I try to convert a complex number into another one with different precision.

EmilyBourne: If you try to convert a complex number into another one with different precision then you will have:
```
arg0 = my_complex_object,
arg1 = LiteralFloat(0)
```
so `_is_cast` will be true.

`_is_cast` simply indicates that the function was called as a cast rather than as a function assembling 2 floats/integers into a complex number.

Before this PR `_is_cast` indicated if a complex was being cast. I am not sure where this extra restriction came from. It meant that `np.complex(1+2j)` would be treated as a cast while `complex(1)` would be treated the same as `complex(1, 0)`

EmilyBourne: Probably the aim was just to reduce the number of cases but if `complex(1)` is treated the same as `complex(1, 0)` then `complex(np.ones(4))` is also treated the same as `complex(np.ones(4), 0)`. This is exactly the case that doesn't work with the intel compiler bug (it also looks harder to treat in C, but I think the unravelling must handle it)

yguclu: Thanks for the answer. We should have sufficient tests to catch any errors here.

yguclu (approved):

pyccel/ast/builtins.py
EmilyBourne: Indeed my first attempt at correcting this line was wrong and was caught :sweat_smile:

Issue #2001, opened 2024-09-27
Crashing Intel tests 


EmilyBourne:
### Describe the bug

Following b739d46f1be6f936d6ab955d88f5e55996ed2f44 the Fortran tests are crashing with the Intel compiler. This went unnoticed as #1932 hadn't been fixed yet.

EmilyBourne:
Related error `double free or corruption (out)`

EmilyBourne:
Better to free Fortran arrays from Fortran?

EmilyBourne:
This is indeed caused by Fortran arrays being deallocated from C. This was not a problem previously as it was a memory leak

jalalium:
Why not write a deallocation routine in Fortran and call it as an extern from C?

EmilyBourne:
> Why not write a deallocation routine in Fortran and call it as an extern from C?

Without returning a 1D array this is difficult to achieve as you would need a deallocation routine for every possible rank

jalalium:
> > Why not write a deallocation routine in Fortran and call it as an extern from C?
> 
> Without returning a 1D array this is difficult to achieve as you would need a deallocation routine for every possible rank

Sorry I am not following, could you provide an example where this would fail to deallocate?

```fortran
   subroutine deallocate_tensor()
      if (allocated(tensor)) then
        deallocate(tensor)
      end if
    end subroutine deallocate_tensor
```

PR #2002, opened 2024-10-01
Fix the worst Intel bugs

EmilyBourne:
This PR incorporate #1999 into the `devel` branch. #1999 included a workaround for an intel compiler bug.
In the `devel` branch the intel branch wasn't working due to #2001 . This bug is not present in the `main` branch as there is a memory leak when returning arrays from functions (these arrays are never freed). The fix for this is non-standard and does not work with the Intel compiler. This is documented in #2001 however, for now, to prevent further Intel bugs being introduced without us noticing, the call to `free` is deactivated for the Intel compiler. This allows the tests to pass

### DIFF
--- a/pyccel/codegen/python_wrapper.py
+++ b/pyccel/codegen/python_wrapper.py
@@ -36,7 +36,6 @@ def create_shared_library(codegen,
                           wrapper_flags,
                           pyccel_dirpath,
                           src_compiler,
-                          c_compiler,
                           wrapper_compiler,
                           sharedlib_modname=None,
                           verbose = False):
@@ -72,9 +71,6 @@ def create_shared_library(codegen,
 
     src_compiler : pyccel.codegen.compiling.compilers.Compiler
         The compiler which should be used to compile the library.
-
-    c_compiler : pyccel.codegen.compiling.compilers.Compiler
-        The compiler which should be used to compile C files.
 
     wrapper_compiler : pyccel.codegen.compiling.compilers.Compiler
         The compiler which should be used to compile the wrapper.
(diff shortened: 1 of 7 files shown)


pyccel/codegen/python_wrapper.py
```diff
@@ -72,6 +73,9 @@ def create_shared_library(codegen,
     src_compiler : pyccel.codegen.compiling.compilers.Compiler
         The compiler which should be used to compile the library.
 
+    c_compiler : pyccel.codegen.compiling.compilers.Compiler
+        The compiler which should be used to compile C files.
+
     wrapper_compiler : pyccel.codegen.compiling.compilers.Compiler
         The compiler which should be used to compile the wrapper.
         Often this is the same as src_compiler but it may be different
```
yguclu, line 81: The difference between these two input parameters is not clear.

pyccel/codegen/python_wrapper.py
```diff
@@ -160,7 +164,7 @@ def create_shared_library(codegen,
 
     # get the include folder path and library files
     recompile_object(cwrapper_lib,
-                      compiler = wrapper_compiler,
+                      compiler = c_compiler,
                       verbose  = verbose)
```
yguclu, line 168: In `pipeline.py` we have the line
```python
wrapper_compiler = Compiler('GNU', 'c', debug)
```
so `wrapper_compiler` is always the GCC compiler. Here you swap it with the C compiler selected by the user, but not so on lines 213-215. Why is that?

EmilyBourne: Sorry this is probably a half-baked solution (it is still in draft). We use GNU to be compatible with Python (we should probably find a way to work out which compiler Python was actually compiled with). However I'm not sure if the functions that are called from the wrapper need to be compiled. I think we are doing because several of these functions are Python functions. However I am not sure that it is actually necessary.
I will probably try using the compiler's C compiler for everything

EmilyBourne: > I will probably try using the compiler's C compiler for everything

This seems to work so I think I will stick with this solution

jalalium (approved):

PR #2004, opened 2024-10-03
Add Fortran support for set initialisation

### LINKED ISSUE #1658: [SETS] Support variable declarations in Fortran
### Describe the feature

### Test Code
```python
def declare_list():
    a : set[int]
```

### Proposed Solution
This feature should be implemented with the https://github.com/Goddard-Fortran-Ecosystem/gFTL library. The Set type should be used.

### Target Language
Fortran

### Pre-requisites
- #1656

EmilyBourne:
Add Fortran support for set initialisation in a similar way to lists. Fixes #1658

### DIFF
--- a/pyccel/codegen/printing/fcode.py
+++ b/pyccel/codegen/printing/fcode.py
@@ -66,7 +66,7 @@ from pyccel.ast.numpyext import NumpyIsFinite, NumpyIsNan
 
 from pyccel.ast.numpytypes import NumpyNDArrayType
 
-from pyccel.ast.operators import PyccelAdd, PyccelMul, PyccelMinus, PyccelAnd, PyccelEq
+from pyccel.ast.operators import PyccelAdd, PyccelMul, PyccelMinus, PyccelAnd
 from pyccel.ast.operators import PyccelMod, PyccelNot, PyccelAssociativeParenthesis
 from pyccel.ast.operators import PyccelUnarySub, PyccelLt, PyccelGt, IfTernaryOperator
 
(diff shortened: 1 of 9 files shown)


EmilyBourne:

pyccel/codegen/printing/fcode.py
```diff
@@ -560,6 +560,23 @@ def _build_gFTL_module(self, expr_type):
...
                 macros.append(MacroDefinition('Vector', expr_type))
                 macros.append(MacroDefinition('VectorIterator', IteratorType(expr_type)))
+            elif isinstance(expr_type, HomogeneousSetType):
+                include = Import(LiteralString('set/template.inc'), Module('_', (), ()))
+                element_type = expr_type.element_type
+                if isinstance(element_type, FixedSizeNumericType):
+                    tmpVar_x = Variable(element_type, 'x')
+                    tmpVar_y = Variable(element_type, 'y')
+                    if isinstance(element_type.primitive_type, PrimitiveComplexType):
+                        lt_def = PyccelAssociativeParenthesis(PyccelLt(NumpyAbs(tmpVar_x), NumpyAbs(tmpVar_y)))
```
jalalium, line 570: The modulus is not enough to define a total order on complex numbers
the following code should output 2 but it outputs 1:
```python
if __name__ == '__main__':
    s = {0+1j , 1+0j }
    print(len(s))
```
the real and imaginary parts should also be compared

jalalium (changes requested):

pyccel/codegen/printing/fcode.py
EmilyBourne: Hmm, good point. Maybe I need to create a function for this. I tried `(x.real == y.real ? x.imag < y.imag : x.real < y.real)` but when the macro was expanded it ran over the maximum line length

EmilyBourne: I have changed the test so it would have failed before my changes [1abdf73](https://github.com/pyccel/pyccel/pull/2004/commits/1abdf7307507c929f836081946a0224c7ec0f46f)

EmilyBourne: The code now uses a function

jalalium: Well done!

jalalium (approved):

Issue #1944, opened 2024-07-22
[DICTS] Support variable declarations in Fortran


EmilyBourne:
### Describe the feature
Minimal printing support should be implemented for dicts. It should be possible to declare a dict using a type annotation.

### Test Code
Provide code which does not currently work but which should do when this issue is fixed:
```python
def declare_list():
    a : dict[int, int]
```

### Proposed Solution
This feature should be implemented with the https://github.com/stclib/STC/ library. The `smap` type should be used.

### Target Language
Fortran

PR #2005, opened 2024-10-03
Add Fortran support for dict initialisation

EmilyBourne:
Add Fortran support for dict initialisation in a similar way to lists. Fixes https://github.com/pyccel/pyccel/issues/1944

### DIFF
--- a/pyccel/codegen/printing/fcode.py
+++ b/pyccel/codegen/printing/fcode.py
@@ -576,7 +576,7 @@ class FCodePrinter(CodePrinter):
                                    MacroDefinition('Key_KINDLEN(context)', KindSpecification(key_type)),
                                    MacroDefinition('Key_LT(x,y)', lt_def)])
                 else:
-                    macros.append(MacroDefinition('Key', key_type))
+                    raise NotImplementedError("Support for dicts whose keys define their own < operator is not yet implemented")
                 if isinstance(value_type, FixedSizeNumericType):
                     macros.extend([MacroDefinition('T', value_type.primitive_type),
                                    MacroDefinition('T_KINDLEN(context)', KindSpecification(value_type))])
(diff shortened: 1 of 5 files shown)


EmilyBourne:

pyccel/codegen/printing/fcode.py
```diff
@@ -560,6 +560,31 @@ def _build_gFTL_module(self, expr_type):
...
+            elif isinstance(expr_type, DictType):
+                include = Import(LiteralString('map/template.inc'), Module('_', (), ()))
+                key_type = expr_type.key_type
+                value_type = expr_type.value_type
+                macros = []
+                if isinstance(key_type, FixedSizeNumericType):
+                    tmpVar_x = Variable(key_type, 'x')
+                    tmpVar_y = Variable(key_type, 'y')
+                    if isinstance(key_type.primitive_type, PrimitiveComplexType):
+                        lt_def = PyccelAssociativeParenthesis(PyccelLt(NumpyAbs(tmpVar_x), NumpyAbs(tmpVar_y)))
```
jalalium, line 572: The modulus is not enough to define a total order on complex numbers
the following code should output 2 but it outputs 1:
```python
if __name__ == '__main__':
    s = {0+1j : 5, 1+0j : 3}
    print(len(s))
```
the real and imaginary parts should also be compared

jalalium (changes requested):

pyccel/codegen/printing/fcode.py
EmilyBourne: Fixed in the same way as for sets: [fa8166d](https://github.com/pyccel/pyccel/pull/2005/commits/fa8166d47e66d9b880898803cc0db56fcf12de0e)
Test added that would have failed with the old method: 67c526d4d

jalalium (approved):
Well done!

PR #2007, opened 2024-10-08
Improve homogeneous containers allocation scheme.

### LINKED ISSUE #1979: Memory leak after list redefinition in C
### Describe the bug
List redefinition binds the variable to the new list without dropping the old one.

### To Reproduce
```python

if __name__ == '__main__':
    a = [1,2,3]
    a = [1,2,3,4,5]
```

### Error details
```c
#include "boo.h"
#include <stdlib.h>
#include <stdint.h>
#ifndef _VEC_INT64_T
#define _VEC_INT64_T
#define i_type vec_int64_t
#define i_key int64_t
#include <stc/vec.h>
#endif // _VEC_INT64_T

int main()
{
    vec_int64_t a;
    a = c_init(vec_int64_t, {INT64_C(1), INT64_C(2), INT64_C(3)});
    a = c_init(vec_int64_t, {INT64_C(1), INT64_C(2), INT64_C(3), INT64_C(4), INT64_C(5)});
    vec_int64_t_drop(&a);
    return 0;
}
```
After compiling with `-g -fsanitize=address`

```bash
=================================================================
==90375==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 32 byte(s) in 1 object(s) allocated from:
    #0 0x7ba46f2b4c38 in __interceptor_realloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:164
    #1 0x598fea4ac9bb in vec_int64_t_reserve __pyccel__/stc/vec.h:291
    #2 0x598fea4ac356 in vec_int64_t_push __pyccel__/stc/vec.h:96
    #3 0x598fea4ac459 in vec_int64_t_put_n __pyccel__/stc/vec.h:122
    #4 0x598fea4ac5b4 in vec_int64_t_from_n __pyccel__/stc/vec.h:124
    #5 0x598fea4ad4b8 in main __pyccel__/prog_boo.c:14
    #6 0x7ba46ee29d8f in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58

SUMMARY: AddressSanitizer: 32 byte(s) leaked in 1 allocation(s).
```

### Expected behavior
The first list should be dropped before it becomes unreferenced

### Language
C

jalalium:
The goal of this PR is to resolve #1979 and to systematically address memory allocation for STC containers.
This PR aims to use one of the following to address allocation of STC containers:
- use STC `init` functions to allocate containers initialized with literal lists, sets, or dictionaries.
- use `reserve` on STC containers when the capacity is known at compile time and the container will be populated with append/add methods.
- use `resize` on STC containers when elements of the container  will be accessed through their indices in the code.
Care is taken to drop containers before they are reassigned or to reuse them whenever possible.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -35,3 +35,4 @@ Contributors
 * Shoaib Moeen
 * Kush Choudhary
 * Emmmanuel Ferdman
+* Amir Movassaghi
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -59,6 +59,7 @@ sudo apt install gfortran
 sudo apt install libblas-dev liblapack-dev
 sudo apt install libopenmpi-dev openmpi-bin
 sudo apt install libomp-dev libomp5
+sudo apt install cmake
 ```
 
 ### Linux Fedora-CentOS-RHEL
@@ -73,6 +74,7 @@ dnf install gfortran
 dnf install blas-devel lapack-devel
 dnf install openmpi-devel
 dnf install libgomp
+dnf install cmake
 exit
 ```
 
@@ -89,6 +91,7 @@ brew install openblas
 brew install lapack
 brew install open-mpi
 brew install libomp
+brew install cmake
 ```
 
 This requires that the Command Line Tools (CLT) for Xcode are installed.
@@ -103,6 +106,7 @@ In an Administrator prompt install git-bash (if needed), a Python3 distribution,
 choco install git
 choco install python3
 choco install mingw
+choco install cmake
 ```
 
 Download x64 BLAS and LAPACK DLLs from <https://icl.cs.utk.edu/lapack-for-windows/lapack/>:
--- a/pyccel/ast/builtin_methods/set_methods.py
+++ b/pyccel/ast/builtin_methods/set_methods.py
@@ -20,7 +20,6 @@ __all__ = (
     'SetDiscard',
     'SetMethod',
     'SetPop',
-    'SetRemove',
     'SetUnion',
     'SetUpdate'
 )
@@ -147,34 +146,6 @@ class SetPop(SetMethod):
     def __init__(self, set_variable):
         self._class_type = set_variable.class_type.element_type
         super().__init__(set_variable)
-
-#==============================================================================
-class SetRemove(SetMethod):
-    """
-    Represents a call to the .remove() method.
-
-    The remove() method removes the specified item from 
-    the set and updates the set. It doesn't return any value.
-
-    Parameters
-    ----------
-    set_variable : TypedAstNode
-        The set on which the method will operate.
-
-    item : TypedAstNode
-        The item to search for, and remove.
-    """
-    __slots__ = ()
-    _shape = None
-    _class_type = VoidType()
-    name = 'remove'
-
-    def __init__(self, set_variable, item) -> None:
-        if not isinstance(item, TypedAstNode):
-            raise TypeError(f"It is not possible to look for a {type(item).__name__} object in a set of {set_variable.dtype}")
-        if item.class_type != set_variable.class_type.element_type:
-            raise TypeError(f"Can't remove an element of type {item.dtype} from a set of {set_variable.dtype}")
-        super().__init__(set_variable, item)
 
 #==============================================================================
 class SetDiscard(SetMethod):
--- a/pyccel/ast/class_defs.py
+++ b/pyccel/ast/class_defs.py
@@ -6,7 +6,7 @@
 This module contains all types which define a python class which is automatically recognised by pyccel
 """
 
-from pyccel.ast.builtin_methods.set_methods  import (SetAdd, SetClear, SetCopy, SetPop, SetRemove,
+from pyccel.ast.builtin_methods.set_methods  import (SetAdd, SetClear, SetCopy, SetPop,
                                                      SetDiscard, SetUpdate, SetUnion)
 from pyccel.ast.builtin_methods.list_methods import (ListAppend, ListInsert, ListPop,
                                                      ListClear, ListExtend, ListRemove,
@@ -163,7 +163,7 @@ SetClass = ClassDef('set',
             PyccelFunctionDef('copy', func_class = SetCopy),
             PyccelFunctionDef('discard', func_class = SetDiscard),
             PyccelFunctionDef('pop', func_class = SetPop),
-            PyccelFunctionDef('remove', func_class = SetRemove),
+            PyccelFunctionDef('remove', func_class = SetDiscard),
             PyccelFunctionDef('union', func_class = SetUnion),
             PyccelFunctionDef('update', func_class = SetUpdate),
             PyccelFunctionDef('__or__', func_class = SetUnion),


pyccel/codegen/printing/ccode.py
```diff
@@ -1688,7 +1691,18 @@ def _print_Allocate(self, expr):
         free_code = ''
         variable = expr.variable
         if isinstance(variable.class_type, (HomogeneousListType, HomogeneousSetType, DictType)):
-            return ''
+            if expr.status in ('allocated', 'unknown'):
```
EmilyBourne, line 1694: Your PR description makes it sound like the unknown case is not handled

pyccel/parser/semantic.py
```diff
@@ -1570,6 +1570,14 @@ def _assign_lhs_variable(self, lhs, d_var, rhs, new_expressions, is_augassign =
                                         shape=a.alloc_shape, status=status))
                             args = new_args
                             new_args = []
+                    elif isinstance(lhs.class_type, (HomogeneousListType, HomogeneousSetType,DictType)):
+                        if isinstance(rhs, (PythonList, PythonDict, PythonSet)):
+                            alloc_type = 'init'
+                        elif rhs.get_attribute_nodes(IndexedElement):
+                            alloc_type = 'resize'
+                        else:
+                            alloc_type = 'reserve'
```
EmilyBourne, line 1579: I don't understand the logic behind this. Can you explain?

pyccel/parser/semantic.py
```diff
@@ -1738,6 +1746,14 @@ def _ensure_inferred_type_matches_existing(self, class_type, d_var, var, is_auga
                                 self.current_ast_node.col_offset))
 
                 else:
+                    alloc_type = None
+                    if isinstance(getattr(var,'class_type'), (HomogeneousListType, HomogeneousSetType,DictType)):
```
EmilyBourne, line 1750: Why do you use getattr instead of:
```suggestion
                    if isinstance(var.class_type, (HomogeneousListType, HomogeneousSetType,DictType)):
```
?

.github/actions/valgrind_run/action.yml
```diff
@@ -9,3 +9,10 @@ runs:
         valgrind --leak-check=full --error-exitcode=1 ./leaks_check
       shell: bash
       working-directory: ./tests/ndarrays
+    
+    - name: Test with valgrind for memory leaks in stc_containers
+      run: |
+        pyccel --language=c --flags="-g -O0" leaks_check.py
+        valgrind --leak-check=full --error-exitcode=1 ./leaks_check
```
EmilyBourne, line 16: Maybe we should also double check the Fortran implementation

EmilyBourne (changes requested):
This looks like a nice clean solution. Congratulations on keeping it to so few lines. I just have a few comments

pyccel/ast/core.py
```diff
@@ -398,16 +398,22 @@ class Allocate(PyccelAstNode):
         In C this provides the size which will be passed to malloc. In Fortran
         this provides the source argument of the allocate function.
 
+    alloc_type : str {'init'|'reserve'|'resize'}, optional
```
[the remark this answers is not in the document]
jalalium: Done [fix](https://github.com/pyccel/pyccel/pull/2007/commits/0f95f4fd3509251ba20413426467824932ca1999)

pyccel/codegen/printing/ccode.py
jalalium: Updated the description

pyccel/parser/semantic.py
jalalium: This is to differentiate between the uses of `resize` and `reserve`; `resize` actually sets the size; and thus the capacity; of the container to the desired size whereas `reserve` only sets its capacity and keeps its size to 0. The former should be used when we wanna access index elements of the container directly while the latter can be used when append/add will be used afterward and we only wanna reserve the adequate size to avoid repetitive reallocations that may stem from size doubling. Therefore, I need to check if the code block in the `rhs` has any `IndexedElement` nodes. My assumption that their existence is tied to them being used on the `lhs` variable might not always be true. Shall I double check this as well?

pyccel/parser/semantic.py
jalalium: That's what i wanted to use at the beginning but I found out that not all variables have that attribute.

.github/actions/valgrind_run/action.yml
jalalium: Just checked that! turns out there are memory issues with `Fortran`. Shall I create an issue for that?

pyccel/parser/semantic.py
EmilyBourne: All semantic objects inheriting from `TypedAstNode` should have this property. So maybe you found a bug? Without a default argument this function is doing the same as my proposition though:

image

So I'm not sure why this would fix anything as is. Maybe the problem was there earlier?

.github/actions/valgrind_run/action.yml
EmilyBourne: Yes please

EmilyBourne: Hopefully I guess I should be able to reuse most of this to fix the Fortran problem

pyccel/parser/semantic.py
EmilyBourne: So the first case handles:
```python
def f(a : list[int]):
    b = a[1:-1]
```
while the second case handles everything else?

Yes it would be good to check this and maybe add this case in the valgrind tests. Could you also add a comment about this here.

pyccel/ast/core.py
```diff
@@ -398,16 +398,23 @@ class Allocate(PyccelAstNode):
...
+        Defines the method of memory allocation exclusively for STC containers.
+        This parameter is irrelevant and should not be used for anything other than STC containers.
```
EmilyBourne, line 403: I'm not a fan of this description. STC containers work very similarly to gFTL containers (and containers in other languages e.g. std containers in C++) so I'm not convinced that this parameter **should not be used for anything other than STC containers**

pyccel/ast/core.py
```diff
@@ -429,11 +436,15 @@ def __init__(self, variable, *, shape, status, like = None):
         if status not in ('allocated', 'unallocated', 'unknown'):
             raise ValueError(f"Value of 'status' not allowed: '{status}'")
 
+        if alloc_type:
+            assert alloc_type in ('init', 'reserve', 'resize')
```
EmilyBourne, line 440:
```suggestion
        assert alloc_type is None or alloc_type in ('init', 'reserve', 'resize')
```

EmilyBourne: Should there also be an assertion e.g.:
```python
assert alloc_type is None || not isinstance(variable.class_type, (HomogeneousListType, HomogeneousSetType, DictType))
```
?

pyccel/parser/semantic.py
jalalium: Sure thing!
by the way, I could handle the issue we have with `b = a` here as well. Shall I do it here or does it need a PR of its own?

pyccel/ast/core.py
jalalium: Yes! this is a very good. I will add it!

.github/actions/valgrind_run/action.yml
jalalium: The only issue is about allocation for list comprehensions. Fortran does not allocate for the list before populating it. I don't think this issue is relevant in this context.

EmilyBourne: Sounds like something that can be fixed using the `resize` attribute :+1:

pyccel/parser/semantic.py
EmilyBourne: Which issue?

jalalium:
```python

if __name__ == "__main__":
    a = [1,2,3]
    b = a
    a[0] = 2
    print(b[0])
```
this code prints 1 instead of 2.
`b` should only be a pointer and not a new list

EmilyBourne: That doesn't sound related to this issue to me so I think we should fix it separately. Can you create an issue?

jalalium: It already exists #2008

.github/actions/valgrind_run/action.yml
jalalium: I created an issue for it #2036

pyccel/parser/semantic.py
EmilyBourne: Ah super :+1:

pyccel/ast/core.py
jalalium: fixed

pyccel/ast/core.py
jalalium: fixed

tests/stc_containers/leaks_check.py
```diff
@@ -0,0 +1,62 @@
...
+    a = {1,2,3,4}# pylint: disable=unused-variable
+
+def conditional_dict(b1: bool):
+    if (b1):
+        a = {1:1,2:2,3:3}# pylint: disable=unused-variable
+    a = {1:1,2:2,3:3,4:4,5:5}# pylint: disable=unused-variable
+
+def slice_assign():
+    a = [1,2,3,4]
+    b = a[1:-1]# pylint: disable=unused-variable
```
EmilyBourne, line 47: If there are this many unused-variables to disable it's probably better to do it for the whole file. You can add an exception to the bot

EmilyBourne (commented):
Looks like you're more or less done. Don't forget to take it out of draft when you're ready for a new review

tests/stc_containers/leaks_check.py
EmilyBourne: https://github.com/pyccel/pyccel/blob/devel/ci_tools%2Fcheck_pylint_commands.py#L13

EmilyBourne: The exception is necessary to stop the bot complaining about the file-level pylint disable command

tests/stc_containers/leaks_check.py
```diff
@@ -0,0 +1,62 @@
+
+# pylint: disable=missing-function-docstring, missing-module-docstring
+def create_list():
```
EmilyBourne, line 3:
```suggestion
# pylint: disable=missing-function-docstring, missing-module-docstring, unused-variable

def create_list():
```

tests/stc_containers/leaks_check.py
```diff
@@ -0,0 +1,62 @@
...
+def create_set():
+    a = {1,2,3}
+
+def create_dict():
+    a = {1:1,2:2,3:3}
+
+def list_reassign():
+    a = [1,2,3]
+    a = [1,2,3,4]
+    a = [i for i in range(15)] # pylint: disable=unnecessary-comprehension
```
EmilyBourne, line 15: Maybe we can just avoid the pylint error?
```suggestion
    a = [i*2 for i in range(15)] # pylint: disable=unnecessary-comprehension
```

tests/stc_containers/leaks_check.py
EmilyBourne: The change in the bot file doesn't make the pylint test pass. It just stops this change from causing an error in the pyccel_lint test

EmilyBourne: Codacy also runs pylint and doesn't use the bot

pyccel/parser/semantic.py
```diff
@@ -1570,6 +1570,14 @@ def _assign_lhs_variable(self, lhs, d_var, rhs, new_expressions, is_augassign =
...
                             args = new_args
                             new_args = []
+                    elif isinstance(lhs.class_type, (HomogeneousListType, HomogeneousSetType,DictType)):
+                        if isinstance(rhs, (PythonList, PythonDict, PythonSet)):
+                            alloc_type = 'init'
+                        elif isinstance(rhs, IndexedElement) or rhs.get_attribute_nodes(IndexedElement):
+                            alloc_type = 'resize'
+                        else:
+                            alloc_type = 'reserve'
+                        new_expressions.append(Allocate(lhs, shape=lhs.alloc_shape, status=status, alloc_type=alloc_type))
```
EmilyBourne, line 1580: I am concerned that this may break when we can return lists/sets/dicts from functions. What would be the correct case for this? `init`?

EmilyBourne (approved):
This all looks good to me. Happy to merge once my future-proofing question has been answered

pyccel/parser/semantic.py
jalalium: done!

EmilyBourne (approved):

PR #2014, opened 2024-10-10
Add Fortran support for set method `pop`

### LINKED ISSUE #2010: [SETS] Add Fortran support for pop()
### Test Code
```python
def foo():
    a = {1,2,1}
    b = a.pop()
    c = a.pop()
    return b,c
```

### Proposed Solution
This feature should be implemented using gFTL. Unfortunately the gFTL implementation of sets does not appear to be very developed for the moment so this issue will probably require a corresponding issue and maybe PR in gFTL.

EmilyBourne:
Add Fortran support for set method `pop`. Fixes #2010 

**Commit Summary**
- Add `MacroUndef` class to avoid macros polluting namespaces
- Simplify code using `_get_comparison_operator` to avoid duplication
- Add `_build_gFTL_extension_module` to create extensions for gFTL types
- Add support for `set.pop`
- Fix Python printing so expressions that can be used in expressions don't end in `\n`
- Activate `set.pop` tests
- Activate `test_init_` functions which don't require specific language support

### DIFF
--- a/pyccel/ast/low_level_tools.py
+++ b/pyccel/ast/low_level_tools.py
@@ -151,9 +151,9 @@ class MacroUndef(PyccelAstNode):
     @property
     def macro_name(self):
         """
-        The name of the macro being defined.
+        The name of the macro being undefined.
 
-        The name of the macro being defined.
+        The name of the macro being undefined.
         """
         return self._macro_name
 
--- a/pyccel/ast/low_level_tools.py
+++ b/pyccel/ast/low_level_tools.py
@@ -12,7 +12,8 @@ from .datatypes import PyccelType
 
 __all__ = ('IteratorType',
            'PairType',
-           'MacroDefinition')
+           'MacroDefinition',
+           'MacroUndef')
 
 #------------------------------------------------------------------------------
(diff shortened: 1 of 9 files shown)


pyccel/ast/low_level_tools.py
```diff
@@ -127,3 +128,32 @@ def object(self):
...
+
+    def __init__(self, macro_name):
+        assert isinstance(macro_name, str)
+        self._macro_name = macro_name
+        super().__init__()
+
+    @property
+    def macro_name(self):
+        """
+        The name of the macro being defined.
```
jalalium, line 154: Shouldn't this be `The name of the macro to be undefined.`

jalalium (approved):
Well done!

pyccel/ast/low_level_tools.py
EmilyBourne: Yes absolutely. Copy paste error

EmilyBourne: Fixed : fd12a6f02

jalalium (approved):
Well done!

PR #2017, opened 2024-10-11
Add support for `in` operator

EmilyBourne:
Add support for `in` operator for `list`, `set` and `dict` containers for C Fortran and Python, as well as for classes.

**Commit Summary**
- Add `PyccelIn` class to describe the `in` operator
- Add C support for `in` operator for list/set/dict
- Add `#define i_use_cmp` to list declaration in C to activate sorting and searching functions
- Add `_define_gFTL_element` function in Fortran code printer to define types and operators for types inside container types
- Reduce duplication in `_build_gFTL_module`
- Add Fortran support for `in` operator for list/set/dict
- Add Python support for `in` operator for list/set/dict
- In semantic stage handle `PyccelIn` for built-in types and class `__contains__` magic method
- Add tests for `in` operator for lists/sets/dicts
- Add tests for `in` operator for classes
- Use `python_only_language` to allow some of the dictionary tests to be activated for other languages


jalalium (approved):
Well done!

PR #2019, opened 2024-10-15
Add support for set.copy

### LINKED ISSUE #1919: [SETS] Add C and Fortran support for `copy()`
### Test Code
```python
def foo():
    a = {1,2,1}
    b = a.copy()
```

### Proposed Solution
This feature should be implemented using the functionalities provided by STC and gFTL

EmilyBourne:
Add C and Fortran support for `set.copy`. Fixes #1919

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,6 @@ All notable changes to this project will be documented in this file.
 -   #1700 : Add Python support for list method `sort()`.
 -   #1696 : Add Python support for list method `copy()`.
 -   #1693 : Add Python support for list method `remove()`.
--   #1739 : Add Python support for set method `clear()`.
--   #1740 : Add Python support for set method `copy()`.
 -   #1750 : Add Python support for set method `remove()`.
 -   #1743 : Add Python support for set method `discard()`.
 -   #1754 : Add Python support for set method `update()`.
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -38,7 +36,8 @@ All notable changes to this project will be documented in this file.
 -   #1690 : Add C support for list method `pop()`.
 -   #1877 : Add C and Fortran Support for set method `pop()`.
 -   #1917 : Add C and Fortran support for set method `add()`.
--   #1918 : Add C and Fortran support for set method `clear()`.
+-   #1918 : Add for set method `clear()`.
+-   #1918 : Add for set method `copy()`.
```
jalalium, line 40: I think this is missing `C and Fortran support`

tests/epyccel/test_epyccel_sets.py
```diff
@@ -77,42 +77,42 @@ def clear_complex():
     python_result = clear_complex()
     assert python_result == pyccel_result
 
-def test_copy_int(python_only_language):
+def test_copy_int(language):
```
jalalium, line 80: Don't you think we should check that the original set is not affected by the `pops` for example? I know it will work now but reflecting their disassociation in the tests will prevent eventual bugs.

jalalium (changes requested):

CHANGELOG.md
EmilyBourne: I didn't want to have 3 lines for every method that we add. This CHANGELOG is already going to be quite big so if we can limit it I think it will make it more readable.

I wanted to just put "added support" if support was added for all the languages that we support.

Maybe this is confusing though next to methods which already had Python support where the languages therefore need to be mentioned explicitly.

What do you think?

tests/epyccel/test_epyccel_sets.py
EmilyBourne: That is a good idea :+1: I will add this to the tests

CHANGELOG.md
jalalium: I agree with you! I got your point when i was looking at the other PR for set union. I fancy `add support` better as well, but in this case it is also missing `support` , it's just `add for`

EmilyBourne: Ah :sweat_smile: oops

EmilyBourne: Fixed 47366167579c65883c3d87e9bf2648eb74da4bf7

jalalium (approved):
Well done!

PR #2022, opened 2024-10-17
Activate update tests

EmilyBourne:
All `update` tests seem to be working already as `update` is simply a multi-element version of `union`. This PR simply activates the existing tests and updates the documentation


tests/epyccel/test_epyccel_sets.py
```diff
@@ -178,14 +178,14 @@ def Discard_wrong_arg():
     python_result = Discard_wrong_arg()
     assert python_result == pyccel_result
 
-def test_update_basic(python_only_language):
+def test_update_basic(language):
```
jalalium, line 181: I think we should support `update` with more than one argument. Is this a choice that has been made?

EmilyBourne: Ah good point, I thought there was something I was supposed to do before marking this as ready. I just couldn't remember what!

EmilyBourne: Fixed : c98b12d

Test added : 5238c30

pyccel/parser/semantic.py
```diff
@@ -5323,36 +5323,42 @@ def _build_SetUpdate(self, expr, args):
             CodeBlock or For containing SetAdd objects.
         """
         if isinstance(expr, DottedName):
-            iterable = expr.name[1].args[0].value
+            iterable_args = [a.value for a in expr.name[1].args]
             set_obj = expr.name[0]
         elif isinstance(expr, AugAssign):
-            iterable = expr.rhs
+            iterable = [expr.rhs]
```
EmilyBourne, line 5329:
```suggestion
            iterable_args = [expr.rhs]
```

jalalium (approved):
Well done!

PR #2030, opened 2024-10-29
Allow iterating over a set

### LINKED ISSUE #2023: Allow iterating over a set
### Describe the feature
It should be possible to iterate over a set.

### Test Code
```python
a = {1,3,4}
sum_a = 0
for ai in a:
     sum_a += ai
```

### Proposed Solution
This should be implemented using `IteratorType` and `Iterable`. The code should be flexible enough to make it easy to do the same for dicts (dict_keys/dict_values) later.

In C the code should be:
```c
c_for_each(Dummy_0000, set_int, a)
{
    ai = *Dummy_0000.ref;
    sum_a += ai;
}
```
In Fortran the code should be:
```fortran
      iter = a % begin()
      last = a % end()
      do while (iter /= last)
         ai = iter%of()
         sum_a = sum_a + ai
         call iter%next()
      end do
```

EmilyBourne:
Add support for iterating over a set. Fixes #2023

**Commit Summary**
- Add class type information for `zip`
- Complete definition of `IteratorType` following documentation
- Add printing support for iterating over a set
- Add iterator tests for sets

### DIFF
--- a/pyccel/ast/low_level_tools.py
+++ b/pyccel/ast/low_level_tools.py
@@ -29,6 +29,7 @@ class IteratorType(PyccelType, metaclass=ArgumentSingleton):
         The container that is iterated over.
     """
     __slots__ = ('_iterable_type',)
+    rank = 0
     def __init__(self, iterable_type):
         self._iterable_type = iterable_type
         super().__init__()
--- a/pyccel/ast/low_level_tools.py
+++ b/pyccel/ast/low_level_tools.py
@@ -43,6 +43,9 @@ class IteratorType(PyccelType, metaclass=ArgumentSingleton):
         """
         return self._iterable_type
 
+    def __str__(self):
+        return f'Iter[{self._iterable_type}]'
+
 #------------------------------------------------------------------------------
 class PairType(PyccelType, metaclass=ArgumentSingleton):
     """
--- a/pyccel/ast/low_level_tools.py
+++ b/pyccel/ast/low_level_tools.py
@@ -29,7 +29,6 @@ class IteratorType(PyccelType, metaclass=ArgumentSingleton):
         The container that is iterated over.
     """
     __slots__ = ('_iterable_type',)
-    rank = 0
     def __init__(self, iterable_type):
         self._iterable_type = iterable_type
         super().__init__()
@@ -45,6 +44,36 @@ class IteratorType(PyccelType, metaclass=ArgumentSingleton):
 
     def __str__(self):
         return f'Iter[{self._iterable_type}]'
+
+    @property
+    def datatype(self):
+        """
+        The datatype of the object.
+
+        The datatype of the object.
+        """
+        return self
+
+    @property
+    def rank(self):
+        """
+        Number of dimensions of the object.
+
+        Number of dimensions of the object. If the object is a scalar then
+        this is equal to 0.
+        """
+        return 0
+
+    @property
+    def order(self):
+        """
+        The data layout ordering in memory.
+
+        Indicates whether the data is stored in row-major ('C') or column-major
+        ('F') format. This is only relevant if rank > 1. When it is not relevant
+        this function returns None.
+        """
+        return None
 
 #------------------------------------------------------------------------------
 class PairType(PyccelType, metaclass=ArgumentSingleton):
--- a/pyccel/codegen/printing/fcode.py
+++ b/pyccel/codegen/printing/fcode.py
@@ -1862,8 +1862,10 @@ class FCodePrinter(CodePrinter):
         rankstr   = ''
 
         # ... print datatype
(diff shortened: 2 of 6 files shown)


pyccel/ast/low_level_tools.py
```diff
@@ -42,6 +42,39 @@
...
+    def order(self):
+        """
+        The data layout ordering in memory.
+
+        Indicates whether the data is stored in row-major ('C') or column-major
+        ('F') format. This is only relevant if rank > 1. When it is not relevant
+        this function returns None.
+        """
+        return None
+
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
Order is included for completeness

pyccel/codegen/printing/fcode.py
```diff
@@ -2565,27 +2566,40 @@ def _print_FunctionalFor(self, expr):
...
-        if expr.iterable.num_loop_counters_required:
-            self.scope.insert_variable(index)
 
-        target   = index
-        my_range = expr.iterable.get_range()
+        if isinstance(iterable_type, (DictType, HomogeneousSetType)):
+            if isinstance(iterable.iterable, Variable):
+                suggested_name = iterable.iterable.name + '_'
+            else:
+                suggested_name = ''
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
This is expected. Ideally this fallback case should not be used

pyccel/ast/low_level_tools.py
```diff
@@ -42,6 +42,39 @@ def iterable_type(self):
...
+    @property
+    def rank(self):
+        """
+        Number of dimensions of the object.
+
+        Number of dimensions of the object. If the object is a scalar then
+        this is equal to 0.
+        """
+        return 0
```
jalalium, line 65: Is this always the case? What if we iterate over a list of lists?

EmilyBourne: The iterator isn't the object inside the container. It is just a container containing a pointer to that object. It still needs unpacking. The only thing you can do with an STC iterator is `elem = iter->ref`. This means that you can never do `iter[i]` so I don't think it needs to be considered as a multi-dimensional object

tests/epyccel/test_epyccel_sets.py
```diff
@@ -520,3 +520,33 @@ def union_int():
     pyccel_result = epyccel_func()
     python_result = union_int()
     assert python_result == pyccel_result
+
+def test_set_iter(language):
```
jalalium, line 524: This doesn't seem to work for  `C` and `Fortran`. There is a pytest param on line 488 that seems to restrict language to Python. I think handling iteration for all languages is within the scope of this PR.

jalalium (changes requested):

tests/epyccel/test_epyccel_sets.py
EmilyBourne: The param on line 488 is a parametrize decorator. It only applies to the function that it is decorating

EmilyBourne: Here is proof that this works for C and Fortran:
image

jalalium: I apologize profusely for the confusion.
I tested this and it did not work:
```python
if __name__ == '__main__':
    a = {1,2,3}
    ss = 0
    for x in a:
        ss += x
    print(ss)
```
that's why i thought that the tests did not run.
but then, why the following works but not the former:
```python
def f():
    a = {1,2,3}
    ss = 0
    for x in a:
        ss += x
    return ss

if __name__ == '__main__':
    print(f())
```

EmilyBourne: What error are you seeing? This code seems to work fine for me

jalalium: So sorry, I had to delete __pyccel__, something was cached and was showing this error:
```
pyccel:
 |error [codegen]: boo.py| Don't know how to index int type
Pyccel has encountered syntax that has not been implemented yet. Please create an issue at https://github.com/pyccel/pyccel/issues and provide a small example of your problem. Do not forget to specify your target language.
```

EmilyBourne: That is the error on the devel branch. Very strange if you're seeing it here. I would guess that you maybe didn't have pyccel installed in editable mode but if that's the case then I've no idea how you've been developing ^^

jalalium: I do have it installed in editable mode. The issue had to do with the `__pyccel__` folder. It does not seem related but deleting it fixed the issue.

jalalium (approved):
Well done! So sorry for the oversight!

PR #2032, opened 2024-11-01
Add missing dependency to installation description

EmilyBourne:
The cmake dependency is currently missing from the installation docs. This PR adds it.


jalalium (approved):
Well done!

PR #2041, opened 2024-11-06
Extend the C min and max functions to more than two variables

jalalium:
This PR aims to extend the C support for min and max functions to more than two variables.
It leverages the `stc/common.h` functionalities to define two variadic functions for min and max in `stdlib/STC_EXTENSIONS/Common_extensions.h`.
This PR also fixed the import scheme for extension headers:
Previously, stc headers and extension headers were included separately as follows:
```C
#ifndef _HSET_INT64_T
#define _HSET_INT64_T
#define i_type hset_int64_t
#define i_key int64_t
#include <stc/hset.h>
#endif // _HSET_INT64_T

#ifndef _TOOLS_SET_HSET_INT64_T
#define _TOOLS_SET_HSET_INT64_T
#define i_type hset_int64_t
#define i_key int64_t
#include <Set_extensions.h>
#endif // _TOOLS_SET_HSET_INT64_T
```
this is problematic because at the end of `stc` headers, we find a call to `#include "priv/template2.h"`
which undefines what was defined by the header unless `imore` is defined:
```C
#ifdef i_more
#undef i_more
#else
#undef i_TYPE
#undef i_type
#undef i_tag
#undef i_imp
#undef i_opt
#undef i_less
#undef i_cmp
#undef i_eq
#undef i_hash
#undef i_capacity
#undef i_raw_class
....
```
This prohibits the use of these in the extension header.

The proposed solution combines the `stc` header with its respective extension and defines `imore`:
```C
#ifndef _TOOLS_SET_HSET_INT64_T
#define _TOOLS_SET_HSET_INT64_T
#define i_type hset_int64_t
#define i_key int64_t
#define i_more
#include <stc/hset.h>
#include <Set_extensions.h>
#endif // _TOOLS_SET_HSET_INT64_T
```


jalalium:
bot checklist

EmilyBourne:
If we use `i_more` then do we need to call `#undef` ourselves to remove the macros polluting the namespace?

jalalium:
> If we use `i_more` then do we need to call `#undef` ourselves to remove the macros polluting the namespace?
The header `#include <stc/priv/template2.h>` that `#undef` was added to the extension headers. So it is taken care of.

pyccel/ast/core.py
```diff
@@ -3927,6 +3927,30 @@ def define_target(self, new_target):
...
+        target_to_remove : str | AsName | iterable[str | AsName]
+            The import target(s) to remove.
+        """
+        assert pyccel_stage != "syntactic"
+        if iterable(target_to_remove):
+            for t in target_to_remove:
+                self._target.pop(t, None)
+        else:
+            self._target.pop(target_to_remove, None)
+
```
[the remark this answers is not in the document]
jalalium: /bot accept
ad hoc use

pyccel/codegen/printing/ccode.py
```diff
@@ -1034,27 +1107,41 @@ def _print_Import(self, expr):
...
-                else:
-                    container_key = self.get_c_type(class_type.element_type)
-                    element_decl = f'#define i_key {container_key}\n'
-                if isinstance(class_type, HomogeneousListType) and isinstance(class_type.element_type, FixedSizeNumericType) \
-                        and not isinstance(class_type.element_type.primitive_type, PrimitiveComplexType):
-                    element_decl += '#define i_use_cmp\n'
-                header_guard_prefix = import_header_guard_prefix.get(source, '')
-                header_guard = f'{header_guard_prefix}_{container_type.upper()}'
-                code += ''.join((f'#ifndef {header_guard}\n',
+                if 'Common' in source:
```
EmilyBourne, line 1110: This seems like a very very general condition. Is it designed to catch multiple files? If not then why not use something more specific to avoid accidental errors?

pyccel/codegen/printing/ccode.py
```diff
@@ -1034,27 +1107,41 @@ def _print_Import(self, expr):
...
+                    element_decl = f'#define i_key {t.local_alias}\n'
+                    header_guard_prefix = import_header_guard_prefix.get(source, '')
+                    header_guard = f'{header_guard_prefix}_{t.local_alias.upper()}'
+                    code += ''.join((f'#ifndef {header_guard}\n',
                         f'#define {header_guard}\n',
-                        f'#define i_type {container_type}\n',
                         element_decl,
+                        '#define i_more\n' if source in import_header_guard_prefix else '',
+                        f'#include <{stc_header_mapping[source]}.h>\n' if source in import_header_guard_prefix else '',
```
EmilyBourne, line 1118: Are the conditions necessary for Common ?

pyccel/stdlib/STC_Extensions/Common_extensions.h
```diff
@@ -0,0 +1,31 @@
...
+        i_key value = va_arg(args, i_key);
+        if (value > max_value) {
+            max_value = value;
+        }
+    }
+    va_end(args);
+    return max_value;
+}
+
+#undef i_key
```
EmilyBourne, line 31: You have `#undef i_key` in `Common_extensions.h` and `List_extensions.h`. Will this work if you want to use `pull_elem` and `min` in the same file?

EmilyBourne (changes requested):

pyccel/codegen/printing/ccode.py
```diff
@@ -673,6 +707,32 @@ def init_stc_container(self, expr, assignment_var):
...
 
+    def invalidate_stc_headers(self, imports):
+        """
+        Invalidate STC headers when STC extension headers are present.
+
+        This function iterates over the list of imports and removes any targets
+        from STC headers if the target is present in their corresponding
+        STC extension headers are present.
+        The STC extension headers take care of including the standard
+        headers making them redundant.
```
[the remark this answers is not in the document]
jalalium: This docstring was ambiguous so I updated it. The idea is that, say i need something from the set extensions for int then simply by importing `Set_extensions` this will directly import the `set` header in STC thus there is no need to import the `set` header for int.

pyccel/codegen/printing/ccode.py
```diff
@@ -317,6 +323,34 @@ def __init__(self, filename, prefix_module = None):
...
+        """
+        import_src = [str(i.source) for i in imports]
+        stc_imports = [i for i in import_src if i.startswith('stc/')]
+        dependent_imports = [i for i in import_src if i in import_header_guard_prefix]
+        non_stc_imports = [i for i in import_src if i not in chain(stc_imports, dependent_imports)]
+        stc_imports.sort()
+        dependent_imports.sort()
+        non_stc_imports.sort()
+        sorted_imports = [imports[import_src.index(name)] for name in chain(non_stc_imports, stc_imports, dependent_imports)]
+        return sorted_imports
```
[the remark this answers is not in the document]
jalalium: Absolutely, fixed [31faf42a](https://github.com/pyccel/pyccel/pull/2041/commits/31faf42a5ab646eead2e54ebcdfa18f3e4a651e4)

pyccel/codegen/printing/ccode.py
```diff
@@ -874,6 +946,7 @@ def _print_Module(self, expr):
         for item in expr.imports:
             if item.source_module and item.source_module is not self._current_module:
                 self.rename_imported_methods(item.source_module.classes)
+        self.invalidate_stc_headers(expr.imports)
```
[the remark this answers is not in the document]
jalalium: You are right! I added here first before i figured out that i needed to add it to _print_ModuleHeader then i forgot to remove it

pyccel/codegen/printing/ccode.py
jalalium: You are right! It is not, I removed all superfluous code

pyccel/codegen/printing/ccode.py
jalalium: fixed

pyccel/stdlib/STC_Extensions/Common_extensions.h
jalalium: Yes, It works just fine. Normally, each header should undefine `i_key`.

pyccel/codegen/printing/ccode.py
```diff
@@ -855,6 +925,8 @@ def _print_ModuleHeader(self, expr):
 
         # Print imports last to be sure that all additional_imports have been collected
         imports = [*expr.module.imports, *self._additional_imports.values()]
+        imports = self.sort_imports(imports)
+        self.invalidate_stc_headers(imports)
```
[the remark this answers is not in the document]
jalalium: fixed

pyccel/codegen/printing/ccode.py
```diff
@@ -1034,27 +1079,41 @@ def _print_Import(self, expr):
...
-                    container_key = self.get_c_type(class_type.element_type)
-                    element_decl = f'#define i_key {container_key}\n'
-                if isinstance(class_type, HomogeneousListType) and isinstance(class_type.element_type, FixedSizeNumericType) \
-                        and not isinstance(class_type.element_type.primitive_type, PrimitiveComplexType):
-                    element_decl += '#define i_use_cmp\n'
-                header_guard_prefix = import_header_guard_prefix.get(source, '')
-                header_guard = f'{header_guard_prefix}_{container_type.upper()}'
-                code += ''.join((f'#ifndef {header_guard}\n',
+                if source == 'Common_extensions':
+                    print(source)
```
EmilyBourne, line 1083:
```suggestion
                if source == 'Common_extensions':
```

pyccel/stdlib/STC_Extensions/Common_extensions.h
jalalium: Yes there is no way to end up with this because including `<List_extensions.h>` automatically implies the following:
```C
#ifndef _TOOLS_LIST_VEC_INT64_T
#define _TOOLS_LIST_VEC_INT64_T
#define i_type vec_int64_t
#define i_key int64_t
#define i_use_cmp
#define i_more
#include <stc/vec.h>
#include <List_extensions.h>
#endif // _TOOLS_LIST_VEC_INT64_T
```
That's why including an extension header should invalidate the original header of the same `i_type`

EmilyBourne: Ok sounds good :+1:

pyccel/codegen/printing/ccode.py
```diff
@@ -708,13 +766,14 @@ def _print_PythonAbs(self, expr):
         else:
             func = "labs"
         return "{}({})".format(func, self._print(expr.arg))
-
-    def _print_PythonMin(self, expr):
+
```
EmilyBourne, line 769:
```suggestion

```

pyccel/codegen/printing/ccode.py
```diff
@@ -317,6 +323,32 @@ def __init__(self, filename, prefix_module = None):
...
+        Returns
+        -------
+        list[Import]
+            A sorted list of the imports.
+        """
+        import_src = [str(i.source) for i in imports]
+        stc_imports = [i for i in import_src if i.startswith('stc/') or i in import_header_guard_prefix]
+        non_stc_imports = [i for i in import_src if i not in stc_imports]
+        stc_imports.sort()
+        non_stc_imports.sort()
```
EmilyBourne, line 348: Are you sure about this sorting? What are you sorting by?

This seems dangerous to me. What happens if you have a list of list of a type beginning with Z?

```python
list[list[Z]]
```
`vector_Z` comes after `vector_vector_Z` but it needs to be printed before.

jalalium: I copied this method from your cspan PR, I didn't really go through it but now that mentioned it you are right it might be problematic

pyccel/codegen/printing/ccode.py
EmilyBourne: @jalalium This change doesn't seem to have gone through right. I wanted to remove your extra whitespace not the whole line :(

jalalium: I fixed it! b631e9f6

CHANGELOG.md
```diff
@@ -54,6 +54,7 @@ All notable changes to this project will be documented in this file.
 -   \[INTERNALS\] Add a `__call__` method to `FunctionDef` to create `FunctionCall` instances.
 -   \[INTERNALS\] Allow the use of magic methods to describe container methods.
 -   \[DEVELOPER\] Added an improved traceback to the developer-mode errors for errors in function calls.
+-   #1980 Extend The C support for min and max to more than two variables
```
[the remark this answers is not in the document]
jalalium: done!

CHANGELOG.md
```diff
@@ -49,6 +49,7 @@ All notable changes to this project will be documented in this file.
 -   #738 : Add support for homogeneous tuples with scalar elements as arguments.
 -   Add a warning about containers in lists.
 -   #2016 : Add support for translating arithmetic magic methods (methods cannot yet be used from Python).
+-   #1980 Extend The C support for min and max to more than two variables
```
EmilyBourne, line 52: To match existing conventions
```suggestion
-   #1980 : Extend The C support for min and max to more than two variables
```

pyccel/codegen/printing/ccode.py
jalalium: I think the issue you hinted at will be best handled elsewhere since it is about sorting import targets and not imports themselves. We can add a method to topologically sort the targets of imports based on their dependency graph.

pyccel/codegen/printing/ccode.py
```diff
@@ -14,8 +14,9 @@
 
 from pyccel.ast.bind_c    import BindCPointer
 
-from pyccel.ast.builtins  import PythonRange, PythonComplex
-from pyccel.ast.builtins  import PythonPrint, PythonType, VariableIterator
+from pyccel.ast.builtins  import PythonRange, PythonComplex, PythonMin, PythonMax
+from pyccel.ast.builtins  import PythonPrint, PythonType
+
```
EmilyBourne, line 19: Bad merge. `VariableIterator` is missing

pyccel/codegen/printing/ccode.py
jalalium: Good idea!

EmilyBourne: I just meant `.sort()` but if it's passing then removing it completely is better. I will have to reexamine #2064

pyccel/codegen/printing/ccode.py
```diff
@@ -1036,27 +1055,40 @@ def _print_Import(self, expr):
         if source.startswith('stc/') or source in import_header_guard_prefix:
             code = ''
```
EmilyBourne, line 1056: Does `Common_extensions` share any code with STC? If not then why not just do:
```suggestion
        if source == 'Common_extensions':
            ...
        elif source.startswith('stc/') or source in import_header_guard_prefix:
            code = ''
```
?

jalalium: You are right! It uses `common.h` but it does not need to specify anything specific to types.

PR #2040, opened 2024-11-06
Fix use of `_additional_code` in for loops

EmilyBourne:
Fix use of `CCodePrinter._additional_code` in for loops. Fixes #2039.

PR #2062, opened 2024-11-07
Update Python minimum versions

### LINKED ISSUE #2006: Python 3.8 End Of Life and Python 3.13
Python 3.8 reached end of life in October 2024 (there does not seem to be a more precise date but 5 years after the original release would put the date at October 14, 2024).

We should therefore update our CI tests to stop running with Python 3.8 by default.

Python 3.13 is also expected to be released today: https://peps.python.org/pep-0719/

This version will also need to be added to tests.

We usually try to ensure that we at least test:
- The lowest valid version
- The newest valid version

This issue will also apply to https://github.com/pyccel/pyccel-benchmarks

EmilyBourne:
Python 3.8 reached end of life in October. Support for Python 3.8 is now dropped from Pyccel. Fixes #2006 . In particular this will allow us to use:
- **container type annotations** (e.g. `list[int]` without needing strings and/or `typingext`)
- `str.removeprefix`
- `str.removesuffix`

### Commit Summary
- Update the Python versions for tests on `devel`/`main` to cover a variety of versions:
    - linux : 3.9-3.13
    - macosx : 3.13
    - windows : 3.9
    - intel : 3.11
    - anaconda_linux : 3.12
    - anaconda_windows : 3.10
- Update the minimum Python version for Pylint
- Update the CHANGELOG
- Update the tests run on the pull requests to cover a variety of versions:
    - linux : 3.9
    - macosx : 3.13
    - windows : 3.11
    - pyccel_lint : 3.11
    - pylint : 3.9
    - spelling : 3.12
    - docs : 3.10
- Delete unused file `ci_tools/devel_branch_tests.py`
- Remove use of `astunparse` package. From Python 3.9 `ast.unparse` can be used directly.
- Remove `pylint: disable` comments related to Python >=3.9 syntax on annotations
- Update `test_comments.py` test to use `ast.unparse` and `tests/parser/scripts/comments.py` to use basic syntax that matches the output of `ast.unparse`


EmilyBourne:
Blocked by #2040

pyccel/ast/numpytypes.py
```diff
@@ -470,4 +471,7 @@
 original_type_to_pyccel_type.update({v:k for k,v in numpy_type_to_original_type.items()})
 original_type_to_pyccel_type[np.bool_] = PythonNativeBool()
 
-NumpyInt = numpy_precision_map[PrimitiveIntegerType(), np.dtype(int).alignment]
+if Version(np.__version__) >= Version("2.0.0"):
+    NumpyInt = NumpyInt64Type()
+else:
+    NumpyInt = numpy_precision_map[PrimitiveIntegerType(), np.dtype(int).alignment]
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
NumPy v2 is tested by the MacOS CI

pyccel/errors/errors.py
```diff
@@ -325,10 +325,7 @@ def report(self,
         if symbol is not None:
             if isinstance(symbol, ast.AST):
                 ast_node = symbol
-                if sys.version_info < (3, 9):
-                    symbol = ast.dump(ast_node)
-                else:
-                    symbol = ast.unparse(ast_node) # pylint: disable=no-member
+                symbol = ast.unparse(ast_node) # pylint: disable=no-member
```
yguclu, line 328: We should not need the Pylint command `# pylint: disable=no-member` any longer, right?

pyproject.toml
```diff
@@ -20,12 +20,12 @@ dependencies = [
     "sympy >= 1.2",
     "termcolor >= 1.0.0",
     "textx >= 2.2",
-    "astunparse >= 1.6.0", # astunparse is only needed for Python3.8, we should use ast.unparse when we drop Python3.8.
     "packaging",
 ]
 
 [project.optional-dependencies]
 test = [
+    "astunparse >= 1.6.0",
```
yguclu, line 28: Why do we need `astunparse` for the tests?

EmilyBourne: It is used to test the comments here:
https://github.com/pyccel/pyccel/blob/b3ad4a99e6e4e6a05e0ba9c76bbcdabb5ead11f1/tests/parser/test_comments.py#L9-L14

pyccel/errors/errors.py
EmilyBourne: Ah I think you are right!

tests/epyccel/test_epyccel_variable_annotations.py
```diff
@@ -122,8 +122,7 @@ def final_annotation():
 
 def test_homogeneous_tuple_annotation(language):
     def homogeneous_tuple_annotation():
-        # Not valid in Python 3.8
-        a : tuple[int, ...] #pylint: disable=unsubscriptable-object
+        a : tuple[int, ...]
         a = (1,2,3)
```
yguclu, line 126: Does the inline annotation work, too?

tests/epyccel/test_epyccel_variable_annotations.py
```diff
@@ -221,7 +219,7 @@ def homogeneous_set_annotation ():
 
 def test_homogeneous_empty_list_annotation_int(language):
     def homogeneous_list_annotation():
-        a: 'list[int]'  # pylint: disable=unused-variable
+        a: 'list[int]'
```
yguclu, line 222: Can we not stop using the string?

EmilyBourne: Yes we can!

pyproject.toml
yguclu: I see. But maybe there is a similar class in the `ast` package?

tests/epyccel/test_epyccel_variable_annotations.py
EmilyBourne: Yes I think so. Odd that it is only tested for inhomogeneous tuples

pyccel/errors/errors.py
EmilyBourne: Fixed [c1f6db5](https://github.com/pyccel/pyccel/pull/2062/commits/c1f6db557c2b03ff0f38e83b95d988dfb02d4e2b)

tests/epyccel/test_epyccel_variable_annotations.py
EmilyBourne: I will take some out of strings but not all to ensure we keep testing strings. We cannot get rid of string annotations completely as there is no satisfactory replacement for NumPy arrays. Strings are also needed for class methods e.g.:
```python
class A:
    def __add__(self, other : 'A'):
        ...
```

pyproject.toml
EmilyBourne: I didn't see anything in the `ast` documentation

yguclu: There is nothing about `Unparser` in the `astunparse` documentation either 😅 ... We are using library internals here...

yguclu: Here is the source code:
- https://github.com/simonpercivall/astunparse/blob/2acce01fcdda2ea32eea835c30ccca21aaff7297/lib/astunparse/unparser.py#L27
- https://github.com/python/cpython/blob/b93b7e566e5a4efe7f077af2083140e50bd2b08f/Lib/ast.py#L666

tests/epyccel/test_epyccel_variable_annotations.py
EmilyBourne: Fixed 044f81a88

yguclu: Yes, we should test strings for the cases that are not covered otherwise!

Should we keep allowing for string annotations in the cases where we have a clean alternative like sets, lists, dicts and tuples?

EmilyBourne: I think we have to because otherwise you have a problem if you want e.g. `tuple[float[:], float[:,:]]`. It makes things easier for the stub files too as I don't have to work out whether or not the final type needs to be in quotes or not.

It doesn't add too much to the code as once the string annotation is parsed the string is discarded and we handle `list[int]` and `'list[int]'` identically.

pyproject.toml
EmilyBourne: I have just done a quick test. It seems this is possible but it is a lot more fiddly to do this test cleanly with `ast.unparse` instead of `astunparse` as `ast.unparse` is lower-level so it preserves less syntactic information. E.g I am currently getting the error:
```
AssertionError: assert 'b=(a+2)' == 'b=a+2'
```
If we are only using it to test comments though then maybe we don't care and I just need to remove unnecessary parentheses from the test file?

EmilyBourne: They seem fairly unnecessary so maybe they were only added to match what astunparse expected? Either that or it was to ensure readability when ` ` are stripped out

EmilyBourne: Similarly `ast.unparse` parses this:
https://github.com/pyccel/pyccel/blob/b3ad4a99e6e4e6a05e0ba9c76bbcdabb5ead11f1/tests/parser/scripts/comments.py#L60-L61
as:
```python
elif d + 1:
```

yguclu: Indeed there is no guarantee that the original source code is preserved, so we cannot make a string comparison. 

Maybe we can just check the commented lines, as this is what the unit test is about...

EmilyBourne: Fixed: f0ccfd5d2

But I am not sure that this test is the most robust (but it wasn't before either)

EmilyBourne: > Maybe we can just check the commented lines, as this is what the unit test is about...

The problem is that they are not so easy to identify after some of these simplifications as the line where they are found can change

yguclu: I think your solution is good enough for now!

yguclu (approved):
Everything looks good to me. Good job!

PR #2063, opened 2024-11-08
Stop limiting the length of strings to 128 characters in Fortran

### LINKED ISSUE #2013: Remove string length limitation
### Describe the feature
Currently the type of a local string is printed as `character(len=280)`, it should be possible to replace this with `character(len=:), allocatable`

EmilyBourne:
Correct the declaration type of strings in Fortran to remove the limit of 128 characters. Fixes #2013

### DIFF
--- a/tests/epyccel/test_epyccel_variable_annotations.py
+++ b/tests/epyccel/test_epyccel_variable_annotations.py
@@ -330,3 +330,19 @@ def test_dict_complex_float(language):
 
     epyc_dict_int_float = epyccel(dict_int_float, language = language)
     assert epyc_dict_int_float() == dict_int_float()
+
+@pytest.mark.parametrize( 'language', (
+        pytest.param("fortran", marks = pytest.mark.fortran),
+        pytest.param("c", marks = [
+            pytest.mark.skip(reason="C has no support for strings. See #2061"),
+            pytest.mark.c]),
+        pytest.param("python", marks = pytest.mark.python)
+    )
+)
+def test_str_declaration(language):
+    def str_declaration():
+        a : str = 'hello here is a very long string with more than 128 characters. This used to be a Fortran limit but now I can hold lots more characters. There is no limit!'
+        return len(a)
+
+    epyc_str_declaration = epyccel(str_declaration, language = language)
+    assert str_declaration() == epyc_str_declaration()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,7 +75,7 @@ All notable changes to this project will be documented in this file.
 -   #1972 : Simplified `printf` statement for Literal String.
 -   #2026 : Fix missing loop in slice assignment.
 -   #2008 : Ensure list/set/dict assignment is recognised as a reference.
--   #2013 : Strop limiting the length of strings to 128 characters.
+-   #2013 : Stop limiting the length of strings to 128 characters.
 
 ### Changed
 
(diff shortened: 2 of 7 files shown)


tests/epyccel/test_epyccel_variable_annotations.py
```diff
@@ -330,3 +330,19 @@ def dict_int_float():
...
+        pytest.param("fortran", marks = pytest.mark.fortran),
+        pytest.param("c", marks = [
+            pytest.mark.skip(reason="C has no support for strings. See #2061"),
+            pytest.mark.c]),
+        pytest.param("python", marks = pytest.mark.python)
+    )
+)
+def test_str_declaration(language):
+    def str_declaration():
+        a : str = 'hello here is a very long string with more than 128 characters. This used to be a Fortran limit but now I can hold lots more characters. There is no limit!'
```
jalalium, line 344: I don't know if this relevant but discrepancies arise because of `\0`. Python considers it as a regular character but Fortran doesn't. This raises issues when computing the length and printing the string.

EmilyBourne: Hmm, you mean that if you do:
```python
def str_len():
    a = '\0'
    print(len(a))
```
You will get different results in Python and Fortran?
Or if you do:
```python
def str_len():
    a = r'\0'
    print(len(a))
```
?

EmilyBourne: I don't know if this is related to this PR. I would be tempted to open a new PR for that. I don't think this PR is the cause of this issue

jalalium: I mean this
```python
def str_len():
    a = '\0\0\0'
    print(len(a))
```
Python will print 3 but Fortran will print 0.
As you said, it is not directly related to this PR however it needs addressing.

What about this?
```Python
if __name__ == '__main__':
    print(len('abc'))
 ```
 Is this supposed to be supported by this PR?

Comment (EmilyBourne) [reply tests/epyccel/test_epyccel_variable_annotations.py:344]: This case:
```python
def str_len():
    a = '\0\0\0'
    print(len(a))
```
looks like a missing case here:
https://github.com/pyccel/pyccel/blob/ace62bbca7e256e1f4c4d6bd50bca39514db7808/pyccel/codegen/printing/fcode.py#L2239

I think we should open a separate issue for this

jalalium: I will create an issue for that!
What about the support for the length of a string literal

EmilyBourne: What about this?
```python
if __name__ == '__main__':
    print(len('abc'))
```
Is this supposed to be supported by this PR?

The aim of this PR was to fix the length limit not to add support for `len`, but I had to support `len` to be able to test. I have just tested your example and it doesn't work. I agree that this should be fixed before merging this PR.

EmilyBourne: Fixed: [41bce1c](https://github.com/pyccel/pyccel/pull/2063/commits/41bce1cd09a3dcf6db8d30440c3c710df4f7d046)
Test added: [d2c7ac2](https://github.com/pyccel/pyccel/pull/2063/commits/d2c7ac232d2fedbb38dbe88e12bc61445ec8cd23)

EmilyBourne: #2075 issue created

PR #2064, opened 2024-11-08
Add C support for local string declarations

### LINKED ISSUE #2061: [STR] Add C support for local string declarations
### Describe the feature
Pyccel contains some minimal support for strings. Eventually we would like to expand on this support, however as a first step C support should be brought up to the level of Fortran support in order to allow strings to be used as keys for dictionaries.

As a first step support should be added for declaring a string locally

### Test Code
```python
def f():
    a = 'abcdef'
    b = '123456'
```

### Proposed Solution
The declaration of strings should be managed via `cstr` from STC. The associated documentation can be found here:
https://github.com/stclib/STC/blob/master/docs/cstr_api.md

The generated code should be:
```c
void f() {
    cstr a;
    a = cstr_lit("abcdef");
    cstr_drop(&a);
    a = cstr_lit("123456");
    cstr_drop(&a);
}
```
In the semantic stage, this will need to be handled with the `Allocate` node. In the codegen stage, this will need to be handled in `_print_Assign`, `_print_Allocate` and `_print_Deallocate`

EmilyBourne:
Add C support for local string declarations. This is a first step towards allowing string keys for dictionaries. Fixes #2061 

**Commit Summary**
- Handle strings in loop unravelling
- Allow extensions to depend on stc folders by using a recognisable dependency that can be swapped for the correct directory
- Fix declaration of strings in C
- Add support for `PyccelArrayShapeElement` of a string in C
- Remove double line break after deallocation
- Add support for string initialisation
- Correct the warning message for array reallocation (maybe this should be removed for non-arrays)
- Ensure strings are allocated in the semantic parser.
- Create a `cstr.c` file to compile string functions independently (avoids duplicate definitions)
- Activate tests
- Add a leak test for string reallocation


EmilyBourne:
Blocked by #2063

EmilyBourne:
Needs `sort_imports`. Blocked by #2041

jalalium (changes requested):
Great Job!
Printing, however, does not work. An easy fix is to use `cstr_data`

EmilyBourne:
> Great Job! Printing, however, does not work. An easy fix is to use `cstr_data`

Good point. I didn't think about anything beyond declaration at this point but `print` is probably the minimum that should be implemented along with the declarations in order for this to be vaguely useful

jalalium (approved):
Well done!

PR #2065, opened 2024-11-08
Fix missing allocations for lists in Fortran

### LINKED ISSUE #2036: Fortran translation does not allocate for list comprehension
### Describe the bug
The translated Fortran code does not allocate memory before attempting to populate the vector, leading to a segmentation fault when the program tries to access an unallocated memory region.

### To Reproduce
```python
if __name__ == "__main__":
    a = [i for i in range(10)]
    print(a[0])
```

### Error details

```fortran
program prog_prog_boo

  use boo
  use Vector_integer8_mod

  implicit none

  integer(i64) :: i
  type(Vector_integer8) :: a
  integer(i64) :: Dummy_0000

  Dummy_0000 = 0_i64
  do i = 0_i64, 9_i64
    a%of(Dummy_0000 + 1_i64) = i
    Dummy_0000 = Dummy_0000 + 1_i64
  end do
  write(stdout, '(I0)', advance="yes") a%of(1_i64)

end program prog_prog_boo
```

### Expected behavior
The vector should be allocated before it is populated.

### Language
Fortran

EmilyBourne:
Use the `Allocate.alloc_type` property to add missing allocations for lists in Fortran. Fixes #2036.

**Commit Summary**
- Activate leak checks for containers for Fortran
- Apply unravelling algorithms to lists
- Remove unnecessary `_print` call in `FCodePrinter._print_CodeBlock`
- Handle `resize` and `reserve` cases for list allocations in Fortran
- Simplify printing of list indexing
- Reactivate list comprehension tests

### DIFF
--- a/.github/actions/valgrind_run/action.yml
+++ b/.github/actions/valgrind_run/action.yml
@@ -14,5 +14,8 @@ runs:
       run: |
         pyccel --language=c --flags="-g -O0" leaks_check.py
         valgrind --leak-check=full --error-exitcode=1 ./leaks_check
+        pyccel-clean -p
+        pyccel --language=fortran --flags="-g -O0" leaks_check.py
+        valgrind --leak-check=full --error-exitcode=1 ./leaks_check
       shell: bash
       working-directory: ./tests/stc_containers
--- a/.github/actions/valgrind_run/action.yml
+++ b/.github/actions/valgrind_run/action.yml
@@ -3,14 +3,14 @@ name: 'Pyccel valgrind script'
 runs:
   using: "composite"
   steps:
-    - name: Test with valgrind for memory leaks
+    - name: Test with valgrind for memory leaks in ndarrays
       run: |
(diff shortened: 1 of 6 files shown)


.github/actions/valgrind_run/action.yml
```diff
@@ -14,5 +14,8 @@ runs:
       run: |
         pyccel --language=c --flags="-g -O0" leaks_check.py
         valgrind --leak-check=full --error-exitcode=1 ./leaks_check
+        pyccel-clean -p
```
jalalium, line 17: The name of this job indicates that it is for `stc containers`, maybe we can add a new job for `gFTL`.

EmilyBourne: Fixed : ded85d518

I prefer to keep similar tests grouped I think it could be clearer if we have a large number of leak files at some point.

But actually I've never worked out how to distinguish these steps inside an action in the output of a test so I don't think it's super critical

jalalium (approved):
Very good job!

PR #2069, opened 2024-11-11
Add support for returning sets from C

### LINKED ISSUE #1664: [SETS] Support sets as results in C
### Describe the feature
Support returning sets from C code

### Test Code
```python
def f():
    a = {1}
    return a
```

### Proposed Solution
The cwrapper should be improved to ensure that sets can be read from C code and passed to Python.

### Prerequisites
#1659

EmilyBourne:
Support sets as results when translating to C. Fixes #1664


EmilyBourne:
Blocked by #2068

PR #2071, opened 2024-11-11
Remove unused symbolic types

EmilyBourne:
Remove unused classes `SymbolicAssign`, `SymbolicPrint` and the method `Assign.is_symbolic_alias`. These objects come from handling of symbolic methods which existed in Pyccel many years ago but is either broken or handled differently now. The type `SymbolicType` makes these objects unnecessary.


jalalium (approved):
Well done!

PR #2073, opened 2024-11-20
Add support for `dict.items`

EmilyBourne:
Make `Iterable` into a super-class instead of a storage class. Functions that can be used in an iterable context must inherit from this class and define useful methods. The method `get_python_iterable_item` must be defined. If the iterable can be converted to a range the methods `get_range` and `get_assign_targets` must also be defined. Organising iterables via inheritance instead of storing the object in the `Iterable` class avoids problems with circular includes as the low-level `Iterable` object no longer depends on a large number of other ast modules.

The new setup is used to add support for `dict.items`.

**Commit Summary**
- Add an `Iterable` sub-class describing `dict.items`.
- Change method names in `Iterable` sub-classes
- Remove unnecessary methods `length` and `__getitem__` in `Iterable` sub-classes
- Add `Iterable` as a super class of objects that can be iterated over.
- Add a `prefer_inhomogeneous` argument to `PythonTuple` which allows the developer to force the creation of an inhomogeneous tuple (e.g. for the target when iterating over an enumerate of integers).
- Add a `VariableIterator` class to represent the `iter()` function that is called implicitly in a for loop
- Move `Iterable` class from `ast.core` to `ast.internals`
- Remove unnecessary line for `Iterable` in sympy_helper.
- Save the `For` target as a tuple after the semantic stage
- Fix insertion of an inhomogeneous tuple into a loop scope.
- Create `_get_for_iterators` function to unify handling of iterators in FunctionalFor/GeneratorExpressions/For loops
- Add tests


pyccel/ast/internals.py
```diff
@@ -342,6 +343,77 @@ def code(self):
...
+    Wrapper around iterable types helping to convert between those
+    types and a range (necessary in low level languages, e.g. C and Fortran).
+
+    If an iterable can be iterated over using a range then this can be done automatically
+    by defining the following 3 functions:
+    - get_range : Returns the range required for the iteration.
+    - get_python_iterable_item : Returns the item of the iterable that will be saved to the
+        variables which are the loop targets.
+    - get_assign_targets : Returns any objects that should be assigned to targets
+        (e.g. var[idx] for the iterable enumerate(var) with targets idx, var_idx).
```
jalalium, line 359: I think this example can be clearer

jalalium (changes requested):
Awesome Job, this refactoring is great!

pyccel/ast/internals.py
EmilyBourne: Better?
e0df50a20

jalalium: Awesome!

jalalium (approved):
Excellent job!

PR #2079, opened 2024-11-27
Fix translation of classes containing comments

### LINKED ISSUE #2078: Comments in classes do not translate
### Describe the bug
Comments in classes do not translate. They should be ignored

### To Reproduce
```python
class A:
        self._x = x

    #-----------------------------
    @property
    def x(self):
        return self._x
```

### Error details
```
pyccel:
 |error [syntactic]: tmp.py [5,4]| <class 'pyccel.ast.core.Comment'> not currently supported in classes (# -----------------------------)
```

### Expected behavior
The comment should be skipped over.

### Language
Syntactic stage problem affects all languages

### Additional context
Problem appears in PyGyro

EmilyBourne:
Fix translation of classes containing comments by discarding the comments. Fixes #2078

### DIFF
--- a/pyccel/parser/syntactic.py
+++ b/pyccel/parser/syntactic.py
@@ -1030,7 +1030,7 @@ class SyntaxParser(BasicParser):
             if isinstance(visited_i, FunctionDef):
                 methods.append(visited_i)
                 visited_i.arguments[0].bound_argument = True
-            elif isinstance(visited_i, Pass):
+            elif isinstance(visited_i, (Pass, Comment)):
                 continue
             elif isinstance(visited_i, AnnotatedPyccelSymbol):
                 attributes.append(visited_i)
--- a/pyccel/parser/syntactic.py
+++ b/pyccel/parser/syntactic.py
@@ -1032,6 +1032,8 @@ class SyntaxParser(BasicParser):
                 visited_i.arguments[0].bound_argument = True
             elif isinstance(visited_i, (Pass, Comment)):
                 continue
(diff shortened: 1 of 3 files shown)


pyccel/parser/syntactic.py
```diff
@@ -1030,7 +1030,7 @@ def _visit_ClassDef(self, stmt):
             if isinstance(visited_i, FunctionDef):
                 methods.append(visited_i)
                 visited_i.arguments[0].bound_argument = True
-            elif isinstance(visited_i, Pass):
+            elif isinstance(visited_i, (Pass, Comment)):
```
jalalium, line 1033: This does not handle cases where a `CodeBlock` of comments is created like in the following:
```Python
class A:
    def __init__(self):
        self.x = 5
    #bad
    #bad
    #bad
    #bad
if __name__ == '__main__':
    a = A()
    print(a.x)
```

This would fix the problem:
```suggestion
            elif isinstance(visited_i, (Pass, Comment)):
                continue
            elif isinstance(visited_i, CodeBlock) and all(isinstance(x, Comment) for x in visited_i.body):
```

jalalium (changes requested):

pyccel/parser/syntactic.py
EmilyBourne: Huh, I didn't realise that would be translated like that

EmilyBourne: Fixed : c5952b929
Test added : 71588d45f

jalalium (approved):
Well done!

PR #2083, opened 2024-11-27
Add support for multi operator expressions

### LINKED ISSUE #2081: Add support for multi operator expressions
### Describe the feature
Multi operator expressions can be useful. It would be nice to have support for this.

### Test Code
```python
if a <= x <= b:
    pass
```

### Proposed Solution
And and operator should be automatically inserted so `a <= x <= b` is understood as `a <= x and x <= b`

### Additional context
Needed by PyGyro

EmilyBourne:
Add support for multi operator expressions by extending the syntactic stage. Fixes #2081

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,15 +33,18 @@ All notable changes to this project will be documented in this file.
 -   #1874 : Add C and Fortran support for the `len()` function for the `list` container.
 -   #1875 : Add C and Fortran support for the `len()` function for the `set` container.
 -   #1908 : Add C and Fortran support for the `len()` function for the `dict` container.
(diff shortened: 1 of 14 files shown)


jalalium (commented):
I think it would be best to fix #2088 first to adequately test this PR. What do you think?

EmilyBourne:
> I think it would be best to fix #2088 first to adequately test this PR. What do you think?

I don't think it should make a difference. This PR combines comparison operators which always return booleans. So I don't think #2088 is relevant.

This PR is also one of the necessary steps for PyGyro so I would like to keep putting those through if there's no urgence for #2088

jalalium (approved):
Well done!

PR #2087, opened 2024-11-28
Fix calling class methods from constructors

### LINKED ISSUE #2085: Can't call a class method from the constructor
### Describe the bug
The visiting of functions was reorganised in #1728 so that functions are annotated when they are called. This allows functions to call functions whose definitions appear later in the file. However this work has not yet been done for class methods. This is problematic as it is common for the `__init__` method to be the first defined in the class. This means that no other methods can be called from the `__init__` method.

### To Reproduce
```python
class A:
        self._x = x
        self._y : float
        self._construct_y()

    def _construct_y(self):
        self._y = self._x + 3
```

### Error details
```bash/fortran/c
pyccel:
 |fatal [semantic]: tmp.py [3,4]| Can't find method private_construct_y in class A (<pyccel.ast.core.ClassDef object at 0x76848ad467a0>)
```

### Expected behavior
The methods should be handled in the same way as normal functions.

### Language
Semantic stage problem affects all languages

### Additional context
The case of `@property` will also need to be considered.
Needed for PyGyro

EmilyBourne:
Visit class methods in the order that they are called. Fixes #2085


EmilyBourne:
Testing blocked by #2089

jalalium (approved):
Well done!

PR #2089, opened 2024-11-28
Add support for @property

### LINKED ISSUE #1834: Add tests for `@property` decorator
### Describe the feature
The `@property` decorator is a very useful tool in classes. So much so that internally we have been using the property decorator without having added support for it in our translations. As a result it seems that this decorator now works out of the box! Currently a warning is raised claiming that we are ignoring the decorator:
```
pyccel:
 |warning [semantic]: tmp_cls.py| Decorator(s) not used (property)
```
However the code translates perfectly and clearly uses this property. We should therefore remove this warning, add tests and improve the documentation to show that this is supported.

### Test Code
The following code:
```python
class A:
        self._x = x

    @property
    def x(self):
        return self._x

if __name__ == '__main__':
    x = 2.0
    y = A(x)
    z = y.x
```
currently raises the above warning but generates the following code:
```fortran
module tmp_cls


  use, intrinsic :: ISO_C_Binding, only : i64 => C_INT64_T , b1 => &
        C_BOOL , f64 => C_DOUBLE
  implicit none

  type, public :: A
    real(f64) :: private_x

    contains
    procedure :: create => a_create
    procedure :: x => a_x
    procedure :: free => a_free
  end type A

  contains


  subroutine a_create(self, x)

    implicit none

    class(A), intent(inout) :: self
    real(f64), value :: x

    self%is_freed = .False._b1
    self%private_x = x

  end subroutine a_create

  function a_x(self) result(Out_0001)

    implicit none

    real(f64) :: Out_0001
    class(A), intent(inout) :: self

    Out_0001 = self%private_x
    return

  end function a_x

  subroutine a_free(self)

    implicit none

    class(A), intent(inout) :: self

    if (.not. self%is_freed) then
      ! pass
      self%is_freed = .True._b1
    end if

  end subroutine a_free

end module tmp_cls
```
```fortran
program prog_prog_tmp_cls

  use tmp_cls

  use, intrinsic :: ISO_C_Binding, only : i64 => C_INT64_T , b1 => &
        C_BOOL , f64 => C_DOUBLE
  implicit none

  real(f64) :: x
  type(A) :: y
  real(f64) :: z

  x = 2.0_f64
  call y % create(x)
  z = y % x()
  call y % free()

end program prog_prog_tmp_cls
```

Similarly for C:
```c
#include "tmp_cls.h"
#include <stdlib.h>
#include <stdbool.h>


void A__init__(struct A* self, double x)
{
    self->is_freed = 0;
    self->private_x = x;
}
double A__x(struct A* self)
{
    return self->private_x;
}
void A__del__(struct A* self)
{
    if (!self->is_freed)
    {
        // pass
        self->is_freed = 1;
    }
}
```
```c
#include "tmp_cls.h"
#include <stdlib.h>
#include <stdbool.h>
int main()
{
    double x;
    struct A y;
    double z;
    x = 2.0;
    A__init__(&y, x);
    z = A__x(&y);
    A__del__(&y);
    return 0;
}
```

EmilyBourne:
The semantic handling and the printing of methods using the `@property` decorator was already working, however official support for the decorator was missing due to the missing wrapper support. This PR adds the wrapping and removes the warning about this decorator. Fixes #1834 

**Commit Summary**
- Add documentation
- Prefer assertions for errors that only affect developers
- Allow the setter of a property to be missing (it is then saved as `None`)
- Add wrapping of a class property. The arguments of such functions are slightly different to those of a basic function.
- Add `-Wno-incompatible-pointer-types` to Python flags to remove unavoidable warnings
- Add tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,7 @@ All notable changes to this project will be documented in this file.
 -   #1980 : Extend The C support for min and max to more than two variables
 -   #2081 : Add support for multi operator expressions
 -   Add support for inhomogeneous tuple annotations.
+-   #1834 : Add support for `@property` decorator.
 -   \[INTERNALS\] Add abstract class `SetMethod` to handle calls to various set methods.
 -   \[INTERNALS\] Added `container_rank` property to `ast.datatypes.PyccelType` objects.
 -   \[INTERNALS\] Add a `__call__` method to `FunctionDef` to create `FunctionCall` instances.
--- a/docs/classes.md
+++ b/docs/classes.md
@@ -190,7 +190,7 @@ MyClass Object created!
 
 ## Class properties
 
-Pyccel now supports class properties (as getters only).
+Pyccel now supports class properties (to retrieve a constant value only).
 
 ### - Python Example
 
--- a/docs/classes.md
+++ b/docs/classes.md
@@ -188,6 +188,233 @@ MyClass Object created!
 ==158858== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
 ```
 
+## Class properties
+
+Pyccel now supports class properties (as getters only).
+
+### - Python Example
+
+```python
+import numpy as np
+
+class MyClass:
+    def __init__(self, param1 : 'int', param2 : 'int'):
+        self._param1 = param1
+        self._param2 = np.ones(param2)
+        print("MyClass Object created!")
+
+    @property
+    def param1(self):
+        return self._param1
+
+    @property
+    def param2(self):
+        return self._param2
+
+if __name__ == '__main__':
+    obj = MyClass1(2, 4)
+    print(obj.param1)
+    print(obj.param2)
+```
+
+### PYTHON _OUTPUT_
+```Shell
+MyClass Object created!
+2
+[1. 1. 1. 1.]
+```
+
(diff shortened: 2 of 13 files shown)


jalalium (commented):
Awesome job! only a small remark that is probably beyond the scope of this PR.

```Python

class A:
    def __init__(self, x : int , y : int):
        self._x = x
        self._y = y
    
    @property
    def sum(self):
        return self.x + self._y

    @property
    def x(self):
        return self._x
```

This works fine in python but gives this error in pyccel:
```bash

pyccel:
 |fatal [semantic]: boo.py| Can't find method x in class A (<pyccel.ast.core.ClassDef object at 0x7cff407cbd00>)
```
It works just fine with functions:
```python
def g():
    return f() + h()

def f():
    return 3

def h():
    return 2
```

EmilyBourne:
> Awesome job! only a small remark that is probably beyond the scope of this PR.

Yes. This is issue #2085 . It will be fixed and tested in #2087

EmilyBourne:
> It works just fine with functions:

But not with functions in classes:
```python
class A:
    def g(self):
        return self.f() + self.h()

    def f(self):
        return 3

    def h(self):
        return 2
```

jalalium (approved):
Well done!

PR #2092, opened 2024-11-29
Fix missing prefix in _print_Allocate in FCodePrinter

### LINKED ISSUE #2091: f-string and `call` missing prefix in `_print_Allocate` in `FCodePrinter`
### Describe the bug
The return statement in the code below is an f-string but it is missing the `f` prefix and the `call` statement.

```python
            elif expr.alloc_type == 'reserve':
                size_code = self._print(expr.shape[0])
                return '{var_code} % reserve({size_code})\n'
```

### To Reproduce
Can't easily reproduce now since we don't reserve much, but can be reproduced after #1977

## Error details

```bash

   24 |   {var_code} % reserve({size_code})
      |    1
Error: Invalid character in name at (1)

```

### Language
Fortran

### Additional context
Needed for #1977

jalalium:
Add missing call and make string into f-string to fix #2091

### DIFF
--- a/pyccel/codegen/printing/fcode.py
+++ b/pyccel/codegen/printing/fcode.py
@@ -2156,7 +2156,7 @@ class FCodePrinter(CodePrinter):
             elif expr.alloc_type == 'reserve':
                 var_code = self._print(expr.variable)
                 size_code = self._print(expr.shape[0])
-                return f'{var_code} % reserve({size_code})\n'
+                return f'call {var_code} % reserve({size_code})\n'
             else:
                 return ''
         elif isinstance(class_type, (HomogeneousContainerType, DictType)):
--- a/pyccel/codegen/printing/fcode.py
+++ b/pyccel/codegen/printing/fcode.py
@@ -2156,7 +2156,7 @@ class FCodePrinter(CodePrinter):
             elif expr.alloc_type == 'reserve':
                 var_code = self._print(expr.variable)
                 size_code = self._print(expr.shape[0])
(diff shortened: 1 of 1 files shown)


EmilyBourne (approved):
I wonder how this got past the coverage bot. Maybe we just said it was future work?

EmilyBourne:
Actually the coverage test seems to be passing. Do you know how this is possible?

jalalium:
> Actually the coverage test seems to be passing. Do you know how this is possible?

I think it is because  [here](https://github.com/pyccel/pyccel/pull/2065/files#diff-d67d33b36c4955f469a4e84486842966bdf7e6bde8aee3cbc1a18e7ea851b28b) the tests are ran even if they are marked as xfail. and since the problem is only at the compilation stage all the code up to compilation, and thus including the line in question, was ran and therefore was marked as covered. If this is the case, xfails may be hiding more code that is not truly covered

EmilyBourne:
> > Actually the coverage test seems to be passing. Do you know how this is possible?
> 
> I think it is because  [here](https://github.com/pyccel/pyccel/pull/2065/files#diff-d67d33b36c4955f469a4e84486842966bdf7e6bde8aee3cbc1a18e7ea851b28b) the tests are ran even if they are marked as xfail. and since the problem is only at the compilation stage all the code up to compilation, and thus including the line in question, was ran and therefore was marked as covered. If this is the case, xfails may be hiding more code that is not truly covered

Ah. I think you are right. This is why we should use skip for unsupported things.
In that case this PR seems good to me

PR #2100, opened 2024-12-05
Ignore __all__ statements

### LINKED ISSUE #2099: Ignore `__all__` statements
### Describe the feature
The `__all__` statement affects how wildcard imports are managed but is otherwise not useful. Pyccel should discard this object.

### Test Code
```python
__all__ = ('f',)

def f():
    pass
```

### Proposed Solution

Pyccel should simply discard this object unless printing to Python

EmilyBourne:
Fix translation of modules containing `__all__` by ignoring `__all__`. The statement is still printed in the Python translation. Fixes #2099


pyccel/parser/semantic.py
jalalium, line 1: I think it would be best to remove the following warning in this case:
```bash
 |warning [semantic]: boo.py| Nested lists are not yet fully supported. Using containers in lists may lead to bugs such as memory leaks (list[str])
```

jalalium (changes requested):
Maybe we can add a doc section about supported magic attributes

pyccel/parser/semantic.py
EmilyBourne: Fixed a24811ade, but I am hoping to remove this warning before v2.0 (it should go with #1959

jalalium (approved):
Well done!

PR #2101, opened 2024-12-08
Change the order of the arguments of `FunctionDef`

### LINKED ISSUE #337: Cannot return a tuple
This should be possible if the tuple is the only object returned as the result of a function can be stored in a tuple

EmilyBourne:
In order to fix #337 the handling of function results must be changed. Specifically, the handling of the case where nothing is returned is non-trivial. It is therefore better to handle this with a default value, however adding a default value requires the order of the arguments of the `FunctionDef` class to be changed. This PR makes that change in preparation


jalalium (approved):
Well done!

PR #2102, opened 2024-12-09
Allow returning sets and lists from Fortran

### LINKED ISSUE #1666: [LISTS] Support lists as results in Fortran
### Describe the feature
Support returning lists from functions in Fortran code

### Test Code
```python
def f():
    a = [1]
    return a
```

### Proposed Solution
The fortran_to_c_wrapper and the c_to_python_wrapper should be improved to ensure that lists can be read from Fortran code and passed through C code to Python.

### Prerequisites
#1657

### LINKED ISSUE #1667: [SETS] Support sets as results in Fortran
### Describe the feature
Support returning sets from functions in Fortran code

### Test Code
```python
def f():
    a = {1}
    return a
```

### Proposed Solution
The fortran_to_c_wrapper and the c_to_python_wrapper should be improved to ensure that sets can be read from Fortran code and passed through C code to Python.

### Prerequisites
#1658

EmilyBourne:
Add support to the Fortran to C wrapper to allow sets and lists to be returned from Fortran. This is done by packing the elements into an array. Fixes #1667 . Fixes #1666

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,11 +33,11 @@ All notable changes to this project will be documented in this file.
 -   #1874 : Add C and Fortran support for the `len()` function for the `list` container.
 -   #1875 : Add C and Fortran support for the `len()` function for the `set` container.
 -   #1908 : Add C and Fortran support for the `len()` function for the `dict` container.
--   #1665 : Add C support for returning lists from functions.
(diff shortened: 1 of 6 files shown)


jalalium (approved):
Well done!

PR #2107, opened 2024-12-14
Fix swap

### LINKED ISSUE #2104: Add support/error for Pythonic swap
### Describe the bug
Ideally add support for a Pythonic swap, if this proves too difficult to implement then an error should be raised to warn users that the translation does not match their expectations

### To Reproduce
```python
if __name__ == '__main__':
    a = 1
    b = 2
    a, b = b, a
    print(a, b)
```

### Error details
```c
int main()
{
    int64_t a;
    int64_t b;
    a = INT64_C(1);
    b = INT64_C(2);
    a = b;
    b = a;
    printf("%"PRId64" %"PRId64"\n", a, b);
    return 0;
}
```

### Expected behavior
Ideally the code should be:
```c
int main()
{
    int64_t a;
    int64_t b;
    int64_t Dummy_0000;
    a = INT64_C(1);
    b = INT64_C(2);
    Dummy_0000 = b;
    b = a;
    a = b;
    printf("%"PRId64" %"PRId64"\n", a, b);
    return 0;
}
```

### Language
All languages (including Python)

### Additional context
When fixing this we should bear in mind that a swap can contain more than 2 variables. The `get_attribute_nodes` function will be useful for this issue

EmilyBourne:
Add support for Pythonic swapping and raise an error for expressions which are too complex. Fixes #2104 

**Commit Summary**
- Add `is_equivalent` method to `pyccel.ast.base.PyccelAstNode.substitute` to allow more flexibility in how items are recognised
- Make `symbol` argument of `PyccelError` into a string to ensure it is printed before `substitute` is called if possible
- Add support for Pythonic swapping 
- Add tests


EmilyBourne:
@jalalium just to give you an update. I am having trouble with indexed elements. You mentioned:

> In addition, support for indexed elements seems fine in LHS and RHS but fails if in both ie:
```python
if __name__ == '__main__':
    a = [2,5]
    a[0],a[1] = a[1],a[0]
    print(a)
```

The method implemented for Variables cannot be extended to IndexedElements as an `IndexedElement` is not a singleton.

I am also worried about assumptions that may be implicit. E.g. in:
```
a[i], a[j] = a[j], a[i]
```
could/should I assume that `i != j`?

yguclu (commented):
Can the proposed code handle this?
```python
a : float
b : int
c : float
d : int
...
a, b, c, d = c, d, a, b
```

EmilyBourne:
> Can the proposed code handle this?
> 
> ```python
> a : float
> b : int
> c : float
> d : int
> ...
> a, b, c, d = c, d, a, b
> ```

Yes, this generates:
```fortran
  Dummy_0000 = a
  a = c
  Dummy_0001 = b
  b = d
  c = Dummy_0000
  d = Dummy_0001
```

yguclu:
> > Can the proposed code handle this?
> > ```python
> > a : float
> > b : int
> > c : float
> > d : int
> > ...
> > a, b, c, d = c, d, a, b
> > ```
> 
> Yes, this generates:
> 
> ```fortran
>   Dummy_0000 = a
>   a = c
>   Dummy_0001 = b
>   b = d
>   c = Dummy_0000
>   d = Dummy_0001
> ```

That's great, but I could not find a similar example among the unit tests.

EmilyBourne:
> > > Can the proposed code handle this?
> > > ```python
> > > a : float
> > > b : int
> > > c : float
> > > d : int
> > > ...
> > > a, b, c, d = c, d, a, b
> > > ```
> > 
> > 
> > Yes, this generates:
> > ```fortran
> >   Dummy_0000 = a
> >   a = c
> >   Dummy_0001 = b
> >   b = d
> >   c = Dummy_0000
> >   d = Dummy_0001
> > ```
> 
> That's great, but I could not find a similar example among the unit tests.

`test_swap_basic_3` ?

yguclu:
> `test_swap_basic_3`

I do not think so. The variables `a, b, c` are all of the same `int` type.

EmilyBourne:
> > `test_swap_basic_3`
> 
> I do not think so. The variables `a, b, c` are all of the same `int` type.

Test added : [c89af55](https://github.com/pyccel/pyccel/pull/2107/commits/c89af5503d69e6bcd5be384f73967ccf6448ec86)

yguclu (changes requested):
Looks good, but should you not check that the correct error is raised when the multiple assignment cannot be handled as a swap?

EmilyBourne:
> Looks good, but should you not check that the correct error is raised when the multiple assignment cannot be handled as a swap?

That would be nice but I don't have any examples. I wrote that as a fallback so the user has a clear error message if my assumptions about what I have parsed turn out to be wrong. I'm not sure what might trigger this. If I found one I would probably try to make it work rather than using it as an error test

yguclu:
> > Looks good, but should you not check that the correct error is raised when the multiple assignment cannot be handled as a swap?
> 
> That would be nice but I don't have any examples. I wrote that as a fallback so the user has a clear error message if my assumptions about what I have parsed turn out to be wrong. I'm not sure what might trigger this. If I found one I would probably try to make it work rather than using it as an error test

All right! I have no further objections then

yguclu (approved):

yguclu (dismissed):
Excellent!

jalalium (approved):
Excellent!

tests/epyccel/test_arrays_multiple_assignments.py
```diff
@@ -59,7 +59,7 @@ def f():
 
     # Check that the warning is correct
     warning_info = [*errors.error_info_map.values()][0][0]
-    assert warning_info.symbol  == 'x'
+    assert warning_info.symbol in ('x', "'x'")
```
yguclu, line 62: The symbol is a string which could be quoted?

EmilyBourne: When tests are run in developer-mode we use `repr` instead of `str` to get more information about the objects causing the errors. `str(PyccelSymbol('x'))` gives `'x'` but `repr(PyccelSymbol('x'))` gives `"'x'"`

yguclu (approved):
Looks good to me

PR #2108, opened 2024-12-14
Add Fortran support for list.pop

EmilyBourne:


pyccel/codegen/printing/ccode.py
```diff
@@ -1592,8 +1592,8 @@ def _print_IndexedElement(self, expr):
             list_var = self._print(ObjectAddress(base))
             container_type = self.get_c_type(base.class_type)
             if assign:
-                assert len(assign) == 1
-                assign_node = assign[0]
+                #assert len(assign) == 1
```
jalalium, line 1595: Why comment the assert?

pyccel/codegen/printing/fcode.py
```diff
@@ -1310,10 +1312,34 @@ def _print_PythonImag(self, expr):
     #========================== List Methods ===============================#
 
     def _print_ListAppend(self, expr):
-        target = expr.list_obj
+        target = self._print(expr.list_obj)
         arg = self._print(expr.args[0])
         return f'call {target} % push_back({arg})\n'
 
+    def _print_ListPop(self, expr):
```
jalalium, line 1319: Could you please also support `listpop` as a function argument? like with `print`.
It fails in `Fortran`.

jalalium (changes requested):

pyccel/codegen/printing/fcode.py
EmilyBourne: Fixed : 67d9555e3bd1108dd000ff7c03ae2296af6f82d7
Test added : 71ee66ba68ee8e33311253f641c0bf9f7337a061

EmilyBourne: Fixed : 67d9555e3bd1108dd000ff7c03ae2296af6f82d7
Test added : 71ee66ba68ee8e33311253f641c0bf9f7337a061

pyccel/codegen/printing/ccode.py
EmilyBourne: Hmm, I don't remember. This probably means I had multiple assign nodes. That's usually indicative of a problem with the AST tree. I meant to go back and hunt it down once everything else was working. I have reverted the change: e514f3c3fab9a5722db9f64bccd3fc48a78f0e34
If the tests still pass then it's ok. Otherwise I will have to hunt this down

EmilyBourne: It seems to be fine now

CHANGELOG.md
```diff
@@ -36,7 +36,7 @@ All notable changes to this project will be documented in this file.
 -   #1665 : Add C and Fortran support for returning lists from functions.
 -   #1689 : Add C and Fortran support for list method `append()`.
 -   #1876 : Add C support for indexing lists.
--   #1690 : Add C support for list method `pop()`.
+-   #1690 : Add C and Fortran support for list method `pop()`.
```
jalalium, line 39: Didn't we agree to use `Add support` when a functionality is fully supported?

pyccel/codegen/printing/fcode.py
```diff
@@ -1310,10 +1312,42 @@ def _print_PythonImag(self, expr):
...
+        list_obj = expr.list_obj
+        target = self._print(list_obj)
+        index_element = expr.index_element
+        if isinstance(expr.current_user_node, Assign):
```
jalalium, line 1323: the following works in C:
```python
if __name__ == '__main__':
    a = [1,2,3]
    print(a.pop(a.pop(0)))
```
Do you think you can support it here as well?

jalalium (changes requested):

CHANGELOG.md
EmilyBourne: Not quite, we agreed to use `Add support` when support for all languages was added. In this case Python support was already available in the previous release.

pyccel/codegen/printing/fcode.py
EmilyBourne: Test added : a3d542c66
Fix pushed : 742ab8115

PR #2113, opened 2024-12-16
Fix type hinting bug for renamed class attributes

### LINKED ISSUE #2111: Using type hinting with class attribute named `data` causes bug in `Fortran`
### Describe the bug
Using type hinting with a class attribute named data causes a double declare in Fortran.

### To Reproduce
```python
class A:
    def __init__(self, x : int):
        self.data :'int'= x

if __name__ == "__main__":
    a = A(4)
    print(a.data)
```

### Error details
```fortran
 type, public :: A
    integer(i64) :: data_0001
    integer(i64) :: data_0001

    contains
    procedure :: create => a_create
    procedure :: free => a_free
  end type A
```

### Expected behavior
Not sure why this only happens with `data`

### Language
Fortran

EmilyBourne:
Fix declaration of class attributes with name conflicts using type annotations. Fixes #2111 .
 The attribute created via the type annotation was not passing the original Python name to the scope which lead to it not being found later. As a result a second (unnecessary but identical) variable was created.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -93,6 +93,7 @@ All notable changes to this project will be documented in this file.
 -   #2082 : Allow the use of a list comprehension to initialise an array.
 -   #2094 : Fix slicing of array allocated in an if block.
 -   #2085 : Fix calling class methods before they are defined.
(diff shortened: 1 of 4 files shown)


jalalium (approved):
Well done!

PR #2116, opened 2024-12-17
Fix integer handling with NumPy 2.0 on Windows

### LINKED ISSUE #2115: NumPy 2.0 changed the default integer for Windows
### Describe the bug

NumPy version 2.0 changed the type of integers used on Windows:

image

https://numpy.org/devdocs/numpy_2_0_migration_guide.html#windows-default-integer

Although our handling of integer types is fairly flexible it does not handle this change correctly.

### To Reproduce
Run CI on Pythoon > 3.10

EmilyBourne:
Fix integer handling with NumPy 2.0 on Windows. The default integer is correctly deduced however the handling in the wrapper was incorrect. This was fixed (fixes #2115). Additionally the dtype error output is cleaned up so the positions of the datatypes in NumPy's enum are not hard coded on our side.

This fix was complicated by the fact that `np.random.randint`'s default dtype is maintained at `np.long` and therefore does not match the default used by all other NumPy functions in v2:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,12 +33,12 @@ All notable changes to this project will be documented in this file.
 -   #1874 : Add C and Fortran support for the `len()` function for the `list` container.
 -   #1875 : Add C and Fortran support for the `len()` function for the `set` container.
 -   #1908 : Add C and Fortran support for the `len()` function for the `dict` container.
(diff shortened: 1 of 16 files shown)


jalalium (approved):
Well done!

PR #2117, opened 2024-12-18
Add C and Fortran support for set.intersection and set.intersection_update

### LINKED ISSUE #1745: intersection_update : Updates the set with the intersection of itself and other sets.
### Feature Description

**Add support for intersection_update() method for sets in the Python translation.**
`intersection_update(*others)` :  Updates the calling set to contain only the elements that are present in both the calling set and the specified other set(s).

### Test Code

This Code Should Work 

```python
set1 = {1, 2, 3}
set2 = {2, 3, 4}
set1.intersection_update(set2)
print(set1)
>>> {2, 3}
```

### Proposed Solution
See description in #1531

EmilyBourne:
Add C and Fortran support for set methods `intersection` and `intersection_update`. Fixes #1745 
Map `set.intersection` to `set.copy` and `set.intersection_update`. Activate existing tests and add tests for `intersection_update`. Remove tests for `intersection` taking lists or tuples as arguments. This can be added back later if it is requested

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ All notable changes to this project will be documented in this file.
 -   #1753 : Add support for set method `union()`.
 -   #1754 : Add support for set method `update()`.
 -   #1744 : Add support for set method `intersection()`.
+-   #1745 : Add support for set method `intersection_update()`.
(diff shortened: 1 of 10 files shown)


jalalium (approved):
Well done!

PR #2118, opened 2024-12-18
Add C and Fortran support for remove/discard

### LINKED ISSUE #1921: [SETS] Add C support for `Discard`
### Test Code
```python
def foo():
    a = {1, 2, 3}
    a.discard(2)
```

### Proposed Solution

EmilyBourne:
Add C and Fortran support for set methods `remove`/`discard`. Fixes #1921

Blocked by #2102


jalalium (commented):
Don't you think we should mention that we do not distinguish between `discard` and `remove` since our `remove` does not raise.

EmilyBourne:
> Don't you think we should mention that we do not distinguish between `discard` and `remove` since our `remove` does not raise.

I'm not sure that we mention this fact anywhere in our docs currently. But Pyccel doesn't do error handling at all. If your code doesn't run without errors in Python then it can crash when you run it with Pyccel. I don't think this is specific to this PR at all, but if it is not mentioned anywhere yet then it should definitely be added to the docs. Maybe this should have a dedicated issue?

EmilyBourne:
@jalalium Issue created: https://github.com/pyccel/pyccel/issues/2133

jalalium:
> @jalalium Issue created: #2133

Awesome!

jalalium (approved):
Well done!

PR #2128, opened 2025-01-04
Add wrapping support for arithmetic magic methods

### LINKED ISSUE #2024: Add wrapping for class magic methods
### Describe the feature
Some support for class magic methods was added in #2016. The current implementation allows the use of operators to call magic methods in the translation but does not allow the user to call these magic methods from Python. The wrapper should be improved to make this possible.

### Test Code
```python
class A:
    def __init__(self, x : int):
        self.x = x

    def __add__(self, other : int):
        return A(self.x+other)

    def __mul__(self, other : int):
        return A(self.x*other)

    def __radd__(self, other : int):
        return A(self.x+other)

    def __iadd__(self, other : int):
        self.x += other
        return self

if __name__ == '__main__':
    my_a = A(4)
    left_add = my_a + 5
    right_add = 7 + my_a
    left_mul = my_a * 2

    print(my_a.x)
    print(left_add.x)
    print(right_add.x)
    print(left_mul.x)

    my_a += 6

    print(my_a.x)

    my_a *= 3

    print(my_a.x)
```
It should be possible to translate the module, then import it and run the main from Python.

### Proposed Solution
This should be implemented as described in the Python docs:
https://docs.python.org/3/c-api/typeobj.html

The functions describing mathematical methods need to be changed to binary functions:
- https://docs.python.org/3/c-api/typeobj.html#c.binaryfunc

They should then be packed into a `PyNumberMethods` instance:
- https://docs.python.org/3/c-api/typeobj.html#c.PyNumberMethods

Finally this instance should be saved in the `PyTypeObject` description:
- https://docs.python.org/3/c-api/typeobj.html#pytypeobject-definition

EmilyBourne:
Add wrapping support for arithmetic magic methods to allow them to be called from Python. Fixes #2024

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,7 +54,7 @@ All notable changes to this project will be documented in this file.
 -   #1583 : Allow inhomogeneous tuples in classes.
 -   #738 : Add support for homogeneous tuples with scalar elements as arguments.
 -   Add a warning about containers in lists.
--   #2016 : Add support for translating arithmetic magic methods (methods cannot yet be used from Python).
+-   #2016 : Add support for translating arithmetic magic methods.
 -   #1980 : Extend The C support for min and max to more than two variables
 -   #2081 : Add support for multi operator expressions
 -   #2061 : Add C support for string declarations.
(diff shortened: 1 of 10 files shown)


jalalium (approved):
Awesome!

PR #2131, opened 2025-01-04
Allow returning strings from functions

EmilyBourne:
Add wrapper support for returning strings from functions. Fortran strings are copied into a C-compatible object which is terminated with the `C_NULL_CHAR`. Both strings are packed into a Python object using `PyBuildValue`

The function `cstr_data` is added to `c_concepts`. In order to avoid circular imports the type check in `pyccel.ast.core.Allocate` is therefore restricted to the semantic stage.

Existing string tests are reactivated.


jalalium (commented):
Well done!
Only a small remark:
This works:
```Python
def one_quote():
    s = 'hello world'
    return s
```
but this does not work in `C`:
```Python
def one_quote():
    return 'hello world'
```
Do you think it is within the scope of this PR to create a temp when a literal string is returned?

jalalium (changes requested):
```Python
s = ''
```
Does not translate well in `Fortran`. Maybe a separate issue?

EmilyBourne:
> Well done! Only a small remark: This works:
> 
> ```python
> def one_quote():
>     s = 'hello world'
>     return s
> ```
> 
> but this does not work in `C`:
> 
> ```python
> def one_quote():
>     return 'hello world'
> ```
> 
> Do you think it is within the scope of this PR to create a temp when a literal string is returned?

Both of these now work. I do not create a temporary. Instead literals are now always printed with `cstr_lit` to ensure they have the correct type. `CStrStr` can be used to recover a bare string.

> ```
> s = ''
> ``` 
> Does not translate well in Fortran. Maybe a separate issue?

See #2198

jalalium (approved):
Well done!

PR #2136, opened 2025-01-09
Use older cmake command line syntax for compatibility

### LINKED ISSUE #2135: CMake command requires 3.21
### Describe the bug
The CMake command currently used in installation requires v3.21. But the CMake in the nvidia docker image is CMake v3.16.3. This is causing problems in https://github.com/pyccel/pyccel-cuda . It can easily be fixed by using `-DCMAKE_INSTALL_PREFIX=` instead of `--install-prefix`

EmilyBourne:
Use older cmake command line syntax for compatibility with older CMake versions (up to 3.13). This allows tests to run in https://github.com/pyccel/pyccel-cuda. Fixes #2135


jalalium (approved):
Well done!

PR #2140, opened 2025-01-09
Fix removal of `test_node` when `rhs.current_user_node` becomes `PyccelAssociativeParenthesis`

jalalium:
This PR fixes an issue where attempting to remove `test_node` from `rhs` fails when `rhs.current_user_node` becomes a `PyccelAssociativeParenthesis` during the `operator(lhs, rhs)` operation. This happens when `rhs` contains more than one element, the `operator(lhs, rhs)` call creates a `PyccelAssociativeParenthesis`, which becomes the `current_user_node` of `rhs`. As a result, a direct call to `rhs.remove_user_node(test_node)` fails since `test_node` no longer matches the `current_user_node`.


EmilyBourne (approved):
Looks good

PR #2143, opened 2025-01-11
Add C support for `dict.pop`

### LINKED ISSUE #2052: [DICT] Add C support for pop
### Describe the feature
Add C support for the dict method `pop`.

### Test Code
```python
def foo():
    a = {1:1.0, 2:2.0}
    b = a.pop(1)
    c = a.pop(3, 4.0)
    return len(a), b, c
```

### Proposed Solution
This should be implemented using the hmap_X_erase method documented here:

emmanuel-ferdman:
Add C support for dict method `pop()`. Fixes #2052

**Commit Summary**
- Implement C support for the `dict.pop` method, including support for default value.
- Add `Dict_extensions.h` to `STC_Extensions` for handling additional dictionary methods in C.
- Add `hmap_X_pop` method in `Dict_extensions.h` for retrieving and removing a key-value pair.
- Add `hmap_X_pop_with_default` method in `Dict_extensions.h` for retrieving and removing a key-value pair with default value.
- Activate `dict.pop` tests for C (ignoring Fortran for now).


emmanuel-ferdman:
To ensure clarity and efficient code review, here is a detailed explanation with examples and notes:

The implementation uses the [STC library methods](https://github.com/stclib/STC/blob/master/docs/hmap_api.md). The dictionary extensions introduces two new methods:
- `hmap_int64_t_int64_t_pop` method - Uses STC's `hmap_X_at` to attempt accessing the value associated with a given key. If the key exists, the value is retrieved, the key-value pair is removed using `hmap_X_erase`, and the value is returned. If the key is not found, the method fails, mirroring the behavior of Python's `dict.pop` when no default value is provided.
- `hmap_int64_t_int64_t_pop_with_default` method - Similar to `hmap_int64_t_int64_t_pop`, but it first checks if the key exists in the dictionary using `hmap_X_contains`. If the key is not found, the method returns the provided default value. If the key is found, the value is retrieved and removed as before (this is done by simply calling `hmap_int64_t_int64_t_pop`.

Please note that I implemented two methods because using a single method would require passing `NULL` (or other value) as the third argument for the "without default value" case. However, this could cause ambiguity since `None` (or `NULL`) could also be explicitly passed as a default value and C does not support default arguments like Python does.

Here are some examples:

Program 1: Basic `dict.pop` with a single key that exists in the dictionary. Expected return value: `6`.
Python Code:
```python
def dict_test1():
        a={1:5,2:6}
        b = a.pop(2)
        return b
```
Generated C Code:
```c
int64_t dict_test1(void)
{
    hmap_int64_t_int64_t a = {0};
    int64_t Dummy_0000;
    int64_t b;
    a = c_init(hmap_int64_t_int64_t, {{INT64_C(1), INT64_C(5)}, {INT64_C(2), INT64_C(6)}});
    Dummy_0000 = hmap_int64_t_int64_t_pop(&a, INT64_C(2));
    b = Dummy_0000;
    hmap_int64_t_int64_t_drop(&a);
    return b;
}
```

Program 2: `dict.pop` with multiple keys that exist in the dictionary. Expected return value: `11`.
```python
def dict_test2():
        a={3:5,2:6}
        b = a.pop(2)
        c = a.pop(3)
        return b + c
```
Generated C Code:
```c
int64_t dict_test2(void)
{
    hmap_int64_t_int64_t a = {0};
    int64_t b;
    int64_t c;
    int64_t Dummy_0000;
    int64_t Dummy_0001;
    a = c_init(hmap_int64_t_int64_t, {{INT64_C(3), INT64_C(5)}, {INT64_C(2), INT64_C(6)}});
    Dummy_0000 = hmap_int64_t_int64_t_pop(&a, INT64_C(2));
    b = Dummy_0000;
    Dummy_0001 = hmap_int64_t_int64_t_pop(&a, INT64_C(3));
    c = Dummy_0001;
    hmap_int64_t_int64_t_drop(&a);
    return b + c;
}
```

Program 3: `dict.pop` with a default value. Expected return value: `5`.
```python
def dict_test3():
        a={3:5,2:6}
        b = a.pop(100,5)
        return b
```
Generated C Code:
```c
int64_t dict_test3(void)
{
    hmap_int64_t_int64_t a = {0};
    int64_t Dummy_0000;
    int64_t b;
    a = c_init(hmap_int64_t_int64_t, {{INT64_C(3), INT64_C(5)}, {INT64_C(2), INT64_C(6)}});
    Dummy_0000 = hmap_int64_t_int64_t_pop_with_default(&a, INT64_C(100), INT64_C(5));
    b = Dummy_0000;
    hmap_int64_t_int64_t_drop(&a);
    return b;
}
```

One final note - it appears that C dictionaries in the current implementation do not yet support strings as valid key types (for example `{"x": 1, "y": 2}`). Due to this limitation, I have not activated the `test_pop_str_keys[c]` test for now. Once string support for dictionary keys is added, this test can be revisited and activated.

pyccel/codegen/printing/ccode.py
```diff
@@ -1350,6 +1356,9 @@ def get_c_type(self, dtype):
             val_type = self.get_c_type(dtype.value_type).replace(' ', '_')
             i_type = f'{container_type}_{key_type}_{val_type}'
             self.add_import(Import(f'stc/{container_type}', AsName(VariableTypeAnnotation(dtype), i_type)))
+            self.add_import(Import(f'{stc_extension_mapping["stc/" + container_type]}',
+                                   AsName(VariableTypeAnnotation(dtype), i_type),
+                                   ignore_at_print=True))
```
EmilyBourne, line 1361: @jalalium didn't you add something so that both of these includes are imported automatically?
I don't like the idea that we have to do this manually. It seems like a good way to regularly forget the extensions import (as seems to have happened here)

pyccel/stdlib/STC_Extensions/Dict_extensions.h
```diff
@@ -0,0 +1,35 @@
+#ifndef DICT_EXTENSIONS_H
+#define DICT_EXTENSIONS_H
```
EmilyBourne, line 2: This file shouldn't have a header guard. A different version of this file needs to be generated for each type of dictionary. The header guard prevents this. You should be able to see this with the following example:

```python
def f(k : int):
    a = {1:1.0, 2:2.0}
    b = {1:True, 2: False}
    if b.pop(k):
        return a.pop(k)
    else:
        return 3.0
```

tests/epyccel/test_epyccel_dicts.py
```diff
@@ -70,21 +70,31 @@ def kwarg_init():
     assert isinstance(python_result, type(pyccel_result))
     assert python_result == pyccel_result
 
-def test_pop_element(python_only_language):
+@pytest.mark.parametrize("language", (
+        pytest.param("c", marks = pytest.mark.c),
+        pytest.param("python", marks = pytest.mark.python)
+    )
+)
+def test_pop_element(language):
```
EmilyBourne, line 78: There is a shortcut for this if you want:
```suggestion
def test_pop_element(stc_language):
```

This can help us find methods that we still have to translate to Fortran

EmilyBourne (changes requested):
Thank you for another contribution @emmanuel-ferdman :smile:  This looks good to me. I just have a couple of comments.

emmanuel-ferdman:
Thank you for the code review :smile:

Commit contains the following changes:
- Removed the header guard in `Dict_extensions.h`.
- Converted tests to use `stc_language`.
- Add one additional test `test_pop_bool_keys`.
- Removed `imports` from `_print_DictPop` as they look they seem unnecessary.

Regarding your interesting example:
```python
def f(k : int):
    a = {1:1.0, 2:2.0}
    b = {1:True, 2: False}
    if b.pop(k):
        return a.pop(k)
    else:
        return 3.0
```
It generates the following C function:
```c
double f(int64_t k)
{
    hmap_int64_t_double a = {0};
    hmap_int64_t_bool b = {0};
    double Dummy_0000;
    bool Dummy_0001;
    double Out_0001;
    a = c_init(hmap_int64_t_double, {{INT64_C(1), 5.0}, {INT64_C(2), 6.0}});
    b = c_init(hmap_int64_t_bool, {{INT64_C(1), 1}, {INT64_C(2), 0}});
    if (Dummy_0001)
    {
        Dummy_0000 = hmap_int64_t_double_pop(&a, k);
        Out_0001 = Dummy_0000;
        hmap_int64_t_double_drop(&a);
        hmap_int64_t_bool_drop(&b);
        return Out_0001;
    }
    else
    {
        Dummy_0001 = hmap_int64_t_bool_pop(&b, k); // <-- PROBLEM
        hmap_int64_t_double_drop(&a);
        hmap_int64_t_bool_drop(&b);
        return 3.0;
    }
    hmap_int64_t_double_drop(&a);
    hmap_int64_t_bool_drop(&b);
}
```
Beside the 'header guard' issue that I fixed, I think there is another issue here - `hmap_int64_t_bool_pop(&b, k)` is evaluated inside the `else` block instead of being computed before the conditional statement. This results in incorrect behavior, as the conditional logic depends on the value of `b.pop(k)` but evaluates it at the wrong point.

This issue is not specific to the `dict.pop` case. It also occurs in other cases, such as:
```python
def test_min():
    if min(5,6):
        return 1
    else:
        return 0
```
Which generates the following C code:
```c
int64_t test_min(void)
{
    int64_t Dummy_0000;
    int64_t Dummy_0001;
    if (((Dummy_0000 < Dummy_0001 ? Dummy_0000 : Dummy_0001) != 0))
    {
        return INT64_C(1);
    }
    else
    {
        Dummy_0000 = INT64_C(5); // <-- PROBLEM
        Dummy_0001 = INT64_C(6);
        return INT64_C(0);
    }
}
```
Here, the variables `Dummy_0000` and `Dummy_0001` used in the `min` function are only evaluated in the `else` block, leading to incorrect behavior because the condition depends on their values.
It seems that this issue occurs specifically within `if-else` blocks, as `if` blocks alone appear to function correctly. 
Since its not specific to `dict.pop`, resolving it may fall outside the scope of this PR, but it should certainly be addressed in future work. Let me know what you think :smile:

pyccel/codegen/printing/ccode.py
```diff
@@ -2810,6 +2819,26 @@ def _print_SetDiscard(self, expr):
...
+        tmp_var = self.scope.get_temporary_variable(class_type.value_type)
+
+        if expr.default_value:
+            default = self._print(expr.default_value)
+            pop_expr = f"{tmp_var.name} = {c_type}_pop_with_default({dict_var}, {key}, {default});"
+        else:
+            pop_expr = f"{tmp_var.name} = {c_type}_pop({dict_var}, {key});"
+        self._additional_code += f"{pop_expr}\n"
+
+        return tmp_var.name
```
EmilyBourne, line 2840: Why do you need to create a temporary variable to hold the result of `pop`? Can't you just return the function call?

EmilyBourne:
@emmanuel-ferdman Thank you for letting me know about this. I have created a dedicated issue. The problem is to due with `_additional_code`. This variable is sometimes necessary in order to execute code that takes multiple lines in one command. For example:
```python
def test_min():
    a = min(5 + 6*7**5,6) + 42
```
which translates to:
```c
void test_min_2(void)
{
    int64_t a;
    int64_t Dummy_0000;
    int64_t Dummy_0001;
    Dummy_0000 = INT64_C(5) + INT64_C(6) * pow(7.0, 5.0);
    Dummy_0001 = INT64_C(6);
    a = (Dummy_0000 < Dummy_0001 ? Dummy_0000 : Dummy_0001) + INT64_C(42);
}
```
The presence of `Dummy_0000` prevents the calculation of `INT64_C(5) + INT64_C(6) * pow(7.0, 5.0)` being carried out multiple times. (The presence of `Dummy_0001` is overkill that I will fix along with #2149 ).

The only way to avoid multiple calls to `pow` and still have `a = min(5 + 6*7**5,6) + 42` is to do this over multiple lines.

Unfortunately there are a few places in the code where we have forgotten to think about `_additional_code` as it is uncommon for it to be necessary in these locations. `If` is apparently one of these places.

The fix is fairly simple. See #2151 . But it does make me wonder why you have this problem in your code. You say you expect the translation of:
```python
def f(k : int):
    a = {1:1.0, 2:2.0}
    b = {1:True, 2: False}
    if b.pop(k):
        return a.pop(k)
    else:
        return 3.0
```
to be:
```c
double f(int64_t k)
{
    hmap_int64_t_double a = {0};
    hmap_int64_t_bool b = {0};
    double Dummy_0000;
    bool Dummy_0001;
    double Out_0001;
    a = c_init(hmap_int64_t_double, {{INT64_C(1), 5.0}, {INT64_C(2), 6.0}});
    b = c_init(hmap_int64_t_bool, {{INT64_C(1), 1}, {INT64_C(2), 0}});
    Dummy_0001 = hmap_int64_t_bool_pop(&b, k); // <-- PROBLEM
    if (Dummy_0001)
    {
        Dummy_0000 = hmap_int64_t_double_pop(&a, k);
        Out_0001 = Dummy_0000;
        hmap_int64_t_double_drop(&a);
        hmap_int64_t_bool_drop(&b);
        return Out_0001;
    }
    else
    {
        hmap_int64_t_double_drop(&a);
        hmap_int64_t_bool_drop(&b);
        return 3.0;
    }
    hmap_int64_t_double_drop(&a);
    hmap_int64_t_bool_drop(&b);
}
```
But I would expet it to simply be:
```c
double f(int64_t k)
{
    hmap_int64_t_double a = {0};
    hmap_int64_t_bool b = {0};
    double Out_0001;
    a = c_init(hmap_int64_t_double, {{INT64_C(1), 5.0}, {INT64_C(2), 6.0}});
    b = c_init(hmap_int64_t_bool, {{INT64_C(1), 1}, {INT64_C(2), 0}});
    if (hmap_int64_t_bool_pop(&b, k))
    {
        Out_0001 = hmap_int64_t_double_pop(&a, k);
        hmap_int64_t_double_drop(&a);
        hmap_int64_t_bool_drop(&b);
        return Out_0001;
    }
    else
    {
        hmap_int64_t_double_drop(&a);
        hmap_int64_t_bool_drop(&b);
        return 3.0;
    }
    hmap_int64_t_double_drop(&a);
    hmap_int64_t_bool_drop(&b);
}
```

EmilyBourne (changes requested):

emmanuel-ferdman:
@EmilyBourne Un, great catch! Initially, I tried using the statement expression blocks `({ ... })`, which required a temporary variable. However, after reading more about the extensions, I switched.

Thank you for the explanation! The code no longer uses `_additional_code`, so now the Python program correcly generates:
```
double f(int64_t k)
{
    hmap_int64_t_double a = {0};
    hmap_int64_t_bool b = {0};
    double Out_0001;
    a = c_init(hmap_int64_t_double, {{INT64_C(1), 5.0}, {INT64_C(2), 6.0}});
    b = c_init(hmap_int64_t_bool, {{INT64_C(1), 1}, {INT64_C(2), 0}});
    if (hmap_int64_t_bool_pop(&b, k))
    {
        Out_0001 = hmap_int64_t_double_pop(&a, k);
        hmap_int64_t_double_drop(&a);
        hmap_int64_t_bool_drop(&b);
        return Out_0001;
    }
    else
    {
        hmap_int64_t_double_drop(&a);
        hmap_int64_t_bool_drop(&b);
        return 3.0;
    }
    hmap_int64_t_double_drop(&a);
    hmap_int64_t_bool_drop(&b);
}
```

EmilyBourne (approved):
This all looks good to me. I just want to check with @jalalium about the handling of extension imports before merging

pyccel/codegen/printing/ccode.py
jalalium: @EmilyBourne Indeed, In #2086, `_print_header` now automatically adds extension headers along with the stc headers. However, I should also have removed the old calls to both imports; I guess this is the cause of this confusion. I need to open a PR for this.

EmilyBourne: What do you want to do for this PR then? Merge as is? Remove this import?

EmilyBourne: Ok. I have created #2156 to track this. I am going to go ahead and merge this PR.

PR #2146, opened 2025-01-13
Fix class constness

### LINKED ISSUE #2145: Wrong constness detected when using class methods
### Describe the bug
The intent of classes is set to `in` due to the fact that they are scalars but the case for classes is not as simple as integers/floats/etc. This leads to problems when trying to use these classes.

### To Reproduce
```python
import numpy as np

class ArrProperties:
    def __init__(self, n : int):
        self._n_pts = n

    @property
    def n_points(self):
        return self._n_pts

def f(a : 'ArrProperties'):
    x = np.zeros(a.n_points)

if __name__ == '__main__':
    arr_properties = ArrProperties(5)
    f(arr_properties)
```

### Error details
```fortran
module wrong_intent

        C_INT64_T , b1 => C_BOOL

  implicit none

  type, public :: ArrProperties
    integer(i64) :: private_n_pts

    contains
    procedure :: create => arrproperties_create
    procedure :: n_points => arrproperties_n_points
    procedure :: free => arrproperties_free
  end type ArrProperties

  contains


  subroutine arrproperties_create(self, n)

    implicit none

    class(ArrProperties), intent(inout) :: self
    integer(i64), value :: n

    self%is_freed = .False._b1
    self%private_n_pts = n

  end subroutine arrproperties_create

  function arrproperties_n_points(self) result(Out_0001)

    implicit none

    integer(i64) :: Out_0001
    class(ArrProperties), intent(inout) :: self

    Out_0001 = self%private_n_pts
    return

  end function arrproperties_n_points

  subroutine arrproperties_free(self)

    implicit none

    class(ArrProperties), intent(inout) :: self

    if (.not. self%is_freed) then
      ! pass
      self%is_freed = .True._b1
    end if

  end subroutine arrproperties_free

  subroutine f(a)

    implicit none

    class(ArrProperties), intent(in) :: a
    real(f64), allocatable :: x(:)

    allocate(x(0:a % n_points() - 1_i64))
    x = 0.0_f64
    if (allocated(x)) then
      deallocate(x)
    end if

  end subroutine f

end module wrong_intent
```

### Expected behavior
The function f should pass `a` with intent `inout`:
```fortran
  subroutine f(a)

    implicit none

    class(ArrProperties), intent(inout) :: a
    real(f64), allocatable :: x(:)

    allocate(x(0:a % n_points() - 1_i64))
    x = 0.0_f64
    if (allocated(x)) then
      deallocate(x)
    end if

  end subroutine f
```

### Language
Fortran

### Additional context
Appears in PyGyro

EmilyBourne:
Fix the constness of class instances in functions and class methods. Fixes #2145.


pyccel/ast/core.py
```diff
@@ -1572,7 +1572,8 @@ def __init__(self, name, *, value = None, kwonly=False, annotation=None, bound_a
             name.declare_as_argument()
 
         if pyccel_stage != "syntactic":
-            self._inout = self.var.rank>0 and not self.var.is_const if isinstance(self.var, Variable) else False
+            self._inout = (self.var.rank > 0 or isinstance(self.var.class_type, CustomDataType)) \
+                        and not self.var.is_const if isinstance(self.var, Variable) else False
```
jalalium, line 1576:
```suggestion
            self._inout = (self.var.rank > 0 or isinstance(self.var.class_type, CustomDataType)) 
            self._inout &= not (isinstance(self.var, Variable) and self.var.is_const)
```

jalalium (approved):
Well done!

pyccel/ast/core.py
EmilyBourne: This correction was incorrect. The "if" originally had precedence over the "and" so the result is False if the var is not a Variable (i.e. if it is a FunctionAddress)

PR #2150, opened 2025-01-13
Fix intel compilation

### LINKED ISSUE #2001: Crashing Intel tests

EmilyBourne:
Auto-generate the capsule clean up functions. These functions are called by Python's garbage collector when arrays go out of scope. By auto-generating these functions we are able to use Fortran code to deallocate Fortran arrays. Fixes #2001. One deallocation function is generated for each data type. In C only one capsule clean up function is generated. The generated functions have a prefix corresponding to the module name to avoid name collisions when loading modules.


EmilyBourne:
This solution was unsuccessful

PR #2151, opened 2025-01-13
Fix use of `_additional_code` in If condition

### LINKED ISSUE #2149: Function call in if condition printed in wrong location
Regarding your interesting example:
```python
def f(k : int):
    a = {1:1.0, 2:2.0}
    b = {1:True, 2: False}
    if b.pop(k):
        return a.pop(k)
    else:
        return 3.0
```
It generates the following C function:
```c
double f(int64_t k)
{
    hmap_int64_t_double a = {0};
    hmap_int64_t_bool b = {0};
    double Dummy_0000;
    bool Dummy_0001;
    double Out_0001;
    a = c_init(hmap_int64_t_double, {{INT64_C(1), 5.0}, {INT64_C(2), 6.0}});
    b = c_init(hmap_int64_t_bool, {{INT64_C(1), 1}, {INT64_C(2), 0}});
    if (Dummy_0001)
    {
        Dummy_0000 = hmap_int64_t_double_pop(&a, k);
        Out_0001 = Dummy_0000;
        hmap_int64_t_double_drop(&a);
        hmap_int64_t_bool_drop(&b);
        return Out_0001;
    }
    else
    {
        Dummy_0001 = hmap_int64_t_bool_pop(&b, k); // <-- PROBLEM
        hmap_int64_t_double_drop(&a);
        hmap_int64_t_bool_drop(&b);
        return 3.0;
    }
    hmap_int64_t_double_drop(&a);
    hmap_int64_t_bool_drop(&b);
}
```
Beside the 'header guard' issue that I fixed, I think there is another issue here - `hmap_int64_t_bool_pop(&b, k)` is evaluated inside the `else` block instead of being computed before the conditional statement. This results in incorrect behavior, as the conditional logic depends on the value of `b.pop(k)` but evaluates it at the wrong point.

This issue is not specific to the `dict.pop` case. It also occurs in other cases, such as:
```python
def test_min():
    if min(5,6):
        return 1
    else:
        return 0
```
Which generates the following C code:
```c
int64_t test_min(void)
{
    int64_t Dummy_0000;
    int64_t Dummy_0001;
    if (((Dummy_0000 < Dummy_0001 ? Dummy_0000 : Dummy_0001) != 0))
    {
        return INT64_C(1);
    }
    else
    {
        Dummy_0000 = INT64_C(5); // <-- PROBLEM
        Dummy_0001 = INT64_C(6);
        return INT64_C(0);
    }
}
```
Here, the variables `Dummy_0000` and `Dummy_0001` used in the `min` function are only evaluated in the `else` block, leading to incorrect behavior because the condition depends on their values.
It seems that this issue occurs specifically within `if-else` blocks, as `if` blocks alone appear to function correctly. 
Since its not specific to `dict.pop`, resolving it may fall outside the scope of this PR, but it should certainly be addressed in future work. Let me know what you think :smile:

_Originally posted by @emmanuel-ferdman in https://github.com/pyccel/pyccel/issues/2143#issuecomment-2587577755_

EmilyBourne:
Fix use of `_additional_code` in If condition. Fixes #2149. Additionally avoid creating variables to hold literals in min/max functions.


tests/epyccel/test_builtins.py
```diff
@@ -122,6 +122,19 @@ def f(x : 'T', y : 'T', z : 'T'):
...
+def test_min_if(language):
+    def f(x : 'int', y : 'int'):
+        if min(x*y, x+y) < 2:
+            return x*y
+        else:
+            return x-y
+
+    epyc_f = epyccel(f, language=language)
+
+    int_args = [randint(min_int, max_int) for _ in range(2)]
```
EmilyBourne, line 134:
```suggestion
    def f(x : 'int', y : 'int'):
        if min(x+x+y, x+y+y) < (x+y):
            return x+y
        else:
            return x-y

    epyc_f = epyccel(f, language=language)

    int_args = [randint(min_int//3, max_int//3) for _ in range(2)]
```

jalalium (changes requested):
I don't think this solves the issue:
```python
def f(x : 'int', y : 'int'):
    if min(min(x+x,+y), min(x+y,y)) < (x+y):
        return x+y
    else:
        return x-y
```
gives
```C
int64_t f(int64_t x, int64_t y)
{
    int64_t Dummy_0000;
    int64_t Dummy_0001;
    int64_t Dummy_0002;
    int64_t Dummy_0003;
    int64_t Dummy_0004;
    Dummy_0000 = (Dummy_0001 < Dummy_0002 ? Dummy_0001 : Dummy_0002);
    Dummy_0003 = (Dummy_0004 < y ? Dummy_0004 : y);
    if ((Dummy_0000 < Dummy_0003 ? Dummy_0000 : Dummy_0003) < x + y)
    {
        return x + y;
    }
    else
    {
        return x - y;
    }
}
```
In addition, for the test you added, don't you think that this prints better:

```C
int64_t f(int64_t x, int64_t y)
{
    if (int64_t_min(2, x + x + y, x + y + y) < x + y)
    {
        return x + y;
    }
    else
    {
        return x - y;
    }
}
```
I think using the ternary operator for 2 arguments needs revisiting, it is not faster than an inline function with today's compilers and it introduces too many temps for things like this `min(min(min(min(min(x,y),y),y),y),y)`

EmilyBourne:
@jalalium Good catch! Indeed if you do:
```python
self._additional_code += self._print(arg)
```
and `self._print(arg)` modifies `self._additional_code` then this modification is lost. The code is equivalent to:
```python
tmp1 = self._additional_code
tmp2 = self._print(arg) # Modifies self._additional_code
self._additional_code = tmp1 + tmp2
```

I have fixed this: [0c5132b](https://github.com/pyccel/pyccel/pull/2151/commits/0c5132b20e31c39797eba0b2a7a4c9f73f266c1a)

Regarding your comment about the printing. I think this is out of scope for this PR which is entirely about the correct use of `self._additional_code`.

If we want to change this then we could open a dedicated issue. I agree that an inline function should be equivalent to the ternary. I'm not sure how the compilers handle the arguments in this case though. The main reason for the temporaries is to avoid recalculating expressions. This is important if the expression is costly, but also if the expression has side effects. E.g:
```python
def get_1():
    print(1)
    return 1

def get_2():
    print(2)
    return 2

if __name__ == '__main__':
    print(min(get_1(), get_2()))
```
prints `1 2 1`. If the translation was `printf("%"PRId64"\n", (get_1() < get_2() ? get_1() : get_2()));` instead of:
```c
    Dummy_0000 = get_1();
    Dummy_0001 = get_2();
    printf("%"PRId64"\n", (Dummy_0000 < Dummy_0001 ? Dummy_0000 : Dummy_0001));
```
Then more values would be printed.

I don't know if an inline function is equivalent to our current implementation or if it would evaluate the expressions multiple times.

I don't know that `int64_t_min(a,b)` is significantly more readable than `(a < b ? a : b)` in C code where min is often implemented with a ternary (https://stackoverflow.com/questions/3437404/min-and-max-in-c)

jalalium (approved):
Well done!

PR #2154, opened 2025-01-15
Fix missing line information when an unknown class method is called

### LINKED ISSUE #2153: Row and column numbers are not displayed in case of an error.
### Describe the bug

Error on
pyccell arrerr0.py --language fortran

### To Reproduce
Python script arrerr0.py:
```
[0] import numpy as np
[1] def f1(n: int, V: float) -> 'float[:]': # That's right: V: 'float[:]'
[2]     U = np.empty(V.size) # Here is error.
[3]     for i in range(V.size):
[4]         U[i] = n * V[i]
[5]     return U
```

### Error details

```
pyccel:
 |fatal [semantic]: arrerr0.py| Can't find method size in class float (<pyccel.ast.core.ClassDef object at 0x79e9718b6290>)
```

### Expected behavior

The row and column numbers of the error position in the error message are expected.

### Language

Fortran

### Additional context

pyccel 1.12.1 from ...сonda/lib/python3.12/site-packages/pyccel (python 3.12)

EmilyBourne:
Fix missing line information when an unknown class method is called by changing the prototype of `ClassDef.get_method`. Fixes #2153


jalalium (approved):
Well done!

PR #2157, opened 2025-01-16
Remove BindCFunctionDefResult

EmilyBourne:
Remove `BindCFunctionDefResult`. It is replaced with the class `BindCVariable` which links new C-compatible variables to their equivalent in Fortran. When returning arrays multiple variables are returned from Fortran this is now represented using tuples. The type `BindCArrayType` subclasses `InhomogeneousTupleType` to describe the tuple `(a.data, *a.shape)` for an array `a`.

The code in `fortran_to_c_wrapper.py` is re-written to use `_extract_X_FunctionDefResult` functions to improve readability.

This PR is important as a first step towards #337 . After #337 we will have 1 `FunctionDefResult` per function so we cannot use `FunctionDefResult` or its subclass `BindCFunctionDefResult` to describe the relationships between array data and shape.

This PR will also make it easier to implement #2060 .

**Commit Summary**
- Make docstrings optional for `_extract_X_FunctionDefResult` to avoid copy/paste docstrings
- Add new datatype class `BindCArrayType`.
- Rename existing class `BindCVariable`->`BindCModuleVariable`
- Add new class `BindCVariable`
- Remove class `BindCFunctionDefResult`
- Store `FunctionDefResult(BindCResultVariable)` in `BindCFunctionDef` instead of `BindCFunctionDefResult`
- Add function `flatten_tuple_var` to obtain all low-level variables in a multi-level tuple.
- Preserve scope when creating external function definitions so inhomogeneous tuple elements can be retrieved for arguments or results.
- Add method `_extract_BindCArrayType_FunctionDefResult` to `CToPythonWrapper` as Fortran and C NDArray wrapping is quite different.
- Introduce `_extract_FunctionDefResult` methods in `FortranToCWrapper` to improve readability.
- Add `_get_bind_c_array` method in `FortranToCWrapper` to avoid code duplication.


pyccel/ast/bind_c.py
```diff
@@ -312,94 +312,49 @@ def wrapping_bound_argument(self):
 
 # =======================================================================================
 
-
-class BindCFunctionDefResult(FunctionDefResult):
+class BindCResultVariable(Variable):
```
EmilyBourne, line 315: Maybe this new class could be useful outside the concept of results in the future. Can we find a better name? `BindCVariable` is already taken (for module variables). Maybe the clearest name involves renaming `BindCVariable`?

EmilyBourne:
/bot run pylint pyccel_lint docs spelling linux

pyccel/ast/bind_c.py
jalalium: I agree! maybe we should rename `BindCVariable`

pyccel/ast/utilities.py
```diff
@@ -839,3 +840,32 @@ def is_literal_integer(expr):
     """
     return isinstance(expr, (int, LiteralInteger)) or \
         isinstance(expr, PyccelUnarySub) and isinstance(expr.args[0], (int, LiteralInteger))
+
+#==============================================================================
+def flatten_tuple_var(expr, scope):
```
jalalium, line 845: Do you intend to use this function with nested inhomogeneous tuples?

pyccel/ast/bind_c.py
EmilyBourne: `BindCResultVariable`->`BindCVariable`
`BindCVariable`->`BindCModuleVariable`
?

pyccel/ast/utilities.py
EmilyBourne: Yes that is the aim. It will become more important when I fix #337

pyccel/ast/bind_c.py
EmilyBourne: Done:
[5520a73](https://github.com/pyccel/pyccel/pull/2157/commits/5520a7341268bc0eaa8e973f07f612b45dd94d0d)
[e940783](https://github.com/pyccel/pyccel/pull/2157/commits/e940783c0fc887fc5bfdaf3ba7331f62f58a5f64)
[9f258dd](https://github.com/pyccel/pyccel/pull/2157/commits/9f258dda14ac986655016134c974813178d8acdc)
[9161b46](https://github.com/pyccel/pyccel/pull/2157/commits/9161b46205cacf5e62eb5f8d2bd2c1899f951a5e)

pyccel/ast/utilities.py
jalalium: That's what I thought, but such a mapping is lossy? shouldn't the strides and shape be computed here as well? or is it taken care of in a separate function?

EmilyBourne: I'm not sure what you mean. `BindCArrayType` inherits from `InhomogeneousTupleType` so the elements of this tuple are the shape and stride variables. You lose the information about how they are linked to one another but this function is only used when that information is not required. Eg to print `a = get_array` we call `flatten_tuple_var` on `a` to get `(a_ptr, a_shape0, a_shape1)` (no strides because this is a return type) so we can print
```
get_array(&a_ptr, &a_shape0, &a_shape1)
```

The way they are linked is still available elsewhere in the code because the result is not saved in a normal `InhomogeneousTupleType` but rather in a `BindCArrayType`.

EmilyBourne: If a function returned `(1, my_array)` then this method would see `(1, (my_array_ptr, my_array_shape))` and would return `(1, my_array_ptr, my_array_shape)`

jalalium: Got it. Thank you!

jalalium (approved):
Well done!

PR #2158, opened 2025-01-16
Allow returning a dictionary from C

### LINKED ISSUE #2059: [DICTS] Support dicts as results in C
### Describe the feature
Support returning dicts from C code

### Test Code
```python
def f():
    a = {1:1.0}
    return a
```

### Proposed Solution
The cwrapper should be improved to ensure that dicts can be read from C code and passed to Python.

EmilyBourne:
Allow returning a dictionary from C. Fixes #2059


jalalium (approved):
Well done!

PR #2160, opened 2025-01-16
Add C support for `dict.clear`

### LINKED ISSUE #2045: [DICT] Add C support for clear
### Describe the feature
C support should be added for `clear`.

### Test Code
```python
def foo():
    a = {1,1.0, 2:2.0}
    a.clear()
    return len(a)
```

### Proposed Solution
This should be implemented using the `hmap_X_clear` method documented here: 

emmanuel-ferdman:
Add C support for dict method `clear()`. Fixes #2045

**Commit Summary**
- Implement C support for the `dict.clear()` method.
- Activate `dict.clear()` tests for C.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 -   #1881 : Add Python support for dict method `copy()`.
--   #1880 : Add Python support for dict method `clear()`.
+-   #1880 : Add Python and C support for dict method `clear()`.
 -   #1720 : Add support for `Ellipsis` as the only index for an array.
 -   #1787 : Ensure STC is installed with Pyccel.
 -   #1656 : Ensure gFTL is installed with Pyccel.
--- a/docs/builtin-functions.md
+++ b/docs/builtin-functions.md
@@ -120,7 +120,7 @@ Python contains a limited number of builtin functions defined [here](https://doc
 
 | Method | Supported |
 |----------|-----------|
-| `clear` | Python-only |
+| `clear` | Python and C |
 | `copy` | Python-only |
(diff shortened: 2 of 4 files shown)


emmanuel-ferdman:
Here are some examples:

Example 1:
```python
def dict_clear_test1():
    a = {1: 1.0, 2: 2.0}
    a.clear()
    return len(a)
```
Generated C Code:
```c
int64_t dict_clear_test1(void)
{
    hmap_int64_t_double a = {0};
    int64_t Out_0001;
    a = c_init(hmap_int64_t_double, {{INT64_C(1), 1.0}, {INT64_C(2), 2.0}});
    hmap_int64_t_double_clear(&a);
    Out_0001 = hmap_int64_t_double_size(&a);
    hmap_int64_t_double_drop(&a);
    return Out_0001;
}
```

Example 2:
```python
def dict_clear_test2():
    a = {True:2, False:1}
    a.clear()
    return len(a)
```
Generated C Code:
```c
int64_t dict_clear_test2(void)
{
    hmap_bool_int64_t a = {0};
    int64_t Out_0001;
    a = c_init(hmap_bool_int64_t, {{1, INT64_C(2)}, {0, INT64_C(1)}});
    hmap_bool_int64_t_clear(&a);
    Out_0001 = hmap_bool_int64_t_size(&a);
    hmap_bool_int64_t_drop(&a);
    return Out_0001;
}
```

Please note that I modified the `dict_clear` method in the `test_dict_clear` test case to return `len(a)` instead of the dictionary itself, as returning a dictionary is not yet implemented in C. We can revert this change after PR #2158 is merged, although I’m not sure if it makes much of a difference.

EmilyBourne (approved):
Thanks again. Everything looks good to me

PR #2168, opened 2025-01-20
Improve `STC` extensions imports

jalalium:
This PR aims to clean `STC` imports by automatically adding extensions`along with the original headers. In addition, extension imports now include the folder name to avoid unrecognized includes in programs or modules that import modules that use stc extensions.


pyccel/codegen/printing/ccode.py
```diff
@@ -1089,6 +1084,8 @@ def _print_Import(self, expr):
         if source == 'stc/common':
             code = ''
             for t in expr.target:
+                self.add_import(Import(stc_extension_mapping[source],(),
+                       ignore_at_print=True))
```
EmilyBourne, line 1088:
```suggestion
                self.add_import(Import(stc_extension_mapping[source], ignore_at_print=True))
```

EmilyBourne: The `target` argument is optional. Better to let `Import` decide on the default

jalalium: Got it! I will commit your suggestion as soon as the test finishes.

EmilyBourne: Oops. Sorry it looks like I misremembered/misread that :eyes:

pyccel/codegen/printing/ccode.py
```diff
@@ -1089,6 +1084,7 @@ def _print_Import(self, expr):
         if source == 'stc/common':
             code = ''
             for t in expr.target:
+                self.add_import(Import(stc_extension_mapping[source], (), ignore_at_print=True))
```
EmilyBourne, line 1087: Does this line need to be in the loop? Surely it only needs adding once per source not once per target?

pyccel/codegen/utilities.py
```diff
@@ -51,19 +51,19 @@
     "cwrapper"        : ("cwrapper", CompileObj("cwrapper.c",folder="cwrapper", accelerators=('python',))),
     "numpy_f90"       : ("numpy", CompileObj("numpy_f90.f90",folder="numpy")),
     "numpy_c"         : ("numpy", CompileObj("numpy_c.c",folder="numpy")),
-    "Set_extensions"  : ("STC_Extensions", CompileObj("Set_Extensions.h",
+    "STC_Extensions/Set_extensions"  : ("STC_Extensions", CompileObj("Set_Extensions.h",
                                                       folder="STC_Extensions",
```
EmilyBourne, line 55: I don't think the folder is needed if you use `STC_Extensions/List_Extension.h`. Ideally the enclosing folder should be specified but I don't know if there's a way to do this?

```suggestion
                                                       folder=".",
```
?

pyccel/codegen/printing/ccode.py
jalalium: Good point!

pyccel/codegen/utilities.py
jalalium:
```Python
"gFTL_functions/Set_extensions"  : ("gFTL_functions", CompileObj("Set_Extensions.inc",
                                                                     folder="gFTL_functions",
                                                                     has_target_file = False,
                                                                     dependencies = (external_libs['gFTL'][1],))),
```
I followed this pattern and I think that your suggestion will create a deadlock because the enclosing folder will be say `__pyccel__` and thus can't be acquired.

EmilyBourne: Ah! Good point. I didn't think about the fact that the folder is used for avoiding deadlocks too. In this case it should indeed be `STC_Extensions` to protect these files

EmilyBourne (approved):
Looks good to me

PR #2170, opened 2025-01-21
Add support for set initialisation from other objects

EmilyBourne:
Add support for set initialisation from other objects via the `set` function. The support is added via `_build_PythonSetFunction`. This removes the need for `PythonSetFunction.__new__`.

While fixing this some bugs were found and fixed. The first bug is relative to temporaries used as arguments to `union` or `intersection`. Temporary objects are now created to store these temporaries.

The second bug is related to the use of `merge` to represent `union` in Fortran. This is not the correct function. In fact `merge` empties the set passed as argument. Instead `insert(a%begin(), a%end())` should be used.

**Commit Summary**
- Remove unnecessary `PythonSetFunction.__new__` function.
- Modify `init_stc_container` to pass `lhs` instead of `Assign`.
- Add implementation of `PythonSet` in `CCodePrinter` in case temporary set literals are used.
- Implement `_print_Allocate` for `PyCodePrinter` in order to handle `reserve` allocations.
- Allow calling `union` on a temporary object.
- Allow calling `intersection` on a temporary object.
- Add `_build_PythonSetFunction`.
- Activate relevant tests.
- Group all `union` tests together.
- Group all `intersection` tests together.
- Add new test `test_set_union_temporaries`.
- Add new test `test_set_intersection_int_temporaries`.
- Fix printing of `set` in Python.
- Fix `union` implementation in Fortran.


jalalium (approved):
Well done!

PR #2171, opened 2025-01-21
Add `min/max` support for `Lists` and `Sets`

jalalium:
This PR aims to add support for functions `min` and `max` with `lists` and `sets`.
Necessary extension functions were added for `STC` and `gFTL`


pyccel/stdlib/STC_Extensions/List_extensions.h
```diff
@@ -19,8 +15,34 @@ static inline i_key _c_MEMB(_pull_elem)(i_type* self, intptr_t pop_idx) {
     }
     return *(itr.ref); // Return the element that is being popped.
 }
+#if defined(_Imaginary_I) || defined(_Complex_I) || defined(_Complex)
+    /* Do nothing */
+#else
```
EmilyBourne, line 20:
```suggestion
#if !defined(_Imaginary_I) && !defined(_Complex_I) && !defined(_Complex)
```
?
I don't know if you need all of these though

jalalium: Probably not! Just to be sure. Can you think of something better?

EmilyBourne: For lists maybe you should use `#if i_use_cmp`?

jalalium: I used it before, but to keep consistent with `sets` I opted to check for a complex `i_key`

pyccel/ast/builtins.py
```diff
@@ -1485,10 +1485,12 @@ def __init__(self, *x):
 
         if isinstance(x, (list, tuple)):
             x = PythonTuple(*x)
+        elif isinstance(x, Variable) and isinstance(x.class_type, (HomogeneousListType, HomogeneousSetType, HomogeneousTupleType)):
+            pass
         elif not isinstance(x, (PythonTuple, PythonList)):
```
EmilyBourne, line 1490:
```suggestion
        elif not (isinstance(x, (PythonTuple, PythonList)) or isinstance(x, Variable) and isinstance(x.class_type, HomogeneousContainerType)):
```
?

pyccel/codegen/printing/ccode.py
```diff
@@ -822,13 +822,47 @@ def _print_PythonRound(self, expr):
     def _print_PythonMinMax(self, expr):
         arg = expr.args[0]
         primitive_type = arg.dtype.primitive_type
+        variadic_args = False
+        can_compare = True
 
-        if primitive_type is PrimitiveFloatingPointType() and len(arg) == 2:
+        if isinstance(primitive_type, (PrimitiveFloatingPointType, PrimitiveIntegerType)):
+            variadic_args = True
```
EmilyBourne, line 829:
```suggestion
        variadic_args = isinstance(primitive_type, (PrimitiveFloatingPointType, PrimitiveIntegerType))
```

pyccel/codegen/printing/ccode.py
```diff
@@ -822,13 +822,47 @@ def _print_PythonRound(self, expr):
...
+
+        if primitive_type is PrimitiveComplexType():
+            can_compare = False
```
EmilyBourne, line 832:
```suggestion
        can_compare = primitive_type is not PrimitiveComplexType()
```

pyccel/codegen/printing/ccode.py
```diff
@@ -822,13 +822,47 @@ def _print_PythonRound(self, expr):
...
+
+        if isinstance(arg, Variable) and isinstance(arg.class_type , HomogeneousTupleType):
+            arg = PythonTuple(*arg)
```
EmilyBourne, line 835: This is not valid. A `HomogeneousTupleType` can have unknown size. In this case `*arg` will return an infinite iterator.

Examples where `HomogeneousTupleType` has unknown size:
```
def f(tup : 'tuple[int,...]'):
    ...

def f(n : int):
    tup = (1, 2, 3)
    longer_tup = tup * n
```

pyccel/codegen/printing/ccode.py
```diff
@@ -822,13 +822,47 @@ def _print_PythonRound(self, expr):
...
+
+        if isinstance(arg, (PythonTuple, PythonList)) and variadic_args:
+            key = self.get_c_type(arg.class_type.element_type)
+            self.add_import(Import('stc/common', AsName(VariableTypeAnnotation(arg.dtype), key)))
+            return  f'{key}_{expr.name}({len(arg.args)}, {", ".join(self._print(a) for a in arg.args)})'
```
EmilyBourne, line 840: I find this string quite hard to read. Maybe break it up slightly:
```suggestion
            args_code = ", ".join(self._print(a) for a in arg.args)
            return  f'{key}_{expr.name}({len(arg.args)}, {args_code})'
```

pyccel/codegen/printing/ccode.py
```diff
@@ -822,13 +822,47 @@ def _print_PythonRound(self, expr):
...
+                c_type = self.get_c_type(class_type)
+                arg_obj = self._print(ObjectAddress(arg))
+                self.add_import(Import('stc/vec', AsName(VariableTypeAnnotation(class_type), c_type)))
+                return f'{c_type}_{expr.name}({arg_obj})'
+            elif isinstance(arg.class_type, HomogeneousSetType) and can_compare:
+                class_type = arg.class_type
+                c_type = self.get_c_type(class_type)
+                arg_obj = self._print(ObjectAddress(arg))
+                self.add_import(Import('stc/hset', AsName(VariableTypeAnnotation(class_type), c_type)))
+                return f'{c_type}_{expr.name}({arg_obj})'
```
EmilyBourne, line 853: There is only 1 line difference between these 2. Maybe unify?
```suggestion
            if isinstance(arg.class_type, (HomogeneousListType, HomogeneousSetType)) and can_compare:
                class_type = arg.class_type
                c_type = self.get_c_type(class_type)
                arg_obj = self._print(ObjectAddress(arg))
                import_loc = 'stc/vec' if isinstance(arg.class_type, HomogeneousListType)) else 'stc/hset'
                self.add_import(Import('stc/vec', AsName(VariableTypeAnnotation(class_type), c_type)))
                return f'{c_type}_{expr.name}({arg_obj})'
```
This may be slightly slower though so I don't know if the trade-off is worth it.

pyccel/codegen/printing/ccode.py
```diff
@@ -822,13 +822,47 @@ def _print_PythonRound(self, expr):
...
+            else:
+                return errors.report(f"{expr.name} in C does not support arguments of type {arg.class_type}", symbol=expr,
+                    severity='fatal')
+        if len(arg) != 2:
+            return errors.report(f"{expr.name} in C does not support {len(arg)} arguments of type {arg.class_type}", symbol=expr,
+                                 severity='fatal')
+        if arg.dtype.primitive_type is PrimitiveFloatingPointType():
```
EmilyBourne, line 860:
```suggestion
        if primitive_type is PrimitiveFloatingPointType():
```

pyccel/codegen/printing/ccode.py
```diff
@@ -846,20 +879,19 @@ def _print_PythonMinMax(self, expr):
...
                 arg2 = self._print(arg2_temp)
-
             op = '<' if isinstance(expr, PythonMin) else '>'
             return f"({arg1} {op} {arg2} ? {arg1} : {arg2})"
         elif len(arg) > 2 and isinstance(arg.dtype.primitive_type, (PrimitiveFloatingPointType, PrimitiveIntegerType)):
             key = self.get_c_type(arg[0].class_type)
             self.add_import(Import('stc/common', AsName(VariableTypeAnnotation(arg.dtype), key)))
             return  f'{key}_{expr.name}({len(arg)}, {", ".join(self._print(a) for a in arg)})'
-        elif isinstance(primitive_type, PrimitiveComplexType):
+        elif isinstance(primitive_type, PrimitiveComplexType) and len(arg) == 2:
```
EmilyBourne, line 888: Is this change still needed?

pyccel/codegen/printing/fcode.py
```diff
@@ -1875,28 +1881,31 @@ def _print_NumpyAmin(self, expr):
         else:
             return f'minval({arg_code})'
         
-    def _print_PythonMin(self, expr):
+    def _print_PythonMinMax(self, expr):
         args = expr.args
```
EmilyBourne, line 1885:
```suggestion
        arg, = expr.args
```
?

pyccel/stdlib/STC_Extensions/List_extensions.h
```diff
@@ -18,5 +19,28 @@ static inline i_key _c_MEMB(_pull_elem)(Self* self, intptr_t pop_idx) {
     }
     return *(itr.ref); // Return the element that is being popped.
 }
+#if defined(i_use_cmp)
+// Function to get the minimum element from the vector
+static inline i_key _c_MEMB(_min)(const Self* self) {
+    i_key min_val = *_c_MEMB(_front)(self);
+    c_foreach(it, Self, *self) {
+        if (*(it.ref) < min_val) {
```
EmilyBourne, line 27: Shouldn't you use `i_less` to ensure that this keeps working if we add complex support later?

pyccel/stdlib/STC_Extensions/Set_extensions.h
```diff
@@ -56,5 +56,34 @@ static inline void _c_MEMB(_intersection_update)(Self* self, Self* other) {
...
     }
 }
+#if defined (i_use_cmp) 
+// Function to get the minimum element from the set
+static inline i_key _c_MEMB(_min)(const Self* self) {
+    _c_MEMB(_iter) itr = _c_MEMB(_begin)(self);
+    i_key min_val = *(itr.ref);
+    while (itr.ref)
+    {
+        if (*(itr.ref) < min_val)
```
EmilyBourne, line 66: Shouldn't you use i_less to ensure that this keeps working if we add complex support later?

pyccel/stdlib/gFTL_functions/Set_extensions.inc
```diff
@@ -41,4 +41,33 @@ contains
...
+    class(Set), intent(in) :: my_set
+    __T_declare_dummy__ :: result
+
+    type(SetIterator) :: iter
+    type(SetIterator) :: last
+
+    iter = my_set%begin()
+    last = my_set%end()
+
+    result = iter%of()
```
EmilyBourne, line 54: Here you rely on the fact that sets are sorted? Can you add a comment to indicate this

pyccel/stdlib/gFTL_functions/Vector_extensions.inc
```diff
@@ -0,0 +1,64 @@
...
+contains
+
+#define __IDENTITY(x) x
+#define __guard __vector_guard
+#include "parameters/T/copy_vector_T_to_internal_T.inc"
+#include "parameters/T/define_derived_macros.inc"
+
+#ifdef __T_IS_COMPLEX
+  #error "Complex types are not supported for min/max operations"
+#else
```
EmilyBourne, line 12: Why do you need this? Complex comparison is available in Fortran I already added it to have sorted sets

EmilyBourne (changes requested):

pyccel/ast/builtins.py
```diff
@@ -1485,10 +1485,12 @@ def __init__(self, *x):
 
         if isinstance(x, (list, tuple)):
             x = PythonTuple(*x)
+        elif isinstance(x, Variable) and isinstance(x.class_type, (HomogeneousListType, HomogeneousSetType, HomogeneousTupleType)):
+            pass
```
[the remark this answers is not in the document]
jalalium: It's because I haven't implemented a printer for `ndarrays` `min` and `max`. Do you think it is best to change this to:
`isinstance(x.class_type, HomogeneousContainerType)` and fail at the printing stage?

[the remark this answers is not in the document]
EmilyBourne: Yes I think so. You catch it at the printing stage anyway so there is no need to catch it twice, especially as this is not a semantic restriction. It is very much language-implementation specific (e.g. this should work for Python)

pyccel/ast/builtins.py
jalalium: fixed!

pyccel/codegen/printing/ccode.py
jalalium: Oops, fixed!

pyccel/codegen/printing/ccode.py
jalalium: fixed!

pyccel/codegen/printing/ccode.py
jalalium: I do not know how to check if the size is known at compile time. Could you please give me a hint?

pyccel/codegen/printing/ccode.py
jalalium: fixed!

pyccel/codegen/printing/ccode.py
jalalium: fixed!

pyccel/codegen/printing/fcode.py
jalalium: fixed!

pyccel/stdlib/STC_Extensions/List_extensions.h
jalalium: fixed!

pyccel/stdlib/STC_Extensions/Set_extensions.h
jalalium: fixed!

pyccel/stdlib/gFTL_functions/Set_extensions.inc
jalalium: fixed!

pyccel/stdlib/gFTL_functions/Vector_extensions.inc
jalalium: fixed!

pyccel/codegen/printing/ccode.py
EmilyBourne: The size is known if `shape[0]` is a literal

EmilyBourne: Does this mean that this implementation only works for tuples/lists of known size?

EmilyBourne: What about explicitly sized tuples? Does it support:
```python
def f(tup : 'tuple[int, int, int]'):
    return max(*tup)
```

EmilyBourne: Hmm, that's a shame as it means that there is no way to call min/max on a tuple that is an argument. Couldn't min/max for tuples be mapped to `NumpyAmin`/`NumpyAmax` in the implementation?

jalalium: Sure they can, I will do it through `cspan` extensions; but it is out of the scope of this PR I think. I will dedicate a separate PR to handle homogeneous tuples.

pyccel/ast/builtins.py
```diff
@@ -1485,10 +1485,11 @@ def __init__(self, *x):
 
         if isinstance(x, (list, tuple)):
             x = PythonTuple(*x)
-        elif not isinstance(x, (PythonTuple, PythonList)):
+        elif not (isinstance(x, (PythonTuple, PythonList))
+                  or isinstance(x, Variable) and isinstance(x.class_type, HomogeneousContainerType)):
```
EmilyBourne, line 1489: I can never remember which takes precedence between `or` and `and`. Maybe add some parentheses for clarity. Same remark on line 1525

pyccel/codegen/printing/ccode.py
```diff
@@ -822,13 +822,40 @@ def _print_PythonRound(self, expr):
...
+                arg_obj = self._print(ObjectAddress(arg))
+                import_loc = 'stc/vec' if isinstance(arg.class_type, HomogeneousListType) else 'stc/hset'
+                self.add_import(Import(import_loc, AsName(VariableTypeAnnotation(class_type), c_type)))
+                return f'{c_type}_{expr.name}({arg_obj})'
+            else:
+                return errors.report(f"{expr.name} in C does not support arguments of type {arg.class_type}", symbol=expr,
+                    severity='fatal')
+        if len(arg) != 2:
+            return errors.report(f"{expr.name} in C does not support {len(arg)} arguments of type {arg.class_type}", symbol=expr,
+                                 severity='fatal')
```
EmilyBourne, line 852: A lot of the errors in this function are things that aren't implemented yet rather than things that we never plan on implementing. Can you use `PYCCEL_RESTRICTION_TODO` in the message to encourage people to open issues if they need these tools to be implemented sooner rather than later

pyccel/codegen/printing/ccode.py
```diff
@@ -822,13 +822,40 @@ def _print_PythonRound(self, expr):
...
+                arg_obj = self._print(ObjectAddress(arg))
+                import_loc = 'stc/vec' if isinstance(arg.class_type, HomogeneousListType) else 'stc/hset'
+                self.add_import(Import(import_loc, AsName(VariableTypeAnnotation(class_type), c_type)))
+                return f'{c_type}_{expr.name}({arg_obj})'
+            else:
+                return errors.report(f"{expr.name} in C does not support arguments of type {arg.class_type}", symbol=expr,
+                    severity='fatal')
+        if len(arg) != 2:
+            return errors.report('\n'.join([f"{expr.name} in C does not support {len(arg)} arguments of type {arg.dtype}",
+                                            PYCCEL_RESTRICTION_TODO]), symbol=expr, severity='fatal')
```
EmilyBourne, line 852: Join seems like overkill for this. You realise that you can just do:
```suggestion
            return errors.report(f"{expr.name} in C does not support {len(arg)} arguments of type {arg.dtype}\n"
                                 + PYCCEL_RESTRICTION_TODO, symbol=expr, severity='fatal')
```
?

EmilyBourne (approved):
Looks good to me

EmilyBourne (approved):

PR #2172, opened 2025-01-21
Restore developer mode for tests

EmilyBourne:
Developer mode for tests was lost after #996 where the developer-mode was set to user by default instead of preserving the current state. The new solution uses an environment variable to allow the default to be manipulated.


jalalium (approved):
Well done! Great initiative!

PR #2184, opened 2025-01-27
Add support for dict.keys

### LINKED ISSUE #1882: [DICT] Add Python support for keys
### Describe the feature
Syntactic and semantic support should be added for `keys`. This should be tested by implementing Python printing

### Test Code
```python
def foo():
    a = {1,1.0, 2:2.0}
    b : list[int] = []
    for bi in a.keys():
        b.append(bi)
    return b
```

EmilyBourne:
Add support for dict.keys. Fixes #1882


jalalium (approved):
Well done!

PR #2188, opened 2025-01-29
Remove pickle

### LINKED ISSUE #2187: Remove pickle
Pickle is currently used to save the AST for files. The aim of this is to speed up the translation. However pickle is only called after the syntactic stage.

A test with openmp.pyh shows negligible time differences between executing with and without pickle.

This is because the pickling is not currently carried out correctly. Pickle errors are caught here:
https://github.com/pyccel/pyccel/blob/35353851088955d5ecc0fc4cf75f41f273866b30/pyccel/parser/base.py#L549

which means that all but the most critical pickle errors are ignored. As a result we are not actually using the pickled version of the file.

The problem can be fixed by adding `__getnewargs__` and `__getnewargs_ex__` functions but a lot of these functions are missing currently as the whole AST is dumped.

After #297 and #641 we will be generating stub files. It may be useful to pickle these stub files which represent a reduced sub-set of the AST:
- `FunctionDef`
- `FunctionDefArgument`
- `FunctionDefResult`
- `Scope`
- `Variable`
- `ClassDef`
- `Interface`
- annotations
- ...

but this should be timed to check the benefits.

Until then it is not helpful to add `__getnewargs__` functions which will quickly become obsolete, nor is it helpful to have large amounts of code which do not lead to any real gain. We should therefore remove the use of pickle for now.

EmilyBourne:
Remove all use of pickling functions. Fixes #2187


jalalium (approved):
Well done!

PR #2191, opened 2025-02-03
Fix missing pointer assignment error

### LINKED ISSUE #2190: Incorrect detection of pointer assignment
### Describe the bug
When a list is assigned to a new variable this new variable is correctly identified as a pointer to the existing list. However if that variable has already been defined as a non-pointer an error should be raised as the type would need to change. Currently the assignment is changed to an elementwise assignment which assumes that the lhs is already assigned and has the same size as the rhs

### To Reproduce
```python
if __name__ == '__main__':
    a = [2,5]
    b = [3]
    b = a
    c = a
    print(a)
```

### Error details
```fortran
program prog_prog_tmp_lst

  use tmp_lst
  use Vector_integer8_mod

  implicit none

  type(Vector_integer8), target :: a
  type(Vector_integer8) :: b
  type(Vector_integer8), pointer :: c
  integer(i64) :: i
  integer(i64) :: i_0001

  a = Vector_integer8([2_i64, 5_i64])
  b = Vector_integer8([3_i64])
  call b % reserve(a % size())
  do i = 0_i64, b % size() - 1_i64
    b%of(i + 1_i64) = a%of(i + 1_i64)
  end do
  c => a
  do i_0001 = 0_i64, a % size() - 1_i64 - 1_i64
    write(stdout, '(I0, A)', advance="no") a%of(i_0001 + 1_i64) , ' '
  end do
        ) , ']'

end program prog_prog_tmp_lst
```
```c
#include <inttypes.h>
#include <stdint.h>
#include <stdio.h>
#include <stdlib.h>
#include "tmp_lst.h"
#ifndef _TOOLS_LIST_VEC_INT64_T
#define _TOOLS_LIST_VEC_INT64_T
#define i_type vec_int64_t
#define i_key int64_t
#define i_use_cmp
#define i_more
#include <stc/vec.h>
#include <STC_Extensions/List_extensions.h>
#endif // _TOOLS_LIST_VEC_INT64_T

int main()
{
    vec_int64_t a = {0};
    vec_int64_t b = {0};
    vec_int64_t* c = {0};
    int64_t i;
    int64_t i_0001;
    a = c_init(vec_int64_t, {INT64_C(2), INT64_C(5)});
    b = c_init(vec_int64_t, {INT64_C(3)});
    vec_int64_t_drop(&b);
    vec_int64_t_reserve(&b, vec_int64_t_size(&a));
    for (i = INT64_C(0); i < vec_int64_t_size(&b); i += INT64_C(1))
    {
        (*vec_int64_t_at_mut(&b,i)) = (*vec_int64_t_at(&a,i));
    }
    c = &a;
    printf("[");
    for (i_0001 = INT64_C(0); i_0001 < vec_int64_t_size(&a) - INT64_C(1); i_0001 += INT64_C(1))
    {
        printf("%"PRId64" ", (*vec_int64_t_at(&a,i_0001)));
    }
    printf("%"PRId64"]\n", (*vec_int64_t_at(&a,vec_int64_t_size(&a) - INT64_C(1))));
    vec_int64_t_drop(&a);
    vec_int64_t_drop(&b);
    return 0;
}
```

### Expected behavior
The behaviour should be the same as for:
```python
import numpy as np
if __name__ == '__main__':
    a = np.array([2,5])
    b = np.array([3])
    b = a
    c = a
    print(a)
```
Ie the following error should be raised:
```
pyccel:
 |error [semantic]: tmp_lst.py [9,4]| Arrays which own their data cannot become views on other arrays (b)
```

### Language
Semantic error affecting all languages

### Additional context
Detected by @jalalium in #2107

EmilyBourne:
Fix condition leading to error about pointing to data with a non-pointer object. Fixes #2190


jalalium (approved):
Well done!

PR #2196, opened 2025-02-06
Fix string comparison

### LINKED ISSUE #2195: Can't compare strings
### Describe the bug
Strings cannot be compared as the comparison triggers non-implemented `ContainerType` methods.

### To Reproduce
```python
ef f():
    option = 'hello'
    if option == 'do this':
        return 1.0
    else:
        return 2.0
```

### Error details
```bash/fortran/c
AttributeError: 'PrimitiveCharacterType' object has no attribute 'switch_basic_type'
```

### Expected behaviour
String comparison should work like integer comparison or floating point comparison.

### Language
Semantic error implies all languages.

### Proposed Solution
Stop `StringType` inheriting from `ContainerType`. This inheritance is not helpful and leads to many exceptions.

### Additional context
Related to #2194

EmilyBourne:
Fix string comparisons. Recurrent problems due to `StringType` inheritance are fixed by removing the inheritance of `ContainerType`. `CStrStr` is used with `strcmp` to implement comparison of strings in C. Fixes #2195.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -113,6 +113,7 @@ All notable changes to this project will be documented in this file.
 -   #2149 : Fix multi-line expressions in `if` conditions.
 -   #2181 : Allow saving an array result of a function to a slice but raise a warning about suboptimal performance.
 -   #2190 : Fix missing error for list pointer assignment.
+-   #2195 : Fix string comparisons.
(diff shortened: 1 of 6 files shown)


jalalium (approved):
Well done!

PR #2200, opened 2025-02-06
Allow passing strings as arguments

### LINKED ISSUE #2194: Add support for strings as arguments
### Describe the feature
It should be possible to pass strings as arguments to functions

### Test Code
```python
def f(option : str):
    if option == 'do this':
        return 1.0
    else:
        return 2.0
```

### Proposed Solution
Most of the tools necessary for this should already be implemented. The main missing piece is the wrapping

### Additional context
This feature was requested by @spossann for Struphy

EmilyBourne:
Add the necessary wrappers to add support for strings as arguments. Fixes #2194 .
Add support for the built-in function `str` so it can be used as a type annotation.
Correct the type of `s[0]` when `s` is a string (should be string, previously char).
Add basic tests.


EmilyBourne:
Marking PR as blocked as I am splitting it in 2. The changes to the wrapper will now go in a dedicated PR.

pyccel/codegen/printing/fcode.py
```diff
@@ -3125,7 +3130,7 @@ def _print_PyccelPow(self, expr):
 
     def _print_PyccelAdd(self, expr):
         if isinstance(expr.dtype, StringType):
-            return '//'.join('trim('+self._print(a)+')' for a in expr.args)
+            return ' // '.join(self._print(a) for a in expr.args)
```
yguclu, line 3133: Was `trim` necessary before this PR?

tests/epyccel/modules/strings_module.py
```diff
@@ -47,3 +49,12 @@ def concatenate_expr():
...
     return v
+
+def string_function_call():
+    s = 'hello'
+    t = str(s)
+    return t
+
+def string_function_call_on_literal():
+    t = str('hello')
+    return t
```
yguclu, line 60: What happens if we return the output of `str` directly,  without the temporary variable `t`?

pyccel/codegen/wrapper/c_to_python_wrapper.py
```diff
@@ -2562,6 +2565,87 @@ def _extract_HomogeneousSetType_FunctionDefArgument(self, orig_var, collect_arg,
...
+        arg_var : Variable | IndexedElement, optional
+            A variable or an element of the variable representing the argument that
+            will be passed to the low-level function call.
+
+        Returns
+        -------
+        list[PyccelAstNode]
+            A list of expressions which extract the argument from collect_arg into arg_var.
+        """
+        assert not bound_argument
```
yguclu, line 2606: I find this more readable and safer (because it does not rely on `__bool__`):
```suggestion
        assert bound_argument is False
```

yguclu:
I have just noticed that there is a conflict in the Fortran-to-C wrapper which should be resolved.

pyccel/codegen/printing/fcode.py
EmilyBourne: No. Trim was used in the original implementation written by Ahmed and Said. This implementation declared strings as fixed length objects so trim was necessary to avoid an excessive number of spaces. However it had the downside of preventing us from doing:
```python
a = 'hello '
b = 'world'
c = a + b
```
with the old implementation with `trim`, `c`would have contained `'helloworld'` instead of `'hello world'`.

The declaration of strings was changed in #2063 ( 6a1c01b888 ) they are now allocatables so the length matches the Python length and `trim` serves no purpose

docs/type_annotations.md
```diff
@@ -105,9 +105,11 @@ Strings are not yet supported as keys in Fortran.
 
 ## Strings
 
-Pyccel contains very minimal support for strings. This is mostly provided to allow the use of strings as keys of dictionaries.
-
-Currently strings can be used as local variables and can be returned from functions.
+Pyccel contains very minimal support for strings. This is mostly provided to allow us to implement strings as keys of dictionaries.
```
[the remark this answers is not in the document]
EmilyBourne: Done [0fef055](https://github.com/pyccel/pyccel/pull/2200/commits/0fef0555c9ddc526f30dd7c0351bc510056b5657)

pyccel/ast/builtins.py
```diff
@@ -1724,6 +1725,36 @@ class PythonIsInstance(PyccelFunction):
     def __init__(self, obj, class_or_tuple):
         super().__init__(obj, class_or_tuple)
 
+#==============================================================================
+class PythonString(PyccelFunction):
```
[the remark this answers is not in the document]
EmilyBourne: Fixed [4f9b756](https://github.com/pyccel/pyccel/pull/2200/commits/4f9b7568468fcdce69b9c55066e8f1c99068f7f7)

pyccel/codegen/wrapper/c_to_python_wrapper.py
EmilyBourne: Fixed [570b49f](https://github.com/pyccel/pyccel/pull/2200/commits/570b49f74b6374cd36f935095c47e29d216051c3)

tests/epyccel/modules/strings_module.py
EmilyBourne: I have added a test: [6a0d593](https://github.com/pyccel/pyccel/pull/2200/commits/6a0d5933155563c349ba2a59f750e04caf014791)

It works as expected. In C:
```c
return cstr_lit("hello");
```
In Fortran:
```fortran
result_0001 = 'hello'
```

EmilyBourne:
> I have just noticed that there is a conflict in the Fortran-to-C wrapper which should be resolved.

Fixed

yguclu (approved):

PR #2201, opened 2025-02-08
Improve error handling

EmilyBourne:
Add a `try`/`except` block in `SyntacticParser._visit`, `SemanticParser._visit`, `CodegenPrinter._print`, and `Wrapper._wrap` to catch any internal errors. Use context information to try to provide an accurate line number. Internal errors arise due to bugs in Pyccel, but also if user code does not execute properly. Tracebacks should be avoided in both these cases, but especially in the latter. The behaviour seen on the `devel` branch can be recovered using the `--developer-mode` flag.

It is extremely difficult to test this kind of error as bugs that would lead to this error should generally be fixed. Nonetheless a test is added for the semantic stage using code that does not execute correctly in Python.


jalalium (approved):
Well done!

PR #2207, opened 2025-02-10
Fix returning an array of unknown literal size

### LINKED ISSUE #2206: Returning an array of unknown literal size doesn't work
### Describe the bug
Returning an array allocated in an if condition leads to Pyccel assuming the wrong size in the wrapper.

### To Reproduce
```python
import numpy as np

def q():
    b = False
    if b:
        return np.ones(1)
    else:
        return np.ones(3)
```

### Error details
```fortran
  subroutine bind_c_q(bound_Out_0001, Out_0001_shape_1) bind(c)

    implicit none

    type(c_ptr), intent(out) :: bound_Out_0001
    integer(i32), intent(out) :: Out_0001_shape_1
    real(f64), allocatable :: Out_0001(:)
    real(f64), pointer :: Out_0001_ptr(:)
    real(f64) :: Out_0001_elem
    integer(i64) :: i

    call q(Out_0001 = Out_0001)
    Out_0001_shape_1 = 1_i64
    bound_Out_0001 = c_malloc(storage_size(Out_0001_elem, kind = &
          c_size_t) * Out_0001_shape_1)
    call C_F_Pointer(bound_Out_0001, Out_0001_ptr, [Out_0001_shape_1])
    do i = 0_i64, size(Out_0001_ptr, kind=i64) - 1_i64
      Out_0001_ptr(i) = Out_0001(0_i64)
    end do

  end subroutine bind_c_q
```

### Expected behavior
The shape should be unknown and therefore deduced from the output

### Language
All

### Additional context
This problem used to be handled with `pyccel.ast.variable.Variable.shape_can_change`.
This problem was detected in #2179

EmilyBourne:
Fix returning an array of unknown literal size. This was broken due to a correct shape but incorrect `alloc_shape`. Fixes #2206

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -114,6 +114,7 @@ All notable changes to this project will be documented in this file.
 -   #2181 : Allow saving an array result of a function to a slice but raise a warning about suboptimal performance.
 -   #2190 : Fix missing error for list pointer assignment.
 -   Lifted the restriction on ndarrays limiting them to rank<15.
+-   #2206 : Fix returning an array of unknown literal size.
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


jalalium (approved):
Well done!

PR #2208, opened 2025-02-11
Add compile time analysis of if blocks

### LINKED ISSUE #2183: Add compile time analysis of if blocks
### Relevant Discussion

Discussed in meeting 

### Describe the feature
If an `if` block has a condition which is always `True` then subsequent `elif` blocks can result in warnings from the compiler about inaccessible code. We would like to use the functions in `ast/sympy_helper.py` to check if a condition is always true or always false.
`SemanticParser._visit_If` should then filter the `IfSection`s based on if the condition is `LiteralTrue`, `LiteralFalse` or a runtime condition discarding inaccessible code. This may result in the code from one if block being returned directly.

### Test Code
```python
def f(a : 'int[:] | int[:,:]'):
    if len(a.shape) == 1:
        print("rank one")
        print(a[0])
    else:
        print("rank two")
        print(a[0,0])
```

### Proposed Solution
- [ ] use `pyccel_to_sympy` to change `IfSection` conditions to `LiteralTrue`/`LiteralFalse` where appropriate
- [x] modify `SemanticParser._visit_If` to filter IfSections

EmilyBourne:
Add compile time analysis of if block conditions to allow differentiating code based on the type or rank. Fixes #2183

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,7 @@ All notable changes to this project will be documented in this file.
 -   #983 : Add support for built-in function `round`.
 -   Add support for `type` as a type annotation.
 -   #2182 : Add support for `isinstance`.
+-   #2183 : Add compile time analysis of if block conditions.
 -   \[INTERNALS\] Add abstract class `SetMethod` to handle calls to various set methods.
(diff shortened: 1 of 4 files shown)


jalalium (approved):
Well done!

PR #2211, opened 2025-02-12
Add missing libraries in intel installation

### LINKED ISSUE #2166: Intel tests no longer find blas/lapack
### Describe the bug
The Intel CI tests using blas/lapack are failing as they are no longer able to locate the libraries.  Probably a step is missing from the installation

EmilyBourne:
Install BLAS and LAPACK for the intel installation (ideally we should use MKL). Fixes #2166. Mark test as passing as #1540 is now fixed. Run `test_combined_for_simd` in non-debug mode to avoid intel bug.


yguclu (approved):
Looks good to me!

PR #2215, opened 2025-02-15
Fix returns from inline functions

### LINKED ISSUE #2214: Returning a local variable from an inline function doesn't work
### Describe the bug
Returning a local variable from an inline function doesn't work

### To Reproduce
```python
def f():
    @inline
    def tmp():
        a = 1
        return a

    b = tmp()
    c = tmp()
    d = tmp() + 3
    e = tmp() * 4
    return b,c,d,e
```

### Error details
```c
int64_t f(int64_t* b, int64_t* c, int64_t* d, int64_t* e)
{
    a = INT64_C(1);
    (*b) = a;
    a = INT64_C(1);
    (*c) = a;
    a = INT64_C(1);
    (*d) = a + INT64_C(3);
    a = INT64_C(1);
    (*e) = a * INT64_C(4);
    return 0;
}
```
```fortran
  subroutine f(b, c, d, e)

    implicit none

    integer(i64), intent(out) :: b
    integer(i64), intent(out) :: c
    integer(i64), intent(out) :: d
    integer(i64), intent(out) :: e

    a = 1_i64
    b = a
    a = 1_i64
    c = a
    a = 1_i64
    d = a + 3_i64
    a = 1_i64
    e = a * 4_i64
    return

  end subroutine f
```

### Expected behaviour
The necessary variables should be allocated

### Language
Fortran and C

### Additional context
Blocking #2179

EmilyBourne:
Fix returning a local variable from an inline function. Fixes #2214

### DIFF
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -616,12 +616,15 @@ class CCodePrinter(CodePrinter):
         func_result_vars = [r.var for r in func.results]
         generated_result_vars = any(not v.is_temp for v in func_result_vars)
         if generated_result_vars:
-            non_temp_func_result_vars = [v for v in func_result_vars if not v.is_temp]
-            result_vars = [r.clone(name = self.scope.get_new_name(r.name)) \
-                        for r in func_result_vars if not r.is_temp]
-            for v in result_vars:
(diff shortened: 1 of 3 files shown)


jalalium (approved):
Well done!

PR #2219, opened 2025-02-17
Add support for `__getitem__`

jalalium:
This PR introduces support for the magic method `__getitem__` . Previously, while basic translation of class methods worked, objects were not recognized as subscriptable because the mapping slot (mp_subscript) in the type object was not properly configured.


pyccel/codegen/printing/cwrappercode.py
```diff
@@ -439,6 +439,18 @@ def _print_PyClassDef(self, expr):
...
+        if '__getitem__' in magic_methods:
+            getitem_wrapper_def = (
+                "static PyObject* __pyccel_getitem_wrapper(PyObject* self, PyObject* key) {\n"
+                "    /* Pack the key into a tuple and call the original __getitem__ wrapper */\n"
+                "    PyObject* args = PyTuple_Pack(1, key);\n"
+                f"    PyObject* result = {magic_methods['__getitem__'].name}((struct PyAObject*)self, args, NULL);\n"
+                "    Py_DECREF(args);\n"
+                "    return result;\n"
+                "}\n"
+            )
```
EmilyBourne, line 452: This looks like a much faster way of implementing the wrapper than what is done for the other magic methods. But won't this lead to significantly more calculations being done by the wrapper?

```c
static PyObject* __pyccel_getitem_wrapper(PyObject* self, PyObject* key) {
    /* Pack the key into a tuple and call the original __getitem__ wrapper */
    PyObject* args = PyTuple_Pack(1, key);
    PyObject* result = bind_c_a_operator_getitem_wrapper((struct PyAObject*)self, args, NULL);
    Py_DECREF(args);
    return result;
}

static PyObject* bind_c_a_operator_getitem_wrapper(struct PyAObject* self_0001, PyObject* args, PyObject* kwargs)
{
    PyObject* i_obj;
    void* self = NULL;
    int64_t i;
    int64_t Out_0001;
    PyObject* Out_0001_obj; 
    static char *kwlist[] = {
        (char*)"i",
        NULL
    }; 
    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "O", kwlist, &i_obj))
    {   
        return NULL;
    }
    self = (struct A*)(self_0001->instance);
    if (!PyIs_NativeInt(i_obj))
    {
        PyErr_SetString(PyExc_TypeError, "Expected an argument of type int for argument i");
        return NULL;
    } 
    i = PyInt64_to_Int64(i_obj);
    Out_0001 = bind_c_a_operator_getitem(self, i);
    Out_0001_obj = Int64_to_NumpyLong(&Out_0001);
    return Out_0001_obj;
}
```
instead of:
```c
static PyObject* __pyccel_getitem_wrapper(struct PyAObject* self, PyObject* i_obj) {
    void* self = NULL;
    int64_t i;
    int64_t Out_0001;
    PyObject* Out_0001_obj;
    self = (struct A*)(self_0001->instance);
    if (!PyIs_NativeInt(i_obj))
    {
        PyErr_SetString(PyExc_TypeError, "Expected an argument of type int for argument i");
        return NULL;
    }
    i = PyInt64_to_Int64(i_obj);
    Out_0001 = bind_c_a_operator_getitem(self, i);
    Out_0001_obj = Int64_to_NumpyLong(&Out_0001);
    return Out_0001_obj;
}
```

jalalium: I think that the extra function call and tuple packing add some overhead, but in practice it's negligible compared to the overall operation cost. Compiler optimizations may inline these calls, and the added flexibility of using a consistent argument parsing mechanism outweighs the slight performance hit. Maybe I can benchmark the two versions?

EmilyBourne: I think it is worth benchmarking because `__getitem__` is rarely linked to a computationally expensive operation. If it is simply indexing an array (as in your example) I would not expect an extra function call, tuple packing, and a call to PyArg_ParseTupleAndKeywords to be negligible. 

In particular the call to `PyArg_ParseTupleAndKeywords` (https://github.com/python/cpython/blob/b93b7e566e5a4efe7f077af2083140e50bd2b08f/Python/getargs.c#L1276) internally calls a rather large function (https://github.com/python/cpython/blob/b93b7e566e5a4efe7f077af2083140e50bd2b08f/Python/getargs.c#L1513) which looks non-trivial. I'm not convinced a compiler can optimise round this function.

> the added flexibility of using a consistent argument parsing mechanism

My example (matching what is donc for the numeric methods) uses the same argument parsing mechanisms. The only thing that changes is that what is usually an output of `PyArg_ParseTupleAndKeywords` becomes the function arguments.

The only things necessary to handle this in the wrapper are:
https://github.com/pyccel/pyccel/blob/b3ad4a99e6e4e6a05e0ba9c76bbcdabb5ead11f1/pyccel/codegen/wrapper/c_to_python_wrapper.py#L1501-L1503
https://github.com/pyccel/pyccel/blob/b3ad4a99e6e4e6a05e0ba9c76bbcdabb5ead11f1/pyccel/codegen/wrapper/c_to_python_wrapper.py#L1512-L1518
https://github.com/pyccel/pyccel/blob/b3ad4a99e6e4e6a05e0ba9c76bbcdabb5ead11f1/pyccel/codegen/wrapper/c_to_python_wrapper.py#L2053-L2054

pyccel/codegen/wrapper/c_to_python_wrapper.py
```diff
@@ -1498,7 +1498,7 @@ def _wrap_FunctionDef(self, expr):
             func_args = [FunctionDefArgument(a) for a in func_args]
             body = []
         else:
-            if in_interface or original_func_name in magic_binary_funcs or original_func_name == '__len__':
+            if in_interface or original_func_name in magic_binary_funcs or original_func_name in ('__len__', '__getitem__'):
```
EmilyBourne, line 1501: Isn't `mp_subscript` a binary function?

jalalium: Sorry, I just got what you meant.
[fixed](https://github.com/pyccel/pyccel/pull/2219/commits/5e755a8574dbd850e45d3055bb66a5f3590261c8)

pyccel/codegen/wrapper/c_to_python_wrapper.py
```diff
@@ -1498,7 +1499,7 @@ def _wrap_FunctionDef(self, expr):
             func_args = [FunctionDefArgument(a) for a in func_args]
             body = []
         else:
-            if in_interface or original_func_name in magic_binary_funcs or original_func_name == '__len__':
+            if in_interface or original_func_name in magic_binary_funcs or original_func_name in ('__len__'):
```
EmilyBourne, line 1502: Bad change
```suggestion
            if in_interface or original_func_name in magic_binary_funcs or original_func_name == '__len__':
```

EmilyBourne (approved):
This all looks good to me

EmilyBourne (approved):

PR #2221, opened 2025-02-18
Fix premature `stc/cspan` import

### LINKED ISSUE #2220: Misordered imports
### Describe the bug
The C standard imports appear after the STC `cspan` import. This may cause problems with integers.

### To Reproduce
```python
def f():
    import numpy as np
    a = np.array([1,2,3])
```

### Error details
```c
#ifndef TMP_ARR_H
#define TMP_ARR_H

#define STC_CSPAN_INDEX_TYPE int64_t
#include <stc/cspan.h>
#ifndef _ARRAY_INT64_1D
#define _ARRAY_INT64_1D
using_cspan(array_int64_1d, int64_t, 1);
#endif // _ARRAY_INT64_1D

#include <stdint.h>
#include <stdlib.h>



void f(void);
#endif // tmp_arr_H
```

### Expected behaviour
`stdint.h` should be imported before it is used by cspan.

### Language
C

### Additional context
This may be related to the problems in #2211

EmilyBourne:
Fix premature `stc/cspan` import. Fixes #2220

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -123,6 +123,7 @@ All notable changes to this project will be documented in this file.
 -   Catch all internal errors arising in the syntactic, semantic, printing or code generation stages.
 -   #2206 : Fix returning an array of unknown literal size.
 -   #2112 : Improve floor division.
+-   #2220 : Fix premature `stc/cspan` import.
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


yguclu (approved):
Looks good

PR #2227, opened 2025-02-20
Correct error message formatting for invalid type annotation

emmanuel-ferdman:
# PR Summary
Small PR - Use f-string for error message in `SyntacticTypeAnnotation` to print the data type.

Before:
```
ValueError: Syntactic datatypes should be strings not {type(dtype)}
```
After:
```
ValueError: Syntactic datatypes should be strings not <class 'int'>
```

### DIFF
--- a/pyccel/ast/type_annotations.py
+++ b/pyccel/ast/type_annotations.py
@@ -274,7 +274,7 @@ class SyntacticTypeAnnotation(PyccelAstNode):
 
     def __init__(self, dtype, order = None):
         if not isinstance(dtype, (str, DottedName, IndexedElement)):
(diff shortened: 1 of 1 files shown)


emmanuel-ferdman:
I believe this change is too minor to include in the changelog, but let me know if you'd prefer me to add it.

EmilyBourne (approved):

emmanuel-ferdman:
@EmilyBourne No, actually, I was reviewing the source code to get a better understanding of the project and happened to come across it by accident. These kinds of bugs are the ones humans aren’t great at spotting. But I’ve encountered similar issues in the past - it's a shame that tools like Pylint can’t catch them.

PR #2239, opened 2025-03-13
Specify the libdirs for a shared Python library

EmilyBourne:
Specify the libdirs for a shared Python library. This ensures that the library will be found during execution if the compilation succeeds.


yguclu (approved):
Looks good! Thanks!

yguclu (approved):

PR #2244, opened 2025-03-15
Fix minor bugs on the devel branch

EmilyBourne:
Fix two minor bugs recently introduced on the devel branch:
- Intel tests are failing due to a double import of `C_BOOL`. This is fixed by moving the import in `pyccel/stdlib/gFTL_functions/Set_extensions.inc` from the module scope to the function scope.
- Windows tests are sometimes failing due to overflow errors. This is fixed by limiting the range of integers used in the test to a range that returns valid results


jalalium (approved):
Well done!

PR #2246, opened 2025-03-16
Fix underflow errors on devel

EmilyBourne:
Windows tests are sometimes failing due to underflow errors. This is fixed by limiting the range of integers used in the test to a range that returns valid results. This was already fixed for overflow errors but `min_int` is a negative number not a small value as for `min_float`.


jalalium (approved):
Well done!

PR #2252, opened 2025-03-18
Describe multiple compilers in the same JSON file

### LINKED ISSUE #2242: Describe multiple compilers in the same JSON file
### Problem

When calling Pyccel with `--language=fortran`, two different compilers are used: in addition to the Fortran compiler, the C wrapper requires a C compiler.

If the user wants to use two unsupported compilers, or specify custom flags for both compilers, he cannot do that through a JSON file (as described [here](https://github.com/pyccel/pyccel/blob/devel/docs/compiler.md#using-an-unsupported-compiler-or-custom-compiler-flags)) because the JSON file only contains information about one compiler. 

### Example Code

Let's say that we want to accelerate a trivial `hello.py` file:
```python
def hello_world(n : int):
    for i in range(n):
        print("Hello, world!")
```

Currently, when using a JSON file with Fortran compiler specs, we get this warning:

```
$ pyccel --compiler=gnu --language=fortran --export-compile-info custom.json
$ pyccel --compiler=custom.json --language=fortran hello.py
[PATH/TO/PYCCEL]/codegen/compiling/compilers.py:96: UserWarning: Language does not match compiler. Using GNU compiler
  warnings.warn(UserWarning("Language does not match compiler. Using GNU compiler"))
```

### Proposed Solution

We should allow JSON files with information about more than one compiler. The compilers' languages must all be different from each other, and one of these languages must be C.

### Notes

This is also useful when running unit tests, which might be using different target languages, as it allows one to provide a single JSON file and then run, for example:

```
$ export PYCCEL_DEFAULT_COMPILER=custom.json
$ pytest pyccel/tests/
```

EmilyBourne:
* Describe multiple compilers in the same JSON file. Fixes #2242
* Let a JSON file define a compiler family with the same name as the file
* Rename parameter in `execute_pyccel` function: `compiler` -> `compiler_family`
* Remove `language` and `family` entries from dictionaries describing the default compilers 
* Update documentation about customization of compiler and use of `PYCCEL_DEFAULT_COMPILER` environment variable

By grouping the information by compiler family we remove the need to have 2 Compiler objects, but need to specify the target language in the compile command.


yguclu (commented):
Looks good. I have a few questions

pyccel/compilers/default_compilers.py
```diff
@@ -290,13 +282,13 @@ def change_to_lib_flag(lib):
...
-available_compilers = {('GNU', 'c') : gcc_info,
-                       ('GNU', 'fortran') : gfort_info,
-                       ('intel', 'c') : icc_info,
-                       ('intel', 'fortran') : ifort_info,
-                       ('PGI', 'c') : pgcc_info,
-                       ('PGI', 'fortran') : pgfortran_info,
-                       ('nvidia', 'c') : nvc_info,
-                       ('nvidia', 'fortran') : nvfort_info}
+available_compilers = {'GNU': {'c' : gcc_info,
+                               'fortran' : gfort_info},
```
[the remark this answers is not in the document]
EmilyBourne: Hmm, I removed `language` because it was already present in the output. I actually don't think the `"family"` key is used. But in the future we could imagine having a folder, e.g. `.pyccel` that we automatically check for configs. In this case we would need a way to differentiate between them so the "family" key could be useful. That said having it inside the language config is certainly a bit odd

docs/compiler.md
```diff
@@ -50,11 +49,11 @@ Python is considered to be an accelerator and must additionally specify `shared_
 The default compilers can provide examples compatible with your system once Pyccel has been executed at least. To export the JSON file describing your setup, use the `--export-compile-info` flag and provide a target file name.
 E.g.
 ```shell
-pyccel --compiler=intel --language=c --export-compile-info=icc.json
+pyccel --compiler=intel --export-compile-info=intel.json
 ```
 once this file has been modified it can then be used with:
 ```shell
-pyccel --compiler=icc.json --language=c <file_to_translate>
+pyccel --compiler=intel.json <file_to_translate>
```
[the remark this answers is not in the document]
EmilyBourne: Yes the `--language` flag now behaves as it usually does when using the `pyccel` command

docs/compiler.md
```diff
@@ -50,14 +49,14 @@ Python is considered to be an accelerator and must additionally specify `shared_
...
```
yguclu, line 56: I think we should mention that the environment variable `PYCCEL_DEFAULT_COMPILER` can be used to specify the path to the JSON file. This is not obvious just by reading the section "Specifying a compiler".

pyccel/codegen/compiling/compilers.py
```diff
@@ -144,55 +138,74 @@ def _get_exec(self, accelerators):
...
+
+        Returns
+        -------
+        iterable[str]
+            An iterable containing the relevant information from the
+            requested property.
         """
         # Use dict keys as an ordered set
-        prop = dict.fromkeys(prop)
+        properties = dict.fromkeys(properties)
```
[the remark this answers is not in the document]
EmilyBourne: Clearer: ac67023ba ?

docs/compiler.md
EmilyBourne: Better : 40af93c24 ?

pyccel/codegen/compiling/compilers.py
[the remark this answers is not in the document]
yguclu: The example is very clear, thanks!

docs/compiler.md
yguclu: Do you want me to do it?

EmilyBourne: If you have a suggestion then feel free to add it :+1:

EmilyBourne: The flags should override the environment variable

https://github.com/pyccel/pyccel/blob/bbff872ce78c57d2c91d1895ca154ba8fe3b7d1a/pyccel/codegen/pipeline.py#L186-L187

yguclu: Thanks, I will mention this

yguclu: What do you think? [d23bed8](https://github.com/pyccel/pyccel/pull/2252/commits/d23bed8d330b848d8453e6b355f4f034497fbebb)

EmilyBourne: I think I misunderstood your question about flags. I thought you were talking about `--X` options not the `--flags` option. A compiler definition does not contain a `flags` field. It has `general_flags`, `release_flags` and `debug_flags`. The flags used in the compilations are `general_flags`, one of `release_flags` and `debug_flags`, and any additional flags passed using the `--flags` option.

yguclu: No, you understood correctly what I meant. I was talking about any option that is also contained in the JSON file. But I was sloppy in my explanation, as you rightly pointed out things are more complicated

EmilyBourne: Corrected 
[583c6f6](https://github.com/pyccel/pyccel/pull/2252/commits/583c6f6b1f6e39468ed4bab4e8b58331998694b2)

yguclu: Thanks!

pyccel/compilers/default_compilers.py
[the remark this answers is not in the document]
yguclu: This is similar to GitHub Actions, where `build_docs.yml` corresponds to the action `build_docs`.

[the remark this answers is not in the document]
EmilyBourne: That sounds like a good idea :+1:

PR #2259, opened 2025-03-24
Add C and Fortran support for list method `insert`

jalalium:
This PR adds C and Fortran printers for list method `insert` and activates the relevant tests

### DIFF
--- a/docs/builtin-functions.md
+++ b/docs/builtin-functions.md
@@ -84,7 +84,7 @@ Python contains a limited number of builtin functions defined [here](https://doc
 | `count` | No |
 | `extend` | Python-only |
 | `index` | No |
-| `insert` | Python-only |
+| **`insert`** | **Yes** |
 | `max` | No |
 | `min` | No |
 | **`pop`** | **Yes** |
--- a/pyccel/stdlib/STC_Extensions/List_extensions.h
+++ b/pyccel/stdlib/STC_Extensions/List_extensions.h
@@ -43,7 +43,26 @@ static inline i_key _c_MEMB(_max)(const Self* self) {
     }
     return max_val;
 }
+
 #endif
+
+// Function to insert value before position idx
+static inline void _c_MEMB(_insert)(Self* self, intptr_t idx, i_key value) {
+    size_t n = _c_MEMB(_size)(self);
+
+    if (idx < 0) idx += (intptr_t)n;
+    if (idx < 0) idx = 0;
+
+    if ((size_t)idx >= n) {
+        _c_MEMB(_push_back)(self, value);
+    }
+    else {
+        _c_MEMB(_insert_at)(self,
+            _c_MEMB(_advance)(_c_MEMB(_begin)(self), idx),
+            value);
+    }
+}
+
 #undef i_type
 #undef i_key
 #undef i_use_cmp
--- a/pyccel/stdlib/gFTL_functions/Vector_extensions.inc
+++ b/pyccel/stdlib/gFTL_functions/Vector_extensions.inc
@@ -56,4 +56,28 @@ contains
     end do
 
   end function __IDENTITY(Vector)_max
+
+  subroutine __IDENTITY(Vector)_insert(my_vector, idx, value)
+    class(Vector), intent(inout)              :: my_vector
+    integer(kind=GFTL_SIZE_KIND), intent(in)  :: idx
+    __T_declare_dummy__, intent(in)           :: value
+  
+    integer :: n, pos, i
+  
+    n = my_vector%size()
+    pos = idx
+  
+    if (pos < 0) pos = pos + n
+    if (pos < 0) pos = 0
+    if (pos > n) pos = n
+  
+    call my_vector%resize(n + 1)
+  
+    do i = n + 1, pos + 1, -1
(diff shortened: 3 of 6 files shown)


docs/builtin-functions.md
```diff
@@ -84,7 +84,7 @@ Python contains a limited number of builtin functions defined [here](https://doc
 | `count` | No |
 | `extend` | Python-only |
```
EmilyBourne, line 85: By the way, unrelated to this PR but isn't `extend` fully implemented? I thought it worked by translating itself into multiple `append` calls in the semantic stage. If that is true, and it works for Python, and `append` works for all languages, doesn't that mean that `extend` should also be available?

jalalium: Yes, I noticed that. It works through `append` and takes either containers or generators as arguments. However, I think if the  argument is another list, a dedicated function should be used to copy the whole list at once to avoid any performance losses.

EmilyBourne: Ah ok. Please create an issue so we remember that this is why there is a difference

jalalium: #2260

pyccel/stdlib/gFTL_functions/Vector_extensions.inc
```diff
@@ -77,4 +77,29 @@ contains
...
+    class(Vector), intent(inout)              :: my_vector
+    integer(i64), intent(in)                  :: idx
+    __T_declare_dummy__, intent(in)           :: value
+  
+    integer(i64) :: n, pos, i
+  
+    n = my_vector%size()
+    pos = idx
+  
+    if (pos < 0) pos = pos + n
```
EmilyBourne, line 92: Why write a new method instead of using https://github.com/Goddard-Fortran-Ecosystem/gFTL/blob/main/include/v2/vector/vector.md#insert ?

pyccel/stdlib/STC_Extensions/List_extensions.h
```diff
@@ -43,7 +43,26 @@ static inline i_key _c_MEMB(_max)(const Self* self) {
     }
     return max_val;
 }
+
 #endif
+
+// Function to insert value before position idx
+static inline void _c_MEMB(_insert)(Self* self, intptr_t idx, i_key value) {
```
EmilyBourne, line 50: Why write a new method instead of using STC's `vec_X_insert_at` ?

jalalium: because the STC function does not allow for negative indices. Do you wanna add a codeblock to calculate the appropriate index and just call their function?

pyccel/stdlib/gFTL_functions/Vector_extensions.inc
EmilyBourne: I assumed negative indices would work in the same way as for arrays

jalalium: insert is very lenient with indices. the index can be way larger than the size of the array, then it would be appended, or too small (negative) then it would be prepended to the array. maybe keep this subroutine as a wrapper around their insert?

EmilyBourne: Ah you're right! I think we even have tests for some of these awkward insides. In that case a function that fixes the indices then calls the existing function sounds perfect

EmilyBourne (approved):
This looks good to me. Although I would prefer a `index` and a `value` property in the `ListInsert` class. The order is not likely to change but I know I usually get it wrong

yguclu:
> This looks good to me. Although I would prefer a `index` and a `value` property in the `ListInsert` class. The order is not likely to change but I know I usually get it wrong

I agree with @EmilyBourne, although the same observation applies to several other `List<METHOD>` classes.

jalalium:
> > This looks good to me. Although I would prefer a `index` and a `value` property in the `ListInsert` class. The order is not likely to change but I know I usually get it wrong
> 
> I agree with @EmilyBourne, although the same observation applies to several other `List<METHOD>` classes.

I agree, I will update this class and create a separate PR to add the properties for all other classes

yguclu (changes requested):
Looks good to me. I just have two minor comments.

pyccel/ast/builtin_methods/list_methods.py
```diff
@@ -200,6 +200,24 @@ def __init__(self, list_obj, index, new_elem) -> None:
...
+        return self._args[0]
+
+    @property
+    def value(self):
+        """
+        The value to insert into the list.
+
+        The value to insert into the list.
+        """
+        return self._args[1]
```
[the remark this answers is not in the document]
jalalium: Got it! Do you suggest to rename it `object`? wouldn't that overwrite the `object` class?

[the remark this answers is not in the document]
yguclu: I had not thought of that! 😆

Given that this is just a property, I think we are safe to call it as we want though!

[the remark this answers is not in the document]
jalalium: Fixed!

yguclu (approved):

PR #2274, opened 2025-04-11
Do not pass include flags to linker

### LINKED ISSUE #2273: [Improvement] Do not pass include directories to linker when creating shared library

We currently pass `-I <INCLUDE-DIR>` to every compilation step, including the final one which creates the C Python extension module as a shared library. This flag is usually passed many times with different absolute paths, and it clutters the output when `pyccel` is used with the option `--verbose`.

### Do you have any suggestions for improvement?

There is no need to pass include directories to the linker, as all object files are already available.

### Record

yguclu:
Fix #2273 by removing `*includes` from the list of arguments to the linker in method `compile_shared_library` of class `Compiler`.

Further, in function `create_shared_library` of module `python_wrapper`, force the compiler language to C (rather than using the backend language) when building the C wrapper for NumPy arrays. This avoids failures with the Intel compilers.

### DIFF
--- a/.github/workflows/intel.yml
+++ b/.github/workflows/intel.yml
@@ -16,7 +16,7 @@ on:
         required: false
         type: string
   push:
-    branches: [devel, main]
+    branches: [devel, main, issue-2273]
 
 env:
   COMMIT: ${{ inputs.ref || github.event.ref }}
@@ -78,6 +78,11 @@ jobs:
         shell: bash
       - name: Coverage install
         uses: ./.github/actions/coverage_install
+      - name: Test ifx
+        run: |
+          pyccel c_arrays.py --verbose
+        shell: bash
+        working-directory: ./tests/pyccel/scripts
       - name: Fortran/C tests with pytest
         id: f_c_pytest
         timeout-minutes: 60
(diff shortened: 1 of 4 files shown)


yguclu:
> /bot run intel

Should we add it to the required PR tests?

EmilyBourne:
> > /bot run intel
> 
> Should we add it to the required PR tests?

I don't think we need to. It is rare that we break only intel. It should be enough to run it on the devel branch but I like to check it manually if I think it might be affected

yguclu:
@EmilyBourne Does the failure looks related to this PR?

EmilyBourne:
> @EmilyBourne Does the failure looks related to this PR?

I'm not sure. It depends on why the file is not found. It is a C file so I wouldn't expect it to be related. But maybe the dependencies ended up in the includes :thinking:

EmilyBourne:
> ifx: warning #10145: no action performed for file '/home/runner/work/pyccel/pyccel/tests/pyccel/scripts/__pyccel__/cwrapper/cwrapper.c'


This seems to be the crux of the problem. I'm not sure why `ifx` is being used for this though. It should be using a C compiler

EmilyBourne:
I found the problem. It has an easy fix so I've fixed it here. It seems this is broken in the devel branch too. I just hadn't noticed

EmilyBourne (approved):

PR #2277, opened 2025-04-11
Allow returning dictionaries from Fortran

### LINKED ISSUE #2060: [DICTS] Support dicts as results in Fortran
### Describe the feature
Support returning dicts from Fortran code

### Test Code
```python
def f():
    a = {1:1.0}
    return a
```

### Proposed Solution
The cwrapper should be improved to ensure that dicts can be read from Fortran code and passed to Python.

EmilyBourne:
Allow returning dictionaries from Fortran. Fixes #2060

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,7 +51,7 @@ All notable changes to this project will be documented in this file.
 -   #1744 : Add support for set method `intersection()`.
 -   #1745 : Add support for set method `intersection_update()`.
 -   #1745 : Add support for set method `isdisjoint()`.
--   #2059 : Add C support for returning dictionaries from functions.
+-   #2059 : Add C and Fortran support for returning dictionaries from functions.
 -   #2164 : Add support for dict indexing.
 -   #1880 : Add support for dict method `clear()`.
 -   #1884 : Add support for dict method `items()`.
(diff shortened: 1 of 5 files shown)


docs/type_annotations.md
```diff
@@ -102,6 +102,7 @@ b : dict[int,bool] = {1: False, 4: True}
 c : dict[int,complex] = {}
 ```
 Strings are not yet supported as keys in Fortran.
```
yguclu, line 104: Supporting strings as keys in Fortran sounds like a very important task, but I could not find an issue about it.

CHANGELOG.md
```diff
@@ -51,7 +51,7 @@ All notable changes to this project will be documented in this file.
 -   #1744 : Add support for set method `intersection()`.
 -   #1745 : Add support for set method `intersection_update()`.
 -   #1745 : Add support for set method `isdisjoint()`.
--   #2059 : Add C support for returning dictionaries from functions.
+-   #2059 : Add C and Fortran support for returning dictionaries from functions.
```
yguclu, line 54: How do you decide if you should mention #2059 or #2277 here?

EmilyBourne: I usually just leave whichever was mentioned first. The two issues are almost a copy/paste of one another

docs/type_annotations.md
EmilyBourne: Hmm. You're right :thinking: I'll open one

EmilyBourne: https://github.com/pyccel/pyccel/issues/2278

yguclu (approved):
Looks good to me!

PR #2281, opened 2025-04-12
Cast insert index to correct type

### LINKED ISSUE #2280: [Bug] Fortran translation of list.insert uses wrong precision for insertion index

### Describe the bug

The Fortran translation of list.insert uses the default C precision for the insertion index. This is often equal to 4, however the implementation in stdlib/ expects an integer with precision 8.

### Expected behavior

There should be a cast to the correct precision

### Add Screenshots

https://github.com/pyccel/pyccel/actions/runs/14420192148/job/40441772048

Image

### Record

EmilyBourne:
The Fortran translation of `list.insert` was using the default C precision for the insertion index. This is often equal to 4, however the implementation in `stdlib/` expects an integer with precision 8. Hence an explicit cast to 64-bit integer was inserted. This fixes #2280.

### DIFF
--- a/pyccel/codegen/printing/fcode.py
+++ b/pyccel/codegen/printing/fcode.py
@@ -1416,7 +1416,7 @@ class FCodePrinter(CodePrinter):
         target = self._print(expr.list_obj)
         type_name = self._print(expr.list_obj.class_type)
         self.add_import(self._build_gFTL_extension_module(expr.list_obj.class_type))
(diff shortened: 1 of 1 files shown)


EmilyBourne:
CHANGELOG is not modified as the Fortran implementation of `insert` is not present in the most recent release.

yguclu (approved):
A one-liner fix... excellent! 🙂

PR #2283, opened 2025-04-14
Change the order of arguments to place results first

### LINKED ISSUE #2282: [Improvement] Change the order of the arguments when there are multiple returns

The following code:
```python
def f(a : int, b : int = 3, c : int = 4):
   return a*b, c
```
is translated to:
```fortran
  subroutine f(a, b, c, result_0, result_1)

    implicit none

    integer(i64), intent(out) :: result_0
    integer(i64), intent(out) :: result_1
    integer(i64), value :: a
    integer(i64), value :: b
    integer(i64), value :: c
```
In order for a user to call this method as: `f(a)` they must therefore know the names of the result variables. The call would be:
```fortran
call f(a, result_0 = result_0, result_1 = result_1)
```
This is unnecessarily complex. If results are printed first then we don't need to know the names. The code would then be:
```fortran
  subroutine f(result_0, result_1, a, b, c)

    implicit none

    integer(i64), intent(out) :: result_0
    integer(i64), intent(out) :: result_1
    integer(i64), value :: a
    integer(i64), value :: b
    integer(i64), value :: c
```
```fortran
call f(result_0, result_1, a)
```

This is a convention that is often seen in C++. The translation of the code above would be:
```cpp
int64_t f(int64_t a, int64_t b = 3, int64_t c = 4, int64_t& result_0, int64_t% result_1);
```
this is not possible so `b` and `c` must either become non-optional or `result_0` and `result_1` must be moved. The reordered version allows the optionals to be kept:
```cpp
int64_t f(int64_t& result_0, int64_t& result_1, int64_t a, int64_t b = 3, int64_t c = 4);
```

### Do you have any suggestions for improvement? (Optional)

This improvement will let us have cleaner stub files as it will no longer be necessary to retrieve the result names

### Record

EmilyBourne:
Change the order of arguments to place results first. Fixes #2282

### DIFF
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -1647,7 +1647,10 @@ class CCodePrinter(CodePrinter):
 
         if n_results > 1 or isinstance(expr.results.var.class_type, InhomogeneousTupleType):
             ret_type = self.get_c_type(PythonNativeInt())
-            arg_vars = result_vars + arg_vars
+            if expr.arguments and expr.arguments[0].bound_argument:
+                arg_vars = arg_vars[:1] + result_vars + arg_vars[1:]
+            else:
+                arg_vars = result_vars + arg_vars
             self._additional_args.append(result_vars) # Ensure correct result for is_c_pointer
         elif n_results == 1:
             ret_type = self.get_declare_type(result_vars[0])
@@ -2305,7 +2308,10 @@ class CCodePrinter(CodePrinter):
             else :
                 args.append(arg_val)
 
-        args = self._temporary_args + args
+        if func.arguments and func.arguments[0].bound_argument:
+            args = args[:1] + self._temporary_args + args[1:]
+        else:
+            args = self._temporary_args + args
         self._temporary_args = []
         args = ', '.join(self._print(ai) for a in args for ai in self.scope.collect_all_tuple_elements(a))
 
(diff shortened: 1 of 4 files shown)


pyccel/codegen/printing/ccode.py
```diff
@@ -1647,7 +1647,10 @@ def function_signature(self, expr, print_arg_names = True):
 
         if n_results > 1 or isinstance(expr.results.var.class_type, InhomogeneousTupleType):
             ret_type = self.get_c_type(PythonNativeInt())
-            arg_vars.extend(result_vars)
+            if expr.arguments and expr.arguments[0].bound_argument:
+                arg_vars = arg_vars[:1] + result_vars + arg_vars[1:]
```
yguclu, line 1651: Would you mind adding a comment here to remind the developer that `arg_vars[0]` is the class object `self` in this case? In the Fortran printer you call this variable `class_arg`, which makes it quite explicit.

pyccel/codegen/printing/ccode.py
```diff
@@ -2305,7 +2308,10 @@ def _print_FunctionCall(self, expr):
             else :
                 args.append(arg_val)
 
-        args += self._temporary_args
+        if func.arguments and func.arguments[0].bound_argument:
+            args = args[:1] + self._temporary_args + args[1:]
```
yguclu, line 2312: Same here please!

yguclu (commented):
Looks good!

pyccel/codegen/printing/ccode.py
EmilyBourne: Done : cf4a46570

yguclu: Thanks

yguclu (approved):

PR #2287, opened 2025-04-17
Add Fortran support for dict.pop

### LINKED ISSUE #2051: [DICT] Add Fortran support for pop
### Describe the feature
Add Fortran support for the dict method `pop`.

### Test Code
```python
def foo():
    a = {1:1.0, 2:2.0}
    b = a.pop(1)
    c = a.pop(3, 4.0)
    return a, b, c
```

### Proposed Solution
This can be implemented using the undocumented `erase` method from gFTL:
https://github.com/Goddard-Fortran-Ecosystem/gFTL/blob/main/include/v2/map/specification.inc

EmilyBourne:
Add Fortran support for dict.pop. Fixes #2051

### DIFF
--- a/pyccel/stdlib/gFTL_functions/Map_extensions.inc
+++ b/pyccel/stdlib/gFTL_functions/Map_extensions.inc
@@ -16,11 +16,10 @@ contains
     __T_declare_dummy__ :: value
 
     type (__map_iterator) :: iter
-    type (__map_iterator) :: new_iter
 
     iter = my_map % find( key )
     value = iter % second()
-    new_iter = my_map % erase( iter )
+    iter = my_map % erase( iter )
 
   end function __IDENTITY(Map)_pop
 
@@ -31,13 +30,12 @@ contains
 
     __T_declare_dummy__ :: value
     type (__map_iterator) :: iter
-    type (__map_iterator) :: new_iter
 
     iter = my_map % find( key )
 
     if (iter /= my_map % end()) then
       value = iter % second()
-      new_iter = my_map % erase( iter )
+      iter = my_map % erase( iter )
     else
       value = default
     endif
--- a/pyccel/stdlib/gFTL_functions/Map_extensions.inc
+++ b/pyccel/stdlib/gFTL_functions/Map_extensions.inc
@@ -35,7 +35,7 @@ contains
 
     iter = my_map % find( key )
 
-    if (iter /= my_map%end()) then
+    if (iter /= my_map % end()) then
       value = iter % second()
       new_iter = my_map % erase( iter )
     else
--- a/pyccel/stdlib/gFTL_functions/Map_extensions.inc
+++ b/pyccel/stdlib/gFTL_functions/Map_extensions.inc
@@ -1,0 +1,48 @@
+#include <map/header.inc>
+
+contains
+
+#define __IDENTITY(x) x
+#define __guard __map_guard
+#include "parameters/Key/copy_map_Key_to_internal_Key.inc"
+#include "parameters/T/copy_map_T_to_internal_T.inc"
+#include "parameters/Key/define_derived_macros.inc"
+#include "parameters/T/define_derived_macros.inc"
+
+  function __IDENTITY(Map)_pop(my_map, key) result(value)
+    class(Map), intent(inout) :: my_map
+    __Key_declare_dummy__, intent(in) :: key
+
+    __T_declare_dummy__ :: value
+
+    type (__map_iterator) :: iter
+    type (__map_iterator) :: new_iter
+
+    iter = my_map % find( key )
+    value = iter % second()
+    new_iter = my_map % erase( iter )
+
+  end function __IDENTITY(Map)_pop
+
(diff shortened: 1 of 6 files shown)


pyccel/stdlib/gFTL_functions/Map_extensions.inc
```diff
@@ -0,0 +1,48 @@
...
+contains
+
+#define __IDENTITY(x) x
+#define __guard __map_guard
+#include "parameters/Key/copy_map_Key_to_internal_Key.inc"
+#include "parameters/T/copy_map_T_to_internal_T.inc"
+#include "parameters/Key/define_derived_macros.inc"
+#include "parameters/T/define_derived_macros.inc"
+
+  function __IDENTITY(Map)_pop(my_map, key) result(value)
```
yguclu, line 12: Is the functions' indentation given on purpose, or by accident?

pyccel/stdlib/gFTL_functions/Map_extensions.inc
```diff
@@ -0,0 +1,48 @@
...
+    __Key_declare_dummy__, intent(in) :: key
+    __T_declare_dummy__, intent(in) :: default
+
+    __T_declare_dummy__ :: value
+    type (__map_iterator) :: iter
+    type (__map_iterator) :: new_iter
+
+    iter = my_map % find( key )
+
+    if (iter /= my_map%end()) then
```
yguclu, line 38: More consistent with the style of this file:
```suggestion
    if (iter /= my_map % end()) then
```

yguclu (commented):
Looks good to me! Just two very minor comments on style..

pyccel/stdlib/gFTL_functions/Map_extensions.inc
EmilyBourne: On purpose to match the usual indentation of a function in a module and what is done in similar files, e.g.
https://github.com/pyccel/pyccel/blob/devel/pyccel%2Fstdlib%2FgFTL_functions%2FVector_extensions.inc

yguclu: Sounds good!

pyccel/stdlib/gFTL_functions/Map_extensions.inc
```diff
@@ -0,0 +1,48 @@
...
+    __Key_declare_dummy__, intent(in) :: key
+
+    __T_declare_dummy__ :: value
+
+    type (__map_iterator) :: iter
+    type (__map_iterator) :: new_iter
+
+    iter = my_map % find( key )
+    value = iter % second()
+    new_iter = my_map % erase( iter )
```
yguclu, line 23: If `new_iter` is just a local iterator, why do we need it?

pyccel/stdlib/gFTL_functions/Map_extensions.inc
```diff
@@ -0,0 +1,48 @@
...
+
+    __T_declare_dummy__ :: value
+    type (__map_iterator) :: iter
+    type (__map_iterator) :: new_iter
+
+    iter = my_map % find( key )
+
+    if (iter /= my_map % end()) then
+      value = iter % second()
+      new_iter = my_map % erase( iter )
```
yguclu, line 40: Same comment as above.

yguclu (changes requested):
Another question...

pyccel/stdlib/gFTL_functions/Map_extensions.inc
yguclu: `new_iter` is not the result, though. The result of this function is `value`.

EmilyBourne: I am talking about the function `my_map % erase()`. Fortran doesn't let us do
```fortran
my_map % erase( iter )
```

And I'm not certain how well it works if I do:
```fortran
iter = my_map % erase( iter )
```
If the latter should work without any issues due to `iter` being passed as argument and used to save the result then I agree that we should use this

yguclu: I see. We would need a subroutine version of `erase`...

EmilyBourne: To remove the result entirely, yes. But this function is provided by gFTL and I don't think it's worth creating a `my_dict_type_erase_subroutine` method just for this

yguclu: In my previous comment I meant that gFTL should provide one 😉

EmilyBourne: `new_iter` removed: 044131b5c4ab62d563df3225408582778ce6d811

pyccel/stdlib/gFTL_functions/Map_extensions.inc
EmilyBourne: `new_iter` removed: 044131b5c4ab62d563df3225408582778ce6d811

PR #2288, opened 2025-04-17
Add C and Fortran support for dict.popitem

### LINKED ISSUE #2053: [DICT] Add C support for popitem
### Describe the feature
Add C support for the dict method `popitem`.

### Test Code
```python
def foo():
    a = {1,1.0, 2:2.0}
    b = a.popitem()
    c = a.popitem()
    return len(a), b, c
```

### Proposed Solution
This should be implemented using the hmap_X_erase method documented here:

### LINKED ISSUE #2054: [DICT] Add Fortran support for popitem
### Describe the feature
Add Fortran support for the dict method `popitem`.

### Test Code
```python
def foo():
    a = {1,1.0, 2:2.0}
    b = a.popitem()
    c = a.popitem()
    return len(a), b, c
```

### Proposed Solution
This can be implemented using the undocumented erase method from gFTL:
https://github.com/Goddard-Fortran-Ecosystem/gFTL/blob/main/include/v2/map/specification.inc

EmilyBourne:
Add C and Fortran support for dict.popitem. Fixes #2054. Fixes #2053 
Fix semantic stage to allow `a, b = c.popitem()` syntax.
Add extra tests.

### DIFF
--- a/pyccel/ast/builtin_methods/dict_methods.py
+++ b/pyccel/ast/builtin_methods/dict_methods.py
@@ -144,6 +144,13 @@ class DictPopitem(DictMethod):
         super().__init__(dict_obj)
 
     def __iter__(self):
+        """
+        Iterate over a popitem to get an example of a key and a value.
+
+        Iterate over a popitem to get an example of a key and a value. This
+        is particularly useful in the semantic stage in order to create the
+        variables representing the key and value objects.
+        """
         return iter((IndexedElement(self, 0), IndexedElement(self, 1)))
 
 #==============================================================================
--- a/pyccel/ast/builtin_methods/dict_methods.py
+++ b/pyccel/ast/builtin_methods/dict_methods.py
@@ -143,6 +143,9 @@ class DictPopitem(DictMethod):
         self._class_type = InhomogeneousTupleType(dict_type.key_type, dict_type.value_type)
         super().__init__(dict_obj)
 
+    def __iter__(self):
+        return iter((IndexedElement(self, 0), IndexedElement(self, 1)))
+
 #==============================================================================
 class DictGet(DictMethod):
     """
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -21,7 +21,7 @@ from pyccel.ast.builtins  import PythonPrint, PythonType, VariableIterator
 
 from pyccel.ast.builtins  import PythonList, PythonTuple, PythonSet, PythonDict, PythonLen
 
-from pyccel.ast.builtin_methods.dict_methods  import DictItems, DictKeys
+from pyccel.ast.builtin_methods.dict_methods  import DictItems, DictKeys, DictPopitem
 
 from pyccel.ast.core      import Declare, For, CodeBlock, ClassDef
 from pyccel.ast.core      import FunctionCall, FunctionCallArgument
@@ -2473,7 +2473,7 @@ class CCodePrinter(CodePrinter):
     def _print_Assign(self, expr):
         lhs = expr.lhs
         rhs = expr.rhs
-        if isinstance(rhs, FunctionCall) and isinstance(rhs.class_type, InhomogeneousTupleType):
+        if isinstance(rhs, (FunctionCall, DictPopitem)) and isinstance(rhs.class_type, InhomogeneousTupleType):
             self._temporary_args = [ObjectAddress(a) for a in lhs]
             code = self._print(rhs)
             self._temporary_args = []
@@ -3043,6 +3043,24 @@ class CCodePrinter(CodePrinter):
             pop_expr = f"{c_type}_pop({dict_var}, {key})"
 
         return pop_expr
+
+    def _print_DictPopitem(self, expr):
+        dict_obj = expr.dict_obj
+        class_type = dict_obj.class_type
+        c_type = self.get_c_type(class_type)
+
+        dict_obj_code = self._print(ObjectAddress(dict_obj))
+
+        assigns = expr.get_direct_user_nodes(lambda u: isinstance(u, Assign))
+        assert len(assigns) == 1
+        lhs = assigns[0].lhs
+
+        key, value = self._temporary_args
+
+        key_code = self._print(key)
+        value_code = self._print(value)
+
+        return f'{c_type}_popitem({dict_obj_code}, {key_code}, {value_code});\n'
 
     def _print_DictGetItem(self, expr):
         dict_obj = expr.dict_obj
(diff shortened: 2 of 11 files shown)


pyccel/codegen/printing/ccode.py
```diff
@@ -3044,6 +3044,24 @@ def _print_DictPop(self, expr):
...
+        assigns = expr.get_direct_user_nodes(lambda u: isinstance(u, Assign))
+        assert len(assigns) == 1
+        lhs = assigns[0].lhs
+
+        key, value = self._temporary_args
+
+        key_code = self._print(key)
+        value_code = self._print(value)
+
+        return f'{c_type}_popitem({dict_obj_code}, {key_code}, {value_code});\n'
```
yguclu, line 3063: Is this code going to be printed always on its own line?

pyccel/codegen/printing/fcode.py
```diff
@@ -2221,7 +2241,7 @@ def _print_Assign(self, expr):
         lhs = expr.lhs
         rhs = expr.rhs
 
-        if isinstance(rhs, (FunctionCall, SetUnion, ListPop)):
+        if isinstance(rhs, (FunctionCall, SetUnion, ListPop, DictPopitem)):
             return self._print(rhs)
```
yguclu, line 2245: So far I had not paid much attention to the fact that, under certain circumstances, we only print the RHS of an assignment. Is this because the LHS is effectively the `out` (or `inout`) parameter of the subroutine call on the RHS?

pyccel/codegen/printing/pycode.py
```diff
@@ -935,7 +935,11 @@ def _print_DictMethod(self, expr):
         dict_obj = self._print(expr.dict_obj)
         method_args = ', '.join(self._print(a) for a in expr.args)
 
-        return f"{dict_obj}.{method_name}({method_args})\n"
+        code = f"{dict_obj}.{method_name}({method_args})"
+        if isinstance(expr.class_type, VoidType):
+            return f'{code}\n'
+        else:
+            return code
```
yguclu, line 942: Can you please explain this change?

BTW, is `_print_DictMethod` a fallback printer? When do we actually use it?

pyccel/parser/semantic.py
```diff
@@ -3829,6 +3829,15 @@ def _visit_Assign(self, expr):
...
                 lhs = PythonTuple(*new_lhs)
+            elif isinstance(rhs, PyccelFunction):
+                assert isinstance(rhs.class_type, InhomogeneousTupleType)
+                r_iter = [self.scope.collect_tuple_element(v) for v in rhs]
+                new_lhs = []
+                for i,(l,r) in enumerate(zip(lhs, r_iter)):
+                    d = self._infer_type(r)
+                    new_lhs.append( self._assign_lhs_variable(l, d, r, new_expressions,
+                                                    arr_in_multirets=r.rank>0 ) )
+                lhs = PythonTuple(*new_lhs)
```
yguclu, line 3840: What is going on here? And how is it related to this PR?

pyccel/stdlib/gFTL_functions/Map_extensions.inc
```diff
@@ -44,5 +44,22 @@ contains
...
 
+  subroutine __IDENTITY(Map)_popitem(my_map, key, value)
+    class(Map), intent(inout) :: my_map
+    __Key_declare_dummy__, intent(out) :: key
+    __T_declare_dummy__, intent(out) :: value
+
+    type (__map_iterator) :: iter
+    type (__map_iterator) :: new_iter
+
+    iter =  my_map%begin()
```
yguclu, line 55: More consistent with the coding style used in this function:
```suggestion
    iter =  my_map % begin()
```

pyccel/stdlib/gFTL_functions/Map_extensions.inc
```diff
@@ -44,5 +44,22 @@ contains
...
+
+    key = iter % first()
+    value = iter % second()
+
+    new_iter = my_map % erase( iter )
```
yguclu, line 60: If `new_iter` is just a local iterator, why do we need it?

yguclu (changes requested):

pyccel/codegen/printing/pycode.py
EmilyBourne: `_print_DictMethod` is a fallback printer. For the Python implementation it is used for all dict methods which don't have default arguments (basically everything except `pop`. The newline should therefore not be printed as there are several dict methods that can be used in expressions. E.g. `d.popitem()[0]` was being printed as:
```python
d.popitem()
[0]
```
On the other hand methods such as`clear` which don't return anything need to be printed on their own line.
This change therefore checks the return type and uses it to decide if a new line is needed

pyccel/codegen/printing/ccode.py
```diff
@@ -3044,6 +3044,24 @@ def _print_DictPop(self, expr):
...
+    def _print_DictPopitem(self, expr):
+        dict_obj = expr.dict_obj
+        class_type = dict_obj.class_type
+        c_type = self.get_c_type(class_type)
+
+        dict_obj_code = self._print(ObjectAddress(dict_obj))
+
+        assigns = expr.get_direct_user_nodes(lambda u: isinstance(u, Assign))
+        assert len(assigns) == 1
+        lhs = assigns[0].lhs
```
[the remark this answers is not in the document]
EmilyBourne: Originally I was collecting the result variables from the `lhs` until I realised that this could be handled with `_additional_args` like all other function calls. This code is also related to your comment below. If there is exactly 1 `Assign` node in the direct users then we are certain that the popitem call would be printed on its own line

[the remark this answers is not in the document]
EmilyBourne: Removed : 94dbe69ac

pyccel/stdlib/gFTL_functions/Map_extensions.inc
EmilyBourne: Removed: bd4535c08

pyccel/ast/builtin_methods/dict_methods.py
```diff
@@ -143,6 +143,9 @@ def __init__(self, dict_obj):
         self._class_type = InhomogeneousTupleType(dict_type.key_type, dict_type.value_type)
         super().__init__(dict_obj)
 
+    def __iter__(self):
+        return iter((IndexedElement(self, 0), IndexedElement(self, 1)))
```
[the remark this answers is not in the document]
EmilyBourne: Done: 33dd81109

pyccel/parser/semantic.py
yguclu: Fascinating! Does this allow to handle something else (on top of `popitem`) which was not handled before?

EmilyBourne: Nothing comes to mind but it should be relevant for any built-in that returns an inhomogeneous tuple

yguclu (approved):

PR #2289, opened 2025-04-17
Add support for dict.values

### LINKED ISSUE #1883: [DICT] Add Python support for values()
### Describe the feature
Syntactic and semantic support should be added for `values`. This should be tested by implementing Python printing

### Test Code
```python
def foo():
    a = {1,1.0, 2:2.0}
    b : list[float] = []
    for bi in a.values():
        b.append(bi)
    return b
```

EmilyBourne:
Add support for `dict.values`. Fixes #1883

### DIFF
--- a/pyccel/ast/builtin_methods/dict_methods.py
+++ b/pyccel/ast/builtin_methods/dict_methods.py
@@ -347,7 +347,8 @@ class DictItems(Iterable):
         """
         Get the object representing the dict.
 
-        Get the object representing the dict.
+        Get the object representing the dict. The name of this method is
+        chosen to match the name of the equivalent method in VariableIterator.
         """
         return self._dict_obj
 
@@ -394,7 +395,8 @@ class DictKeys(Iterable):
         """
         Get the object representing the dict.
 
-        Get the object representing the dict.
+        Get the object representing the dict. The name of this method is
+        chosen to match the name of the equivalent method in VariableIterator.
         """
         return self._dict_obj
 
@@ -481,7 +483,8 @@ class DictValues(Iterable):
         """
         Get the object representing the dict.
 
-        Get the object representing the dict.
+        Get the object representing the dict. The name of this method is
+        chosen to match the name of the equivalent method in VariableIterator.
         """
         return self._dict_obj
 
--- a/pyccel/ast/builtin_methods/dict_methods.py
+++ b/pyccel/ast/builtin_methods/dict_methods.py
@@ -26,6 +26,7 @@ __all__ = ('DictClear',
            'DictPop',
            'DictPopitem',
            'DictSetDefault',
+           'DictValues',
            )
 
 #==============================================================================
--- a/pyccel/ast/builtin_methods/dict_methods.py
+++ b/pyccel/ast/builtin_methods/dict_methods.py
@@ -13,7 +13,7 @@ This module contains objects which describe these methods within Pyccel's AST.
 from pyccel.ast.datatypes import InhomogeneousTupleType, VoidType, SymbolicType
 from pyccel.ast.internals import PyccelFunction, Iterable, PyccelArrayShapeElement
 from pyccel.ast.literals  import LiteralInteger
-from pyccel.ast.variable  import IndexedElement
+from pyccel.ast.variable  import IndexedElement, Variable
 
 
 __all__ = ('DictClear',
@@ -412,8 +412,8 @@ class DictKeys(Iterable):
         list[TypedAstNode]
             A list of objects that should be assigned to variables.
         """
-        item = DictPopitem(self._dict_obj)
-        return [IndexedElement(item, 0)]
+        class_type = self._dict_obj.class_type.key_type
+        return [Variable(class_type, '_', memory_handling = 'heap' if class_type.rank > 0 else 'stack')]
 
 #==============================================================================
 class DictGetItem(DictMethod):
@@ -500,5 +500,5 @@ class DictValues(Iterable):
         list[TypedAstNode]
             A list of objects that should be assigned to variables.
         """
-        item = DictPopitem(self._dict_obj)
-        return [IndexedElement(item, 1)]
+        class_type = self._dict_obj.class_type.value_type
+        return [Variable(class_type, '_', memory_handling = 'heap' if class_type.rank > 0 else 'stack')]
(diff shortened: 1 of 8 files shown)


pyccel/ast/builtin_methods/dict_methods.py
```diff
@@ -451,3 +452,50 @@ def key(self):
...
+        super().__init__(1)
+
+    @property
+    def variable(self):
+        """
+        Get the object representing the dict.
+
+        Get the object representing the dict.
+        """
+        return self._dict_obj
```
yguclu, line 486: Class `DictMethod` calls this property `dict_obj`. Maybe we should do the same here for consistency?

The name `variable` appears to be used also in `DictItems` and `DictKeys`.

pyccel/ast/builtin_methods/dict_methods.py
```diff
@@ -451,3 +452,50 @@ def key(self):
...
+        Returns two objects that could be a key and a value of the dictionary.
+        These elements are used to determine the types of the loop targets.
+
+        Returns
+        -------
+        list[TypedAstNode]
+            A list of objects that should be assigned to variables.
+        """
+        item = DictPopitem(self._dict_obj)
+        return [IndexedElement(item, 1)]
```
yguclu, line 501: Why do we use `DictPopitem` here? `d.popitem()` modifies the dictionary `d` in place, while `d.values()` does not.

pyccel/ast/builtin_methods/dict_methods.py
```diff
@@ -451,3 +452,50 @@ def key(self):
...
+        """
+        Get the item of the iterable that will be saved to the loop targets.
+
+        Returns two objects that could be a key and a value of the dictionary.
+        These elements are used to determine the types of the loop targets.
+
+        Returns
+        -------
+        list[TypedAstNode]
+            A list of objects that should be assigned to variables.
```
yguclu, line 498: I have a few comments here:

- The docstring says that it returns "the item", or "two objects that could be a key and a value", but it really only returns the `value`, which is `d.popitem()[1]`.
- This function returns a list as explained, but with only one object. So I would clarify that.
- Given that `item := (key, value)` the Python jargon, I feel that this function should use the word `value` in its name and docstring.

yguclu (changes requested):
Very nice! I have a few comments

pyccel/ast/builtin_methods/dict_methods.py
EmilyBourne: This method returns an object with the type of the target. It is used in the semantic stage. Eg if we have:
```python
for v in d.values():
```
The type of `v` is deduced by parsing:
```python
v = d.popitem()[1]
```

I didn't want to create a new object to handle this as it is only used during the assignment. However now I think about it again, this might not be the correct solution. For multi dimensional objects this will result in `v` being a heap object when it should be an alias

yguclu: Gotcha. How do you want to proceed?

EmilyBourne: I can probably replace the call to DictPopitem with a Variable. I'll have a look on Monday or Tuesday

pyccel/ast/builtin_methods/dict_methods.py
EmilyBourne: The name `variable` is chosen to match the name used in `VariableIterator`:
https://github.com/pyccel/pyccel/blob/d25ae63327dccde8e0813c4b9a15d692cc77536b/pyccel/ast/builtins.py#L1634-L1663

This is useful as these objects act in an almost identical way in the same context. If I made `.items()` `.keys()` and `.values()` behave like `.pop` instead of like `a` in `for ai in a:` then I would need to treat `for ai in a:` and `for ai in a.keys():` differently even if `a` represents a `dict`. I can add a comment

pyccel/ast/builtin_methods/dict_methods.py
EmilyBourne: Done : 915968b20

pyccel/ast/builtin_methods/dict_methods.py
EmilyBourne: Docstrings fixed: 4f95dbd34

pyccel/ast/builtin_methods/dict_methods.py
yguclu: I see.

But then, why not calling this property `variable` in `DictMethod` too? I find it confusing that the `self` of a method is called with different names across our codebase. I do not have a strong preference for the name used, I just feel that we should pick any name and stick to it.

EmilyBourne: I don't call the property `variable` in `DictMethod` as the object may not be a variable. I don't have a good example for dicts but if we take the case of sets (`SetMethod` has a `set_obj` property) we may have:
```
a.copy().union(c)
```
Here the `set_obj` of `SetUnion` is `a.copy()` which is not a `Variable`. In order to avoid anyone assuming that it is a `Variable` I preferred to use `obj`.
This is not the case for an `iterable`. If the argument to `VariableIterator`/`DictKeys`/`DictValues`/`DictItems` is an expression then a temporary variable will be created in the semantic stage to ensure that it operates on a variable.

> I find it confusing that the self of a method is called with different names across our codebase.

I get this but the concept of `Iterable` and `DictMethod` are very different so it is not surprising that their implementation differs on certain points. It would be possible to replace `variable` in `VariableIterator` and `dict_obj` in `DictMethod` (and `set_obj` in `SetMethod` and `list_obj` in `ListMethod`) with `obj` but I feel that this would be a lot less descriptive. E.g in `ListInsert` we currently have the properties `list_obj`, `index` and `object` (where the names `index` and `object` are chosen to match Python's docstring of `list.insert`). `obj`, `index` and `object` would be much less clear. And `variable`, `index` and `object` would be misleading in the case of `f(my_list_ref).insert(0,4)`.

Similarly without reading the docstring I'm not sure I would be certain what `VariableIterator.obj` represents. I would guess the iterator object not the object being iterated over

EmilyBourne: If you have a suggestion that avoids these pitfalls then we could change but as this affects methods across several files unrelated to this PR I think it would be better to put this in its own PR

pyccel/ast/builtin_methods/dict_methods.py
EmilyBourne: This should be good enough for this PR though. If it is wrong then the fix is simply heap->alias

pyccel/ast/builtin_methods/dict_methods.py
yguclu: Yes, let's leave this to a separate discussion/issue/PR

yguclu (approved):

PR #2294, opened 2025-04-23
Fix LAPACK installation on windows

### LINKED ISSUE #1091: Windows will not run lapack or blas tests
Windows will not run lapack or blas tests. They fail with the error:
```
liblapack.dll: cannot open shared object file: No such file or directory
```
It is not clear what command should be used to avoid this error. Several things were tried in PR #959 but with no success. If anyone has a windows machine they can test on then this would be useful

_Originally posted by @EmilyBourne in https://github.com/pyccel/pyccel/issues/959#issuecomment-948542498_

EmilyBourne:
The Windows CI is failing as the University of Tennessee has changed the permissions on its website. It is therefore no longer possible to download the compiled dll files from this site.

This PR changes the installation process to collect LAPACK from the GitHub reference repository. It is compiled with CMake and the result is cached to avoid recompiling every time we use a Windows runner. This new installation fixes #1091

### DIFF
--- a/.github/actions/windows_install/action.yml
+++ b/.github/actions/windows_install/action.yml
@@ -40,7 +40,7 @@ runs:
         #cd build
         #cmake --help
         #echo "cmake -DCMAKE_INSTALL_PREFIX=${Env:MINGW_DIR} .."
-        cmake -D CMAKE_INSTALL_PREFIX="${Env:MINGW_DIR}" -B build -S .
+        cmake -D CMAKE_INSTALL_PREFIX="${Env:MINGW_DIR}" -D CMAKE_C_COMPILER=gcc -D CMAKE_Fortran_COMPILER=gfortran -B build -S .
         echo "cmake done"
         cmake --build build -j 4 --target install
         ls ${Env:MINGW_DIR}
--- a/.github/actions/windows_install/action.yml
+++ b/.github/actions/windows_install/action.yml
@@ -38,6 +38,9 @@ runs:
         cd lapack-3.12.1
         $GCC_EXE=(Get-Command gcc).Path
         $GFORTRAN_EXE=(Get-Command gfortran).Path
+        gcc --version
+        gfortran --version
+        echo "${Env:GCC_EXE}  ${Env:GFORTRAN_EXE}"
         cmake -D CMAKE_INSTALL_PREFIX="${Env:MINGW_DIR}" -D CMAKE_C_COMPILER="${Env:GCC_EXE}" -D CMAKE_Fortran_COMPILER="${Env:GFORTRAN_EXE}" -B build -S .
         echo "cmake done"
         cmake --build build -j 4 --target install
--- a/.github/actions/windows_install/action.yml
+++ b/.github/actions/windows_install/action.yml
@@ -39,7 +39,7 @@ runs:
         $GCC_EXE=((Get-Command gcc).Path -replace '\\', '/')
         $GFORTRAN_EXE=((Get-Command gfortran).Path -replace '\\', '/')
         echo "$GCC_EXE  $GFORTRAN_EXE  ${Env:MINGW_DIR}"
-        cmake -D CMAKE_INSTALL_PREFIX="${Env:MINGW_DIR}" -D CMAKE_C_COMPILER="$GCC_EXE" -D CMAKE_Fortran_COMPILER="$GFORTRAN_EXE" -B build -S .
+        cmake  -G "MinGW Makefiles" -D CMAKE_INSTALL_PREFIX="${Env:MINGW_DIR}" -D CMAKE_C_COMPILER="$GCC_EXE" -D CMAKE_Fortran_COMPILER="$GFORTRAN_EXE" -B build -S .
         echo "cmake done"
         cmake --build build -j 4 --target install
         ls ${Env:MINGW_DIR}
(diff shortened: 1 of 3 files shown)


.github/actions/windows_install/action.yml
```diff
@@ -21,30 +21,62 @@ runs:
         echo $EXE
         $BIN_DIR=(Split-Path -Parent ${EXE}) # Find folder containing c compiler
         $MINGW_DIR=(Split-Path -Parent ${BIN_DIR}) # Find the enclosing mingw directory
-        $LIBRARY_DIR="$MINGW_DIR/lib"
-        echo $LIBRARY_DIR | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
-        echo "LIBRARY_DIR=${LIBRARY_DIR}" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
+        echo $MINGW_DIR | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+        echo "MINGW_DIR=${MINGW_DIR}" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
```
yguclu, line 26: Now you are adding the enclosing `mingw` directory to GITHUB_PATH instead of just its `lib` subdirectory. Does this change make a difference, or is it just for better clarity in this action file?

.github/actions/windows_install/action.yml
yguclu, line 1: I suggest to store the LAPACK library version in a string variable (now `"3.12.1"`), so that it can be more easily changed in the future.

.github/actions/windows_install/action.yml
```diff
@@ -21,30 +21,62 @@ runs:
...
+          C:/mingw64/lib/cmake/lapack-3.12.0/lapack-targets-release.cmake
+          C:/mingw64/lib/pkgconfig/lapack.pc
+          C:/mingw64/lib/cmake/lapack-3.12.0/lapack-config.cmake
+          C:/mingw64/lib/cmake/lapack-3.12.0/lapack-config-version.cmake
+          C:/mingw64/lib/pkgconfig/blas.pc
+          C:/mingw64/lib/libblas.dll.a
+          C:/mingw64/bin/libblas.dll
+          C:/mingw64/lib/liblapack.dll.a
+          C:/mingw64/bin/liblapack.dll
+        key: ${{ runner.os }}-gcc-lapack-v3.12.1
```
yguclu, line 46: Are we using version `3.12.0` or `3.12.1`?

.github/actions/windows_install/action.yml
```diff
@@ -21,30 +21,61 @@ runs:
...
+    - name: Save Lapack Cache
+      id: cache-lapack-save
+      uses: actions/cache/save@v3
+      with:
+        path: |
+          C:/mingw64/lib/cmake/lapack-3.12.0/lapack-targets.cmake
+          C:/mingw64/lib/cmake/lapack-3.12.0/lapack-targets-release.cmake
+          C:/mingw64/lib/pkgconfig/lapack.pc
+          C:/mingw64/lib/cmake/lapack-3.12.0/lapack-config.cmake
+          C:/mingw64/lib/cmake/lapack-3.12.0/lapack-config-version.cmake
```
yguclu, line 72: Here we have `3.12.0` again.

yguclu (changes requested):

.github/actions/windows_install/action.yml
yguclu: Definitely, it is because we installed the dynamic libraries into the lib/ folder

.github/actions/windows_install/action.yml
EmilyBourne: I am installing v3.12.1:

https://github.com/pyccel/pyccel/blob/6425fa2d298e7fc8e1efdec1b800d9d7110f1312/.github/actions/windows_install/action.yml#L53

But I am caching the locations that CMake spat out. It looks like LAPACK forgot to update this when they created the tag. The tag was created 8th Jan: https://github.com/Reference-LAPACK/lapack/releases/tag/v3.12.1
but the version was update 12th Jan:
https://github.com/Reference-LAPACK/lapack/commit/3e242fc8d5d38f0425c6018fe88abf31e1a9e80f

.github/actions/windows_install/action.yml
EmilyBourne: I think it is pretty easy to change with sed or vim. On the other hand I have no idea how well powershell deals with string variables in the middle of expressions and don't have a windows machine to test on. I also don't think that such variables could be used in the caching unless there is a way to make them GitHub variables instead. If we can then that would be an option. Although as seen below that will not clarify everything as the cached files are in the wrong location

.github/actions/windows_install/action.yml
EmilyBourne: I have removed this line: [6425fa2](https://github.com/pyccel/pyccel/pull/2294/commits/6425fa2d298e7fc8e1efdec1b800d9d7110f1312)

.github/actions/windows_install/action.yml
yguclu: Ah, I have just read your answer to my other comment 😆 

Never mind, there is no point in using a string variable if the LAPACK team does not update the names consistently

.github/actions/windows_install/action.yml
yguclu: Is there a way to get these locations automatically from CMake?

yguclu: I agree

yguclu (approved):

PR #2299, opened 2025-04-24
Make the contents of Fortran modules private by default

### LINKED ISSUE #2144: Importing a module in `Fortran` does not account for name collisions with aliases
### Describe the bug
When importing a module in `Fortran`, The name collision resolution mechanism does not account for local aliases used in the imported module. This is especially critical with `pyc_math_f90`; using `i32` as a variable like it is the case in `tests/pyccel/scripts/expressions.py` causes compilation errors.

### To Reproduce
```python
import numpy as np
from math import gcd

if __name__ == "__main__":
    i32 = 1
    print(gcd(i32, i32))
```
Here, Importing pyc_math_f90 introduces the name collision, even when we only wanna import  a single function from the module.

Pyccel however has no issues with the collision inside the same module. 
```python
if __name__ == "__main__":
    i64 = 1
```
The code above translates to:
```Fortran
program prog_prog_soo
  use soo
  use, intrinsic :: ISO_C_Binding, only : C_INT64_T
  implicit none
  integer(C_INT64_T) :: i64
  i64 = 1_C_INT64_T
end program prog_prog_soo
```
### Error details

```bash
__pyccel__/prog_soo.f90:11:21:

    4 |   use pyc_math_f90
      |      2               
......
   11 |   integer(i64) :: i32
      |                     1
Error: Symbol ‘i32’ at (1) conflicts with symbol from module ‘__iso_c_binding’, use-associated at (2)
__pyccel__/prog_soo.f90:13:2:

   13 |   i32 = 1_i64
      |  1
Error: Named constant ‘i32’ in variable definition context (assignment) at (1)
```

### Expected behavior
- The name collision should be extended to imported modules.
- `from math import x` should translate to ` use pyc_math_f90, only: x`
- for the time, being a quick fix is to make the aliases private in `pyc_math_f90`

### Language
Fortran

EmilyBourne:
Make the contents of Fortran modules private by default instead of public. Only non-private functions, non-private variables, interfaces and classes should be explicitly public. Fixes #2144


EmilyBourne:
No test has been added as I cannot think of one. When there is an `import` statement in Python we correctly translate this to `use X, only Y`. We only use `use X` for the module of a program or for a stdlib file. This bug is therefore only really dangerous for users using their code in an external project.

pyccel/stdlib/gFTL_functions/Set_extensions.inc
```diff
@@ -1,9 +1,15 @@
 
 #include <set/header.inc>
+#define __IDENTITY(x) x
```
yguclu, line 3: What is the purpose of this macro?

pyccel/stdlib/gFTL_functions/Vector_extensions.inc
```diff
@@ -1,8 +1,11 @@
 #include <vector/header.inc>
+#define __IDENTITY(x) x
```
yguclu, line 2: Same question as in the previous file

yguclu (approved):
Looks good to me!

PR #2301, opened 2025-04-24
Fix printing of lists

### LINKED ISSUE #2097: Print does not handle empty lists well
### Describe the bug
When printing lists, the actual code prints the inner elements except the last one, then prints the last element along with `]`. The last element is obtained by `*vec_int64_t_at(&a,vec_int64_t_size(&a) - INT64_C(1))` in `C` and ` write(stdout, '(I0, A)', advance="no") a%of(a % size() - 1_i64 + 1_i64 &` in `Fortran`. These accesses are prohibited when the length of the list is 0/

### To Reproduce
```python
if __name__ == '__main__':
    a : 'list[int]'= []
    print(a)
```

### Error details

```c
    printf("[");
    for (i = INT64_C(0); i < vec_int64_t_size(&a) - INT64_C(1); i += INT64_C(1))
    {
        printf("%"PRId64" ", (*vec_int64_t_at(&a,i)));
    }
    printf("%"PRId64"]\n", (*vec_int64_t_at(&a,vec_int64_t_size(&a) - INT64_C(1))));
    vec_int64_t_drop(&a);
```
```fortran
  do i = 0_i64, a % size() - 1_i64 - 1_i64
    write(stdout, '(I0, A)', advance="no") a%of(i + 1_i64) , ' '
  end do
        ) , ']'
```
### Expected behavior
A first check on the length of the array should be performed before actually printing.

### Language
C, Fortran

EmilyBourne:
Fix printing of lists by adding an if block round the printing of the final element. This code handles lists but also arrays so special handling is added for the common case of an array of known (non-zero) size. Fixes #2097


pyccel/codegen/printing/fcode.py
```diff
@@ -3177,10 +3184,10 @@ def _print_If(self, expr):
 
         for i, (c, e) in enumerate(expr.blocks):
 
-            if i == 0:
-                lines.append("if (%s) then\n" % self._print(c))
-            elif i == len(expr.blocks) - 1 and isinstance(c, LiteralTrue):
+            if i == len(expr.blocks) - 1 and isinstance(c, LiteralTrue):
                 lines.append("else\n")
+            elif i == 0:
+                lines.append("if (%s) then\n" % self._print(c))
```
EmilyBourne, line 3190:
```suggestion
                lines.append(f"if ({self._print(c)}) then\n" )
```

yguclu (approved):
Looks good

PR #2302, opened 2025-04-24
Print deallocate in a more compact form

EmilyBourne:
The deallocation of arrays takes 3 lines which draws attention unnecessarily to the fact that it handles the case where the allocation occurred conditionally (e.g. in an if block). This PR changes the print to a 1 line form to neaten the code.

### DIFF
--- a/pyccel/codegen/printing/fcode.py
+++ b/pyccel/codegen/printing/fcode.py
@@ -2416,9 +2416,7 @@ class FCodePrinter(CodePrinter):
             return ''
         elif isinstance(class_type, (NumpyNDArrayType, HomogeneousTupleType, StringType)):
             var_code = self._print(var)
-            code  = 'if (allocated({})) then\n'.format(var_code)
-            code += '  deallocate({})\n'     .format(var_code)
-            code += 'end if\n'
+            code  = f'if (allocated({var_code})) deallocate({var_code})\n'
             return code
         else:
             errors.report(f"Deallocate not implemented for {class_type}",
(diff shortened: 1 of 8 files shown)


EmilyBourne:
This is useful for the readability of the PASC poster

pyccel/codegen/printing/fcode.py
```diff
@@ -231,11 +231,11 @@
     'C_BOOL'                : 'b1'
 }
 
-inc_keyword = (r'do\b', r'if\b',
+inc_keyword = (r'do\b', r'if \(.*?\) then$',
                r'else\b', r'type\b\s*[^\(]',
                r'(elemental )?(pure )?(recursive )?((subroutine)|(function))\b',
                r'interface\b',r'module\b(?! *procedure)',r'program\b')
-inc_regex = re.compile('|'.join('({})'.format(i) for i in inc_keyword))
+inc_regex = re.compile('|'.join(f'({i})' for i in inc_keyword))
```
yguclu, line 238: Could you please explain the change in `inc_keyword`?

yguclu (commented):
Looks good. I just have one question

pyccel/codegen/printing/fcode.py
EmilyBourne: Previously we indented after a line starting with the word if. This means that we would indent after:
```fortran
if (allocated(arr)) deallocate(arr)
```
because it starts with `if`. The new regex means we indent after a line starting with the word if and ending with the word `then`

yguclu (approved):

PR #2311, opened 2025-05-22
Fix Intel compilation

EmilyBourne:
The `options` parameter of the `ClassDef` class was unused. However its default parameter (`public`) was conflicting with the global public declaration of module elements. This parameter has been removed. This fixes the `intel` CI workflow, which is added to the PR tests to prevent future regressions.


yguclu (approved):

EmilyBourne:
@yguclu I was going to wait 5 mins to pick up the results of the previous run

PR #2312, opened 2025-05-29
Fix integer linspace

### LINKED ISSUE #2126: Wrong result when using integer dtype in np.linspace
### Describe the bug
When the dtype and all the arguments of a np.linspace are integers the array calculated by Pyccel does not match that calculated by NumPy. NumPy uses integer division to calculate the step while Pyccel uses division of doubles

### To Reproduce
```python
import numpy as np

if __name__ == '__main__':
    a = np.linspace(-393, 5, 7, False, dtype=np.int32)
    print(a)
```

### Error details
Python output:
```
[-393 -337 -280 -223 -166 -109  -52]
```
Fortran output:
```
[-393 -336 -279 -222 -165 -108 -51]
```
C  output:
```
[-393 -336 -279 -222 -165 -108 -51]
```

### Language
Fortran/C

### Additional context
There is a test function related to this here:
https://github.com/pyccel/pyccel/blob/fede2e66eadd734a3af9e683f6075cad5cfc741c/tests/epyccel/recognised_functions/test_numpy_funcs.py#L5555-L5564

but it is not called in the test itself. The call is added in a comment in #2124

EmilyBourne:
Fix wrong result when using integer dtype in `np.linspace`. The difference between our previous implementation and NumPy's implementation is that we were relying on `int` to do the rounding, while NumPy were using `floor`. This is fixed for both C and Fortran. Fixes #2126 . The code is further cleaned up to use f-strings instead of relying on a string template and calling `format` as a last step.

### DIFF
--- a/tests/epyccel/recognised_functions/test_numpy_funcs.py
+++ b/tests/epyccel/recognised_functions/test_numpy_funcs.py
@@ -1,6 +1,7 @@
 # pylint: disable=missing-function-docstring, missing-module-docstring
 import os
 import sys
+from typing import TypeVar
 import pytest
 from numpy.random import rand, randn, uniform
 from numpy import isclose, iinfo, finfo, complex64, complex128
@@ -5456,7 +5457,8 @@ def test_numpy_linspace_scalar(language):
         for i in range(len(x)):
             result[i] = x[i]
 
-    def test_linspace_int(start : int, end : int, step : int, endpoint : bool):
+    FI = TypeVar('FI', float, int)
+    def test_linspace_int(start : FI, end : FI, step : int, endpoint : bool):
         return np.linspace(start, end, step, endpoint, dtype=np.int32)
 
     integer8 = randint(min_int8, max_int8 // 2, dtype=np.int8)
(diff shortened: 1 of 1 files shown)


yguclu (commented):
According to the description of issue #2126, the bug only arises in the case where **all** the arguments of the call to `np.linspace` are integer. I suspect that `start` and `stop` can be floats as long as `dtype` is an integer type. Could you please check this?

yguclu:
> > According to the description of issue #2126, the bug only arises in the case where **all** the arguments of the call to `np.linspace` are integer. I suspect that `start` and `stop` can be floats as long as `dtype` is an integer type. Could you please check this?
> 
> Given the fix I agree. I can confirm that on `devel`:
> 
> ```python
> import numpy as np
> 
> if __name__ == '__main__':
>     a = np.linspace(-393.0, 5.0, 7, False, dtype=np.int32)
>     print(a)
> ```
> 
> gives the same bug as reported in #2126

Thanks for checking! Do we have a unit test for this additional case? We should...

yguclu (approved):
Looks good to me!

PR #2314, opened 2025-05-29
Fix class scoping

### LINKED ISSUE #2093: Scoping wrong for class functions
### Describe the bug
It is not possible to name a local variable with the same name as a class method. This is because the scope of the class method is currently inside the scope of the class. This is wrong, in reality none of the methods or variables of the class can be accessed without using `self.` the scope of the class method should therefore be inside the global scope.

### To Reproduce
```python
class A:
        self._x = x

    @property
    def x(self):
        """
        Bla
        """
        return self._x

    def translate(self, y : float):
        x = self.x
        return x + y
```

### Error details
```bash/fortran/c
pyccel:
 |fatal [semantic]: tmp.py [14,8]| Object has already been defined with type '<class 'pyccel.ast.core.FunctionDef'>'. Type 'float' in assignment is incompatible (x=float)
```

### Expected behavior
The scope of the class method should therefore be inside the global scope.

### Language
Semantic error affects all languages.

### Additional context
Used by PyGyro but easy to avoid

EmilyBourne:
Fix class scoping. Methods should have the Module as `parent_scope` (used for name clashes), but the resulting function should be saved into the scope of the class. This can cause problems for retrieving the name of the function itself so the syntactic<->semantic mapping of the name of a function is added to the scope of that function in the `create_new_function_scope` in the semantic stage. Fixes #2093 

**Commit Summary**
- Add assertions to `FunctionDef` and `ClassDef` to ensure the scopes are well-formed
- Exit class scopes before visiting methods so they have the correct parent
- Correct size of homogeneous tuple in wrapper (not clear why this error became apparent in this PR)
- Rename `self._current_function`->`self._current_function_name`
- Use a list to store `Parser._current_function_name` to ensure history can be cleanly retrieved
- When a function is inserted into the scope, use it to update a new variable `self._current_function` (useful to avoid searching for the function in the scope, which is difficult for a class method where the scope is not readily available)
- Move `Parser._annotate_the_called_function_def` from `Parser` to children `SyntacticParser` and `SemanticParser` to allow a more complex implementation for `SemanticParser`
- Use 2 name arguments for `SemanticParser._annotate_the_called_function_def` to ensure correct handling of syntactic and semantic names
- Call correct scope from `collect_tuple_element` if the argument is a `DottedVariable`
- Remove unused code
- Use new `current_function_name` attribute to access `self._current_function_name` to avoid code duplication
- Change the arguments of `SemanticParser._annotate_the_called_function_def` to make the `function_call_args` argument compulsory. This reduces code duplication
- Insert functions into the correct scope
- Add a test (from the issue) with a class with an attribute whose name conflicts with a variable found in the other methods of the class

### DIFF
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -3661,10 +3661,11 @@ class SemanticParser(BasicParser):
                 cls_def = semantic_lhs_var.lhs.cls_base
                 insert_scope = cls_def.scope
                 cls_def.add_new_attribute(semantic_lhs_var)
-                lhs = lhs.name
+                lhs_scope_name = lhs.name.name[-1]
             else:
                 insert_scope = self.scope
-                lhs = lhs.name
+                lhs_scope_name = lhs.name
+            lhs = lhs.name
 
             if semantic_lhs_var.class_type is TypeAlias():
                 pyccel_stage.set_stage('syntactic')
@@ -3685,7 +3686,7 @@ class SemanticParser(BasicParser):
                 return EmptyNode()
 
             try:
-                insert_scope.insert_variable(semantic_lhs_var, lhs)
+                insert_scope.insert_variable(semantic_lhs_var, lhs_scope_name)
             except RuntimeError as e:
                 errors.report(e, symbol=expr, severity='error')
 
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -1365,22 +1365,25 @@ class SemanticParser(BasicParser):
         func: FunctionDef|Interface
             The new annotated function.
         """
+        name = old_func.name
+        cls_base_syntactic = old_func.get_direct_user_nodes(lambda p: isinstance(p, ClassDef))
+        if cls_base_syntactic:
+            cls_name = cls_base_syntactic[0].name
+            cls_base = self.scope.find(cls_name, 'classes')
+            cls_scope = cls_base.scope
+            key = 'classes'
+            search_name = cls_name
+        else:
+            key = 'functions'
+            search_name = old_func.name
         # The function call might be in a completely different scope from the FunctionDef
         # Store the current scope and go to the parent scope of the FunctionDef
-        old_scope            = self._scope
-        old_current_function = self._current_function_name
-        names = []
-        sc = old_func.scope if isinstance(old_func, FunctionDef) else old_func.syntactic_node.scope
-        while sc.parent_scope is not None:
+        old_scope = self._scope
+        old_current_function = self._current_function
+        old_current_function_name = self._current_function_name
+        sc = self.scope
+        while search_name not in getattr(sc, key):
             sc = sc.parent_scope
-            if not sc.name is None:
-                names.append(sc.name)
-        names.reverse()
-        self._current_function_name = names
-
-        while names:
-            sc = sc.sons_scopes[names[0]]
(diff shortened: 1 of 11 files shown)


EmilyBourne:
/bot run linux docs pyccel_lint pylint spelling

pyccel/parser/semantic.py
```diff
@@ -355,6 +355,47 @@ def annotate(self):
...
+        -------
+        Scope
+            The new scope for the function.
+        """
+        child = self.scope.new_child_scope(syntactic_name, **kwargs)
+        child.local_used_symbols[syntactic_name] = semantic_name
+        child.python_names[semantic_name] = syntactic_name
+
+
+        self._scope = child
```
yguclu, line 394: One blank line should be enough here

pyccel/parser/semantic.py
```diff
@@ -665,26 +706,19 @@ class instance persists after a call to a class method, and the arguments of a
         """
         assert not isinstance(exceptions, Variable)
         for i in self._allocs[-1]:
-            if isinstance(i, DottedVariable):
-                if isinstance(i.lhs.class_type, CustomDataType) and self._current_function != '__del__':
-                    continue
```
yguclu, line 670: What happened here?

pyccel/parser/semantic.py
```diff
@@ -1358,47 +1392,65 @@ def _annotate_the_called_function_def(self, old_func, function_call_args=None):
...
+        while new_scope.parent_scope is not None:
+            new_scope = new_scope.parent_scope
+            if not new_scope.name is None:
+                scope_names.append(new_scope.name)
+        scope_names.reverse()
+
+        # Use scope_names to find semantic scopes
+        while scope_names:
+            new_scope = new_scope.sons_scopes[scope_names[0]]
+            scope_names = scope_names[1:]
```
yguclu, line 1429: I have the feeling that this code can be rewritten with no reverse() and no redefining of the list in the loop.

yguclu (changes requested):
Looks really good to me! I just have a couple of questions

pyccel/parser/semantic.py
EmilyBourne: The coverage flagged it as unused and I couldn't find an example that triggers it. I don't know if that is because it was always unused or if my changes had an effect. It is possible that the allocated objects are now correctly associated with the scope of their class so there are no `DottedVariable`s in scopes anymore

pyccel/parser/semantic.py
EmilyBourne: Fixed [c024b87](https://github.com/pyccel/pyccel/pull/2314/commits/c024b8736cd968746fa234c412d99cdf4d42133f)

pyccel/parser/semantic.py
EmilyBourne: Fixed [10683bf](https://github.com/pyccel/pyccel/pull/2314/commits/10683bfdb7ea47803f29a4d2634b3e32618fb29e)

pyccel/parser/semantic.py
EmilyBourne: Maybe we should think about using a tool such as [`black`](https://black.readthedocs.io/en/stable/index.html) to ensure consistent formatting?

yguclu: Yes! Good idea!

PR #2315, opened 2025-05-29
Fix missing index in loop unravelling for loop of length 1

### LINKED ISSUE #2042: Loop over single element is not correctly handled
### Describe the bug
When unravelling vector expressions loops of length 1 are not turned into for loops as such loops are not necessary. However a loop index is still generated but then is never initialised.

### To Reproduce
```python

@stack_array('a','b')
def f():
    import numpy as np
    a = np.array([[1,2,3], [4,5,6]])
    b = a[1:,:] + a[:1,:] # Equivalent to b = a[1,:] + a[0,:]
```

### Error details
Provide the generated code:
```c
#include "tmp_arr2.h"


void f(void)
{
    int64_t array_dummy[INT64_C(6)];
    t_ndarray a = (t_ndarray){
        .nd_int64=array_dummy,
        .shape=(int64_t[]){INT64_C(2), INT64_C(3)},
        .strides=(int64_t[2]){0},
        .nd=2,
        .type=nd_int64,
        .is_view=false
    };
    stack_array_init(&a);
    int64_t array_dummy_0001[INT64_C(3)];
    t_ndarray b = (t_ndarray){
        .nd_int64=array_dummy_0001,
        .shape=(int64_t[]){INT64_C(1), INT64_C(3)},
        .strides=(int64_t[2]){0},
        .nd=2,
        .type=nd_int64,
        .is_view=false
    };
    stack_array_init(&b);
    int64_t i;
    int64_t i_0001;
    int64_t Dummy_0000[] = {INT64_C(1), INT64_C(2), INT64_C(3), INT64_C(4), INT64_C(5), INT64_C(6)};
    memcpy(&a.nd_int64[INT64_C(0)], Dummy_0000, 6 * a.type_size);
    for (i_0001 = INT64_C(0); i_0001 < INT64_C(3); i_0001 += INT64_C(1))
    {
        GET_ELEMENT(b, nd_int64, INT64_C(0), i_0001) = GET_ELEMENT(a, nd_int64, i + INT64_C(1), i_0001) + GET_ELEMENT(a, nd_int64, i, i_0001);
    }
}
```
here we see that `i` is used but is not initialised.

### Expected behaviour
The index `i` should be replaced with 0. Ideally expressions such as `i + INT64_C(1)` should be simplified to `INT64_C(1)`

### Language
C

EmilyBourne:
Fix missing index in loop unravelling for loop of length 1. This is a general fix which works but leads to expressions such as `0+4`. Fixes #2042.

Use `LiteralInteger(0)` as index when size is known to be 0 to ensure better simplification. I am not 100% certain that this includes all locations where indices may be generated so I have also left the general fix in place.

Pre-existing tests which fail consistently in debug mode (now used in the CI) are reactivated.

### DIFF
--- a/pyccel/ast/utilities.py
+++ b/pyccel/ast/utilities.py
@@ -288,14 +288,13 @@ def insert_index(expr, pos, index_var):
     >>> insert_index(expr, 0, i)
     IndexedElement(c, i) := IndexedElement(a, i) + IndexedElement(b, i)
     """
+    if expr.rank==0 or -pos > expr.rank:
+        return expr
+
     if expr.shape and expr.shape[pos] == 1:
         index_var = LiteralInteger(0)
 
-    if expr.rank==0:
-        return expr
-    elif isinstance(expr, (Variable, ObjectAddress)):
-        if expr.rank==0 or -pos>expr.rank:
-            return expr
+    if isinstance(expr, (Variable, ObjectAddress)):
 
         # Add index at the required position
         indexes = [Slice(None,None)]*(expr.rank+pos) + [index_var]+[Slice(None,None)]*(-1-pos)


yguclu (approved):
Looks good to me

PR #2318, opened 2025-06-02
Replace .pyh files in stdlib/ with .pyi files

### LINKED ISSUE #2173: Remove header syntax in stdlib
### Describe the change
In order to deprecate header syntax we need to update the `.pyh` files in `stdlib/internal`. These will eventually become `.pyi` files but it is already possible to write these files with almost identical syntax to what will be used in the `.pyi` files.

E.g.:
```python
#$ header function omp_get_ancestor_thread_num(int32) results(int32)
```
should become:
```python
def omp_get_ancestor_thread_num(level : 'int32') -> 'int32':
    ...
```

This ought to be easy to automatise, however it may be harder than it seems as the argument names are all missing from the current implementation.

EmilyBourne:
Replace .pyh files in stdlib/ with .pyi files. Fixes #2173 . The new files are almost entirely auto-generated from the old files by temporarily commenting the lines which make Pyccel stop before generating a translation of these files. Files which have not been working for a while or have never worked (fftw.pyh, hdf5.pyh) were luckily short so they could be translated manually.


EmilyBourne:
To merge after #297

pyccel/parser/semantic.py
```diff
@@ -2827,47 +2828,12 @@ def _visit_Module(self, expr):
...
-                    else:
-                        errors.report(IMPORTING_EXISTING_IDENTIFIED,
-                                symbol=name,
-                                severity='fatal')
+            if self.metavars.get('external', False):
+                for f in funcs:
+                    f.is_external = True
+                for c in classes:
+                    for m in c.methods:
+                        m.is_external = True
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
This code is identical to the code above. The case of classes is included for completeness

pyccel/ast/core.py
```diff
@@ -2418,6 +2413,12 @@ def is_external(self):
         from a f77 module """
         return self._is_external
 
+    @is_external.setter
+    def is_external(self, is_external):
+        """
+        """
+        self._is_external = is_external
```
yguclu, line 2420: Why not checking if the argument is a `bool` as was done originally in the `__init__`?

docs/header-files.md
```diff
@@ -26,7 +24,8 @@ from header import omp_get_num_threads
 print('number of threads is :', omp_get_num_threads())
 ```
 Pyccel can compile the Python file with the following command: `pyccel example.py --openmp`
-, It will then create the executable file `example`
+, it will then create the executable file `example`.
```
yguclu, line 27: Could we not place the comma at the end of the previous line?

yguclu (changes requested):
Looks good! I have a couple of comments

docs/header-files.md
```diff
@@ -64,7 +63,7 @@ We then create a static library using these commands:
 -   `gfortran -c funcs.f90`
 -   `ar rcs libfuncs.a funcs.o`
 
-In order to use this library the user needs to create a header file, we call it  `funcs_headers.pyh`
+In order to use this library the user needs to create a header file, we call it  `funcs_headers.pyi`
 ```python
 #$ header metavar module_name      = "funcs"
 #$ header metavar ignore_at_import = True
```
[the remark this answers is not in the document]
EmilyBourne: I did this in #2329 which actually deprecates this syntax

docs/header-files.md
EmilyBourne: If you can't see why this line break is necessary either then I'm removing it: e70a9f958

I assumed there was some kind of problem with a code snippet in the middle of a row. But I use these a lot and I am not aware of any problem

pyccel/ast/core.py
EmilyBourne: Fixed [a2c04d1](https://github.com/pyccel/pyccel/pull/2318/commits/a2c04d1a6564ca26f3d910ee5955b277b5b97d2c)

Issue #2204, opened 2025-02-10
Bot is sometimes encountering API rate limits


EmilyBourne:
### Describe the bug
The bot is sometimes encountering API rate limits which prevent it from reporting the results of tests.
E.g. https://github.com/pyccel/pyccel/actions/runs/13189748142/job/36820142477

This slows development as it means tests cannot be completed so code cannot be merged.

Currently the bot does the following:
1. checks if PR is ready when marked as ready. A PR is ready if:
    - There is a sufficient description
    - The checklist is completed
2. allows checks to be requested via a comment
3. ensures that tests are not rerun if no relevant files have been changed
4. leaves comments on PRs indicating coverage
5. manages flags
6. requests reviews

Some of these tasks could be managed without authenticating as the bot. Doing this would reduce our risk of encountering these limits.

In particular:
- [ ] Point 3 could be handled using https://github.com/fkirc/skip-duplicate-actions
- [ ] Point 2 could be managed using `gh` (https://cli.github.com/manual/gh_workflow_run) without using the bot's authentication
- [ ] Point 4 should be able to be managed using `gh`. Maybe this step should be externalised to an action that could be used by other repos. GitHub's marketplace contains similar tools but none allow certain lines of missing coverage to be flagged as acceptable

PR #2321, opened 2025-06-03
Try to reduce API rate limit issues

EmilyBourne:
The bot occasionally hits API rate limits. This usually happens while looking for duplicate runs. This PR changes the handling to let https://github.com/fkirc/skip-duplicate-actions deal with this step.
Further fast static analysis checks (spelling, pylint, pyccel_lint, docs) are automatically triggered on non-draft PRs.

This fixes the main problem raised in #2204 but there are still recommendations in that issue that could be applied

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -23,6 +23,7 @@ on:
       - opened
       - reopened
       - synchronize
+      - ready_for_review
 
 env:
   COMMIT: ${{ inputs.ref || github.event.ref }}
--- a/.github/workflows/pyccel_lint.yml
+++ b/.github/workflows/pyccel_lint.yml
@@ -23,6 +23,7 @@ on:
       - opened
       - reopened
       - synchronize
+      - ready_for_review
 
 env:
   COMMIT: ${{ inputs.ref || github.event.ref }}
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -23,6 +23,7 @@ on:
       - opened
       - reopened
       - synchronize
+      - ready_for_review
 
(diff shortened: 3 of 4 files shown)


EmilyBourne:
- [ ] Write a clear PR description
- [ ] Add tests to check your code works as expected
- [ ] Update Changelog

.github/workflows/linux.yml
```diff
@@ -57,6 +57,14 @@ jobs:
       matrix: ${{fromJson(needs.matrix_prep.outputs.matrix)}}
 
     steps:
+      - id: duplicate_check
+        uses: fkirc/skip-duplicate-actions@v5
+        with:
+          # All of these options are optional, so you can remove them if you are happy with the defaults
+          skip_after_successful_duplicate: 'true'
+          paths: '["pyccel/**/*.py", "tests/**/*.py"]'
+          paths_ignore: '["pyccel/version.py"]'
```
yguclu, line 66: Is this the recommended way to use  `skip-duplicate-actions`? Using it as a job step and adding an `if` condition to every subsequent step looks a bit cumbersome to me. Maybe one can run it as an independent job and have the other jobs depend on it?

EmilyBourne: We have the choice, but if you run it as a separate job then you can't make that job a requirement for the PR. Note that I am skipping all the steps which run the test, but not the steps which report the result

EmilyBourne:
> Looks good to me

I think there's one small thing missing. The tests should have triggered when I took this PR out of draft. It is important to have the correct version so you can mark it as Required. I have just fixed this

yguclu (approved):

PR #1813, opened 2024-03-29
Deprecate 'real' and 'integer' type annotations

EmilyBourne:
Deprecate 'real' and 'integer' type annotations. Fixes https://github.com/pyccel/pyccel/issues/1786

PR #2324, opened 2025-06-04
Complete deprecation of real as a type annotation

### LINKED ISSUE #2323: [Improvement] Deprecate `real` in type annotations

Following the Fortran language, Pyccel used to accept the term `real` as a synonym for `float`. We should remove it from our codebase and deprecate any use of it.

### Do you have any suggestions for improvement? (Optional)

_No response_

### Record

EmilyBourne:
According to the CHANGELOG `real` was deprecated as a type annotation in #1813 . In reality this annotation was still working. This is now fixed and a clean error is raised for unrecognised type annotations. Fixes #2323


yguclu (approved):
Looks good to me!

PR #2327, opened 2025-06-06
Stop modifying variables to add `Final` annotation

### LINKED ISSUE #2167: Error INTENT definition in function with function and array as argument
### Describe the bug
When implementing the function for finding the root of an equation using Newton's method, it is necessary to pass the function as an argument. It is convenient for the arguments of this function (except one) to be passed through a variable in the form of an array (in Python, this is done through a tuple). In the code below, after the pyccel command, an error is generated.

Error: Interface mismatch in dummy procedure ‘f’ at (1): INTENT mismatch in argument 'dummy_0001'.

### To Reproduce
Python file:
```
# pyccel func_arg.py --language fortran

import numpy as np

#%% Constants

TINY1 = 2.2250738585072014e-308 # np.finfo(np.float64).tiny
MAX_ITERATIONS1 = 100
EPS1 = 1.e-8

#%% Functions

def newton_halfdiv(f: '(float)(float, float[:])', args: 'float[:]',
                   x1: float, x2: float) -> float:
    if (x1 > x2):
        xl1 = x2
        xr = x1
        xl = xl1
    else:
        xl = x1
        xr = x2
    dx1 = 1.e-6
    fl0 = f(xl, args)
    fr0 = f(xr, args)
    if (fl0 * fr0 > 0):
        print("ERROR in GetRoot_newton: no root (f(x_left) * f(x_right) > 0)")
        return -1.
    xm = 0.5 * (xl + xr)
    for i in range(MAX_ITERATIONS1):
        fm0 = f(xm, args)
        fm1 = f(xm + dx1, args)
        dfm = (fm1 - fm0) / dx1
        dx = fm0 / (dfm + TINY1 * np.sign(dfm)) # avoid division by 0
        root = xm - dx
        if ((root < xl) or (root > xr)):
            if (fl0 * fm0 > 0.):
                xl = xm
                fl0 = fm0
            else:
                xr = xm
                fr0 = fm0
            if (np.abs(xr - xl) < EPS1):
                root = xm
                return root
            xm = 0.5 * (xl + xr)
        else:
          xm = root
    return root

def f1(x: float, args: 'float[:]') -> float:
    res = x * args[0] - args[1] * np.cos(x)
    return res

#%% Program

args1 = np.array([1., 1.], dtype = np.float64)
x0 = newton_halfdiv(f1, args1, 0., np.pi / 2.)

#%% End
```

### Error details
```
$ pyccel func_arg.py --language fortran --debug

ERROR at compilation stage
  File "/home/vadter/Programs/conda3/bin/pyccel", line 8, in <module>
    sys.exit(pyccel())
             ^^^^^^^^
  File "/home/vadter/Programs/conda3/lib/python3.12/site-packages/pyccel/commands/console.py", line 266, in pyccel
    execute_pyccel(filename,
  File "/home/vadter/Programs/conda3/lib/python3.12/site-packages/pyccel/codegen/pipeline.py", line 378, in execute_pyccel
    src_compiler.compile_module(compile_obj=mod_obj,
  File "/home/vadter/Programs/conda3/lib/python3.12/site-packages/pyccel/codegen/compiling/compilers.py", line 355, in compile_module
  File "/home/vadter/Programs/conda3/lib/python3.12/site-packages/pyccel/codegen/compiling/compilers.py", line 500, in run_command
    raise RuntimeError(err_msg)
/home/vadter/Documents/!Docs/mathprog/Python/PYCCEL/7_tuples/__pyccel__/func_arg.f90:130:26:

  130 |       x0 = newton_halfdiv(f1, args1, 0.0_f64, 3.141592653589793_f64 / &
      |                          1
Error: Interface mismatch in dummy procedure ‘f’ at (1): INTENT mismatch in argument 'dummy_0001'
```

As I understand it, in the Fortran file func_arg.f90 in the pyccel folder, the function f1 in the interface for the args (Dummy_0001) argument is described as intent(inout), but in the function there is intent(in).

### Expected behavior
Same intent definition in func_arg.f90

### Language
Fortran

### Additional context
The method with an array instead of arguments seems to me the most convenient in this case. Maybe I'm wrong and it could be done more elegantly?

max-models:
Closes #2167 

I removed the block of code that attempted to determine which function arguments were modified and should be marked as `inout`. If the user manually marks an argument with `const`, then it will still be `intent:in`, but otherwise `intent:inout` is always used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -168,7 +168,7 @@ All notable changes to this project will be documented in this file.
 -   #1668 : Fix handling of `is not None` check to ensure it is always checked before accessing the variable.
 -   #802 : Add if blocks in Python output to ensure support for implementations that differ for different types.
 -   Fix casting of arrays in Python translation.
--   #2167 : Do not automatically treat variables as input-only.
+-   #2167 : Stop modifying variables to add `Final` annotation.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -168,7 +168,7 @@ All notable changes to this project will be documented in this file.
 -   #1668 : Fix handling of `is not None` check to ensure it is always checked before accessing the variable.
 -   #802 : Add if blocks in Python output to ensure support for implementations that differ for different types.
 -   Fix casting of arrays in Python translation.
--   #2167 : Fix error in INTENT in/inout
+-   #2167 : Do not automatically set variables to intent in.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -168,7 +168,7 @@ All notable changes to this project will be documented in this file.
 -   #1668 : Fix handling of `is not None` check to ensure it is always checked before accessing the variable.
 -   #802 : Add if blocks in Python output to ensure support for implementations that differ for different types.
 -   Fix casting of arrays in Python translation.
--   #2167 : Do not automatically set variables to intent in.
+-   #2167 : Do not automatically treat variables as input-only.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -168,6 +168,7 @@ All notable changes to this project will be documented in this file.
 -   #1668 : Fix handling of `is not None` check to ensure it is always checked before accessing the variable.
 -   #802 : Add if blocks in Python output to ensure support for implementations that differ for different types.
 -   Fix casting of arrays in Python translation.
+-   #2167 : Fix error in INTENT in/inout
 
 ### Changed
 
--- a/tests/codegen/fcode/scripts/macros.py
+++ b/tests/codegen/fcode/scripts/macros.py
@@ -1,4 +1,4 @@
-# pylint: disable=missing-function-docstring, missing-module-docstring, undefined-variable
+# pylint: disable=missing-function-docstring, missing-module-docstring
 from numpy import zeros
 
 # .....................................
@@ -15,11 +15,9 @@ def f1(n):
     x = zeros(n, 'int')
     __f(x)
 
-# TODO not working yet, x is intent(inout)
-#      but it is not inferred as inout
 #$ header function f2(int [:])
 def f2(x):
-    __f(x)
+    __f(x) # pylint: disable=undefined-variable
 # .....................................
 
 # .....................................
--- a/tests/codegen/fcode/scripts/macros.py
+++ b/tests/codegen/fcode/scripts/macros.py
@@ -19,9 +19,8 @@ def f1(n):
 #      but it is not inferred as inout
 #$ header function f2(int [:])
 def f2(x):
-   __f(x)
+    __f(x)
 # .....................................
-
 
 # .....................................
 #           2d case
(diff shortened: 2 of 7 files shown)


max-models:
Tested with  `mwe_2167.py`

```python
def f(x: 'float[:]') -> float:
    res = x[1] - x[0]
    return res

def g(x: 'const float[:]') -> float:
    res = x[1] - x[0]
    return res
```

Translate with `pyccel mwe_2167.py --language fortran`:

`__pyccel__/mwe_2167.f90`:

```fortran
module mwe_2167

        C_DOUBLE

  implicit none

  public :: f
  public :: g

  private

  contains

  function f(x) result(res)

    implicit none

    real(f64) :: res
    real(f64), intent(inout) :: x(0_i64:)

    res = x(1_i64) - x(0_i64)
    return

  end function f

  function g(x) result(res)

    implicit none

    real(f64) :: res
    real(f64), intent(in) :: x(0_i64:)

    res = x(1_i64) - x(0_i64)
    return

  end function g

end module mwe_2167
```

So `x` is now `intent(inout)` unless explicitly specified as const (as in `g(x)`)

EmilyBourne (approved):
Looks good to me :+1:

CHANGELOG.md
```diff
@@ -168,6 +168,7 @@ All notable changes to this project will be documented in this file.
 -   #1668 : Fix handling of `is not None` check to ensure it is always checked before accessing the variable.
 -   #802 : Add if blocks in Python output to ensure support for implementations that differ for different types.
 -   Fix casting of arrays in Python translation.
+-   #2167 : Do not automatically set variables to intent in.
```
EmilyBourne, line 171: I am not sure about this description. "intent in" is only a Fortran thing, but the problem was only a Fortran thing so maybe it is ok :thinking: 

An opinion @yguclu ?

max-models:
Why is the linting taking forever?

CHANGELOG.md
max-models: How about "Do not automatically treat variables as input-only"?

EmilyBourne:
> Why is the linting taking forever?

Because I left a typo in the workflow and the fix that I put on another branch hasn't been merged. If you click through you can see that it's finished. Pyccel_lint passed, pylint didn't (due to macros.py).

You can disable the pylint message in that file with `#pylint:disable=undefined-variable`

I'm working on the clean version

CHANGELOG.md
max-models: Ok, sounds fine to me.

max-models: Fixed in [d213c3c](https://github.com/pyccel/pyccel/pull/2327/commits/d213c3c52cc96fb810b6ec65eff9b66de12d258f)

EmilyBourne:
Don't forget to take the branch out of draft once your happy

tests/codegen/fcode/scripts/macros.py
```diff
@@ -1,4 +1,4 @@
-# pylint: disable=missing-function-docstring, missing-module-docstring
+# pylint: disable=missing-function-docstring, missing-module-docstring, undefined-variable
```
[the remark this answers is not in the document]
max-models: Isn't "it's written in rust" a sufficient reason to call it superior? 😅 You are right though, there are still many things from pylint that are not covered, I like it because it replaces black + isort and it is very very fast, the ruff formatter works great too.

PR #2332, opened 2025-06-09
Remove code and docs referring to pickling

EmilyBourne:
Remove docs about obsolete pickling step. The pickling was removed a while ago (in an unreleased version) as it was not helpful but the code still contains a few references to pickling. More critically the installation was still installing `pyccel-init` which didn't point to a module that exists. The docs also contained several references to pickling.

The **installation** test in the CI was equally broken as the test workflow was not renamed from pickle.yml. As the CI chooses the test using the filename this had to be updated. However the workflow cannot be called until it is merged to devel


EmilyBourne:
Installation cannot be checked as bot calling of installation tests is broken. This PR fixes this. I don't expect new installation problems

yguclu (approved):
👍

PR #2337, opened 2025-06-11
Correct base for workflows triggered without inputs

EmilyBourne:
Correct base for workflows triggered without inputs. Base is an input (usually devel) but is not provided when the workflow is triggered by taking it out of draft. This leads to an empty diff so no errors are reported.


yguclu (commented):
When does it happen that the `base` branch is undefined? When the `pull_request` event type is `ready_for_review`?

yguclu:
> > When does it happen that the `base` branch is undefined? When the `pull_request` event type is `ready_for_review`?
> 
> Yes exactly. I noticed this in the OpenMP PR which was marked as ready and passed tests but shouldn't have passed the doc tests

Maybe we should point this out to the developers of GitHub Actions. It looks like a bug to me

PR #2340, opened 2025-06-12
Deprecate "const" modifier

### LINKED ISSUE #2339: [Deprecation] Remove support for const in favour of Final

Remove support for const in favour of Final in type annotations. E.g:
```python
def f(a : 'const int[:]'):
    pass
```
should always be written:
```python
def f(a : 'Final[int[:]]'):
    pass
```
This is important as:
```python
TypeVar('T', 'const int[:]', 'const float[:]')
```
raises:

Image

This indicates that we should move away from annotations with spaces

### Record

EmilyBourne:
Deprecate un Pythonic const annotation in favour of Python's `Final`. Fixes #2339

**Commit Summary**
- Bug fix in CI. Reference the base SHA which is more precise than the ref
- Remove docs about `const` keyword. `Final` is documented by Python
- Remove `const` from the grammar
- Update error messages
- Use `Final` in fftw and bug fix some nonsense type decorators
- Use `Final` instead of `const` in tests

### DIFF
--- a/tests/errors/semantic/non_blocking/CONST_ASSIGNED_ARGUMENT2.py
+++ b/tests/errors/semantic/non_blocking/CONST_ASSIGNED_ARGUMENT2.py
@@ -1,5 +1,5 @@
 # pylint: disable=missing-function-docstring, missing-module-docstring
-from typing import Final #pylint: disable=unused-import
+from typing import Final
 
 def array_int32_1d_add_const( x : 'Final[int32[:]]', y : 'int32[:]' ):
     x[:] += y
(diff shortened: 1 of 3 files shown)


tests/errors/semantic/non_blocking/CONST_ASSIGNED_ARGUMENT2.py
```diff
@@ -1,4 +1,5 @@
 # pylint: disable=missing-function-docstring, missing-module-docstring
+from typing import Final #pylint: disable=unused-import
 
-def array_int32_1d_add_const( x : 'const int32[:]', y : 'int32[:]' ):
+def array_int32_1d_add_const( x : 'Final[int32[:]]', y : 'int32[:]' ):
```
EmilyBourne, line 4:
```suggestion
def array_int32_1d_add_const( x : Final['int32[:]'], y : 'int32[:]' ):
```

yguclu (approved):
Looks good to me

PR #2344, opened 2025-06-13
Allow language to be set using a capitalised name (Fortran, C, Python)

EmilyBourne:
Allow language to be set using a capitalised name (Fortran, C, Python) via both Pyccel's command line tool and `epyccel`.


yguclu (approved):
Looks good to me

PR #2345, opened 2025-06-13
Deprecate macros

### LINKED ISSUE #960: Macro default arguments can be expressions
**Describe the bug**
Currently only a small subset of expressions are allowed in macros. This subset needs expanding slightly in order to be able to describe scipy correctly. For example the lapack function [`dgetri`](http://www.netlib.org/lapack/explore-html/dd/d9a/group__double_g_ecomputational_ga56d9c860ce4ce42ded7f914fdb0683ff.html) is defined by [scipy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.lapack.dgetri.html) as follows:
```
Parameters:
lu : input rank-2 array(‘d’) with bounds (n,n)
piv : input rank-1 array(‘i’) with bounds (n)

Returns:
inv_a : rank-2 array(‘d’) with bounds (n,n) and lu storage
info : int

Other Parameters:
overwrite_lu : input int, optional
Default: 0

lwork : input int, optional
Default: max(3*n,1)
```

Currently the default value for the final argument `lwork` cannot be specified. A textx error is raised:
```
textx.exceptions.TextXSyntaxError: None:1:55: error: Expected ',' or ')' at position (1, 55) => 'v, lwork=3**a.shape[0'.
```

**Additional information**
I have seen the following expressions used by scipy:
- Multiplication
- Max (often `max(1,N)` it is perhaps sufficient to set this to `N`)

### LINKED ISSUE #961: acros need default
**Describe the bug**
A clear and concise description of what you would like to be implemented.

**Test Code**
```python
Code here
```

**Language**
If appropriate, please specify which language the python code is translated to (Fortran by default)

### LINKED ISSUE #1372: [COVERAGE] : Macro functions
### Untested code

The untested code is : https://github.com/pyccel/pyccel/blob/48df47e8006cb05aa10857325d09b7f6cbaf9dd5/pyccel/parser/semantic.py#L2410-L2417
The problem was discovered in PR : https://github.com/pyccel/pyccel/pull/1353

EmilyBourne:
Remove macros. Fixes #1372 . Fixes #961 . Fixes #960 
This PR does not deal with the existing macros so it breaks the use of the 4 functions from `scipy.linalg.lapack` which had partial support (which didn't match the scipy interface).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -240,7 +240,7 @@ All notable changes to this project will be documented in this file.
 -   #1487 : Remove support for `#$ header` syntax.
 -   #1812 : Stop allowing multiple main blocks inside a module.
 -   Removed `templates` argument from `lambdify`. Replaced with `context_dict`.
--   #2339 : Remove `const` type modifier which cannot be used in objects such as `TypeVar`.
+-   #2339 : Remove `const` type modifier in favour of `typing.TypeVar`.
 -   #1287 : Remove unused method `BasicParser.copy()`.
 -   #2345 : Remove undocumented macro syntax in favour of inline methods.
 -   #2345 : Break support for `scipy.linalg.lapack.dgbtrf`, `scipy.linalg.lapack.dgbtrs`, `scipy.linalg.lapack.dgetrf`, and `scipy.linalg.lapack.dgetrs`.


CHANGELOG.md
```diff
@@ -240,8 +240,10 @@ All notable changes to this project will be documented in this file.
 -   #1487 : Remove support for `#$ header` syntax.
 -   #1812 : Stop allowing multiple main blocks inside a module.
 -   Removed `templates` argument from `lambdify`. Replaced with `context_dict`.
--   #2339 : Deprecated `const` type modifier which cannot be used in objects such as `TypeVar`.
+-   #2339 : Remove `const` type modifier which cannot be used in objects such as `TypeVar`.
 -   #1287 : Remove unused method `BasicParser.copy()`.
+-   #2345 : Remove undocumented macro syntax in favour of inline methods.
+-   #2345 : Broken support for `scipy.linalg.lapack.dgbtrf`, `scipy.linalg.lapack.dgbtrs`, `scipy.linalg.lapack.dgetrf`, and `scipy.linalg.lapack.dgetrs`.
```
yguclu, line 246: If you use the imperative form everywhere, this should read
```suggestion
-   #2345 : Break support for `scipy.linalg.lapack.dgbtrf`, `scipy.linalg.lapack.dgbtrs`, `scipy.linalg.lapack.dgetrf`, and `scipy.linalg.lapack.dgetrs`.
```

CHANGELOG.md
```diff
@@ -240,8 +240,10 @@ All notable changes to this project will be documented in this file.
 -   #1487 : Remove support for `#$ header` syntax.
 -   #1812 : Stop allowing multiple main blocks inside a module.
 -   Removed `templates` argument from `lambdify`. Replaced with `context_dict`.
--   #2339 : Deprecated `const` type modifier which cannot be used in objects such as `TypeVar`.
+-   #2339 : Remove `const` type modifier which cannot be used in objects such as `TypeVar`.
```
yguclu, line 243: I would explain more simply
```suggestion
-   #2339 : Remove `const` type modifier in favour of `typing.TypeVar`.
```

pyccel/parser/semantic.py
```diff
@@ -3871,56 +3841,9 @@ def _visit_Assign(self, expr):
...
-                    else:
-                        return func_call
-                else:
-                    # TODO treat interface case
-                    errors.report(PYCCEL_RESTRICTION_TODO,
-                                  bounding_box=(self.current_ast_node.lineno, self.current_ast_node.col_offset),
-                                  severity='fatal')
+            rhs = self._visit(rhs)
+            if isinstance(rhs, (PythonMap, PythonZip, PythonEnumerate, PythonRange)):
+                errors.report(f"{type(rhs)} cannot be saved to variables", symbol=expr, severity='fatal')
```
yguclu, line 3846: This check looks very general to me, so why adding it only to the case where `rhs` is the result of a `FunctionCall`?

EmilyBourne: Hmm, I didn't add anything in this PR so if it's showing up as new it is simply because the indentation has changed. But I would think that the listed objects can only be created by a `FunctionCall` hence why it is here.

Even if you do
```python
return range(10)
```
Internally we would understand that as
```python
result = range(10)
return result
```
And would still end up at this line

yguclu: Sorry, I really thought that you had added these lines in this PR!

Thanks for the explanation anyway!

yguclu (approved):
Looks good to me!

PR #2346, opened 2025-06-15
Add verbosity levels to Pyccel

### LINKED ISSUE #2253: Improve output of `pyccel --verbose` command
### Current Status

When running `pyccel --verbose hello.py` on the CLI, Pyccel prints the compilation commands but does not provide any further information.

### Proposal for Improvement

We can probably think of three levels of verbosity:

0. No output is printed, as is done now without the `--verbose` flag;
1. Print at least some basic information about what Pyccel is doing, and what files it is generating (this needs a discussion);
2. In addition to what is printed at level 1, print the full compiler commands as is done now with the `--verbose` flag.

EmilyBourne:
Add four levels of verbosity to Pyccel:

0. No output is printed, as is done now without the --verbose flag;
1. Print basic information about which files Pyccel is treating at which stage, and what files it is generating;
2. In addition to what is printed at level 1, print the full compiler commands as is done now with the --verbose flag.
3. In addition to what is printed at level 2, print the detail of the `_visit`, `_print`, and `_wrap` calls.

Fixes #2253


yguclu (changes requested):
Great job!

I only have a minor suggestion for improving a docstring.

PR #2351, opened 2025-06-19
Fix missing import

EmilyBourne:
When creating the release branch I noticed an import which was accidentally deleted although it is used in the file. This PR puts it back.

### DIFF
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -52,7 +52,7 @@ from pyccel.ast.low_level_tools import IteratorType, MemoryHandlerType, ManagedM
 
 from pyccel.ast.mathext  import math_constants
 
-from pyccel.ast.numpyext import NumpyFull, NumpyArray, NumpySum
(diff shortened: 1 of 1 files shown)


EmilyBourne:
> I do not understand. All tests are passing on `devel`...

That is just a coverage issue. The object is used here:
https://github.com/pyccel/pyccel/blob/devel/pyccel/codegen/printing/ccode.py#L2795

This is the fallback case for `Constant` which we do not use currently and which will either be removed with #1811 or will start being used and tested

EmilyBourne:
I'm surprised Codacy didn't catch it though

yguclu (approved):

yguclu:
> > I do not understand. All tests are passing on `devel`...
> 
> That is just a coverage issue. The object is used here: https://github.com/pyccel/pyccel/blob/devel/pyccel/codegen/printing/ccode.py#L2795
> 
> This is the fallback case for `Constant` which we do not use currently and which will either be removed with #1811 or will start being used and tested

Could you add a word about this to the PR description?

PR #2354, opened 2025-06-19
Fix Import regression

### LINKED ISSUE #2353: [Bug] ImportError for submodules

### Describe the bug

Pyccel raises a semantic error when trying to compile a script that imports a previously compiled submodule.

**Example**:

a/foo.py:
```python
def f() -> float:
    return 1.0
```

bar.py:

```python
import a.foo as foo

def g():
    x = foo.f()
```

Testing:

```bash
❯ pyccel a/foo.py
❯ pyccel bar.py
pyccel:
 |fatal [semantic]: bar.py [4,8]| Trying to import an identifier that already exists in the namespace. Hint: use import as (foo)
```

This works with the previous release of pyccel. I don't really understand the hint, am I importing `foo` incorrectly?

This issue seems similar to https://github.com/pyccel/pyccel/issues/1500, but there the error occurs during the syntax stage.

If I change bar.py like this, then compiling works:

```python
import a.foo as imported_foo

def g():
    x = imported_foo.f()
```

### Expected behavior

I expect to be able to import a module with `import a.b.c as c`, this is what we do in Struphy. Example: https://gitlab.mpcdf.mpg.de/struphy/struphy/-/blob/devel/src/struphy/pic/pushing/pusher_kernels.py?ref_type=heads#L6-9

### Add Screenshots

_No response_

### Record

EmilyBourne:
Previously imports like `import a.b.c as d` were saved as `d` however this makes it difficult to reuse parsing when 2 files use `a.b.c` with different names. This was therefore changed so the key is now `a.b.c`. The case `import a.b.c as d` was correctly handled, but `import a.b.c as c` was not as it was previously a special case. The special case has now been corrected and a test is added. Fixes #2353

### DIFF
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -587,7 +587,7 @@ class SemanticParser(BasicParser):
             found_from_import_name = True
 
         if imp is not None:
-            if found_from_import_name or imp.source == source:
+            if found_from_import_name or imp.source_module.name == source:
                 imp.define_target(target)
             else:
                 errors.report(IMPORTING_EXISTING_IDENTIFIED,
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -587,7 +587,7 @@ class SemanticParser(BasicParser):
             found_from_import_name = True
 
         if imp is not None:
-            if found_from_import_name or imp.source_module.name == source:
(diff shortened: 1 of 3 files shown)


yguclu (approved):
Looks good to me!

max-models:
That was quick! @EmilyBourne Thank you solving the issue!

PR #2352, opened 2025-06-19
Version 2.0.0

EmilyBourne:
Prepare the release of v2.0.0

PR #2358, opened 2025-06-19
Add support for bitwise operators with NumPy arrays

EmilyBourne:
Add support for bitwise operators with NumPy arrays. This improves coverage on sections that were introduced for this purpose but were never fully implemented and tested. This is necessary for the tests to pass on #2352

### DIFF
--- a/pyccel/ast/numpytypes.py
+++ b/pyccel/ast/numpytypes.py
@@ -321,7 +321,6 @@ class NumpyNDArrayType(HomogeneousContainerType, metaclass = ArgumentSingleton):
     @lru_cache
     def __and__(self, other):
         elem_type = self.element_type
-        assert isinstance(elem_type, PythonNativeBool)
         out_type = NumpyNDArrayType(elem_type, self.rank, self.order)
         if isinstance(other, FixedSizeNumericType):
             out_type.switch_basic_type(elem_type and other)
--- a/pyccel/ast/numpytypes.py
+++ b/pyccel/ast/numpytypes.py
@@ -321,10 +321,14 @@ class NumpyNDArrayType(HomogeneousContainerType, metaclass = ArgumentSingleton):
     @lru_cache
     def __and__(self, other):
         elem_type = self.element_type
+        assert isinstance(elem_type, PythonNativeBool)
+        out_type = NumpyNDArrayType(elem_type, self.rank, self.order)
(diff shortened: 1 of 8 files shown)


EmilyBourne:
/bot run pylint pyccel_lint docs spelling linux

pyccel/ast/numpytypes.py
```diff
@@ -322,9 +322,9 @@ def __radd__(self, other):
     def __and__(self, other):
         elem_type = self.element_type
         if isinstance(other, FixedSizeNumericType):
-            return NumpyNDArrayType(elem_type and other)
+            return self.switch_basic_type(elem_type & other)
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
Tested in Python3.10 and up

tests/epyccel/recognised_functions/test_numpy_operators.py
```diff
@@ -0,0 +1,204 @@
+# pylint: disable=missing-function-docstring, missing-module-docstring
+import os
+from typing import TypeVar
+from numpy.random import randint
+import numpy as np
+import pytest
```
yguclu, line 6: I am surprised that Pylint does not complain about not separating the standard library imports from the third-party ones... 🤷‍♂️

tests/epyccel/recognised_functions/test_numpy_operators.py
```diff
@@ -0,0 +1,204 @@
...
+        return a & b
+
+    epyc_f = epyccel(f, language=language)
+    epyc_g = epyccel(g, language=language)
+    epyc_h = epyccel(h, language=language)
+
+    for t_x in int_types:
+        for t_y in int_types:
+            x = randint(2 if t_x is bool else 127, size=(2,3,4), dtype = t_x)
+            y = randint(2 if t_y is bool else 127, size=(2,3,4), dtype = t_y)
```
yguclu, line 29: Should we check a case that requires broadcasting? Here we could set to `1` any of the dimensions of the array `y`.

yguclu: Indeed, we could drop the first dimension.

Not related to this PR: why is the case of 1 not working, given that we can already drop one dimension? In any case we must check the dimensions of the arrays at runtime

yguclu (approved):
Looks good to me!

PR #2362, opened 2025-06-21
Ensure submodules are present when installing pyccel

EmilyBourne:
Ensure submodules are present and environment is clean before building `gFTL` during installation


yguclu (approved):
LGTM

PR #2374, opened 2025-06-26
Add links for the PyPi page

EmilyBourne:
Add links for the PyPi page in the `pyproject.toml` file.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,6 +43,10 @@ pyccel-test = "pyccel.commands.pyccel_test:pyccel_test_command"
 
 [project.urls]
 Homepage = "https://github.com/pyccel/pyccel"
+Documentation = "https://pyccel.github.io/pyccel/"
+Source = "https://github.com/pyccel/pyccel"
+Tracker = "https://github.com/pyccel/pyccel/issues"
+"Release notes" = "https://github.com/pyccel/pyccel/blob/devel/CHANGELOG.md"
 
 [tool.hatch.version]
 path = "pyccel/version.py"


pyproject.toml
```diff
@@ -43,6 +43,10 @@ pyccel-test = "pyccel.commands.pyccel_test:pyccel_test_command"
 
 [project.urls]
 Homepage = "https://github.com/pyccel/pyccel"
+Documentation = "https://pyccel.github.io/pyccel/"
+Source = "https://github.com/pyccel/pyccel"
+Tracker = "https://github.com/pyccel/pyccel/issues"
+"Release notes" = "https://github.com/pyccel/pyccel/blob/devel/CHANGELOG.md"
```
yguclu, line 49: We will have to update this link when we make a new release... I suppose that could be done automatically? Maybe we just need to have a "release checklist" to remind us of these small things.

EmilyBourne: Do we need to? On the devel branch new developments are explicitly listed in a "Unreleased" section so this should still be a valid link for a release

EmilyBourne: Otherwise it should be possible to do something with an install hook: https://hatch.pypa.io/latest/how-to/config/dynamic-metadata/#implement-hook

to have a dynamic value that depends on the version (as the tag)

EmilyBourne: Alternatively it seems we can link to the CHANGELOG which probably makes more sense as we don't have a compact Release Notes anywhere other than linkedin
image

EmilyBourne: Done [8a92e8f](https://github.com/pyccel/pyccel/pull/2374/commits/8a92e8fd40b5bc01bd2344af401ad2d56d344742)

yguclu: The link I was suggesting gives information for all the releases. One just needs to click on a previous version once they get there! And it also allows them to diff between releases...

Anyway, I also like your last solution which is probably the simplest.

PR #2378, opened 2025-07-02
Rename `set_variable` to `set_obj`

EmilyBourne:
Rename `set_variable` to `set_obj` to make it clear that this object is not always a `Variable`.


pyccel/ast/builtin_methods/set_methods.py
yguclu, line 1: Since we are looking at renaming variables, I might as well ask here. Why is `*args` sometimes renamed as `*others`?

PR #2402, opened 2025-07-23
Use absolute paths to avoid move command

EmilyBourne:
Use absolute paths to avoid move command for the final shared library and program. This makes testing easier as the outputted compiler commands create objects directly in their expected locations.

Fix anaconda CI by using `pip install .[test]` to ensure all pytest plugins are installed.


yguclu (approved):
Looks good to me!

PR #2407, opened 2025-07-30
Fix missing undef in `Managed_memory.h`

EmilyBourne:
The file `Managed_memory.h` defined the macro `i_is_arc` which affects STC's internals but it forgot to undefine this macro at the end of the file. This is easy to fix but the errant `i_is_arc` in the program lead to multi-level containers stealing references as they did not create copy methods. This is the desired behaviour as we write our code in such a way that the initialiser is passed a Managed_memory object with the correct ownership. However undefining the macro meant that we retrieved the default behaviour where `_clone` is called for class arguments of containers. This PR therefore additionally defines a `_steal` method which is used by containers to steal references to arguments. It is then the responsibility of the printer to ensure that the reference count is accurate.

### DIFF
--- a/.github/actions/valgrind_run/action.yml
+++ b/.github/actions/valgrind_run/action.yml
@@ -5,20 +5,20 @@ runs:
   steps:
     - name: Test with valgrind for memory leaks in ndarrays
       run: |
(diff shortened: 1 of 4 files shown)


yguclu (approved):
Looks good to me!

PR #2411, opened 2025-08-04
Fix bug when returning an expression that is translated to multiple lines

### LINKED ISSUE #2406: [Bug] Handling of `&` for sets in return

### Describe the bug

The following fails to compile:
```python
def f(a : set[int], b : set[int]):
    return a & b
```

The generated code is:
```c
hset_int64_t f(hset_int64_t* a, hset_int64_t* b)
{
    hset_int64_t_reserve(&result, hset_int64_t_size(a));
    hset_int64_t_intersection_update(&result, b);
    return hset_int64_t_clone((*a));
}
```
The error is:
```
/home/emily/Code/test/tmp/__pyccel__/tmp_set.c: In function ‘f’:
/home/emily/Code/test/tmp/__pyccel__/tmp_set.c:7:27: error: ‘result’ undeclared (first use in this function); did you mean ‘_m_result’?
    7 |     hset_int64_t_reserve(&result, hset_int64_t_size(a));
      |                           ^~~~~~
      |                           _m_result
/home/emily/Code/test/tmp/__pyccel__/tmp_set.c:7:27: note: each undeclared identifier is reported only once for each function it appears in
```

### Expected behavior

The expected code is:
```c
hset_int64_t f(hset_int64_t* a, hset_int64_t* b)
{
    hset_int64_t result = {0};
    hset_int64_t_reserve(&result, hset_int64_t_size(a));
    result = hset_int64_t_clone((*a));
    hset_int64_t_intersection_update(&result, b);
    return result;
}
```

### Add Screenshots

_No response_

### Record

EmilyBourne:
Fix bug when returning an expression that is translated to multiple lines. Fixes #2406 

The code tries to avoid printing unnecessary assignments (e.g `return 1` should be `return 1;` not `result = 1; return result;`). However the case of `a & b` was not taken into account. This translates to multiple expressions (Allocate, Assign, SetIntersectionUpdate). As the assignment is not the last element it shouldn't be removed from the print.

### DIFF
--- a/pyccel/codegen/printing/pycode.py
+++ b/pyccel/codegen/printing/pycode.py
@@ -650,6 +650,9 @@ class PythonCodePrinter(CodePrinter):
             return 'return\n'
 
         if expr.stmt:
+            to_print = [l for l in expr.stmt.body \
+                            if not ((isinstance(l, Assign) and isinstance(l.lhs, Variable) and l.lhs in result_vars)
+                                     or isinstance(l, UnpackManagedMemory))]
             # Get expressions that should be printed as they are. Assignments to result variables are not
             # printed as the rhs can be inlined
             if any(expr.stmt.body.index(p) < expr.stmt.body.index(a) for p in to_print for a in expr.stmt.body if isinstance(a, Assign)):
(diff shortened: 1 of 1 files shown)


pyccel/codegen/printing/ccode.py
```diff
@@ -2437,11 +2437,11 @@ def _print_Return(self, expr):
                 # make sure that stmt contains one assign node.
                 last_assign = last_assign[-1]
                 variables = last_assign.rhs.get_attribute_nodes(Variable)
-                unneeded_var = not any(b in vars_in_deallocate_nodes or b.is_ndarray for b in variables) and \
+                unneeded_var = not any(b in vars_in_deallocate_nodes or b.rank for b in variables) and \
```
yguclu: Could you please explain this change? Does this allow handling sets in addition to arrays?

EmilyBourne: Yes exactly. This enables handling sets, dits, and lists as well as arrays. In other words it now handles any object which needs deallocating

PR #2412, opened 2025-08-05
Remove dead code

EmilyBourne:
Remove unused undocumented decorators:
- `sympy`
- `bypass`

Remove code associated with the `sympy` decorator. This code populates `SympyFunction` objects which then remain unused. The class is therefore removed along with the code that creates these objects.

### DIFF
--- a/pyccel/decorators.py
+++ b/pyccel/decorators.py
@@ -21,12 +21,6 @@ __all__ = (
 )
 
 
-def sympy(f):
-    return f
-
-def bypass(f):
-    return f
-
 def pure(f):
     return f
 
--- a/pyccel/decorators.py
+++ b/pyccel/decorators.py
@@ -10,13 +10,11 @@ import warnings
 
 __all__ = (
     'allow_negative_index',
-    'bypass',
     'elemental',
     'inline',
     'private',
     'pure',
     'stack_array',
-    'sympy',
     'types',
 )
 
--- a/pyccel/ast/core.py
+++ b/pyccel/ast/core.py
@@ -86,7 +86,6 @@ __all__ = (
     'Return',
     'SeparatorComment',
     'StarredArguments',
-    'SympyFunction',
     'While',
     'With',
 )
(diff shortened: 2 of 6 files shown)


pyccel/decorators.py
```diff
@@ -10,66 +10,135 @@
...
+    """
     return f
 
 def elemental(f):
+    """
+    Indicate that the function can also be used element-wise.
+
+    This decorator indicates that the function can be applied element-wise to
+    array arguments. The function operates on scalar inputs and the wrapper
+    will be provided for the annotatied arguments, but in the low-level code
```
yguclu, line 69: Typo: annotatied -> annotated
```suggestion
    will be provided for the annotated arguments, but in the low-level code
```

pyccel/decorators.py
```diff
@@ -10,66 +10,135 @@
...
     return f
 
 def inline(f):
-    """Indicates that function calls to this function should
-    print the function body directly"""
+    """
+    Indicate that the function should be inlined in the low-level code.
+
+    This decorator indicates that the function should be inlined where it is
+    called. Calls are replace with the function body.
```
yguclu, line 89: Typo: replace -> replaced. And maybe we can give more explanation:
```suggestion
    called. Calls are replaced with the function body, where the
    arguments are substituted for the function parameters.
```

pyccel/decorators.py
```diff
@@ -10,66 +10,135 @@
...
     def identity(f):
         return f
     return identity
 
-def allow_negative_index(f,*args):
+def allow_negative_index(*args):
     """
+    Indicate that arrays can be accessed with negative indexes.
+
     Decorator indicates that all arrays mentioned as args can be accessed with
```
yguclu, line 128: Maybe the word "This" is missing?
```suggestion
    This decorator indicates that all arrays mentioned as args can be accessed with
```
Or otherwise,
```suggestion
    Decorator indicating that all arrays mentioned as args can be accessed with
```

yguclu (changes requested):
Looks good to me!

I have just three minor comments

PR #2414, opened 2025-08-12
Ensure printing of imports from Fortran intrinsic libraries is reproducible

EmilyBourne:
Currently the printing of imports from Fortran intrinsic libraries prints the imported elements from a set. As a result their order is random leading to slightly different Fortran files being generated. This PR sorts these imports so they are always printed in the same order.

Fix issue when using a DottedVariable in an inline function.


yguclu:
/bot run coverage linux intel macosx windows

PR #2415, opened 2025-08-13
Incorporate type modifiers into typing system

EmilyBourne:
Some type information does not directly represent a type but simply how an object of this type behaves in certain circumstances. This is notably the case for `Final` and `Optional`. Currently this information is stored in the `Variable` class however this is not really sufficient (particularly for `Optional`) as printing of other objects may need this information (e.g. `dict.get` should be printed differently if `None` can be returned).

Incorporating this information into the datatype is therefore necessary for `Optional` but this is non-trivial as the code often relies on statements like `isinstance(dtype, NativeIntegerType)` which should return true if dtype is `Optional[NativeIntegerType]`. Many other changes are required to make `Optional` work as a general concept so in order to break up the work `Final` is first incorporated into the typing system to work as a prototype. `Optional` will then be implemented following this pattern.

In order to ensure that `isinstance(final_native_integer_type_obj, NativeIntegerType)` returns true, `FinalNativeIntegerType` must inherit from `NativeIntegerType`. In order for `FinalListNativeIntegerType` to follow a similar pattern and to know how to initialise itself it is simpler for all datatypes to be singletons. Previously things like `HomogeneousListType` were `ArgumentSingleton`s meaning that there was 1 instance for each possible argument type. In order to create a singleton with the correct inheritance a new class method `get_new` is added to build the necessary type on the fly. `@cache` is used to ensure we don't have multiple types representing the same concept.

### DIFF
--- a/pyccel/ast/datatypes.py
+++ b/pyccel/ast/datatypes.py
@@ -8,11 +8,11 @@
 """
 Classes and methods that handle supported datatypes in C/Fortran.
 """
-from functools import lru_cache
+from functools import cache
 
 import numpy
 
-from pyccel.utilities.metaclasses import Singleton, ArgumentSingleton
+from pyccel.utilities.metaclasses import Singleton
 from .basic import iterable
 
 __all__ = (
@@ -65,7 +65,7 @@ class PrimitiveType(metaclass=Singleton):
     _name = '__UNDEFINED__'
 
     def __init__(self): #pylint: disable=useless-parent-delegation
-        # This __init__ function is required so the ArgumentSingleton can
+        # This __init__ function is required so the Singleton can
         # always detect a signature
         super().__init__()
 
@@ -149,7 +149,7 @@ class PyccelType:
         return self._name
 
     def __init__(self): #pylint: disable=useless-parent-delegation
-        # This __init__ function is required so the ArgumentSingleton can
+        # This __init__ function is required so the Singleton can
         # always detect a signature
         super().__init__()
 
@@ -313,7 +313,7 @@ class PythonNativeBool(PythonNativeNumericType):
     _primitive_type = PrimitiveBooleanType()
     _precision = -1
 
-    @lru_cache
+    @cache
     def __add__(self, other):
         if isinstance(other, PythonNativeBool):
             return PythonNativeInt()
@@ -322,7 +322,7 @@ class PythonNativeBool(PythonNativeNumericType):
         else:
             return NotImplemented
 
-    @lru_cache
+    @cache
     def __and__(self, other):
         if isinstance(other, PythonNativeBool):
             return PythonNativeBool()
@@ -342,7 +342,7 @@ class PythonNativeInt(PythonNativeNumericType):
     _primitive_type = PrimitiveIntegerType()
     _precision = numpy.dtype(int).alignment
 
-    @lru_cache
+    @cache
     def __add__(self, other):
         if isinstance(other, PythonNativeBool):
             return self
@@ -351,7 +351,7 @@ class PythonNativeInt(PythonNativeNumericType):
         else:
             return NotImplemented
 
-    @lru_cache
+    @cache
     def __and__(self, other):
         if isinstance(other, PythonNativeNumericType):
             return self
@@ -370,7 +370,7 @@ class PythonNativeFloat(PythonNativeNumericType):
     _primitive_type = PrimitiveFloatingPointType()
     _precision = 8
 
-    @lru_cache
+    @cache
     def __add__(self, other):
         if isinstance(other, PythonNativeComplex):
             return other
@@ -391,7 +391,7 @@ class PythonNativeComplex(PythonNativeNumericType):
     _primitive_type = PrimitiveComplexType()
     _precision = 8
 
-    @lru_cache
+    @cache
     def __add__(self, other):
         if isinstance(other, PythonNativeNumericType):
             return self
@@ -432,7 +432,7 @@ class GenericType(FixedSizeType):
     _name = 'Generic'
     _primitive_type = None
 
-    @lru_cache
+    @cache
     def __add__(self, other):
         return other
 
@@ -750,7 +750,7 @@ class StringType(ContainerType, metaclass = Singleton):
     def __hash__(self):
         return hash(self.__class__)
 
-class HomogeneousTupleType(HomogeneousContainerType, TupleType, metaclass = ArgumentSingleton):
+class HomogeneousTupleType(HomogeneousContainerType, TupleType, metaclass = Singleton):
     """
     Class representing the homogeneous tuple type.
 
@@ -766,11 +766,16 @@ class HomogeneousTupleType(HomogeneousContainerType, TupleType, metaclass = Argu
     __slots__ = ('_element_type', '_order')
     _container_rank = 1
 
-    def __init__(self, element_type):
-        assert isinstance(element_type, PyccelType)
-        self._element_type = element_type
-        self._order = 'C' if (element_type.order == 'C' or element_type.rank == 1) else None
-        super().__init__()
+    @classmethod
+    @cache
+    def get_new(cls, element_type):
+        def __init__(self):
+            self._element_type = element_type
+            self._order = 'C' if (element_type.order == 'C' or element_type.rank == 1) else None
+            HomogeneousContainerType.__init__(self)
+
+        return type(f'HomogeneousTuple{type(element_type)}', (HomogeneousTupleType,),
+                    {'__init__' : __init__})()
 
     def __str__(self):
         return f'tuple[{self._element_type}, ...]'
@@ -795,7 +800,7 @@ class HomogeneousTupleType(HomogeneousContainerType, TupleType, metaclass = Argu
         # TODO: Remove this specialisation if tuples are saved in lists instead of ndarrays
         return isinstance(shape, tuple) and len(shape) == self.rank
 
-class HomogeneousListType(HomogeneousContainerType, metaclass = ArgumentSingleton):
+class HomogeneousListType(HomogeneousContainerType, metaclass = Singleton):
     """
     Class representing the homogeneous list type.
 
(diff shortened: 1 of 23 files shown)


pyccel/ast/datatypes.py
```diff
@@ -197,6 +199,53 @@ def shape_is_compatible(self, shape):
...
+        Get the parameterised Final type Final[underlying_type].
+
+        Parameters
+        ----------
+        underlying_type : PyccelType
+            The type which is characterised as final.
+        """
+        if isinstance(underlying_type, FinalType):
+            return underlying_type
+
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
Included to avoid `Final[Final[int]]`

.pylintrc
```diff
@@ -38,10 +38,6 @@ load-plugins=
 # Pickle collected data for later comparisons.
 persistent=yes
 
-# When enabled, pylint would attempt to guess common misconfiguration and emit
-# user-friendly hints instead of false-positive error messages.
-suggestion-mode=yes
```
EmilyBourne: This config option no longer exists. In the pyccel_lint CI it gives:
```
.pylintrc:1:0: E0015: Unrecognized option found: suggestion-mode (unrecognized-option)
```

pyccel/ast/datatypes.py
```diff
@@ -197,6 +199,55 @@ def shape_is_compatible(self, shape):
...
+        Get the parameterised Final type Final[underlying_type].
+
+        Parameters
+        ----------
+        underlying_type : PyccelType
+            The type which is characterised as final.
+        """
+        if isinstance(underlying_type, FinalType):
+            return underlying_type
```
yguclu: Should we not check that `underlying_type` is indeed a `PyccelType` instance?

pyccel/ast/datatypes.py
```diff
@@ -197,6 +199,55 @@ def shape_is_compatible(self, shape):
...
+            Get the type that is indicated as const.
+
+            Get the type that is indicated as const.
+            """
+            return self._underlying_type
+        return type(f'Final{type_class.__name__}', (FinalType, type_class,),
+                    {'__init__' : __init__,
+                     '__hash__' : __hash__,
+                     '__eq__' : __eq__,
+                     'underlying_type': property(get_underlying_type)})()
```
yguclu: Are we not missing the square brackets in the type name?

```suggestion
        return type(f'Final[{type_class.__name__}]', (FinalType, type_class,),
                    {'__init__' : __init__,
                     '__hash__' : __hash__,
                     '__eq__' : __eq__,
                     'underlying_type': property(get_underlying_type)})()
```

Further, when I try to use this constructor passing it a subclass of `PyccelType` I get this error:
```python
TypeError: type.__new__() takes exactly 3 arguments (0 given)
```

pyccel/utilities/metaclasses.py
```diff
@@ -37,7 +37,7 @@ def __init__(cls, name, bases, dct):
 
     def __call__(cls, *args, **kwargs):
         index = (*args, *sorted(kwargs.items()))
-        existing_instance = cls._instances.get(index, None)
+        existing_instance = next((i for i in cls._instances if i is index), None)
```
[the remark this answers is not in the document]
EmilyBourne: Hmm, I forgot about this change.

> BTW, a quick search shows that this ArgumentSingleton class is only used in `pyccel/ast/low_level_tools.py` now. Maybe it is no longer needed?

I agree, it looks like the use in `pyccel/ast/low_level_tools.py` should also be updated to follow the same pattern as elsewhere.

> How is this supposed to work?

It is supposed to work as before. I think if I changed it, it must be because `.get()` uses an equality check to look for keys instead of an identity check. So `cls._instances.get(index, None)` would be equivalent to:
```python
existing_instance = next((i for i in cls._instances if i == index), None)
```
There are non-identical types which are considered equal. E.g. `PyccelNativeInt()` and `ConstType[PyccelNativeInt()]`

pyccel/ast/datatypes.py
```diff
@@ -603,7 +654,7 @@ def switch_basic_type(self, new_type):
         """
         assert isinstance(new_type, FixedSizeType)
```
yguclu: Given that this function only accepts `FixedSizeType` instances, the docstring should be updated.

pyccel/utilities/metaclasses.py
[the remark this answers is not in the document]
EmilyBourne: Ah, you are right. This probably shows that this is unused :sweat_smile:

pyccel/ast/datatypes.py
```diff
@@ -795,62 +851,90 @@ def shape_is_compatible(self, shape):
...
+        element_type : PyccelType
+            The type which is stored in the homogeneous list.
+        """
+        def __init__(self):
+            self._element_type = element_type
+            self._order = 'C' if (element_type.order == 'C' or element_type.rank == 1) else None
+            HomogeneousContainerType.__init__(self)
+
+        return type('HomogeneousList{type(element_type)}', (HomogeneousListType,),
+                    {'__init__' : __init__})()
```
yguclu: Have you forgotten the `f` in front of the formatted string?

pyccel/parser/syntactic.py
```diff
@@ -1496,7 +1496,7 @@ def _visit_TypeAlias(self, stmt):
         name = self._visit(stmt.name)
         self._in_lhs_assign = False
         rhs = self._treat_type_annotation(stmt.value, self._visit(stmt.value))
-        type_annotation = UnionTypeAnnotation(VariableTypeAnnotation(TypeAlias(), is_const = True))
+        type_annotation = UnionTypeAnnotation(VariableTypeAnnotation(FinalType.get_new(TypeAlias())))
```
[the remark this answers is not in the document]
yguclu: According to the latest Python [docs](https://docs.python.org/3/library/typing.html#typing.TypeAlias), `TypeAlias` was introduced in Python 3.10.

Curiously, `TypeAlias` is deprecated since Python 3.12 in favor of the `type` statement 🙃

pyccel/ast/datatypes.py
EmilyBourne: Done [306609b](https://github.com/pyccel/pyccel/pull/2415/commits/306609b415f7ed71b116144f01d7410e588445e4)

pyccel/ast/datatypes.py
yguclu: Could you please explain how you decide if a class in this module should be a `Singleton` or not?

This is not very clear to me. For example, I was expecting `PyccelType` to be a singleton as its `__init__` function definitely hints at that.

pyccel/ast/type_annotations.py
```diff
@@ -43,15 +43,11 @@ class VariableTypeAnnotation(PyccelAstNode):
...
         The requested Python type of the variable.
-
-    is_const : bool, default=False
-        True if the variable cannot be modified, false otherwise.
     """
-    __slots__ = ('_class_type', '_is_const')
+    __slots__ = ('_class_type',)
     _attribute_nodes = ()
-    def __init__(self, class_type : PyccelType, is_const : bool = False):
+    def __init__(self, class_type : PyccelType):
```
yguclu: Could you please add a blank line before the `__init__`?
```suggestion
    __slots__ = ('_class_type',)
    _attribute_nodes = ()

    def __init__(self, class_type : PyccelType):
```

pyccel/parser/semantic.py
```diff
@@ -3015,12 +3005,6 @@ def _visit_FunctionTypeAnnotation(self, expr):
         res_type = self._visit(expr.result)
         return UnionTypeAnnotation(FunctionTypeAnnotation(arg_types, res_type))
 
-    def _visit_TypingFinal(self, expr):
-        annotation = self._visit(expr.arg)
-        for t in annotation:
-            t.is_const = True
-        return annotation
```
yguclu: Could you explain why you could remove this function?

pyccel/ast/datatypes.py
```diff
@@ -893,53 +977,67 @@ def order(self):
...
+                order = None
         else:
-            self._order = None
+            order = None
 
-        super().__init__()
+        def __init__(self):
+            self._element_types = args
+
+            self._datatype = dtype if all(d == dtype for d in possible_types) else self
```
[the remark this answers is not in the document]
EmilyBourne: No I don't think so. There are fundamental differences in how homogeneous and non-homogeneous tuples are translated:

 | Homogeneous | Inhomogeneous
---|---|---
Indexing | Can index with a variable | Compile-time indexing only
Translation | Saved like an array | Multiple variables printed

In particular the second option may be preferred for optimisation reasons so I think it is better if users can request it.

pyccel/ast/datatypes.py
yguclu: Me neither! Apparently any string is accepted...

pyccel/parser/semantic.py
EmilyBourne: This was related to the code above. `TypingFinal` was visited as the base of the indexed element but needed special treatment to set `is_const` on the resulting `VariableTypeAnnotation`. This is no longer the case so it is sufficient to add the `_static_type` property to this class

pyccel/ast/datatypes.py
```diff
@@ -603,7 +652,7 @@ def switch_basic_type(self, new_type):
         """
         assert isinstance(new_type, FixedSizeType)
         cls = type(self)
-        return cls(self.element_type.switch_basic_type(new_type))
+        return cls.get_new(self.element_type.switch_basic_type(new_type))
```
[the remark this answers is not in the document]
yguclu: I see. Given that we are not using abstract classes which inherit from `abc.ABC`, I suggest to add a default `get_new` method to the top-level class. This method should raise a `NotImplementedError`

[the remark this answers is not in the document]
EmilyBourne: Done : fe2935804

pyccel/ast/datatypes.py
```diff
@@ -576,10 +642,10 @@ def element_type(self):
 
         The PyccelType describing an element of the container.
         """
-        return self._element_type # pylint: disable=no-member
+        return self._element_type
```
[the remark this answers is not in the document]
EmilyBourne: Pylint doesn't look at the slots when checking this. E.g. the following code:
```python
class A:
    __slots__ = ('_sub_class_obj',)

    def method(self):
        print(self._sub_class_obj)
```
raises:
```
tmp_lint.py:6:14: E1101: Instance of 'A' has no '_sub_class_obj' member (no-member)
```
This is a common pattern in this file which is why I have disabled the `no-member` error for this file.

I think it should be possible to fix this with a `__init__` definition. From what I remember we didn't do this for `PyccelType` as we wanted to be able to define objects at the class level instead of the instance level. But I don't think there is any advantage to this when all instances are singletons anyway

PR #2419, opened 2025-08-14
Fix missing errors

EmilyBourne:
After #2201 all errors raised by Pyccel are caught. This is good for users as no tracebacks are dumped into their terminals. However it means that our error tests are less reliable. Previously they checked that a `PyccelError` was raised to ensure that there was a clean error but now all errors are `PyccelError`s. As a result some clean errors with explanations have been replaced by INTERNAL ERRORs. To fix this this PR adds a line to the beginning of every error test containing a snippet of the expected error string. Errors that had disappeared have been reinstated.


jalalium (approved):
Well done!

PR #2423, opened 2025-08-18
Fix wrapper bug when returning an instance of a class

### LINKED ISSUE #2248: Pyccel Fails to Compile Interdependent Classes A and B Due to Circular Dependency in getitem
### Describe the bug
Pyccel fails to compile a minimal example of two interdependent classes. In this example, class A’s getitem method returns an instance of class B, while class B’s constructor expects an instance of A. Pyccel does not correctly resolve the circular dependency between A and B during the translation process.

### To Reproduce
```python
class A:
    def __init__(self, value: int):
        self.value = value

    def __getitem__(self, index: int):
        return B(self)

class B:
    def __init__(self, a: A):
        self.a = a
```

### Error details
```bash
pyccel:
 |fatal [semantic]: boo.py [6,15]| Undefined function (B)

```
### Expected behavior
Pyccel should recognize class `B`

### Language
C, Fortran, Python

EmilyBourne:
Fix wrapper bug when returning an instance of a class from the method of a preceding class. Fixes #2248 

The issue arose due to the fact that the type had not yet been declared at the moment it was used.

Additionally the type of an inhomogeneous tuple is fixed. The use of the wrong type caused a warning but did not cause bugs as the pointer pointed to the correct type.


jalalium (approved):
Well done!

PR #2427, opened 2025-08-19
Fix bug when specifying ND array shape with a tuple variable.

### LINKED ISSUE #2426: [Bug] Wrong rank of array with tuple shape

### Describe the bug

When an array is initialised and a variable containing a tuple is used for the shape Pyccel does not use the tuple to deduce the rank. Instead it assumes a rank 1 object.

**Reproducer**
```python
import numpy as np

def f():
    shape = (1, 2)
    a = np.zeros(shape) # Pyccel says rank 1
    b = np.zeros((1, 2)) # Pyccel says rank 2

    c = a[0,0]

    print(a.shape)
    print(b.shape)

if __name__ == '__main__':
    f()
```

Translating this leads to:
```
  File "/home/emily/Code/pyccel/pyccel/ast/variable.py", line 668, in __init__
    assert len(indices) <= rank
           ^^^^^^^^^^^^^^^^^^^^
AssertionError
```
due to the line `c = a[0,0]`

### Expected behavior

A tuple variable should be valid if the shape is known.

### Add Screenshots

_No response_

### Record

EmilyBourne:
Fix bug when specifying ND array shape with a tuple variable. Fixes #2426


jalalium (changes requested):
The following still does not work:
```
import numpy as np

def g():
    return (1,2)

def f():
    shape = (1, 2)
    a = np.zeros(g())
    b = np.zeros((1, 2))
    c = a[0,0]
    print(a.shape)
    print(b.shape)
if __name__ == '__main__':
    f() 
```
Do you think it is within the scope of this PR?

EmilyBourne:
> The following still does not work:
> 
> ```
> import numpy as np
> 
> def g():
>     return (1,2)
> 
> def f():
>     shape = (1, 2)
>     a = np.zeros(g())
>     b = np.zeros((1, 2))
>     c = a[0,0]
>     print(a.shape)
>     print(b.shape)
> if __name__ == '__main__':
>     f() 
> ```
> 
> Do you think it is within the scope of this PR?

I am not sure it is entirely within scope as the cause is completely different. However it is a 2 line fix so I will add it :+1:

EmilyBourne:
> The following still does not work:

Fixed

jalalium (approved):
looks good to me.

PR #2430, opened 2025-08-20
Fix interface parsing from imported function

### LINKED ISSUE #2429: [Bug] Interfaces not correctly parsed from .pyi file

### Describe the bug

Interfaces defined with the `@overload` decorator are not correctly parsed. The call to `insert_function` leads to the implementations being overwritten leaving only the final implementation.

### Expected behavior

This should work

### Add Screenshots

_No response_

### Record

- [ ] I have starred the repository

EmilyBourne:
Fix interface parsing from imported function. Fixes #2429 

This issue was fixed by #2413 however it is still important to add the fix to the CHANGELOG and add a dedicated test.


tests/pyccel/project_rel_imports/project/folder1/mod1.py
yguclu, line 1: Issue #2430 talks explicitly about "Interfaces defined with the `@overload` decorator", but here we use `TypeVar` instead. Could you please explain how this unit test is relevant?

tests/pyccel/project_rel_imports/runtest.py
```diff
@@ -1,5 +1,7 @@
 # pylint: disable=missing-function-docstring, missing-module-docstring
 from project.folder2.mod3 import one_hundred_plus_sum_to_n_squared
+from project.folder1.mod1 import add2
 
 if __name__ == '__main__':
     print(one_hundred_plus_sum_to_n_squared(4))
+    print(add2(1.0, 4.0))
```
yguclu, line 7: Issue #2430 says that "The call to `insert_function` leads to the implementations being overwritten leaving only the final implementation." In fact here we are using the second and last implementation of `add2`, which works on float numbers. So I expect this unit test to work even before this PR...

yguclu (changes requested):

tests/pyccel/project_rel_imports/runtest.py
EmilyBourne: The first implementation is used in `project/folder1/mod2.py` I added this to make sure that both implementations were being tested rather than to trigger the test

tests/pyccel/project_rel_imports/project/folder1/mod1.py
EmilyBourne: This is a multi-file test. The files are translated one by one. First this file is translated which generates a `mod1.pyi` file which will be parsed in place of this file by all the subsequent files (`mod2.py`, `mod3.py` and `runtest.py`). The `@overload` decorator appears in the `mod1.pyi` file of any file containing Interfaces. `TypeVar`s are a standard way to define an interface.

yguclu: Thanks! I had not realized that this was a multi-file test.

PR #2432, opened 2025-08-20
Fix missing type checks

### LINKED ISSUE #2431: [Bug] Missing type checks

### Describe the bug

The type checks are missing for functions which are themselves called from Interfaces

### Expected behavior

The type checks should be generated

### Add Screenshots

_No response_

### Record

EmilyBourne:
The type check was missing for functions which are themselves called from Interfaces as the check to see if a function is in an Interface was travelling up through the `FunctionCall`. Excluding the `FunctionCall` from the search tree fixes this problem. Fixes #2431


pyccel/codegen/wrapper/c_to_python_wrapper.py
```diff
@@ -1481,7 +1481,7 @@ def _wrap_FunctionDef(self, expr):
             a_var = a.var
             func_scope.insert_symbol(getattr(a_var, 'original_var', a_var).name)
 
-        in_interface = len(expr.get_user_nodes(Interface)) > 0
+        in_interface = len(expr.get_user_nodes(Interface, (FunctionCall,))) > 0
```
yguclu, line 1484: Why not writing the keyword parameter `excluded_nodes` explicitly as you have done below?

tests/epyccel/test_epyccel_functions.py
```diff
@@ -448,6 +448,19 @@ def f(a : T, b : T):
     with pytest.raises(TypeError):
         epyc_f(3.5, 4)
 
+@pytest.mark.parametrize( 'language', (
+        pytest.param("fortran", marks = pytest.mark.fortran),
+        pytest.param("c", marks = pytest.mark.c),
+    )
+)
```
yguclu, line 455: In Python we do not have type checks because these are generated by the Python wrapper?

pyccel/codegen/wrapper/c_to_python_wrapper.py
EmilyBourne: Fixed : 6acee633b

PR #2434, opened 2025-08-20
Fix shape calculation for resized class member variables

### LINKED ISSUE #2433: [Bug] Wrong size calculation

### Describe the bug

If the shape of an array is known in one context but not in another, it is assumed that the shape is the same in both cases, but this is not necessarily true.

### Reproducer

```python
import numpy as np

class A:
    def __init__(self, b : bool, a : 'int[:]'):
        if b:
            self.c = np.array([1,2,3])
        else:
            self.c = np.array(a)

        print(len(self.c))


if __name__ == '__main__':
    arr = np.array([1,2,3,4,5,6])
    a1 = A(True, arr)
    a2 = A(False, arr)
```

Python output:
```
3
6
```

Pyccel output (F/C)
```
3
3
```

### Add Screenshots

_No response_

### Record

EmilyBourne:
Fix shape calculation for resized class member variables. Fixes #2433

### DIFF
--- a/tests/epyccel/test_epyccel_classes.py
+++ b/tests/epyccel/test_epyccel_classes.py
@@ -501,3 +501,25 @@ def test_class_globals_visitation_order(language):
     a_l = modnew.A()
 
     assert a_py.x == a_l.x
+
+def test_class_variables(language):
+    import classes.class_variables as mod
+    modnew = epyccel(mod, language = language)
+
+    a_py = mod.A(True)
+    a_l = modnew.A(True)
+
+    assert a_py.x == a_l.x
+    assert a_py.z == a_l.z
+    assert a_py.get_4() == a_l.get_4()
+    assert np.array_equal(a_py.y, a_l.y)
+    assert a_py.get_y_len() == a_l.get_y_len()
+
+    a_py = mod.A(False)
+    a_l = modnew.A(False)
+
+    assert a_py.x == a_l.x
+    assert a_py.z == a_l.z
+    assert a_py.get_4() == a_l.get_4()
+    assert np.array_equal(a_py.y, a_l.y)
+    assert a_py.get_y_len() == a_l.get_y_len()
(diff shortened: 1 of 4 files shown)


tests/epyccel/test_epyccel_classes.py
```diff
@@ -501,3 +501,25 @@ def test_class_globals_visitation_order(language):
...
+    assert a_py.get_y_len() == a_l.get_y_len()
+
+    a_py = mod.A(False)
+    a_l = modnew.A(False)
+
+    assert a_py.x == a_l.x
+    assert a_py.z == a_l.z
+    assert a_py.get_4() == a_l.get_4()
+    assert np.array_equal(a_py.y, a_l.y)
+    assert a_py.get_y_len() == a_l.get_y_len()
```
yguclu, line 525: How about a parametric unit test?

```suggestion
@pytest.mark.parametrize('bool_param', [True, False])
def test_class_variables(language, bool_param):
    import classes.class_variables as mod
    modnew = epyccel(mod, language = language)

    a_py = mod.A(bool_param)
    a_l = modnew.A(bool_param)

    assert a_py.x == a_l.x
    assert a_py.z == a_l.z
    assert a_py.get_4() == a_l.get_4()
    assert np.array_equal(a_py.y, a_l.y)
    assert a_py.get_y_len() == a_l.get_y_len()
```

EmilyBourne: Code duplication reduced: [175e225](https://github.com/pyccel/pyccel/pull/2434/commits/175e225d461c6ed3bd5b81d642e7abb7ce4fce58)

yguclu (changes requested):

tests/epyccel/test_epyccel_classes.py
```diff
@@ -510,3 +510,17 @@ def test_class_globals_visitation_order(language):
     a_l = modnew.A()
 
     assert a_py.x == a_l.x
+
+def test_class_variables(language):
+    import classes.class_variables as mod
+    modnew = epyccel(mod, language = language)
+
+    for bool_val in (True, False)
```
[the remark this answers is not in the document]
EmilyBourne: :facepalm: yes

PR #2461, opened 2025-10-09
Fix memory leak when memory is allocated in a function

### LINKED ISSUE #2292: [Bug] Memory leak when allocating memory via a function

### Describe the bug

When memory is allocated in a function Pyccel correctly recognises that an `Allocate` node is unneeded, however this means that if the same variable is used to store a new array later it is not recognised as allocated. This leads to a memory leak in the C code

### Expected behavior

A deallocation should be added before the second function call

### Add Screenshots

```python
import numpy as np

def get_a():
    a = np.array([1,2,3,4])
    return a

def get_b():
    b = np.array([5,6,7,8])
    return b

if __name__ == '__main__':
    x = get_a()
    print(x[0])
    x = get_b()
    print(x[0])
```

Image

### Record

EmilyBourne:
It is difficult to locate function memory allocation a posteriori. However it is easy to identify while the call is being examined. In order to fix #2292 we therefore add a new `alloc_type` to `Allocate`. The `alloc_type` `'function'` indicates that the allocation should not be printed, but any necessary deallocations should be. This `Allocate` node is easy to locate later which allows us to rely on the old logic to add missing deallocations. The argument `heap_mem_in_multirets` is also added to `_ensure_inferred_type_matches_existing` to identify a variable requiring heap allocation, which is collected as one of the returns in a function returning multiple objects. This allows the functions to be recognised without heap variables being falsely identified as pointers (as previously they were seen to point to the function result element). Valgrind tests are added to ensure proper deallocation.

### DIFF
--- a/pyccel/parser/semantic.py
+++ b/pyccel/parser/semantic.py
@@ -1574,7 +1574,7 @@ class SemanticParser(BasicParser):
 
         return func
 
-    def is_memory_allocated_in_function(self, lhs, rhs, mem_in_multirets):
+    def is_memory_allocated_in_function(self, lhs, rhs, heap_mem_in_multirets):
         """
         Determine whether a memory chunk was allocated in a function.
 
@@ -1588,8 +1588,8 @@ class SemanticParser(BasicParser):
             The array variable.
         rhs : TypedAstNode
             The expression used to define the array variable.
-        mem_in_multirets : bool
-            If True, the variable that will be created is found
+        heap_mem_in_multirets : bool
+            If True, the variable that will be created in heap memory, is found
             in a multi-values return, False otherwise.
 
         Returns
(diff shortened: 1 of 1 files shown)


pyccel/codegen/printing/fcode.py
```diff
@@ -2177,6 +2177,22 @@ def _print_Assign(self, expr):
...
+                    return ''
+                elif expr.status == 'unknown':
+                    var_code = self._print(expr.variable)
+                    return (f'if (allocated({var_code})) then\n'
+                            f'  deallocate({var_code})\n'
+                             'end if\n')
+
+                elif expr.status == 'allocated':
+                    var_code = self._print(expr.variable)
+                    return f'deallocate({var_code})\n'
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
Tested in memory leak testing

pyccel/parser/semantic.py
```diff
@@ -1986,6 +2018,9 @@ def _ensure_inferred_type_matches_existing(self, class_type, d_var, var, is_auga
         rhs : TypedAstNode
             The right hand side of the expression : lhs=rhs.
             If is_augassign is False, this value is not used.
+        arr_in_multirets : bool, default: False
+            If True, the variable that will be created is an array
+            in multi-values return, false otherwise.
```
yguclu:
```suggestion
            in multi-values return, False otherwise.
```

pyccel/parser/semantic.py
```diff
@@ -1986,6 +2018,9 @@ def _ensure_inferred_type_matches_existing(self, class_type, d_var, var, is_auga
         rhs : TypedAstNode
             The right hand side of the expression : lhs=rhs.
             If is_augassign is False, this value is not used.
+        arr_in_multirets : bool, default: False
```
yguclu: Should we call this variable `mem_in_multirets` instead? That would be consistent with the method `is_memory_allocated_in_function`

yguclu (changes requested):
Good job fixing this bug! I have some minor comments

PR #2467, opened 2025-10-15
Make class `__init__`/`__del__` methods optional when wrapping

### LINKED ISSUE #2464: [Feature Request] Remove the need for __init__ and __del__ in class wrapper

### Feature Description

Following #2422 it is possible to bind low-level functions to a Python class description. However the use of this is limited by how closely the implementation follows Pyccel's translation. In particular a version of `__init__` and `__del__` are currently required but this should be unnecessary.

See also [this conversation](https://github.com/pyccel/pyccel/pull/2422#discussion_r2420805139) about `final` subroutines in Fortran.

### Use Case

Any class which doesn't contain allocatables probably won't have a `__del__` equivalent

### Benefits

_No response_

### Add Screenshots

_No response_

### Priority

High

### Record

EmilyBourne:
Stop requiring .pyi files to describe `__init__` and `__del__` methods for classes (for wrapping). Fixes #2464

**Commit Summary**
- Allow `_print_Deallocate` to print nothing if the class doesn't have a `__del__` function
- Ensure that a `__del__` function is wrapped in the Fortran to C wrapper so `deallocate` is called
- Only create `__init__`/`__del__` methods for classes if we are not checking a stub file
- Replace `ConstructorCall` with `EmptyNode` for classes which have no `__init__` method
- Add test with no `__init__`/`__del__` methods
- Add test using Fortran `final` keyword

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -31,7 +31,7 @@ def stc_language(request):
     return request.param
 
 @pytest.fixture(autouse=True)
-def skipif_by_language(request, language):
+def skipif_by_language(request):
     """
     Looks for the decorator `skipif_by_language(condition, *, language, reason)`
     and skips the test when the condition is met for a given language.
@@ -46,14 +46,16 @@ def skipif_by_language(request, language):
         assert 'reason'   in mark.kwargs.keys(), preamble + "you need to specify reason=STRING"
 
         condition = mark.args[0]
-        lang      = mark.kwargs['language']
+        language  = mark.kwargs['language']
         reason    = mark.kwargs['reason']
 
         assert isinstance(condition, bool), preamble + "condition must be bool"
-        assert isinstance(  lang, str), preamble + "language must be str"
+        assert isinstance(language, str), preamble + "language must be str"
         assert isinstance(reason, str), preamble + "reason must be str"
 
-        if condition and lang == language:
+        assert 'language' in request.fixturenames, preamble + "test must depend on the language fixture"
+
+        if condition and request.getfixturevalue('language') == language:
             pytest.skip(reason)
 
 def move_coverage(path_dir):


pyccel/parser/semantic.py
```diff
@@ -3658,19 +3659,6 @@ def _visit_FunctionCall(self, expr):
             d_methods = cls.methods_as_dict
             method = d_methods.pop('__init__', None)
```
yguclu: Could we call this variable `init_method` or just `init`? The name `method` is a bit confusing 😕

EmilyBourne: Done [615c52d](https://github.com/pyccel/pyccel/pull/2467/commits/615c52d27eefeeec4121b13176dbb19baa2070f6)

tests/pyccel/test_pyccel_wrap.py
```diff
@@ -117,10 +118,28 @@ def test_array_methods():
...
+def test_class_final_fortran_keyword():
+    check_pyccel_wrap_and_call_translation('final_destroy', 'runtest_final_destroy', 'fortran')
+    valgrind = shutil.which('valgrind')
+    if valgrind:
+        cwd = Path(__file__).parent / 'wrap_scripts' / 'fortran_tests'
+        subprocess.run([valgrind, sys.executable, cwd / 'runtest_final_destroy.py'], cwd = cwd, check = True)
+        exe_file = cwd / 'runtest_final_destroy'
+        if sys.platform == "win32":
+            exe_file = exe_file.with_suffix('.exe')
+        subprocess.run([valgrind, exe_file], cwd = cwd, check = True)
```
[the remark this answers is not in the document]
EmilyBourne: I am not sure. Potentially we don't need to care given that the memory should be released when the program ends anyway. But it does seem like it would be cleaner. In the discussions that I found that talked about this issue no-one seems to be doing this usually.
E.g.
https://community.intel.com/t5/Intel-Fortran-Compiler/Fortran-FINAL-routine-called-multiple-times/m-p/1147089

yguclu (approved):
Looks good to me!

PR #2472, opened 2025-10-23
Add C support for nested functions

### LINKED ISSUE #601: Nested Functions not working for C language
**Describe the bug**
Pyccel cant generate a working code in the case of nested functions.
It fails in the compilation stage because the inner function is not generated at all.

**To Reproduce**
```python
def func():
  def printer():
    print("hello world!")
  printer()
```

the generated code is the following:
```bash/fortran/c
#include <test.h>
#include <stdlib.h>

void func(void)
{

    printer();
}
```
The prototype of the inner function is available in the header file but there is no generated function called `printer`

**Language**
C

EmilyBourne:
Add C support for nested functions. Fixes #601


yguclu (changes requested):
Good job!

A comment in method [`get_variables`](https://github.com/pyccel/pyccel/blob/43089bbda787a790b9db2796c7658a77a4acbc1a/pyccel/parser/semantic.py#L502-L524) of our semantic parser reads that the method "needs more tests when we have nested functions". Do you think this is the case?

PR #2477, opened 2025-11-02
Fix passing non-contiguous slice

### LINKED ISSUE #2443: [Bug] Can't pass non-contiguous array

### Describe the bug

It is not possible to pass a non-contiguous array to a Pyccel function because the wrapper checks the flags `NPY_ARRAY_C_CONTIGUOUS` and `NPY_ARRAY_F_CONTIGUOUS` to check if the array is valid. This is more restrictive than it should be

### Expected behavior

The wrapper should only check if the strides are strictly increasing  or decreasing

### Add Screenshots

_No response_

### Record

EmilyBourne:
Fix passing non-contiguous slice by improving the methods `_check_pyarray_order` and `get_strides_and_shape_from_numpy_array`. It is sufficient to check that the strides are correctly ordered to determine the ordering. The shapes and strides then require 1 additional piece of information in order to correctly pass the input to low-level code. Fixes #2443

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,8 @@ All notable changes to this project will be documented in this file.
 -   #2458 : Fix incoherent import printing in Python and `.pyi` files.
 -   #2460 : Fix missing error when overwriting a constant (e.g. `np.pi`).
 -   #2462 : Fix erroneous additional import in stub file.
+-   #2479 : Fix missing `@property` decorator in stub files.
+-   #2479 : Fix Fortran wrapping of methods with `@property` decorator returning inhomogeneous tuples.
 -   #2443 : Fix passing non-contiguous slices of arrays.
 -   Rename `main` function when translating to C.
 
--- a/pyccel/codegen/printing/pycode.py
+++ b/pyccel/codegen/printing/pycode.py
@@ -283,8 +283,9 @@ class PythonCodePrinter(CodePrinter):
             res = f' -> {self._get_type_annotation(result.var)}'
         else:
             res = ' -> None'
+        dec = self._handle_decorators(func.decorators)
         self.exit_scope()
-        return ''.join((wrapping, overload, f"def {name}({args}){res}:\n",
+        return ''.join((dec, wrapping, overload, f"def {name}({args}){res}:\n",
                         self._indent_codestring(body)))
 
     def _handle_decorators(self, decorators):
--- a/pyccel/codegen/wrapper/c_to_python_wrapper.py
+++ b/pyccel/codegen/wrapper/c_to_python_wrapper.py
@@ -1956,7 +1956,7 @@ class CToPythonWrapper(Wrapper):
 
         # Cast the C variable into a Python variable
         get_val_result_var = getattr(get_val_result, 'original_function_result_variable', get_val_result.var)
-        result_wrapping = self._extract_FunctionDefResult(get_val_result_var, True)
+        result_wrapping = self._extract_FunctionDefResult(get_val_result_var, True, expr.getter)
         res_wrapper = result_wrapping['body']
         c_results = result_wrapping['c_results']
         getter_result = result_wrapping['py_result']
@@ -1969,9 +1969,10 @@ class CToPythonWrapper(Wrapper):
 
         if isinstance(expr.getter.original_function, DottedVariable):
             wrapped_var = expr.getter.original_function
+
+            res_wrapper.extend(self._incref_return_pointer(getter_args[0], getter_result, wrapped_var))
         else:
             wrapped_var = expr.getter.original_function.results.var
-        res_wrapper.extend(self._incref_return_pointer(getter_args[0], getter_result, wrapped_var))
 
         getter_body = [*setup,
                        *arg_code,
--- a/tests/pyccel/project_class_imports/project/shapes/Square_mod.py
+++ b/tests/pyccel/project_class_imports/project/shapes/Square_mod.py
@@ -8,18 +8,22 @@ class Square:
         self.c = c
         self.d = d
 
-    def get_corner_1(self):
+    @property
+    def corner_1(self):
         x,y = self.a.get_val()
         return x,y
 
-    def get_corner_2(self):
+    @property
+    def corner_2(self):
         x,y = self.b.get_val()
         return x,y
 
-    def get_corner_3(self):
+    @property
+    def corner_3(self):
         x,y = self.c.get_val()
         return x,y
 
-    def get_corner_4(self):
+    @property
+    def corner_4(self):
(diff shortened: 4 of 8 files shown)


EmilyBourne:
- [ ] Update Changelog

EmilyBourne:
> There appears to be just one failure: `test_array_int32_2d_C_scalar_idiv_stride[c]`. See [this job](https://github.com/pyccel/pyccel/actions/runs/19134802286/job/54684090776).

Yes, this is why this PR is still in draft. The technique that we are using to slice the array would work if the strides were related to the underlying 1D data, but this is not the case for our slicing. Now that I have worked out the problem I do know how to fix it at least

pyccel/ast/operators.py
```diff
@@ -784,6 +784,8 @@ def __new__(cls, arg1 = None, arg2 = None, simplify = False):
                 dtype = cls._calculate_type(arg1, arg2)
                 return convert_to_literal(arg1.python_value - arg2.python_value,
                                           dtype)
+            if isinstance(arg1, PyccelAdd) and arg1.args[1] == arg2:
+                return arg1.args[0]
```
yguclu: How about the similar case where the first argument of `PyccelAdd` is equal to `arg2`?

```suggestion
            if isinstance(arg1, PyccelAdd):
                if arg1.args[1] == arg2:
                    return arg1.args[0]
                elif arg1.args[0] == arg2:
                    return arg1.args[1]
```

pyccel/stdlib/cwrapper/cwrapper.c
```diff
@@ -309,18 +318,37 @@ static char* _check_pyarray_rank(PyArrayObject *a, int rank, bool allow_empty)
...
+    else if (flag == NPY_ARRAY_F_CONTIGUOUS) {
+        int nd = PyArray_NDIM(a);
+        npy_intp* np_strides = PyArray_STRIDES(a);
+        for (int i = 1; i<nd; ++i) {
+            valid = valid & (np_strides[i-1] <= np_strides[i]);
+        }
+    }
+    else {
+        valid = PyArray_CHKFLAGS(a, flag);
+    }
```
yguclu: What flags can this function process? All those handled by [`PyArray_CHKFLAGS`](https://numpy.org/devdocs/reference/c-api/array.html#c.PyArray_CHKFLAGS) or just some of them?

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -6238,9 +6238,9 @@ def copy_array_to_C(a : Y):
         return b
 
     arr_1d = randint(min_int, max_int, size=5)
-    arr_2d = uniform(min_float64 / 2, max_float64 / 2, size=(3,4))
-    arr_3d = (uniform(min_float64 / 2, max_float64 / 2, size=(3,4,5)) \
-            + uniform(min_float64 / 2, max_float64 / 2, size=(3,4,5))*1j).T
+    arr_2d = uniform(-100,  100, size=(3,4))
+    arr_3d = (uniform(-100, 100, size=(3,4,5)) \
+            + uniform(-100, 100, size=(3,4,5))*1j).T
```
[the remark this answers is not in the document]
EmilyBourne: This change was made while debugging. I forgot to revert it

pyccel/stdlib/cwrapper/cwrapper.c
```diff
@@ -149,27 +149,36 @@ void get_strides_and_shape_from_numpy_array(PyObject* arr, int64_t shape[], int6
...
+            for (int i = 0; i < nd-1; ++i) {
+                base_shape[i] = np_strides[i+1] / current_shape;
+                current_shape *= base_shape[i];
+                strides[i+1] = np_strides[i+1] / current_shape;
             }
+            base_shape[nd-1] = np_shape[nd-1] * strides[nd-1];
+        }
+        for (int i = 0; i < nd; ++i) {
+            ubounds[i] = np_shape[i] * strides[i];
         }
```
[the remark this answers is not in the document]
EmilyBourne: > By the way, could we not use PyArray_BASE and PyArray_SHAPE to get base_shape directly?

This would cause problems if an array has been sliced such that the base array has a different number of dimensions compared to the view.
E.g
```python
a = np.ones((3,4,5))
b = a[:,2,:]
# PyArray_SHAPE(PyArray_BASE(b)) = (3,4,5)
# base_shape = (12,5)
```

pyccel/stdlib/cwrapper/cwrapper.c
EmilyBourne: All of them, but I don't know what is available and I don't think we currently use this after this PR

tests/epyccel/recognised_functions/test_numpy_funcs.py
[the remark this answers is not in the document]
EmilyBourne: Reverted: [bc0951d](https://github.com/pyccel/pyccel/pull/2477/commits/bc0951dfd9d157c7503cad56a7ce5f526e2096b6)

pyccel/stdlib/cwrapper/cwrapper.c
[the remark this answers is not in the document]
EmilyBourne: Done : [4f14854](https://github.com/pyccel/pyccel/pull/2477/commits/4f14854103f2c19950d916b275906e2e75c83b7c)

pyccel/stdlib/cwrapper/cwrapper.c
EmilyBourne: Comment modified : [4f14854](https://github.com/pyccel/pyccel/pull/2477/commits/4f14854103f2c19950d916b275906e2e75c83b7c)

pyccel/stdlib/cwrapper/cwrapper.c
```diff
@@ -139,37 +139,80 @@ PyObject	*Float_to_NumpyDouble(float *d)
...
+ * E.g.
+ * ```python
+ * a = np.ones((2,3,4,5))
+ * b = a[:,2,::2,:]
+ * ```
+ * Here b has dimensionality 3 so the base_shape, ubounds and strides each contain
+ * 3 elements:
+ *   - base_shape : [4,4,5]
+ *   - ubound : [4,4,5]
+ *   - strides : [3,2,1]
```
yguclu: As discussed in a separate chat, it appears that this example cannot be reproduced by this function in its current form. The result that I obtain is

- `base_shape` : `(2, 6, 10)`
- `ubound` : `(2, 2, 5)`
- `strides` : `(1, 1, 1)`

This should be double checked. If correct, we should either update the function `get_strides_and_shape_from_numpy_array` or its docstring.

PR #2487, opened 2025-11-10
Fix linux CI Python input

EmilyBourne:
All the workflow dispatch CIs handle the Python version as a string allowing `3.10` to be passed. For the linux CI there is an extra layer of obfuscation (due to the use of matrix). During this obfuscation the version is handled without specifying the type. This means that the string type is lost. As a result this workflow (and only this workflow) currently needs to be triggered with `'3.10'`. This also means that the linux tests run by the bot are also currently not working. This PR fixes this by quoting the received string to preserve the type.


yguclu (approved):
Looks good to me

PR #2490, opened 2025-11-11
Stop `FinalType` being passed in assignment

### LINKED ISSUE #2489: [Bug] Semantic error when passing `Final` argument to `@inline` function
### Bug description

A strange bug appears when passing a `Final` argument to an `@inline` function. Apparently, a `Final` temporary variable is created and later assigned to. This causes an error in the `_visit_Assign` method of our semantic parser.

### Expected behavior

Since the inline function does not attempt to modify the input argument, Pyccel should translate the code with no errors. In particular, no `Final` temporary variable should be created.

### Minimal reproducible example

Let the contents of file `final_inline_bug.py` be:
```python
from typing import Final
from pyccel.decorators import inline

def finite_geometric_series(r : Final[float], n : Final[int]) -> float:
    a : float = 1.0
    for k in range(1, n + 1):
        a += power(r, k)
    return a

@inline
def power(x : Final[float], n : Final[int]) -> float:
    return x ** n
```
This error is raised:
```
$ pyccel final_inline_bug.py

pyccel:
 |error [semantic]: final_inline_bug.py [12,4]| Cannot modify variable marked as Final (Dummy_0000)
```
The code is translated correctly without the `@inline` decorator.

### Record

- [x] I have searched the existing issues and could not find a similar one

EmilyBourne:
Stop `Final` attribute being used for the new type of an object if the final object is copy-assigned. Fixes #2489 

No changes are made to the CHANGELOG as this bug was introduced in this version.

### DIFF
--- a/tests/epyccel/test_epyccel_variable_annotations.py
+++ b/tests/epyccel/test_epyccel_variable_annotations.py
@@ -122,7 +122,6 @@ def test_final_annotation(language):
 
 def test_final_annotation_transmission(language):
     def final_annotation():
-        from typing import Final
         a : Final[int] = 3
         b = a
         return b


yguclu (approved):
Looks good to me!

PR #2492, opened 2025-11-12
Fix class definition retrieval for `Final[MyClass]`

### LINKED ISSUE #2491: [Bug] Missing class definition when using `Final[MyClass]`

### Describe the bug

The `class_base` is retrieved with `self.scope.find(str(class_type), 'classes') or get_cls_base(class_type)` however if `class_type` is qualified (e.g. Final[] or Annotated[]) then the string does not match that expected by the scope. As a result the `class_base` is not found.

### Reproducer

```python
class A:
    def __init__(self, n : int):
        self._my_vals = np.ones(n)

    def print_vals(self : Final[A]):
        print(self._my_vals)
```

### Add Screenshots

_No response_

### Record

EmilyBourne:
Fix the class definition retrieval for qualified custom class instances. This includes objects such as `Final[MyClass]`. A new class method `get_cls_base` is added to `SemanticParser` to reduce code duplication. The function `get_cls_base` in `pyccel.ast.class_defs` is renamed as `get_builtin_cls_base` and only used for built-in datatypes. Fixes #2491.

Fix class name for `NumpyNDArrayType` subclasses.

Correct import of renamed classes in stub files.

The CHANGELOG is not updated as this bug was introduced via #2415

### DIFF
--- a/pyccel/codegen/printing/pycode.py
+++ b/pyccel/codegen/printing/pycode.py
@@ -12,7 +12,7 @@ from pyccel.ast.builtins   import PythonMin, PythonMax, PythonType, PythonBool, 
 from pyccel.ast.builtins   import PythonComplex, DtypePrecisionToCastFunction, PythonTuple, PythonDict
 from pyccel.ast.builtin_methods.list_methods import ListAppend
 from pyccel.ast.core       import CodeBlock, Import, Assign, FunctionCall, For, AsName, FunctionAddress, If
-from pyccel.ast.core       import IfSection, FunctionDef, Module, PyccelFunctionDef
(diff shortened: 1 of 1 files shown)


pyccel/codegen/printing/pycode.py
yguclu: Could you please explain the changes in this file?

EmilyBourne: Yes exactly. In `get_builtin_cls_base` I removed `if isinstance(class_type, CustomDataType): return None`. The class base should not be `None` and if the object is used having `None` as a class base will cause problems. You can get away with it in a stub file as the objects are not used.

EmilyBourne: I have added the sentence
> Correct import of renamed classes in stub files.

PR #2495, opened 2025-11-12
Fix unnecessary pointer creation for inline return

### LINKED ISSUE #2494: [Bug] Pointer assignment is forbidden in pure functions

### Describe the bug

Assigning pointers in pure functions leads to:
```
Bad target in pointer assignment in PURE procedure at (1)
```
Pure functions are important for working parallelism. In most cases the user should be able to avoid the use of pointers, however one tricky use case is for class properties. If a property provides a getter for an array, the method will return a pointer. This can be worked around using `@inline`. However currently Pyccel generates an Assign statement to handle the call to the inlined function. This introduces the pointer assignment.

### Expected behavior

The pointer assignment should be avoided for the result of an inlined function

### Add Screenshots

_No response_

### Record

EmilyBourne:
Remove unnecessary pointer creation for inline return. Fixes #2494 

**Commit Summary**
- Fix incorrect name printing in Python (in case of class method whose name collides with a global function)
- When an inline function `foo()` contains one return of the form `return X` translate `a = foo()` to `a = X` instead of `Dummy = X; a = Dummy`
- Add a test which fails on `devel` due to unnecessary dummy pointers being created in a `@pure` procedure.

### DIFF
--- a/.dict_custom.txt
+++ b/.dict_custom.txt
@@ -9,6 +9,7 @@ SymPy
 SymPy's
 BLAS
 LAPACK
+LAPACK's
 MPI
 OpenMP
 Fortran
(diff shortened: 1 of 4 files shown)


yguclu (commented):
Although this PR is not ready for review yet, it passes all the tests so I have reviewed it. It looks good to me! I just have one question

yguclu (approved):

PR #2501, opened 2025-11-14
Fix stdlib chaining

### LINKED ISSUE #2500: [Bug] Bad chaining of stdlib imports
# Is there an existing issue for this?

# Is this issue mentioned in the CHANGELOG?

# Describe the bug

There is a small bug in how stdlib imports are chained in a multi-file project. This eventually leads to a linker error.

# Reproducer

## file1.py

```python
def get_floor(a : int, b : int):
    return a//b
```

## file2.py

```python
from file1 import get_floor

def level_1_floor_div(a : int, b : int):
    return get_floor(a, b) // get_floor(b, a)
```

## file3.py

```python
from file2 import level_1_floor_div

def level_2_floor_div():
    return level_1_floor_div(40, 34)
```

When trying to import this final file we see the following error:
```
----> 1 import file4

ImportError: /home/emily/Code/pyccel/tests/pyccel/project_stdlib_imports/file3.cpython-313-x86_64-linux-gnu.so: undefined symbol: __pyc_math_f90_MOD_pyc_floor_div_i64
```

This is explained by the stub file

## file3.pyi (generated)

```python
#$ header metavar printer_imports="pyc_math_f90pyc_math_f90"
from file3 import level_1_floor_div
from typing import TypeVar
from pyccel.decorators import low_level

@low_level('level_2_floor_div')
def level_2_floor_div() -> 'int':
    ...
```

The printer imports are wrong and so therefore are the linker command

### Record

EmilyBourne:
Fix the chaining of stdlib imports across multiple files. Fixes #2500

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@ All notable changes to this project will be documented in this file.
 -   #2479 : Fix Fortran wrapping of methods with `@property` decorator returning inhomogeneous tuples.
 -   #2443 : Fix passing non-contiguous slices of arrays.
 -   #2436 : Fix list comprehension with variable slice containing negative indices as an iterable.
+-   #2496 : Fix renaming classes when importing.
 -   #2500 : Fix bad chaining of `stdlib` imports.
 -   Rename `main` function when translating to C.
 
--- a/pyccel/codegen/printing/fcode.py
+++ b/pyccel/codegen/printing/fcode.py
@@ -2312,7 +2312,13 @@ class FCodePrinter(CodePrinter):
         return f"{iterable_type}_Iterator"
 
     def _print_CustomDataType(self, expr):
-        return expr.low_level_name
+        while hasattr(expr, 'underlying_type'):
+            expr = expr.underlying_type
+        try:
+            name = self.scope.get_import_alias(expr, 'cls_constructs')
+        except RuntimeError:
+            name = expr.low_level_name
+        return name
 
     def _print_DataType(self, expr):
         return self._print(expr.name)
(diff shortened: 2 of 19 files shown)


EmilyBourne:
- [ ] Update Changelog

EmilyBourne:
It is extremely difficult to create a reproducer for this. At least 3 files are required. Now that I am removing duplicates, at least 2 stdlibs are also required. I am not convinced that it is worth the effort to target such a specific corner case.

yguclu (commented):
Do we have a unit test that would fail on `devel` prior to these changes?

yguclu:
> It is extremely difficult to create a reproducer for this. At least 3 files are required. Now that I am removing duplicates, at least 2 stdlibs are also required. I am not convinced that it is worth the effort to target such a specific corner case.

Sorry, I did not read your comment before. This is a corner case, but you need it fixed for PyGyro right?

EmilyBourne:
> Do we have a unit test that would fail on `devel` prior to these changes?

As per the message above https://github.com/pyccel/pyccel/pull/2501#issuecomment-3534134414

It may be possible to construct one, but it would require at least 3 files. It is also very much a corner case, and I am not convinced that I can construct a unit test which would fail for this but also for a similar error and that would continue to fail in the future. This seems like a lot of effort for a 3 line fix

EmilyBourne:
> > It is extremely difficult to create a reproducer for this. At least 3 files are required. Now that I am removing duplicates, at least 2 stdlibs are also required. I am not convinced that it is worth the effort to target such a specific corner case.
> 
> Sorry, I did not read your comment before. This is a corner case, but you need it fixed for PyGyro right?

Yes, PyGyro caught this issue

PR #2503, opened 2025-11-16
fix pyccel-clean to remove empty directories left behind after a recursive clean

triangle-motelti:
In the `pyccel_clean.py` file, the `pyccel_clean` function has a bug that prevents it from cleaning directories recursively as intended.

When the function encounters a subdirectory, it correctly calls itself to clean that subdirectory. However, after the recursive call returns, it doesn't do anything with the now-empty subdirectory. An empty directory that previously contained Pyccel-generated files (like `__pyccel__`) will be left behind.

The `pyccel_clean` function should check if a subdirectory is empty after a recursive call. if it is, the function should remove it.

This change ensures that after recursively cleaning a directory, the function checks if the directory has become empty and, if so, removes it. This will result in a much cleaner directory tree after running the command.

### DIFF
--- a/.github/actions/check_for_stc/action.yml
+++ b/.github/actions/check_for_stc/action.yml
@@ -15,6 +15,4 @@ runs:
         print(pyccel_loc)
         print("${{ inputs.not_editable }}")
         assert exists(join(pyccel_loc,'extensions/STC/include/stc'))
-        if ${{ inputs.not_editable }}:
-          assert not exists(join(pyccel_loc,'extensions/STC/docs'))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,7 @@ All notable changes to this project will be documented in this file.
 -   #2510 : Fix missing initialisation of container aliases in C.
 -   #2483 : Fix calls to `math.sqrt` using dotted syntax.
 -   #2483 : Fix calls to `cmath.sqrt` using dotted syntax.
+-   #2484 : Fix optional `TypeVar` arguments.
 -   Rename `main` function when translating to C.
 
 ### Changed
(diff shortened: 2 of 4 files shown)


triangle-motelti:
- [ ] Write a clear PR description
- [ ] Add tests to check your code works as expected
- [ ] Update Changelog
- [ ] Update AUTHORS file

pyccel/commands/pyccel_clean.py
```diff
@@ -12,75 +12,86 @@
...
+def remove_pyccel_folders(file_name):
+    """Remove __pyccel__ or __epyccel__ folders"""
+    if os.path.basename(file_name).startswith(("__pyccel__", "__epyccel__")):
+        shutil.rmtree(file_name, ignore_errors=True)
+        return True
+    return False
+
+
+def remove_pyccel_files(file_name, remove_shared_libs=False, remove_programs=False):
+    """Remove files generated by Pyccel based on type and flags"""
```
yguclu: Could you please expand this docstring and add the `Parameters` and `Returns` sections?

pyccel/commands/pyccel_clean.py
```diff
@@ -12,75 +12,86 @@
...
-def pyccel_clean(path_dir = None, recursive = True, remove_shared_libs = False, remove_programs = False):
-    """
-    Remove folders generated by Pyccel.
 
-    Remove `__pyccel__X` and `__epyccel__X` folders as well
-    as any Python shared libraries from the directory path_dir.
-    The folders `__pyccel__X` are called `__pyccel__` unless
-    they were generated using `pytest-xdist`.
+def remove_pyccel_folders(file_name):
+    """Remove __pyccel__ or __epyccel__ folders"""
```
yguclu: Could you please expand this docstring and add the `Parameters` and `Returns` sections?

PR #2514, opened 2025-11-20
Fix optional `TypeVar` arguments

### LINKED ISSUE #2484: [Bug] Optional TypeVar argument causes runtime failure at function call
### Bug description

If an optional function parameter is declared to have `TypeVar` type, a `TypeError` is raised when the function is called without the corresponding argument. This is probably caused by a bug in the Python wrapper.

### Expected behavior

It should be possible to call the function without providing the optional argument, which is the case when using a classical type annotation without `TypeVar`.

### Minimal reproducible example

Let the file `optional_typevar_bug.py` contain
```python
from typing import TypeVar

T = TypeVar('T', float, complex)

def max_abs(a : T, b : T = None) -> float:
    if b is None:
        return abs(a)
    else:
        return max(abs(a), abs(b))
```
Then
```python
$ pyccel optional_typevar_bug.py
$ ipython
In [1]: from optional_typevar_bug import max_abs

In [2]: max_abs(3., -4.)
Out[2]: 4.0

In [3]: max_abs(1j, 1+1j)
Out[3]: 1.4142135623730951

In [4]: max_abs(3.5)
Cell In[4], line 1
----> 1 max_abs(3.5)

TypeError: Unexpected type for argument b. Received <class 'NoneType'>
```
The function `max_abs` does not fail if `TypeVar` is not used.

### Record

- [x] I have searched the existing issues and could not find one similar to this

EmilyBourne:
Fix optional `TypeVar` arguments by leaving the type check for the wrapper of the implementation of the function. Previously the wrapper of the interface checked the type of all `TypeVar` arguments to determine which implementation to call. Now it checks the type of all non-optional `TypeVar` arguments. When the implementation has been chosen, all arguments whose type has not yet been checked are checked for correctness. Fixes #2484

### DIFF
--- a/.github/actions/check_for_stc/action.yml
+++ b/.github/actions/check_for_stc/action.yml
@@ -15,6 +15,4 @@ runs:
         print(pyccel_loc)
         print("${{ inputs.not_editable }}")
         assert exists(join(pyccel_loc,'extensions/STC/include/stc'))
-        if ${{ inputs.not_editable }}:
-          assert not exists(join(pyccel_loc,'extensions/STC/docs'))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,7 @@ All notable changes to this project will be documented in this file.
 -   #2488 : Use MPI wrapper executables to compile MPI code on Windows.
 -   #2494 : Avoid creating unnecessary pointers when inlining functions.
 -   #2504 : Expose annotated `@inline` functions in the generated Python shared library.
+-   #2508 : Place if ternary expressions in parentheses.
 -   \[INTERNALS\] Ensure low-level names are chosen in the parsing so objects are never renamed in the code generation stage.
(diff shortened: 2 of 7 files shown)


tests/epyccel/test_epyccel_default_args.py
```diff
@@ -96,3 +97,23 @@ def test_changed_precision_arguments(language):
...
+            return max(abs(a), abs(b))
+
+    f = epyccel(max_abs, language = language)
+
+    # ...
+    assert f(2.9) == max_abs(2.9)
+    assert f(2.9, 2.3) == max_abs(2.9, 2.3)
+    # ...
+    assert f(2.9+3j) == max_abs(2.9+3j)
+    assert f(2.9+3j, 2.9+3j) == max_abs(2.9+3j, 2.9+3j)
```
yguclu: For the sake of completeness, I would not pass the same argument twice. E.g.:

```suggestion
    assert f(2.9+3j, 2.3+4j) == max_abs(2.9+3j, 2.3+4j)
```

PR #2519, opened 2025-11-26
Use scikit-build-core to pre-compile STC during installation

### LINKED ISSUE #2516: [Bug] Slow translation times

### Describe the bug

A recent PR has significantly slowed down Pyccel. This can be seen in the pyccel-benchmarks
## Old

old benchmarks

## Current

new benchmarks

### Expected behavior

Nothing particularly problematic has been added so this should be investigated

### Add Screenshots

_No response_

### Record

EmilyBourne:
Use `scikit-build-core` as the backend for the Pyccel installation to ensure gFTL is provided in sdist installations and to ensure that STC is installed alongside Pyccel. This will improve compile times as in most cases, STC will only be compiled once, at installation. Fixes #2516 

**Commit Summary**
- Update checks for `STC` and `gFTL` to look for generated files using `importlib.resources` which should find packages regardless of their location (which may be different for editable or non-editable installations).
- Deploy to PyPI using sdist instead of wheel so `STC` is compiled on the target machine. This means we don't need to deploy 1 wheel per architecture/OS
- Add a workflow to test the sdist installation
- Correct gitignore filter
- Remove `install_scripts` folder as these scripts are superceded by the CMake install procedure
- Add a `CMakeLists.txt` describing Pyccel installation
- Stop automatically adding folder containing libraries to includes
- Use `importlib.resources` to locate `STC` and `gFTL`
- Search for installed `STC` then search for a version installed by Pyccel using the same compiler family. Finally fall back to installing `STC` from source
- Search for installed `gFTL` then use the version installed by Pyccel
- Update `pyproject.toml` to use `scikit-build-core` as the backend

### DIFF
--- a/.github/workflows/sdist.yml
+++ b/.github/workflows/sdist.yml
@@ -48,7 +48,7 @@ jobs:
 
   Sdist-installation:
     runs-on: ubuntu-latest
-    name: Deploy tests
+    name: Test deployment
     needs: Python_version_picker
 
     steps:


pyproject.toml
```diff
@@ -78,9 +93,13 @@ include = [
...
   "pyccel/extensions/STC/meson.build",
-  "pyccel/extensions/STC/meson_options.txt"
+  "pyccel/extensions/STC/meson_options.txt",
+  "pyccel/extensions/gftl_install"
 ]
 
-[tool.hatch.build.targets.wheel.hooks.custom]
-path = "install_scripts/hook.py"
-require-runtime-dependencies = true
+sdist.exclude = ['*']
```
yguclu: What files are we excluding exactly?

.github/actions/check_for_gftl/action.yml
```diff
@@ -9,11 +9,11 @@ runs:
...
-        from os.path import exists, join
-        pyccel_loc = pyccel.__path__[0]
-        print(pyccel_loc)
-        assert exists(join(pyccel_loc,'extensions/gFTL/install/GFTL-1.13/include/v2'))
-        if ${{ inputs.not_editable }}:
-          assert not exists(join(pyccel_loc,'extensions/gFTL/doc'))
+        import importlib.resources
+        gftl_folder = importlib.resources.files(f'pyccel.extensions.gftl_install')
+        gftl_folder_gen = importlib.resources.as_file(gftl_folder)
+        with gftl_folder_gen as f:
```
yguclu: Since `importlib.resources.as_file` returns a context manager, I think it is safer to call it in the `with` block:

```suggestion
        with importlib.resources.as_file(gftl_folder) as f:
```

.github/actions/check_for_stc/action.yml
```diff
@@ -9,10 +9,11 @@ runs:
...
-        import pyccel, os
-        from os.path import exists, join
-        pyccel_loc = pyccel.__path__[0]
-        print(pyccel_loc)
-        print("${{ inputs.not_editable }}")
-        assert exists(join(pyccel_loc,'extensions/STC/include/stc'))
+        import importlib.resources
+        stc_folder = importlib.resources.files(f'pyccel.extensions.stc_install_GNU')
+        stc_folder_gen = importlib.resources.as_file(stc_folder)
+        with stc_folder_gen as f:
```
yguclu: As above:
```suggestion
        with importlib.resources.as_file(stc_folder) as f:
```

pyccel/codegen/compiling/library_config.py
```diff
@@ -499,26 +521,25 @@ def install_to(self, pyccel_dirpath, installed_libs, verbose, compiler):
...
-        self._discovery_method = 'CMake'
         sep = ';' if sys.platform == "win32" else ':'
         CMAKE_PREFIX_PATH = os.environ.get('CMAKE_PREFIX_PATH', '').split(sep)
-        os.environ['CMAKE_PREFIX_PATH'] = ':'.join(s for s in (*CMAKE_PREFIX_PATH, str(dest_dir))
-                                                   if s and Path(s).exists())
 
-        return new_obj
+        gftl_installation = importlib.resources.files('pyccel.extensions.gftl_install')
+        gftl_installation_gen = importlib.resources.as_file(gftl_installation)
+        with gftl_installation_gen as f:
```
yguclu: Same comment as above. I believe we should only create the context manager in the `with` block:

```suggestion
        with importlib.resources.as_file(gftl_installation) as f:
```

.github/workflows/sdist.yml
```diff
@@ -0,0 +1,118 @@
...
+        echo "version=${{ inputs.python_version }}" >> $GITHUB_ENV
+      shell: bash
+    - id: set-python_version
+      run: |
+        echo "python_version=$version" >> $GITHUB_OUTPUT
+      shell: bash
+
+  Wheel-installation:
+    runs-on: ubuntu-latest
+    name: Deploy tests
```
yguclu: Why calling this job "Deploy tests"?

docs/installation.md
```diff
@@ -111,11 +111,13 @@ pacman -S mingw-w64-x86_64-gcc-fortran
...
 pacman -S mingw-w64-x86_64-cmake
 pacman -S mingw-w64-x86_64-meson
-pacman -S m4
 pacman -S pkg-config
 # Install MPI and Lapack for library support
 pacman -S mingw-w64-x86_64-msmpi
 pacman -S mingw-w64-x86_64-lapack
+
+# Install if compiling from source
+pacman -S m4
```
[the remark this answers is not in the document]
EmilyBourne: My aim was to not be stylistically consistent. All other dependencies are compulsory for the installation to work properly or to use the code properly. This is only temporarily necessary during the source compilation (actually from source not from sdist)

.github/workflows/sdist.yml
EmilyBourne: Because it tests if you can create the sdist object that is created during the deployment

PR #2524, opened 2025-12-03
Convert Fortran array shapes to int32 for output to avoid warning

EmilyBourne:
Convert Fortran array shapes to int32 for output to avoid warning in bind c file.


yguclu (approved):
Looks good to me!

PR #2529, opened 2025-12-08
Fix `pyccel test` command

EmilyBourne:
Fix `pyccel test` command by ensuring that `pyccel clean` is called between commands and that `sys.exit` is called to raise errors. Missing language-agnostic tests are also run.

A minor bug fix is also included for an incorrect key in Errors.


yguclu (approved):
Looks good to me

PR #2541, opened 2026-01-09
Add axis/keepdims/initial arguments to amin/amax

EmilyBourne:
Use the new `NumpyReduction` class to add support for the `axis`, `keepdims`, and `initial` arguments of `numpy.amin` and `numpy.amax`. These functions are equivalent to `a.min()` and `a.max()` syntax so the arguments are also added for these classes.

### DIFF
--- a/tests/epyccel/recognised_functions/test_numpy_funcs.py
+++ b/tests/epyccel/recognised_functions/test_numpy_funcs.py
@@ -2751,7 +2751,7 @@ def test_min_real(language):
 
     f1 = epyccel(min_call, language = language)
     x = rand(10)
-    assert isclose(f1(x), min_call(x), rtol=RTOL, atol=ATOL)
+    assert np.array_equal(f1(x), min_call(x))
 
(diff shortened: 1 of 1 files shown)


pyccel/codegen/printing/fcode.py
```diff
@@ -1869,12 +1869,15 @@ def _print_NumpyAmax(self, expr):
...
+            code = f'amax({array_arg})'
         elif is_bool: # Convert INTEGER to LOGICAL
             zero = self._print(LiteralInteger(0))
-            return f'maxval({arg_code}) /= {zero}'
+            code = f'maxval({arg_code}) /= {zero}'
         else:
-            return f'maxval({arg_code})'
+            code = f'maxval({arg_code})'
+        if expr.initial is not None:
+            code = f'minval([{self._print(expr.initial)}, {code}])'
```
EmilyBourne:
```suggestion
            code = f'maxval([{self._print(expr.initial)}, {code}])'
```

PR #2557, opened 2026-01-28
Deprecate `pyccel config export --compiler-config`

### LINKED ISSUE #2555: [Improvement] Deprecate `pyccel config export --compiler-config`

As mentioned by @EmilyBourne in https://github.com/pyccel/pyccel/pull/2554#discussion_r2728730656:

> The problem is that `pyccel config export --compiler-config` is a nonsensical command. It is included because `pyccel config` inherits the way of choosing the configuration from `pyccel compile` but `pyccel config export --compiler-config my_config.json exported_config.json` is equivalent to `cp my_config.json exported_config.json`

Therefore, the flag `--compiler-config` should be removed from the `pyccel config export` subcommand. As a first step, this should be deprecated in a dedicated PR.

### Record

EmilyBourne:
Deprecate `pyccel config export --compiler-config FILE1.json FILE2.json` which is equivalent to `cp FILE1.json FILE2.json`. Fixes #2555

### DIFF
--- a/pyccel/commands/pyccel_compile.py
+++ b/pyccel/commands/pyccel_compile.py
@@ -101,7 +101,7 @@ def setup_pyccel_compile_parser(parser):
     add_common_settings(group)
     group.add_argument('--export-compiler-config', action='store_true',
                         help='Export all compiler information to a JSON file with the given path (relative or absolute). '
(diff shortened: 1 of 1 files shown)


yguclu (changes requested):
Good job! I just have a couple of minor comments

Issue #791, opened 2021-03-04
Some numpy functions are missing thier printer in C


mostafaMamoni:
- [ ] mod
- [ ] norm
- [ ] matmul
- [ ] prod
- [ ] linspace
- [ ] where
- [ ] shape

**Language**
C

EmilyBourne:
Related to #789

EmilyBourne:
Are you sure about `shape` ? I think it is translated to a different object (e.g. a `PythonTuple` of `PyccelArraySize`) but I think it still can be translated

mostafaMamoni:
> Are you sure about `shape` ? I think it is translated to a different object (e.g. a `PythonTuple` of `PyccelArraySize`) but I think it still can be translated

i am not sure that's why i put it last, so it can be removed from the list.

yguclu:
What about `dot`? If we try to import it from Numpy we get this error:
```bash
ERROR at annotation (semantic) stage
pyccel:
 |error [semantic]: test_numpy.py [1]| Function 'dot' from module 'numpy' is not currently supported by pyccel (from numpy import dot)
```

EmilyBourne:
> @[yguclu](https://github.com/yguclu) yguclu pinned this issue [on 9 Apr 2021](https://github.com/pyccel/pyccel/issues/791#event-4572751020)

Most of the remaining functions require code to be written potentially both for Fortran and C. I think we should decide what our strategy should be for auto-translating modules to provide necessary code. This will probably require a meeting. I'm not sure we have time for that at the moment though. Should we unpin this issue in the meantime to free up the spot for issues that we want solved sooner?

yguclu:
We urgently need the function `np.sum` in order to run some of our benchmarks with the C printer. @jalalium Do you think you could have a look at it?

For now we just need something that works. In the future we will search for the most efficient implementation, which is robust with respect to the accumulation of round-off errors.

jalalium:
> We urgently need the function [`np.sum`](https://numpy.org/doc/stable/reference/generated/numpy.sum.html) in order to run some of our benchmarks with the C printer. @jalalium Do you think you could have a look at it?
> 
> For now we just need something that works. In the future we will search for the most efficient implementation, which is robust with respect to the accumulation of round-off errors.
> 
> For starters I would only support the single-argument case, i.e. `np.sum(a)`.

Sure, I will work on it right away!

PR #2562, opened 2026-02-02
Add C support for np.matmul

### LINKED ISSUE #244: Improvements in Matmul function
Now that the basic version is working in 2f9678ae106a98ba865830f12fd1386488141dcd and #235 we should 

1) add more tests with different shapes and order
2) make the two failing unit tests work:
https://github.com/pyccel/pyccel/blob/2f9678ae106a98ba865830f12fd1386488141dcd/tests/epyccel/test_arrays.py#L779-L808
3) check `shape` of the output carefully

### LINKED ISSUE #767: Numpy.matmul doesn't work for arrays with rank>2
**Describe the bug**
[see below](https://github.com/pyccel/pyccel/issues/767#issuecomment-815020738)

**To Reproduce**
```python
    @types('int[:,:,:]')
    def test_int(arr):
        from numpy import matmul, shape
        a = matmul(arr, arr)
        return a[0]
```

compilation error message:
```
   19 |     a = matmul(arr,arr)
      |               1
Error: 'matrix_a' argument of 'matmul' intrinsic at (1) must be of rank 1 or 2
```

**Expected behavior**
a generated code that works with arrays with rank>2 in Fortran.

**Language**
Fortran

**Additional context**
- The tests are needed for this case.
- [numpy-functions.md](https://github.com/pyccel/pyccel/blob/master/tutorial/numpy-functions.md) should be updated after resolving this issue.

EmilyBourne:
Improve `np.matmul` translation. Support for higher rank inputs is added which fixes #767. The associated tests close #244. A C printer is also added (see #791).

`matmul` can be called as a batched function but after unravelling the function call must still be 2D. To reflect this a new property is added `is_indexable` to indicate if an index can be inserted into the expression.

**Commit Summary**
- Add new `PyccelFunction.is_indexable` property
- Add `dtype` and `order` arguments to `NumpyMatmul`
- Handle the shape of a `matmul` call for a variety of shapes
- Handle unravelling of functions that operate on non-scalar values
- Add stc dependency to math stdlib
- Add `_inline_vecdot` method to `CCodePrinter` to avoid code duplication between `_print_NumpyVecdot` and `_print_NumpyMatmul` called with 1D arguments
- Add C support for matmul
- Remove unnecessary if block in semantic.py
- Add implementations of matrix multiplications in stdlib/math
- Enable C tests for matmul
- Don't test boolean matmul as this is not implemented in stdlib/math and is not very useful (to avoid a new type being introduced the number of int type tests is also reduced)
- Add tests with batches
- Use `arange` for matmul tests so the order is tested properly

### DIFF
--- a/tests/epyccel/test_arrays.py
+++ b/tests/epyccel/test_arrays.py
@@ -1977,9 +1977,9 @@ def test_array_float_1d_1d_prod(language):
 def test_array_float_2d_1d_matmul(language):
     f1 = arrays.array_float_2d_1d_matmul
     f2 = epyccel( f1 , language = language)
-    A1 = np.arange(6, dtype=float).reshape(3, 2) + 1
+    A1 = np.arange(1, 7, dtype=float).reshape(3, 2)
     A2 = np.copy(A1)
-    x1 = np.arange(2, dtype=float) + 30
+    x1 = np.arange(30, 32, dtype=float)
     x2 = np.copy(x1)
     y1 = np.empty([3])
     y2 = np.empty([3])
@@ -1991,9 +1991,9 @@ def test_array_float_2d_1d_matmul(language):
 def test_array_float_2d_1d_matmul_creation(language):
     f1 = arrays.array_float_2d_1d_matmul_creation
     f2 = epyccel( f1 , language = language)
-    A1 = np.arange(6, dtype=float).reshape([3, 2]) + 1
+    A1 = np.arange(1, 7, dtype=float).reshape([3, 2])
     A2 = np.copy(A1)
-    x1 = np.arange(2, dtype=float) - 10
+    x1 = np.arange(-10, -8, dtype=float)
     x2 = np.copy(x1)
     y1 = f1(A1, x1)
     y2 = f2(A2, x2)
@@ -2003,9 +2003,9 @@ def test_array_float_2d_1d_matmul_creation(language):
 def test_array_float_2d_1d_matmul_order_F_F(language):
     f1 = arrays.array_float_2d_1d_matmul_order_F
     f2 = epyccel( f1 , language = language)
-    A1 = np.arange(6, dtype=float).reshape([3, 2], order='F') + 1
+    A1 = np.arange(1, 7, dtype=float).reshape([3, 2], order='F')
     A2 = np.copy(A1)
-    x1 = np.arange(2, dtype=float) + 10
+    x1 = np.arange(10, 12, dtype=float)
     x2 = np.copy(x1)
     y1 = np.empty([3])
     y2 = np.empty([3])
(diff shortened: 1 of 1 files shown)


pyccel/stdlib/math/CMakeLists.txt
```diff
@@ -20,6 +20,7 @@ target_include_directories(pyc_math_c
 target_link_libraries(pyc_math_c
     PRIVATE
     m
+    stc
 )
```
yguclu: Should we not use the same indentation here as in the `if` case above?

pyccel/stdlib/math/pyc_math_c.c
```diff
@@ -122,3 +123,86 @@ extern inline int8_t py_floor_div_int8_t(int8_t x, int8_t y);
...
+PYC_MATVECMUL_TYPE(int16_t       , array_int16_t_2d       , array_int16_t_1d)
+PYC_MATVECMUL_TYPE(int32_t       , array_int32_t_2d       , array_int32_t_1d)
+PYC_MATVECMUL_TYPE(int64_t       , array_int64_t_2d       , array_int64_t_1d)
+PYC_MATVECMUL_TYPE(float         , array_float_2d         , array_float_1d)
+PYC_MATVECMUL_TYPE(double        , array_double_2d        , array_double_1d)
+PYC_MATVECMUL_TYPE(float_complex , array_float_complex_2d , array_float_complex_1d)
+PYC_MATVECMUL_TYPE(double_complex, array_double_complex_2d, array_double_complex_1d)
+
+#define PYC_VECMATMUL_TYPE(TYPE, TYPE2D, TYPE1D) \
+void pyc_vecmatmul_##TYPE(TYPE1D out, TYPE1D a, TYPE2D b) \
```
yguclu: Same question as above

yguclu (changes requested):
Thanks for this PR, it looks really good to me!

I just have a couple of comments

tests/epyccel/test_arrays.py
```diff
@@ -1974,21 +1974,12 @@ def test_array_float_1d_1d_prod(language):
...
-        pytest.param("fortran", marks = pytest.mark.fortran),
-        pytest.param("python", marks = pytest.mark.python)
-    ]
-)
 def test_array_float_2d_1d_matmul(language):
     f1 = arrays.array_float_2d_1d_matmul
     f2 = epyccel( f1 , language = language)
-    A1 = np.ones([3, 2])
-    A1[1,0] = 2
+    A1 = np.arange(6, dtype=float).reshape(3, 2)
```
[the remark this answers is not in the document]
EmilyBourne: No longer starting at 0 : b17e8dcbb2a49ead250ba0ee7bb473f9c7aa9e4d

PR #2567, opened 2026-02-25
Fix Pyccel installation

EmilyBourne:
Following the release of `scikit-build-core` v0.12.0 the way in which files are included and excluded has changed. This has broken Pyccel's installation as can be seen in [our CI](https://github.com/pyccel/pyccel/actions/runs/22393404647/job/64826093764?pr=2470) and in the [Pyccel benchmarks CI](https://github.com/pyccel/pyccel-benchmarks/actions/runs/22392771289/job/64818735337)

<details>

<summary>
Documentation of the scikit-build-core changes
</summary>

The relevant section of the [scikit-build-core CHANGELOG](https://scikit-build-core.readthedocs.io/en/latest/about/changelog.html#version-0-12-0) is:
> This release adds a new sdist.inclusion-mode setting, with a new default: ignored directories are no longer traversed unless a file is specifically allowed inside of it via the base path. This should be faster and less surprising. The old behavior (“classic”) and no gitignore (“manual”) are also provided.

The [relevant documentation about the new setting](https://scikit-build-core.readthedocs.io/en/latest/api/scikit_build_core.settings.html#scikit_build_core.settings.skbuild_model.SDistSettings.inclusion_mode) is:

> Method to use to compute the files to include and exclude.
> 
> The methods are:
> 
> “default”: Process the git ignore files. Shortcuts on ignored directories.
> 
> “classic”: The behavior before 0.12, like “default” but does not shortcut directories.
> 
> “manual”: No extra logic, based on include/exclude only.
> 
> If you don’t set this, it will be “default” unless you set the minimum version below 0.12, in which case it will be “classic”.

</details>

The new setting is now used which increases the minimum required version of `scikit-build-core`. In order to prevent future changes to `scikit-build-core` breaking Pyccel again the setting `minimum-version` is added to guarantee backwards compatibility.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -73,6 +73,8 @@ logging.level = "INFO"
 # Hook sdist build to insert gFTL compilation and avoid m4 requirement
 sdist.cmake = true
 
+sdist.inclusion-mode = "manual"
+
 sdist.include = [
   "AUTHORS",
   "CHANGELOG.md",
@@ -100,8 +102,6 @@ sdist.include = [
   "pyccel/extensions/STC/meson_options.txt",
   "pyccel/extensions/gftl_install"
 ]
-
-sdist.inclusion-mode = "manual"
 
 [tool.scikit-build.metadata.version]
 # Set version info


yguclu (changes requested):
Looks good to me. I have a single suggestion!

PR #2572, opened 2026-02-28
Enable NumPy array functions in compound expressions

### LINKED ISSUE #2566: Cannot use temporary NumPy array in expression with C
### Bug description

The following code cannot be translated to C

```python
from numpy  import zeros
from pyccel import epyccel

def f():
    a = zeros(3) + 2
    return a.sum()

fc = epyccel(f, language='C')
assert f() == fc()
```

because a `PyccelCodegenError` is raised:

```
$ python temp_array_in_expr.py 

pyccel:
 |fatal [codegen]: ~/tmp/bug/__epyccel__/mod_sfy5e6xmit8v.py [2,4]| Loop unravelling cannot handle extraction of function calls which return arrays as this requires allocation. Please place the function call on its own line (a := <pyccel.ast.numpyext.NumpyZeros object at 0x111cf6b20> + 2)
  File "~/tmp/bug/temp_array_in_expr.py", line 8, in <module>
    fc = epyccel(f, language='C')
  File "~/Repos/PYCCEL/pyccel/pyccel/commands/epyccel.py", line 611, in epyccel
    raise type(e)(str(e)) from None
pyccel.errors.errors.PyccelCodegenError: Loop unravelling cannot handle extraction of function calls which return arrays as this requires allocation. Please place the function call on its own line
```

The code generation works correctly using the Fortran or Python printers. I expect that the code snippet above should also be supported by our C printer.

### Record

emmanuel-ferdman:
Allow array-creating functions like `zeros()` and `ones()` to be used directly in expressions when targeting C. Previously, code like `a = zeros(3) + 2` would fail because the loop unravelling didn't handle array temporaries properly. Now it creates the right temporary variables and indexes them correctly in the generated loops.

Fixes #2566

---

## Pull request checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -38,35 +38,57 @@ jobs:
       - uses: actions/checkout@v6
         with:
           path: origin
+          fetch-depth: 0
       - name: Get PR info
         id: pr_info
         run: |
           cd origin # Run in origin so gh connects to target repo
-          PR_ID=$(gh pr list --state open --json number,headRefOid \
-                          --jq '.[] | select(.headRefOid == "${{ github.event.workflow_run.head_sha }}") | .number')
-          echo "PR_ID=${PR_ID}" >> $GITHUB_ENV
+
+          # Define HEAD_SHA and PR_REPO environment variables
           if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
             HEAD_SHA="${{ github.sha }}"
-            echo "head_sha=${HEAD_SHA}" >> $GITHUB_OUTPUT
-            echo "pr_repo=${{ inputs.pr_repo || github.repository }}" >> $GITHUB_OUTPUT
-            if [ -n "${{ inputs.run_id }}" ]
-            then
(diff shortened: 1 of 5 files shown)


tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -2139,6 +2139,22 @@ def create_ones_3_val():
     assert isclose(f3_val(), create_ones_3_val(), rtol=RTOL, atol=ATOL)
     assert matching_types(f3_val(), create_ones_3_val())
 
+def test_ones_in_expression(language):
+    def ones_plus_scalar():
+        from numpy import ones
+        a = ones(3) + 2
+        return a.sum()
```
yguclu: For this function I suggest creating an integer array here, and then verifying the result with an exact equality check.
```suggestion
    def ones_plus_scalar():
        from numpy import ones
        a = ones(3, dtype=int) + 2
        return a.sum()
```

yguclu (commented):
@emmanuel-ferdman Thanks for applying the changes. Before proceeding with merging this PR we should wait for #2575 which is almost ready

PR #2606, opened 2026-03-19
Make class methods virtual by default

EmilyBourne:
In C function pointers are required in order to dispatch function calls at runtime. This is necessary for class inheritance. This PR adds the necessary print methods to make methods virtual by default. This is a first step towards handling inheritance. The next PR will add support for the `@final` decorator for classes and methods in order to retrieve performance 

---

## Pull request checklist

- [x] Update CHANGELOG.md


docs/classes.md
```diff
@@ -74,76 +74,104 @@ MyClass Object created!
 ### - Header File Equivalent
 
 ```C
-struct MyClass {
+struct cls_test__MyClass {
     int64_t param1;
     array_double_1d param2;
     bool is_freed;
+    void (*get_param)(struct cls_test__MyClass* self);
```
yguclu, line 81: Shouldn't this read
```suggestion
    void (*cls_test__MyClass__get_param)(struct cls_test__MyClass* self);
```
?

docs/classes.md
```diff
@@ -74,76 +74,104 @@ MyClass Object created!
...
 }
 /*........................................*/
+
 /*........................................*/
-void MyClass1__init__(struct MyClass1* self)
+void cls_test__MyClass1__init(struct cls_test__MyClass1* self)
 {
+    // Save virtual function addresses
+    self->Method2 = cls_test__MyClass1__Method2;
+    self->Method1 = cls_test__MyClass1__Method1;
```
yguclu, line 152: In which order are these lines printed?

Would it be possible to print them alphabetically?

pyccel/codegen/printing/ccode.py
```diff
@@ -3850,13 +3872,54 @@ def _print_Del(self, expr):
...
+
+        # Print __init__ with injected function pointer assignments
+        init_method = expr.get_method("__init__")
+        init_printed = self._print(init_method)
+        if virtual_methods:
+            init_lines = init_printed.split("\n")
+            self_name = init_method.arguments[0].var.name
+            fp_assignments = "    // Save virtual function addresses\n" + "".join(
+                f"    {self_name}->{m.cls_name} = {m.name};\n" for m in virtual_methods
+            )
```
yguclu, line 3955: See my previous comment about the ordering of these assignments

docs/classes.md
EmilyBourne: No I don't think so. The function pointer is a member of the class so the scoping already differentiates between different methods called `get_param`. The object `cls_test__MyClass__get_param` is a free function taking an instance of type `struct cls_test__MyClass*`. `get_param` is a function pointer which may be set to point at this free function, or at a different free function with a compatible signature (i.e. the function that overrides `get_param` in a subclass).

docs/classes.md
EmilyBourne: It is printed in the order they are stored, which is the order in which they are visited. Here the main is:
```python
if __name__ == '__main__':
    obj = MyClass1()
    obj.Method1(obj.Method2())
    obj.param.get_param()
```
which means `Method2` gets visited first. It should be possible to make this alphabetical though

EmilyBourne: Done : [81d0f86](https://github.com/pyccel/pyccel/pull/2606/commits/81d0f8679305695989cf0c45a0cd1b24374ba587)

pyccel/codegen/printing/ccode.py
EmilyBourne: Done : [81d0f86](https://github.com/pyccel/pyccel/pull/2606/commits/81d0f8679305695989cf0c45a0cd1b24374ba587)

yguclu: Actually, I was thinking of using `m.cls_name` as the sorting key, not `m.name`🤔

Do you see a reason for preferring one to the other?

EmilyBourne: Ah you are right, they are probably almost equivalent though. `m.name` should be `m.cls_name` with the prefix

EmilyBourne: Changed : [74a6055](https://github.com/pyccel/pyccel/pull/2606/commits/74a605597800194badd9a5c2a47738980e1afb0d)

yguclu (approved):

PR #2608, opened 2026-03-23
Stop `pr_clean_up` workflow triggering when `pr_tests` is skipped

EmilyBourne:
Stop `pr_clean_up` workflow triggering when `pr_tests` is skipped. When the workflow is skipped everything that ran passed so the PR may be marked as `needs_initial_review` even if the PR is in draft.

---

## Pull request checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/.github/workflows/pr_clean_up.yml
+++ b/.github/workflows/pr_clean_up.yml
@@ -14,6 +14,8 @@ jobs:
   update_status:
     name: Update PR status
     runs-on: ubuntu-latest
+    # Possible conclusions : ~action_required~, cancelled, failure, ~neutral~, skipped, stale, success, timed_out, startup_failure, null
+    if: ${{ github.event.workflow_run.conclusion == 'success' || github.event.workflow_run.conclusion != 'failure' }}
 
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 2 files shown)


.github/workflows/pr_clean_up.yml
```diff
@@ -14,6 +14,8 @@ jobs:
   update_status:
     name: Update PR status
     runs-on: ubuntu-latest
+    # Possible conclusions : ~action_required~, cancelled, failure, ~neutral~, skipped, stale, success, timed_out, startup_failure, null
+    if: ${{ github.event.workflow_run.conclusion == 'success' || github.event.workflow_run.conclusion != 'failure' }}
```
yguclu: I am confused by the usage of the bash "or" operator `||` here, given that you consider every conclusion which is not `'failure'` and `success` is already one of them. It seems to me that your `if` condition is equivalent to the simple
```bash
if: ${{ github.event.workflow_run.conclusion != 'failure' }}
```
unless I am missing something here?

EmilyBourne: Oops this is a typo. It should be:
```suggestion
    if: ${{ github.event.workflow_run.conclusion == 'success' || github.event.workflow_run.conclusion == 'failure' }}
```

yguclu (approved):
LGTM!

PR #2611, opened 2026-04-07
Setup plugin structure

EmilyBourne:
Setup a plugin structure which allows internal or external plugins to extend existing functionalities. This is particularly useful when the extensions are dependent on parameters (e.g. OpenMP extensions which depend on the chosen standard). The structure is tested using a small plugin which prints comments in the generated code indicating the line of Python code that was used to generate the subsequent line(s) of code.

Plugins are implemented using the `pluggy` package.

Plugins are implemented through the following compulsory hooks:
- `get_description`
- `add_cli_options`
- `read_cli_arguments`

and the following optional hooks:
- `get_extra_accelerators`
- `get_updated_syntactic_methods`
- `get_updated_semantic_methods`
- `get_codegen_class`
- `get_updated_codegen_methods`
- `get_wrapper_codegen_class`
- `get_updated_wrapper_codegen_methods`
- `get_wrapper_class`
- `get_updated_wrapper_methods`
- `get_build_generation_class`
- `get_updated_build_generation_methods`

For now `get_extra_accelerators` is not linked in, as this requires a different usage framework and cannot be tested with a tiny example. The support will be important when OpenMP is handled via plugins.

The proposed setup should allow a variety of plugins, for example pyccel-cuda should be able to be written as a plugin with this setup.

**Commit Summary**
- Allow plugins to avoid documenting `_visit_X`, `_print_X` and `_extract_X` methods that extend common Pyccel classes.
- Allow `from pyccel import hookimpl` so external libraries can create plugins.
- Call plugin utilities to get main classes (`SyntaxParser`, `SemanticParser`, `CodePrinter`, `Wrapper`, `BuildSystemHandler`) so they can be extended by plugins if necessary
- Change method used to find which wrapper classes are needed in `Wrappergen` to deduce this information from information that can be provided by a plugin
- Update CLI tool to collect and display plugin arguments
- Remove unhelpful main from `syntactic.py`
- Add a line-annotation plugin

---

## Pull request checklist

- [x] Update CHANGELOG.md


pyccel/commands/pyccel_wrap.py
```diff
@@ -110,14 +120,21 @@ def pyccel_wrap(*, filename, language, output, **kwargs) -> None:
...
+    handle_plugin_arguments(plugin_manager, kwargs)
+
     # ...
 
     execute_pyccel_wrap(
-        filename, language=language.lower(), folder=output or filename.parent, **kwargs
+        filename,
+        language=language.lower(),
+        folder=output or filename.parent,
+        plugin_manager=plugin_manager**kwargs,
```
EmilyBourne:
```suggestion
        plugin_manager=plugin_manager,
        **kwargs,
```

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,418 @@
...
+        except AttributeError:
+            continue
+
+        BaseClass = type(
+            name + BaseClass.__name__,
+            (BaseClass,),
+            {m.__name__: m for m in new_methods},
+        )
+
+    return BaseClass
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
Strategy is similar to that of `get_semantic_class` which is tested. This section is left for other plugins to use

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,418 @@
...
+                continue
+            break
+
+    if BaseClass is None:
+        raise ValueError(f"{build_gen_method} build generation method is not available")
+
+    for plugin in plugin_manager.get_plugins():
+        name = plugin_manager.get_name(plugin)
+        try:
+            new_methods = plugin.get_updated_build_generation_methods(build_gen_method)
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
Strategy is similar to that of `get_semantic_class` which is tested. This section is left for other plugins to use

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,418 @@
...
+            new_methods = plugin.get_updated_codegen_methods(language)
+        except AttributeError:
+            continue
+
+        BaseClass = type(
+            name + BaseClass.__name__,
+            (BaseClass,),
+            {m.__name__: m for m in new_methods},
+        )
+
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
Strategy is similar to that of `get_semantic_class` which is tested. This section is left for other plugins to use

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,418 @@
...
+                continue
+            break
+
+    if BaseClass is None:
+        raise ValueError(f"{language} language is not available")
+
+    for plugin in plugin_manager.get_plugins():
+        name = plugin_manager.get_name(plugin)
+        try:
+            new_methods = plugin.get_updated_codegen_methods(language)
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
Strategy is similar to that of `get_semantic_class` which is tested. This section is left for other plugins to use

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,418 @@
...
+            )
+        except AttributeError:
+            continue
+
+        BaseClass = type(
+            name + BaseClass.__name__,
+            (BaseClass,),
+            {m.__name__: m for m in new_methods},
+        )
+
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
Strategy is similar to that of `get_semantic_class` which is tested. This section is left for other plugins to use

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,418 @@
...
+        raise ValueError(
+            f"Wrapping from {start_language} to {target_language} is not available"
+        )
+
+    for plugin in plugin_manager.get_plugins():
+        name = plugin_manager.get_name(plugin)
+        try:
+            new_methods = plugin.get_updated_wrapper_methods(
+                start_language, target_language
+            )
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
Strategy is similar to that of `get_semantic_class` which is tested. This section is left for other plugins to use

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,418 @@
...
+            new_methods = plugin.get_updated_codegen_methods(language)
+        except AttributeError:
+            continue
+
+        BaseClass = type(
+            name + BaseClass.__name__,
+            (BaseClass,),
+            {m.__name__: m for m in new_methods},
+        )
+
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
Strategy is similar to that of `get_semantic_class` which is tested. This section is left for other plugins to use

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,418 @@
...
+            new_methods = plugin.get_updated_syntactic_methods()
+        except AttributeError:
+            continue
+
+        BaseClass = type(
+            name + BaseClass.__name__,
+            (BaseClass,),
+            {m.__name__: m for m in new_methods},
+        )
+
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
Strategy is similar to that of `get_semantic_class` which is tested. This section is left for other plugins to use

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,418 @@
...
+    type
+        A (possibly new) class derived from BaseClass augmented with the plugin methods.
+    """
+    for plugin in plugin_manager.get_plugins():
+        name = plugin_manager.get_name(plugin)
+        try:
+            new_methods = plugin.get_updated_semantic_methods()
+        except AttributeError:
+            continue
+
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
Only used by plugins that don't implement `get_updated_semantic_methods`

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,418 @@
...
+    Raises
+    ------
+    ValueError
+        If BaseClass is `None` and no plugin provides a class for language.
+    """
+    if BaseClass is None:
+        for plugin in plugin_manager.get_plugins():
+            try:
+                BaseClass = plugin.get_codegen_class(language)
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
This can only be tested by a plugin implementing a new language

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,418 @@
...
+            except AttributeError:
+                continue
+            break
+
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
This can only be tested by a plugin implementing a new language

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,418 @@
...
+    Raises
+    ------
+    ValueError
+        If BaseClass is `None` and no plugin provides a suitable wrapper class.
+    """
+    if BaseClass is None:
+        for plugin in plugin_manager.get_plugins():
+            try:
+                BaseClass, target_language = plugin.get_wrapper_class(start_language)
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
This can only be tested by a plugin implementing a new language

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,418 @@
...
+            except (AttributeError, TypeError):
+                continue
+            break
+
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
This can only be tested by a plugin implementing a new language

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,418 @@
...
+        If BaseClass is `None` and no plugin provides a class for language.
+    """
+    if BaseClass is None:
+        for plugin in plugin_manager.get_plugins():
+            try:
+                BaseClass = plugin.get_codegen_class(language)
+            except AttributeError:
+                continue
+            break
+
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
This can only be tested by a plugin implementing a new language

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,418 @@
...
+    Raises
+    ------
+    ValueError
+        If BaseClass is `None` and no plugin provides a class for build_gen_method.
+    """
+    if BaseClass is None:
+        for plugin in plugin_manager.get_plugins():
+            try:
+                BaseClass = plugin.get_build_generation_class(build_gen_method)
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
This can only be tested by a plugin implementing a new language

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,418 @@
...
+            except AttributeError:
+                continue
+            break
+
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
This can only be tested by a plugin implementing a new language

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,418 @@
...
+    Raises
+    ------
+    ValueError
+        If BaseClass is `None` and no plugin provides a class for language.
+    """
+    if BaseClass is None:
+        for plugin in plugin_manager.get_plugins():
+            try:
+                BaseClass = plugin.get_wrapper_codegen_class(language)
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
This can only be tested by a plugin implementing a new language

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,418 @@
...
+                continue
+            break
+
+    if BaseClass is None:
+        raise ValueError(f"{language} language is not available")
+
+    for plugin in plugin_manager.get_plugins():
+        name = plugin_manager.get_name(plugin)
+        try:
+            new_methods = plugin.get_updated_wrapper_codegen_methods(language)
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
Strategy is similar to that of get_semantic_class which is tested. This section is left for other plugins to use

pyproject.toml
```diff
@@ -26,6 +26,7 @@ dependencies = [
     "pybind11",
     "ninja",
     "meson",
+    "pluggy" #TODO: Version
```
yguclu, line 29: I think we should mention this new dependency in the PR description, and possibly also in the CHANGELOG

developer_docs/plugins.md
```diff
@@ -0,0 +1,22 @@
...
+Pyccel's plugin system is built on [pluggy](https://pluggy.readthedocs.io/).
+
+## How it works
+
+Available hooks are declared in [`pyccel/plugins/hookspecs.py`](../pyccel/plugins/hookspecs.py). Plugins implement the three hooks required for the command line interface, as well as any additional hooks relevant for the functionality. The helpers in [`pyccel/plugins/plugin_tools.py`](../pyccel/plugins/plugin_tools.py) build augmented `parser/codegen/wrapping` classes by layering plugin methods on top of the base classes.
+
+## Writing a plugin
+
+1. Create a module that imports `hookimpl` from `pyccel.plugins` and decorates each hook implementation with it.
+2. Publish it as a Python package with the entry-point group `pyccel`.
```
yguclu: What do you mean with "entry-point group"? The name of the package has to be in the form `pyccel-<PLUGIN-NAME>`? Is this enforced somehow?

developer_docs/plugins.md
```diff
@@ -0,0 +1,22 @@
...
+
+Pyccel's plugin system is built on [pluggy](https://pluggy.readthedocs.io/).
+
+## How it works
+
+Available hooks are declared in [`pyccel/plugins/hookspecs.py`](../pyccel/plugins/hookspecs.py). Plugins implement the three hooks required for the command line interface, as well as any additional hooks relevant for the functionality. The helpers in [`pyccel/plugins/plugin_tools.py`](../pyccel/plugins/plugin_tools.py) build augmented `parser/codegen/wrapping` classes by layering plugin methods on top of the base classes.
+
+## Writing a plugin
+
+1. Create a module that imports `hookimpl` from `pyccel.plugins` and decorates each hook implementation with it.
```
yguclu: The PR documentation reads that `hookimpl` can be imported directly from the Pyccel package with `from pyccel import hookimpl`, and in fact the file `pyccel.__init__.py` supports that. Is there a reason to prefer the syntax `from pyccel.plugins import hookimpl` instead?

pyccel/plugins/LineAnnot/plugin.py
```diff
@@ -0,0 +1,68 @@
...
+    Read any arguments from the kwargs dictionary.
+
+    The LineAnnot plugin has no specific arguments to read, so this hook is a
+    no-op.
+
+    Parameters
+    ----------
+    kwargs : dict
+        The keyword arguments passed to the pipeline.
+    """
```
yguclu, line 54: Is there no need to decorate this function with `@hookimpl`?

pyccel/plugins/hookspecs.py
```diff
@@ -0,0 +1,320 @@
...
+    Parameters
+    ----------
+    parser : ArgumentParser
+        The argument parser to which any options should be added.
+    cli_tool : str
+        The name of the tool being used.
+    """
+
+
+def read_cli_arguments(kwargs: dict):
```
yguclu, line 50: `@hookspec` is not needed if the function does not return anything?

pyccel/plugins/hookspecs.py
```diff
@@ -0,0 +1,320 @@
...
+@hookspec
+def get_extra_accelerators() -> Iterable[str]:
+    """
+    Return the names of any additional accelerators required by the plugin.
+
+    Return the names of any additional accelerators required by the plugin.
+    These accelerators affect the compilation of the code and should appear
+    in the compiler configurations.
+
+    This hook is optional. It should only be implemented if it is needed for the implementation.
```
yguclu: TBH I am not totally sure of the purpose of an optional hook. If I understand things correctly, this sentence should be a bit clearer:

```suggestion
    This hook is optional. It should only be implemented if the plugin needs it.
```

pyccel/plugins/hookspecs.py
yguclu, line 1: Could you please explain the mechanism that makes the first three hooks mandatory, and the others optional? How is this enforced?

pyccel/plugins/hookspecs.py
```diff
@@ -0,0 +1,320 @@
...
+    Iterable[str]
+        Names of the accelerators made available by the plugin
+        (e.g. ``['openmp']``).
+    """
+
+
+@hookspec
+def get_updated_syntactic_methods() -> Iterable[FunctionType]:
+    """
+    Return methods to be added to or to override methods in the syntactic parser class.
```
yguclu, line 88: Is the name of the returned function what determines if a specific method is overridden?

pyccel/plugins/hookspecs.py
```diff
@@ -0,0 +1,320 @@
...
+    Parameters
+    ----------
+    language : str
+        The target language (e.g. 'cuda').
+
+    Returns
+    -------
+    type
+        The code-generation class for the requested language or None if this
+        plugin doesn't implement the requested language.
```
yguclu, line 150: I would expect the returned type to be a subclass of `CodePrinter`. Do you think that could be too restrictive for a plugin?

BTW, why do we need both functions `get_codegen_class` and `get_updated_codegen_methods`? The class should already have its own methods, right? Are these alternative ways to achieve the same result?

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,418 @@
...
+        )
+
+        plugin.add_cli_options(parser=group, cli_tool=cli_tool)
+
+    parser.set_defaults(plugin_manager=plugin_manager)
+
+
+def handle_plugin_arguments(plugin_manager, kwargs):
+    """
+    Use kwargs to unregister unused plugins and collect arguments.
```
yguclu, line 77: With "arguments", do you mean CLI arguments?

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,418 @@
...
+            plugin_manager.unregister(plugin)
+        elif not kwargs[name]:
+            plugin_manager.unregister(plugin)
+            kwargs.pop(name)
+        else:
+            plugin.read_cli_arguments(kwargs)
+            kwargs.pop(name)
+
+
+def get_syntactic_class(plugin_manager, BaseClass):
```
yguclu: I do not understand why the argument `BaseClass` is needed, given that the base class is (probably) always `SyntaxParser`.

BTW, even if you expect multiple plugins to redefine `BaseClass` one after another, this function is not called recursively AFAICT. So I would return a variable with a different name, e.g. `SyntacticClass`

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,418 @@
...
+        BaseClass = type(
+            name + BaseClass.__name__,
+            (BaseClass,),
+            {m.__name__: m for m in new_methods},
+        )
+
+    return BaseClass
+
+
+def get_semantic_class(plugin_manager, BaseClass):
```
yguclu: I have the same comment as the one about `get_syntactic_class`

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,418 @@
...
+    Raises
+    ------
+    ValueError
+        If BaseClass is `None` and no plugin provides a suitable wrapper class.
+    """
+    if BaseClass is None:
+        for plugin in plugin_manager.get_plugins():
+            try:
+                BaseClass, target_language = plugin.get_wrapper_class(start_language)
+            except (AttributeError, TypeError):
```
yguclu, line 288: Under what circumstances would a `TypeError` be raised?

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,418 @@
...
+    ValueError
+        If BaseClass is `None` and no plugin provides a class for language.
+    """
+    if BaseClass is None:
+        for plugin in plugin_manager.get_plugins():
+            try:
+                BaseClass = plugin.get_codegen_class(language)
+            except AttributeError:
+                continue
+            break
```
yguclu: Do we stop searching for a base class at the first plugin which implements one? If so, I would add a comment line about this.

I also think that an `else` block would be clearer here:
```suggestion
            else:
                # Stop searching for a codegen class at the first plugin providing one
                break
```

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,418 @@
...
+        The source language of the wrapper (e.g. 'fortran').
+    target_language : str or None
+        The target language of the wrapper (e.g. 'python').
+
+    Returns
+    -------
+    BaseClass : type
+        A (possibly new) class derived from BaseClass augmented with the plugin methods.
+    target_language : str
+        The target language of the wrapper (e.g. 'python').
```
yguclu: `target_language` is both a parameter and a return variable; I find this confusing, as the input argument does not seem to be used. Could you please explain?

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,418 @@
...
+        BaseClass = type(
+            name + BaseClass.__name__,
+            (BaseClass,),
+            {m.__name__: m for m in new_methods},
+        )
+
+    return BaseClass, target_language
+
+
+def get_wrapper_codegen_class(plugin_manager, BaseClass, language):
```
yguclu, line 319: About this function, I have the same comments as for `get_wrapper_class`

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,418 @@
...
+    ValueError
+        If BaseClass is `None` and no plugin provides a class for build_gen_method.
+    """
+    if BaseClass is None:
+        for plugin in plugin_manager.get_plugins():
+            try:
+                BaseClass = plugin.get_build_generation_class(build_gen_method)
+            except AttributeError:
+                continue
+            break
```
yguclu: As in my comment to `get_codegen_class`:

```suggestion
            # Stop searching for a build-gen class at the first plugin providing one
            break
```

tests/pyccel/test_pyccel.py
```diff
@@ -1881,3 +1886,30 @@ def test_complex_numbers(language):
...
+    output_folder = os.path.join(folder, "__pyccel__la__")
+
+    pyccel_commands = (
+        "-t --line_annotation --language=" + language + f" --output={output_folder}"
+    )
+
+    try:
+        os.rmdir(output_folder)
+    except OSError:
+        pass
```
yguclu: `os.rmdir` could also raise a `FileNotFoundError`, which is not handled here. How about this instead?

```suggestion
    shutil.rmtree(output_folder, ignore_errors=True)
```

pyccel/commands/pyccel_compile.py
```diff
@@ -148,19 +158,23 @@ def pyccel_compile(*, filename, language, output, **kwargs):
         The target language Pyccel is translating to.
     output : str
         Path to the working directory.
```
yguclu: `None` is also accepted

yguclu (changes requested):
This looks very useful to me! Great job!

I have a few comments and questions

pyccel/plugins/hookspecs.py
EmilyBourne: This comes simply from the call site. The normal way to call the hooks is:
https://github.com/pyccel/pyccel/blob/4e3953be994b41d26d006f800f7a901ca88e7c58/pyccel/plugins/plugin_tools.py#L99

If the plugin does not implement the hook then the code will fail to call the function.

However I don't think it is useful/necessary to ask every plugin to implement all the hooks for all the different stages (in particular when we add an optimisation stage, the related hooks will work similarly and compulsory hooks would cause pre-existing plugins to fail). Therefore when calling the optional hooks I use:
https://github.com/pyccel/pyccel/blob/4e3953be994b41d26d006f800f7a901ca88e7c58/pyccel/plugins/plugin_tools.py#L163-L166
the try/except catches the error that is raised when the plugin doesn't implement the hook.

pyccel/plugins/hookspecs.py
EmilyBourne: > I would expect the returned type to be a subclass of CodePrinter.

I agree. I'm not sure what the standard way to annotate that is though. If I put:
```suggestion
    CodePrinter
        The code-generation class for the requested language or None if this
        plugin doesn't implement the requested language.
```
then this would be an instance of a `CodePrinter` not the class type itself.

> BTW, why do we need both functions get_codegen_class and get_updated_codegen_methods? The class should already have its own methods, right? Are these alternative ways to achieve the same result?

`get_codegen_class` gets a clean codegen class for the requested language. This is how we can add support for a new language. E.g. `get_codegen_class('cuda')`. The method is only called if a class which prints the language was not already found:
https://github.com/pyccel/pyccel/blob/4e3953be994b41d26d006f800f7a901ca88e7c58/pyccel/plugins/plugin_tools.py#L209-L215

`get_updated_codegen_methods` is the way to extend what an existing printer can do. E.g. if you wanted to add a plugin so the Fortran printer maps NumPy calls to calls in a custom library (e.g. soledge). This should be used rather than completely overriding the existing class to allow multiple plugins to act on the same class.

pyccel/plugins/plugin_tools.py
EmilyBourne: CLI arguments if using `pyccel`, `kwargs` if using `epyccel`

pyccel/plugins/plugin_tools.py
EmilyBourne: For syntactic and semantic the `BaseClass` argument is not really needed. I included it to match the pattern used for codegen, etc but I can remove it.

> BTW, even if you expect multiple plugins to redefine `BaseClass` one after another, this function is not called recursively AFAICT. So I would return a variable with a different name, e.g. `SyntacticClass`

Plugins should not redefine `BaseClass`, the redefinition is done by this function:
https://github.com/pyccel/pyccel/blob/4e3953be994b41d26d006f800f7a901ca88e7c58/pyccel/plugins/plugin_tools.py#L131-L135
The loop ensures that `BaseClass` is updated for each plugin

pyccel/plugins/plugin_tools.py
EmilyBourne: What if is this else supposed to match? Does `try` have an `else` clause? I didn't know this :sweat_smile:

pyccel/plugins/plugin_tools.py
EmilyBourne: The `target_language` on input describes the language we are wrapping to. It is used here:
https://github.com/pyccel/pyccel/blob/4e3953be994b41d26d006f800f7a901ca88e7c58/pyccel/plugins/plugin_tools.py#L287-L290

On output it describes the language we managed to wrap to. In theory this should be the same, but it may be None on entry if we are adding support for a new language. E.g. when wrapping cuda, Pyccel does not support cuda so we will not find a wrapper here:
https://github.com/pyccel/pyccel/blob/4e3953be994b41d26d006f800f7a901ca88e7c58/pyccel/codegen/wrappergen.py#L78
This means we don't know what language we are wrapping to. Most languages wrap to C and then go from C to Python, but some languages (e.g. c++) provide dedicated tools to wrap from that language directly to Python.

After we have called `get_wrapper_class`:
https://github.com/pyccel/pyccel/blob/4e3953be994b41d26d006f800f7a901ca88e7c58/pyccel/plugins/plugin_tools.py#L275
we then know what wrapper we are dealing with exactly.

In theory it should be sufficient to just know the start language, but I am not 100% sure that this is the case (e.g. in theory you can wrap C++ via C instead of using pybind), so I preferred to implement the general case.

developer_docs/plugins.md
EmilyBourne: This refers to the `entry_point` argument in `setup.py` described in pluggy's docs:
https://pluggy.readthedocs.io/en/stable/index.html#the-plugin

pyproject.toml
EmilyBourne: Done : bc2b07baddac7e249b63ea645cf91316affa2120

developer_docs/plugins.md
EmilyBourne: Better : 7728f4b62
I am not 100% sure what goes in this entry-point block. But it should be findable in pluggy's docs if someone wants to add an external plugin. I can look into it more if I change `pyccel-cuda` into a plugin rather than a forked repo

developer_docs/plugins.md
EmilyBourne: Not at all, I had it in `pyccel.plugins` at first, then I moved it but forgot to update the docs.
Fixed : 2ac16e51b

pyccel/plugins/LineAnnot/plugin.py
EmilyBourne: I just forgot. Fixed : 5da57395f869208ae8c45866376841f694626e3e

EmilyBourne: I'm slightly surprised that it still works

pyccel/plugins/hookspecs.py
EmilyBourne: I just forgot. Fixed : 14e7aedc6dfddbf88a86338148f7d347aae2e59f

pyccel/plugins/hookspecs.py
EmilyBourne: Changed : 1d5e622b2

pyccel/plugins/hookspecs.py
EmilyBourne: Yes. This is what I was trying to explain with the sentence:
> The name of the method will remain the same once it is added to the class.

Do you have a clearer suggestion?

pyccel/plugins/hookspecs.py
EmilyBourne: Comment added to hopefully help clarify : 64eb7f651

developer_docs/plugins.md
yguclu: I see. The `pluggy` docs probably refer to this page in the documentation of `setuptools`: https://setuptools.pypa.io/en/latest/userguide/entry_point.html

pyccel/plugins/hookspecs.py
yguclu: Clearer, I am not sure 😅 But I can be more verbose. I am going to add a suggestion shortly

pyccel/plugins/plugin_tools.py
EmilyBourne: Hmm, I think I have just remember why I did this. If I don't pass `SyntaxParser` to this function then I need to import it in the module. I think this may have created a cyclic dependency because the methods in this file are imported in `parser/parser.py`

EmilyBourne: I can't reproduce the error though, I will remove the argument

EmilyBourne: Done : e29c8c61cc9251f33f8d205b6c44b70655c8f3c0

pyccel/plugins/plugin_tools.py
EmilyBourne: More documentation added : 8dca0c0c583b9e53c50345bf01cc42c50c37bb71

pyccel/plugins/plugin_tools.py
EmilyBourne: https://github.com/pyccel/pyccel/commit/e29c8c61cc9251f33f8d205b6c44b70655c8f3c0

pyccel/plugins/plugin_tools.py
EmilyBourne: If a plugin implements `get_wrapper_class` but not the requested language (e.g. `start_language` is cuda but this is a rust plugin) then it will return `None`. `a, b = None` returns a `TypeError`

developer_docs/plugins.md
EmilyBourne: Ah! You are absolutely right. This is a more useful link

EmilyBourne: Added : 2e3c96c6c

pyccel/plugins/hookspecs.py
yguclu: Thanks for the updated docstrings; I think they are clear now!

As for documenting the fact that the returned type should be a subclass of `CodePrinter` (or another base class depending on the specific case), we could just write it explicitly in the description line below the type specification

pyccel/plugins/plugin_tools.py
EmilyBourne: More docs : ec3047ac3

pyccel/plugins/plugin_tools.py
yguclu: Thanks; could you add a line about that in the docstring?

pyccel/plugins/plugin_tools.py
EmilyBourne: Fixed : 42a638f85

Although this is comments plural. The only other comment I see is about `target_language` which is not relevant here. Did I miss something?

pyccel/plugins/plugin_tools.py
EmilyBourne: Fixed : https://github.com/pyccel/pyccel/commit/42a638f85cad663d9d455bb44a5bf623a0713302

pyccel/commands/pyccel_compile.py
EmilyBourne: Fixed : 0e734f9c9c4a15d8ae442c051eac5e10c4a854bb

pyccel/plugins/plugin_tools.py
yguclu: Indeed, I would be in favor of removing the parameter `BaseClass` where it is not needed.

Wherever possible, I would use a dedicated return variable (e.g. `NewBaseClass`) instead of overwriting the input argument

pyccel/plugins/hookspecs.py
EmilyBourne: Done : 7edf63c1b

pyccel/plugins/plugin_tools.py
EmilyBourne: It is possible, but only by doing `NewBaseClass = BaseClass` at the start of the function. I'm not sure why this would be useful?  This also means we return an object called `NewBaseClass` even if no plugins modified the base class

pyccel/plugins/plugin_tools.py
yguclu: Thanks for expanding the docstring!

pyccel/plugins/plugin_tools.py
yguclu: I just prefer to avoid overwriting an existing variable unless absolutely necessary

pyccel/plugins/plugin_tools.py
yguclu: I had no idea that `a, b = None` returns a `TypeError`.. good to know!

pyccel/plugins/plugin_tools.py
EmilyBourne: I'm not sure it makes sense here though. There are 2 places where we overwrite:
1. When finding the value to use instead of None
2. When updating the value in a loop

For the first we could do:
```python
if BaseClass is None:
    NewBaseClass = plugin.get_codegen_class(language)
else:
    NewBaseClass = BaseClass
```
but this seems a little contrived. It is extremely common to do:
```python
def f(a = None):
    if a is None:
        a = {} # not def f(a = {}) as we want a fresh dict at each function call
```
For the second we could do:
```python
NewBaseClass = BaseClass
for plugin in plugin_list:
    NewBaseClass = type(...)
```
but then we are still overwriting

yguclu: Just my personal opinion. I find it OK to overwrite a variable in a loop, given that I see no other clean way to proceed in that case. I would avoid overwriting a variable in any other case, especially if it is an input argument.

This does not look much contrived to me:
```python
NewBaseClass = BaseClass or plugin.get_codegen_class(language)
```
As for the name of the returned variable, we could just drop it from the docstring if we think that it could create confusion

pyccel/plugins/plugin_tools.py
yguclu: It was just a typo, sorry

pyccel/plugins/hookspecs.py
```diff
@@ -0,0 +1,333 @@
...
+
+@hookspec
+def get_updated_syntactic_methods() -> Iterable[FunctionType]:
+    """
+    Return methods to be added to or to override methods in the syntactic parser class.
+
+    The returned callables are injected into a dynamically created subclass
+    of the base syntactic parser, so each callable must accept `self` as
+    its first argument. The name of the method will remain the same once it
+    is added to the class.
```
yguclu:
```suggestion
    its first argument. If the name of the method matches one of
    `SyntaxParser`, the original method is overwritten. Otherwise, the
    method is simply added to the new subclass.
```

pyccel/plugins/hookspecs.py
yguclu: Would [this](https://github.com/pyccel/pyccel/pull/2611#discussion_r3588589922) work?

pyccel/plugins/plugin_tools.py
EmilyBourne: Done : 654d7a3e3

pyccel/plugins/hookspecs.py
EmilyBourne: I made a small modification and added an extra sentence : [a040b80](https://github.com/pyccel/pyccel/pull/2611/commits/a040b80733f24ae9ea883510e03fe60a2be39d08)

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,442 @@
...
+    if BaseClass is None:
+        for plugin in plugin_manager.get_plugins():
+            try:
+                NewBaseClass = plugin.get_build_generation_class(build_gen_method)
+            except AttributeError:
+                continue
+            else:
+                # Stop searching for a build generation class at the first plugin providing one
+                break
+    else:
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
This can only be tested by requesting a language that we don't implement

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,442 @@
...
+    if BaseClass is None:
+        for plugin in plugin_manager.get_plugins():
+            try:
+                NewBaseClass = plugin.get_wrapper_codegen_class(language)
+            except AttributeError:
+                continue
+            else:
+                # Stop searching for a wrapper codegen class at the first plugin providing one
+                break
+    else:
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
This can only be tested by requesting a language that we don't implement

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,442 @@
...
+    if BaseClass is None:
+        for plugin in plugin_manager.get_plugins():
+            try:
+                NewBaseClass, target_language = plugin.get_wrapper_class(start_language)
+            except (AttributeError, TypeError):
+                continue
+            else:
+                # Stop searching for a wrapper class at the first plugin providing one
+                break
+    else:
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
This can only be tested by requesting a language that we don't implement

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,442 @@
...
+            new_methods = plugin.get_updated_codegen_methods(language)
+        except AttributeError:
+            continue
+
+        NewBaseClass = type(
+            name + NewBaseClass.__name__,
+            (NewBaseClass,),
+            {m.__name__: m for m in new_methods},
+        )
+
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
Strategy is similar to that of get_semantic_class which is tested. This section is left for other plugins to use

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,442 @@
...
+    return NewBaseClass
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
Strategy is similar to that of get_semantic_class which is tested. This section is left for other plugins to use

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,442 @@
...
+
+    Raises
+    ------
+    ValueError
+        If BaseClass is `None` and no plugin provides a class for build_gen_method.
+    """
+    if BaseClass is None:
+        for plugin in plugin_manager.get_plugins():
+            try:
+                NewBaseClass = plugin.get_build_generation_class(build_gen_method)
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
This can only be tested by a plugin implementing a new language

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,442 @@
...
+
+    Raises
+    ------
+    ValueError
+        If BaseClass is `None` and no plugin provides a class for language.
+    """
+    if BaseClass is None:
+        for plugin in plugin_manager.get_plugins():
+            try:
+                NewBaseClass = plugin.get_wrapper_codegen_class(language)
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
This can only be tested by a plugin implementing a new language

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,442 @@
...
+
+    Raises
+    ------
+    ValueError
+        If BaseClass is `None` and no plugin provides a suitable wrapper class.
+    """
+    if BaseClass is None:
+        for plugin in plugin_manager.get_plugins():
+            try:
+                NewBaseClass, target_language = plugin.get_wrapper_class(start_language)
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
This can only be tested by a plugin implementing a new language

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,442 @@
...
+    if BaseClass is None:
+        for plugin in plugin_manager.get_plugins():
+            try:
+                NewBaseClass = plugin.get_codegen_class(language)
+            except AttributeError:
+                continue
+            else:
+                # Stop searching for a codegen class at the first plugin providing one
+                break
+    else:
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
This can only be tested by a plugin implementing a new language

pyccel/plugins/plugin_tools.py
```diff
@@ -0,0 +1,442 @@
...
+
+    Raises
+    ------
+    ValueError
+        If BaseClass is `None` and no plugin provides a class for language.
+    """
+    if BaseClass is None:
+        for plugin in plugin_manager.get_plugins():
+            try:
+                NewBaseClass = plugin.get_codegen_class(language)
```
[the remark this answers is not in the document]
EmilyBourne: /bot accept
This can only be tested by a plugin implementing a new language

pyccel/plugins/hookspecs.py
yguclu: Thanks for commit a040b80; that looks perfect to me!

yguclu (approved):
Ready to merge!

PR #2618, opened 2026-07-01
Improve slicing performance in C code

EmilyBourne:
Slicing in C involves a call to a `cspan_slice`. This external function has a cost which makes it hard for the compiler to optimise this and surrounding code. Cspan provides `cspan_submd<RANK>` functions to construct a lower rank submd spans. This is equivalent to `cspan_slice` but faster as it is based on a macro design. Additionally minor slice related bug fixes are incorporated which also improve speed.

**Commit summary**
- Use calls to `cspan_submd<RANK>` when possible
- Ensure the variable generated to save a temporary slice is marked as an alias (prevents unnecessary stack memory allocation)
- Translate expressions such as `a[:]` to `a` to avoid printing unnecessary slicing in the generated code

---

We can see the impact of this change by comparing the benchmarking results from devel:
performance on devel
to the benchmarking results from this branch:
performance on ebourne_c_slicing_performance with a jump in C performance for the time stepping cases

In particular, improvements can be seen for the time stepping methods.

---

## Pull request checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -2227,7 +2227,7 @@ class CCodePrinter(CodePrinter):
                     and expr.rank < base.rank
                     and all(isinstance(i, Slice) for i in inds[expr.rank :])
                 ):
-                    indices_code = ", ".join(indices[: expr.rank])
+                    indices_code = ", ".join(indices[: -expr.rank])
                     return (
                         f"({c_type})cspan_submd{base.rank}({base_code}, {indices_code})"
                     )
--- a/pyccel/codegen/printing/ccode.py
+++ b/pyccel/codegen/printing/ccode.py
@@ -2221,15 +2221,24 @@ class CCodePrinter(CodePrinter):
                         )
                     indices.append(self._print(idx))
                 base_code = self._print(ObjectAddress(base))
-                if expr.rank < 4 and base.rank < 5 and expr.rank < base.rank and all(isinstance(i, Slice) for i in inds[expr.rank:]):
-                    indices_code = ", ".join(indices[:expr.rank])
-                    return f"({c_type})cspan_submd{base.rank}({base_code}, {indices_code})"
+                if (
+                    expr.rank < 4
+                    and base.rank < 5
+                    and expr.rank < base.rank
+                    and all(isinstance(i, Slice) for i in inds[expr.rank :])
+                ):
+                    indices_code = ", ".join(indices[: expr.rank])
+                    return (
+                        f"({c_type})cspan_submd{base.rank}({base_code}, {indices_code})"
+                    )
                 else:
                     indices_code = ", ".join(f"{{{i}}}" for i in indices)
                     return f"cspan_slice({base_code}, {c_type}, {indices_code})"
             else:
                 new_type = base.class_type.switch_rank(expr.rank, expr.order)
-                tmp_var = self.scope.get_temporary_variable(new_type, shape=expr.shape, memory_handling='alias')
+                tmp_var = self.scope.get_temporary_variable(
+                    new_type, shape=expr.shape, memory_handling="alias"
+                )
                 assign = AliasAssign(tmp_var, expr)
                 code = self._print(assign)
                 self._additional_code += code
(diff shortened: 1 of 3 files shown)


yguclu:
Looking at the benchmark results which you posted, it appears that the performance of Numba has dropped considerably on this branch. This does not make much sense to me, can you see an explanation?

EmilyBourne:
> Looking at the benchmark results which you posted, it appears that the performance of Numba has dropped considerably on this branch. This does not make much sense to me, can you see an explanation?

The exact numbers are here:
https://github.com/pyccel/pyccel-benchmarks/compare/4081a754a66fffff68803971d62def004e70a3d0...5d56b8d9dac45dc056754ce666c632d97f345661

I did have that issue on an earlier version. I put it down to the fact that we are using `timeit` instead of `pyperf` (which we use for a version) and bad luck with a busy runner. On this version though I only see the decrease for FD - Laplace. The columns are shorter, but that is because the max value on the y axis has changed now that midpoint-fixed is so much faster.

EmilyBourne:
In any case I have not touched the examples so any changes cannot be related to this PR. It is possible that the tests on devel and the tests on this branch were done with different numba versions though. The devel tests are from 16.04.2026

EmilyBourne:
> In any case I have not touched the examples so any changes cannot be related to this PR. It is possible that the tests on devel and the tests on this branch were done with different numba versions though. The devel tests are from 16.04.2026

Actually we can see this in the diff. The only change is on a patch version

pyccel/codegen/printing/ccode.py
```diff
@@ -2219,13 +2219,32 @@ def _print_IndexedElement(self, expr):
...
                             allow_negative_indexes,
                         )
-                    indices.append("{" + self._print(idx) + "}")
-                indices_code = ", ".join(indices)
+                    indices.append(self._print(idx))
                 base_code = self._print(ObjectAddress(base))
-                return f"cspan_slice({base_code}, {c_type}, {indices_code})"
+                if (
+                    expr.rank < 4
+                    and base.rank < 5
```
yguclu, line 2228: Could you please add a comment line here? E.g.
```suggestion
                    # cspan_submd{N} exists only for N = 2, 3, 4,
                    # where N is the rank of the array to be sliced
                    expr.rank < 4
                    and base.rank < 5
```

pyccel/parser/semantic.py
```diff
@@ -4265,7 +4265,25 @@ def _visit_IndexedElement(self, expr):
...
             args = [self._visit(idx) for idx in expr.indices]
 
-            if len(args) == 1 and isinstance(
+            try:
+                in_lhs_assign = (
+                    isinstance(expr.current_user_node, Assign)
+                    and expr.current_user_node.lhs is expr
+                )
+            except AssertionError:
+                in_lhs_assign = False
```
yguclu, line 4276: Where do you expect an `AssertionError` to be raised?

pyccel/parser/semantic.py
```diff
@@ -4265,7 +4265,25 @@ def _visit_IndexedElement(self, expr):
...
+
+            if not in_lhs_assign and all(
+                not isinstance(class_type, HomogeneousListType)
+                and isinstance(a, Slice)
+                and a.start is None
+                and a.stop is None
+                and a.step is None
+                for a in args
+            ):
+                return var
```
yguclu, line 4288: Could you please add a comment line here, too? E.g.

```suggestion
            ):
                # Replace a[:] with a to avoid printing
                # unnecessary slicing in generated code
                return var
```

yguclu (changes requested):
Looks good to me! I have just a couple of minor comments

pyccel/parser/semantic.py
EmilyBourne: When there is more than 1 parent node. This can happen when handling more complex expressions. I had this problem in the tests

yguclu: I see, thanks! I would add a comment line here, too:
```suggestion
            # The property `current_user_node` of `PyccelAstNode`
            # works for objects with only one user node
            except AssertionError:
                in_lhs_assign = False
```

yguclu (approved):

PR #2625, opened 2026-07-08
POC: Reduce CI execution time

EmilyBourne:
When using Pyccel, the compilation is the main bottleneck. For most languages, compiling 1 file containing 10 functions is much faster than compiling 10 files each containing 1 function. In Pyccel's CI we take the latter approach so that a failure in 1 test doesn't crash all tests in the file.

This PR presents a proof-of-concept to get the best of both worlds. A new function `epyccel_module_with_fallback` is added which attempts to translate a module. If the translation succeeds then the functions from the translated module are used. If the translation fails then an instance of `LazyPerFunctionEpyccel` is created. This class allows functions to be translated individually (and cached) as and when they are used. This way, if all functions succeed then the tests should run much faster, and otherwise we fall back to the previous situation allowing us to easily identify exactly which test failed.

The file `tests/epyccel/test_epyccel_lists.py` is used to test this. All functions that are tested in C, Fortran and Python are moved to `tests/epyccel/modules/lists.py`. These functions are then accessed via a fixture `epyc_lists_mod` which uses `epyccel_module_with_fallback`.

On devel the command `time pytest test_epyccel_lists.py` shows:
```
real	1m17.614s
user	1m6.743s
sys	0m11.981s
```
on this branch the same command shows:
```
real	0m13.114s
user	0m12.384s
sys	0m1.793s
```

The CHANGELOG is not modified. This is left for a new PR which applies the proof-of-concept to more files in `tests/epyccel`. This second PR should speed up the CI. This will be critical to add new language support (in particular to unstick #2537)

---

## Pull request checklist

- [x] Update CHANGELOG.md

### DIFF
--- a/tests/epyccel/utilities.py
+++ b/tests/epyccel/utilities.py
@@ -42,7 +42,6 @@ class LazyPerFunctionEpyccel:
 
     def __getattr__(self, name):
         if name not in self._cache:
-            print("Translating function")
             self._cache[name] = epyccel(
                 getattr(self._pymod, name), language=self._language
             )
--- a/tests/epyccel/utilities.py
+++ b/tests/epyccel/utilities.py
@@ -25,7 +25,7 @@ class epyccel_test:
 
 
 # ==============================================================================
-class _LazyPerFunctionEpyccel:
+class LazyPerFunctionEpyccel:
     """
     Fallback proxy used when a whole-module `epyccel()` translation fails.
 
@@ -61,4 +61,4 @@ def epyccel_module_with_fallback(pymod, language):
     try:
         return epyccel(pymod, language=language)
     except Exception:
-        return _LazyPerFunctionEpyccel(pymod, language)
+        return LazyPerFunctionEpyccel(pymod, language)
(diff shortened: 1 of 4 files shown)


tests/epyccel/utilities.py
```diff
@@ -22,3 +23,43 @@ def compare_epyccel(self, *args):
...
+    """
+
+    def __init__(self, pymod, language):
+        self._pymod = pymod
+        self._language = language
+        self._cache = {}
+
+    def __getattr__(self, name):
+        if name not in self._cache:
+            print("Translating function")
```
yguclu: Did you want to also print the name of the function being translated?

EmilyBourne: Ah good idea. Otherwise this could be removed. It is only visible with -s. It was a debugging print originally

EmilyBourne: Removed: 3a6720aae

yguclu (approved):
This approach sounds very useful! I look forward to its extensive use across our epyccel unit tests

PR #2627, opened 2026-07-14
Speed up CI by grouping translations

EmilyBourne:
Use the `epyccel_module_with_fallback` function to group translations in `tests/epyccel`. For each `epyccel/X/test_Y.py` containing functions that are translated individually a file `epyccel/X/modules/Y.py` is created. Functions are moved to this file so `epyccel_module_with_fallback` can be used to accelerate the tests. xfailing or skipped tests are left in the original test file (or in some cases moved from an existing module back to the test file) so module files can be translated in as many cases as possible. The tool `epyccel_test` is removed, the tool `compare_epyccel` is modified to take pre-compiled functions instead of functions to be pyccelised.

---

## Pull request checklist

Please tick off items when you have completed them or determined that they are not necessary for this pull request:

- [x] Write a clear PR description
- [x] Ensure you appear in the AUTHORS file
- [x] Add tests to check your code works as expected
- [x] Update documentation if necessary
- [x] Update CHANGELOG.md
- [x] Ensure any relevant issues are linked
- [x] Ensure new tests are passing

PR #2628, opened 2026-07-14
Clean up test utilities

EmilyBourne:
Several functionalities are used across multiple tests. This leads to code duplication and an associated maintenance burden. This is sometimes avoided by importing from local files, but this causes its own problems as the tests are not a module with `__init__.py` files. To fix this a folder `tests/test_utils/` is added to group such utility functions, classes and variables.

**Commit Summary**
- Add the `tests/test_utils/` folder
- Add `tests/test_utils/` to the `$PYTHONPATH` using `pytest.ini`
- Move `ATOL` and `RTOL` tolerances to `tests/test_utils/tolerances.py` to unify values across files
- Move `min_<dtype>` and `max_<dtype>` to `tests/test_utils/tolerances.py`
- Rename `tests/epyccel/utilities.py` ->`tests/test_utils/epyccel_utilities.py`
- Move `matching_types` to `tests/test_utils/epyccel_utilities.py`
- Remove `randint` from `test_numpy_func.py` this was added to handle unexpected integer types in NumPy<v2 and is no longer needed. We can simply specify `dtype=int` when calling the normal NumPy `randint` function.
- Use `min_<dtype>` and `max_<dtype>` variables in `test_arrays.py` (continue to use in other files, but now use the version from `tests/test_utils/tolerances.py`).
- Move `compare_epyccel` function to `tests/test_utils/epyccel_utilities.py` (to be removed in #2627)

---

## Pull request checklist

- [x] Update CHANGELOG.md


tests/test_utils/tolerances.py
```diff
@@ -0,0 +1,43 @@
...
+
+min_int16 = iinfo("int16").min
+max_int16 = iinfo("int16").max
+
+# Use int32 for Windows compatibility
+min_int = iinfo(np.int32).min
+max_int = iinfo(np.int32).max
+
+min_int32 = iinfo("int32").min
+max_int32 = iinfo("int32").max
```
yguclu: Could you please remind me of the difference between `np.iinfo("int32")` and `np.iinfo(np.int32)`?

tests/test_utils/epyccel_utilities.py
```diff
@@ -35,21 +35,31 @@ class LazyPerFunctionEpyccel:
...
+                **self._epyccel_kwargs,
             )
         return self._cache[name]
 
+    @property
+    def language(self):
+        """
+        The language we are translating to.
+        """
+        return self._language
```
yguclu, line 58: It should be possible to specify the type here

```suggestion
    @property
    def language(self):
        """
        str: The language we are translating to.
        """
        return self._language
```

tests/epyccel/recognised_functions/test_numpy_types.py
```diff
@@ -3,13 +3,13 @@
 
 import numpy as np
 import pytest
-from numpy.random import uniform
-from test_numpy_funcs import (
+from epyccel_utilities import matching_types
+from numpy.random import randint, uniform
```
yguclu: Shouldn't importing from `numpy.random` come before the one from `epyccel_utilities`? I suppose the linter thinks that `epyccel_utilities` is a third-party package

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -5,38 +5,34 @@
 
 import numpy as np
 import pytest
-from numpy import finfo, iinfo, isclose
-from numpy.random import rand, randn, uniform
+from epyccel_utilities import matching_types
+from numpy import isclose
+from numpy.random import rand, randint, randn, uniform
```
yguclu: Here, too, I believe that importing from `epyccel_utilities` should come after importing from `numpy`
```suggestion
from numpy import isclose
from numpy.random import rand, randint, randn, uniform
from epyccel_utilities import matching_types
```

tests/epyccel/recognised_functions/test_numpy_funcs.py
```diff
@@ -2074,7 +2040,7 @@ def create_full_val_auto(val: T):
         a = full(3, val)
         return a[0]
 
-    integer32 = randint(low=min_int32, high=max_int32, dtype=np.int32)
+    integer32 = randint(low=min_int32, high=max_int32)
     integer = int(integer32)
```
yguclu, line 2045: Why this change? Is the conversion from a 32-bit integer to a 64-bit one problematic on the machines where the Python `int` is indeed 64-bit?

tests/epyccel/test_arrays.py
```diff
@@ -4860,41 +4874,25 @@ def test_dtype_conversion_to_bool_from_other_types(language):
...
-    integer16 = randint(
-        low=iinfo("int16").min, high=iinfo("int16").max, size=size, dtype=np.int16
-    )
-    integer32 = randint(
-        low=iinfo("int32").min, high=iinfo("int32").max, size=size, dtype=np.int32
-    )
-    integer64 = randint(
-        low=iinfo("int64").min, high=iinfo("int64").max, size=size, dtype=np.int64
-    )
+    integer = np.array(randint(low=min_int32, high=max_int32, size=size), dtype=int)
```
yguclu, line 4878: Is this not equivalent?

```suggestion
    integer = randint(low=min_int32, high=max_int32, size=size, dtype=int)
```

tests/epyccel/test_epyccel_mod.py
```diff
@@ -1,23 +1,17 @@
...
 # Relative and absolute tolerances for array comparisons in the form
-# numpy.isclose(a, b, rtol, atol). Windows has larger round-off errors.
-if sys.platform == "win32":
-    RTOL = 1e-13
-    ATOL = 1e-14
-elif os.environ.get("PYCCEL_DEFAULT_COMPILER", "GNU") == "intel":
+# numpy.isclose(a, b, rtol, atol). Intel seems to use a different algorithm
+if os.environ.get("PYCCEL_DEFAULT_COMPILER", "GNU") == "intel":
     RTOL = 1e-10
     ATOL = 1e-10
```
yguclu, line 15: Not related to this PR, but I prefer to ask here rather than forgetting..

I cannot remember why such a large tolerance is needed on Intel. Maybe it uses a "fast-math" algorithm for the modulo operation? Is this problem language-specific, or is it the same for C, C++, and Fortran?

tests/symbolic/test_symbolic.py
```diff
@@ -8,16 +8,14 @@
...
 
 from pyccel import lambdify as pyc_lambdify
 from pyccel.codegen.codegen import Codegen
 from pyccel.errors.errors import Errors
 from pyccel.naming import name_clash_checkers
 from pyccel.parser.parser import Parser
 
-RTOL = 1e-13
-ATOL = 1e-14
-
```
yguclu: I find it a bit strange to import `RTOL` (=2e-14) and `ATOL` (=1e-15) and then multiply them by a factor of 100 everywhere they are used. Are the tolerances originally used in this file too small?

tests/test_utils/epyccel_utilities.py
```diff
@@ -59,6 +69,29 @@ def epyccel_module_with_fallback(pymod, language):
...
+
+# ==============================================================================
+def compare_epyccel(f, language, *args):
+    """
+    Pyccelize `f`, call both versions with `args`, and assert the outputs match.
+    """
+    f2 = epyccel(f, language=language)
+    out1 = f(*args)
+    out2 = f2(*args)
+    assert all(r1 == r2 for r1, r2 in zip(out1, out2))
```
yguclu: This is slightly different than the original `compare_epyccel` function. If `f` returns a NumPy array among its outputs, `r1 == r2` will return a NumPy array of `bool`; is this the intended behavior of this function?

tests/epyccel/test_multiple_results.py
```diff
@@ -38,4 +31,4 @@ def test_expr_float_int_bool(language):
     def expr_complex_int_bool(n: "int"):
         return 0.5 + n * 1j, 2 * n, n == 3
 
-    compare_epyccel(expr_complex_int_bool, 3, language=language)
+    compare_epyccel(expr_complex_int_bool, language, 3)
```
yguclu, line 34: Good catch! I had not realized that we had several different definitions of the `compare_epyccel` function, with slightly different signatures 😅

yguclu (changes requested):
Good job! I have just a few comments and questions.

In the PR description we should mention that the file `tests/epyccel/utilities.py` has been renamed to `tests/test_utils/epyccel_utilities.py`. You write that the functions `matching_types` and `compare_epyccel` have been moved to that file, but not where the file originates from.

We should also mention the variables `min_<dtype>` and `max_<dtype>` are used across the `epyccel` tests, not just in `test_arrays.py`

tests/test_utils/tolerances.py
EmilyBourne: I'm not sure there is one. I just copied the syntax already used elsewhere. I agree that we should standardise, but I don't have a preference between the 2 if you do?

tests/epyccel/recognised_functions/test_numpy_types.py
EmilyBourne: Yes probably. In order to change this though we would have to provide extra information to pylint and isort, otherwise the CI won't pass after rearranging. I can do that in this PR or in a dedicated PR as you prefer?

tests/epyccel/recognised_functions/test_numpy_funcs.py
EmilyBourne: See above, can't reorder without triggering the CI

tests/epyccel/test_epyccel_mod.py
EmilyBourne: I'm not sure we ever investigated this, we just assumed that it must be using a different algorithm. From memory, the issue was the same for C and Fortran, C++ support hasn't been added for this file yet (maybe it is in #2537)

tests/symbolic/test_symbolic.py
EmilyBourne: The tolerances originally used are fine but I thought it was better to use the same tolerances throughout the tests as they are slightly arbitrary. Originally I multiplied by 100 because there are 100 values, but looking more closely these tests don't sum values, so there is no reason for this. The question then is why this 1 test needs higher tolerances than everything else?

tests/test_utils/tolerances.py
yguclu: If that works, I prefer `np.int32` for consistency with the other types

tests/epyccel/recognised_functions/test_numpy_types.py
yguclu: Are you sure that the CI would not pass? `numpy` is definitely a third-party package, after all.

Ah sorry, the problem is that we should print the imports in alphabetical order

EmilyBourne: Yes. If I change the order, isort changes it back

EmilyBourne: For isort the fix is: https://isort.readthedocs.io/en/latest/configuration/options.html#known-local-folder

tests/test_utils/tolerances.py
EmilyBourne: Fixed : 24545b18e

tests/epyccel/recognised_functions/test_numpy_types.py
yguclu: If a one-liner works, I would apply it to this PR

tests/epyccel/test_epyccel_mod.py
yguclu: All right, thanks! Could you create a short issue about this?

tests/epyccel/test_arrays.py
EmilyBourne: It is equivalent on my machine. But I think that on windows they are not equivalent

tests/epyccel/recognised_functions/test_numpy_types.py
EmilyBourne: It is a 1 line fix but isort will probably then modify several other files. It is probably worth doing here anyway though

tests/symbolic/test_symbolic.py
yguclu: I guess that the symbolic expressions of the mappings are quite complicated, and there could be numerical cancellation. E.g. the factor `(1-x1)` in `PolarMapping` should cause that when `x1` is close to `1`

EmilyBourne: Could be. What do you recommend for this file?

tests/epyccel/test_arrays.py
EmilyBourne: I have just run a test on the CI:
https://github.com/pyccel/pyccel/blob/741b2d03a97f0feba34b409d651d16f20002447f/.github/workflows/windows_test.yml#L27-L31
The output is:
```
int32
int64
```
which confirms my suspicion. I think `randint` was one of the only functions that wasn't moved over to int64 with NumPy 2.0

yguclu: That is worrying! Luckily we do not use Windows ourselves 😅

tests/symbolic/test_symbolic.py
yguclu: In order to make things clear and explicit, I would leave the original local definitions of `RTOL` and `ATOL` in this file. Those values seemed to work fine, and have nothing to do with the global values in `tolerances.py` anyway

tests/epyccel/recognised_functions/test_numpy_funcs.py
EmilyBourne: This was an accident when I was going through every `randint` in this file. Fixed : 382d99e6ff4e2cb3bc176c930c6491ed2cab2e06

tests/epyccel/recognised_functions/test_numpy_types.py
EmilyBourne: Fix applied: 8866619529f5167fd860bc3d03e444e2fb8cb23a
Isort applied: c6663cd08db1d21db05bb454c1d749facc913825

tests/epyccel/test_epyccel_mod.py
EmilyBourne: https://github.com/pyccel/pyccel/issues/2630

tests/symbolic/test_symbolic.py
EmilyBourne: Done : 620612953

tests/test_utils/epyccel_utilities.py
EmilyBourne: This should generalise : 652cc831f

tests/epyccel/test_epyccel_mod.py
yguclu: Thanks!

tests/test_utils/epyccel_utilities.py
yguclu: Almost! [`array_equal`](https://numpy.org/devdocs/reference/generated/numpy.array_equal.html) should be the correct function for this purpose

EmilyBourne: But the results are not arrays in all the files

EmilyBourne: This particular variant was copied from `tests/epyccel/test_multiple_results.py` where the results are not usually arrays

EmilyBourne: Never mind. A quick test shows that array_equal accepts scalars. Fixed : 1f065ceeb

yguclu: Just be careful: `array_equal` returns a `bool`, which does not have an `all()` method

EmilyBourne: Good catch. Fixed : a4c6f58424c116265b24777fd87799664c1ae1c6

yguclu:
The one remaining task is to update the PR description as I suggested in my first review comment

EmilyBourne:
> The one remaining task is to update the PR description as I suggested in my first review comment

Sorry, I forgot about this. Done

yguclu (approved):
LGTM!

